PR #1203, opened 2016-04-05
mothur recipe: pin boost version and add blast-legacy dependency

shiltemann:


bgruening:
You need to increase the build number, I guess.
@shiltemann why is this change needed? Does it not work with later versions? Can you use <1.60.0?

shiltemann:
yes it fails for me without this, "cannot find libboost_1.57.0.so" (off top of my head). travis in mothur branch of iuc is out of date i think, it's not running the tests at least, we could fix that first and see if we get the same error there before we merge if you like? 

is using < better than ==? (just curious)

shiltemann:
it seems to look specifically for version 1.57.0 judging fron the error message I get, so maybe == is needed? (i.e.  does conda install the highest version below 1.60 if you specify with < ? or if you already have say 1.56 installed will it just use that? )

bgruening:
@johanneskoester if the original version of mothur was compiled against 1.57.0 and now there is a new version of boost 1.60.0 this will cause this error. Am I right? What is the best way to solve this from day one, not after we encounter an error.

bgruening:
@yhoogstrate and me had an error today that indicates that == can produce wrong environments and we figured out that `- boost 1.57.0` should also work.

shiltemann:
ah, good to know :)

shiltemann:
yeah, I guess I could probably change this to use 1.60.0 as well then, because mothur will get rebuilt with it? ..and should maybe just fix all dependencies to current versions just to be safe then (unless theres a better mechanism to avoid things like this)

johanneskoester:
It is possible to use jinja templating in recipes. This can be used to pin the boost version to one set via an environment variable. I have updated our build script to allow that by setting the environment variable ``CONDA_BOOST``, see PR #1207.

shiltemann:
ah, awesome, thanks :)

johanneskoester:
I suggest to wait with this PR until I have merged PR #1207.

johanneskoester:
It's merged!

johanneskoester:
Do we need the exact blast-legacy dependency?

shiltemann:
@johanneskoester I did this for the sake of reproducibility. A different version of blast could potentially lead to different outputs from some of the mothur tools, and to me it would be nice if we can count on the fact that if we install this build number of mothur today we will get the exact same outputs from the tool as we would when we install it a month from now, regardless of whether somebody has created a new version of blast in the meantime..

johanneskoester:
I hear you, but this has to be handled on a different level. A recipe shall give the minimal constraints to be able to successfully build and run the package. The way to model reproducibility like you would like to have is to use environment.yml files. There you can fix a particular version of blast and all other stuff you think is critical. There is even a command that allows you to do this easily: `conda env export`, which will output a yaml description of all installed packages with their versions. Storing this together with your analysis will allow you to reproduce exactly the same results on any other system.

johanneskoester:
The main reason for this: imagine you have a different package, that also wants to have a particular version of a dependency. Although it would be technically possible, it is suddenly not any more because the authors of the packages decided to constrain them this way, rather than let the user decide which versions he wants to use.

shiltemann:
sure, I'm still new to conda, so thanks for explaining, this sounds very reasonable.

@bgruening how would we go about doing this in Galaxy tools? can you specify this in some way? i.e. "install mothur version X with blast dependency version Y"? ..normally I would say have blast as a separate dependency, but mothur doesn't look on the PATH for blast, it expects it in a particular location in one of its subfolders :/ ..any ideas on the best way to handle this?

johanneskoester:
Oh, if mothur does that, I would say the best way would be to patch it. Sounds like a simple ``sed`` invocation could fix it.

johanneskoester:
Another possibility would be to put relay scripts into `blast/bin`, which invoke blast as e.g.
```
blastall $@
```

This way, the e.g. the blastall in `$PATH` will be called.

shiltemann:
ah, good idea, thanks! :)

bgruening:
@shiltemann yes. We will solve this problem on the Galaxy side. We are discussion at the moment different possibilities. One is to ship a requirements.txt file which specifies the entire environment used by a tool. So every detailed version.

shiltemann:
ah, perfect :)

johanneskoester:
Is this ready to merge?

shiltemann:
yes, this is good to go, updating the branch now

PR #1207, opened 2016-04-05
Refactor dependency version pinning, include boost.

johanneskoester:
This allows to pin the boost version in recipes using jinja2 templating. E.g., you can write

```
package:
  name: mypackage
  version: 0.1

build:
  number: 0
  string: boost{{CONDA_BOOST}}_{{PKG_BUILDNUM}}

requirements:
  build:
    - boost >=1.55,{{CONDA_BOOST}}*

  run:
    - boost {{CONDA_BOOST}}*
```

### DIFF
--- a/scripts/build-packages.py
+++ b/scripts/build-packages.py
@@ -93,7 +93,7 @@ def filter_recipes(recipes):
     msgs = lambda py: [
         msg for msg in
         sp.run(
-            ["conda", "build", "--numpy", CONDA_NPY, "--python", py,
+            ["conda", "build", "--python", py,
              "--skip-existing", "--output"] + recipes,
             check=True, stdout=sp.PIPE, universal_newlines=True
         ).stdout.split("\n")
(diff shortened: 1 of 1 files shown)


bgruening:
Like it! Is there any standard we can follow? Wondering what other projects are doing.

johanneskoester:
The way I posted above is what the Continuum guys favor. It mimics the way Python and Numpy pinning with x.x is hardcoded. At some point, we need to decide which other libs should be pinned.

bgruening:
Great! Thanks a lot!

acaprez:
Hmm, so the Travis build for `master` I started a few minutes ago failed with `NameError: name 'CONDA_NPY' is not defined`.  I wonder if that's related to the changes here in `build-packages.py`?

johanneskoester:
Yes it is. I hope to fix it with PR #1211. Sorry for that :-)!

acaprez:
No problem - thanks!

PR #1212, opened 2016-04-06
Add Trf 4.07b binary

remimarenco:


remimarenco:
@bgruening If you can check that what I am doing is good that would be awesome! And then I will merge (or you can do it if you want to ;) )

recipes/trf/meta.yaml
```diff
@@ -0,0 +1,19 @@
...
+  name: trf
+  version: "4.07b"
+
+source:
+  fn: trf407b
+  url: https://tandem.bu.edu/trf/downloads/trf407b.linux64 # [linux]
+  url: https://tandem.bu.edu/trf/downloads/trf407b.macos64 # [osx]
+
+build:
+  number: 1
```
bgruening: Build number can/should be 0 in a first iteration.

bgruening: license and a description is missing

recipes/trf/meta.yaml
```diff
@@ -0,0 +1,19 @@
+package:
+  name: trf
+  version: "4.07b"
+
+source:
+  fn: trf407b
+  url: https://tandem.bu.edu/trf/downloads/trf407b.linux64 # [linux]
+  url: https://tandem.bu.edu/trf/downloads/trf407b.macos64 # [osx]
```
bgruening: In general we recommend to compile from source, as you never know if these binaries are relocatable and will work on old Linux machines.
But if this is to complicated I'm fine with this, I have seen this before.

recipes/trf/meta.yaml
```diff
@@ -0,0 +1,19 @@
+package:
+  name: trf
+  version: "4.07b"
+
+source:
+  fn: trf407b
+  url: https://tandem.bu.edu/trf/downloads/trf407b.linux64 # [linux]
```
bgruening: Btw. there is a [linux64] trick, just not sure how it is called. But linux is also linux-32

recipes/trf/meta.yaml
remimarenco: Well...I wish I could take the sources! But did not find them :/ https://tandem.bu.edu/trf/downloads/

bgruening: Than it's ok ... maybe you can ask them for the source and enhance this recipe in build-1 if you get the source.

recipes/trf/meta.yaml
```diff
@@ -0,0 +1,19 @@
...
+build:
+  number: 0
+  skip: False
+
+test:
+  files:
+    - test-data/dbia3.fasta
+  commands:
+    - trf 2>&1 | grep Tandem > /dev/null
+    - trf dbia3.fasta 2 7 7 80 10 50 500 2>&1 | grep "Tandem" > /dev/null
```
bgruening: I like this test data example! Well done!

recipes/trf/meta.yaml
remimarenco: Asked!

jxtx: Very unlikely to get the source. TRF is not under an open license. 

There are open alternatives – phobos, mreps, maybe VNTRseek. Would one of those work instead?

remimarenco: Will see with folks of my project if they are useful. Could be nice to add them anyway here! Thanks @jxtx

PR #1213, opened 2016-04-06
multiqc: ensure all files are world readible in install since multiqc…

chapmanb:
… copies resources to final output directory

### DIFF
--- a/recipes/multiqc/build.sh
+++ b/recipes/multiqc/build.sh
@@ -1,3 +1,4 @@
 #!/bin/bash
 
 $PYTHON setup.py install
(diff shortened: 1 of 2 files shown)


mbargull:
Hi @chapmanb, I don't quite understand why the `chmod` is needed. The `setup.py` doesn't look to being doing anything special. Could you elaborate?

chapmanb:
I don't remember the full context but multiqc reads some resource files that get copied over into the python install. If multiqc got installed by root and then run as a regular user it wouldn't have permissions to read/copy them and failed. This ensures that these are all user readable to avoid problems at runtime.

I'm not sure if this has been addressed in the meantime but that was the motivation.

PR #1225, opened 2016-04-06
Change CNVkit license to Apache 2.0

etal:
The upstream CNVkit license is Apache 2.0, see: https://github.com/etal/cnvkit/blob/master/LICENSE

### DIFF
--- a/recipes/cnvkit/meta.yaml
+++ b/recipes/cnvkit/meta.yaml
@@ -37,5 +37,5 @@ test:
 
 about:
   home: https://github.com/etal/cnvkit
-  license: MIT
+  license: Apache License 2.0
   summary: Copy number variant detection from targeted DNA sequencing


kyleabeauchamp:
Oops, sorry about that.  If you want this change to overwrite the uploaded packages, you will need to increment the build number from 0 to 1.

kyleabeauchamp:
AFAIK the CI setup only uploads packages if either the version number is increased or the build number of increased.

kyleabeauchamp:
See #1226

kyleabeauchamp:
I should clarify: incrementing the build number doesn't actually over-write the package, but it uploads a new package with a larger build number.  Then, whenever a user does `conda install xyz`, the latest build number will be downloaded and installed.  

This means that to actually *delete* a build requires pinging Johannes.

etal:
It doesn't matter much, they're both permissive licenses.  Will this change to the metadata show up for future releases, then?

johanneskoester:
Yes.

PR #1257, opened 2016-04-09
abyss: Remove google-sparsehash from run

sjackman:
google-sparsehash is a build dependency only.


bgruening:
@sjackman thanks @sjackman!

sjackman:
As a heads up, the next release will remove `sqlite` as a required dependency. It's already committed to `master`.

PR #1264, opened 2016-04-11
WIP: CrossMap pysam fix

yhoogstrate:
CrossMap is incompatible with pysam(.csamtools) 0.9.* and will terminate abnormally with the following error:
```
pysam.utils.SamtoolsError: 'samtools returned with error 1: stdout=, stderr=[bam_sort] Use -T PREFIX / -o FILE to specify temporary and final output files
Usage: samtools sort [options...] [in.bam]
Options:
  -l INT     Set compression level, from 0 (uncompressed) to 9 (best)
  -m INT     Set maximum memory per thread; suffix K/M/G recognized [768M]
  -n         Sort by read name
  -o FILE    Write final output to FILE rather than standard output
  -T PREFIX  Write temporary files to PREFIX.nnnn.bam
  -@, --threads INT
             Set number of sorting and compression threads [1]
      --input-fmt-option OPT[=VAL]
               Specify a single input file format option in the form
               of OPTION or OPTION=VALUE
  -O, --output-fmt FORMAT[,OPT[=VAL]]...
               Specify output format (SAM, BAM, CRAM)
      --output-fmt-option OPT[=VAL]
               Specify a single output file format option in the form
               of OPTION or OPTION=VALUE
      --reference FILE
               Reference sequence FASTA FILE [null]
```

CrossMap is incompatible with pysam(.csamtools) 0.7.7 and 0.8.3 and will terminate abnormally with the following error without further information:
```
[1]    57033 abort (core dumped) 
```

CrossMap (using bam files, not using sam files) is compatible with pysam v0.6 and will terminate abnormally with the following error:
```
@ 2016-04-11 14:03:55: Sort "/tmp/tmpwM938m/files/000/dataset_3.dat.bam" ...
[bam_header_read] bgzf_check_EOF: Invalid argument
/tmp/tmpwM938m/job_working_directory/000/3/tool_script.sh: line 9: 51929 Segmentation fault      (core dumped) 
```

Using 0.6 is an improvement compared to using 0.9.0, but it still doesn't solve `CrossMap.py bam *.chain *.bam out.bam`, it only solves using a sam file.


yhoogstrate:
Hey @chapmanb I saw that you have looked into CrossMap and it's dependencies. Could you share your opinion on this PR before I merge?

Thanks in advance,

Youri

bgruening:
Looks fine for me. Maybe you can add a comment to the recipe so this get not lost.

chapmanb:
Youri;
Have you tried raising an issue about this with the CrossMap team? It might be possible to get an updated version of CrossMap that works with bam and doesn't require pinning pysam to an old version. With all the incompatibilities between versions I'm a bit worried that having an old 0.6 pysam in the recipe would cause problems with any tools installed alongside CrossMap that require recent versions.

yhoogstrate:
Hey @chapmanb, I don't think it's fair to expect from a package maintainer to be maintaining code to keep up to date with upstream dependencies that are being made backwards incompatible, especially if CrossMap itself ships with fixed version of pysam (0.7.5 https://github.com/yhoogstrate/bioconda-recipes/blob/c2e1555b5f519e46258f7ed1bed8330d05cae6d1/recipes/crossmap/build.sh#L5).

A rolling release model on major releases of dependencies is really a curse as many bioinformatics libraries suffer from backwards incompatibility, and will again in the future. Is there no way in Conda you can use per tool a different and fixed (major) version of a dependency (in this case pysam)?

That CrossMap gave an error with pysam 0.7.7 is a point I will ask to the maintainers.

johanneskoester:
If crossmap is incompatible with newer versions of pysam, it is fine to set a maximum (and minimum) version in the recipe. I still would try to encourage the devs to port. As far as I remember the API differences are not that big.
Technically, the solution to handle conflicting dependencies in conda is to create separate environments. There is even a command to invoke something in another environment without activating it first.

bgruening:
So I take @johanneskoester comment as `ready to merge` but encourage upstream devs to fix pysam deps.
I have written the author and will merge now.

yhoogstrate:
Thank you both @bgruening and @chapmanb :+1:

PR #1282, opened 2016-04-12
RMBlast recipe, based on NCBI-BLAST recipe

pvanheus:
RMBlast is a variant of BLAST used by RepeatMaster. It is available from NCBI and has a build process similar to NCBI BLAST+.

This recipe builds RMBlast from source on Linux rather than pulling the binary from the NCBI FTP site.


pvanheus:
The build failed due to timeout, since the BLAST source build takes a very long time.

recipes/rmblast/build.sh
```diff
@@ -0,0 +1,26 @@
...
+
+    cp $SRC_DIR/bin/rmblastn $PREFIX/bin
+
+elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+    echo "Platform: Linux"
+
+    cd $SRC_DIR/c++
+    ./configure --prefix=$PREFIX --with-mt --without-debug
+
+    make -j4
```
johanneskoester: I think we should remove the -j4 here. Travis CI VM only provides 1.5 cores (I know this comes from the original blast recipe).

recipes/rmblast/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    fn:  ncbi-rmblastn-2.2.28-src.tar.gz # [linux]
+    url: ftp://ftp.ncbi.nlm.nih.gov/blast/executables/rmblast/2.2.28/ncbi-rmblastn-2.2.28-src.tar.gz # [linux]
+    md5: fb5f4e2e02ffcb1b17af2e9f206c5c22 # [linux]
+    fn: ncbi-rmblastn-2.2.28-universal-macosx.tar.gz # [osx]
+    url: ftp://ftp.ncbi.nlm.nih.gov/blast/executables/rmblast/2.2.28/ncbi-rmblastn-2.2.28-universal-macosx.tar.gz # [osx]
+    md5: 421c5ddf2c425800d934562c965e8706 # [osx]
+
+build:
+  number: 0
+  skip: False
```
johanneskoester: Above line can be removed.

pvanheus:
It still fails to build in time. What is the reason to build from source rather than fetch binaries from NCBI?

bgruening:
We usually don't know if binaries in general can run on old HPC systems. bioconda is compiling against CentOS-5. This guarantees a certain HPC compatibility.

pvanheus:
Thanks @bgruening, then I don't know how to solve this build problem :(

bgruening:
@pvanheus not sure why this takes longer as the blast package. Given that the original blast package also takes binaries for osx, I would say it's ok to go for binaries as soon as we switch to a new build system.

pvanheus:
@bgruening what new build system?

BTW this recipe passes this test:

docker run -t --rm -v $(pwd)/bioconda-recipes:/bioconda-recipes bioconda/bioconda-builder:latest --package /bioconda-recipes/recipes/rmblast

johanneskoester:
We will switch to building on anaconda.org soon. I might have time to set that up in the next week. I'm sorry for the delay.

johanneskoester:
With the recent fixes to blast, I noticed that it hardly fits into the time limits. But there are times when travis is fast enough. You might have to restart the build several times until it happens to be within the limit.

johanneskoester:
Can you sync with master and try again?

PR #1284, opened 2016-04-12
UCSC utilities: ucsc-rmfadups

yhoogstrate:


johanneskoester:
Please don't be too fast with the merging. I want to avoid that the ucsc tools aggregate in the travis jobs. This would be the same effect as having one big PR. In the future, once we have migrated our build system to anaconda.org, this won't be an issue any more.

PR #1302, opened 2016-04-13
Using involucro

thriqon:


thriqon:
I'm going to merge this if it is green because I have to test if it's correctly working in this exact moment ;-)

thriqon:
Sorry, wrong upstream

PR #1313, opened 2016-04-13
enable py3k builds for biom-format

bgruening:


wasade:
:+1: once build passes

PR #1320, opened 2016-04-14
New recipe: meme

saketkc:


recipes/meme/build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+#
+MEME_ETC_DIR=${PREFIX}/etc
+cpanm YAML
+cpanm HTML::PullParser
```
bgruening: It would be better to use these modules as dependencies and not install them via `cpanm`.

saketkc: Thanks @bgruening, but it leads to many recursive perl dependencies some of them failing tests.

bgruening: I know :( but this would be the proper way.
I'm not stopping this PR, just letting you know what the best-practise is in bioconda.

saketkc: Sure @bgruening. I would want to modify the dependencies so they are packaged as well. However I am going to merge this for now and keep iterating.

PR #1325, opened 2016-04-15
Add recipe for Goatools

bebatut:


recipes/goatools/meta.yaml
```diff
@@ -0,0 +1,26 @@
...
+build:
+  number: 0
+  skip: False
+source:
+  fn: goatools-0.5.9.tar.gz
+  md5: 2d5f1f9319dbf59850bf92cfa71e92ee
+  url: https://pypi.python.org/packages/source/g/goatools/goatools-0.5.9.tar.gz
+requirements:
+  build:
+    - python 2.7*
```
bgruening: I guess a better way to do this is to just write `- python` here and above `skip: True # [py3k]`

bebatut: Ok I will test like this

recipes/goatools/meta.yaml
```diff
@@ -0,0 +1,26 @@
...
+  fn: goatools-0.5.9.tar.gz
+  md5: 2d5f1f9319dbf59850bf92cfa71e92ee
+  url: https://pypi.python.org/packages/source/g/goatools/goatools-0.5.9.tar.gz
+requirements:
+  build:
+    - python 
+    - setuptools
+    - fisher
+  run:
+    - python 2.7*
```
bgruening: Still one python 2.7 missing :)

bebatut: Oh, I missed it

recipes/goatools/meta.yaml
```diff
@@ -0,0 +1,26 @@
+about:
+    home: 'https://github.com/tanghaibao/goatools'
+    summary: "Python scripts to find enrichment of GO terms"
```
bgruening: license is BSD, sorry have not spotted this earlier. If you submit your PR from the main branch I can fix those for you.

bebatut: I had it :)

PR #1331, opened 2016-04-15
Adding rdp-tools

colinbrislawn:
Adding support for the RDP Tools.
This is my first contribution to this repo, so I would appreciate any feedback @bioconda/core 
Thank you all!

### DIFF
--- a/recipes/rdp-tools/meta.yaml
+++ b/recipes/rdp-tools/meta.yaml
@@ -10,6 +10,8 @@ source:
 requirements:
   build:
     - ant
+  run:
+    - java
 
 build:
   number: 0
--- a/recipes/rdp-tools/meta.yaml
+++ b/recipes/rdp-tools/meta.yaml
@@ -16,6 +16,12 @@ build:
 
 test:
   commands:
+    - java=java
+    - if [ -z "${JAVA_HOME:=}" ]; then
+    -   if [ -e "$JAVA_HOME/bin/java" ]; then
+    -     java="$JAVA_HOME/bin/java"
+    -   fi
+    - fi
     - java -Xmx1g -jar $PREFIX/bin/classifier.jar classify -c 0.5 -o usga_classified.txt -h soil_hier.txt $SRC_DIR/classifier/samplefiles/USGA_2_4_B_trimmed.fasta
 
 about:
--- a/recipes/rdp-tools/meta.yaml
+++ b/recipes/rdp-tools/meta.yaml
@@ -1,0 +1,37 @@
+
+package:
+  name: rdp-tools
+  version: "2.0.2"
+
+source:
+  git_url: https://github.com/rdpstaff/RDPTools.git
+  git_rev: 2.0.2
+
+requirements:
+  build:
+    - ant
+
+build:
+  number: 0
+  script:
+#    - brew update              # [osx]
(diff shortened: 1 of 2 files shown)


recipes/rdp-tools/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+#    - brew update              # [osx]
+#    - brew install ant         # [osx]
+
+    - git config --global url."https://".insteadOf git://
+    - git submodule init
+    - git submodule update
+
+    - make
+
+    - mkdir -p $PREFIX/bin
```
bgruening: This is usually handled in a build.sh file. Please have a look at similar recipes.

recipes/rdp-tools/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - ant
+
+build:
+  number: 0
+  script:
+#    - brew update              # [osx]
+#    - brew install ant         # [osx]
+
+    - git config --global url."https://".insteadOf git://
+    - git submodule init
```
bgruening: I don't think this is needd if you use git_url. Recommended is a tarball if there is one packed with all submodules included.

recipes/rdp-tools/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+    - make
+
+    - mkdir -p $PREFIX/bin
+    - mv *.jar $PREFIX/bin
+    - mv lib $PREFIX/bin
+
+test:
+  commands:
+    - java -Xmx1g -jar $PREFIX/bin/classifier.jar classify -c 0.5 -o usga_classified.txt -h soil_hier.txt $SRC_DIR/classifier/samplefiles/USGA_2_4_B_trimmed.fasta
```
bgruening: Java programs are shipped with a special wrapper that makes is more easily to call from the commandline line. Moreover, the $PREFIX is not available during runtime, afaik.
Please have a look at: https://github.com/bioconda/bioconda-recipes/tree/master/recipes/opsin/2.1.0 for a nice example.

bgruening:
@colinbrislawn thanks for your contribution!

recipes/rdp-tools/meta.yaml
colinbrislawn: Unfortunately there is not. In fact, this github repo refers to several other repos which include the dependencies. It's kind of a mess.

recipes/rdp-tools/meta.yaml
colinbrislawn: Unfortunately, RDP tools does not have this kind of wrapper. 😞  This is their example command: https://github.com/rdpstaff/classifier/blob/master/README#L84 

I'll see if $PREFIX works or if there is a suitable replacement.

recipes/rdp-tools/meta.yaml
colinbrislawn: Good catch!

colinbrislawn:
Works on OSX. On linux
`/anaconda/conda-bld/test-tmp_dir/run_test.sh: line 3: java: command not found`

Any ideas? Should I list java as a runtime dependency?

druvus:
@colinbrislawn It was recently added to bioconda as rdptools.

PR #1334, opened 2016-04-16
New: gsort and vcfanno from @brentp

chapmanb:

### DIFF
--- a/recipes/gsort/meta.yaml
+++ b/recipes/gsort/meta.yaml
@@ -1,0 +1,23 @@
+package:
+  name: gsort
+  version: '0.0.1'
+
+source:
+  fn: gsort_darwin_amd64 # [osx]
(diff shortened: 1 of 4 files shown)


recipes/gsort/meta.yaml
```diff
@@ -0,0 +1,23 @@
+package:
+  name: gsort
+  version: '0.0.1'
+
+source:
+  fn: gsort_darwin_amd64 # [osx]
+  url: https://github.com/brentp/gsort/releases/download/v0.0.1/gsort_darwin_amd64 # [osx]
```
bgruening: @chapmanb I thought we recommend to build from source? Is this not true anymore or does the build time exceed the travis limit?

chapmanb: These are pre-built static GO binaries so hopefully will work pretty cleanly across most systems. I'm personally a fan of using binaries where it makes sense so am not very strict about trying to build everything from source.

PR #1343, opened 2016-04-19
F/networkxgmml

xguse:

@bioconda/core 

NetworkX does not like the formats that cytoscape exports as.  Even when exported from cytoscape as graphml (a common format between them) networkx will not import correctly.  This library allows networkx to import graphs exported from cytoscape in the XGMML format.

This is a simple `conda skeleton pypi networkxgmml` followed by setting the meta.yaml tags: build.number / build.skip.


johanneskoester:
Thanks!

PR #1371, opened 2016-04-22
Pin boost

johanneskoester:

This PR does two things:

1. All (I hope) recipes depending on boost are changed such that the boost version is pinned, via a jinja2 mechanism (see GUIDELINES.md).
2. Our build system is modified, such that multiple recipes can be build in parallel jobs on travis CI. This is achieved by calculating the (i) DAG of recipes, (ii) searching for connected components (iii) distributing the connected components to separate Travis CI jobs (iv) toposorting and building each component.

Part two is very useful if many recipes are modified in a single PR (like here). Without the modification, the PR can easily exceed the build time limits on Travis.
The number of spawned jobs can be configured in the `travis.yml`.


recipes/blast/build.sh
```diff
@@ -18,7 +18,7 @@ elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
     cd $SRC_DIR/c++
     ./configure --prefix=$PREFIX
 
-    make -j4
+    make -j2
```
bgruening: We could also use `-j$(nproc)`

recipes/variantbam/meta.yaml
```diff
@@ -4,19 +4,24 @@ package:
 
 build:
   number: 0
-  skip: True # [osx]
+  skip: True
```
bgruening: skip: True looks wrong to me

bgruening:
Very nice improvements!

recipes/variantbam/meta.yaml
johanneskoester: The variantbam build fails with an error that seems unrelated to the pinning changes. I have added the skip in order to notify the author once this is merged.

recipes/blast/build.sh
johanneskoester: A plain `-j` would achieve the same I think. However, Travis provides 1.5 cores. I am not sure how this is reported via nproc.

johanneskoester:
Because building blast takes so long, I had to temporarily increase the number of spawned jobs to 8. I will revert this to 4 once the packages have been uploaded.

johanneskoester:
@bioconda/all please note above changes to our build system, and the new pinning mechanism.

bgruening:
Awesome! Thanks @johanneskoester! Any reason to not leave it to 8 cores. This https://github.com/bioconda/bioconda-recipes/pull/1282 can profit from it as well.

chapmanb:
Johannes -- awesome, thank you for the boost fix. I'll take a look at the variantbam problem and see if I can revive that recipe.

johanneskoester:
Thanks, Brad!

johanneskoester:
@bgruening the thing is that spawning 8 jobs per default means that we have to wait for idle jobs to complete in most cases, because the usual case is that one recipe is updated in a PR. The thing is that travis does not start all 8 jobs immediately. Hence, I thought 4 is a good compromise. We can however tune this from time to time, and first try 8 jobs for now.

PR #1387, opened 2016-04-25
Update Graphlan package

bebatut:


recipes/graphlan/meta.yaml
```diff
@@ -3,7 +3,7 @@ about:
     summary: "GraPhlAn is a software tool for producing high-quality circular representations of taxonomic and phylogenetic trees. It focuses on concise, integrative, informative, and publication-ready representations of phylogenetically- and taxonomically-driven investigation."
 package:
   name: graphlan
-  version: '1.0.0'
+  version: '0.9.7'
```
bgruening: Why is that? You are still downloading the `hg_tag: 1.0`.
The problem here is that conda will take the latest version by default. So `conda install graphlan` will install version 1.0 I guess.

bebatut: The version number is different in scripts. But I can maintain 1.0.0.
This update was more for `matplotlib`, to avoid a warning

PR #1391, opened 2016-04-25
samtools: updated to version 1.3.1

croth1:

### DIFF
--- a/recipes/samtools/1.3/build.sh
+++ b/recipes/samtools/1.3/build.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+sed -i.bak 's/^CPPFLAGS =$//g' Makefile
+sed -i.bak 's/^LDFLAGS  =$//g' Makefile
+
+sed -i.bak 's/^CPPFLAGS =$//g' htslib-1.3/Makefile
+sed -i.bak 's/^LDFLAGS  =$//g' htslib-1.3/Makefile
+
+export CPPFLAGS="-I$PREFIX/include"
+export LDFLAGS="-L$PREFIX/lib"
+
+make install prefix=$PREFIX
--- a/recipes/samtools/1.3/meta.yaml
+++ b/recipes/samtools/1.3/meta.yaml
@@ -1,0 +1,27 @@
+package:
+  name: samtools
+  version: "1.3"
+
+build:
+  number: 2
+  skip: False
+
+source:
+  fn: samtools-1.3.tar.bz2
+  url: https://github.com/samtools/samtools/releases/download/1.3/samtools-1.3.tar.bz2
+
+requirements:
+  build:
+  - ncurses
+  - zlib
+  run:
+  - ncurses
+  - zlib
+
+about:
+  home: https://github.com/samtools/samtools
+  license: MIT
+  summary: Tools for dealing with SAM, BAM and CRAM files
+test:
+  commands:
+    - samtools --help
--- a/recipes/samtools/build.sh
+++ b/recipes/samtools/build.sh
@@ -3,8 +3,10 @@
 sed -i.bak 's/^CPPFLAGS =$//g' Makefile
 sed -i.bak 's/^LDFLAGS  =$//g' Makefile
 
-sed -i.bak 's/^CPPFLAGS =$//g' htslib-1.3.1/Makefile
-sed -i.bak 's/^LDFLAGS  =$//g' htslib-1.3.1/Makefile
+sed -i.bak 's~s/@CURSES_LIB@/-lcurses/g~s|@CURSES_LIB@|-lcurses|g~g' Makefile
+sed -i.bak "s~-lcurses~$(ncurses${CONDA_NCURSES%.*}-config --libs)~g" Makefile
+sed -i.bak 's/^CPPFLAGS =$//g' htslib-$PKG_VERSION/Makefile
+sed -i.bak 's/^LDFLAGS  =$//g' htslib-$PKG_VERSION/Makefile
 
 export CPPFLAGS="-I$PREFIX/include"
 export LDFLAGS="-L$PREFIX/lib"
--- a/recipes/samtools/meta.yaml
+++ b/recipes/samtools/meta.yaml
@@ -1,21 +1,26 @@
+{% set version = "1.3.1" %}
 package:
   name: samtools
-  version: "1.3.1"
+  version: {{ version }}
 
 build:
-  number: 0
-  skip: False
+  number: 1
+  script_env:
+   - CONDA_NCURSES
 
 source:
-  fn: samtools-1.3.1.tar.bz2
-  url: https://github.com/samtools/samtools/releases/download/1.3.1/samtools-1.3.1.tar.bz2
+  fn: samtools-{{ version }}.tar.bz2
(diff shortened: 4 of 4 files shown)


croth1:
@johanneskoester, it seems that travis is becoming a bottleneck. It takes > 2h per CI run, which makes only around 10 checked commits per day.

johanneskoester:
I wonder if the ncurses dependency should also be pinned? I think an update to ncurses 6.0 would break the samtools build, right?

jakirkham:
This is a good point. I got a bug report about missing symbols, but if it works with 5.0 and not 6.0 that seems like something that needs pinning.

croth1:
@johanneskoester good catch! Updated the PR.

johanneskoester:
In fact, I'd rather like to pin the ncurses version via an environment variable, similar to boost (e.g. in the blast recipe). I will update the infrastructure to allow this ASAP (maybe tomorrow).

The reason is: I am quite sure that samtools will still build with ncurses 6.x. By using an environment variable, we can globally update ncurses for all depending recipes.

daler:
I needed a bug fix in 1.3.1 for a project, only saw 1.3 in bioconda, and updated to 1.3.1 in f7eae60df2eee  without checking open PRs or pinning ncurses.

I clearly need to stay on top of the PRs better, I apologize for missing this.

Should we remove the existing anaconda tarball and should I revert the recipe back to 1.3?

daler:
I also noticed that pysam is pinned to `samtools ==1.3` so that recipe may need updating.

croth1:
I updated the PR. In general things are tricky here. Without `sed -i.bak 's/-lcurses/-lncurses/g' Makefile`, samtools links against system's ncurses, and conda's ncurses is simply ignored. So this should be the right thing to do.

However this will bring problems for everyone using the conda-forge channel. It seems they'll run into https://github.com/conda-forge/ncurses-feedstock/issues/9. I tried to understand what's happening there, but the issue seems messy and I currently cannot think of the reason for the missing symbols there.

Maybe https://github.com/conda/conda/issues/2410 could work as a workaround. I guess we should defer this PR until this problem is solved.

jakirkham:
I really still think this is a pinning issue as @johanneskoester has mentioned. Missing symbols normally occur when you have something compiled against something with those symbols and you try to load them from a library that doesn't have them.

johanneskoester:
Indeed. Can you please pin ncurses as it is done in [this recipe](https://github.com/bioconda/bioconda-recipes/blob/16b7615c615ff78202eb08dab17bf1554c3c36b3/recipes/mothur/1.25.0/meta.yaml)? Thanks!

croth1:
@johanneskoester, done.
@jakirkham, pinning did not have an influence on the choice of package. Both conda-forge and r have ncurses-5, conda-forge has a higher build number, so conda's solver chooses this version. Apart from bioconda's devs now being able to update the version globally, nothing changed.

jakirkham:
So, I have no idea whether your issue, @croth1, is related to this issue ( https://github.com/conda-forge/ncurses-feedstock/issues/9 ) opened by @cel4, but we have all of those symbols ( `_cbreak`, `_keypad`, `_stdscr` ). They are just in `libtinfo`. Some people choose to lump those symbols in `libncurses`. We did not. I suppose we could change that. If you think that is very important, we can take that into consideration.

Still this is not an uncommon configuration for `ncurses` to be built with. So, it would not be unreasonable to get in touch with the devs maintaining this software and have them include `-ltinfo` in their build scripts. Though you could affect this behavior now using the appropriate environment variables.

jakirkham:
We may try to change `ncurses` so that `libncurses` includes all the symbols that would be in `libtinfo`. See this PR ( https://github.com/conda-forge/ncurses-feedstock/pull/10 ), which will do exactly that. We will still keep links for the old `libtinfo` libraries so they can pull their symbols from `libncurses` in the end.

croth1:
@jakirkham, thanks for figuring this out. Indeed linking to `libtinfo` as well does the trick. I have no strong opinion on whether conda-forge's ncurses should provide these symbols.

However, I find it tricky to rewrite my recipe so that it works in both cases. I would somehow check if `libtinfo` exists and then patch the linking flags accordingly.

I guess the easiest way for that is to use `pkg-config`. This however requires `--enable-pc-files` in conda-forge's configuration of ncurses.

Also conda's `pkg-config` recipe only seems to be packaged for linux for some reason. I can use the native one on OS X - not so nice, but would work.

jakirkham:
> thanks for figuring this out. Indeed linking to `libtinfo` as well does the trick. 

Glad to help. Also, glad to hear it worked for you. 😄

> I guess the easiest way for that is to use `pkg-config`. This however requires `--enable-pc-files` in conda-forge's configuration of ncurses.

Absolutely, we should already be doing this. This change should be non-controversial so I'll try to add a separate PR. Will update ( https://github.com/conda-forge/ncurses-feedstock/pull/10 ) respectively.

Even though combining the symbols into `libncurses` could be controversial, there is only one thing that depends on it now (`erlang`) and it isn't getting so much use yet. Our `python` package will depend on it, but it is not out yet. So, we should be able to make a change there.

> Also conda's `pkg-config` recipe only seems to be packaged for linux for some reason. I can use the native one on OS X - not so nice, but would work.

We have `pkg-config` in conda-forge and it works nicely on Mac and Linux. We make heavy use of it. Please give it a try if it would help.

croth1:
Oh, things are more complicated than I have expected. OS X does not ship pkg-config by default. Conda-forge has it, but since we are not using conda-forge at the moment, we would have to copy it here to actually use it.

I think at this point it would make sense to test if all packages are compatible to conda-forge and then add the conda forge channel to bioconda's default channels. Copying and maintaining general purpose packages inside bioconda will just bind energy.

I will defer the testing until conda-forge's version contains the pkg-config files.

croth1:
depends on: https://github.com/bioconda/bioconda-recipes/pull/1388

jakirkham:
Adding pkg-config files in this PR ( https://github.com/conda-forge/ncurses-feedstock/pull/11 ), @croth1.

croth1:
This is a version that uses `ncurses-config` instead of `pkg-config` to determine the libraries to link against. This is a little hacky, but it seems to work with ncurses from all channels.

jakirkham:
So, we have a PR ( https://github.com/conda-forge/ncurses-feedstock/pull/11 ) to add `pkg-config` files to the `ncurses` package. It seems to work fine. Still it would be nice to get some more feedback from people to make sure there isn't something I have missed and to make sure it works well for people here.

jakirkham:
A new version of `ncurses` has been released at `conda-forge` and it should contain `pkg-config` files. Please try it out and let us know how it works.

chapmanb:
Christian;
I incoporated your versioning and ncurses fixes into #1650. I had to do ncurses a different way -- by using the samtools configure script -- since curses is no longer in the base package. I am also not sure about include gcc/libgcc -- I think the current recommendation has moved back to not including them.

I hope this fixes what you needed here. If not, definitely happy to merge a PR with more fixes. Sorry for not merging this directly but hope that helps get you moving.

johanneskoester:
@chapmanb I think we haven't decided about gcc yet. In fact, it seems like at least Björn and me are not convinced by the strategy of conda-forge to not use gcc as a dependency. I would suggest to go on specifying gcc and libgcc in the bioconda packages until a final decision has been made.

PR #1392, opened 2016-04-25
julia: Update to 0.4.5

yesimon:

### DIFF
--- a/recipes/julia/build.sh
+++ b/recipes/julia/build.sh
@@ -1,3 +1,2 @@
 #!/bin/sh
-make
-env PREFIX=$PREFIX prefix=$PREFIX make install
+cp -r * $PREFIX
--- a/recipes/julia/build.sh
+++ b/recipes/julia/build.sh
@@ -1,2 +1,3 @@
 #!/bin/sh
-cp -r * $PREFIX
+mv * "$PREFIX"
+mv "$PREFIX/LICENSE.md" .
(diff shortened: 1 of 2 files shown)


recipes/julia/build.sh
```diff
@@ -1,3 +1,2 @@
 #!/bin/sh
-make
-env PREFIX=$PREFIX prefix=$PREFIX make install
+cp -r * $PREFIX
```
johanneskoester: Won't this also copy over LICENSE.txt from the package?

yesimon: Hmm how about this way. It seems like LICENSE.md is the only unusual file in the tar.

recipes/julia/build.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
-make
-env PREFIX=$PREFIX prefix=$PREFIX make install
+mv * "$PREFIX"
+mv "$PREFIX/LICENSE.md" .
```
johanneskoester: Why not rm?

yesimon: It seems like the build process checks for the presence of the license file after running the `build.sh` or else it fails.

johanneskoester: I see. This is because of the `license_file` statement in the meta.yaml. Seems to be necessary, because julia specifies some exceptions to MIT license.

johanneskoester:
Thanks!

PR #417, opened 2015-12-09
Add jellyfish to OSX whitelist

kyleabeauchamp:

PR #1400, opened 2016-04-26
add bioepic

daler:

### DIFF
--- a/recipes/bioepic/bld.bat
+++ b/recipes/bioepic/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
+:: http://docs.continuum.io/conda/build.html
+:: for a list of environment variables that are set during the build process.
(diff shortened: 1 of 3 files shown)


daler:
@johanneskoester are there any recent build changes that might cause [this error](https://travis-ci.org/bioconda/bioconda-recipes/jobs/125933117#L245)? py27 seems to build fine but then for py34/35 it seems to be looking for the very package it's trying to build.

daler:
@johanneskoester nevermind, it seems it was silently failing from the `functools32` dependency. Sorry for the noise.

daler:
cc @endrebak --
1. Do you plan on changing the name any time soon? If so we should probably hold off on merging this.
2. OSX build is failing due to jellyfish dependency (see #417). Since jellyfish is only used for the effective genome size supplementary script, do you think it's reasonable to just skip the jellyfish dependency altogether, and therefore gain an OSX-installable package? bedtools works on OSX so it seems fine to keep it as a dependency for bam support.

endrebak:
Thanks for the bioconda add.

1) I do not plan on changing the name really, as I have no better ideas. And if it is added to bioconda, I will not change it.

2) I think making it available for OS X would be great. I can just add an error message about `jellyfish2` not being installed and tell the users to `brew install jellyfish` (from `homebrew-science`). 

(epic might not work on all OS X instances though, because `zcat` is sometimes called `gzcat` and so on. But I will handle those errors when bugreports are filed.)

PR #1402, opened 2016-04-27
Test all packages independently of updates.

johanneskoester:

This PR prepares for testing of **all** recipes without rebuilding them (using `conda build --test`), in a special repository called **bioconda-tests**. This will enable us to see if an update to the master branch breaks an already uploaded package. Also it will show us if a package breaks due to updates in channels we depend on (like the r channel). The bioconda-tests repository will be configured to be tested every day using http://nightli.es.


johanneskoester:
@bioconda/core please have a look at this. Having this running upon each merge into master seems ideal, but might cause a lot of overhead. We could come up with some scheduling (test every week) instead. However, I am not yet sure what would be the best way to implement that.

johanneskoester:
@bioconda/core please have a look at this. Having this running upon each merge into master seems ideal, but might cause a lot of overhead. We could come up with some scheduling (test every week) instead. However, I am not yet sure what would be the best way to implement that.

bgruening:
Wow, this seems to be a great idea! What about a second repository to which we push after merging to master to trigger the tests. Conda core could be subscribed to this new repo. We could also think about moving other infrastructure things into this new repo.

chapmanb:
Brilliant work, this is incredible. I agree that weekly seems like a good time frame for running this but not sure if it's possible to schedule CI tasks like that.

bgruening:
As a workaround we could touch a file and tell Travis to only test of the timestamp is older than X.

johanneskoester:
Thanks! Well, there is a service for nightly travis triggers. But the idea of Björn is very promising as well. This will work I guess. I will have a look ASAP.

daler:
This is a great idea. Along with the recent subdags additions, these are some elegant improvements!

johanneskoester:
Thanks! Btw. I think the weekly tests can be realized by comparing git commit dates between the master branch and the testall branch. No file touching needed.

PR #1405, opened 2016-04-27
Add dinopy

HenningTimm:

@johanneskoester this should hopefully solve your dependency for simlord.


johanneskoester:
Very nice, thanks!

johanneskoester:
@HenningTimm as far as I know, cython is usually a build dependency only. Or am I mistaken?

HenningTimm:
@johanneskoester that is true. I will change that with the next update.

Since we pack the cythonized .c files it should even be possible to build dinopy without cython at all, but I have not tested that yet.

Issue #1012, opened 2016-03-09
retire osx-whitelist?


daler:
Currently the default is to not build for OSX unless it is explicitly put on `osx-whitelist`. It seems like it would be cleaner to build everything for OSX by default, and failing recipes can then use 
```
build:
  skip: True  # [osx]
```
in the meta.yaml. This way all information about what platforms the recipe runs on is contained in the meta.yaml (rather than having to cross-reference with osx-whitelist). I'm certainly guilty of forgetting to add things to the whitelist. If this happens to other people too, building for OSX by default should increase the availability of OSX packages in bioconda. 

Setting up the build system to do this shouldn't be that tricky. The hard part will be in figuring out which recipes should be retrofitted to have `skip: True #[osx]`.

bgruening:
:+1: for a osx-blacklist. But I suspect than many recipes are not building anymore because the tarball is missing. I have this problem in my add-zlib branch :(

daler:
Do you mean tarballs for bioconductor packages? Or more generally?

bgruening:
Mostly bioc packages. I'm working on a solution for this to fix it once and for all, just need time to write it down and open an own issue for it.

daler:
For the most recent bioc release (month or two ago?) I did a round converting as many to bioaRchive as possible, but many didn't have a tarball there yet. I bet a lot of those have had an incremental release in the meantime, wiping the previous version on the bioc servers.

I feel your pain. To be continued on another issue perhaps . . . but like you say, it still needs to be resolved before removing osx-whitelist.

johanneskoester:
Good idea. I totally support the removal of osx-whitelist. When I introduced it, I wasn't aware of the skip instruction.

kyleabeauchamp:
+1

bgruening:
@daler what do you think about this start? https://github.com/bioconda/bioconda-recipes/compare/kill_osx_whitelist?expand=1

I can work on this further. The idea is
1. add `skip:True` to every file that is currently not on the whitelist
2. add them to the whitelist
3. if all tools are on the whitelist, drop the whitelist

What do you think?

daler:
I think this is a good plan. The disadvantage is that it does not find recipes that work on OSX but were never added to the whitelist by recipe authors. To address this I have an OSX development machine and was starting down the more painful route of deleting osx-whitelist first, and then only adding the `skip: True` to a recipe when it failed to build.

[...]

PR #1409, opened 2016-04-27
remove skip:True [osx] for recipe flash

odoublewen:

I've confirmed that this recipe [flash] builds fine on osx64.  Per issue #1012, can we remove the skip: True for [osx]?

### DIFF
--- a/recipes/flash/meta.yaml
+++ b/recipes/flash/meta.yaml
@@ -1,7 +1,5 @@
 build:
   number: 0
-  skip: False
-
 package:
   name: flash
   version: '1.2.11'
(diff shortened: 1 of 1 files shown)


johanneskoester:
Sure, thanks! Can you please remove the entire line?

odoublewen:
Sure! (I actually removed the line initially, but then noticed that other recipes (like [blast](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/blast/meta.yaml)) have `skip: False` and thought it might be a kind of convention.)  But I am happy to remove the line.

Two naive questions: 
1) Should I increment the build number from 0 to 1?  I suppose not.
2) Am I suppose to upload **my** build to bioconda, or does that somehow happen automatically (due to Travis CI)?

Thanks

bgruening:
@odoublewen it's fine as it is. This will not update the Linux build but will create a OS-X build automatically. You don't need to update it by yourself. Look at conda search in a few minutes :)

Issue #1247, opened 2016-04-08
Packaging GATK-protected


johanneskoester:
I just had a look at this repository:
https://github.com/broadgsa/gatk-protected/

These seem to be the official sources of GATK including the non-commercial licensed parts (e.g. the callers I think). We have other non-commercial license software in bioconda. Hence, my question is, can't we package this as well? Both the anaconda page and the bioconda homepage will clearly state that this is non-commercial use only.

@bioconda/core any thoughts?

johanneskoester:
Looking into the license, this passage might be problematic:

> 2.2 No Sublicensing or Additional Rights. LICENSEE shall not sublicense or distribute the PROGRAM, in whole or in part, without prior written permission from BROAD. LICENSEE shall ensure that all of its users agree to the terms of this Agreement. LICENSEE further agrees that it shall not put the PROGRAM on a network, server, or other similar technology that may be accessed by anyone other than the LICENSEE and its employees and users who have agreed to the terms of this agreement.

kyleabeauchamp:
That is pretty clearly a no-go IMHO.  Furthermore, even in the absence of such a clause, I don't think it's safe to assume that you have "redistribution" rights for binaries.

kyleabeauchamp:
For example, all of the latest UCSC tools.

kyleabeauchamp:
Also regarding ucsc, we should stop using varies in license field.

chapmanb:
You aren't allowed to re-distribute GATK, unfortunately. This has been the source of a lot of effort on building workarounds in bcbio. I'm having a conversation with the Broad team about allowing it in docker, but it's not yet clear if it will be possible. I'll ask their thoughts on bioconda as part of this conversation.

johanneskoester:
Great, thanks Brad!
@kyleabeauchamp we are building ucsc packages from source, right? Does that also have these problems?

kyleabeauchamp:
AFAIK yes.  Here is the license info:

http://genome.ucsc.edu/license/

It's noncommerical, so we at the very least need each yaml file to say that this component has a non-commericial license.  It's not clear to me that you are allowed to redistribute without their permission, even for academics.

johanneskoester:
Well, the source code and binaries are freely available for academic use. Question is if redistribution via bioconda and installation by academic users can be seen as academic use. At least they don't disallow redistribution.

kyleabeauchamp:
At the very least, we need to fix the license section in the yaml files.

[...]

PR #1413, opened 2016-04-28
Gatk V3.5

jerowe:

Hi Johannes,

This PR is to update the version of GATK (gatk-framework) from 3.4.x to 3.5. It also updates the bin script.

It follows the same structure as the 3.4.x release. Credit really goes to this guy for the build system. https://github.com/chapmanb/gatk . ;-)

Best,
Jillian

### DIFF
--- a/recipes/gatk-framework/gatk-framework
+++ b/recipes/gatk-framework/gatk-framework
@@ -6,17 +6,7 @@
 set -o pipefail
 export LC_ALL=en_US.UTF-8
 
-# Find original directory of bash script, resovling symlinks
-# http://stackoverflow.com/questions/59895/can-a-bash-script-tell-what-directory-its-stored-in/246128#246128
-SOURCE="${BASH_SOURCE[0]}"
-while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
-    DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
-    SOURCE="$(readlink "$SOURCE")"
-    [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
-done
-DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
-
-JAR_DIR=$DIR
+JAR_DIR="$(cd "$(dirname "$0")" && pwd -P)"
 
 java=java
 if [ -e "$JAVA_HOME/bin/java" ]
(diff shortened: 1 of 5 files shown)


recipes/gatk-framework/gatk-framework
```diff
@@ -6,17 +6,7 @@
...
-SOURCE="${BASH_SOURCE[0]}"
-while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
-    DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
-    SOURCE="$(readlink "$SOURCE")"
-    [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
-done
-DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
-
-JAR_DIR=$DIR
+JAR_DIR="$(cd "$(dirname "$0")" && pwd -P)"
```
johanneskoester: I don't immediately get this. Can you explain why this can replace the code on the left?

johanneskoester:
Thanks @jerowe!
@chapmanb, this is your baby. Can you have a look?

chapmanb:
Jillian;
Thanks for doing this. It looks like your code package gatk-protected, so should be a different recipe. The original gatk-framework is only the MIT licensed parts of the code (I can bump that part to 3.5 if it is useful to folks) so does not include variant calling and other license restricted code.

For gatk-protected, it's not clear we're allowed to redistribute it due to the terms of the licensing. See the discussion in #1247 for current thoughts. Thanks for helping move this forward.

jerowe:
Hi @chapmanb and @johanneskoester 

Thanks for putting in the work to make the uberjar in the first place! I'm not a java person, and I never would have figured that out. If I'd known you were still activate I would have asked you for help. ;-)

I was a bit unclear on the licensing bit myself, but I thought it would check out as ok since the software is free for non profit use. My thoughts were the responsibility was on the user once the licensing was clear. I also put a sentence when you start up the gatk-framework stating as such. I took a look at the discussion. 

I see there was talking of having a register script. I'd be happy to work on that if no one else can. We need GATK 3.5 for an inhouse pipeline, and its getting that up and running is a priority for me right now.

Best,
Jillian

jerowe:
Taking a closer look at the licensing is making me a bit paranoid, so I'm going to delete the repo I forked from gatk-protected.

jerowe:
There is an email in the licensing info. I'm going to send off a quick email and ask.

jerowe:
@chapmanb , if you wanted to update your GATK to 3.5 that would be awesome.

btw, the more I thought about it the more I realized I should have just asked you first. I'm really sorry if I stepped on your toes! I get tunnel vision and don't think sometimes. ;-)

chapmanb:
Jillian -- thanks much, I appreciate you getting things started with moving to GATK 3.5. I pushed a new recipe built off GATK 3.5 now, so hopefully this will work for your analyses. Thanks again.

jerowe:
@chapman, thanks for the new release! ;-)

PR #1414, opened 2016-04-28
add IPO R library from a GitHub repositiory

lecorguille:

@bioconda/core

I'm not used to install R library from a GitHub repositiory. So let me know if there is better way to do that.
Many thanks


johanneskoester:
You should specify the git repo directly in the meta.yaml (see conda docs). Importantly, it would also need a git tag or a commit, in order to be reproducible.

lecorguille:
Obviously!
Not use a tag (unavailable) had disturbed me. I didn't think that commits could be download.
Many thanks!

lecorguille:
Hum ....
**[r-ipo](https://github.com/bioconda/bioconda-recipes/blob/r-ipo-part2/recipes/r-ipo/meta.yaml)** needs [bioconductor-xcms](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bioconductor-xcms/1.46.0/meta.yaml) which needs [bioconductor-mz](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bioconductor-mzr/meta.yaml) which need libnetcdf

Once apon a time ....

Some hours ago, on my OSX, `conda build r-ipo` asked for `@rpath/libnetcdf.7.dylib` for `mzR.so`. So I add `- libnetcdf 4.2.1.1` in my deps. Travis was all green.

Just after, I tried to remove `- libnetcdf 4.2.1.1`. I don't know why it works suddenly on my OSX. But Travis-OSX-SUBDAG=0 fails

When I tried to restore `- libnetcdf 4.2.1.1` on my OSX, this time the build fails because it ask for `@rpath/libnetcdf.11.dylib` provided by libnetcdf 4.3.3.1 set in the bioconductor-mzr :cry: 

So I specify the 4.3.3.1 version in the mzr meta build1 2.4.1-r3.2.2_1. But Travis-OSX-SUBDAG=0 still fail because it use the first build  2.4.1-r3.2.2_0 and by consequence libnetcdf-4.4.0 is used :sob: 

So my question:
- what is SUBDAG=0?
- why is it using the build 0 of mzr instead of the 1?

Thanks a lot

daler:
Not sure if it's the cause of any problems, but libnetcdf 4.3.3.1 in the meta.yaml (e.g. [this line](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bioconductor-mzr/meta.yaml#L21))  should be specified as
```yaml
- 'libnetcdf ==4.3.3.1'
```

rather than
```yaml
- libnetcdf\t4.3.3.1
```
which will cause YAML parsing errors.

lecorguille:
The same :/

The travis test use:
```
libnetcdf:                 4.4.0-1
bioconductor-mzr:          2.4.1-r3.2.2_0
```

lecorguille:
Sorry to insist!

But is anybody can explain me why when I install IPO dependencies on my OSX or on our CentOS, it installs
```
bioconductor-camera: 1.26.0-0
bioconductor-mzr:    2.4.1-r3.2.2_1
bioconductor-xcms:   1.46.0-0
libnetcdf:           4.3.3.1-3
```
But for [this OSX -> SUBDAG=0, travis test](https://travis-ci.org/bioconda/bioconda-recipes/jobs/128809516), those package are installed?
```
libnetcdf:                 4.4.0-1
bioconductor-mzr:          2.4.1-r3.2.2_0
```
Why does it install the build 0 of bioconductor-mzr?
All the 7 other tests pass!

lecorguille:
Let skip OSX in first place.
Personnally, I need it for CentOS :)

Travis :+1:

johanneskoester:
Some of the versions might not have been build on osx, causing a different set of dependencies to be installed.

recipes/r-ipo/meta.yaml
```diff
@@ -0,0 +1,106 @@
...
+    - lib/
+
+# Suggests: parallel
+requirements:
+  build:
+    - r
+    - bioconductor-xcms
+    - r-rsm
+    - bioconductor-camera
+    #- libnetcdf 4.2.1.1
```
johanneskoester: I might have missed this in the discussion, but I understood that libnetcdf is a dependency right? Then, it should be in here. However, it should be pinned similar to what we do with boost (see e.g. the recipe clever-toolkit and the file `scripts/env_matrix.yml`.

lecorguille: libnetcdf is not a direct dependency of r-ipo.
libnetcdf is need my bioconductor-mzr which is need by bioconductor-camera which is need by r-ipo

bioconductor-mzr passed all the tests (and so bioconductor-camera), both in the master.
So I can't figure out why when installing r-ipo, on this particular test OSX->SUBDAG=0, it fails.

johanneskoester:
Ok, I know what happens here. We have an old package of bioconductor-mzr that did not depend on libnetcdf or did not pin the version. The new one hardcodes the version, but that causes a conflict with another package, such that the old, broken bioconductor-mzr is used.

I have deleted the broken builds. Let's see what happens.

In the long run, we need to use proper pinning of libnetcdf in all packages to resolve such conflicts.

johanneskoester:
Can you reactivate osx support please?

lecorguille:
Done :+1: 
Many many thanks

bgruening:
Awesome! Thanks @lecorguille and @johanneskoester !

lecorguille:
Great!

PR #1422, opened 2016-04-28
recipes to build ibis-framework and associated deps

xguse:

Hello @bioconda/core.

This PR adds Wes McKinney's (@wesm) out-of-core mem dataframe manipulation framework.  It is not _specifically_ bio-related but I have noticed the need to play with csvs large enough that my main machine is not happy.  I expect others have or will experience this very soon.  Since it might be considered a little off topic, I wanted to flag you guys for your opinion.

Gus

### DIFF
--- a/recipes/fastavro/bld.bat
+++ b/recipes/fastavro/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
+:: http://docs.continuum.io/conda/build.html
+:: for a list of environment variables that are set during the build process.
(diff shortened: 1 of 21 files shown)


wesm:
I'm not sure I agree with the way that Ibis is being described in this patch. If you have a Hadoop cluster running (and Impala, too), then Ibis may be a useful solution for you, but otherwise it is too early.

xguse:
No sqlite?

Sent by MailWise<http://www.mail-wise.com/installation/2> – See your emails as clean, short chats.


-------- Original Message --------

wesm:
You can use SQLite, so if you have an existing SQLite database this could be useful. Let me know how it goes

xguse:
@wesm: Thanks; I will continue to play with it using sqlite and let you how things work out.

Others: Unfortunately, even though I have gotten it to build for my own situation (python3, linux64) after tweaking the conda build recipes a bit, some of the deps' meta.yaml files just really don't seem to like the python2 situation and I don't have too much more time to use troubleshooting the solution right this second. I am closing this for now, but will probably come back in a week or so and tackle it for good.

PR #1436, opened 2016-04-30
Generalize dependency pinning using an environment variable matrix

johanneskoester:

This PR generalizes the pinning mechanism. In the future, we can add new environment variables for pinning recipes using `scripts/env_matrix.yml`. Our build script will automatically build all combinations. It won't build a recipe twice if it is not affected by a particular combination. E.g., non-python recipes will be build only once, although we have two python versions in our matrix.

Still, I think we should keep the number of parallel versions minimal. I would suggest to let Python be the only case where we support more than one version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,17 +56,17 @@ Bioconductor packages will be built using:
 
 Then, you can test it in the docker container with:
 
-    docker run -v `pwd`:/bioconda-recipes bioconda/bioconda-builder --packages your_package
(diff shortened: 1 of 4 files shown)


johanneskoester:
@bioconda/core please note above changes. If we need to pin more libs apart from boost, we can do that with `scripts/env_matrix.yml` from now on.

chapmanb:
Brilliant, this is a nice abstraction. Thanks for doing this.

bgruening:
Nice!

PR #1443, opened 2016-05-02
Adding recipes for perl HPC::Runner-* and BioX::Workflow

jerowe:

Adding recipes for HPC::Runner::MCE, HPC::Runner::Slurm, HPC::Runner::PBS and BioX::Workflow


recipes/perl-biox-workflow/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+   # - fix.patch
+
+# build:
+  # If this is a new build for the same version, increment the build
+  # number. If you do not include this key, it defaults to 0.
+  # number: 1
+
+requirements:
+  build:
+    - perl
```
johanneskoester: perl-threaded is a special version of perl, compiled with threading support. Please don't depend on both perl and perl-threaded. Just on perl-threaded. Thanks!

jerowe:
@johanneskoester , I changed it to just perl-threaded. Using threaded perl for biox-workflow is kind of overkill anyways.

recipes/perl-biox-workflow/meta.yaml
johanneskoester: I still see a perl instead of perl-threaded here. I am pretty sure that depending on both in two different packages will cause problems.

PR #886, opened 2016-02-26
Add ncbi-vdb recipe

saketkc:
This and `ngs-sdk` package  are requirements for NCBI's [sra-tools](https://github.com/ncbi/sra-tools). I am working on `sra-tools`'s recipe.

PR #76, opened 2015-10-14
Add ncbi-vdb

percyfal:
Add ncbi-vdb. Building sra-tools requires both the build and source files of ncbi-vdb. These have been placed in $PREFIX/share/ncbi.

PR #1451, opened 2016-05-02
sra-tools plus dependencies

simonvh:

Ok, this is a first go at trying to get fastq-dump (and other utilities from the NCBI sra-tools) to work. Apparently sra-tools also expects the source of both ngs-sdk and ncbi-vdb to be present at compile time. The way I solve this now is to create, according to FSH standards, a src directory. However, there might be a different or better way to solve this.

@bioconda/core any ideas?

Also, I've skipped osx for now, as I have no means to test if it works.


daler:
So the built packages of `ngs-sdk` and `ncbi-vdb` each also contain their own source in `/src/ncbi`, and building `sra-tools` requires both the built `ngs-sdk` and `ncbi-vdb` *and* their respective sources. Is that correct?

How big is the source?

If small, this seems fine. If big, one alternative might be to create source packages for `ngs-sdk` and `ncbi-vdb`, so the `sra-tools` meta.yaml looks like:

```yaml
build:
  - ngs-sdk-src
  - ncbi-vdb-src
run:
  - ngs-sdk
  - ncbi-vdb
```

But in this case `-src` and built versions should probably be pinned in order to avoid potential mismatches between source and compiled packages.

As for OSX, you could always remove the `skip: True` and let travis-ci test it to see if you get lucky!

simonvh:
> So the built packages of ngs-sdk and ncbi-vdb each also contain their own source in /src/ncbi, and building sra-tools requires both the built ngs-sdk and ncbi-vdb and their respective sources. Is that correct?

Yes, this is correct as far as I can determine. 

Both src dirs together are ~64M. Not exactly small, on the other hand, anyone installing sra-tools is bound to use much more disk space then that ;-). Of course I would be fine with make separate source packages, but it seems that it will become even more fiddly than it is now.

daler:
Nice, OSX worked!

Agree with sra-tools users not being stingy about space; 64M combined shouldn't be an issue and is a small price to pay for maintainability of the recipes.

@bioconda/core, other thoughts?

bgruening:
I guess we should consider that bioconda does not have infinite space on anaconda.org.

simonvh:
Yes, I can see that would be a consideration. However, with separate
packages that would still be the same, as these source packages would still
have to be stored, right? In this case it would mainly concern the user
system.

Op 3 mei 2016 18:04 schreef "Björn Grüning" <notifications@github.com>:
>
> I guess we should consider that bioconda does not have infinite space on
anaconda.org.

bgruening:
Another idea would be to ship both tools with one package and delete the source after make?

daler:
In this case it would be all 3 packages in 1 (since sra-tools needs source and build of both ngs-sdk and ncbi-vdb). 

That said, it's not clear that ngs-sdk and ncbi-vdb would be used on their own, so bundling them all might be a good strategy.

simonvh:
Indeed, it would then be only the sra-tools package, which would include the other two. However, the ngs-sdk and ncbi-vdb source is also used by ngs-bam, ngs-python and ngs-java, which are language-specific bindings to the SDK. So then ideally these would also need to be included. If we do this, then indeed the source can be deleted after the make.

My vote would be for separate packages, as it would then be possible to install SDK bindings without having sra-tools, but this might be just a hypothetical situation. So if you prefer to have one package then I'll create that. 

One other thing is that I noticed there are two similar open PRs #886 and #76, which I didn't notice before (sorry @percyfal and @saketkc).

bgruening:
I think all in one makes sense :-) 
Thanks for working on this!

simonvh:
This should be it then. It's become a rather complicated recipe but it works :).

Note that I have used [this piece of code](https://github.com/stuarteberg/conda-multisrc-example) to download multiple sources using conda's mechanism, which will prevent downloading of the source files every time the recipe is built. Might be useful as example for other recipes that need to download multiple files.

daler:
Thanks for working out all these details. That multi-download script looks very useful.

I also vote for the separate packages but I'm a little too late :). I suppose they could always be split back out when/if the `ngs-*` bindings are needed for other recipes. The important part is that we finally have a working `sra-tools`!

saketkc:
Awesome, thanks a ton!

percyfal:
@simonvh No problem, I planned to finish the sra-tools package before going on parental leave but got stuck on the way. I have no time to work on this now so if you get it done, all the better. Note that there is a recipe for ngs-sdk, if that's any help. Cheers!

simonvh:
@bioconda/core just to let you know that the build is correct as far as I've tested ;)

bgruening:
@daler are you fine with merging it? I think it looks good :)

daler:
Yep, looks good!

PR #1457, opened 2016-05-03
Pin ncurses version with our new pinning mechanism.

johanneskoester:

Define an ncurses version to use and update recipes by pinning ncurses to that version.
@bioconda/core I am unsure if we need to do something like this for all shared library dependencies...
Any comments?


bgruening:
:+1:

bgruening:
Thanks Johannes!

daler:
Seems like it's easy enough to add a new env var and `sed` recipes if/when any other dependencies become problematic rather than preemptively changing them all over now.

Probably a dumb question, but what are some examples of other shared lib dependencies?

bgruening:
zlib, libxml?, libpng?

PR #1475, opened 2016-05-05
change shebang

bgruening:

### DIFF
--- a/recipes/krona/2.6/build.sh
+++ b/recipes/krona/2.6/build.sh
@@ -3,3 +3,5 @@ mkdir -p $PREFIX/opt/krona
 mv ./* $PREFIX/opt/krona
 cd $PREFIX/opt/krona
 ./install.pl --prefix=$PREFIX
+
+sed -i.bak 's|#!/usr/bin/perl|#!/usr/bin/env perl|' $PREFIX/bin/* 
--- a/recipes/krona/2.6/build.sh
+++ b/recipes/krona/2.6/build.sh
@@ -4,4 +4,4 @@ mv ./* $PREFIX/opt/krona
 cd $PREFIX/opt/krona
 ./install.pl --prefix=$PREFIX
 
-sed -i.bak 's|#!/usr/bin/perl|#!/usr/bin/env perl|' $PREFIX/bin/* 
+LANG=C sed -i.bak 's|#!/usr/bin/perl|#!/usr/bin/env perl|' $PREFIX/bin/* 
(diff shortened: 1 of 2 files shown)


yesimon:
Is this a normal practice where we replace shebang lines even when it works fine when `install.pl` is run and the conda bin path is in `$PATH`? Changing the shebang lines should not be necessary.

bgruening:
@yesimon how does it work if `/usr/bin/perl` does not exist?

yesimon:
@bgruening Ok, it looks like some other packages do change the shebang, but it seems like patches are preferred to sed to be cautious. Otherwise, you can also send up a PR to the KronaTools repo which is active.

bgruening:
@yesimon sure! This was a quick and dirty hack and it did not even work ;)
I need more time to work on this :(

odoublewen:
I think the quick and dirty hack failed b/c Krona has a space after the shebang, so your regex misses.

Anyway, I made a PR on the Krona repo.
https://github.com/marbl/Krona/pull/13

bgruening:
@odoublewen you rock! Thanks! I haven't looked at the error yet.

jerowe:
I think the proper way to do this is to use Module::Build. That way the perl scripts point to the perl they were installed with.

https://gist.github.com/jerowe/c08dd31e83380b7dc14621139979c11a - Is a quick and dirty example, but I want to make it more robust.

This will more fix things like distributions (PFAM, trinotate, miravelvet-sl) having a mix of perl shebang lines. Sometimes they're '#!/usr/bin/perl', which is very, very wrong, sometimes they're '#!/usr/bin/env perl', which is better but still not perfect. I saw '#/usr/bin/perl' yesterday, and that just made me mad. ;-)

In general I'm also moving towards a direction where I go through each package, see if it has perl distros, and install the *pm-s to the perl lib folder, using Module::Build. 

I feel like this makes things much more user friendly. Conda is already built for isolated environments, so we don't need to worry about cluttering up spaces. A fully bundled pfam/trinotate/metavelvet/etc should be sitting off in its own corner with its own perl and @INC. Part of the point of conda installs is that we shouldn't have a long list of instructions. It shouldn't be 'install this, change your perl @INC to this, make sure your perls are pointing in the right place and/or that your modules are loaded in a particular order'. 

I think Module::Build has it right. On systems where you might have many perls installed you want to make sure your script is calling to its own perl and @INC.

Think of all the tickets it will save admins and rejoice!

Issue #1372, opened 2016-04-22
Replacing/updating hall-lab-svtools


ernfrid:
I've been significantly updating the hall-lab svtools suite and am narrowing in on a release. I was thrilled to see that there is already a bioconda recipe for an earlier version and would like to make a conda recipe for the new version. 

I had a question though, the bioconda recipe is for hall-lab-svtools. The package itself is called svtools and I've uploaded it to pip under that name. Is it possible to remove or rename the old package? What would be your recommendation?

bgruening:
I would keep the old version and add a comment to the yaml that this one is deprecated and replaced by the new one.

johanneskoester:
For me, it would be fine to remove the old recipe from the repository, and just keep the packages in the bioconda channel. This would further decrease the chance that people install the old one. @brainstorm this is your recipe, what do you think?

chapmanb:
Dave;
Thanks for the work on svtools and updating the bioconda package. I'm the originator of the hall-lab-svtools name and was probably being over specific. Happy to call it whatever you prefer and agree with Johannes that we can rename the current one to svtools and then update. Thanks again.

brainstorm:
@johanneskoester, IIRC the reason for the naming was to differentiate from the different forks and name collisions the tool was having at the time. It's totally ok with me to remove (rename) the old tool and upload the new one with `svtools` name, much better, IMHO. Thanks all!

johanneskoester:
@ernfrid, looks like we have a decision here. Feel free to take over the existing recipe, rename it to svtools and update the version. Do you want me to add you as a bioconda member?

ernfrid:
Got it! Thanks very much everybody. I really appreciate the guidance.

@johanneskoester - Definitely add me. That sounds great.

Just so you're all aware. I'm still waiting on some documentation before releasing so there may be a delay before you see a new recipe.

bgruening:
@ernfrid thanks for your contribution. I will close this issue now.

PR #1479, opened 2016-05-05
Rename and update hall-lab-svtools

ernfrid:

As discussed in #1372, this PR renames hall-lab-svtools to svtools and updates the recipe to the newest version. I also maintained the older version like what is done with [bxpython](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/bx-python), but it is a little unclear to me if that is desirable. 

I am new to conda and bioconda, but have followed the guidelines and updated the meta.yaml to mimic the earlier recipe where necessary. @bioconda/core, I would greatly appreciate a some review to make sure this recipe is up to snuff.


kyleabeauchamp:
LGTM.

As an minor nit, I wonder if you might to be more explicit about minor version numbering on various libraries.  I know for example that scipy, pandas, and statsmodels have had various API-breaking changes over the years, so it might be good to have some sort of minimum version as a guideline.

chapmanb:
Dave -- thanks so much for this and for all the work on svtools.

ernfrid:
Thanks Brad and Kyle. 

Kyle - I'll look into the minimum versions for scipy et al and try to get those in for the next release.

PR #1485, opened 2016-05-06
Fixed a problem with the last IDR version bump PR.

nboley:

I forgot to change 
fn: 2.0.2.tar.gz 
to
fn: 2.0.3.tar.gz
in my previous PR.

Not sure how I missed that - really sorry about that.

### DIFF
--- a/recipes/idr/meta.yaml
+++ b/recipes/idr/meta.yaml
@@ -8,7 +8,7 @@ source:
   md5: 14219095e41650221634e6097842568d
 
 build:
-  number: 0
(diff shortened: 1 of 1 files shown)


daler:
No worries -- it shouldn't matter too much since it's just the file to which conda-build will download the URL. No version information is taken from that filename, it just seems to be good practice to match them to the URL when possible.

But just to be sure, could you bump the build number here to 1?

nboley:
Done - thanks again.

Issue #1487, opened 2016-05-06
Broken conda-build


johanneskoester:
Dear @bioconda/core. I have recently merged a PR of Continuum into our conda-build fork. I trusted them, hence I merged without prior testing. Unfortunately, this broke our fork [see here](https://travis-ci.org/bioconda/bioconda-recipes/jobs/128226989). I am moving to a new apartment this weekend. If somebody needs a quick fix before I have internet again, please have a look at our fork and above link. It should be fairly simple to fix (seems to be a refactoring issue).

Sorry for the inconvenience!!!

chapmanb:
Johannes;
I think I've fixed the issues with this branch and things seems to be building. In addition to the immediate error with refactoring, I had to fix the recursive check in this branch that looks for meta.yaml files and fails on your directory structure since we have multiple of these when there are older recipe versions. I'm trying to get the builds re-run now -- they were running into time limit errors but hopefully a restart will get the backlog cleared.

chapmanb:
Hi all -- I'm not sure if something else is broken from the merge as I'm geting timeouts on all of the builds:

https://travis-ci.org/bioconda/bioconda-recipes/builds/128226983

I'm not familiar enough with the build system to know the best way to debug and need to appeal for some help. Sorry not have this resolved on a first pass.

daler:
travis-ci is having issues with OSX builds today, and to rule out problems with linux and/or docker builds I forced building of a single recipe. [That worked](https://travis-ci.org/bioconda/bioconda-recipes/jobs/128312342).

Locally in a docker container, the build script seems to want to build a LOT of packages. It correctly skips a bunch, but not all, and it's not clear to me why it wants to build the ones it does. So it seems we're timing out on travis-ci because the build script tries to build too many packages.

Since I don't have more time for this at the moment but I wanted to have a working build system, in 8e92922 the build script temporarily uses the pre-merge commit from the conda-build fork.

chapmanb:
Ryan;
Thanks for investigating this and for rolling things back to keep it working. Hopefully Johannes will have more of an idea about the excessive package builds once he's settled into his new place.

PR #1489, opened 2016-05-06
Fix Fiji Life-Line version 2014 November 25 for OS X

gregvonkuster:


gregvonkuster:
@bgruening so sorry for the delay on this - I have been really swamped.  It took me a bit, but I discovered the issues on OS X were related to Java 3d libraries which we do not use so I've eliminated the files from the source that required linking during the build.  I believe this should now work for both Linux and OS X.

daler:
After fixing the build problems (unrelated to this PR; see #1487) I re-ran the travis test and it passed . . . so, merging!

bgruening:
Thanks!

PR #1510, opened 2016-05-09
Test new zlib less Docker

bgruening:

### DIFF
--- a/recipes/bwa/build.sh
+++ b/recipes/bwa/build.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
 
+sed -i.bak 's/^INCLUDES=$//g' Makefile
+
 export CPPFLAGS="-I$PREFIX/include"
+export INCLUDES="-I$PREFIX/include"
 export LDFLAGS="-L$PREFIX/lib"
 
 make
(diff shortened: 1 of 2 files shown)


bgruening:
This one actually should fail with the new zlib less Docker image.

kyleabeauchamp:
Maybe because we don't have a test on this one?  AFAIK just running `bwa` should print the help string...

bgruening:
lol @kyleabeauchamp you are absolutely right. Why does this one have no test? I just picked it by chance :)
Thanks for the pointer! We should implement a linting of yaml files. At lest one test, if cmake as dependency, gcc/llvm as well and so on.

kyleabeauchamp:
It also seems to be non-compliant w.r.t. our GCC / LLVM annotation conventions as well?  Not sure.

kyleabeauchamp:
It might be because of stdout versus stderr printing stuff, which can be tricky to handle for some executables.  Someone probably knows how to do it though :)

bgruening:
Give me a sec I will work on this.

bgruening:
Ok let's see if this is finally failing :)

bgruening:
@kyleabeauchamp still not failing. But we have enhanced it :)

kyleabeauchamp:
Agreed.

bgruening:
@kyleabeauchamp isn't it nice - it's failing! :)

PR #1517, opened 2016-05-10
Added perl package from cpan

matthdsm:

dependency for latest Ensembl VEP


recipes/perl-file-copy-recursive/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+  build:
+    - perl-threaded
+    - perl-pathtools
+
+  run:
+    - perl-threaded
+    - perl-pathtools
+
+test:
+  # Perl 'use' tests
```
bgruening: Can you add a test?

matthdsm:
it seems I issued this pull request way too early...
there is another perl dependecy short to build this package, but building this dependency creates a strew of other dependencies that aren't available in the bioconda channel. Should I submit a recipe for for each and every dependency that isn't available of is there another solution?

M

bgruening:
You can submit everything in one PR. Please note that the skeleton command has a `--recursive` feature.

leipzig:
looks like perl-file-copy-recursive did not get merged. Were there more issues with this?

PR #1526, opened 2016-05-10
fiji_fix

bgruening:


gregvonkuster:
@bgruening This looks fine to me if this makes things work - it does clutter the bin directory with all of the source though.  Should the same change be made for the OS X section?

bgruening:
@gregvonkuster yes, but more importantly it fails the tests :(

PR #1527, opened 2016-05-10
updateTaxonomy.sh improvements

odoublewen:

Added symlink so that updateTaxonomy.sh is accessible to user in PATH. 

Also, the default taxonomy directory is being left out during packaging, presumably b/c it is empty.  The placeholder file makes sure it is present for user after install.

I think it would be helpful if a message is emitted to user upon installation of this package.  I guess the user won't see the echo message from the build.sh script.  What is the better way to do this?  Does the meta.yaml support some kind of  "message" tag for this purpose?

### DIFF
--- a/recipes/krona/2.5/build.sh
+++ b/recipes/krona/2.5/build.sh
@@ -3,3 +3,5 @@ mkdir -p $PREFIX/opt/krona
 mv ./* $PREFIX/opt/krona
 cd $PREFIX/opt/krona
 ./install.pl --prefix=$PREFIX
+ln -s $PREFIX/opt/krona/updateTaxonomy.sh $PREFIX/bin/ktUpdateTaxonomy.sh
+echo 1 > $PREFIX/opt/krona/taxonomy/placeholder
--- a/recipes/krona/2.5/meta.yaml
+++ b/recipes/krona/2.5/meta.yaml
@@ -4,7 +4,7 @@ about:
   summary: 'Krona Tools is a set of scripts to create Krona charts from several Bioinformatics
             tools as well as from text and XML files.'
 build:
-  number: 3
+  number: 4
 package:
   name: krona
   version: 2.5
--- a/recipes/krona/2.5/post-link.sh
+++ b/recipes/krona/2.5/post-link.sh
@@ -1,0 +1,14 @@
+echo "
+Krona installed.  You still need to manually update the taxonomy
+databases before Krona can generate taxonomic reports.  The update
+script is ktUpdateTaxonomy.sh.  The default location for storing
+taxonomic databases is $PREFIX/opt/krona/taxonomy
+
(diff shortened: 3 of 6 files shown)


odoublewen:
@yesimon What do you think of my changes here?  Also, I am confused at the Travis build fail.  When I look at the [log](https://travis-ci.org/bioconda/bioconda-recipes/jobs/129278457) the line showing the error says `FileNotFoundError: [Errno 2] No such file or directory: '/anaconda/envs/_build/bin/ImageJ-linux32'` -- But what does ImageJ have to do with this package??  Did something go haywire in TravisCI?

odoublewen:
@bgruening the build fail is really strange -- it seems that it tried to build both krona and fiji, which I notice was the PR that you made moments before this one.  Is it normal for TravisCI to try to bundle sequential PRs into a single test?

yesimon:
I don't think it's "right" to add a symlink to `updateTaxonomy.sh` when it's such a generic name. Since it is called by the user it should probably be unifed with the rest of KronaTools and symlinked as `ktUpdateTaxonomy.sh`.

It seems like there is a mechanism to emit messages - please see: http://conda.pydata.org/docs/building/build-scripts.html and https://github.com/conda/conda/blob/91847aab60b571a5d89b5e2a582612b3cd383f7e/conda/install.py#L1008, probably by echoing to `$PREFIX/.messages.txt` in a `post-link.sh` script.

The build fail is probably caused by the merged in commit relevant to fiji earlier today.

odoublewen:
Thanks!  Actually, it's funny, I initially did exactly what you suggest (prefixing the kt) but then thought maybe I was being too clever and might confuse long time users with the name change.  But I agree, it is better to keep the namespace of the kt commands consistent.

Thanks for the tip for .messages.txt and post-link.sh -- it works great.  Pushing update now.

I noticed that fiji was pulling from ImageJ and causing the build failure... but still puzzled why it got pulled in since this PR is concerning only commits to krona.

yesimon:
One last thing - instead of making a placeholder file, I would prefer `mkdir -p $PREFIX/opt/krona/taxonomy` unless `updateTaxonomy.sh` requires more than that.

odoublewen:
Yes, this was my preference too, and was what I tried initially.  Actually, there is no need to mkdir, b/c an empty dir is already present in the tarball.  But for some reason it goes away during the conda packaging -- not sure why. Putting the placeholder file made it stay.

yesimon:
Sounds good. Thanks!

bgruening:
@odoublewen Travis has some problems afaik and the fiji package has also one. I will work on this.
If you have time can you please look at this: https://github.com/bioconda/bioconda-recipes/pull/1475/
I saw that the shebang is not `/usr/bin/env perl` for krona which can cause errors if your perl is in a non-standard location.

Thanks!

PR #1528, opened 2016-05-10
Update bcbreport.

lpantano:

### DIFF
--- a/recipes/bcbreport/meta.yaml
+++ b/recipes/bcbreport/meta.yaml
@@ -1,12 +1,12 @@
 package:
   name: bcbreport
-  version: "0.99.28"
+  version: "0.99.29"
 build:
   number: 0
   skip: True # [osx]
 source:
   git_url: https://github.com/lpantano/bcbio.coverage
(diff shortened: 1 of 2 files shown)


lpantano:
Hi @bgruening,

do you know what is going on with Fiji package? it seems the file you get from the URL is corrupted, so everything is failing. Do you have any idea how to fix this, since this is already in master, all PR will fail.

thanks  a lot!

lpantano:
Hi @gregvonkuster, I had to comment the symlink of the 32 version because it was failing. It seems now there is another file for linux32 that is not the same that this one: http://downloads.imagej.net/fiji/Life-Line/fiji-linux64-20141125.tar.gz

I downloaded this file, and this line fails because the file doesn't exists:

`ln -s $PREFIX/ImageJ-linux32 $PREFIX/bin/ImageJ-linux32`

gregvonkuster:
@lpantano Sorry about this - I'm running Linux64, so had not tested this.  I assume @bgruening has corrected this problem with this PR: https://github.com/bioconda/bioconda-recipes/pull/1526.

PR #1530, opened 2016-05-11
Added metilene recipe

jfallmann:


bgruening:
Jörg I own you :beers: for this!

jfallmann:
Sebastian wants one too :smile:

recipes/metilene/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+requirements:
+  build: 
+    - llvm # [osx]
+    - gcc # [linux]
+
+  run: 
+    - bedtools >=2.24
+    - perl-threaded
+    - r-ggplot2 >=2.0.0
+    - libgcc # [linux]
```
bgruening: Not everytime when you specify gcc as build dep you need libgcc as  run-dependency. So maybe you can get rid of it.

bgruening:
No documentation on OS-X ;)
`install: /anaconda/envs/_build/share/doc/metilene: No such file or directory`

@s-will we can do!

recipes/metilene/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+make -j ${CPU_COUNT}
+cp metilene ${PREFIX}/bin
+cp metilene_input.pl ${PREFIX}/bin
+cp metilene_output.pl ${PREFIX}/bin
+cp metilene_output.R ${PREFIX}/bin
+install manual.pdf ${PREFIX}/share/doc/metilene
```
s-will: We do not create the directory yet. One could e.g. add
install -d ${PREFIX}/share/doc/metilene
to explicitely create the dir or run
install -D manual.pdf ${PREFIX}/share/doc/metilene/manual.pdf

s-will:
Hi @jfallmann, it seems to work as intended only after
diff --git a/recipes/metilene/build.sh b/recipes/metilene/build.sh
index f0c34ef..4f2865d 100644
--- a/recipes/metilene/build.sh
+++ b/recipes/metilene/build.sh
@@ -4,5 +4,7 @@ cp metilene ${PREFIX}/bin
 cp metilene_input.pl ${PREFIX}/bin
 cp metilene_output.pl ${PREFIX}/bin
 cp metilene_output.R ${PREFIX}/bin
+
+install -d ${PREFIX}/share/doc/metilene
 install manual.pdf ${PREFIX}/share/doc/metilene
 install README ${PREFIX}/share/doc/metilene


Please also uncomment the tests of the perl scripts. At least on my system they work (no idea what went wrong before)

jfallmann:
@s-will done, 
but tests fail for me,
+ metilene --help
+ grep VERSION
 [VERSION]
+ metilene_input.pl --help
+ grep usage
TESTS FAILED: metilene-0.2.6-0

recipes/metilene/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - perl-threaded
+    - bedtools >=2.24
+    - r-ggplot2 >=2.0.0
+    - libgcc # [linux]
+
+test:
+  commands:
+    - metilene --help 2>&1 | grep VERSION
+    - unset PERL5LIB
+    - unset PERL_LOCAL_LIB_ROOT
```
bgruening: Oo this should not be needed. Why is this so. I guess this does not guarantee that the tool is working after installation.

jfallmann: No this should not be needed, and it does not guarantee that the tool is working after install, but only if one has set certain perl system variables to use local versions of modules.
These Variables like PERL5LIB, PERL_LOCAL_LIB_ROOT, PERL_CPANM_HOME are not handled correctly by conda, so even in an environment where I use conda-perl, the PERL5LIB still points to my local module dir, which results in a version clash. Bu unsetting these variables one can build and install and has to run the scripts with an additional -I $path_to_environment_perl_modules.
Not what I was expecting, but it seems a similar problem occurs to python users when setting their python module path to something not global, see [http://conda.pydata.org/docs/troubleshooting.html#conda-claims-installed](url).
The solution should not be to unset all these variables in my .bashrc, as this will render my setup completely useless and I still want to be able to use my system perl which is newer than the conda perl

bgruening: I have no clue about perl. @acaprez can you help here?  Thanks!

s-will:
Congratulations, @jfallmann for getting this running. Of course, the unsets in the test should not be necessary for successful testing on Travis (or with locally with docker). So, it would be good to remove them again to get a clean metilene package first.

Since conda seems to have general issues with "unclean" environments for scripting languages, it would be great to find more general solutions. One possibility would be to call scripts (in this case perl scripts but similar for other scripting languages) from wrapper scripts that clean the environment appropriately (another would be to patch the scripts). This could be done by the packager (@jfallmann, maybe you wanna try this for metilene or ViennaNGS) or ultimately by conda or bioconda in the process of package building (of course this would require careful consideration of pros and cons). In some way this reminds me of the patching of binaries as it is already done by conda (patchelf). What do you think @bgruening; were there already discussions/thoughts on this?

bgruening:
@s-will not that I know of. Imho a package manager can assume a clean environment and the admin should take care about a clean state. If the package should ensure a clean state you open Pandora's box. 
@jfallmann in your case can you not prepand something to the `source activate` script that cleans/reset your environment?

acaprez:
I've run into this issue before on our machines.  In our case it's users having local perl modules installed that were built with the system perl, which is older.  We actually use modules (Lmod) to make all software available, including Conda packages, in order to provide a uniform environment.  So we handle it at the Lmod level - basically unset PERL5LIB, etc. on module load and restore the original values on unload.  That approach is by no means perfect, but it does seem to work most of the time.   You could use the `etc/conda/activate.d` / `deactivate.d` functionality to essentially do the same thing per-package.  As @bgruening mentions though, the question is whether or not that's a good idea.

jfallmann:
Ok, @s-will @bgruening  i cleaned the recipe and it compiles, will work on a fix for the perl problem
@acaprez thanks for the hints, will play around and see what works best

recipes/metilene/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+  build:
+    - llvm # [osx]
+    - gcc # [linux]
+
+  run:
+    - perl-threaded
+    - bedtools >=2.24
+    - r-ggplot2 >=2.0.0
+    - libgcc # [linux]
+
```
bgruening: I think R is missing here.

bgruening:
Feel free to merge after you have added R as runtime dep. Thanks!

bgruening:
Nice! Thanks a lot! Beer for you both!

PR #1541, opened 2016-05-11
Adding recipe for AMOS

eclarke:

Request review by @bioconda/core 

AMOS hasn't been updated in a while, and the most recent released version contains a bug under more recent versions of GCC. Building directly from their SourceForge repo avoids this error, but adds about ~1.5GB of test data. To avoid this, I've cloned the source tree into my repo, pruned the test data, and gzipped it. This resulting tarball is what the recipe uses, but I'm willing to re-evaluate that method if it's not a good approach.


bgruening:
@eclarke maybe you can delete the test-data after installing it into $PREFIX?

eclarke:
Thanks for the suggestion, @bgruening. I've removed the local tarball dependency so now it builds directly from their SourceForge repository.

recipes/amos/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  summary: 'A Modular, Open-Source whole genome assembler'
+  license: Artistic License
+
+source:
+  git_url: http://git.code.sf.net/p/amos/code
+  git_rev: 9bd658
+
+build:
+  number: 3
+  # OSX failing due to missing dependencies in OSX branch
```
bgruening: Not for this PR, but what packages are missing? Can we activate them?

recipes/amos/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+requirements:
+  build:
+    - gcc
+    - zlib
+    - boost 
+    - autoconf
+    - blat
+    - mummer
+    - perl-statistics-descriptive
+    - perl-xml-parser
```
bgruening: I guess you also need to depend on `perl-threaded`.
@bioconda/core do you think this is necessary or is the implicit dependency on perl enough? I'm more towards explicit.

bgruening:
This looks very good! I will would like to wait a little bit and see if someone has an opinion on the perl question.

eclarke:
Thanks! I have no problem adding an explicit dependency to `perl-threaded`.

recipes/amos/meta.yaml
johanneskoester: Yes, definitely explicit. For python this is even mandatory, because the pinning does not work otherwise. Let's just do the same with perl.

bgruening:
Thanks!

PR #1545, opened 2016-05-12
Adding treemix recipe

hardingnj:

Potential problem with gsl

### DIFF
--- a/recipes/treemix/meta.yaml
+++ b/recipes/treemix/meta.yaml
@@ -1,0 +1,29 @@
+package:
+  name: treemix
+  version: 1.12
+
+build:
+  number: 0
+  skip: True # [osx]
+
+source:
+  fn: treemix-1.12.tar.gz
+  url: https://bitbucket.org/nygcresearch/treemix/downloads/treemix-1.12.tar.gz
+  md5: 183f605751a0fa445e249db840907104
+
+requirements:
+  build:
+  - gsl >=1.6
+  - boost >=1.42
+
+  run:
+  - gsl >=1.16
+  - boost >=1.42
+
+about:
+  home: http://pritchardlab.stanford.edu/software.html
+  license: GNU GENERAL PUBLIC LICENSE v3
+  summary: TreeMix is a method for inferring the patterns of population splits and mixtures in the history of a set of populations
+test:
+  commands:
+    - treemix --help | grep 1.12
--- a/recipes/treemix/meta.yaml
+++ b/recipes/treemix/meta.yaml
@@ -13,11 +13,15 @@ source:
 
 requirements:
   build:
+  - libgcc
+  - gcc ==4.8.5
   - zlib
   - gsl >=1.6
   - boost >=1.42
 
   run:
+  - libgcc
+  - gcc ==4.8.5
   - zlib
   - gsl >=1.16
   - boost >=1.42
--- a/recipes/treemix/meta.yaml
+++ b/recipes/treemix/meta.yaml
@@ -5,6 +5,8 @@ package:
 build:
   number: 0
   skip: True # [osx]
+  string: boost{{CONDA_BOOST}}__gsl{{CONDA_GSL}_{{PKG_BUILDNUM}}
+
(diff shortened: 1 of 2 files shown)


hardingnj:
@bioconda/core Having some trouble with zlib and gsl, something I am missing perhaps?

kyleabeauchamp:
Maybe because it's building GSL with CPPFlags set but not CFLAGS.  AFAIK GSL is a C library so maybe it's not smart enough to find ZLIB without CFLAGS?

hardingnj:
Thanks. Still no joy. Do you mean to set both CFLAGS and CPPFLAGs?

snewhouse:
I'm still struggling with this. same problems. need to learn more programming skills...sure it's a simple thing we're missing. sorry for the unhelpful reply

hardingnj:
Fixed. --with-boost=${PREFIX} was the missing piece. I had also mis-specified LDFLAG. 
Can I get a review from @bioconda/core before merge please?

recipes/treemix/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+source:
+  fn: treemix-1.12.tar.gz
+  url: https://bitbucket.org/nygcresearch/treemix/downloads/treemix-1.12.tar.gz
+  md5: 183f605751a0fa445e249db840907104
+
+requirements:
+  build:
+  - libgcc
+  - gcc ==4.8.5
```
bgruening: Can we just specify gcc here. The entire GCC question is not finally cleared but I would vote to just use gcc for the time being.

bgruening:
@hardingnj awesome! Can you look at the GCC stuff please.

recipes/treemix/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - zlib
+  - gsl >=1.6
+  - boost >=1.42
```
bgruening: Can you pin the GCC version with: 
```
- boost {{CONDA_BOOST}}*
```

recipes/treemix/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  fn: treemix-1.12.tar.gz
+  url: https://bitbucket.org/nygcresearch/treemix/downloads/treemix-1.12.tar.gz
+  md5: 183f605751a0fa445e249db840907104
+
+requirements:
+  build:
+  - libgcc
+  - gcc ==4.8.5
+  - zlib
+  - gsl >=1.6
```
bgruening: Can you pin gsl please:
```
- gsl {{CONDA_GSL}}*
```

recipes/treemix/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  skip: True # [osx]
+
+source:
+  fn: treemix-1.12.tar.gz
+  url: https://bitbucket.org/nygcresearch/treemix/downloads/treemix-1.12.tar.gz
+  md5: 183f605751a0fa445e249db840907104
+
+requirements:
+  build:
+  - libgcc
```
bgruening: libgcc is not needed as build. It ***can** be needed as runtime.

recipes/treemix/meta.yaml
johanneskoester: Please see the C++ section of the GUIDELINES.md. It points to an example recipe.

hardingnj:
Sorry for update delay. Have fixed those issues. Am I ok to merge?

recipes/treemix/meta.yaml
```diff
@@ -0,0 +1,34 @@
+package:
+  name: treemix
+  version: 1.12
+
+build:
+  number: 0
+  skip: True # [osx]
```
johanneskoester: You need to add a build string, similar to the salmon recipe. Otherwise, conda won't be able to detect that the package needs a rebuild when CONDA_GSL or CONDA_BOOST are updated.

bgruening:
Feel free to merge - looks very nice!

PR #1569, opened 2016-05-15
Update seqtk version and add explicit zlib dependencies

kyleabeauchamp:

cc @bgruening  this should be similar to your bwa changes.

### DIFF
--- a/recipes/seqtk/build.sh
+++ b/recipes/seqtk/build.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+export C_INCLUDE_PATH=${PREFIX}/include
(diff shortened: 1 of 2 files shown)


bgruening:
Awesome @kyleabeauchamp !
I hope it will work, it does not work for every package: https://github.com/bioconda/bioconda-recipes/pull/1568

bgruening:
@kyleabeauchamp can you try: `export CPLUS_INCLUDE_PATH=${PREFIX}/include`

kyleabeauchamp:
It's not that, it's that one of the binaries in seqtk was separated to a different repo.  I think my current commit should fix things.

PR #1578, opened 2016-05-16
Update clark to version 1.2.3

eclarke:


recipes/clark/1.2.2-b/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+#strictly use anaconda build environment
+CC=${PREFIX}/bin/gcc
```
bgruening: I don't think this is necessary. Is it?

eclarke: I don't know, probably not. I left it in because it was in the previous version of the recipe, but I can try it without.

druvus: It should not be needed. It is my fault that they are there but I haven't had time to remove the lines.

eclarke:
@druvus @bgruening thanks for the comments. It seems to work fine without the flags so I think it's ready to be merged.

bgruening:
@eclarke thanks for your work here! This is fantastic!

PR #1586, opened 2016-05-18
dammit plus dependencies

simonvh:

Adds the following recipes (dammit plus dependencies):

* busco
* transdecoder
* emboss
* doit
* nose-capturestderr
* ficus
* pyinotify

This is a first check (partly to see if it builds on OS X).


simonvh:
@bioconda/core two things that I'm unsure about:
1) Is it OK to specify a libgcc dependency with a specific version (d6d137f)?
2) There is no way to specify conda packages of Ruby gems, right? Currently I added a `gem install` command in the `build.sh` (d06f2e0), as a pragmatic way to install this gem.

recipes/busco/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+mkdir -p $PREFIX/bin/
+cp BUSCO_v1.2.py $PREFIX/bin
+ln -s $PREFIX/bin/BUSCO_v1.2.py $PREFIX/bin/BUSCO_v1.1b1.py
```
johanneskoester: I don't get the purpose of line number 5. Can you explain?

recipes/dammit/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+    - numpy
+    - pandas
+    - numexpr >=2.3.1
+    - khmer >=2.0
+    - sphinx >1.3.1
+    - sphinx_rtd_theme >=0.1.9
+    - nose ==1.3.4
+    - nose-capturestderr ==1.0
+    - ficus >=0.1
+    - doit >=0.29.0
```
johanneskoester: Do above versions come from the package itself?

recipes/emboss/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+source:
+  fn: EMBOSS-6.5.7.tar.gz
+  url: ftp://emboss.open-bio.org/pub/EMBOSS/old/6.5.0/EMBOSS-6.5.7.tar.gz
+  md5: 6a2cb3f93d5e9415c74ab0f6b1ede5f0
+
+requirements:
+  build:
+    #
+  run:
+    #
```
johanneskoester: If there are really no dependencies, you can remove the entire requirements section.

recipes/transdecoder/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  skip: True # [osx]
+
+source:
+  fn: transdecoder-2.1.0.tar.gz
+  url: https://github.com/TransDecoder/TransDecoder/archive/v2.1.0.tar.gz
+  md5: f86c688e13ced3cc51fed3d508a2ea05
+
+requirements:
+  build:
+    #
```
johanneskoester: Can you also add gcc here please?

recipes/transdecoder/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+source:
+  fn: transdecoder-2.1.0.tar.gz
+  url: https://github.com/TransDecoder/TransDecoder/archive/v2.1.0.tar.gz
+  md5: f86c688e13ced3cc51fed3d508a2ea05
+
+requirements:
+  build:
+    #  
+  run:
+    - libgcc >=4.9.1 # GOMP_4.0 requirement
```
johanneskoester: This is ok I think. libgcc is backward compatible, so other package depending on libgcc should still work, even if this leads to a newer libgcc being pulled.

recipes/dammit/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+gem install crb-blast
```
johanneskoester: Where will the gem be installed? It is important that it is installed into $PREFIX. If that is ensured, the approach would be ok.

recipes/busco/build.sh
simonvh: Sure! Apparently the `BUSCO` package comes with an executable that includes the version in the name. The `dammit` package in turn depends on a specific `BUSCO` executable. So this is how I solved that. Another way, which might be more elegant, is to symlink the exutable to a version independent name (ie `BUSCO.py`) and than patch the `dammit` code to run this executable. I admit to taking the easy way out here...

recipes/dammit/meta.yaml
simonvh: Yup, these are the requirements from the `dammit` docs.

recipes/emboss/meta.yaml
simonvh: Will do.

simonvh:
Thanks for the review Johannes.

recipes/transdecoder/meta.yaml
simonvh: Will do.

recipes/dammit/build.sh
simonvh: I think it will be, as the `ruby` is a $PREFIX version. I will double-check to make sure.

recipes/busco/build.sh
johanneskoester: I must say that I prefer to have an executable called ``busco``. Do you think this would be feasible? At least, I definitly don't want busco 1.2 to install a file indicating version 1.1b1 into bin.

recipes/dammit/build.sh
johanneskoester: Ok. Of course, it would be nice to have a separate package for crb-blast then.

recipes/dammit/meta.yaml
johanneskoester: Ok, that's fine.

recipes/busco/build.sh
simonvh: Yes, it would be possible. I forgot why I even created this hacky solution, instead of just installing BUSCO 1.1b1, but now I'm looking at it I remember that it has to with Python incompatibilities as 1.1b1 doesn't run on  2.7, which in turn is needed for some other dependencies. Anyway, I'll look into fixing this.

recipes/transdecoder/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  skip: True # [osx]
+
+source:
+  fn: transdecoder-2.1.0.tar.gz
+  url: https://github.com/TransDecoder/TransDecoder/archive/v2.1.0.tar.gz
+  md5: f86c688e13ced3cc51fed3d508a2ea05
+
+requirements:
+  build:
+    - gcc
```
simonvh: @johanneskoester There is no `gcc` version that satisfies in `>= 4.9.1` in the current channels, even if there is `libgcc 5.2.0`. This seems to work fine as as now.

johanneskoester: Well, this is weird :-). How come it compiles with the gcc < 4.9.1 then?

simonvh: Beats me :). I just know that `libgcc-5.2.0` fixes a run-time problem with `transdecoder`, but that it compiles fine with `gcc-4.8.5`.

johanneskoester: Ok, then let's keep it like that.

PR #1592, opened 2016-05-18
Adding recipe for hapbin

hardingnj:

Uses gzipped files for tests


hardingnj:
Only thing I am not sure about is the tests. 

Authors provide some test files for running tool. They need to be decompressed and tool to run. Takes <20s. 

- Is this too long?
- Is the use of find/unzip ok, or should that be in `build.sh`. I figured here, as no point gunzipping if tests not being run.

johanneskoester:
For testing, it is sufficient to check wether e.g. --version of --help works. We don't want to reproduce real test cases here.

hardingnj:
Done. Think I am ok to merge pending tests?

PR #1607, opened 2016-05-19
Fix for openssl issue during build.

roryk:

This fixes the local building for me, it wasn't picking up openssl during the build. I could be a moron though, so a check would be a good plan before merging.

### DIFF
--- a/recipes/ucsc-gff3togenepred/include.patch
+++ b/recipes/ucsc-gff3togenepred/include.patch
@@ -5,7 +5,7 @@
  
  HG_DEFS=-D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -D_GNU_SOURCE -DMACHTYPE_${MACHTYPE}
 -HG_INC+=-I../inc -I../../inc -I../../../inc -I../../../../inc -I../../../../../inc
-+HG_INC+=-I../inc -I../../inc -I../../../inc -I../../../../inc -I../../../../../inc -I /anaconda/include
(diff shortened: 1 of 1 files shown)


bgruening:
It makes a lot sense to me. I just don't understand why this worked before. I will leave this open so other can chime in.

roryk:
Me too for both statements. :)

recipes/ucsc-gff3togenepred/include.patch
```diff
@@ -5,7 +5,7 @@
  
  HG_DEFS=-D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -D_GNU_SOURCE -DMACHTYPE_${MACHTYPE}
 -HG_INC+=-I../inc -I../../inc -I../../../inc -I../../../../inc -I../../../../../inc
-+HG_INC+=-I../inc -I../../inc -I../../../inc -I../../../../inc -I../../../../../inc -I /anaconda/include
++HG_INC+=-I../inc -I../../inc -I../../../inc -I../../../../inc -I../../../../../inc -I /anaconda/include -I ${PREFIX}/include
```
johanneskoester: Do we really need the /anaconda/include? I would prefer the recipe to be independent of our setup. I think $PREFIX/include should be sufficient.

roryk: Sweet, deleting unnecessary stuff sounds good to me.

johanneskoester:
Looks good, thanks!

PR #1620, opened 2016-05-23
New: gvcf-regions Update: vardict-java

chapmanb:

- vardict-java fixes out of range errors
- gvcf-regions converts gVCF output files into callable regions

### DIFF
--- a/recipes/gvcf-regions/build.sh
+++ b/recipes/gvcf-regions/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+mkdir -p $PREFIX/bin
+sed -i '1i#!/opt/anaconda1anaconda2anaconda3/bin/python' gvcf_regions.py
+chmod a+x gvcf_regions.py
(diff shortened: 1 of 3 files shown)


recipes/gvcf-regions/build.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 mkdir -p $PREFIX/bin
-sed -i '1i#!/opt/anaconda1anaconda2anaconda3/bin/python' gvcf_regions.py
+sed -i $'1i\\\n#!/opt/anaconda1anaconda2anaconda3/bin/python\n' gvcf_regions.py
```
croth1: I think on OS X, using -i always requires a prefix. So `sed -i.bak` should work.

chapmanb: Thanks so much for the reminder on this. I pushed a new version of the recipe which should, fingers crossed, support OSX correctly now with sed. Thanks again.

PR #1643, opened 2016-05-27
MGLTools Package

leobiscassi:
@bgruening I fixed the problems mentioned previously, can you review this PR?

### DIFF
--- a/recipes/mgltools/meta.yaml
+++ b/recipes/mgltools/meta.yaml
@@ -2,12 +2,14 @@ about:
     home: http://mgltools.scripps.edu/
     license_file: LICENSES
     summary: MGLTools is an program for visualization and analisys of molecular structures.
+
 package:
     name: mgltools
     version: 1.5.6
+
 build:
   number: 0
-  skip: False
+
 source:
     fn:  mgltools_x86_64Linux2_1.5.6.tar.gz                                                                             # [linux]
     url: http://mgltools.scripps.edu/downloads/downloads/tars/releases/REL1.5.6/mgltools_x86_64Linux2_1.5.6.tar.gz      # [linux]
@@ -15,11 +17,11 @@ source:
     fn:  mgltools_i86Darwin9_1.5.6.tar.gz                                                                               # [osx]
     url: http://mgltools.scripps.edu/downloads/downloads/tars/releases/REL1.5.6/mgltools_i86Darwin9_1.5.6.tar.gz        # [osx]
     md5: c515b187cddeec72fe5a81ecb4a92d6b                                                                               # [osx]
+
 requirements:
   build:
-    - gcc  # [linux]
-    - llvm # [osx]
   run:
+
 test:
     commands:
       - pmv --help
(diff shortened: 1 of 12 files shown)


recipes/mgltools/meta.yaml
```diff
@@ -0,0 +1,25 @@
...
+    url: http://mgltools.scripps.edu/downloads/downloads/tars/releases/REL1.5.6/mgltools_i86Darwin9_1.5.6.tar.gz        # [osx]
+    md5: c515b187cddeec72fe5a81ecb4a92d6b                                                                               # [osx]
+requirements:
+  build:
+    - gcc  # [linux]
+    - llvm # [osx]
+  run:
+test:
+    commands:
+      - pmv --help
```
bgruening: Can you test please your two python scripts here as well?

recipes/mgltools/meta.yaml
```diff
@@ -0,0 +1,25 @@
...
+    summary: MGLTools is an program for visualization and analisys of molecular structures.
+package:
+    name: mgltools
+    version: 1.5.6
+build:
+  number: 0
+  skip: False
+source:
+    fn:  mgltools_x86_64Linux2_1.5.6.tar.gz                                                                             # [linux]
+    url: http://mgltools.scripps.edu/downloads/downloads/tars/releases/REL1.5.6/mgltools_x86_64Linux2_1.5.6.tar.gz      # [linux]
```
bgruening: Are these source packages, or binary distributions? If you download two separate binaries you should use the template syntax to specify when to use which URL. Eg. # [osx].

Compiling from source in encouraged if possible :)

leobiscassi: Hi @bgruening , about the template syntax, I made it.

``` yml
source:
fn:  mgltools_x86_64Linux2_1.5.6.tar.gz # [linux]
url: http://mgltools.scripps.edu/downloads/downloads/tars/releases/REL1.5.6/mgltools_x86_64Linux2_1.5.6.tar.gz # [linux]
md5: e3f62852c4d442a292dedbe6c79540ca # [linux]
fn:  mgltools_i86Darwin9_1.5.6.tar.gz  # [osx]
url: http://mgltools.scripps.edu/downloads/downloads/tars/releases/REL1.5.6/mgltools_i86Darwin9_1.5.6.tar.gz # [osx]
md5: c515b187cddeec72fe5a81ecb4a92d6b # [osx]
```
There are anything wrong?
This is a binary distribution. Thanks for your suggestions! :)

recipes/mgltools/meta.yaml
leobiscassi: Ok, I'll make it, thanks for suggestion! :)

bgruening:
If these packages are precompiled and you do not compile them with the install script you can remove the GCC/llvm. Sorry for the confusion I can hardly check from my mobile.

leobiscassi:
Hi, I analysed the install script and makes sense execute only one installation because is the exactly the same script. 
About the packages, if flag (-c) was used *.pyc or *.pyo will be generated, but in this case I don't use this flag, the requirements still makes sense?
Sorry for the confusion too.

PR #1645, opened 2016-05-28
CrossMap: added samtools dependency

yhoogstrate:

This is an update for CrossMap which uses samtools  under the hood. Samtools has changed certain arguments in 1.0.0 and are incompatible with CrossMap, which is the reason I fixed the version.


bgruening:
Youri can you please increase the build-number otherwise the tests will not run.

yhoogstrate:
Thank you Björn

bgruening:
Thanks!

PR #1655, opened 2016-05-31
New recipe: moca

saketkc:

### DIFF
--- a/recipes/moca/bld.bat
+++ b/recipes/moca/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
+:: http://docs.continuum.io/conda/build.html
+:: for a list of environment variables that are set during the build process.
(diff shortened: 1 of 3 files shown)


saketkc:
@johanneskoester I am not sure I understand the DAG 0 failure. Is somehow the recipe being bumped above its dependencies?

johanneskoester:
Let's do it step by step. In the osx build, pycairo is missing, because it is not available for osx in the default channel. I guess there is a good reason for this. Hence, if this is a necessary dependency of moca, you should add a `skip: True # [osx]` to the build section.

johanneskoester:
The error in subdag 0 is indeed weird, and I don't understand it at all. @bioconda/core, has anybody seen something like this in another PR? Is it just too late and I am missing something obvious?

tomkinsc:
I may have encountered a similar strange error before on my local machine. If I remember right, the cause was that some of the requirements had their own requirements, some of which were incompatible (pinned to a specific version rather than an open range, and conda couldn't resolve the dependencies), and conda yielded that error rather than its usual message about incompatible dependencies. Does the recipe build on your local machine, @saketkc?

saketkc:
Thanks @tomkinsc and @johanneskoester! It builds on my local machine, though only on py27. I need to check what's going on since I don't get any errors on travis: https://travis-ci.org/saketkc/moca for py3+. I have skipped building on `osx` though.

PR #1657, opened 2016-05-31
added dependency for lof plugin

matthdsm:

### DIFF
--- a/recipes/variant-effect-predictor/build.sh
+++ b/recipes/variant-effect-predictor/build.sh
@@ -15,6 +15,8 @@ sed -i.bak 's@/convert_cache.pl@/vep_convert_cache.pl@' scripts/variant_effect_p
 sed -i.bak 's@use strict;@use strict;\nuse FindBin qw($RealBin);\nuse lib $RealBin;@' scripts/variant_effect_predictor/convert_cache.pl
 # Find plugins in install directory
 sed -i.bak 's@$config->{dir_plugins} ||=.*@$config->{dir_plugins} ||= $RealBin;@' scripts/variant_effect_predictor/variant_effect_predictor.pl
+# Change location where INSTALL.pl looks for the zlib headers
+sed -i -e "s@/usr/include/zlib.h@${PREFIX}/include@" scripts/variant_effect_predictor/INSTALL.pl
 cp scripts/variant_effect_predictor/convert_cache.pl $target/vep_convert_cache.pl
 cp scripts/variant_effect_predictor/INSTALL.pl $target/vep_install.pl
 cp scripts/variant_effect_predictor/variant_effect_predictor.pl $target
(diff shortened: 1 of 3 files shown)


recipes/variant-effect-predictor/build.sh
```diff
@@ -29,3 +29,4 @@ rm -rf t/
 wget --no-check-certificate https://raw.githubusercontent.com/ensembl-variation/VEP_plugins/master/CADD.pm
 wget --no-check-certificate https://raw.githubusercontent.com/ensembl-variation/VEP_plugins/master/dbNSFP.pm
 wget --no-check-certificate https://raw.githubusercontent.com/konradjk/loftee/master/LoF.pm
```
bgruening: Can we use here a specific revision for all scripts or modules?

matthdsm: I don't really think so, the ensembl-variation repo only has a specific branch for the latest  version of vep.
the loftee repo does have some version tags, but the last tag was made over a year ago, while production ready changes still get pushed to the master branch

bgruening: You could use such a link isn't it: https://raw.githubusercontent.com/Ensembl/VEP_plugins/435d16729a518a83b83787985987f8f2c2782e2c/dbNSFP.pm

matthdsm: true, but that doesn't really relate to some version does it? This just fixes to plugin to some arbitrary point in the history of the repo. How would we pinpoint the right commit?

bgruening: What is the `right commit`? The problem is that you don't know what you get with the current approach. You will get the script from the time of build. If you rebuild the package you can get an entire new script in worst case. With fixing a revision you are guaranteed to at least get every time the same script.

matthdsm: Ok, that's also a valid point. 
I'll just point the url's to to the latest commit on the master branch. The plugin url's can then be updated together with the vep version.

bgruening: Thansk!

matthdsm:
@bgruening 
Checks failed because zlib.h library header is missing. What do you propose I do about this? Do I create a new vep recipe? of is this something that's just missing in the Travis platform?

bgruening:
Depend on `- zlib` as build and runtime dependency. It can be that you need to add `export C_INCLUDE_PATH=$PREFIX/include` to your build.sh script as well.

Thanks!

matthdsm:
Ok, so I've added another sed statement to change the hardcoded path of the zlib library. we'll see how this goes

matthdsm:
Huraah! This seems to have fixed it!

bgruening:
Great work! Thanks!

matthdsm:
yay!

Issue #1519, opened 2016-05-10
Java wrapper script


bgruening:
Currently we use a bash script to wrap Java programs ins Bioconda. This is a port to python of the same script. I think this is more maintainable and we can define a strict dependency on python. Whereas the old wrapper depends on bash, which is not often available in minimal Docker containers and there is no dependency yet (alternatively we could write one).

I'm not sure this is worth the effort and defining python as dependency for a Java program also looks wired. On the other side this script should be more portable and maintainable.

ping @chapmanb @bioconda/core 

```
#!/usr/bin/env python
#
# Wrapper script for Java Conda packages that ensures that the java runtime
# is invoked with the right options. Adapted from the bash script (http://stackoverflow.com/questions/59895/can-a-bash-script-tell-what-directory-its-stored-in/246128#246128).
#

#
# Program Parameters
#

jar_file = 'PeptideShaker-1.1.3.jar'

default_jvm_mem_opts = ['-Xms512m', '-Xmx1g']

# !!! End of parameter section. No user-serviceable code below this line !!!


import sys
from os import access, getenv, system, X_OK
from os.path import dirname, realpath, join as join_path
from subprocess import call


def real_dirname(path):
    """Return the symlink-resolved, canonicalized directory-portion of path."""
    return dirname(realpath(path))


def java_executable():
    """Return the executable name of the Java interpreter."""
    java_home = getenv('JAVA_HOME')
    java_bin = join_path('bin', 'java')

    if java_home and access(join_path(java_home, java_bin), X_OK):
        return join_path(java_home, java_bin)
    else:
        return 'java'


def jvm_opts(argv):
    """Construct list of Java arguments based on our argument list.

    The argument list passed in argv must not include the script name.
    The return value is a 3-tuple lists of strings of the form:
      (memory_options, prop_options, passthrough_options)
    """
    mem_opts = []
    prop_opts = []
    pass_args = []

    for arg in argv:
        if arg[0:2] == '-D':
            prop_opts.append(arg)
        elif arg[0:3] == '-XX':
            prop_opts.append(arg)
        elif arg[0:3] == '-Xm':
            mem_opts.append(arg)
        else:
            pass_args.append(arg)

    # In the original shell script the test coded below read:
    #   if [ "$jvm_mem_opts" == "" ] && [ -z ${_JAVA_OPTIONS+x} ]
    # To reproduce the behaviour of the above shell code fragment
    # it is important to explictly check for equality with None
[...]

PR #1665, opened 2016-06-01
fix logic in selecting java version

dakl:

Previously java would be set if JAVA_HOME was UNset, which I assume was not the intended behaviour. This would cause picard to fail if the system installation was not the correct version to use. 

For me, picard 2.3.0 used `/bin/java` which was version 1.7.x instead of the version installed by conda.

### DIFF
--- a/recipes/picard/picard.sh
+++ b/recipes/picard/picard.sh
@@ -19,7 +19,8 @@ JAR_DIR=$DIR
 
 java=java
 
-if [ -z "${JAVA_HOME:=}" ]; then
+# if JAVA_HOME is set (non-empty), use it. Otherwise keep "java"
+if [ ! -z "${JAVA_HOME:=}" ]; then
   if [ -e "$JAVA_HOME/bin/java" ]; then
       java="$JAVA_HOME/bin/java"
   fi


dakl:
As I'm myself a little confused of the correctness of this, I'm tagging @bioconda/core.

brainstorm:
This looks like a good extra "robustness" check before calling java, Daniel. Nothing wrong with it AFAICT, other than the TravisCI stuff you reported:

dakl:
Yes, that's a strange one. There is one test set in `meta.yaml`:

```

test:
  commands:
    - "picard 2>&1 | grep 'USAGE: PicardCommandLine'"
```

bgruening:
@dakl this is a nice addition and many other tools would profit from it I guess.
As a related note what do you think about this one: https://github.com/bioconda/bioconda-recipes/issues/1519
Want to work on this :) ?

dakl:
@bgruening sure thing!

dakl:
Even though this is merged, I'm still getting the old `picard.sh` when I `conda install -c bioconda picard=2.3.0`. Any idea why?

PR #1668, opened 2016-06-01
Use the new stable version of conda-build instead of our fork.

johanneskoester:

With the new stable release, we should be able to use upstream conda-build again.


bgruening:
Your previous changes were removed, is this intended?

PR #1669, opened 2016-06-01
Recipe for Augustus

jerowe:

Hello,

Augustus is a software available here:
http://bioinf.uni-greifswald.de/augustus/downloads/

It does RNASeq annotation


recipes/augustus/3.2.2/meta.yaml
```diff
@@ -0,0 +1,119 @@
...
+  skip: True # [osx]
+requirements:
+  build:
+      - libgcc
+      - gcc
+      - bamtools
+      - perl-app-cpanminus
+      - perl-yaml
+      - perl-dbi
+      - boost
```
bgruening: Can you please pin your boost dependency like here: https://github.com/bioconda/bioconda-recipes/blob/28f764787310141bc74bf4ee90ace6432b9fa6b0/recipes/salmon/meta.yaml#L21
Please also not the build-string change.

recipes/augustus/3.2.2/build.sh
```diff
@@ -0,0 +1,26 @@
...
+make
+mkdir -p $PREFIX/bin
+mkdir -p $PREFIX/scripts
+mkdir -p $PREFIX/config
+mv bin/* $PREFIX/bin/
+mv scripts/* $PREFIX/scripts/
+mv config/* $PREFIX/config/
+
+#Install perl reqs
+cpanm Cwd File::Basename FileHandle File::Path File::Spec::Functions Getopt::Long IO::File List::Util Term::ANSIColor
```
bgruening: Ideally we would have this as extra packages.

recipes/augustus/3.2.2/meta.yaml
jerowe: Thanks! I was looking through the recipes to see how people were handling boost, but apparently not enough. I'm making the changes now.

recipes/augustus/3.2.2/build.sh
jerowe: What are you thinking? An augustus-deps package?

I'm not sure what the gain is on something like that. Ultimately I want to be to hand my users one command, 

`conda install -c bioconda my-thing`

 and then I want for them to actually be able to use it. If its missing all kinds of dependencies I will have a million tickets for things that aren't installed. 

`conda install -c bioconda my-thing my-thing-deps`

Would also accomplish this. ;-)

bgruening: Creating sinlge packages and depending on it, like you did here `- perl-yaml`.
I know this is a pain with perl, but this would be awesome to have.

jerowe: Aaah I gotcha. I thought you meant a separate package for the dependencies. I'll work on that.

recipes/augustus/3.2.2/meta.yaml
jerowe: So I'm having trouble getting it compiled without the boost libs.

I made the changes to the meta.yaml as in salmon.

This is the output from building on the bioconda image.
`+ make
In file included from genbank.cc:15:0:
../include/genbank.hh:26:48: fatal error: boost/iostreams/filtering_stream.hpp: No such file or directory
 #include <boost/iostreams/filtering_stream.hpp>
`

If I don't include these two lines in my build.sh:

export CXXFLAGS="-DUSE_BOOST -I${BOOST_INCLUDE_DIR} -L${BOOST_LIBRARY_DIR}"
export LDFLAGS="-L${BOOST_LIBRARY_DIR}"

jerowe: @bgruening 

When I try to pin the boost dep I get the following error on building

`
Extracting download
Applying patch: '/bioconda-recipes/recipes/augustus/3.2.2/patches/bam2hints.Makefile.patch'
patching file auxprogs/bam2hints/Makefile
Applying patch: '/bioconda-recipes/recipes/augustus/3.2.2/patches/filterBam.src.Makefile.patch'
patching file auxprogs/filterBam/src/Makefile
Applying patch: '/bioconda-recipes/recipes/augustus/3.2.2/patches/filterBam.Makefile.patch'
patching file auxprogs/filterBam/Makefile
Error: Failed to render jinja template in /bioconda-recipes/recipes/augustus/3.2.2/meta.yaml:
'CONDA_BOOST' is undefined
`

I just did a docker build, so I don't think that's the issue.

recipes/augustus/3.2.2/build.sh
jerowe: @bgruening 

The conda skeleton cpan has a recursive option, and I'm working on a quick bootstrap to use that to add the perl modules, but I'm getting an error with Test::Deep. I've reported the error to conda build here:  [https://github.com/conda/conda-build/issues/994](conda-build 994)

recipes/augustus/3.2.2/meta.yaml
bgruening: Have you used the latest mast branch. We added CONDA_BOOST a month back or so. It's defined here: https://github.com/bioconda/bioconda-recipes/blob/master/scripts/env_matrix.yml
and it is working for other recipes.

jerowe: Ah. That seemed so obvious once you said it. ;-) Thanks so much for all your help!

PR #1660, opened 2016-05-31
bump cyvcf2

brentp:
* [x] I have read the guidelines above.
* [ ] This PR adds a new recipe.
* [x] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

PR #1676, opened 2016-06-02
Update guidelines

daler:

update guidelines with gcc, llvm, and zlib info

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -109,10 +109,37 @@ the `meta.yaml` files in above examples).
 In general, standard `make` should work. Other build tools (e.g., `autoconf`)
 and compilers (e.g., `gcc`) should be specified in the build requirements.
 
+We have not yet decided whether to have `gcc` as a conda package or to assume
+it is in the build environment. Until this decision is made, please add `gcc`
+(for Linux packages) and `llvm` (for OSX packages) to the `meta.yaml` as
+follows:
+
+```yaml
+requirements:
+  build:
+    - gcc   # [not osx]
+    - llvm  # [osx]
+
+  run:
+    - libgcc    # [not osx]
+```
+
+If the package uses `zlib`, then please see the `ZLIB` section below.
+
 * example requiring `autoconf`: [srprism](recipes/srprism)
 * simple example: [samtools](recipes/samtools)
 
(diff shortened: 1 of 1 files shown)


GUIDELINES.md
```diff
@@ -109,10 +109,37 @@ the `meta.yaml` files in above examples).
...
+follows:
+
+```yaml
+requirements:
+  build:
+    - gcc   # [not osx]
+    - llvm  # [osx]
+
+  run:
+    - libgcc    # [not osx]
```
bgruening: `libgcc` is not always needed.

GUIDELINES.md
```diff
@@ -200,3 +227,41 @@ cluttering the output in the Travis-CI build environment.
...
+```
+
+When building the package, you may get an error saying that zlib.h can't be
+found -- despite having zlib listed in the dependencies. This is becuase the
+location of `zlib` often has to be specified in the `build.sh` script, which
+can be done like this:
+
+```bash
+export CFLAGS="-I$PREFIX/include"
+export LDFLAGS="-L$PREFIX/lib"
```
bgruening: This is also not needed at any time. I would say in most cases it is not needed.

daler: Which is not needed? 

Seems like e.g. [verse](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/verse/build.sh) needed the CPATH;  [samtools](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/samtools/build.sh) and [cyvcf2](https://github.com/bioconda/bioconda-recipes/pull/1660/files) needed the CFLAGS and LDFLAGS. Since it's a difficult (for me at least) to resolve problem, I figured this was a good place to put the solution.

Is there a rule of thumb we could provide to suggest when this would or wouldn't be needed?

GUIDELINES.md
daler: Is there a rule of thumb to figure out if libgcc is needed? Maybe the strategy should be to first try without libgcc, then if tests fail then add it back in?

bgruening: I don't now of any rule of thumb. Your idea seems good to me.

GUIDELINES.md
bgruening: In many cases `export C_INCLUDE_PATH=$PREFIX/include` is sufficient.  If this gets overwritten in some Makefile, `CFLAGS="-I$PREFIX/include"` is equivalent. But his one also get purged `CFLAGS=` in many Makefiles. LDFLAGS and CPPFLAGS are not so frequently used to fix these issue as far as I know.

PR #1683, opened 2016-06-02
add (installer for) gatk-protected (full, license-restricted GATK)

tomkinsc:
The license for the full version of GATK prevents it from being redistributed. This adds a recipe for a GATK package that provides the commands “gatk” and “gatk-register”. The former is a thin wrapper around the gatk.jar file, and the latter is an installer that copies a user-supplied GenomeAnalysisTK.jar file (or unpacked tar.bz2 file) to the conda environment. It includes helpful messages that instruct the user on what to do.

### DIFF
--- a/recipes/gatk/build.sh
+++ b/recipes/gatk/build.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+BINARY_HOME=$PREFIX/bin
+PACKAGE_HOME=$PREFIX/opt/$PKG_NAME-$PKG_VERSION
+
+mkdir -p $PREFIX/bin
+mkdir -p $PACKAGE_HOME
+
+cp $RECIPE_DIR/gatk.sh $PACKAGE_HOME/gatk.sh
+
+SOURCE_FILE=$RECIPE_DIR/gatk-register.sh
+DEST_FILE=$PACKAGE_HOME/gatk-register.sh
+
+
+echo "#!/bin/bash" > $DEST_FILE
+echo "PKG_NAME=$PKG_NAME" >> $DEST_FILE
+echo "PKG_VERSION=$PKG_VERSION" >> $DEST_FILE
+
+cat $SOURCE_FILE >> $DEST_FILE
+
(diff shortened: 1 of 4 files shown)


tomkinsc:
@bioconda/core, please look over this recipe and let me know what you think.
This address #1247, based on an initial [description](https://github.com/bioconda/bioconda-recipes/issues/1247#issuecomment-208547450) by @johanneskoester.

chapmanb:
Chris -- nice one. Thanks for doing this. The only thing I noticed was that GATK 3.6 is java 8 friendly so we can remove the pin on java-jdk in the requirements. Thanks again.

tomkinsc:
Nice catch, thanks. I've changed it to allow java-jdk >=7.

johanneskoester:
Great work, this is exactly what we needed!

recipes/gatk/build.sh
```diff
@@ -0,0 +1,25 @@
...
+mkdir -p $PACKAGE_HOME
+
+cp $RECIPE_DIR/gatk.sh $PACKAGE_HOME/gatk.sh
+
+SOURCE_FILE=$RECIPE_DIR/gatk-register.sh
+DEST_FILE=$PACKAGE_HOME/gatk-register.sh
+
+
+echo "#!/bin/bash" > $DEST_FILE
+echo "PKG_NAME=$PKG_NAME" >> $DEST_FILE
```
tomkinsc: Is there a better way to get the conda package variables into a non-build script?

johanneskoester: At least for me that looks ok.

tomkinsc:
OK to merge?

xguse:
This is seriously great. Thanks for your time on this!

yhoogstrate:
The gatk3-register message refers to the main download page that only shows GATK4. Would it make sense to refer to: https://software.broadinstitute.org/gatk/download/archive ?

tomkinsc:
@yhoogstrate: Yes, that makes sense. I've opened #14989 to change the URL.

PR #1710, opened 2016-06-08
gatk: Use anaconda java for gatk-register

chapmanb:

Ensures gatk-register uses the anaconda install java so it gets the
correct version, otherwise can fail with Java errors if 1.7 in
PATH before anaconda. (cc @tomkinsc)


tomkinsc:
Thanks for this, @chapmanb.

PR #1712, opened 2016-06-08
new stacks version 1.40

abretaud:

### DIFF
--- a/recipes/stacks/1.37/meta.yaml
+++ b/recipes/stacks/1.37/meta.yaml
@@ -1,13 +1,12 @@
 package:
   name: stacks
-  version: "1.40"
+  version: "1.37"
 build:
   number: 1
   skip: True # [osx]
 source:
-  fn: stacks-1.40.tar.gz
-  md5: 6e37053d3a5fcbbe2b421be9624b19fa
-  url: http://catchenlab.life.illinois.edu/stacks/source/stacks-1.40.tar.gz
+  fn: stacks-1.37.tar.gz
+  url: http://catchenlab.life.illinois.edu/stacks/source/stacks-1.37.tar.gz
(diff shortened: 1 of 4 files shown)


recipes/stacks/1.37/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+  version: "1.37"
+build:
+  number: 1
+  skip: True # [osx]
+source:
+  fn: stacks-1.37.tar.gz
+  url: http://catchenlab.life.illinois.edu/stacks/source/stacks-1.37.tar.gz
+requirements:
+  build:
+    - gcc
```
bgruening: can you please add # [linux] and # [osx] for llvm, same for libgcc

recipes/stacks/1.37/meta.yaml
```diff
@@ -0,0 +1,27 @@
+package:
+  name: stacks
+  version: "1.37"
+build:
+  number: 1
+  skip: True # [osx]
```
bgruening: If you skip osx you probably do not need to include llvm.
The build number should be 0 if this is the first version.

bgruening: Oh I see it's copied and already exists. In this case leave it like it is. Just saying for the next recipe the `# [linux]` would be nice and you should adjust the 1.40 version to it.

Thanks!

abretaud:
Hum, it seems like the Travis job is stuck (even after triggering a new one), I don't understand why...
Have you already experienced this kind of problem?

daler:
Depending on the load on the travis-ci servers sometimes it can take a while for the job to start. It's uncommon, but I've seen it take almost an hour to start a job.

PR #1715, opened 2016-06-08
Aria2 1.23.0

acaprez:

This adds the `aria2` package, and needed dependencies `libssh2` and `c-ares`.  

Question for @bioconda/core - is there a recommended way to do testing for library only packages?  The `libssh2` and `c-ares` packages are both library-only.  What I've done previously is just test to see if the main libs and header files exist, so I did that here as well.  Just wondering if there's an alternate way that's preferred?

### DIFF
--- a/recipes/aria2/build.sh
+++ b/recipes/aria2/build.sh
@@ -2,6 +2,15 @@
 
 export PKG_CONFIG_PATH=${PREFIX}/lib/pkgconfig
 
-./configure --prefix=${PREFIX} --with-libz --with-libxml2 --with-libssh2 --with-libcares --with-sqlite3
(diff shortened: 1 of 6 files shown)


johanneskoester:
That sounds reasonable to me. I would do it the same way.

acaprez:
Ok - thanks.  Good to merge this then?

johanneskoester:
Yes, looks good! Thanks!

bgruening:
I have seen very tiny files which you compile in the test to test the library.
On Jun 8, 2016 7:22 PM, Johannes Köster <notifications@github.com> wrote:That sounds reasonable to me. I would do it the same way.

PR #1727, opened 2016-06-10
batch 13 of ucsc updates

daler:

### DIFF
--- a/recipes/ucsc-pslcat/include.patch
+++ b/recipes/ucsc-pslcat/include.patch
@@ -2,10 +2,10 @@
 +++ kent/src/inc/common.mk
 @@ -17,7 +17,7 @@
  endif
- 
+
  HG_DEFS=-D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -D_GNU_SOURCE -DMACHTYPE_${MACHTYPE}
 -HG_INC+=-I../inc -I../../inc -I../../../inc -I../../../../inc -I../../../../../inc
-+HG_INC+=-I../inc -I../../inc -I../../../inc -I../../../../inc -I../../../../../inc -I /anaconda/include
- 
++HG_INC+=-I../inc -I../../inc -I../../../inc -I../../../../inc -I../../../../../inc -I ${PREFIX}/include
+
  # to check for Mac OSX Darwin specifics:
  UNAME_S := $(shell uname -s)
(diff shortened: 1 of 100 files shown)


daler:
@bioconda/core does this mean we're out of space?

https://travis-ci.org/bioconda/bioconda-recipes/jobs/136724166

kyleabeauchamp:
If we are, it might be good at some point to clear out all the old builds of recipes.  I think a lot of "semi-broken" builds have been uploaded over the years.  For example, if one forgets an optional dependency or a linker option and then increments the build number.

daler:
Is there a good way to identify those builds?

kyleabeauchamp:
I think there's a command line client to probe the anaconda cloud, so we could just write a script to enumerate the builds and keep track of the most recent build for each version / OS.

bgruening:
I'm not a fan of removing all builds and only keeping the latest. This will break all pipelines that uses a requirements.txt file with strict dependencies for reproducibility.
If we are really over quota we have a more substantial problem to solve. We are not a year old and have filled 100GB, either we restrict ourself and do not accept any new packages which is a shame especially considering how fast this community is growing or we are bagging Continuum to provide us more storage. 

We could also ask other projects to give us more storage or we can start collection money. Removing old packages will not solve the problem it can postpone it maybe,

johanneskoester:
We can ask continuum for more space at some point. However, at the moment, we use only 18GB of the 100GB.
Hence, the error must mean something else. Maybe a temporary problem?

johanneskoester:
Apart from that, requirements.txt or environment.yaml should maybe not point to build numbers, only to version numbers. With `conda env export`, there is the `--no-builds` flag that achieves this.

bgruening:
> Apart from that, requirements.txt or environment.yaml should maybe not point to build numbers, only to version numbers. With `conda env export`, there is the `--no-builds` flag that achieves this.

This is not the default and there are reasons for it I guess :).
An other argument is the build-number per se, if this would not be needed we should always overwrite the version and not require a bump of this number. I'm happy like it is and more importantly I'm really happy that we only use 18GB currently :)

daler:
If we're not close to quota I'll just try restarting the build then. Thanks.

daler:
Sorry for the noise, it seems to have been some temporary issue with the anaconda servers. Everything looks fine here now.

PR #1730, opened 2016-06-10
Update canu to version 1.3

jashapiro:

### DIFF
--- a/recipes/canu/meta.yaml
+++ b/recipes/canu/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: canu
-  version: 1.1
+  version: 1.3
 
 source:
   # Canu currently needs to come from git to get its version string right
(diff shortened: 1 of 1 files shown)


bgruening:
@jashapiro thanks for the update. Unfortunately it is failing on OSX because of missing OpenMP support in clang. Has this changed in the recent version of canu?
You can try to get rid of llvm for OSX, GCC should have OpenMP support. @johanneskoester would this be ok with you? Any other ideas?

johanneskoester:
Of course. In fact, in the recent compiler meeting with conda-forge and Continuum, we discussed to entirely move to the packaged gcc, also for osx. This is a valid option also for the other teams.

bgruening:
@jashapiro thanks a lot.
ping @jxtx

PR #1737, opened 2016-06-13
add genomicassertions

dakl:

I wrote a package to make assertions in VCF and BAM files which is useful for integration tests of workflows. I'm sure it could be useful to a larger audience. 

https://github.com/dakl/genomicassertions

@bioconda/core, opinions?

### DIFF
--- a/recipes/genomicassertions/bld.bat
+++ b/recipes/genomicassertions/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
+
+:: Add more build steps here, if they are necessary.
+
+:: See
+:: http://docs.continuum.io/conda/build.html
+:: for a list of environment variables that are set during the build process.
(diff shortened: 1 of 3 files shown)


dakl:
Bump

croth1:
@dakl, you are mixing tabs and spaces for indentation in the setup.py file. Python forces you to be consistent - either choose tabs or spaces.  Usually python code is indented by 4 spaces per level.

You can also check travis' report:

```
+ /anaconda/envs/_build/bin/python setup.py install
   File "setup.py", line 16
     lines = (line.strip() for line in handle
                                            ^
 TabError: inconsistent use of tabs and spaces in indentation
```

dakl:
Argh - I'm having these issues where the whitespace is perfectly correct, and then Sourcetree mangles it when I commit. Usually it doesn't actually cause a problem which is a bit strange, but I'm guessing that different versions of python have a different strategy on enforcing consistent whitespaceing since the installation worked on some versions.

Anyway, I've updated now and I guess we'll see if travis is happier this time. That said, some builds errored, ie crashed before `script` the previous time, unrelated the the actual module being installed.

bgruening:
LGTM! Thanks @dakl and @croth1!

PR #1760, opened 2016-06-15
Add LEfSe version 1.0.7

blankenberg:

We prefix with `lefse-` during symlinking some scripts names that IMHO are too generic: 
```
format_input.py 
plot_cladogram.py
plot_features.py
plot_res.py
```

@bioconda/core new software review needed

### DIFF
--- a/recipes/lefse/build.sh
+++ b/recipes/lefse/build.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+
+outdir=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+mkdir -p $outdir
+mkdir -p $PREFIX/bin
+
+cp -R * $outdir
+chmod +x $outdir/run_lefse.py
+chmod +x $outdir/format_input.py
+chmod +x $outdir/lefse2circlader.py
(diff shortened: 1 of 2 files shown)


bgruening:
Upps: `TabError: inconsistent use of tabs and spaces in indentation` :(

blankenberg:
@bgruening whats the best way to deal with this, as the tabs exists in the actual software (https://bitbucket.org/nsegata/lefse/src/e3cabe93a0d1772ef01aa0f5d27bc50e8a074b72/run_lefse.py?at=default&fileviewer=file-view-default lines lines 24-28).

I could add a patch that fixes formatting in the external package, but seems sort of out of scope for a packaging task.

johanneskoester:
Yeah, I think Python 3 is more strict. If you want, you should contact the authors about this. They might be interested to adapt to Python 3 if that allows them to reach more users in Bioconda. For now, skipping Python 3 is ok. Regarding the changed script names, can you add a summary to `extra -> notes`?

blankenberg:
Very good point about adding in the script renames to the notes. Thanks!

bgruening:
Thanks @blankenberg!

PR #1777, opened 2016-06-17
Try building jellyfish on osx

kyleabeauchamp:


tomkinsc:
Looks like the yaggo dependencies are not satisfied since there isn't a conda package available for ruby (because it is built in to macOS?). So yaggo could add a `# [not osx]` selector to its ruby dependency.

kyleabeauchamp:
Yeah, I gave up because I remember having issues with the yaggo / ruby builds a few months ago.  Also this one was low priority for me so I backburnered it.

PR #1820, opened 2016-06-23
omit upper bound from jdk version for older trinity

tomkinsc:


tomkinsc:
I could use help figuring out how to satisfy the ncurses dependency.

johanneskoester:
Sure. You need to encode the ncurses version into the build string. See [this section](https://github.com/bioconda/bioconda-recipes/blob/master/GUIDELINES.md#cc) of the guidelines for examples (e.g. salmon).

PR #1837, opened 2016-06-25
raxml: compile single cpu bin too

abretaud:


recipes/raxml/meta.yaml
```diff
@@ -12,6 +12,7 @@ requirements:
   run:
 test:
   commands:
+    - raxmlHPC -h > /dev/null
```
bgruening: You need to update the build-number.

abretaud:
Oh yes, thanks!

PR #1841, opened 2016-06-27
Updated vsearch to version 2.0.0

mdehollander:

now supports reading from stdin


colinbrislawn:
Hey, thanks for fixing the vsearch build after they updated it!

PR #1839, opened 2016-06-26
Bambamc

apetkau:
This adds bambamc https://github.com/gt1/bambamc which is used to add sam/bam support for smalt.

* [x] I have read the guidelines above.
* [x] This PR adds a new recipe.
* [ ] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

PR #1853, opened 2016-06-28
Add recipe for bambamc 0.0.50.

acaprez:

### DIFF
--- a/recipes/bambamc/build.sh
+++ b/recipes/bambamc/build.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+export C_INCLUDE_PATH=${PREFIX}/include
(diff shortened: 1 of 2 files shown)


acaprez:
Oops, looks like this is a duplicate of #1839.  Package name is `bambamc` here vs. `libbambamc` there.  @bioconda/core is there a convention to prefix `lib` for library-only packages?  I usually try to stick with using the repo/package name directly whenever possible.

johanneskoester:
Indeed, there is no such convention. We try to use the name that the author has chosen.

johanneskoester:
@bioconda/core  and preference about how to proceed? Deprecate the old package and use this one, or stick with the old name?

bgruening:
+1 to deprecate the old one.

PR #1856, opened 2016-06-29
Feature/metavelet

jerowe:

MetaVelvet : An extension of Velvet assembler to de novo metagenome assembly from short sequence reads

This PR adds a recipe for the metavelvet assembler.

It also includes a convention I would like to use for all perl packages.

Some softwares with perl do hacky things to @INC at run time. Presumably they do this to allow for non root access, or to keep their software isolated, but we already accomplish these two with anaconda. With that in mind perl scripts/modules should be installed using the usual Module::Build (or Extutils::Makemaker, whichever) build system. This ensures that

1. Libraries (*.pm) go to their proper place.
2. Any perl scripts (*pl) have the correct shebang line. Previously I was using a catchall #!/usr/bin/env perl, but I realized this is not the optimal way of doing things and Module::Build does it better. Instead of #!/usr/bin/env perl, it gives the actual shebang perl as its installed. This makes sure that if one has multiple conda/perl libraries loaded, that each perl script is pointing at the perl it was installed with.


jerowe:
Merge?

leobiscassi:
Hi @jerowe, your tests pass and I think the scripts are good! I'll merge it.

jerowe:
Thanks, @leobiscassi!

PR #1865, opened 2016-06-30
Update meta.yaml

davidliwei:
* [x] This PR dates an existing recipe.

Update mageck to version 0.5.4.

### DIFF
--- a/recipes/mageck/meta.yaml
+++ b/recipes/mageck/meta.yaml
@@ -1,10 +1,10 @@
 package:
   name: mageck
-  version: "0.5.3"
+  version: "0.5.4"
 
 source:
   git_url: https://bitbucket.org/liulab/mageck.git
-  git_rev: v0.5.3
+  git_rev: v0.5.4
 
 build:
   number: 0


bgruening:
Thanks!

johanneskoester:
This is weird. I can't see mageck-0.5.4 in the channel: https://anaconda.org/bioconda/mageck/files.
Any idea what's going on here?

johanneskoester:
Ah, I guess travis is slow again... Seems that the master build is not yet finished.

bgruening:
@johanneskoester I think the test-repository that runs every 24h slows testing here down as well. I saw some cases where no test is started until the large tests are finished.

johanneskoester:
Yes, it could be that quotas are per account (makes also a lot more sense than per repository. I will try to find a way to increase the testing intervals.

johanneskoester:
Perfect: travis now has a cron job feature (see [here](https://docs.travis-ci.com/user/cron-jobs/)). I will play around with it later today.

bgruening:
Great! We could create a separate account specifically for testing that is independent of the organisation, isn't it?

johanneskoester:
Sure, but that would be cheating somehow. Let's try weekly intervals first.

bgruening:
Let me know if this works, otherwise I will ask them again, they are really open for such kind of issues for open community projects like bioconda.

johanneskoester:
We should have a weekly cron job now.

bgruening:
Great thanks!

PR #1866, opened 2016-06-30
Feature/metavelvet sl

jerowe:

Adding a recipe for metavelvet-sl


jerowe:
Why were the osx builds stopped?

Why are there osx builds when I specify skip: True #osx anyways? I've been wondering about that!

PR #1881, opened 2016-07-01
Feature/trinotate

jerowe:

Recipe for trinotate. 

https://trinotate.github.io/


recipes/trinotate/Trinotate.clean
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env perl
```
bgruening: Is this script used? Same for the other cleanup script.

jerowe: Good point. I had them around for when I was building patches, but they should be removed.

recipes/trinotate/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  number: 1
+  skip: True # [osx]
+
+package:
+  name: trinotate
+  version: '3.2.2'
+
+requirements:
+  build:
+      - libgcc
```
bgruening: libgcc does not need to be an build requirement. It can be a run-requirement. Also please double check if you really need the gcc as dependency at all. Isn't this just a perl script?
Can we try to remove the skip-osx step?

Thanks a lot for working on this. It's also on my todo list for quite some time :)

jerowe: No problem. I sympathize with having too many items on the to do list. ;-)

I'll try and go back through some of my packages and build them with osx. I may need to pick your brain for that on some of the more complicated compilations, but at least the perl ones should be fairly straightforward.

jerowe:
Merge?

recipes/trinotate/meta.yaml
```diff
@@ -0,0 +1,42 @@
+about:
+  home: https://trinotate.github.io/
+  license: ''
+  summary: 'Trinotate is a comprehensive annotation suite designed for automatic
+    functional annotation of transcriptomes, particularly de novo assembled
+    transcriptomes, from model or non-model organisms'
+
+build:
+  number: 1
```
bgruening: I'm sorry to be so nitpicking ... but the build number starts usually with 0

recipes/trinotate/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+requirements:
+  build:
+      - perl-app-cpanminus
+      - perl-threaded
+      - perl-module-build
+      - perl-dbi
+      - perl-file-find-rule
+  run:
+      - libgcc
```
bgruening: libgcc can be remove.
More for information please look at: https://github.com/bioconda/bioconda-recipes/blob/master/GUIDELINES.md#cc

bgruening:
After this you can merge it :)
Thanks again!

recipes/trinotate/meta.yaml
jerowe: Nitpicky is good. I want things done properly. ;-)

bgruening: Cool! Feel free to merge it as soon as it turns green :)

jerowe:
Merging.

As always, thanks for your help @bgruening!

PR #1887, opened 2016-07-02
Feature/trinity

jerowe:

I'm updating trinity to 2.2.0, and adding support for anaconda zlib. Also made some changes to the scripts so they only look in the $PREFIX/bin for executables instead of using relative directories.


jerowe:
Anyone want to weigh in on this?

recipes/trinity/2.1.1/build.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+BINARY=Trinity
+BINARY_HOME=$PREFIX/bin
+TRINITY_HOME=$PREFIX/opt/trinity-$PKG_VERSION
+
+cd $SRC_DIR
```
bgruening: I don't think this step is needed.

recipes/trinity/2.1.1/build.sh
```diff
@@ -0,0 +1,42 @@
...
+
+make inchworm_target
+make chrysalis_target
+
+cd $SRC_DIR/trinity-plugins/
```
bgruening: same here, SRC_DIR can be removed.

jerowe: The files in the 2.1.1 recipe are not actually the recipe I created, but the original recipe.

recipes/trinity/meta.yaml
```diff
@@ -1,44 +1,52 @@
...
-    home: 'https://github.com/trinityrnaseq/trinityrnaseq'
-    license: "https://raw.githubusercontent.com/trinityrnaseq/trinityrnaseq/master/LICENSE.txt"
-    summary: "Trinity assembles transcript sequences from Illumina RNA-Seq data"
+  home: https://github.com/trinityrnaseq/trinityrnaseq/
+  license: ''
+  summary: 'Trinity RNA-Seq de novo transcriptome assembly'
+
 build:
-  number: 6
+  number: 1
```
bgruening: 0

recipes/trinity/meta.yaml
```diff
@@ -1,44 +1,52 @@
...
-source:
-  fn: v2.1.1.tar.gz
-  md5: 45c1ad13c331d6a1976f6a9b5b015daa
-  url: https://github.com/trinityrnaseq/trinityrnaseq/archive/v2.1.1.tar.gz
+  version: '2.2.0'
+
 requirements:
   build:
-    - perl-threaded
+    - libgcc
```
bgruening: gcc really needed?
OS-X builds?

jerowe: Yes, there is a make command.

Do you think this will build on OS-X? I'll remove the skip and we'll see!

jerowe: Its missing dependencies for the OS-X build.

```
Error:  Packages missing in current osx-64 channels: 
  - jellyfish 2.2.3
  - collectl
  - fastool
  - parafly
  - slclust
```

jerowe: I'll try and see if I can build the dependencies on OS-X as well. I think they're just java, so it should be doable.

Just because I'm wondering, how could a docker container have OS-X? Isn't OS-X heavily licensed? I didn't even think you could have a VM with it...

bgruening: @jerowe travis has some VM with it, don't ask me how :)
I try to fix jellyfish but this seems to be a rabbit hole: https://github.com/bioconda/bioconda-recipes/pull/1889

jerowe: I will just keep on believing that travis is magical. ;-)

I can take a look at it, but like I said I have no experience with compiling software on OSX, and I don't know how much help I'll be.

Should we push the linux build and have a separate PR for the OSX one?

jerowe: Also, what do you want to do with the 2.1.1 build of trinity? It doesn't build on the bioconda docker image, or at least I couldn't get it to build. How is it in there? 

I could update that recipe. Its probably fine for people who have zlib installed as system dependency, so using zlib as an anaconda dependency wouldn't be too bad.

bgruening: Yes sounds like a plan, not sure how long it need to get jellyfish under osx.
W

bgruening: fastool, jellyfish and parafly are done.

jerowe: Cool! I'm going to merge this for now so I can check it off of my spreadsheet.

PR #1894, opened 2016-07-04
Feature/discovar

jerowe:

### DIFF
--- a/recipes/discovar/build.sh
+++ b/recipes/discovar/build.sh
@@ -1,12 +1,12 @@
 #!/bin/sh
 set -x -e
 
-export CC=${PREFIX}/bin/gcc
-export CXX=${PREFIX}/bin/g++
+#export CC=${PREFIX}/bin/gcc
+#export CXX=${PREFIX}/bin/g++
 
-export INCLUDE_PATH="${PREFIX}/include"
-export LIBRARY_PATH="${PREFIX}/lib"
-export LD_LIBRARY_PATH="${PREFIX}/lib"
+#export INCLUDE_PATH="${PREFIX}/include"
+#export LIBRARY_PATH="${PREFIX}/lib"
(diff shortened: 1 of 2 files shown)


recipes/discovar/meta.yaml
```diff
@@ -0,0 +1,22 @@
+about:
+  home: https://www.broadinstitute.org/software/discovar
+  license: ''
+  summary: 'Suitable for variant calling with reference and de novo assembly of small genomes.'
+
+build:
+  number: 1
```
bgruening: build numbers starts with 0

recipes/discovar/meta.yaml
```diff
@@ -0,0 +1,22 @@
+about:
+  home: https://www.broadinstitute.org/software/discovar
+  license: ''
```
bgruening: license is missing.

recipes/discovar/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -x -e
+
+export CC=${PREFIX}/bin/gcc
```
bgruening: Wondering if this is really needed.

jerowe: I think so. I want to make sure it uses the anaconda gcc

recipes/discovar/meta.yaml
jerowe: fixed and fixed

recipes/discovar/build.sh
bgruening: This should not be necessary, afaik.

jerowe: I'll commit a build without it. I suppose you're right. Is ${PREFIX}/bin already the first entry in PATH?

jerowe:
Merge?

bgruening:
Thanks!

jerowe:
Thanks!

PR #1901, opened 2016-07-05
First commit for piler

oinizan:
This PR add a new recipe so I mention @bioconda/core to get a review.

### DIFF
--- a/recipes/piler/meta.yaml
+++ b/recipes/piler/meta.yaml
@@ -1,0 +1,28 @@
+package:
+  name: piler
+  version: "0.1"
+
+build:
+  number: 0
+  skip: True # [osx]
+
+source:
+  fn:  piler_source.tar.gz
+  url: http://www.drive5.com/piler/piler_source.tar.gz
+  md5: 6f351fb437d0cc6a1ec84f41998621c9
+
+requirements:
+  build:
+  - gcc
+  run:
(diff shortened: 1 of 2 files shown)


recipes/piler/meta.yaml
```diff
@@ -0,0 +1,28 @@
+package:
+  name: piler
+  version: "0.1"
+
+build:
+  number: 0
+  skip: True # [osx]
```
bgruening: Is this really needed. Would be great to compile it an OS-X as well :)

recipes/piler/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+source:
+  fn:  piler_source.tar.gz
+  url: http://www.drive5.com/piler/piler_source.tar.gz
+  md5: 6f351fb437d0cc6a1ec84f41998621c9
+
+requirements:
+  build:
+  - gcc
```
bgruening: - gcc # [linux]
- llvm # [osx]

recipes/piler/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+  run:
+  - libgcc
```
bgruening: - libgcc # [linux]

oinizan:
Ok for osx compatibility

PR #1904, opened 2016-07-05
Mreps recipe

oinizan:

This PR add a new recipe so I mention @bioconda/core to get a review.


recipes/piler/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+  skip: True # [osx]
+
+source:
+  fn:  piler_source.tar.gz
+  url: http://www.drive5.com/piler/piler_source.tar.gz
+  md5: 6f351fb437d0cc6a1ec84f41998621c9
+
+requirements:
+  build:
+  - gcc
```
bgruening: You have both tools in this PR.
I have commended on the other already that it would be wonderful if you could try to also compile it under OSX.

Thanks.

oinizan:
For Piler, compilation on osx failed:
> ld: library not found for -lcrt0.o
> clang: error: linker command failed with exit code 1 (use -v to see invocation)
> make: *** [piler2] Error

It seems that error come from the the [static compilation] (http://stackoverflow.com/questions/3801011/ld-library-not-found-for-lcrt0-o-on-osx-10-6-with-gcc-clang-static-flag).
@bgruening: Should I patch the Makefile to remove the -static flag ?

bgruening:
This would be great. I don't want to force you to spend time on making this work on osx, but it would be beautiful to have it on osx as well.

oinizan:
Ok, I will try to patch it.

oinizan:
I close this PR now. Piler + Mreps in a next PR soon.

PR #1907, opened 2016-07-05
Fiji recipe fixes

gregvonkuster:

1. The fiji.sh script is not needed since the ImageJ2 executable encapsulates Java calls.
2. The bunwarpjj.sh executable is still needed because bUnwarpJ provides an API, allowing tools to execute it using Java calls from the command line.  The command line built for these tools will be updated to call the bunwarpj.sh script instead of invoking Java.


bgruening:
Ping @gregvonkuster

gregvonkuster:
@bgruening Sorry, I looked at this a few days back, and just not quite sure why the test failed for this.  Should I just delete this PR for now?

croth1:
@gregvonkuster, the recipe installs a binary named `ImageJ-macosx` instead of `ImageJ` on macOS, but the test script explicitly tries to call `ImageJ` which is doomed to fail. You could try symlinking `$PREFIX/bin/ImageJ-macosx` to `$PREFIX/bin/ImageJ` for the macOS build.

gregvonkuster:
@croth1 Thanks for pointing this out - I'm not quite sure how I missed this.  I'll close this PR and open a new one with a recipe that works in the way I intended as soon as I get a chance.

PR #1912, opened 2016-07-06
Use bioconda utils

johanneskoester:

This PR switches from our previous scripts to the separate bioconda-utils python package (which is basically just a generalization of our scripts).
In the first pass, the switch is made for OSX. Next step is to do the same for Linux.


bgruening:
Wuhu! Fantastic! General question do we want to advertise this library beyond bioconda? Ranaming?

bgruening:
I have restarted travis, now it passes. Thanks Johannes.

bgruening:
Are we ready to merge it?

johanneskoester:
We need to activate it on linux first. This is a bit more complex, since the dockerfile has to be changed. I would like to change it in a way such that we don't need to rebuild the docker image when updating bioconda-utils. One way would be to have a bash script that first pip installs it from the repo, and then passes through any arguments. Also, the dockerfile should be moved to either a separate directory or the bioconda-utils repository.

daler:
@johanneskoester, which version of `conda-build` should we be using? Do you think pinning conda-build would resolve [this error](https://travis-ci.org/bioconda/bioconda-recipes/jobs/147533037#L490)?

johanneskoester:
We should pin to the current stable version. Any idea why this PR changes so many files compared to master? What happened??

daler:
Yeah it strangely seems to have collected a few weeks of commits. Maybe because they're interleaved with comments, and they remain for context or something? Anyway I just opened #2058 which is this same branch, plus pinned versions in `scripts/requirements.txt`. Simplifies things quite a bit.

PR #1924, opened 2016-07-06
Added in tn93 1.0.2

ericenns:
* [X]  I have read the guidelines above.

new recipe @bioconda/core

### DIFF
--- a/recipes/tn93/meta.yaml
+++ b/recipes/tn93/meta.yaml
@@ -8,11 +8,16 @@ source:
 
 requirements:
   build:
-    - gcc
+    - gcc   # [not osx]
+    - llvm  # [osx]
     - cmake
   run:
-    - libgcc
+    - libgcc # [not osx]
 
 test:
   commands:
     - tn93 -h 2>&1 | grep 'compute Tamura Nei 93 distances between aligned sequences' > /dev/null
+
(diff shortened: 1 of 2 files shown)


recipes/tn93/meta.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: tn93
+  version: 1.0.2
+
+source:
+  git_rev: v1.0.2
+  git_url: https://github.com/veg/tn93.git
+
+requirements:
+  build:
+    - gcc
```
bgruening: @ericenns can you adopt this please to our guidelines: https://github.com/bioconda/bioconda-recipes/blob/master/GUIDELINES.md#cc
Thanks!

recipes/tn93/meta.yaml
```diff
@@ -0,0 +1,18 @@
...
+    - cmake
+  run:
+    - libgcc
+
```
bgruening: Can you also please add a `about` section: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bamtools/meta.yaml#L31

ericenns:
Ok I added in an about section and the stuff for not osx and osx. The software does not contain any license so I left it out.

PR #1927, opened 2016-07-07
Piler + Mreps recipes 

oinizan:

This PR add 2 new recipes so I mention @bioconda/core to get a review.


recipes/piler/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  build:
+  - gcc # [linux]
+  - llvm # [osx]
+
+  run:
+  - libgcc # [linux]
+
+test:
+  commands:
+    - piler2 -version # [linux]
```
bgruening: Uhh, why is that?
I don't think it is good to have different behaviour for different OS. Can we symlink the one to the other and make it consistent?

recipes/mreps/meta.yaml
```diff
@@ -0,0 +1,29 @@
+package:
+  name: mreps
+  version: "0.1"
+
+build:
+  number: 0
+
+source:
+  git_url: https://github.com/gregorykucherov/mreps.git
```
bgruening: Can you add a revision, or better use a tar.gz as download with checksum?

oinizan:
@bgruening For Piler: Instead of a symlink I choose to call the executable piler (and not piler 2, see the sed in the Makefile). It is more consistent with the "real" version of piler (1.0). What do you think ?

recipes/mreps/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  url: http://mreps.univ-mlv.fr/mreps-2.6.tar
+  md5: 4cc5444676501adeddf09f0aaf68dde8
+
+requirements:
+  build:
+  - gcc # [linux]
+  - llvm # [osx]
+
+  run:
+  - libgcc
```
bgruening: only linux

recipes/mreps/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  files:
+    - DmelChr4.fa
+  commands:
+    - mreps -res 3 -exp 3.0 -from 10000 -to 12000 DmelChr4.fa
+
+about:
+  home: http://mreps.univ-mlv.fr/ 
+  license: GNU GENERAL PUBLIC LICENSE
+  license_file: LICENSE.md
+  summary: GNU GENERAL PUBLIC LICENSE Version 2, June 1991
```
bgruening: This is the tool summary not the license :)

bgruening:
@oinizan this is a great fix! Thanks! This fix should also go upstream I think.

oinizan:
@bgruening let me know ...

bgruening:
Great! Thanks!

PR #1928, opened 2016-07-07
flow-related recipe

cristhomas:
* [x] This PR adds several new recipe.
@bioconda/core


recipes/r-mba/meta.yaml
```diff
@@ -0,0 +1,86 @@
...
+package:
+  name: r-mba
+  # Note that conda versions cannot contain -, so any -'s in the version have
+  # been replaced with _'s.
+  version: "0.0_8"
+
+source:
+  fn: MBA_0.0-8.tar.gz
+  url:
+    - http://cran.r-project.org/src/contrib/MBA_0.0-8.tar.gz
```
bgruening: can you please add checksums here

recipes/r-rfoc/meta.yaml
```diff
@@ -0,0 +1,93 @@
...
+package:
+  name: r-rfoc
+  # Note that conda versions cannot contain -, so any -'s in the version have
+  # been replaced with _'s.
+  version: "3.3_3"
+
+source:
+  fn: RFOC_3.3-3.tar.gz
+  url:
+    - http://cran.r-project.org/src/contrib/RFOC_3.3-3.tar.gz
```
bgruening: Checksums?

recipes/r-robustbase/meta.yaml
```diff
@@ -0,0 +1,94 @@
...
+package:
+  name: r-robustbase
+  # Note that conda versions cannot contain -, so any -'s in the version have
+  # been replaced with _'s.
+  version: "0.92_6"
+
+source:
+  fn: robustbase_0.92-6.tar.gz
+  url:
+    - http://cran.r-project.org/src/contrib/robustbase_0.92-6.tar.gz
```
bgruening: I'm wondering why these checksums are not automatically included ...

bgruening:
What a awesome contribution! Thanks a lot!
@cristhomas most of the other packages uses `r` and not `r-base` maybe you can change this? And feel free to remove some of the comments in your file.

bgruening:
`ERROR: dependency ‘BH’ is not available for package ‘flowCore’` do you know what BH is?

cristhomas:
That's odd about the checksum, I didn't think to check because they were included on the recipe I generated yesterday. I'll do that now. Also the 'BH' thing I don't know. It worked fine on my machine, I'll look into it.

recipes/r-robustbase/meta.yaml
cristhomas: so looks like other recipe for r packages in the existing bioconda recipe don't have md5 either. I wonder if it just doesn't include them automatically when you build your recipe with 'conda skeleton'?

bgruening: I think so and we should fix this. Sorry to bother you, it's fine. Leave it like it is for now.

cristhomas: that's ok, i fixed the files already =)

cristhomas:
so how did you get the 'BH' dependency error?

bgruening:
Oh is it this one? https://cran.r-project.org/web/packages/BH/index.html

cristhomas:
it probably is yes, if it's not installed for you - it's installed on my system.

bgruening:
I'm wondering why it is working in your Docker container. It shouldn't as this package is not installed in Docker and travis is using the same container.

cristhomas:
aaaah, I didn't test in a docker container. maybe that's why. I'm sorry, I didn't realize it would make a difference.

bgruening:
Ah! No big deal. I think if you add this one package and then the build dependency this should be fine :)

cristhomas:
ok let me do that now

cristhomas:
Let me know if this works now

bgruening:
travis will tell you, I need to leave soon, sorry.

cristhomas:
np
thanks for your help

bgruening:
It failed but because of `Error: in section 'source': unknown key 'rpaths'`
I guess your last recipe has something wired in it :)

cristhomas:
argh! neverending =)
I'll fix and check.

cristhomas:
this worked for me in the docker container.

recipes/r-geomap/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+package:
+  name: r-geomap
+  version: "2.3_8"
+source:
+  fn: GEOmap_2.3-8.tar.gz
+  url:
+    - http://cran.r-project.org/src/contrib/GEOmap_2.3-8.tar.gz
+    - http://cran.r-project.org/src/contrib/Archive/GEOmap/GEOmap_2.3-8.tar.gz
+  md5: 3207abd337906a9345fc6d2be6c2c212
+  rpaths:
```
bgruening: This one here is wrong.

`rpath` should not be in source but in `build:`

cristhomas: thanks for checking. I corrected it. hopefully this was the last mistake.

cristhomas:
hmmm. I didn't do Rgraphviz because it's already in the bioconda recipe. Not sure how to proceed.

bgruening:
Please merge my PR :)
The problem is that some packages are not available on OSX.
This is a general problem and more complicated, so I have skip this for the moment.

cristhomas:
ok!
Didn't I merge your PR already?

bgruening:
There should be a second one, isn't it?

bgruening:
Wuhu! Great! These will be available in 20-30 minutes on anaconda.org

cristhomas:
Awesome, thanks for your help!!

PR #1932, opened 2016-07-08
Kmergenie

jerowe:

This is a recipe for kmergenie.

But I have no idea how to structure it. Help?


jerowe:
OSX build is failing. It seems to be timing out.

bgruening:
The real problem is this one on OSX:
```
+ sed -i 's/print sys.hexversion>=0x02050000/print(sys.hexversion>=0x02050000)/' makefile
sed: 1: "makefile": invalid command code m
```

OSX has a different version of sed :crying_cat_face: 
And the command needs to look like that to be compatible with both sed versions.

```
sed -i.bak 's/print sys.hexversion>=0x02050000/print(sys.hexversion>=0x02050000)/' makefile
```

jerowe:
Why? Why does osx have to mess with sed?  Our entire infrastructure is based on find, grep and sed.

recipes/kmergenie/build.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+set -x -e
+
+export GCC=${PREFIX}/bin/gcc
```
bgruening: This needs to go :)

bgruening:
Now we have this one
```
make: /anaconda/envs/_build/bin/g++: No such file or directory
```
I think if you remove the exports we should be fine.

To answer you question about `sed`. Historical reasons :)
OSX is using the BSD tools and the BSD-sed whereas most of the Linux distributions are using GNU coreutils. Both are not in any case compatible. For this reason we decided in Galaxy to annotate all tools with gnu-coreutils that are using sed, awk etc... so that we can guarantee reproducible results even on OSX and BSD derivatives.

jerowe:
```
~~Historical reasons ~~
```
```
At some point some guy through a hissy fit over something completely ridiculous
```

Fixed that for you.

bgruening:
Let's see what supervisor Travis is saying.
Thanks for being so patient and helping us to improve our recipes!

jerowe:
Thanks for being so patient and helping me fix my recipes. ;-)

bgruening:
A never ending story, but we are one step closer :)

http://stackoverflow.com/questions/19242275/re-error-illegal-byte-sequence-on-mac-os-x

```
LC_ALL=C sed -i.bak 's' doobar
```

recipes/kmergenie/build.sh
```diff
@@ -14,8 +14,9 @@ export CPPFLAGS="-I${PREFIX}/include"
 
 mkdir -p $PREFIX/bin
 
+LC_ALL=C sed -i.bak 's' doobar
```
bgruening: Oh no this was just a terrible example ;)
You should probably do: `export LC_ALL=C`
And this is just a guess. Sorry for the confusing.

bgruening:
Wuhu!

PR #1936, opened 2016-07-08
genblasta and genblastg

jerowe:

### DIFF
--- a/recipes/genblasta/meta.yaml
+++ b/recipes/genblasta/meta.yaml
@@ -14,7 +14,6 @@ package:
 
 requirements:
   run:
-      - libgcc
       - blast
 
 source:
--- a/recipes/genblasta/meta.yaml
+++ b/recipes/genblasta/meta.yaml
@@ -1,0 +1,22 @@
+about:
+  home: http://genome.sfu.ca/genblast/download.html
+  license: 'GNU'
+  summary: 'genBlast is a program suite, consisting of two programs: genBlastA and genBlastG.
+      genBlastA parses local alignments, or high-scoring segment pairs (HSPs)
+      produced by local sequence alignment programs such as BLAST and WU-BLAST and identify groups of HSPs.'
+
+build:
+  number: 0
+  skip: True # [osx]
+
+package:
+  name: genblasta
+  version: '1.0.4'
+
+requirements:
+  run:
+      - libgcc
+
+source:
+  fn: genblast_v1.0.4
+  url: http://genome.sfu.ca/genblast/latest/genblasta_v1.0.4_linux_x86_64
--- a/recipes/genblasta/meta.yaml
+++ b/recipes/genblasta/meta.yaml
@@ -7,7 +7,6 @@ about:
 
 build:
   number: 0
-  skip: True # [osx]
 
 package:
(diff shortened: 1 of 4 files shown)


recipes/genblasta/meta.yaml
```diff
@@ -0,0 +1,23 @@
...
+build:
+  number: 0
+  skip: True # [osx]
+
+package:
+  name: genblasta
+  version: '1.0.4'
+
+requirements:
+  run:
```
bgruening: build requirements missing?
What about osx?

recipes/genblastg/meta.yaml
```diff
@@ -0,0 +1,23 @@
...
+build:
+  number: 0
+  skip: True # [osx]
+
+package:
+  name: genblastb
+  version: '1.38'
+
+requirements:
+  run:
```
bgruening: build requirements? 
And OSX :)

jerowe: They are precompiled binaries.

One, I think genblasta, has a precompiled for osx, but it has a lot of disclaimers. I'll give a try to the source versions and we'll see how that goes. ;-)

jerowe: Will this work?

```
url: ftp://path/to/binary # [linux]
url: ftp://path/to/binary # [osx]
```

?

recipes/genblasta/meta.yaml
jerowe: Huh. I kind of feel like it might be too good to be true, but both builds with precompiled linux/osx test fine. What do you think?

This is the downloads page.

http://genome.sfu.ca/genblast/download.html

bgruening: @jerowe the problem with precompiled binaries is that you often don't know how they where produced. bioconda and conda-forge are trying hard to have one unified build infrastructure based on a as old as possible cent-os. As a result these binaries are relocatable to any other OS, which is not true if you compile them on a recent Ubuntu for example.
So we very much favour compiling from source if possible.

jerowe: @bgruening, fair enough. I'll work on the compilation from source, but I probably won't get to these until tomorrow.

jerowe: I can find source code for genblastg, but not genblasta. I emailed the author, and we'll see if maybe I can get my hands on that. 

Would you like for me to just leave genblasta out for now?

bgruening: There is no strict rule for not letting precombiled binaries in so I would say leave it, but fix it as soon as you can :)

recipes/genblasta/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+build:
+  number: 0
+
+package:
+  name: genblasta
+  version: '1.0.4'
+
+requirements:
+  run:
+      - libgcc
```
bgruening: if you are copying you don't need libgcc

jerowe: I got rid of it.

Merge after this build?

bgruening:
Merge on your convenience :)

PR #1940, opened 2016-07-08
Added in python-bioext 0.17.4

ericenns:

### DIFF
--- a/recipes/python-bioext/build.sh
+++ b/recipes/python-bioext/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+$PYTHON setup.py install
+
+# Add more build steps here, if they are necessary.
+
+# See
+# http://docs.continuum.io/conda/build.html
+# for a list of environment variables that are set during the build process.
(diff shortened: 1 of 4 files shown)


ericenns:
I am at a loss for why this is failing any idea @bioconda/core

ericenns:
Looks like it was due to python-consensuscore not having python version in the build string

ericenns:
Not sure why osx build is failing, I don't have a Mac to test on. I might just set it to not build for osx on Monday if that is ok.

bgruening:
Me neither, but can you try to add `export DYLD_FALLBACK_LIBRARY_PATH="${PREFIX}/lib"` maybe this will help.

ericenns:
Ok trying the flag, I guess we will see if that works.

ericenns:
Thanks for the advice @bgruening it ended up just needing a patch to look for osx specific library file so .dylib instead of .so.6

bgruening:
Awesome! Thanks!

PR #1955, opened 2016-07-11
flock (clustering for flow data)

cristhomas:

@bioconda/core

### DIFF
--- a/recipes/flock/bld.bat
+++ b/recipes/flock/bld.bat
@@ -1,3 +1,0 @@
-"%R%" CMD INSTALL --build .
-if errorlevel 1 exit 1
-
--- a/recipes/flock/bld.bat
+++ b/recipes/flock/bld.bat
@@ -1,0 +1,3 @@
+"%R%" CMD INSTALL --build .
+if errorlevel 1 exit 1
+
--- a/recipes/flock/build.sh
+++ b/recipes/flock/build.sh
@@ -1,9 +1,5 @@
 #!/bin/bash
-
-
-cd flock
-mkdir bin
-cd bin
-cc -o flock1 ../src/flock1.c ../src/find_connected.c -lm
-cc -o flock2 ../src/flock2.c -lm
-cc -o cent_adjust ../src/cent_adjust.c -lm
+cd flock/src
+cc -o $PREFIX/bin/flock1 flock1.c find_connected.c -lm
+cc -o $PREFIX/bin/flock2 flock2.c -lm
+cc -o $PREFIX/bin/cent_adjust cent_adjust.c -lm
--- a/recipes/flock/build.sh
+++ b/recipes/flock/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+
+cd flock
+mkdir bin
+cd bin
+cc -o flock1 ../src/flock1.c ../src/find_connected.c -lm
+cc -o flock2 ../src/flock2.c -lm
+cc -o cent_adjust ../src/cent_adjust.c -lm
--- a/recipes/flock/meta.yaml
+++ b/recipes/flock/meta.yaml
@@ -1,0 +1,10 @@
+package:
+  name: flock
+  version: "1.0"
+
+source:
+  
+  git_url: https://github.com/cristhomas/immport-test.git
+
+about:
+  home: https://sourceforge.net/projects/immportflock/
--- a/recipes/flock/meta.yaml
+++ b/recipes/flock/meta.yaml
@@ -2,7 +2,9 @@ package:
   name: flock
   version: "1.0"
 source:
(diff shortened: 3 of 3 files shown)


recipes/flock/meta.yaml
```diff
@@ -0,0 +1,13 @@
+package:
+  name: flock
+  version: "1.0"
+source:
+  git_url: https://github.com/cristhomas/immport-test.git
```
bgruening: Please specify the git_rev or use a tarball.
Btw. why is this different from the sourceforge page?

recipes/flock/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+
+cd flock
+mkdir bin
+cd bin
+cc -o flock1 ../src/flock1.c ../src/find_connected.c -lm
```
bgruening: You need to move the binaries to $PREFIX/bin at some point.

recipes/flock/bld.bat
```diff
@@ -0,0 +1,3 @@
+"%R%" CMD INSTALL --build .
```
bgruening: can be removed I guess.

recipes/flock/meta.yaml
cristhomas: it's not, I was trying to make it work, and for some reason couldn't get it to work with the sourceforge page.

cristhomas: (the sourceforge doesn't host a tarball, and using a tarball from a git repo was giving me unpacking errors.)

recipes/flock/bld.bat
cristhomas: the line or the file?

bgruening: The file this is a left over isn't it from copying?

cristhomas: I created the file because it says to create one in the conda tuto =)
if it doesn't have to be there i'll remove it.

recipes/flock/meta.yaml
cristhomas: git_tag ok?

bgruening: This one does not work? https://github.com/cristhomas/immport-test/archive/12b153db8d1e0192c449296d9348d2f33d279293.tar.gz

cristhomas: thank you, yes it does. 
how did you get it to work? I couldn't find any help online with the error message I was getting.

bgruening: It's a github trick. Is the commit hash and the archive url. The ending `tar.gz` can be exchanged to `.zip` of example.

cristhomas: so many things to learn!
thank you =)

recipes/flock/meta.yaml
```diff
@@ -0,0 +1,15 @@
...
+source:
+  fn: 12b153db8d1e0192c449296d9348d2f33d279293.tar.gz
+  url: https://github.com/cristhomas/immport-test/archive/12b153db8d1e0192c449296d9348d2f33d279293.tar.gz
+  md5: 0d4821d92847d0f9376249bec7ce8ee8
+requirements:
+  build:
+    - gcc   # [not osx]
+    - llvm  # [osx]
+  run:
+    - libgcc    # [not osx]
```
johanneskoester: There should also be a test section. You can just test whether --help or something equivalent works. We only want to know if the binary loads correctly. Also see the [guidelines](GUIDELINES.md).

cristhomas: Unfortunately the author did not provide a --help or --version or equivalent. Can I add a test file?

What am I missing from the guidelines?

cristhomas: Oh right. There is no license in the program either =/
I did not write it, so I can't fix that part.

bgruening: Yes you can include test files. If the program spits out some useful help message just with `flock` this is also good to go.

cristhomas:
I ended up not adding a test file, but just testing on the output message when run without a file. Let me know if I need to do anything else.

bgruening:
This looks great! You could think about renaming the binary to flock. I don't think anyone expect the name flock1, but maybe it's just me.

cristhomas:
right, so actually there are 3 binaries in there, flock1, flock2 and cent_adjust. it made sense to distribute them together because the clustering tool we have in galaxy offers either flock1 or flock2, and cent_adjust is the program that compares individual events to the stack of events. 
Also it's written in C, I know nothing about C and I didn't want to break it =)

cristhomas:
But I guess I could have three bioconda package, one for each binary? what is best practice?

bgruening:
Ok, thats a strange program, but that's fine! Thanks! Ready to merge! :)

cristhomas:
Great, thank you!

PR #1960, opened 2016-07-11
Update R version to 3.3.0.

johanneskoester:


kyleabeauchamp:
What is iCONDA_R?

scripts/env_matrix.yml
```diff
@@ -4,6 +4,7 @@ CONDA_PY:
   - "3.5"
 CONDA_BOOST: "1.60"
 CONDA_PERL: "5.22.0"
+iCONDA_R: "3.3.0"
```
bgruening: `i`?
I think the next version that will hit the `r` channel will be at least 3.3.1

johanneskoester:
A typo :-). You guys are fast ;-)!

scripts/env_matrix.yml
johanneskoester: We already have 3.3.0 in the R channel, right?

johanneskoester: This is basically also a test how easy it is to update R. If it goes flawlessly, I'll try to follow the R channel more closely.

bgruening: No we don't. We are blocking on https://github.com/conda/conda-recipes/issues/683
It will be released this week hopefully.

johanneskoester: You are of course right. I was mislead by the front page of the R package on anaconda.org, which suggests 3.3.0. This seems to be due to the windows version which is already at 3.3.0.

Thanks! If it will be fixed this week, I can just keep this PR open.

bgruening:
@johanneskoester R3.3 is ready! :)
ping @daler!
Do we want to coordinate this effort a little bit? Any ideas?
I guess Ryan has already a PR in the pipeline :)

daler:
That was a month or more ago, that PR will need updating . . . let me work on that this morning and we can see where we stand. I suspect travis-ci will be the bottleneck, so I'm not sure how much splitting the effort will help.

I'll get the `bioconductor-*` and their `r-*` dependencies going and report back.

bgruening:
@daler we have now 75min build time. So this will help hopefully a bit. Currently no PR is running for 6h because bioconda-test is running.

PR #1962, opened 2016-07-12
Add Segway and genomedata

### LINKED ISSUE #1953: Add recipes for Genomedata and Segway
Could you please add recipes for the Genomedata dense genomic data storage system and the Segway semi-automated genome annotation software?

http://genomedata.hoffmanlab.org/
http://segway.hoffmanlab.org/

Genomedata is a prerequisite for Segway.

Thank you!

dpryan79:

@bioconda/core: I should note that this adds a number of dependencies in addition to Segway and Genomedata. I'm not sure if there's a best practice regarding that. If so, let me know. This should close #1953


recipes/hgtools/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+requirements:
+  build:
+    - python
+    - setuptools
+    - setuptools_scm
+
+  run:
+    - python
+    - setuptools
+    - setuptools_scm
```
bgruening: Wondering if this is really needed.

recipes/segway/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - numpy
+    - forked-path
+    - hdf5
+    - hgtools
+    - colorbrewer
+    - drmaa >=0.4a3
+
+  run:
+    - python
+    - setuptools
```
bgruening: See above.

bgruening:
Submitting multiple packages at once is totally ok :)

recipes/genomedata/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+    - forked-path
+    - pytables >=3.0
+    - textinput
+    - gcc
+    - hdf5
+    - hgtools
+
+  run:
+    - python
+    - setuptools
```
dpryan79: Probably not needed, good catch. I'll go through and remove those throughout.

For C-based packages that use gcc, are we normally requiring libgcc at the runtime? I wouldn't normally do that, but I've seen some discussion surrounding it here so I added it in.

bgruening: This heavily depends on your program, I was unable to come up with a rule or recommendation when to include libgcc and when not ... it is not needed everytime, but hard to tell when it is needed.

dpryan79:
@bgruening Setuptools has been removed as a runtime dependency. Anything else that you've noticed?

bgruening:
This looks fine! Thanks a bunch!

PR #1975, opened 2016-07-13
Bioconductor r3.3.1 update [WIP]

daler:

First attempt at updating bioconductor packages for R 3.3.1.

The changes here so far are largely a result of running the following:
```bash
docker run --net="host" --rm --entrypoint /bin/bash -it -v `pwd`:/bioconda-recipes bioconda/bioconda-builder

# the rest of the following is inside the interactive container
git clone https://github.com/bioconda/bioconda-utils
(cd bioconda-utils && python setup.py develop)
conda install --file scripts/bioconductor/requirements.txt

# clear cache in case same version has been added to bioaRchive
rm -rf cached_bioconductor_tarballs

# full update
scripts/bioconductor/update-bioconductor-packages-wrapper.sh
```

There are several recipes that had existing modifications on top of a default bioconductor recipe. For example, some had `gcc` or `perl-threaded` as a dependency. In those cases I only merged the updated `source` block and any otherwise updated dependencies and left those custom modifications alone.

These potentially problematic packages include:

* bioconductor-shortread
* bioconductor-rgraphviz
* bioconductor-mosaics
* bioconductor-impute
* bioconductor-genefilter
* bioconductor-dose
* bioconductor-gosemsim
* bioconductor-diffbind

Other packages that might be problematic but weren't touched here are `bioconductor-mzr` (libnetcdf has a pinned version) and `bioconductor-flowcore` (various customizations in existing recipe).

I don't expect these to build cleanly at the moment, but it serves as a starting point. For example I believe we have to replace all occurrences of `- r` with `- {{CONDA_R}}`, right?

(xref #1960, https://github.com/conda/conda-recipes/issues/683, @bgruening, @johanneskoester)


bgruening:
@daler I guess we need to replace them with `-r {{CONDA_R}}` and also change the build_string isn't it? Maybe this should be a separate PR?
@johanneskoester can we kill the conda-testing repository? It blocks bioconda for an entire day.

bgruening:
@daler I have also some changes for DESeq2 here: https://github.com/bioconda/bioconda-recipes/pull/1972
It needs also some small modifications to support tximport. I will rebase my branch as soon as you merged this one. Ok?

daler:
I don't expect this to be merged quickly . . . you might want to push the deseq2 changes to this PR so everything can be tested together.

Also, the bioconductor updating script I rand does not create subdirectories and move existing recipes to a subdir as you've done in #1972. The bioconda channel has these older versions; is that sufficient or should the script do the recipe archiving to subdirs?

bgruening:
My personal opinion is to keep old versions in github visible in cases we need to fix old versions or if we need to rebuild everything (can happen earlier as we think, there were a crucial conda bug recently). But there is no official decision yet and I know this was a debatable topic.

daler:
In general I agree. But specifically for bioconductor packages we need to ensure bioaRchive tarballs for old recipes to work. For example, trying to rebuild the old bioconductor-s4vectors/0.8.7 fails because of a 404 error trying to download a no-longer-existing tarball from bioconductor.

I think this would take more work in terms of trawling the git log for version changes, triggering bioaRchive tarball creation for those, and then building older recipes in subdirs where the meta.yaml refers to the bioarchive tarball. I would prefer to get everything updated first, and then go back to retrofit the repo for older version.

bgruening:
@daler I have plans to mirror all packages to an FTP server during build-time: https://github.com/mulled/auto-mulled/issues/2
This will solve this issue in a long run. We then just need to include a fall-back URL or make this part of bioconda-lib ... if 404 than try backup repo.
Does this sound good to you?

daler:
mulled/auto-mulled#2 sounds fantastic. Once that's up and running, adding fallback URLs should be easy!

bgruening:
@daler glad you like it! Let's see when I find time for it. But thats the plan. It will also automatically build docker container for every package :)

daler:
@bgruening, I'm considering closing this one and opening a new PR where `-r {{CONDA_R}}` and `build:string: r{{CONDA_R}}_{{PKG_BUILDNUM}}` are updated for all `r-*` and `bioconductor-*` recipes. It's going to have to be done anyway, might as well do it now.

Also since it's such a big change I think I will split that PR into toposorted chunks to avoid swamping even our extended 75 min limits. So: don't spend too much time on this PR!

bgruening:
Ok! Sounds fantastic. Can I propose an other change as well?
Can you move everything into a `r` subfolder? We need to move all packages into it's own namespace.
I need to do this as well for all perl-* packages. So that our main directory is a little bit smaller.

daler:
Sure. I think I will do it in increments though for easier tracking through the git history.

daler:
Closing this in favor of #1976

johanneskoester:
@bgruening I think bioconda-tests is too important to kill it. However, I can move the cron job to a different time, e.g. at the weekend?

bgruening:
@johanneskoester sorry, maybe too harsh. I do think it is important but, if it already failed and I thought it would be ok to kill the other runs. It takes currently more than 9h and in this time no other PR will be tested.

johanneskoester:
We could indeed try to make it fail early on the first error. But that would mean that we could end up waiting another week to see the next error. What about just doing it between saturday and sunday? I think we can live with having no other tests running during that time, can't we? I can also think about some kind of whitelisting: i.e., we only test recipes that have either failed last time or have been updated in the meantime.

bgruening:
@johanneskoester over the weekend seems good for me. But as a long term solution we should only test the packages from let's say the last 2 weeks. My guess is that if bioconda keeps growing like it was in the last 6 month we will also get a lot of PR over the weekend.

johanneskoester:
The problem is that an error can occur e.g. half a year later, e.g. when a certain dependency is updated that erroneously hasn't been pinned. The modus we have now guarantees to be able to see such problems.

PR #1976, opened 2016-07-13
Bioconductor r3.3.1 update conda r

daler:

This PR replaces #1975. Same idea, but using an updated `scripts/bioconductor/bioconductor_skeleton.py` and `scripts/env_matrix.yaml` that add the CONDA_R placeholder to build strings and dependencies.

This first round is the batch of packages that only depend on R to make sure they build correctly on travis-ci. Subsequent rounds will add more (toposorted) packages.


recipes/bioconductor-biobase/meta.yaml
```diff
@@ -1,23 +1,23 @@
...
+  string: r{{CONDA_R}}_{{PKG_BUILDNUM}}
+  number: 2
   rpaths:
     - lib/R/lib/
     - lib/
 requirements:
   build:
     - 'bioconductor-biocgenerics >=0.3.2'
-    - 'r >=2.10'
+    - 'r {{CONDA_R}}'
```
johanneskoester: We don't need the CONDA_R here. I am pretty sure that the R version is pinned automatically, similar to Python and Perl.

PR #1987, opened 2016-07-15
Upgrade numpy env to 1.11

kyleabeauchamp:
- [ ] I have read the guidelines above.
- [x] This PR does something else (explain below).

@bioconda/core  According to continuum, the latest version of anaconda has upgraded numpy: https://www.continuum.io/blog/developer-blog/anaconda-4-1-released

Perhaps we should do this as well?  Not sure of all the implications so I'll wait for someone else to discuss and merge.


johanneskoester:
I agree, we should follow continuum here. The PR just needs to fix bxpython which can't seem to find zlib (I guess it was compiled before we removed zlib from the docker image).

kyleabeauchamp:
OK I tried the obvious things with no luck.  Maybe we can ping the person who made the recipe?

bgruening:
@kyleabeauchamp thanks for working on this! I have created a PR against your PR which should hopefully fix the bx-python issue.

kyleabeauchamp:
Still nope

bgruening:
@kyleabeauchamp actually it worked, but there two more recipes to be patched.
You have a new PR.

bgruening:
`zlib` is now fixed the remaining error is interesting,
@bioconda/core I try to keep old recipes to be able to fix them afterwards. With the environment variables this does cause problems because very old may not build with a recent numy, r, boost version.

Is it possible to only build the latest version of an recipe if we change the build-matrix? Or maybe we do not build anything if we change the env-matrix. Instead we rebuild the entire dependency chain if a new package is submitted?

@kyleabeauchamp to fix this PR I see two options, include a skip=True to the old version of bx-python or remove it from the repository. I would prefer the first one for the time being or the new blacklist options from @daler.

kyleabeauchamp:
Yeah the new error appears to be an API breaking in some easy_install functions?

bgruening:
My simply guess is that the old version of bx-python does not work with a newer numpy version.

johanneskoester:
Not rebuilding when the env matrix changes is not really convenient, because it will result in inconsistencies (e.g., some packages updated to the new version of the dependency, some not). Adding a `skip: True` would be one way. Removing the recipe would be another (it will still be present in the history). In a sense, I would prefer the latter way, because then we would have the following policy:

For all recipes appearing in the repo we support building them with the current env matrix. This way, we also finally have a policy when to move a recipe to a subfolder with the version name. We do this whenever we want or have to continue supporting that version. This also means that such recipes effectively block the update of the env matrix.

Finally, we can further introduce a change to bioconda-utils that allows to overwrite the env matrix by placing it locally next to the `meta.yaml`. In this case, it would mean the following. If you want to go on to support this version of bx-python, we place a copy of the env matrix to the recipe, and update bioconda-utils. If it is not important for you, I would vote for removing the recipe (aka deprecating it).

bgruening:
> Not rebuilding when the env matrix changes is not really convenient, because it will result in inconsistencies (e.g., some packages updated to the new version of the dependency, some not). 

Why is this a problem? If no package is updated it will not be used. Lets take the R packages as example, why do we need to update all packages? Would it not be more convenient to update all packages that belongs to the (e.g.) DESeq2 dependency chain, if and only if I want to have a new DESeq2 version? Or the other way around, let's assume no one updates an R package in half a year. All packages can stay with R3.2 and everything will work, no need for R3.3, isn't it? Today we just waste space on anaconda.org.

> Adding a `skip: True` would be one way. Removing the recipe would be another (it will still be present in the history). In a sense, I would prefer the latter way, because then we would have the following policy:

I don't consider removing a package a user-friendly way. The argument with `it's in the git history` is very weak imho. It's kind of hard to find the correct package out of the history. 
Assuming you found it, you need to add it back to the master branch - at this time you still have the problem with the env_matrix and we probably need to remove this package after rebuild, which makes it even harder to find the correct one the next time.

I guess part of this is the reason conda-forge has it's different model dealing with github repos.

> For all recipes appearing in the repo we support building them with the current env matrix. This way, we also finally have a policy when to move a recipe to a subfolder with the version name. We do this whenever we want or have to continue supporting that version. 

Who decides when is whenever? I'm not sure there is `someone` that can decide this. We probably need to do this for all or none. 

> This also means that such recipes effectively block the update of the env matrix.

> Finally, we can further introduce a change to bioconda-utils that allows to overwrite the env matrix by placing it locally next to the `meta.yaml`. In this case, it would mean the following. If you want to go on to support this version of bx-python, we place a copy of the env matrix to the recipe, and update bioconda-utils. If it is not important for you, I would vote for removing the recipe (aka deprecating it).

`you` and `we` assume package maintainers like in conda-forge. We do not have that, we have guidelines and I feel we need to be strict here.

What about the following idea:
We only build the latest recipe with the env-matrix of a certain package (this can either be the meta.yml in the root dir of the package, or the the `latest` version of all subdirs). All the other packages are not touched. If someone wants to update a old recipe he/she has to manually pin the versions and change the build-string to reflect the old env_matrix that might not exist anymore.

johanneskoester:
> We only build the latest recipe with the env-matrix of a certain package (this can either be the meta.yml in the root dir of the package, or the the latest version of all subdirs). All the other packages are not touched. If someone wants to update a old recipe he/she has to manually pin the versions and change the build-string to reflect the old env_matrix that might not exist anymore.

To me, that also sounds good. The only problem might be to find the latest version if no primary package exists. What about being strict and requiring a primary recipe in the root dir of the package? I still think there will be use cases where we might need to support overwriting the env matrix for certain packages though, but we can probably wait until/if such a case occurs.

> > Not rebuilding when the env matrix changes is not really convenient, because it will result in inconsistencies (e.g., some packages updated to the new version of the dependency, some not).
> Why is this a problem? If no package is updated it will not be used. Lets take the R packages as example, why do we need to update all packages? Would it not be more convenient to update all packages that belongs to the (e.g.) DESeq2 dependency chain, if and only if I want to have a new DESeq2 version? Or the other way around, let's assume no one updates an R package in half a year. All packages can stay with R3.2 and everything will work, no need for R3.3, isn't it? Today we just waste space on anaconda.org.

It is good to have all packages consistently rebuild against e.g. a common R version, because otherwise you won't be able to mix them in the same environment. Imagine we have DESeq2 rebuild with R 3.3. Then somebody might need to use limma, which is still R 3.2. So far so good, conda would choose R 3.2 for the environment. However if that person needs a third package that has been added more recently, that one would probably have been build only for R 3.3 (or might even require R 3.3 as a minimum version), hence causing a conflict. If we ensure support for the newest version, we avoid this problem.

> I don't consider removing a package a user-friendly way. The argument with it's in the git history is very weak imho. It's kind of hard to find the correct package out of the history.
Assuming you found it, you need to add it back to the master branch - at this time you still have the problem with the env_matrix and we probably need to remove this package after rebuild, which makes it even harder to find the correct one the next time.

Well, there is a way to find old versions in the git history. All we need to do is to add a tag (e.g. packagename-version) whenever a new version occurs. We can even do that automatically in travis.

bgruening:
> To me, that also sounds good. The only problem might be to find the latest version if no primary package exists. What about being strict and requiring a primary recipe in the root dir of the package? I still think there will be use cases where we might need to support overwriting the env matrix for certain packages though, but we can probably wait until/if such a case occurs.

Sounds great, let's do it!

> It is good to have all packages consistently rebuild against e.g. a common R version, because otherwise you won't be able to mix them in the same environment. Imagine we have DESeq2 rebuild with R 3.3. Then somebody might need to use limma, which is still R 3.2. So far so good, conda would choose R 3.2 for the environment. However if that person needs a third package that has been added more recently, that one would probably have been build only for R 3.3 (or might even require R 3.3 as a minimum version), hence causing a conflict. If we ensure support for the newest version, we avoid this problem.

I know, just wondering how frequently this use-case is and if this occurs if the dev can not fix it by triggering the rebuild of needed packages.

>> I don't consider removing a package a user-friendly way. The argument with it's in the git history is very weak imho. It's kind of hard to find the correct package out of the history.
Assuming you found it, you need to add it back to the master branch - at this time you still have the problem with the env_matrix and we probably need to remove this package after rebuild, which makes it even harder to find the correct one the next time.

>Well, there is a way to find old versions in the git history. All we need to do is to add a tag (e.g. packagename-version) whenever a new version occurs. We can even do that automatically in travis.

I don't think that tags are meant to be used like this and we will end up with several 10.000 of tags. There will be probably many problems, starting from github handling such an amount of tags :)

johanneskoester:
> I don't think that tags are meant to be used like this and we will end up with several 10.000 of tags. There will be probably many problems, starting from github handling such an amount of tags :)

From what people say, git itself won't have a problem. Github could have, but this would just be the web interface, similar to displaying thousands of files, or thousands of repositories as it occurs with conda-forge. At least, the tag interface uses paging already. Hence, it might work.

> Sounds great, let's do it!

The question is whether we should wait a bit, because of [this PR](https://github.com/conda/conda-build/pull/966). Our current solution works (although not perfect). I hesitate to invest more work in this if Mike is working on a global solution for upstream conda build.

druvus:
Closing as bioconda is now using numpy 1.12

PR #1999, opened 2016-07-19
rpy2: bump version to 2.8.2

croth1:

### DIFF
--- a/recipes/rpy2/meta.yaml
+++ b/recipes/rpy2/meta.yaml
@@ -1,24 +1,25 @@
+{% set VERSION = "2.8.2" %}
+{% set SHA256 = "2c1a313df4e64236dcfe1078ce847b8e3c180656c894928d3a4b391aacb9b24c" %}
+
 package:
   name: rpy2
-  version: "2.7.8"
+  version: {{ VERSION }}
 
 source:
-  fn: rpy2-2.7.8.tar.gz
-  url: https://pypi.python.org/packages/source/r/rpy2/rpy2-2.7.8.tar.gz
-  md5: 56ca162bca76bb9c3f935f099c916196
-#  patches:
-   # List any patch files here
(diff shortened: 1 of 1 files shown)


croth1:
Hmh, apparently `testRpyMemory` segfaults in our build environment. 

```
def testRpyMemory(self):
        x = rinterface.SexpVector(range(10), rinterface.INTSXP)
        y = rinterface.SexpVector(range(10), rinterface.INTSXP)
        x_rid = x.rid
        self.assertTrue(x_rid in set(z[0] for z in rinterface.protected_rids()))
        del(x)
        gc.collect(); gc.collect()
        self.assertFalse(x_rid in set(z[0] for z in rinterface.protected_rids()))
```

Can anyone test if the failure only pops up in bioconda's build environment on linux? I only have arch linux here, and unfortunately Continuum's r package does not work on arch.

@lgautier any idea what could trigger the segfault here? It's very suspicious that macOS apparently works and only linux segfaults, though.

dmaticzka:
Looks like the old error was replaced by new ones: Now complains about missing R_X11.so and libXt.so.6 libraries during testing. Can't see why, though. At least R_X11.so should be there because of the r dependencies...

lgautier:
@croth1 The segfault could be anything from hardware issues with the RAM to mismatching shared libraries (different at build time and at run time). rpy2 is developed on Linux, and there is a Docker container (see http://rpy2.bitbucket.org/) to try out whether the error is local to your install/conda or more generally observered.

croth1:
@dmaticzka, yes. I didn't follow up here, because I think it would be better to prepare moving general r-stuff to conda-forge. This will take some time, though.

dmaticzka:
@croth1 nice, no hurry. Just wanted to mention this since I stumbled over it.

PR #2003, opened 2016-07-19
Feature/transabyss

jerowe:

Transabyss recipe

http://www.bcgsc.ca/platform/bioinfo/software/trans-abyss


jerowe:
No python-igraph in osx.

recipes/transabyss/build.sh
```diff
@@ -0,0 +1,31 @@
...
+mv transabyss-merge $TOPDIR/scripts/
+
+cd $TOPDIR/scripts/
+2to3 -w *
+cd $TOPDIR/utilities
+2to3 -w *
+
+cd $TOPDIR
+cd utilities
+sed -i.bak 's/from utilities import/from transabyss import/' *.py
```
bgruening: Is this necessary with your patch?

jerowe: If I want to keep it with python3, then yes.

I suppose I could build it with python2. I add a 

```
skip: True #[py35]
```

What do you think? 2to3 or just keep it as python 2?

jerowe: What do you think, @bgruening, keep it as python2 or run 2to3?

yhoogstrate: @jerowe Does:

```
try:
	from utilities import ...
except:
	from transabyss import ...
```
make it universal in this case?

jerowe: I'm not sure what you mean by universal. 

Transabyss is packaged as a set of python scripts, but not as a python package (no setup.py). I want it deployed in conda as a python package, and I don't want for it to include the namespace 'utilities' because that is probably the single most ambiguous namespace of all time.

jerowe:
I'm going to merge this, but if anyone has any opinions please let me know and I will rebuild.

PR #2007, opened 2016-07-19
gsort 0.0.3

brentp:


brentp:
anyone have an idea on this? it's downloading a linux 64 bit binary and seg-faulting on that machine. it's a golang binary...

johanneskoester:
Looks like a bug in the software. Maybe it would be good to report this to upstream?

brentp:
I am the author of gsort, but I'm using go's machinery to build a static binary so I'm not sure how to debug further, thought there might be something with travis...

johanneskoester:
You could run the build process during packaging instead of using pre-built binaries. Maybe the error becomes more obvious then.

bgruening:
Oh yes a GO-compiler would be nice to have ;)

PR #2044, opened 2016-07-25
Feature/detonate

jerowe:

Recipe for Detonate. 

http://deweylab.biostat.wisc.edu/detonate/


jerowe:
@bgruening , could you take a look at the patches for samtools included in the detonate package? I think I got everything (linking against r ncurses instead of curses), but it would be nice if you could look too. ;-)

bgruening:
Looks good to me, but you are missing tests :)

PR #2051, opened 2016-07-26
Add tbl2asn 25.0

nsoranzo:

This continues the work initiated by @tomkinsc in #1661.

### DIFF
--- a/recipes/tbl2asn/meta.yaml
+++ b/recipes/tbl2asn/meta.yaml
@@ -1,0 +1,28 @@
+package:
+    name: tbl2asn
+    version: 24.3
+source:
+    fn: tbl2asn.gz # [osx]
+    url: ftp://ftp.ncbi.nih.gov/toolbox/ncbi_tools/converters/by_program/tbl2asn/mac.tbl2asn.gz # [osx]
+    md5: 2a92d2021a0838659f606822197f60ee # [osx]
+    # Currently only the binaries are posted for Linux, and they use GLIBC_2.7, making them incompatible with the
+    # GLIBC_2.5-based CentOS 5 test environment used by bioconda.
+    # 
+    # fn: tbl2asn.gz # [linux]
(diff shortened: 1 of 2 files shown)


recipes/tbl2asn/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+    sha256: 0f95c6c2872b2beb908eb42a64b15ed282424332faab3c442b43fc8b99548e58 # [linux64]
+
+build:
+    number: 0
+
+# Currently only the binaries are available, and they use GLIBC_2.7 on Linux,
+# making them incompatible with our GLIBC_2.5-based CentOS 5 test environment
+test:
+    commands:
+        - tbl2asn --help | grep arguments # [osx]
```
tomkinsc: I'm not wild about skipping tests under Linux, but if the alternative is no tbl2asn package we can probably live with it. Can we test executability post-install via a `post-link.sh` script?  If the script fails with a non-zero exit code, no package metadata is written, and the package is not considered installed. It may have to check for some environment variable to determine if it is being called during build vs. install (skip the tbl2asn call if build, run if install).

nsoranzo: @tomkinsc I agree with your concern, and I really don't understand why the @ncbi keeps releasing only the binaries and overwriting previous versions. :disappointed: 
Unfortunately I don't know how to implement your idea and I need this package as a dependency for another one ([Prokka](http://www.vicbioinformatics.com/software.prokka.shtml)) which I'm going to submit as soon as this one is merged. Is it ok for you if we merge this as it is and you open a PR when you implement the `post-link.sh`?

bgruening: @nsoranzo I'm also fine with skipping the entire build for osx in this case.

nsoranzo: @bgruening osx works fine, why skip it?

bgruening: Oh true. I'm fine with merging and I will write my ncbi contact to do something about the source.

PR #2052, opened 2016-07-27
add recipe for cosi2

tomkinsc:


recipes/cosi2/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+#strictly use anaconda build environment
+export CC=${PREFIX}/bin/gcc
+export CXX=${PREFIX}/bin/g++
```
bgruening: Why is this needed? It shouldn't be?

PR #2053, opened 2016-07-27
Feature/cap mirseq

jerowe:
* [x] I have read the guidelines above.
* [x] This PR adds a new recipe.
* [ ] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

Recipe for cap mirseq
http://bioinformaticstools.mayo.edu/research/cap-mirseq/

PR #2058, opened 2016-07-27
Add bioconda utils

daler:

Use bioconda-utils to build packages.

#1912 seems to have recorded all changes in the past month. This PR is for the same branch, but hopefully simplified.


daler:
@johanneskoester latest conda-build (1.21.7) has issues with `--dirty` (xref: https://github.com/conda/conda-build/pull/1148); 1.21.6 seems ok.

scripts/travis-setup.sh
```diff
@@ -1,25 +1,37 @@
...
+# Install bioconda-utils with miniconda version of python
+git clone https://github.com/bioconda/bioconda-utils.git
+(cd bioconda-utils && python setup.py install)
+
 
-    # install conda
-    curl -O https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
-    sudo bash Miniconda3-latest-MacOSX-x86_64.sh -b -p /anaconda
-    sudo chown -R $USER /anaconda
+if [[ $TRAVIS_OS_NAME != "linux" ]]
```
johanneskoester: Is this special case for osx still needed?

daler: Currently, yes: [this line](https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/utils.py#L281) sends the channels to the docker container but if not `docker` then [this line](https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/utils.py#L303) does not.

I think a better solution might be to add the channels to the config.yml; then the only special-casing will be whether to use `--docker` or not.

recipes/boltons/meta.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: boltons
-  version: {{ environ['GIT_DESCRIBE_TAG'] }}
+  version: {{ GIT_DESCRIBE_TAG }}
```
johanneskoester: This might not throw an error, but it should result in an empty version instead. This is because with --dirty, the environment variable can't be known, right? I would suggest to either just copy paste the git_rev value or define it via jinja.

scripts/travis-setup.sh
johanneskoester: Indeed that would be great. I don't have the time in the next days. Could you do that?

recipes/boltons/meta.yaml
daler: Ah, right. I'll just paste the git_rev.

scripts/travis-setup.sh
daler: I don't really have time either! But I will try to take stab at it.

scripts/travis-setup.sh
```diff
@@ -1,37 +1,10 @@
...
-
-    # setup bioconda channel
-    conda config --add channels bioconda
-    conda config --add channels r
-    conda config --add channels file://anaconda/conda-bld
-
-    # setup bioconda-utils
-    pip install git+https://github.com/bioconda/bioconda-utils.git
-fi
+pip install git+https://github.com/bioconda/bioconda-utils.git@docker-use-config
```
johanneskoester: I merged your docker-use-config branch into the master of bioconda-utils. Hence we can remove the branch tag here.

johanneskoester:
The permission error is weird... It works for me locally.

daler:
Yeah I can't figure that out. Works fine locally for me as well.

johanneskoester:
Maybe the problem is that the docker container runs with a different user than the travis process, see [here](https://denibertovic.com/posts/handling-permissions-with-docker-volumes/).

daler:
Yeesh. Looks like that could be the trouble. So we would need to keep maintaining a docker image for this? Or maybe just running everything with sudo on travis-ci?

Also, it's suspicious that this takes 18 mins to fail. I'm starting to sprinkle logger.debug() calls in bioconda-utils, and it looks like filter_recipes() is taking most of the time. Wasn't this an issue multiple times with conda-build? Should we go back to using your fork of it?

johanneskoester:
I really hate these frequent regressions in conda build. I think I tested it with 1.21.5 and at that time, the filtering was still fast. Maybe we should try that first, because my fork is a bit outdated and I don't want to maintain it.
There should be a solution to set the user without modifying the container. Just via the docker API. I would think that setting the `user` argument to the current user id could already work in our case (see [here](http://docker-py.readthedocs.io/en/stable/api/#create_container)).

daler:
OK, I'll try 1.21.5 and setting the user.

johanneskoester:
I just checked 1.21.6, and it needs 1min 23sec for filtering the whole repository on my machine (only a simple laptop with a normal internet connection). Approximately the same time as 1.21.5. So there is no regression. Maybe it was just a temporal slowlyness on travis ci?

daler:
Phew!


Looks like the slowness was due to bioconda channel not being checked when filtering recipes, resulting in pretty much everything looking like it needed to be built. I also had a syntax error when passing the args to conda build inside the docker container, causing that to complain and I think that's what was giving the permission errors.

Seems OK now. @johanneskoester if you're OK with the logging branch changes in bioconda-utils, please merge (or fix) so this can go back to using master.

Next up:

- [ ] building a list of R-dependent recipes to add to a temporary blacklist (`utils.get_dag` to the rescue)
- [ ] add a CONDA_R env var
- [ ] ensure we're skipping those blacklisted recipes that would otherwise trigger a build with R 3.3.1
- [ ] start incrementally removing recipes from the blacklist to trigger builds with R 3.3.1 and start merging PRs with master

daler:
I think we're good to go with this PR, but it would be good to have some more eyes on it before merging.

r-blacklist
```diff
@@ -0,0 +1,331 @@
...
+recipes/bioconductor-iranges/2.4.7
+recipes/bioconductor-isomirs
+recipes/bioconductor-jmosaics
+recipes/bioconductor-kegg.db
+recipes/bioconductor-kegggraph
+recipes/bioconductor-keggrest
+recipes/bioconductor-limma
+recipes/bioconductor-limma/3.26.7
+recipes/bioconductor-limma/3.28.2
+recipes/bioconductor-limma/3.28.6
```
bgruening: Assuming that the latest version in in root. I would not touch any older package for the migration to R 3.3.
In other words I would only migrate packages from the root of every folder and leave older package as they are, so they should not be needed on this blacklist.

johanneskoester:
Looks good to me!

johanneskoester:
Thanks a lot for your help on this Ryan! Great work!

r-blacklist
daler: I agree that older versions shouldn't be upgraded to R 3.3. But these infrastructure changes are generic, not just for R. For now I think the solution would be to keep those old versions on an "old version, do not migrate" blacklist.  The bioconda-utils tools make this easy to do, the only cost is in having another list to maintain -- which shouldn't be too bad since it won't change much.

bgruening: :+1:

bgruening:
Yeah, awesome work @daler!

daler:
@bioconda/all I believe our builds are currently broken -- builds are failing, yet travis-ci is reporting success. Need to figure out what's going on.

daler:
I think it got it working in #2070 and in https://github.com/bioconda/bioconda-utils/tree/permission-test

daler:
@bioconda/all, we're back to the old build system for now (as of #2071). Looking at recent PRs merged with master, updates from @chapmanb and @tomkinsc would have been affected (#2066, #2068). Since in #2071 these are building fine, I believe that in the end the only effect is that your updated recipes would not have been uploaded to anaconda. They should be there now, but please double-check that everything is as you expect.

In addition there are some open PRs that were affected:

- @dmaticzka (#2072)
- @eggzilla (#2067)
- @jerowe (#2053)

For these, please merge master branch back into your respective PRs to try testing again. Sorry for the inconvenience while we work out the new build system . . . we'll clearly need better tests-for-the-tests.

PR #2067, opened 2016-07-30
RNAlien and Stack

eggzilla:
* [x] This PR adds 2 new recipes.


recipes/stack/build.sh~
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
bgruening: This file can be removed.

eggzilla:
Nice, thanks for fixing and adding pandoc

eggzilla:
As expected it fails again after rebase

PR #2070, opened 2016-07-31
Bioconda utils r 3 3 1

daler:

Updates bioconductor packages (after running `scripts/bioconductor/update-bioconductor-packages-wrapper.sh`), and adds them to the r-blacklist to prevent everything from building on R 3.3.1 simultaneously. Then just one (bioconductor-biocgenerics) is uncommented in order to confirm successful building.

Importantly, this fixes some build issues with `bioconda-utils` where incorrect users/permissions were causing failures, but travis was still reporting successful builds.


daler:
There seems to still be a permission issue, but only showing up in somewhat obscure cases (matplotlib in py27: https://travis-ci.org/bioconda/bioconda-recipes/jobs/148601578#L2747)

bgruening:
@daler it is this error I assume: https://github.com/matplotlib/matplotlib/pull/5954
We could blacklist the potential packages for the time being or use the one from conda-forge (1.5.2) with this fix included?

bgruening:
@daler what is not clear to me why python is searching for 1001. The dev user has uid 1000. Maybe we can hack the container in the meantime and create a random user with 1001 within bioconta-utils?

daler:
It's because we're running the container with the `--user` argument to avoid other permission issues. I was trying to avoid a chmod 777 on the conda-bld dir so that bioconda-utils would be useful on local machines as well. I see now that the problem is that the uid passed via the `--user` arg may not exist in the docker container, in this case I'm guessing $USER on travis-ci has uid 1001.

daler:
@bgruening and yep, seems like it's that matplotlib issue, so anything in py27 that depends on matplotlib could get that. I think it would be better to solve it from the build side to avoid having to deal with channel-specific dependencies.

daler:
@bgruening Currently trying https://github.com/bioconda/bioconda-utils/commit/c97eeefbb74a6834be3ec200e5211df6c5befcd8

daler:
Goal:

Given an arbitrary uid from the host, send the uid to the docker container. If it doesn't exist then create it, and have the container run everything under that uid. The reason for this is that we'd like the docker container to save its work in the host's conda-bld dir, which will have permissions for a particular uid (1001 on travis-ci, but could be anything).

Trying to add a uid to the startup script https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/bioconda_startup.sh, doesn't work (see [here](https://travis-ci.org/bioconda/bioconda-recipes/jobs/148679262#L621) because that very command is trying to be run by UID 1001. Omitting the `--user 1001` arg to `docker run`, I'm getting `Operation not permitted` errors when trying to create the user.

@johanneskoester @bgruening I'm stumped at the moment. Any ideas for getting this running while still using the continuumio/conda_builder_linux image?

bgruening:
@daler I don't think it's worth the pain. I would remove the --user I guess from the container. Docker user mapping feature is not completed yet :(

johanneskoester:
Should we revert to the old build engine until this is fixed?

bgruening:
@johanneskoester https://github.com/bioconda/bioconda-recipes/pull/2071

johanneskoester:
Ah, thanks! @daler, let me know if I can help. I don't want to mess with bioconda-utils while you are still trying things.

daler:
@johanneskoester please take a look and see what you can do -- I won't be working on this for a couple of days. Mess with it as much as you need!

daler:
@johanneskoester -- in case it's useful, some notes on where I left off and where effort might be best focused:

To allow any user to build on their own machine, I think maintaining our own docker image following the [post](https://denibertovic.com/posts/handling-permissions-with-docker-volumes/) you pointed out earlier will be the way to go. This seems like the most generally useful approach, with the added bonus that it would be really easy for recipe authors to test and use recipes locally.

An easier but less general solution is to hard-code a UID 1001 into whatever image we use, say derived from continuumio/conda_builder_linux.

A third solution is to bypass the `--docker` option in bioconda-utils, and run a wrapper along the lines of the build-packages.py that installs bioconda-utils and then calls it.

PR #2071, opened 2016-08-01
revert to previous build system

daler:
While we work out the issues with new build system, this reverts back to using the bioconda/bioconda-builder docker container and only uses bioconda-utils for OSX. This way we at least have a working build system . . .

### DIFF
--- a/recipes/bcbio-nextgen/meta.yaml
+++ b/recipes/bcbio-nextgen/meta.yaml
@@ -3,7 +3,7 @@ package:
   version: '1.0.0a0'
 
 build:
-  number: 2
+  number: 3
   skip: True # [not py27]
 
 source:
@@ -36,6 +36,7 @@ requirements:
     - ipython-cluster-helper
     - ipywidgets
     - joblib
+    - libsodium >=0.4,<1.0
     - logbook
     - lxml
     - matplotlib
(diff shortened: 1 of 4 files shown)


daler:
@chapmanb the build system has been in flux last couple of days and the new stuff it not working correctly. So I'm trying to revert in this PR, but I'm not sure how to fix this bcbio error on OSX: https://travis-ci.org/bioconda/bioconda-recipes/jobs/148761718#L1277. Any ideas?

johanneskoester:
Thanks for the reverting @daler. The problem with bcbio-nextgen is that it depends on pyzmq from the default channel. pyzmq depends on zeromq which in turn depends on libsodium. On linux, libsodium has been pinned to a particular version. On macos, there is neither a pin nor an actual dependency. I guess this is a bug in the zeromq package. @chapmanb, do you have a temporal solution, e.g. depending on libsodium in bcbio-nextgen? Or should we build-skip the version of the package on osx?

chapmanb:
Ryan and Johannes -- thanks for all the work on the build system. Sorry about the error, I hadn't seen this before but happy to resolve it. I'm not sure about the best way to test/fix to make this easier for you -- if I update bcbio now so it works on OSX (either by adding libsodium or skipping), would that help this out?

johanneskoester:
Yes, that would be perfect, thanks!

chapmanb:
Ryan and Johannes -- I have a fixed version of bcbio/libsodium in #2073. The unix build is failing, I assume due to current build issues, but the OSX version now passes. Do you all have magical powers to merge that so it'll stop blocking you? Thanks much.

daler:
Thanks very much Brad. The changes are small enough, I'll just check them out into this PR.

PR #2075, opened 2016-08-01
Bioconda utils r 3 3 1 from Ryan with fixed UID

bgruening:

@johanneskoester @daler this is ugly I know but I think it is working. Can you both have a look?


scripts/travis-setup.sh
```diff
@@ -1,28 +1,20 @@
...
-    curl -O https://repo.continuum.io/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
-    sudo bash Miniconda3-latest-MacOSX-x86_64.sh -b -p /anaconda
-    sudo chown -R $USER /anaconda
-    mkdir -p /anaconda/conda-bld/osx-64 # workaround for bug in current conda
-    mkdir -p /anaconda/conda-bld/linux-64 # workaround for bug in current conda
-    export PATH=/anaconda/bin:$PATH
-    conda install -y --file $SCRIPT_DIR/requirements.txt
-    conda index /anaconda/conda-bld/linux-64 /anaconda/conda-bld/osx-64
+    sudo apt-get update -qq
+    sudo apt-get install docker-engine --no-install-recommends -y -o Dpkg::Options::="--force-confmiss" -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confnew"
```
daler: Doesn't travis-ci come with docker?

scripts/travis-setup.sh
```diff
@@ -1,28 +1,20 @@
...
+[[ $TRAVIS_OS_NAME = "linux" ]] && tag=Linux || tag=MacOSX
+curl -O https://repo.continuum.io/miniconda/Miniconda3-latest-${tag}-x86_64.sh
+bash Miniconda3-latest-${tag}-x86_64.sh -b -p /tmp/anaconda
+export PATH=/tmp/anaconda/bin:$PATH
+conda install -y --file scripts/requirements.txt
+pip install git+https://github.com/bioconda/bioconda-utils.git@bgruening
+chmod -R 777 /tmp/anaconda
+ls -lrth /tmp/anaconda
+echo "$UID"
+echo "$USER"
```
daler: prob don't need these last lines, they were leftover from all the debugging

daler:
Looks like the `bgruening` branch in bioconda-utils is forcing everything (even on host?) to UID=1000. This means recipe authors need to be UID=1000 on their machine in order to test locally. Is that true?

If so, it works and we can move forward with all the R stuff, but we should definitely warn about this in the README and eventually have a better solution.

daler:
Also I just manually restarted the travis-ci jobs that failed apparently due to github network issues.

daler:
Nice, it's working! Is there a `set -x` somewhere that's making the conda testing so verbose?

scripts/travis-setup.sh
bgruening: Yes it does, this was part of my debugging. This will install the latest docker version on Linux. If you think we should stick with an older version, I can remove it.

bgruening:
I'm not aware of any special `set -x` maybe thats part of the container?
@daler I know this is a bad hack, just wanted to get it working and get the R stuff done.
I will think about a better solution over the next weeks, now that we know how to get it working we can probably use your `getuid()` magic in a more general way.

daler:
OK, `set -x` as part of the container's conda-build install is what I was guessing. And I'm fine with hacks if it means as I can use new/updated Bioconductor packages sooner!

bgruening:
@daler feel free to merge it ;)
I will try to find some time to fix it in a better way over the weekend.

scripts/travis-setup.sh
daler: Ah. I think its fine, as long as it doesn't take too long to install.

johanneskoester: I would like to keep everything as simple as possible. So if we don't need it, let's remove it.

bgruening:
I cleaned up the PR and tried to get rid of the hard-coded UID, no luck so far. 
See here: https://github.com/bioconda/bioconda-utils/pull/8

I would like to get this merged so we can work on the R packages. In parallel we can try to make this more pleasant for our users and try new strategies.

johanneskoester:
Ok, let's do it like that. Feel free to merge.

bgruening:
@daler @johanneskoester I changed to track the master branch of bioconda-utils. If you thinks it's ready please hit the button ;)

johanneskoester:
Why is it saying "Pulling docker image" after the tests have finished? See [here](https://travis-ci.org/bioconda/bioconda-recipes/jobs/149115407#L330).

bgruening:
Maybe because we are writing the Docker logs after everything else: https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/utils.py#L332

johanneskoester:
But that would only happen in case of an error. If that would be the case, the build would not succeed.

bgruening:
I'm two lines above the if condition, but this does not explain it because it was a debug log and not warning. The messages are coming from https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/utils.py#L427

bgruening:
@johanneskoester https://github.com/bioconda/bioconda-utils/pull/9

johanneskoester:
Ok. I would be fine with merging.

bgruening:
Go for it @johanneskoester!
And thanks for the careful review!

johanneskoester:
Thanks for the continuing awesome work on sorting this out, @bgruening and @daler!

johanneskoester:
Mhm, current master builds are failing again. I have quickly fixed a syntax error in bioconda-utils introduced by the last PR. However, the builds are failing all over the place with different or even no obvious error messages...

daler:
One issue is that recent packages have been bioconductor/R packages and it looks like bioaRchive is having connection issues (https://travis-ci.org/bioconda/bioconda-recipes/jobs/149169354#L644). I'm also getting bioaRchive timeouts when trying locally.

daler:
@bgruening any bioaRchive issues you know about?

bgruening:
@daler yes, we have such bad luck. Unbelievable ....

daler:
OK that gives us time to work out other things . . . like why is it trying to upload a package that already exists, as [here](https://travis-ci.org/bioconda/bioconda-recipes/jobs/149169354#L652)? It should not have tried building that package in the first place if it exists.

bgruening:
I like your positive attitude ;)
I'm in contact with the bioarchive admins and it seems the server will be back in 1,5h.

bgruening:
@daler server is back, I will trigger all travis jobs. Let's see if this was a real error.

johanneskoester:
@bgruening @daler, I guess it would be best to revert this PR now. Do you agree? After reverting, we can update all later PRs with the master branch to have the old build system again.

daler:
Yep sounds good to me.

PR #2078, opened 2016-08-02
rdkit: add pillow as runtime dependency

dmaticzka:


bgruening:
Can you also add a test that triggers the failure without this dependency?

dmaticzka:
Good point, I will solicit one.

PR #2079, opened 2016-08-02
 add r-inline recipe

sebastian-luna-valero:

Add recipe for:
https://cran.r-project.org/web/packages/inline/index.html

@bioconda/core please review

### DIFF
--- a/recipes/r-inline/bld.bat
+++ b/recipes/r-inline/bld.bat
@@ -1,0 +1,8 @@
+"%R%" CMD INSTALL --build .
+if errorlevel 1 exit 1
+
+@rem Add more build steps here, if they are necessary.
+
+@rem See
+@rem http://docs.continuum.io/conda/build.html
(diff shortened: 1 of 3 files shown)


johanneskoester:
Looks good. Btw., you can also create a PR for more than one recipe. As long as it won't build longer than an hour, it should be fine.

sebastian-luna-valero:
Thanks! 

I was doing one by one due to dependencies between them. The docker image was complaining when testing locally. Is this not a problem on Travis?

bgruening:
@sebastian-luna-valero the bioconda build environment should create a DAG of your dependencies and build one after the other in the correct order. Everything else is a bug. We are currently working heavily on the build-environment, so it can be that we introduced a bug. Let us know if this is the case.

sebastian-luna-valero:
Great, many thanks for your hard work!

PR #2083, opened 2016-08-02
Upgrade some r packages

bgruening:


bgruening:
@daler @johanneskoester can you please look at this error: https://travis-ci.org/bioconda/bioconda-recipes/jobs/149434862

I have the bad feeling that the compiler versions between the old container and the new one are not compatible.

johanneskoester:
At first sight, this looks like a problem in the code of r-jpeg. We should use the same GCC as before, since we are using the GCC package, not the compiler in the container, right?

But given the problems with other PRs (e.g. packages not being uploaded), we should consider the following:
* revert to the old build system
* write proper tests for bioconda-utils
* once it is thoroughly tested, try again

What do you think?

bgruening:
> At first sight, this looks like a problem in the code of r-jpeg. We should use the same GCC as before, since we are using the GCC package, not the compiler in the container, right?

There are more errors in this link, like checkmate and I have seen much more that looks very similar.
I share your assumption, that's why I'm puzzled ... Maybe one of these build-environments was/is not using the conda-gcc?

> But given the problems with other PRs (e.g. packages not being uploaded), we should consider the following:
* revert to the old build system
* write proper tests for bioconda-utils
* once it is thoroughly tested, try again

I don't like to revert once again, but it's probably the best idea at this point. @johanneskoester @daler are you available for a google hangout today?

bgruening:
@johanneskoester here are more: https://travis-ci.org/bioconda/bioconda-recipes/jobs/149445370
Maybe this is a very easy fix, but I don't see it.

daler:
How about an intermediate step of swapping the original bioconda-builder image for the continuumio image for the container? It's a one-line edit in config.yml and it should help isolate container-specific problems from general build-system problems.

bgruening:
@daler I have done this. Still there is the new startup script that does a lot of stuff that can be dangerous. Please feel free to use this PR to try things out. I'm not sure I can work on this during the next hours.

daler:
True. We should be able to configure that script in config.yml as well. In fact, it's probably better there since it's container-dependent.  I might be able to work on some of this in the afternoon.

I agree that we should revert again until we have a system as well-tested as we can get it, that can be swapped in with minimal disruption.

I think writing proper tests should be straightforward. Then again, I thought upgrading the build system would be straightforward until I actually tried it!

johanneskoester:
No time for a hangout today, sorry Björn. However, I am about to revert. Then, we can sort out what's going on without pressure.

bgruening:
@daler what do you think, can we install bioconda-utils into the container to get things to work quickly to upgrade the R packages? This will reduce the complexity a lot I think of the build-system for the moment.

johanneskoester:
You can even just modify our docker image to have bioconda-utils in addition to the current build system. This way, we don't break anything and we can easily test it in a PR after the docker image has been rebuild.

daler:
Installing bioconda-utils into the container in addition to the existing build-packages.py sounds good to me. We can add a flag to build-packages.py to send the rest of its args to bioconda-utils build, that way it's just a toggle in travis-run.sh for which build system to use.

PR #2085, opened 2016-08-02
Revert "rdkit: add pillow as runtime dependency"

bgruening:
Reverts bioconda/bioconda-recipes#2078

### DIFF
--- a/recipes/rdkit/2016.03.1/meta.yaml
+++ b/recipes/rdkit/2016.03.1/meta.yaml
@@ -13,7 +13,7 @@ source:
     - rdconfig.patch [win]
 
 build:
-  number: 1
+  number: 0
   skip: True # [win or py3k or osx]
   string: py{{CONDA_PY}}_boost{{CONDA_BOOST}}_{{PKG_BUILDNUM}}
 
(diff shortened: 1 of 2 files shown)


bgruening:
@johanneskoester @daler this worries me a little bit. I think all the other PR are not build and are working. This one isn't with the new Docker container. With the old one it worked as it seems. Any ideas? I tried to add gcc as dependency in https://github.com/bioconda/bioconda-recipes/pull/2084 because I thought it's the build-in gcc version from the Container, but is was also not working.

Ideas welcome. I will revert this PR for the time being.
ping @dmaticzka

johanneskoester:
Sorry, I don't get the problem yet (maybe it is still too early for me). The resulting package is not working?

bgruening:
The PR of this package was working with the old build-system, merging into the new build-system broke it as it seems.

johanneskoester:
Ok. Looks like it can't link to some boost stuff (see [here](https://travis-ci.org/bioconda/bioconda-recipes/jobs/149254778#L4211)). Maybe that error did not occur in our own container because lib was leaked in from CentOS.

johanneskoester:
Ok. Looks like it can't link to some boost stuff (see [here](https://travis-ci.org/bioconda/bioconda-recipes/jobs/149254778#L4211)). Maybe that error did not occur in our own container because lib was leaked in from CentOS.

PR #2087, opened 2016-08-03
Switch back to old build system because of continuing problems.

johanneskoester:
Reverts bioconda/bioconda-recipes#2075. Unfortunately, the original PR had multiple other changes. But we need to make this step **now**, in order to not completely loose track of broken packages.


johanneskoester:
Closing in favor of #2089.

johanneskoester:
Reopening because #2089 fails to build many of the updated R packages. This PR will also undo the R package updates. This is unfortunate, but our only chance to quickly get back to the old system.

PR #2095, opened 2016-08-04
Fix EMBOSS package

bgruening:


bgruening:
Ok, it turns out EMBOSS is hard-compiling the path to the ACD files and testing does not remove the ACD files from the build-step. We probably need to wrap every single tool and export the ACD_PATH before calling the binary.

PR #2096, opened 2016-08-04
Increase build number

marcelm:

Forgotten in last PR


marcelm:
Forget this - there is no need to increase the build version since the only change is that now an OS X package exists.

PR #2100, opened 2016-08-04
Bcl2fastq nextseq recipe

dkoppstein:

Creates a package for [this bcl2fastq2 nextseq wrapper](https://github.com/brwnj/bcl2fastq).

NB: bcl2fastq2 is Illumina software that is freely available but whose license does not allow redistribution. Thus, this package assumes bcl2fastq2 is elsewhere in the PATH. 

Per guidelines, cc'ing @bioconda/core for review.


recipes/bcl2fastq-nextseq/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  git_url: https://github.com/brwnj/bcl2fastq.git
+  git_rev: add0bd6669ced224223529d2ae3a8931431f8857
+
+build:
+  number: 0
+  skip: True # [not py27]
+
+requirements:
+  build:
+    - python >=2.7,<3
```
johanneskoester: Usually, we don't restrict the Python version here, because this is already done by skip above. Also, in a previous conda version this resulted in having builds that are not pinned to the Python release. Can you check if the create package contains *_py27* in the file name? If yes, I am fine with this.

kyleabeauchamp:
FWIW, at some point it might possible to convince Illumina to distribute the actual bcl2fastq package via conda.  Because of their license restrictions, I imagine they might want to have their own channel with a clickthrough agreement.  It's possible but unlikely that they would include allow it on Bioconda, as that would require legal changes to the license.

dkoppstein:
@johanneskoester in fact, the python restrictions resulted in a tarball without the `py27` so I removed them. The build now passes and the new tarball contains `py27`. 

For future reference, from the [bcl2fastq2 license](http://support.illumina.com/content/dam/illumina-support/documents/documentation/software_documentation/bcl2fastq/bcl2fastq2-v2-16-EULA.pdf):

> You may (i) make one copy of the Software solely for purposes of having a backup copy, provided that you reproduce on that copy all copyright notices and any other confidentiality or proprietary legends that are on the original copy of the Software, or (ii) transfer the Software to a single hard disk provided you keep the original solely as a backup copy. You understand that Licensor may update the Software at any time but in doing so, incurs no obligation to furnish such updates to you pursuant to this Agreement. Licensor reserves all rights in the Software not expressly granted to you in this Agreement

Agree with @kyleabeauchamp that it would be nice to contact them to have this important software available on bioconda (since from experience, building it on a non-CentOS system can be a pain).

dkoppstein:
OK, the build passes and the tarball is prefixed bcl2fastq-nextseq-0.1.0-py27_0, so I'll go ahead and merge it @johanneskoester. Thanks!

PR #2102, opened 2016-08-05
perl-excel-writer-xlsx-threaded

lecorguille:

Because this package is the only one with depend on perl and not on perl-threaded, Galaxy fails to find perl-threaded package.
https://github.com/workflow4metabolomics/tool-bank-golm-lib_search/issues/2


bgruening:
You need to increase the build number :-)

lecorguille:
Sure :)

lecorguille:
Help!
Why does it need perl?
Is it the same issue: https://github.com/bioconda/bioconda-recipes/pull/1414#issuecomment-219008937 ?

PR #2106, opened 2016-08-06
Build on osx

bgruening:


croth1:
Seems like the macOS version is not available yet. Do you have to increase the build number in order to trigger the upload?

bgruening:
:(, it's on our R-blacklist, because we try to migrate everything to R 3.3. How urgent is this for you?
ping @daler, seems that we can not update an R package now.

daler:
Until we include `CONDA_R: 3.3.1` in `env_matrix.yml`, we should be able to build for R 3.2.

PR #2107, opened 2016-08-08
New recipe: IgDiscover

marcelm:

@bioconda/core: This is a new recipe; I would be glad for a review.

### DIFF
--- a/recipes/igdiscover/meta.yaml
+++ b/recipes/igdiscover/meta.yaml
@@ -1,0 +1,47 @@
+about:
+  home: https://igdiscover.readthedocs.io/
+  license: 'MIT'
+  summary: 'Analyze antibody repertoires and discover new V genes'
+package:
+  name: igdiscover
+  version: "0.3"
+source:
+  fn: igdiscover-0.3.tar.gz
+  url: https://pypi.python.org/packages/ba/ff/bc398601c92d05bf502bcd3f49897b20cd3a0fe41cde0f17c8f073222ad4/igdiscover-0.3.tar.gz#md5=132e771db3e168018a613c0900b560d9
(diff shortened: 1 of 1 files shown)


recipes/igdiscover/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+package:
+  name: igdiscover
+  version: "0.3"
+source:
+  fn: igdiscover-0.3.tar.gz
+  url: https://pypi.python.org/packages/ba/ff/bc398601c92d05bf502bcd3f49897b20cd3a0fe41cde0f17c8f073222ad4/igdiscover-0.3.tar.gz#md5=132e771db3e168018a613c0900b560d9
+  md5: 132e771db3e168018a613c0900b560d9
+requirements:
+  run:
+    - python >=3.3
```
johanneskoester: Constraining the Python version here usually results in conda-build not pinning the Python version in the package. This is clearly a shortcoming in conda, but until it is solved, we have to omit the constraint here and rely solely on the skip option below.

recipes/igdiscover/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - scipy ==0.16.1
+    - pyyaml
+    - muscle
+    - pear
+    - flash
+    - igblast ==1.4.0
+    - vsearch ==1.9.7
+    - fastqc
+  build:
+    - python >=3.3
```
johanneskoester: Same as above.

johanneskoester:
Nice, is this the tool you have told me about at the ISMB, right? Feel free to merge after my comments have been addressed.

marcelm:
Yes, that’s the tool! Thanks for the comments; I’ll address them.

PR #2144, opened 2016-08-15
Add recipe for mzml2isa, including dependencies

RJMW:

@bioconda/core

### DIFF
--- a/recipes/chainmap/bld.bat
+++ b/recipes/chainmap/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install 
+if errorlevel 1 exit 1
+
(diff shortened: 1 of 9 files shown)


RJMW:
@bgruening - Please, can you have a look at the Travis CI build. It seems that the order in which the packages/dependencies are build is incorrect. The builds inside the (local) docker container run without any issues. Any suggestions?

bgruening:
Super Ralf!

RJMW:
Thanks for the quick reply! Indeed skipping the py3k works. It would be great if we can fix the DAG generator at some point. Shall we open an issue for this?

bgruening:
Please do so and attach the yaml please. It's kind of unusal how you pined your version.

PR #2154, opened 2016-08-16
Add recipe for nmrml2isa

RJMW:

@bioconda/core

### DIFF
--- a/recipes/nmrml2isa/meta.yaml
+++ b/recipes/nmrml2isa/meta.yaml
@@ -8,7 +8,7 @@ source:
   md5: 3f56e532da5ab8d7ee62bd32c1d21253
 
 build:
-  skip: True # [py26 and py3k]
+  skip: True # [py3k]
   number: 0
   entry_points:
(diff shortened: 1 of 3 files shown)


recipes/nmrml2isa/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+build:
+  skip: True # [py26 and py3k]
+  number: 0
+  entry_points:
+    - nmrml2isa = nmrml2isa.parsing:run
+
+requirements:
+  build:
+    - python >=2.7,<3
```
johanneskoester: Constraining Python like this leads to conda not pinning to the minor Python version. This usually leads to problems later on. The canonical way to avoid this, is to just use build-skip and keep Python in run and build requirements unconstrained. Can you also fix your mzml2isa recipe in this way, please?

johanneskoester:
If the tests pass, this is ready for merging.

RJMW:
Thanks for checking the recipes. I have updated both recipes as requested.

PR #2157, opened 2016-08-17
Update SPAdes to 3.9.0

mdehollander:

NEW: rnaSPAdes pipeline for de novo transcriptome assembly from RNA-Seq data.

CHANGE: Improved memory consumption in metagenomic pipeline.

FIX: Several minor bugs.

### DIFF
--- a/recipes/spades/meta.yaml
+++ b/recipes/spades/meta.yaml
@@ -1,12 +1,12 @@
 package:
   name: spades
-  version: '3.8.1'
+  version: '3.9.0'
 
 source:
-  fn: SPAdes-3.8.1-Linux.tar.gz                                              # [linux]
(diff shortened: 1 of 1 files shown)


mdehollander:
@bgruening Is there something wrong with the osx image? The installation of this software package should be straight forward. No compiling, just unpacking the binaries... The travis build complains that there is no output for 10 minutes...

bgruening:
Travis was just hanging. Now everything is ok.
@mdehollander ideally this would be compiled from source to make then really work on HPC.

mdehollander:
@bgruening Thanks. I already thought it was something like that. Indeed, using the source files seems better. This time I only adjusted the version and I didn't want to do a complete makeover of the original installation files. With the next update I will have a look at building it from source.

PR #2174, opened 2016-08-18
Build r-jpeg on osx as testing for bioconda-utils

bgruening:


johanneskoester:
Apparently it did not build anything. r-jpeg is in the R channel by now (in the same version). We should remove it from bioconda (see PR #2189). For this test, I guess you have to choose another package.

PR #2177, opened 2016-08-19
smalt - adding recipe for SMALT aligner

notestaff:
* [X] This PR adds a new recipe.  @bioconda/core

### DIFF
--- a/recipes/smalt/meta.yaml
+++ b/recipes/smalt/meta.yaml
@@ -6,26 +6,24 @@ package:
   version: "0.7.6"
 source:
   fn: smalt-0.7.6-static.tar.gz
-  url: https://sourceforge.net/projects/smalt/files/smalt-0.7.6-static.tar.gz/download
+  url: https://sourceforge.net/projects/smalt/files/smalt-0.7.6-static.tar.gz
   md5: 57ed011836fda8fb5e34604dd1858b5e
 requirements:
   build:
     - gcc
-    - autoconf
-    - automake
     - libtool
     - pkg-config
-    - python
     - bambamc
     - zlib
+    - python >=2.6,<3
   run:
     - bambamc
     - zlib
     - libgcc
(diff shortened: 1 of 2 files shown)


notestaff:
@bioconda/core

recipes/smalt/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - gcc
+    - autoconf
+    - automake
+    - libtool
+    - pkg-config
+    - python
+    - bambamc
+    - zlib
+  run:
+    - bambamc
```
bgruening: zlib also needs to be a runtime dependency

notestaff: thanks for catching, added zlib

recipes/smalt/meta.yaml
```diff
@@ -0,0 +1,32 @@
+build:
+  number: 0
+
+package:
+  name: smalt
+  version: "0.7.6"
+source:
+  fn: smalt-0.7.6-static.tar.gz
+  url: https://sourceforge.net/projects/smalt/files/smalt-0.7.6-static.tar.gz/download
```
bgruening: the leading `/download` can be removed

recipes/smalt/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+  md5: 57ed011836fda8fb5e34604dd1858b5e
+requirements:
+  build:
+    - gcc
+    - autoconf
+    - automake
+    - libtool
+    - pkg-config
+    - python
```
bgruening: Is python really needed, couldn't find a reference.

notestaff: python 2 is used for testing ("make check") during the build.
this fails on osx so for now i've disabled building there.
i've removed python from runtime dependencies.

PR #2178, opened 2016-08-19
adding recipes for the iva viral genome assembler, and for its two dependencies kmc and smalt .

notestaff:


notestaff:
@bioconda/core

PR #2182, opened 2016-08-19
Add recipe for icqsol version 0.3.20

gregvonkuster:


recipes/icqsol/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - numpy
+    - triangle
+    - vtk
+
+requirements:
+  build:
+    - cmake
+    - numpy
+    - python
+    - pycsg ==0.3.12
```
bgruening: Is this strict pinning needed?

recipes/icqsol/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+  imports:
+    - csg
+    - icqsol
+    - numpy
+    - triangle
+    - vtk
+
+requirements:
+  build:
+    - cmake
```
bgruening: Please also depend on `gcc # [linux]` and `llvm #[osx]`

recipes/icqsol/meta.yaml
gregvonkuster: In these specific cases where I've set versions, yes because even small changes to these underlying packages most often produce different results for the Galaxy CSG tools that depend on them.  So new versions of these underlying packages will require new Galaxy tool versions.  This is not the case for numpy or python.

recipes/icqsol/meta.yaml
gregvonkuster: ok, I thought this may be required, but I was bumping into issues on Linux because I'm running gcc 6.1 and it looks like some of the required packages here (e.g., vtk) require older versions of gcc, so conda was not compiling this correctly on Linux for me.  I'll add these in though - thanks!

recipes/icqsol/meta.yaml
gregvonkuster: @bgruening Do you think that since the recipes for pycsg and pytriangle specify git-revs that this is sufficient, and that I can eliminate the pinned versions in this recipe?  If so, I will.  Thanks!

bgruening: The thing is if someone updates a new package of pycsg and then you install icqsol you will get this new dependency. If this is working, everything is fine. If you expect a future version breaking this recipe you should bin a version. Maybe <= works, ofter >= is used as well.

recipes/icqsol/meta.yaml
bgruening: You can always use the Docker container to test your recipes.

recipes/icqsol/meta.yaml
gregvonkuster: Ok, my understanding is that if these versions are pinned like this and if someone creates a new version of pycsg, when this version of icqsol is installed, the pinned version of pycsg (i.e., 0.3.12) will come with it.  If I understand correctly, I think we should leave these pinned.  I've tested this on OS X and Linux.  Thanks!

gregvonkuster:
@bgruening Not quite sure how to fix this failing test.  I've tried the directions using conda info but didn't see anything helpful as to what may be in conflict.  Do you have any ideas?

```
...Error: The following specifications were found to be in conflict:
  - gcc
  - vtk ==6.3.0 -> python 2.7*
```

bgruening:
I have it. vtk is only available in py27 and the failing tests you see is on py3k. You need to skip the py3k builds.

bgruening:
Thanks Greg.

PR #2183, opened 2016-08-19
bump lumpy to latest git version and expose cnv-to-bedpe script.

brentp:

### DIFF
--- a/recipes/lumpy-sv/build.sh
+++ b/recipes/lumpy-sv/build.sh
@@ -11,6 +11,8 @@ make
 cp bin/lumpy $PREFIX/bin
 cp scripts/lumpyexpress $PREFIX/bin
 
+cp scripts/cnvantor_to_bedpes.py $PREFIX/bin
+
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
Does this mean you are endorsing the current `git` as something like a "release"?

brentp:
no, this isn't yet an official release. but it contains changes to helper scripts.

do you have a suggestion?

kyleabeauchamp:
I forget: will `conda install` prefer the `-git` version over the previous 2.12?  I'm just curious what this means in terms of versioning.

brentp:
I'm not sure.

chapmanb:
Brent -- thanks for tackling this. If this only contains changes to helper scripts I'd bump the build number from 1 to 2 instead of adding in the extra `git` string. I also don't know what that does but doesn't seem necessary since we were already using a git hash tied to the version.

brentp:
ok. I removed the git string and bumped the version.

PR #2184, opened 2016-08-19
bzip2 - added recipe for bzip2 and libbzip2

notestaff:

Adding this to enable compiling the kmc k-mer counter from source, but of course many other things also have libbzip2 as a dependency.

@bioconda/core

### DIFF
--- a/recipes/bzip2/build.sh
+++ b/recipes/bzip2/build.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+set -euf -o pipefail
+
+make
+make PREFIX=$PREFIX install
+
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
Isn't this already in the default channel?

notestaff:
Couldn't find it in
https://github.com/conda/conda-recipes
Is that the full list of the default channel recipes?

kyleabeauchamp:
I think that list is incomplete.  Look at anaconda.org.

tomkinsc:
I usually check for packages in the default channel via `conda search  <packagename>`. So, `conda search bzip2` shows it in the `defaults` channel (and `conda info` shows more).

PR #2189, opened 2016-08-20
Remove r-jpeg because it is in the r channel by now.

johanneskoester:

Also, our version number seems to be wrong. It should be 0.1-8...


bgruening:
Thanks, will update the other PR now, give me 60min :)

PR #2192, opened 2016-08-20
upgrade dendextend to 1.2.0

bgruening:


bgruening:
Ok, these `r-*` packages seems to work. @johanneskoester please note the R3.3.1 packages, `r-fastcluster-1.1.20-r3.3.1_0` will be created as the R version is not set to an older one. So the first R3.3. packages are already build in bioconda.

PR #2211, opened 2016-08-24
Add recipe for VarScan version 2.3.7

bow:
* [ x ] I have read the guidelines above.
* [ x ] This PR adds a new recipe.

@bioconda/core This is technicaly a new recipe, but it's actually almost the same as the existing recipe in the top level directory of varscan. The actual changes are only:

    * Updating the version number (to 2.3.7)
    * Changing the URL for downloading the JAR

I've tested this locally in my conda environment and also using docker without any errors.

### DIFF
--- a/recipes/varscan/2.3.7/meta.yaml
+++ b/recipes/varscan/2.3.7/meta.yaml
@@ -1,0 +1,19 @@
+about:
+    home: http://dkoboldt.github.io/varscan/
+    license: The Non-Profit Open Software License version 3.0 (NPOSL-3.0)
+    summary: variant detection in massively parallel sequencing data
+package:
+    name: varscan
+    version: 2.3.7
+build:
+  number: 0
+source:
+    fn: VarScan-2.3.7.jar
+    url: http://downloads.sourceforge.net/project/varscan/VarScan.v2.3.7.jar
+requirements:
+  build:
+  run:
+    - java-jdk
+test:
+    commands:
+      - varscan mpileup2cns --help
--- a/recipes/varscan/2.3.7/varscan.sh
+++ b/recipes/varscan/2.3.7/varscan.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+# VarScan executable shell script
+set -eu -o pipefail
+
+set -o pipefail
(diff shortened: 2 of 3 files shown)


recipes/varscan/2.3.7/meta.yaml
```diff
@@ -0,0 +1,19 @@
...
+    license: The Non-Profit Open Software License version 3.0 (NPOSL-3.0)
+    summary: variant detection in massively parallel sequencing data
+package:
+    name: varscan
+    version: 2.3.7
+build:
+  number: 0
+source:
+    fn: VarScan-2.3.7.jar
+    url: http://downloads.sourceforge.net/project/varscan/VarScan.v2.3.7.jar
```
bgruening: can you please include a checksum here? Thanks!

recipes/varscan/2.3.7/varscan.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
bgruening: If you are eager to port this over to the new python wrapper: https://github.com/bioconda/bioconda-recipes/blob/master/GUIDELINES.md#java

This would be great.

recipes/varscan/2.3.7/meta.yaml
bow: That would be useful indeed ~ added in 7e7fe90

recipes/varscan/2.3.7/varscan.sh
bow: Hmm..I suppose you meant the Java wrapper? In that case, I see that the existing `varscan.sh` script seems to conform the guideline still. Am I perhaps missing something?

bgruening: Mh I though about this one: https://github.com/bioconda/bioconda-recipes/tree/master/recipes/peptide-shaker

bow: Aha, it's updated now in ea1e63f :).

yhoogstrate:
Nice one @bow

bow:
Hey thanks for the merge @yhoogstrate ~ and @bgruening for the review :)

Issue #2150, opened 2016-08-16
umi_tools not working after conda update


matthdsm:
Hi,

recently I've submitted the `umi_tools` recipe to bioconda. Everything seemed to work fine up until I updated conda to the latest version. When I try to run the tools I get the following error

```
Traceback (most recent call last):
  File "/Users/matdsmet/miniconda2/bin/umi_tools", line 5, in <module>
    from pkg_resources import load_entry_point
  File "/Users/matdsmet/miniconda2/lib/python2.7/site-packages/setuptools-25.1.6-py2.7.egg/pkg_resources/__init__.py", line 2953, in <module>
  File "/Users/matdsmet/miniconda2/lib/python2.7/site-packages/setuptools-25.1.6-py2.7.egg/pkg_resources/__init__.py", line 2939, in _call_aside
  File "/Users/matdsmet/miniconda2/lib/python2.7/site-packages/setuptools-25.1.6-py2.7.egg/pkg_resources/__init__.py", line 2966, in _initialize_master_working_set
  File "/Users/matdsmet/miniconda2/lib/python2.7/site-packages/setuptools-25.1.6-py2.7.egg/pkg_resources/__init__.py", line 635, in _build_master
  File "/Users/matdsmet/miniconda2/lib/python2.7/site-packages/setuptools-25.1.6-py2.7.egg/pkg_resources/__init__.py", line 943, in require
  File "/Users/matdsmet/miniconda2/lib/python2.7/site-packages/setuptools-25.1.6-py2.7.egg/pkg_resources/__init__.py", line 829, in resolve
pkg_resources.DistributionNotFound: The 'umi-tools==0.2.3' distribution was not found and is required by the application
```
I already tried to `force install` the recipe and conda to resolve any issues.
Anyone have an idea what could be causing this?

`conda list`output:

```
# packages in environment at /Users/matdsmet/miniconda2:
#
bcftools                  1.3.1                         0    bioconda
curl                      7.45.0                        1    bioconda
htslib                    1.3.1                         0    bioconda
pysam                     0.9.1                    py27_0    bioconda
python-dateutil           2.3                      py27_0    bioconda
samtools                  1.3.1                         2    bioconda
umi_tools                 0.2.3                         1    bioconda
conda                     4.1.11                   py27_0  
conda-env                 2.5.2                    py27_0  
cython                    0.24.1                   py27_0  
future                    0.15.2                   py27_0  
mkl                       11.3.3                        0  
[...]

PR #2223, opened 2016-08-26
Added python3 dependency

matthdsm:

added python3 dependency for system that use 2.7 by default,
also bumped build number


matthdsm:
should fix #2150

Issue #1971, opened 2016-07-13
Poster badge


ewels:
Quoting @tomkinsc in #1951:
> Maybe we should have a logo available similar to the GitHub badge for groups to add to conference posters. Something like:
> ```
> ┌ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┐                                         
> 
> │                         │ Reproducible research.                  
>        Bioconda logo        Install the software from bioconda:     
> │                         │ conda install ____________________      
>                             See: bioconda.github.io                 
>└ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ─ ┘                                         
> ```

Mockup of something along those lines here:  http://beta.tallphil.co.uk/bioconda-badge/

[Source here](https://github.com/bioconda/bioconda-recipes/files/360845/poster_badge.zip). Resulting images look like this:

---
![image](https://cloud.githubusercontent.com/assets/465550/16797099/397e447e-48e6-11e6-94e9-e5a8af8fb3a2.png)

---
![image](https://cloud.githubusercontent.com/assets/465550/16797109/44b14422-48e6-11e6-9590-473e3c2822e4.png)

---

To do:

- [x] Add `-c bioconda` (suggested by @bgruening)
- [x] Make the `-c bioconda` optional, with checkbox
- [x] Make a compact version with smaller logo font
- [x] Make monochrome versions in black & white for coloured backgrounds?
- [x] Make a static version - JS only

bgruening:
Love it!!!

ewels:
Added `-c bioconda` to the page, though now only packages with short names will fit on the wide version. I'll update the SVG to be wider.

tomkinsc:
Wonderful! Looks great! Nice work, @ewels.

While it's more "correct", I'd say we can omit the "-c bioconda" part since it clutters the design and deemphasizes how easy it is to install something from bioconda. I guess I see configuring conda to use the bioconda channel as part of the initial setup process for most people, like installing miniconda. It's described on the GitHub page and doesn't need to be on the poster (especially when we include the url for the GitHub page). Maybe the generator could include a checkbox option to add it?

ewels:
I like the idea of making it optional. Then people can choose (probably depending on how long their package name is).

[...]

PR #2227, opened 2016-08-26
add badge generator page

tomkinsc:
This is a static HTML page derived from one created by @ewels in #1971. It is capable of rendering SVG and PNG badges with custom text, suitable for promoting bioconda packages in posters and presentations. This commit differs from the original in that the badges were re-created in Inkscape so the SVGs are accessible to those without Illustrator. The svg files are now loaded in the page dynamically via javascript so the badge files can be edited externally, obviating the need to copy+paste from inline svg source. Inkscape will preserve tspans, so it is possible to edit and see design changes immediately. The modifications also add javascript to download a base64-encoded data string representation of the SVG designs, as well as canvas-rendered PNG files at 4x resolution. The “console” text has a genetic font monospace font family now instead of only Monaco. The SVG files scale correctly in Inkscape (despite fixed-font for the console text). Also added a "very compact" badge design.

Added a link to the docs, with the badge generator files stored in _static/. 

Of note is that because the SVG files are loaded dynamically via javascript, cross-origin browser restrictions prevent the static page from working correctly when viewed locally from a `file://` URL, since most browsers restrict it. On an actual server it works fine, so local previews can be accomplished by spinning up a simple server in python via `python -m http.server` (py3) or `python -m SimpleHTTPServer`.

screen shot 2016-08-26 at 12 03 08 pm


kyleabeauchamp:
Figures look great, thanks!

PR #2230, opened 2016-08-28
Feature/perl recipes

jerowe:

I'm adding in some perl packages.


jerowe:
@bgruening , what happens if I require perl-threaded, but somewhere in the dep tree perl is also required?

daler:
Is it a bioconda package that requires perl? If so it should be changed; otherwise I think the solution is to build up the dep tree from the bottom with perl-threaded.

jerowe:
@daler, I'm fairly certain they are packages built from conda skeleton. I think the default should just be perl-threaded, but sometimes its perl and sometimes its perl-threaded. 

I don't have any particular issue with going and rebuilding the dep tree, but I don't want to break compatibility with any packages that depend upon perl (no threads).

daler:
Only in bioconda do we have the perl-threaded requirement, so you'll have to edit the output of the general conda-skeleton from perl to perl-threaded.

daler:
With respect to breaking compatibility, there should (theoretically) be no vanilla perl packages in bioconda, so that (again theoretically) breaking compatibility should not be an issue. But I guess the issue is that if you have a recipe that doesn't exist in bioconda but exists in the default conda channel, and that default conda channel does not use perl-threaded, then it would be pulled in and would be using the wrong interpreter.

Not sure at this point what the best solution would be, other than manually checking which packages are brought in as dependencies and making sure a bioconda version with perl-threaded exists for all of them.

Also, I agree we should pin. Might be tricky though since some of those "pins" are built in to conda. E.g., [this](https://github.com/bioconda/bioconda-recipes/blob/master/scripts/env_matrix.yml#L7) is where we're pinning the perl version but it doesn't specify threaded or not.

bioconda-utils will have a skeleton that will create the proper dependencies as well as a linter that will identify and flag existing non-perl-threaded recipes, which will hopefully cut down on the perl-packaging headaches.

jerowe:
@daler,  I think there are perl deps listed as -perl, but for perl packages and not bioconda packages.

I wouldn't mind making it a project to go through and rebuild all perl-* packages to make sure there is no vanilla perl, but I wont be able to tackle that this month. Perl is a fairly major part of our infrastructure though, and I'd rest easier knowing I'm only getting perl-threaded. So let me know please!

daler:
It would be fantastic to have better/more complete perl support. I think you've packaged up the majority of Perl packages here, so it will be great to have you working on this.

I know there are some changes coming with conda to handle namespaces (see [here](https://gist.github.com/mcg1969/da5aec380d2ed083b79ddcf151ca16f1)) and perl will soon be in its own namespace. This may also require rebuilding a bunch of packages. So I would say to hold off on major work until we understand how those changes are going to affect us.

jerowe:
Its good to see that everyone is thinking about this sort of thing. ;-) Like I said, I can't get to it this month anyways. I'm perfectly happy to wait. I will have to try to follow the namespaces conversation also.

PR #2232, opened 2016-08-28
Add recipe for Ococo

karel-brinda:

@bioconda/core

### DIFF
--- a/recipes/ococo/build.sh
+++ b/recipes/ococo/build.sh
@@ -13,3 +13,5 @@ export LIBRARY_PATH=${PREFIX}/lib
 make
 mkdir -p $PREFIX/bin
 cp ococo $PREFIX/bin
+cp ococo.1 ${PREFIX}/share/man/man1
+
--- a/recipes/ococo/build.sh
+++ b/recipes/ococo/build.sh
@@ -16,7 +16,7 @@ g++ -v || true
 echo "CXX"
 $CXX -v || true
 
-make
+make VERBOSE=1
 mkdir -p $PREFIX/bin
 cp ococo $PREFIX/bin
 cp ococo.1 ${PREFIX}/share/man/man1
--- a/recipes/ococo/build.sh
+++ b/recipes/ococo/build.sh
@@ -11,6 +11,10 @@ export LIBRARY_PATH=$LIBRARY_PATH:${PREFIX}/lib
 	git clone git://github.com/samtools/htslib
 )
 
+echo "g++"
+g++ -v || true
+echo "CXX"
+$(CXX) -v || true
 
 make
 mkdir -p $PREFIX/bin
--- a/recipes/ococo/build.sh
+++ b/recipes/ococo/build.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 export C_INCLUDE_PATH=${PREFIX}/include
-export CPPFLAGS="-I$PREFIX/include"
+#export CPPFLAGS="-I$PREFIX/include"
 export CPLUS_INCLUDE_PATH=${PREFIX}/include
 export LIBRARY_PATH=${PREFIX}/lib
 
--- a/recipes/ococo/build.sh
+++ b/recipes/ococo/build.sh
@@ -1,9 +1,9 @@
 #!/bin/bash
 
-export C_INCLUDE_PATH=${PREFIX}/include
+export C_INCLUDE_PATH=$C_INCLUDE_PATH:${PREFIX}/include
 #export CPPFLAGS="-I$PREFIX/include"
-export CPLUS_INCLUDE_PATH=${PREFIX}/include
-export LIBRARY_PATH=${PREFIX}/lib
+export CPLUS_INCLUDE_PATH=$CPLUS_INCLUDE_PATH:${PREFIX}/include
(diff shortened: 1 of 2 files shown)


karel-brinda:
Do you have any idea why I am still getting this error? ZLib should be installed.

```
In file included from caller.h:5:0,
                 from ococo.h:3,
                 from main.cpp:1:
stats.h:14:18: fatal error: zlib.h: No such file or directory
 #include <zlib.h>
                  ^
compilation terminated.
```

daler:
Tough to tell why since it's typically a package-dependent issue, but there are some other things to try in the [zlib section of the guidelines](https://github.com/bioconda/bioconda-recipes/blob/master/GUIDELINES.md#zlib).

karel-brinda:
@daler Thank you very much. Now I have the following problem:
```
ococo: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.11' not found (required by ococo)
```
Do you have any idea about how to fix it?

daler:
Can you try depending on libgcc as well?

karel-brinda:
@daler Thank you very much. On Linux it is ok now. There is still a problem on OS X:
```
./types.h:4:10: fatal error: 'cinttypes' file not found
#include <cinttypes>
         ^
1 error generated.
```
The reason is probably an old version of Xcode (I have never seen this error on my OS X laptop).

daler:
Hmm. Easy option is to skip the OSX build; otherwise you may have to figure out a [patch](http://conda.pydata.org/docs/building/meta-yaml.html#patches) to include to make this work.

bgruening:
Can you include libgcc as run time dependency.
On Aug 29, 2016 11:56 AM, Karel Břinda <notifications@github.com> wrote:@daler Thank you very much. Now I have the following problem:

ococo: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.11' not found (required by ococo)


Do you have any idea about how to fix it?

karel-brinda:
@daler I could fix directly Ococo but I don't know how. I can't reproduce this error outside Bioconda. All Travis tests for Ococo pass on OS X, see https://travis-ci.org/karel-brinda/ococo/jobs/155734173.

daler:
Definitely agree that it's a bioconda-specific issue and not even a travis-ci issue, since your tests pass on their osx environment . . . does it help if you change libgcc to be only for linux, e.g., 

```yaml
  run:
    - libgcc    # [not osx]
```

daler:
Looks like that didn't work either. Sorry I'm not much help on this front. @bioconda/core, any ideas on why our OSX LLVM is not finding a standard C++ header file?

karel-brinda:
I managed to fix the last problem by this update:  https://github.com/karel-brinda/ococo/commit/fdbd0877c3cc69dd5388c83a1b986d458fcc89ce#diff-0a1f40fc150a973726797865e772a344L4

Nevertheless, now I am getting another strange error. It looks like that there is some strange compiler.

```
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C ./src
c++ -std=c++11 -Wall -Wextra -Wno-missing-field-initializers -g -O2  -c misc.cpp -I ../ext/htslib
c++ -std=c++11 -Wall -Wextra -Wno-missing-field-initializers -g -O2  -c main.cpp -I ../ext/htslib
In file included from main.cpp:1:
In file included from ./ococo.h:3:
In file included from ./caller.h:5:
./stats.h:143:9: error: use of undeclared identifier 'errno'
    if (errno != 0 || fp == nullptr) {
        ^
In file included from main.cpp:1:
In file included from ./ococo.h:4:
./consensus.h:12:50: warning: unused parameter 'params' [-Wunused-parameter]
                                 const params_t &params) {
                                                 ^
1 warning and 1 error generated.
make[1]: *** [main.o] Error 1
make: *** [ococo] Error 2
```

daler:
Looks like your [Makefile](https://github.com/karel-brinda/ococo/blob/master/Makefile#L15) is using `CXXFLAGS` but in the build.sh in this PR you're exporting a new `CPPFLAGS`. Do you need to export a `CXXFLAGS` in this PR's build.sh as well?

Another thing to try is look in the `find recipes -name "*.patch"` files to see if anyone has a fix for similar problems.

karel-brinda:
@daler It finally works, I had to remove ``errno``. It looks like there is some strange C++ compiler.

daler:
Whew, nice!

Very strange about the `errno`. I'm guessing the reason is the LLVM package dependency we're using to maintain cross-package compatibility (rather than the default travis-ci OSX environment). Were you able to find any information on it or was it trial-and-error? Anything we can add to the guidelines to streamline similar work in the future?

karel-brinda:
I was a long trial-and-error chain. It looks like that the compiler does not fully support C++ standard. It might be also a problem of my configuration.

The problem with ``cinttypes`` can be fixed by the following (I found this trick on https://bugs.launchpad.net/gearmand/+bug/1358084):
```cpp
#ifdef _LIBCPP_VERSION
	#include <cinttypes>
#else
	#include <tr1/cinttypes>
#endif
```

Concerning ``errno``, I did not find any way how to make it work. So I removed it from Ococo.

daler:
Ok thanks.

Looks good by the way, feel free to merge when you're ready.

daler:
@bgruening when you get a chance -- do you know what's happening here?

PR #2240, opened 2016-08-30
Updated planemo to version 0.29.1

blankclemens:

Had to update galaxy-lib to 16.7.10 and add the aenum dependency as a new recipe.

Should the old versions for galaxy-lib and planemo remain?

@bioconda/core


bgruening:
Great thanks!

PR #2241, opened 2016-08-30
WIP: Adding recipe for HLA-MA

holtgrewe:

@bioconda/core

Hi, this adds a recipe for the HLA-MA tool.

The code builds fine with `conda build` but fails in Docker. Could someone point me in the right direction to resolve this?

### DIFF
--- a/recipes/hlama/meta.yaml
+++ b/recipes/hlama/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: hlama
-  version: "0.2"
+  version: '0.3'
 
 build:
   number: 0
@@ -19,6 +19,8 @@ requirements:
     - snakemake ==3.7.1
   run:
     - python
+    - optitype >=1.2
+    - yara >=0.9.6
     - snakemake ==3.7.1
 
 test:
--- a/recipes/hlama/meta.yaml
+++ b/recipes/hlama/meta.yaml
@@ -9,8 +9,8 @@ build:
     - hlama = hlama.app:main
 
 source:
-  fn: hlama-develop.tar.gz
-  url: https://github.com/bihealth/hlama/archive/develop.tar.gz
+  fn: hlama-0.3.tar.gz
+  url: https://github.com/bihealth/hlama/archive/v0.3.tar.gz
 
 requirements:
   build:
--- a/recipes/hlama/meta.yaml
+++ b/recipes/hlama/meta.yaml
@@ -4,7 +4,7 @@ package:
 
 build:
   number: 0
-  skip: True # [py27]
+  skip: True # [py27 or osx]
   entry_points:
     - hlama = hlama.app:main
 
--- a/recipes/hlama/meta.yaml
+++ b/recipes/hlama/meta.yaml
@@ -1,0 +1,31 @@
+package:
+  name: hlama
+  version: "0.2"
+
+build:
+  number: 0
+  skip: True # [py27]
+  entry_points:
+    - hlama = hlama.app:main
+
+source:
+  fn: hlama-develop.tar.gz
+  url: https://github.com/bihealth/hlama/archive/develop.tar.gz
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - snakemake ==3.7.1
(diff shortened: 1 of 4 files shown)


blankclemens:
Activating the console output in the test command shows an error, maybe that can help:

    + hlama --help
      File "/anaconda/envs/_test/bin/hlama", line 4, in <module>
        import hlama.app
    ImportError: No module named 'hlama'

holtgrewe:
Thanks, how did you activate it?

blankclemens:
i just didn't put the output on `dev/null` ;) :

`hlama --help` instead of `hlama --help &>/dev/null`

But I don't know how to interpret this error - building with conda works fine, I can confirm that.

recipes/hlama/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  fn: hlama-develop.tar.gz
+  url: https://github.com/bihealth/hlama/archive/develop.tar.gz
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - snakemake ==3.7.1
+  run:
+    - snakemake ==3.7.1
```
bgruening: Runtime dependency on python is missing.

recipes/hlama/meta.yaml
```diff
@@ -0,0 +1,31 @@
+package:
+  name: hlama
+  version: "0.1"
+
+build:
+  number: 1
```
bgruening: New packages should start with 0

holtgrewe:
OK, thanks for the help.

I have created a patch for FRED-2/OptiType#36 to make it support Python 3. If this is committed, I'll include an update for the OptiType recipe and then directly depend on this so everything is in Python 3 and dandy. Otherwise, I'll think of something...

bgruening:
@holtgrewe is this ready to go?

holtgrewe:
@bgruening as optitype now supports Python 3, let me simplify the installation a bit first

recipes/hlama/meta.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: hlama
-  version: "0.2"
+  version: '0.3'
```
bgruening: Can you please use a stable URL like: https://github.com/bihealth/hlama/archive/v0.3.tar.gz

holtgrewe:
Hm, I don't understand the problem here. Something in the dependency chain does not support Python 3.5. I think the problem is coming from installation on Python 3.5... What can I do to fix this problem?

https://travis-ci.org/bioconda/bioconda-recipes/jobs/159298139

bgruening:
@holtgrewe the reason is that pyomo is not available for python 3.5. One way would be to update pyomo and build it for python 3.5 as well. But this involves more work as you see here: https://github.com/bioconda/bioconda-recipes/pull/2312

Feel free to improve this branch.

holtgrewe:
Could I somehow skip the 3.5 build for now so we can merge this now and then upgrade pyomo somehow?

bgruening:
You can add a skip: True #[py35]
On Sep 14, 2016 8:22 AM, Manuel Holtgrewe <notifications@github.com> wrote:Could I somehow skip the 3.5 build for now so we can merge this now and then upgrade pyomo somehow?

recipes/hlama/meta.yaml
```diff
@@ -0,0 +1,33 @@
+package:
+  name: hlama
+  version: '0.3'
+
+build:
+  number: 0
+  skip: True # [py27 or py35 or osx]
```
holtgrewe: @bgruening I tried and it did not work. Can you help here? Thanks

bgruening: Try to skip py3k. I will look at it in the evening.

holtgrewe: Ah, HLA-MA only works for Python 3 because it depends on Snakemake.

No worries, no hurry. Thanks! 🙇

holtgrewe: ping, any idea?

PR #2246, opened 2016-08-30
Compile FastTree with double precision

wasade:

Use double precision with FastTree. It defaults to single precision which can result in numerical stability issues. See concerns [here](http://darlinglab.org/blog/2015/03/23/not-so-fast-fasttree.html)

### DIFF
--- a/recipes/fasttree/build.sh
+++ b/recipes/fasttree/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-gcc -O3 -finline-functions -funroll-loops -Wall -o FastTree FastTree.c -lm
+gcc -O3 -DUSE_DOUBLE -finline-functions -funroll-loops -Wall -o FastTree FastTree.c -lm
 
 mkdir -p $PREFIX/bin
 cp FastTree $PREFIX/bin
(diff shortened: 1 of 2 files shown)


bgruening:
@wasade can you please increase the build number as well. Thanks.

bgruening:
Urgs. Now we have a real problem here. The checksum does not match anymore. Probably due to changes in the .c file which is not versioned. Can you please use cargo-port https://github.com/galaxyproject/cargo-port/ to backup this file. Include the new link here, increase the probably new version number and set the build number back to 0. Thanks.

wasade:
Don't know what that is, so I just made the URL version specific. Same MD5. Sound good?

bgruening:
If you find the old file that fits the checksum even better. The crucial part is to make the downloaded file sustainable. Cargo-port can help with this.

wasade:
Right:

```bash
$ md5 FastTree-2.1.8.c
MD5 (FastTree-2.1.8.c) = c7e85689a26cabba241378d4f633c2fa
```

bgruening:
Thanks @wasade!

PR #2269, opened 2016-09-02
Building vt for osx

druvus:

### DIFF
--- a/recipes/vt/build.sh
+++ b/recipes/vt/build.sh
@@ -1,4 +1,8 @@
 #!/bin/bash
+export CFLAGS="-I$PREFIX/include"
+export LDFLAGS="-L$PREFIX/lib"
+export CPATH=${PREFIX}/include
+
 make
 mkdir -p $PREFIX/bin
 cp vt $PREFIX/bin
--- a/recipes/vt/meta.yaml
+++ b/recipes/vt/meta.yaml
@@ -12,7 +12,7 @@ build:
 
 requirements:
   build:
-    - gcc   # [not osx]
+    - gcc ==4.8.2  # [not osx]
     - llvm  # [osx]
     - zlib
 
--- a/recipes/vt/meta.yaml
+++ b/recipes/vt/meta.yaml
@@ -14,9 +14,11 @@ requirements:
   build:
     - gcc   # [not osx]
     - llvm  # [osx]
+    - zlib
 
   run:
     - libgcc    # [not osx]
+    - zlib
 
 test:
   commands:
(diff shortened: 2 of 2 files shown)


druvus:
Putting building vt for osx on hold for the moment. I can't get it to build using the current centos image or for osx. Building nicely in a clean environment on my mac. 

@bioconda/core any hints what is causing the errors?

johanneskoester:
Yes, we have seen that before. Apparently, Continuum managed to break compatibility of gcc 4.8.5 with CentOS 5. We can pin to the older 4.8.2, in order to avoid this problem for now.
@bioconda/core Mike Sarahan from Continuum does not seem willingly to investigate this further. Hence, I think we should move to CentOS 6 soon. 
I know this is controversial, since some old clusters won't be supported any more. But on the other hand, we would get compatibility with conda forge, and sooner or later, Continuum will also build only on CentOS 6 since they replaced their build infrastructure with that of conda-forge as far as I understood.

bgruening:
@johanneskoester +1 for me to use the conda-forge build system based on CentOS-6. I for myself will also have problems with this move and need to convince my IT department to migrate, but I think this is the correct way for the project.

kyleabeauchamp:
Agreed in big picture, modulo any outstanding issues that may need resolution.

image

johanneskoester:
I see three possible paths:

1. introduce a CONDA_GCC env var to set GCC version globally (that var would need to be referenced in any gcc based recipe using jinja).
2. Omit the pinning and directly try to update our docker container to CentOS 6.
3. Derive a docker image from the condaforge/linux-anvil image. All we need to add is our build scripts and their dependencies.

bgruening:
3.!

daler:
I prefer 3 as well.

I feel like waiting for clusters to upgrade past CentOS5 means we lose this opportunity to align our build system with conda-forge at the ground level. It will otherwise be hard to know when the "right" time is (what's the threshold number of clusters on CentOS5 required to keep us on CentOS5? How do we monitor the fraction of clusters on CentOS5 in the first place?).

If we go CentOS6 should set a deadline to help those using CentOS5 plan for how to manage the changes. We could run tests on both 5 and 6 and make the results available so those on 5 know which packages won't work. I'd like to generalize the build system so that folks can put bioconda recipes into their own repo (and channel) to easily build packages for other OSs as needed for those problematic packages. In other words, I favor going to 6 but while providing tools and info to help users on 5.

druvus:
Closing this as it doesn't build for osx and @bgruening opened #2283 that takes care of the zlib issues.

PR #2289, opened 2016-09-05
Gd jpeg

bgruening:

### DIFF
--- a/recipes/perl-gd/meta.yaml
+++ b/recipes/perl-gd/meta.yaml
@@ -6,7 +6,7 @@ source:
   url: https://cpan.metacpan.org/authors/id/L/LD/LDS/GD-2.56.tar.gz
   md5: c4b3afd98b2c4ce3c2e1027d101a8f1e
 build:
-  number: 2
+  number: 3
 
 requirements:
   build:
--- a/recipes/perl-gd/meta.yaml
+++ b/recipes/perl-gd/meta.yaml
@@ -15,10 +15,12 @@ requirements:
     - perl-threaded >=5.22.0
     - libgd
     - perl-module-build
+    - jpeg
   run:
     - libgcc # [not osx]
     - perl-threaded >=5.22.0
     - libgd
+    - jpeg
 test:
   imports:
     - GD


bgruening:
@johanneskoester here is the perl bug again with the old buildsystem :(

johanneskoester:
The bug is gone when using conda 2.0.0, I'm working on it in that test branch. However, it points to another problem that we need to solve soon: since we use perl-threaded, the automatic pinning does not work for us. We either need to do that manually for all perl packages or rename our threaded perl build to "perl". This would still work with the new channel priorities. Also, we could encode the threaded feature as a proper "feature", similar to how continuum is doing it with e.g. MKL.

bgruening:
@johanneskoester has anyone tried to get the threaded feature upstream?
I can imagine the conda-forge people would like to have this in the main perl package. I think this would be much more preferable.

johanneskoester:
Indeed. I never really bothered with perl, but @tomkinsc might know.

tomkinsc:
There's a minor performance penalty to using single-threaded perl scripts on `perl-threaded`, but IMHO the benefits for performance and compatibility justify renaming `perl-threaded` to `perl`. (Maybe making a `perl-singlethreaded`) I haven't contacted anyone upstream about this.

bgruening:
Oh this is great! Than lets do sed 's/perl-threaded/perl/' as soon as we rebuild everything and @johanneskoester merges the build infrastructure.

@tomkinsc do you mind to submit a PR against https://github.com/conda-forge/perl-feedstock

tomkinsc:
@bgruening: [Submitted](https://github.com/conda-forge/perl-feedstock/pull/6).

tomkinsc:
Update: The build of perl with threads on conda-forge [is failing](https://circleci.com/gh/conda-forge/perl-feedstock/8), and I'm not sure of the fix (especially in their build environment). Any ideas, @bioconda/core?
https://github.com/conda-forge/perl-feedstock/pull/6#issuecomment-245012337

johanneskoester:
Sorry, I have no experience with Perl unfortunately (or maybe fortunately ;-)).

PR #2293, opened 2016-09-06
require libgcc

bgruening:

### DIFF
--- a/recipes/sqt/meta.yaml
+++ b/recipes/sqt/meta.yaml
@@ -32,3 +32,4 @@ test:
 build:
   skip: True  # [py27]
   script: python3 setup.py install
+  number: 1
--- a/recipes/sqt/meta.yaml
+++ b/recipes/sqt/meta.yaml
@@ -16,6 +16,8 @@ requirements:
     - pysam <0.9.0
     - matplotlib
     - seaborn
+    - libgcc
+
   build:
     - python
     - cutadapt


bgruening:
I get this error:

```
docker run -it quay.io/mulled/sqt:0.7.0--py35_0 sqt --version
  File "/usr/local/bin/sqt", line 11, in <module>
    load_entry_point('sqt==0.7.0', 'console_scripts', 'sqt')()
  File "/usr/local/lib/python3.5/site-packages/sqt/__main__.py", line 42, in main
    module = importlib.import_module('.commands.' + command_name, 'sqt')
  File "/usr/local/lib/python3.5/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 986, in _gcd_import
  File "<frozen importlib._bootstrap>", line 969, in _find_and_load
  File "<frozen importlib._bootstrap>", line 958, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 665, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "/usr/local/lib/python3.5/site-packages/sqt/commands/readlenhisto.py", line 13, in <module>
    import matplotlib.pyplot as plt
  File "/usr/local/lib/python3.5/site-packages/matplotlib/pyplot.py", line 29, in <module>
    import matplotlib.colorbar
  File "/usr/local/lib/python3.5/site-packages/matplotlib/colorbar.py", line 32, in <module>
    import matplotlib.artist as martist
  File "/usr/local/lib/python3.5/site-packages/matplotlib/artist.py", line 14, in <module>
    from .transforms import (Bbox, IdentityTransform, TransformedBbox,
  File "/usr/local/lib/python3.5/site-packages/matplotlib/transforms.py", line 39, in <module>
    from matplotlib._path import (affine_transform, count_bboxes_overlapping_bbox,
ImportError: libstdc++.so.6: cannot open shared object file: No such file or directory
```

Some module needs to require libgcc. I'm not sure which one, hopefully this fixes it.
ping @marcelm

marcelm:
Thanks, but why hasn’t Travis detected this?

bgruening:
The build environment has unfortunately libgcc installed. I have a separated project for creating build artifacts independent of a build environment and here is no libgcc installed.

Feel free to merge once it is reads :)

marcelm:
Ahh, now I get to merge *your* patch, how satisfying ;-)

bgruening:
;) please! :)

marcelm:
Hm, it seems to me that it is actually matplotlib that requires libgcc.

johanneskoester:
The policy for the anaconda channel and also conda-forge changed recently. Now, every package that needs compilation with gcc should depend on the packaged gcc and libgcc. However, this will only gradually appear on the channels I guess. Maybe it would be more consistent to remove the libgcc dependency from sqt, and have matplotlib fixed in conda-forge (since we want to depend on conda-forge soon).

marcelm:
Right, that would be more consistent since libgcc is not a direct dependency of sqt. A libgcc dependency was also added to another package with PR #2278, where it also seems that matplotlib was the actual cause.

bgruening:
It's not as easy. matplotlib does not depend on gcc, but one of its dependencies and this seems to be missing libgcc. I was not able to track this down yet due to time constraints.
I would still arguing that we need to make your packages here work. Ideally we would not allow to merge a non-working sqt and other. But we can not track this yet, so I went to fix this until we depend on conda-forge and can track this down more efficiently and fix it with conda-forge.

marcelm:
Then I suggest that  a comment should be added next to a dependency if it exists only as a work-around. Otherwise, it’s likely that the work-around becomes permanent.

johanneskoester:
Indeed, and there should be a comment mentioning it is a temporal fix.

PR #2299, opened 2016-09-08
add lz4

tomkinsc:

see: https://cyan4973.github.io/lz4/

This recipe is for the C binary, not the Python, etc. bindings. 

There is an “lz4” package in the default conda channel, but it is for the Python bindings.


bgruening:
Mh, this means it has a name conflict, isn't it? Maybe we need to rename this one?

PR #2316, opened 2016-09-13
add older Trinity back again

tomkinsc:
There are algorithmic changes in Trinity v2 and later, and the recipe for this dated version should be preserved to serve users who prefer the older algorithm.

### DIFF
--- a/recipes/trinity/date.2011_11_26/meta.yaml
+++ b/recipes/trinity/date.2011_11_26/meta.yaml
@@ -1,0 +1,40 @@
+about:
+    home: 'https://github.com/trinityrnaseq/trinityrnaseq'
+    license: "https://raw.githubusercontent.com/trinityrnaseq/trinityrnaseq/master/LICENSE.txt"
+    summary: "Trinity assembles transcript sequences from Illumina RNA-Seq data"
+build:
+  number: 5
+  string: ncurses{{CONDA_NCURSES}}_{{PKG_BUILDNUM}}
+package:
+  name: trinity
+  version: 'date.2011_11_26'
+source:
+  fn: trinityrnaseq_r2011-11-26.tgz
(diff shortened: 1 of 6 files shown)


recipes/trinity/date.2011_11_26/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - bzip2
+    - zlib
+    - ncurses {{CONDA_NCURSES}}*
+extra:
+  notes: "Trinity version date.2011_11_26 relies on a custom wrapper script, symlinked to bin/: Trinity-runner.sh. This script changes the current working directory to that of the Trinity.pl perl script before calling Trinity, so relative paths work as expected (particularly for Trinity plugins). All arguments are passed to Trinity.pl (via $@)"
+test:
+    commands:
+        # This version of Trinity does not have a quick function to exit 0
+        # So at least check that we can compile the file
+        - "$PREFIX/bin/Trinity-test &> /dev/null"
```
bgruening: I have problems with theses tests for the Docker container, because $PREFIX is not defined in a normal context. Do you think a plain `Trininty-test` will work?

tomkinsc: Not quite sure why that's there at all. It should work. I'll open a follow-on PR.

tomkinsc: Related to this, is there a way we can include Docker build information in each PR in this repo? It would be nice if there were an automated way to know if we've broken something, so @bgruening doesn't have to manually write a comment. A comment bot, or maybe something optional in the test panel?

bgruening: As soon as we have the new build system I will spend a few days and try to implement the additional Docker testing in our build-system. It will take a few minutes in addition but will also find a few more errors.

PR #2325, opened 2016-09-14
New formula: cwltest

roryk:

This is necessary to upgrade to the more recent versions of cwltool.

https://github.com/common-workflow-language/cwltest

Pinging @bioconda/core for a review. Thanks ya'll!

### DIFF
--- a/recipes/cwltest/meta.yaml
+++ b/recipes/cwltest/meta.yaml
@@ -1,0 +1,37 @@
+package:
+  name: cwltest
+  version: '1.0.20160907111242'
+
+source:
+  fn: cwltest-1.0.20160907111242.tar.gz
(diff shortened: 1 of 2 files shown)


recipes/cwltest/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+requirements:
+  build:
+    - python
+    - setuptools
+    - schema-salad ==1.14.20160708181155
+    - typing
+
+  run:
+    - python
+    - setuptools
```
bgruening: Do we need setuptools as run time dependency?

johanneskoester: Sometimes packages depend on certain setuptools modules to find resources. But I don't know if it is the case here. @roryk are you aware of that?

roryk: Thanks, it doesn't need setuptools to run, afaik, so I dropped it from the requirement.

roryk:
Thanks!

PR #2334, opened 2016-09-16
Update vcfanno to 0.1.0.

roryk:

Adds sha256 sums as well. This recipe is broken, and I could use a hand figuring out why. patchelf produces a binary that core dumps:

```
/home/vagrant/local/share/bcbio/anaconda/conda-bld/vcfanno_1474044392940/test_tmp/run_test.sh: line 3:  5683 Segmentation fault      (core dumped) vcfanno
```

I put a link to the original binary that doesn't core dump on linux64 and the patched one that does:

https://dl.dropboxusercontent.com/u/2822886/vcfanno-binaries.tar

If I don't update the recipe and try to rebuild the original files they also core dump. Any ideas?

### DIFF
--- a/recipes/vcfanno/build.sh
+++ b/recipes/vcfanno/build.sh
@@ -1,5 +1,4 @@
 #!/bin/bash
-
 chmod a+x vcfanno
 mkdir -p $PREFIX/bin
 cp vcfanno $PREFIX/bin/vcfanno
--- a/recipes/vcfanno/meta.yaml
+++ b/recipes/vcfanno/meta.yaml
@@ -1,12 +1,14 @@
 package:
   name: vcfanno
-  version: '0.0.11'
+  version: '0.1.0'
 
 source:
-  fn: vcfanno_0.0.11-beta_darwin_amd64.zip # [osx]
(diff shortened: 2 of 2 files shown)


roryk:
Huh-- wonder why it is broken on my local machine.

roryk:
I think it is setting rpath incorrectly, if I set it manually `patchelf --set-rpath /home/vagrant/local/share/bcbio/anaconda/lib ./vcfanno-work` it works. It sets it to `$ORIGIN/../lib` during the build.

roryk:
I'm stumped why this doesn't build properly on any machine I try, or in the docker image, but passes the tests. @bioconda/core anyone got more of a clue than me?

bgruening:
@roryk with which image you are testing? We have migrated the build infrastructure 2 days ago and are now building in a new container with centos-6. Are you using this container?

roryk:
Thanks @bgruening: 

bioconda/bioconda-builder   latest              c17007ccc653        3 months ago        953 MB

Is there a different one I should be using?

bgruening:
We are now using https://github.com/bioconda/bioconda-utils. Unfortunately, we need to update our documentation but we plane a few more changes before we do this. Sorry for the inconvenience.
The best I can offer at the moment is the travis config :(

roryk:
Thanks!

PR #2335, opened 2016-09-16
Add perl dependencies to fastx_toolkit

mvdbeek:

runtime dependencies. This is required for fasta_clipping_histogram.pl.
Also fixes tests (all fastx_* -h commands exit with exit-code 1) by
grepping for `Part of FASTX Toolkit 0.0.14 by A. Gordon
(assafgordon@gmail.com)`.

@bgruening, can you have a look at this? In particular, I don't understand why simply saying `fasta_clipping_histogram.pl` will use the system's perl (hence this awkward hack to retrieve the path to `fasta_clipping_histogram.pl`). I would have assumed for perl to be on PATH, but that perl won't have GD.pm in @INC.

### DIFF
--- a/recipes/fastx_toolkit/meta.yaml
+++ b/recipes/fastx_toolkit/meta.yaml
@@ -8,7 +8,7 @@ source:
 
 build:
   preserve_egg_dir: True
-  number: 2
+  number: 3
   skip: False
 
 requirements:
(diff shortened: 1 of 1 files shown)


recipes/fastx_toolkit/meta.yaml
```diff
@@ -25,11 +25,16 @@ requirements:
     - nose
     - libgtextutils
     - gnuplot
+    - libgd >=2.1.1post 1
```
bgruening: can you add `perl-threaded` as well please

PR #2342, opened 2016-09-18
forget bowtie and bowtie2 in CIRCexplorer2 run section

kepbod:


bgruening:
Is it possible to inlcude a test for thise deps?

kepbod:
These depends are not essential. But it will be very convenient to include them.

kepbod:
I did not know how to include a test for them.

kepbod:
Do you have any good ideas? @bgruening

bgruening:
Just check if bowtie is in the path. Just want to make sure this does not get lost anymore :)

kepbod:
Thanks for your suggestions. When running the programs, they will check all the depends. In addition, all the essential depends will be checked during the installation. So I think it is not necessary to include some tests for the unessential depends. =)

PR #2344, opened 2016-09-19
Kinwalker

HassanAmr:


HassanAmr:
Hi @bgruening,

This doesn't pass my machine's test for the same reason bamhash didn't. So, I hope it is OK me doing a pull request to see what Travis says.

HassanAmr:
Also,
- What should be the version number? At the bottom of the man file kinwalker.1 there is 0.1 written there. While the Galaxy tool wants 1.0. Should I just set it to 1.0?
- What license should it have? Uni Leipzig?

s-will:
Hi, I am looking into this.
1) there seems to be no proper versioning for kinwalker, which explains the confusion. I guess, you got the "0.1" from the man page? At least this is the only place, I can find any kind of version number.
From the code, the version should be 0.1; no idea, why Galaxy assumes 1.0
2) for the license, my best bet would be to go with the one of the Vienna RNA package

Unfortunately Michael Geis is not actively developing this anymore and probably nobody else took responsibility.
Maybe @raumzeit has additional ideas (maybe ask the co-authors; is there a newer version?; how best name the vienna rna package license?).

HassanAmr:
Hi @s-will,

Yes, I got the version number from the man page.

RaumZeit:
Hi @HassanAmr @s-will ,
the ViennaRNA Package ships a bugfixed version of Kinwalker as well (Since it is not actively developed any more, I guess this is the latest version). However, Kinwalker is an opt-in, so one has to configure the ViennaRNA Package with --with-kinwalker.

bgruening:
@RaumZeit thanks!
@s-will can you add this to your vienna_rna builds? Than we can close this PR I think.

HassanAmr:
Ok, thanks, all :)

PR #2348, opened 2016-09-19
add cromwell

tomkinsc:

See: https://github.com/broadinstitute/cromwell

This PR adds cromwell, a WDL execution engine, as well as its dependencies: scala and sbt (scala build tool). 

The scala and sbt wrappers are rather ugly, and are ported from other conda channels. Ideally the Python-based jar wrapper would be adapted to pass the default flags to sbt (assistance on this branch is welcome). For now I just want cromwell to work.

### DIFF
--- a/recipes/cromwell/build.sh
+++ b/recipes/cromwell/build.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+set -eu -o pipefail
+
+outdir=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+mkdir -p $outdir
+mkdir -p $PREFIX/bin
+
+cd $SRC_DIR
+sbt assembly
+
+cp target/scala-*/cromwell-*.jar $outdir/cromwell.jar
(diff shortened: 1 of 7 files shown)


bgruening:
Imho scala should be in conda-forge. Its a main language that should not be in a science channel.

tomkinsc:
How do we want to handle dependencies that are not in the default conda channel, or bioconda?
Scala probably should be in conda-forge, but it is not there currently, and we certainly have more than a few dependencies that are not themselves directly science-related. 
Can we rearrange [once we depend on the conda-forge channel](https://github.com/bioconda/bioconda-recipes/blob/2c40ed8bbd95847bfe7ee7c3a98e75180f05ef06/config.yml#L10)?

bgruening:
We can and we should imho. Depending on conda-forge will happen soon I think, it just takes a brave soul and some rebuilding effort.

tomkinsc:
Merging this in until we can rely on scala and sbt from conda-forge.

PR #2352, opened 2016-09-19
inforna

HassanAmr:


HassanAmr:
Hi @bgruening,
I only noticed that there is already a branch called inforna when I was trying to push this one. So, I renamed it, and created the pull request in case you might still be interested.

The same problem as with bamhash and kinwalker happens here (Test fails on my machine because the file isn't written to _t_env, which is the test environment). I think the cp and mv commands in the build.sh (but I think only the last command though) are not replicated for _t_env. should I report this issue to conda_build?

HassanAmr:
The OSX test fails when building, so I will skip it for now.
But the linux test fails because it encounters exit code 1. 
As per the below, ```./INFO-RNA-2.1.2  -h``` returns 1 for its exit code. should I fix that
```
[hassan@hassan-pc bin]$ ./INFO-RNA-2.1.2  -h

call: ./INFO-RNA-2.1.2 "structure" [-c "constraints"] 
                   [-v "allowed mismatches (binary vector)"]
                   [-n max. number of allowed mismatches] [-r random init]
                   [-f[ACGUMRWSYKVHDBN] assignment where free bases are set to]
                   [-F[mp]] [-R [repeats]] [-S search strategy] [-m]
                   [-s length multiplier for max. number of steps during SLS] 
                   [-N neighbor choice] [-p prob. to accept worse neighbors]



General options: 
---------------------
 -c "constraints"	 Constraining sequence given in IUPAC code.

Options of the initializing part:
-----------------------------------
 -r 		 random initializing sequence
 -f[ACGUMRWSYKVHDBN]	 Give the assignment (in IUPAC code) where free bases
 			 in loop regions (that give no energy fraction) are
 			 set to.

Options of the local search part:
-----------------------------------
 -v "allowed mismatches" 	 binary vector,
 				 "1" means a mismatch conc. the constraints is allowed,
 				 "0" means a mismatch is forbidden

 -n max. mismatch number	 Maximal number of allowed mismatches in a given
				 interval. If nothing further is given, the interval 
				 includes the whole sequence.

 -F[mp]		 Use minimum energy (-Fm) (default), partition function
 		 folding (-Fp) or both (-Fmp).

 -R[repeats]	 Number of repeating the local search step.

 -S strategy	 Search strategy used during the local search step: 
                            1 - adaptive walk
                            2 - full local search
                            3 - stochastic local search (default)

 -m		 Kind of counting the step during the local search. Here, only
 		 accepted mutations are counted, while usually all tested 
 		 neighbors are counted.

 -s multiplier	 The maximal number of step during the stochastic local
 		 search is given as product of the structure length and a 
 		 multiplier. The multiplier is set to 10 by default.

 -N neighbor choice	 Kind of order the candidate neighbors during the local
			 search.
                            1 - randomly
                            2 - energy dependent (default)

 -p probability	 Probability to accept worse neighbors during the stochastic
 		 local search. It is set to 0.1 by default.

[hassan@hassan-pc bin]$ echo $?
1
```

bgruening:
Fixing sounds good https://github.com/BackofenLab/INFO-RNA
Otherwise, you can do something like `/INFO-RNA-2.1.2  -h | grep "allowed mismatches"`

recipes/inforna/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - viennarna
+
+
+  run:
+    - libgcc    # [not osx]
+    - viennarna
+
+test:
+  commands:
+    - INFO-RNA-2.1.2 "(...)"
```
bgruening: Also nice!

recipes/inforna/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+about:
+  home: http://rna.informatik.uni-freiburg.de/INFORNA/Input.jsp
```
bgruening: Please use this one.
https://github.com/BackofenLab/INFO-RNA

recipes/inforna/meta.yaml
```diff
@@ -0,0 +1,32 @@
+package:
+  name: inforna
+  version: "2.1.2"
+
+build:
+  skip: True # [osx] 
+
+source:
+  url: http://www.bioinf.uni-freiburg.de/Software/INFO-RNA/versions/INFO-RNA-2.1.2.tar.gz
```
bgruening: Can we use this one? https://github.com/BackofenLab/INFO-RNA/releases

bgruening:
This looks great, thanks!

HassanAmr:
My pleasure :)

recipes/inforna/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+make clean
+make all VIENNA=$PREFIX
+
+mv INFO-RNA-2.1.2 $PREFIX/bin
```
yhoogstrate: Is it worth adding a prefix not containing the version number?:
`ln -s INFO-RNA-2.1.2 "$PREFIX/bin/INFO-RNA"`

yhoogstrate: Including the dash depends on how the authors named it. Having both in and using a symlink is fine to me too. This is also in concordance with linux conventions AFAIK

HassanAmr: OK, thanks :)

bgruening:
Thanks!

HassanAmr:
My pleasure :)

PR #2355, opened 2016-09-20
Update viennarna (2.2.10); enable kinwalker

s-will:

### DIFF
--- a/recipes/viennarna/2.2.9/build.sh
+++ b/recipes/viennarna/2.2.9/build.sh
@@ -14,8 +14,7 @@ else ## linux
 fi
 
 ./configure --prefix=$PREFIX \
-            --without-perl --without-python \
-            --with-kinwalker \
+            --without-perl --without-python --disable-c11 \
             ${extra_config_options} && \
 make -j${CPU_COUNT} && \
 make install
--- a/recipes/viennarna/2.2.9/meta.yaml
+++ b/recipes/viennarna/2.2.9/meta.yaml
@@ -1,0 +1,29 @@
+package:
+  name: viennarna
+  version: 2.2.9
+
+build:
+  number: 0
+
+source:
+  fn: ViennaRNA-2.2.9.tar.gz
+  url: http://www.tbi.univie.ac.at/RNA/packages/source/ViennaRNA-2.2.9.tar.gz
+  sha256: cf3806fdad51405ac537e3636f4fa6f56d6b3297841ab4736a84ddd8841d5108
+
+requirements:
+  build:
+    - gcc   # [linux]
+    - llvm  # [osx]
+  run:
+    - libgcc # [linux]
+
+test:
+  commands:
+   - RNAfold --version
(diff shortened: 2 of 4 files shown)


s-will:
Hi @bgruening, could you please help? This seems to fail for reasons inside the conda build system. (See the several errors in the OSX build output)

s-will:
Interestingly, it newly complains about missing perl. Let's see:)

bgruening:
Perl is also a run dependency? Thanks @s-will for looking into it.

s-will:
I can't imagine that perl is required to run all the main tools. There are some Utils written in perl, which are written to some extra bin dir: share/ViennaRNA/bin/, which is usually not even in the path. 
Do you think this justifies to include perl for run-time?

The whole wining about perl dependencies is new to me (and maybe 2.2.10) anyway.

bgruening:
I don't know how often the utils are used. No opinion here sorry.

bgruening:
Wow, this is a nice one: https://travis-ci.org/bioconda/bioconda-recipes/jobs/161312111#L1670
What is `chmod 644 "/anaconda/conda-bld/vienna_rna/share/doc/ViennaRNA/html"/*` doing?

s-will:
I would say it makes sense to omit the perl run-time dependency to keep this lean. (Maybe Ronny @raumzeit can get rid of the compile time dependency if possible? Why is this even there?).

The chmod is in the documenation Makefile.am of the package. Then, this is rather a bug in the Vienna package; I can try to patch this (but Ronny, please fix in 2.2.11!). However, I wonder, why this happens only in the OSX build -- must be other length limits

RaumZeit:
Thanks for pointing out that the number of files in the HTML documentation might exceed the maximum number of arguments. This will be fixed in the next release of ViennaRNA.
@s-will Your patch for Makefile.in won't do what you expect. A ```chmod -R 644 dir/``` will certainly fail since the execution bit for ```dir/``` will be removed first. Consequently, the files within ```dir/``` won't be accessible anymore for processing by ```chmod```. The only portable way to do this is to loop over the content of ```dir```, as in your first patch:
```
{ for i in "$(DESTDIR)$(htmldir)"/*; do test -f "$$i" && chmod 644 "$$i"; done }
chmod 755 "$(DESTDIR)$(htmldir)"
{ for i in "$(DESTDIR)$(htmldir)"/search/*; do test -f "$$i" && chmod 644 "$$i"; done }
chmod 755 "$(DESTDIR)$(htmldir)"/search
```
And btw, @bgruening , The chmod'ing in the install rule of ViennaRNAs documentation is a crude fix for wrong file attributes generated by some doxygen versions.

s-will:
Thanks @raumzeit; I realized too late that there is a general problem with this. Now, I switched to the loop again.

bgruening:
@RaumZeit thanks for the explanation, this explains a lot.
Seems to work, thanks again @s-will!

PR #2361, opened 2016-09-20
Update: arvados-cwl-runner arvados-python-client

chapmanb:

Updates extracted from previous attempts due to issues with resolving
openssl, to hopefully demonstrate problem, fix and get merged.

### DIFF
--- a/recipes/arvados-cwl-runner/meta.yaml
+++ b/recipes/arvados-cwl-runner/meta.yaml
@@ -1,12 +1,11 @@
 package:
   name: arvados-cwl-runner
-
-  version: '1.0.20160715171107'
+  version: '1.0.20161031135838'
 
 source:
-  fn: arvados-cwl-runner-1.0.20160715171107.tar.gz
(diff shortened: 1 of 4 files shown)


chapmanb:
@bioconda/core -- I'm running into an issue with openssl that I'm not sure how to work around and could use some help. I'm seeing this failure:
```
ERROR:bioconda_utils.utils:Error: PREFIX too long: 255 characters, but only 128 allowed
Error: post-link failed for: anaconda::openssl-1.0.2d-0
```
which I can't seem to replicate locally using bioconda-build. Locally I get openssl-1.0.2h and everything works okay but in Travis I get an older version and this failure.

I see some discussion around this with @brentp here (https://github.com/conda/conda/issues/1087#issuecomment-207559278) but it looks like that has been fixed/resolved, although I don't understand all the issue.

It also looks like there might a fix if we can update the build version we use:

https://github.com/conda/conda-build/pull/1312

Thanks for any pointers or help on this.

brentp:
we have been hitting this the last couple of weeks with openssl as well. I think it's happening as conda is changing from 128 to 255 characters for the prefix and some builds have it and some do not. I don't have a good solution though.

bgruening:
@chapmanb is this still an issue?

chapmanb:
Björn -- thanks so much for the push to re-check this. I bumped the versions and retried and it looks like this works with the current version of conda-build. Awesome. Thanks again.

PR #2367, opened 2016-09-21
New stacks 1.42 version

abretaud:

### DIFF
--- a/recipes/stacks/1.40/build.sh
+++ b/recipes/stacks/1.40/build.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+export CPATH=${PREFIX}/include
+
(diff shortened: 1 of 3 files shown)


yhoogstrate:
@abretaud Are you sure this test works?

```
$ doesnotwork 2>&1 | grep doesnotwork
zsh: command not found: doesnotwork
$ echo $?
0
```
I think this also gives a True if nothing has been installed?

abretaud:
Oh yes you're right!
I just kept the tests without seeing this potential problem, I'm going to fix this, thank you!

yhoogstrate:
```
    - command -v cstacks
    - command -v pstacks
    - command -v process_radtags
    - command -v clone_filter
```
would already do it right?

PR #2370, opened 2016-09-21
Stacks: better tests

abretaud:

Following the discussion in #2367

### DIFF
--- a/recipes/stacks/meta.yaml
+++ b/recipes/stacks/meta.yaml
@@ -2,7 +2,7 @@ package:
   name: stacks
   version: "1.42"
 build:
-  number: 1
+  number: 2
   skip: True # [osx]
 source:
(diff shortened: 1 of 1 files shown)


abretaud:
@yhoogstrate before merging, does it look good to you like this?

yhoogstrate:
If the light turns green, yes! `command -v` exits with `1` if the tool can not be found or not be invoked (http://pubs.opengroup.org/onlinepubs/009696899/utilities/command.html).

abretaud:
Green! Let's go
@bgruening this 1.42 version will magically appear [there](https://quay.io/repository/mulled/stacks?tab=tags), right? is there a delay for this?

bgruening:
It will be compiled and tested again in travis after merge, if this succeeds it will be pushed. This is the delay.

PR #2372, opened 2016-09-21
WIP: add mummer as dependency to quast

bgruening:

### DIFF
--- a/recipes/quast/build.sh
+++ b/recipes/quast/build.sh
@@ -1,5 +1,21 @@
 #!/bin/bash
 
+export INCLUDE_PATH="${PREFIX}/include"
+export LIBRARY_PATH="${PREFIX}/lib"
+export LD_LIBRARY_PATH="${PREFIX}/lib"
+export BOOST_INCLUDE_DIR=${PREFIX}/include
+export BOOST_LIBRARY_DIR=${PREFIX}/lib
+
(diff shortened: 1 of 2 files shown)


natefoo:
Quast depends on [E-MEM](http://www.csd.uwo.ca/~ilie/E-MEM/) and MUMmer if using the reference (`-R`) option. It will not use external versions of these, however. It attempts to build them at runtime in `opt/quast-4.1/libs/{E-MEM-linux,MUMmer3.223-linux}`. It may be possible to use the bioconda MUMmer with some linking tricks.

lparsons:
I'm a bit in favor of building the packaged E-MEM since that is what quast is expecting and it's what non-bioconda users will be doing. Provides a more "standard" install, IMHO, even if it's not as "clean" from the dependency perspective.

recipes/quast/build.sh
```diff
@@ -8,6 +8,8 @@ mkdir -p $QUAST_HOME
 
 cp -R $SRC_DIR/* $QUAST_HOME
 
+make -C $PREFIX/opt/quast-4.1/libs/E-MEM-linux
```
natefoo: Probably also need `make -C $PREFIX/opt/quast-4.1/libs/MUMmer3.23-linux` here

bgruening:
@natefoo can you try the latest version?

PR #2376, opened 2016-09-22
updating stringtie to 1.3

jerowe:

Updating stringtie to 1.3

### DIFF
--- a/recipes/stringtie/1.2.4/meta.yaml
+++ b/recipes/stringtie/1.2.4/meta.yaml
@@ -2,15 +2,12 @@ about:
   home: http://ccb.jhu.edu/software/stringtie/
   license: Artistic License 2.0
   summary: Transcriptome assembly and quantification for RNA-seq
-
 build:
-  number: 1
+  number: 0
   skip: True # [osx]
-
 package:
   name: stringtie
   version: 1.2.4
-
 requirements:
   build:
     - gcc   # [linux]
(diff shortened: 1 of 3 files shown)


recipes/stringtie/meta.yaml
```diff
@@ -2,12 +2,15 @@ about:
...
   summary: Transcriptome assembly and quantification for RNA-seq
+
 build:
-  number: 0
+  number: 1
   skip: True # [osx]
+
 package:
   name: stringtie
   version: 1.2.4
```
yhoogstrate: 1.3.0 ?

recipes/stringtie/1.2.4/meta.yaml
```diff
@@ -0,0 +1,25 @@
...
+build:
+  number: 0
+  skip: True # [osx]
+package:
+  name: stringtie
+  version: 1.2.4
+requirements:
+  build:
+    - gcc   # [linux]
+    - llvm  # [osx]
```
yhoogstrate: Does this work when you have `  skip: True # [osx]` (line7) ?

jerowe: Well, it builds.

Let me see if I can build it with OSX. I copied over the meta file from 1.2.4, but lets see.

bgruening: As a reference, this # [osx] is a template language, it will be stripped out if you are not on OSX. So it has no influence in the end. It is just not a good "coding" style. On the other hand if we leave it in the guy who cares about OSX builds later on has less work :)

jerowe: It builds on OSX!

jerowe: Thanks for the reference, @bgruening!

bgruening: Than you should remove "skip: True # [osx]" :)

jerowe: Removed! 

Merge?

bgruening:
:+1: Thanks!

PR #2379, opened 2016-09-23
enable conda-forge and R3.3

bgruening:

This PR enables conda-forge and R3.3.
As far as I can tell conda-forge should not be a big problem at least the Docker container building all with conda-forge enables since two weeks.

ping @johanneskoester, @daler


jerowe:
@bgruening , could you go into what this means for existing R packages? Should they be rebuilt for R-3.3?

bgruening:
@jerowe yes exactly. I have now all packages on the blacklist. Removing than one by one will rebuild them for R3.3. Help is really needed as soon as this one gets merged.

(not realy one by one, but in chunks)

bgruening:
@bioconda/core any comments on this one? I think this is a big step for us so I would like to get someone else to merge this :)

daler:
@bioconda/all: please note that as of this merge, we are now using the conda-forge channel and using R 3.3 for building R packages. This is long overdue, but is a big change so there is the potential of unforeseen difficulties . . . please bear with us and report issues as you find them.

What this means to you:
 - over the next days we will be upgrading all R packages to R 3.3. If you plan on adding a new R recipe, you might want to wait until this upgrading has been completed.
 - if you add a package that doesn't exist in bioconda, but does exist in conda-forge, it will appear that your build is being skipped by bioconda's travis-ci. In that case, you don't have to do any merges with bioconda and instead you can use the conda-forge package directly.

bgruening:
Makes my day!!!!

tomkinsc:
Probably time to update the docs to mention conda-forge:
https://github.com/bioconda/bioconda-recipes/blame/master/docs/index.rst#L54-L55

daler:
Thanks @tomkinsc, added to #2401

kyleabeauchamp:
Great work!  

Looking at our Dockerfile (https://github.com/bioconda/bioconda-recipes/blob/master/scripts/Dockerfile), it looks like we're still based on Centos5.  Are we still planning to upgrade that to Centos6 later, in order to be consistent with Conda-Forge?

daler:
We should probably delete that Dockerfile to avoid confusion -- in the [`travis-run.sh`](https://github.com/bioconda/bioconda-recipes/blob/master/scripts/travis-run.sh#L8) script we're already using CentOS6 via the `condaforge/linux-anvil` image.

bgruening:
:+1: for deleting the Dockerfile. We need to overhaul our documentation but I think we can wait with this until we have merged your bioconda-utils work.

kyleabeauchamp:
OK it also says Centos5 in the docs as well, so it was doubly confusing :)

jerowe:
Have all bioconda packages been rebuilt on centos6? Or only newer packages?

PR #2389, opened 2016-09-24
add nucleosome prediction tool

bgruening:


bgruening:
@johanneskoester @daler any idea about: https://travis-ci.org/bioconda/bioconda-recipes/jobs/162440037#L492

Unfortunately, I could not find the source for the included binary so I need to ship it. But the error looks strange, I though that a binary compiled on an old system will always run on a new one? 
Including libgcc as run time dependency will include libstdc++.so.6 - not libstdc++.so.5. Symlinking libstdc++.so.6 to libstdc++.so.5 also does not work. Any idea how to fix this? Where do I get libstdc++.so.5?

kyleabeauchamp:
Have you been able to run the binary manually on your machine or manually in the container?

bgruening:
On my machine yes, but only after install `libstdc++5`.

johanneskoester:

PR #2390, opened 2016-09-24
Bump IgBLAST to 1.5.0

marcelm:

Keep also the old 1.4.0 version since the output format has changed
slightly, and programs out there that parse it may need to be updated.

### DIFF
--- a/recipes/igblast/meta.yaml
+++ b/recipes/igblast/meta.yaml
@@ -1,20 +1,31 @@
 package:
   name: igblast
-  version: "1.4.0"
-source:
-  fn: ncbi-igblast-1.4.0-x64-linux.tar.gz  # [not osx]
-  fn: ncbi-igblast-1.4.0-universal-macosx.tar.gz  # [osx]
-  url: ftp://ftp.ncbi.nih.gov/blast/executables/igblast/release/1.4.0/ncbi-igblast-1.4.0-x64-linux.tar.gz  # [not osx]
-  url: ftp://ftp.ncbi.nih.gov/blast/executables/igblast/release/1.4.0/ncbi-igblast-1.4.0-universal-macosx.tar.gz  # [osx]
+  version: "1.5.0"
+
 about:
   home: http://www.ncbi.nlm.nih.gov/projects/igblast/
   license: Public Domain
   summary: A tool for analyzing immunoglobulin (IG) and T cell receptor (TR) sequences.
+
+source:
+  fn: ncbi-igblast-1.5.0-x64-linux.tar.gz  # [not osx]
+  fn: ncbi-igblast-1.5.0-universal-macosx.tar.gz  # [osx]
+  md5: 984de09c68701cc75dd49c8935aa78a1  # [not osx]
+  md5: 22cdb563a11cc41a753a5b4d7043e2d7  # [osx]
+  url: ftp://ftp.ncbi.nih.gov/blast/executables/igblast/release/1.5.0/ncbi-igblast-1.5.0-x64-linux.tar.gz  # [not osx]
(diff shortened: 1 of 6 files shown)


marcelm:
How strange! I wrote the recipe already a week ago or so, and now 1.6.0 has disappeared from the FTP server. I’ll investigate.

marcelm:
?

bgruening:
Upps, sorry!!!!

marcelm:
No worries! I haven’t received a reply from the author, yet, so I think I’ll turn this into a bump to version 1.5.0 instead, but I’ll wait one or two more days.

marcelm:
Jian Ye, who is IgBLAST’s author, replied and said that 1.6.0 had an error and will be restored soon. Version 1.5.0 already contains the features I am interested in, so I changed this PR to is a bump to version 1.5.0.

marcelm:
@bioconda/core Could anyone have a look? The recipe is failing with the message [`wget: command not found`](https://travis-ci.org/bioconda/bioconda-recipes/jobs/163338709#L491). I’ve added `wget` to the build requirements, but it still does not work. Many other recipes I checked that use `wget` in the build script don’t even have it in their dependencies.

bgruening:
Please use curl, this should work.

marcelm:
Ok, I will do so. It’s still a regression, though.

marcelm:
Hm, I actually need wget because I cannot make curl do `wget -r -nH --cut-dirs=5 ...` (recursive download).

johanneskoester:
You should be able to set wget as a build dependency, because it is in the anaconda/default channel. You are right, it is a regression, likely caused by our recent change to use the same CentOS 6 docker image as conda-forge. But it is actually a good regression, because it forces us to put a tool that is not guaranteed to be present into the build dependencies.

bgruening:
@johanneskoester the `wget` package is just the library not the binary.

marcelm:
Yes, `wget` on anaconda is the [Python wget module](https://pypi.python.org/pypi/wget).

johanneskoester:
Oh, that's a pity. So the logical next step would be to add wget to the conda-forge channel, right? Or temporarily to bioconda, and move it over at some point.

marcelm:
Finally! This has taken a while to figure out ...

bgruening:
@marcelm for some reason your are missing zlib as runtime dependency: `/usr/local/bin/../share/igblast/bin/igblastn: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
`

recipes/igblast/meta.yaml
```diff
@@ -1,20 +1,31 @@
...
+  fn: ncbi-igblast-1.5.0-x64-linux.tar.gz  # [linux]
+  fn: ncbi-igblast-1.5.0-universal-macosx.tar.gz  # [osx]
+  md5: 984de09c68701cc75dd49c8935aa78a1  # [linux]
+  md5: 22cdb563a11cc41a753a5b4d7043e2d7  # [osx]
+  url: ftp://ftp.ncbi.nih.gov/blast/executables/igblast/release/1.5.0/ncbi-igblast-1.5.0-x64-linux.tar.gz  # [linux]
+  url: ftp://ftp.ncbi.nih.gov/blast/executables/igblast/release/1.5.0/ncbi-igblast-1.5.0-universal-macosx.tar.gz  # [osx]
+
 build:
-  number: 6
+  number: 1
```
bgruening: please start next time with number: 0 if you have a new version

marcelm: Sorry, my bad.

marcelm:
Hm, the test passed, so I assumed everything was fine. But it seems that Travis did not really build the package. I’ll open a new PR to fix this.

PR #2384, opened 2016-09-23
R migration part I

bgruening:
* [x] I have read the guidelines above.
* [ ] This PR adds a new recipe.
* [x] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

PR #2408, opened 2016-09-26
Part1.5

daler:

Cleans up the R blacklist, new list in `sorted-r-blacklist` and is used by the config.

- de-duped r-blacklist
- packages with older-version subdirs are listed explicitly
- list is toposorted. If we start commenting out packages at the top and work our way down, all dependencies will built in the correct order
- commented-out packages from #2384 are also commented out here

`sorted-r-blacklist` created with:

```python
import networkx as nx
from bioconda_utils.utils import get_dag, get_recipes

d, n2r = get_dag(get_recipes('recipes'), restrict=False)
dependent_on_r = []
for dep in ['r', 'r-base']:
    for pkg in d['r']:
        for recipe in n2r[pkg]:
            dependent_on_r.append(recipe)

dependent_on_r = [recipe for pkg in d['r'] for recipe in n2r[pkg]]

# add anything else that's in the current blacklist, just to be sure.
commented_packages = []
with open('r-blacklist') as fin:
    for line in fin:
        line = line.strip()
        if len(line) == 0:
            continue
        pkg = line.replace('#', '')
        pkg = line.split('/')[1]
        dependent_on_r.extend(n2r[pkg])
        if line.startswith('#'):
            commented_packages.extend(n2r[pkg])

sorted_r_dependent = [recipe for pkg in nx.topological_sort(d) for recipe in n2r[pkg] if recipe in dependent_on_r]
with open('sorted-r-blacklist', 'w') as fout:
    for i in sorted_r_dependent:
        if i in commented_packages:
            i = '#' + i
        fout.write(i + '\n')
```


bgruening:
Thanks!

PR #2413, opened 2016-09-27
R 3.3 migration, part 4

daler:

Had a little trouble with `r-ash` (needs `lsb_release`?). Just to keep the ball rolling I kept it uncommented in `sorted-r-blacklist` and put some comments in that file next to other packages that depend on it.

Also deleted `r-blacklist` to avoid confusion.

ping @bgruening


sorted-r-blacklist
```diff
@@ -20,15 +20,19 @@
...
-recipes/r-calibrate
+#recipes/r-maldiquant
+#recipes/r-fda
+#recipes/r-compute.es
+#recipes/r-snowfall
+#recipes/r-permute
+#recipes/r-funr
+#recipes/r-calibrate
+
+# r-ash needs gfortran; even after adding gcc it complains about needing
```
bgruening: I saw this once, but was this a error or a warning? If an error we need to patch the source I guess :(

daler: It didn't build because of that error.

PR #2431, opened 2016-09-27
bump samtools build after switch to conda-forge

tomkinsc:

I seem to be having trouble with the OSX build of samtools (`dyld: Symbol not found: _stdscr`) after depending on conda-forge. Triggering Travis to rebuild to see if it is just me...

### DIFF
--- a/recipes/htslib/meta.yaml
+++ b/recipes/htslib/meta.yaml
@@ -2,7 +2,7 @@ package:
   name: htslib
   version: "1.3.1"
 build:
-  number: 0
+  number: 1
   skip: False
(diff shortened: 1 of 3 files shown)


kyleabeauchamp:
Any resaon not to do pysam and htslib at the same time?  They all have a lot in common.

kyleabeauchamp:
I guess particularly pysam, which builds a vendored samtools

tomkinsc:
Great point. Pysam and htslib are now also represented in this PR.

bgruening:
Thanks!!

PR #2432, opened 2016-09-27
Add ProteoWizard Conda recipe

chambm:
* [X ] This PR adds a new recipe.

@bioconda/core

### DIFF
--- a/recipes/proteowizard/meta.yaml
+++ b/recipes/proteowizard/meta.yaml
@@ -1,0 +1,29 @@
+{% set version = "3_0_9992" %}
+
+package:
+  name: proteowizard
+  version: {{ version }}
+
+source:
+  url: http://teamcity.labkey.org/guestAuth/repository/download/bt81/365914:id/pwiz-src-{{ version }}.tar.bz2
+  sha1: 4bf02eb83755b3df822a2d35b39d577be84f8282
+
+build:
+  script: bash -c "./quickbuild.sh -j{{ CPU_COUNT }} address-model=64 executables && cp build-linux-x86_64/gcc-release-x86_64/* $PREFIX/bin"
+  skip: True # [osx]
+
+requirements:
+  build:
+    - gcc   # [linux]
+  run:
+    - libgcc # [linux]
+
+
+about:
+  home: https://proteowizard.sourceforge.net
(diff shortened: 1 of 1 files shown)


bgruening:
Fantastic Matt!

bernt-matthias:
Hi @chambm and @bgruening, do you think its possible to add the libraries to this recipe, they are needed here: https://github.com/bioconda/bioconda-recipes/pull/19219

Also wondering if its possible/advisable to make this recipe not use the shipped boost and zlib but make it depend on other recipes.

chambm:
> Hi @chambm and @bgruening, do you think its possible to add the libraries to this recipe, they are needed here: #19219
I think the better way to go now is to use the Windows build via Docker/Singularity. I don't see a lot of point supporting Linux/OSX builds of pwiz when they're missing the main benefit of using pwiz - the vendor readers. :) I originally created this recipe for putting pwiz and pwiz-derived tools (like [Bumbershoot](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bumbershoot/meta.yaml)) in Galaxy. But since making the official Docker container I'm not inclined to maintain these anymore.

Thus I recommend making a MaRaCluster image that derives from the pwiz Docker image and just add its executables to it. Although you'd have to be careful to build MaraCluster with approximately the same version of pwiz as is in the pwiz container (same vendor DLL versions). Are there some contexts I'm not thinking about where Conda is preferable (or necessary) compared to Docker/Singularity?

> Also wondering if its possible/advisable to make this recipe not use the shipped boost and zlib but make it depend on other recipes.
Probably possible but not something I'm spending time on (because of the reasoning above).

glormph:
I can think of a usecase where I'd prefer to use my own container rather than the pwiz docker. (Happens a lot in my nextflow pipelines, which I try to deploy with only one container full of conda recipes.) I started to try something (#19635)  but even just incrementing build nr doesnt work at the moment.

bernt-matthias:
Thanks for the comment @glormph . 

In my opinion the much better solution would be to convert the vendor formats at the first step of any pipeline to a proper format using pwiz. 

Then processing, e.g. with MaRaCluster, can be done even without vendor format support. But for the compilation of MaRaCluster the libraries are needed.

PR #2436, opened 2016-09-27
Updated recipe for poretools from 0.5.1 to 0.6.0

malloryfreeberg:


recipes/poretools/meta.yaml
```diff
@@ -9,16 +9,19 @@ requirements:
         - python
         - watchdog >=0.8.3
         - rpy2 >=2.4.2
-        - h5py >=2.0
+        - h5py >=2.2.0
 
     run:
         - python
         - hdf5 >=1.8.7
         - r >=3.0.0
```
druvus: r and rpy are not a dependencies for version 0.6.0

recipes/poretools/meta.yaml
```diff
@@ -1,24 +1,24 @@
 package:
     name: poretools
-    version: "0.5.1"
+    version: "0.6.0"
 source:
     git_url: https://github.com/arq5x/poretools
```
druvus: Same here:
<pre>
source:
  fn: poretools_v0.6.0.tar.gz
  url: https://github.com/arq5x/poretools/archive/v0.6.0.tar.gz
  md5: 7fe9dd17b6fdefb976039c00b95ac18a
</pre>

recipes/poretools/meta.yaml
```diff
@@ -1,24 +1,25 @@
...
-        - h5py >=2.0
-        - watchdog >=0.8.3
+        - h5py >=2.2.0
         - qcli >=0.1.1
+        - matplotlib
+        - seaborn
+        - pandas
 
 build:
     skip: True # [not py27 or osx]
```
druvus: It is building for me on osx with the following sections: 
<pre>
requirements:
    build:
        - python
        - setuptools
        - hdf5 >=1.8.7
        - h5py >=2.2.0
        - matplotlib
        - seaborn
        - pandas

    run:
        - python
        - hdf5 >=1.8.7
        - h5py >=2.2.0
        - matplotlib
        - seaborn
        - pandas

build:
    skip: True # [not py27]
</pre>

druvus: License info is also missing. It is GPL according to setup.py

malloryfreeberg:
Thank you for the suggestions and feedback. This is my first time submitting to bioconda, so I'm not 100% sure of the protocols yet :]

druvus:
Thanks for updating it! It has been on my todo list for while to update poretools. It is good to have some more nanopore people contributing. 

I started to look through the v0.5.1 recipe and source code more carefully and I think the dependencies are not correct. I think the dependencies are valid for v0.5.0 and everything related to R was phased out in v0.5.1 when they swapped to seaborn instead of ggplot2 in R. I could not find references to rpy2 in the python code. 

I have successfully used the following in the recipe for v0.5.1 (i.e. no R dependencies). 
<pre>
requirements:
    build:
        - python
        - setuptools
        - hdf5 >=1.8.7
        - h5py >=2.0
        - matplotlib
        - seaborn
        - pandas

    run:
        - python
        - hdf5 >=1.8.7
        - h5py >=2.0
        - matplotlib
        - seaborn
        - pandas
</pre>

You could also build it for osx (i.e. remove osx from the skip line for both recipes)

malloryfreeberg:
I had initially installed poretools v0.5.1 using conda on a cluster running CentOS 6.7. It worked fine. Should I leave the v0.5.1 recipe alone since it does work as-is, or update it with the correct set of dependencies you mentioned?

druvus:
I tested the current Bioconda poretools package in a clean environment and you are right it is working. 

It made look into the code differences between poretools versions using squiggle.py that is used for plotting as a test case.
 
__squiggle.py (v0.5.0)__
<pre>
import os
import sys
import rpy2.robjects as robjects
import rpy2.robjects.lib.ggplot2 as ggplot2
from rpy2.robjects.packages import importr
</pre>

__squiggle.py (v0.5.1)__
<pre>
import os
import sys
import pandas as pd
import seaborn as sns
from matplotlib import rcParams
rcParams.update({'figure.autolayout': True})
from matplotlib import pyplot as plt
</pre>

The code in the bioconda poretools package is identical to v0.5.0 and not to v0.5.1. It seems wrong version got included, maybe because the recipe was lacking git_rev information. 

I would recommend to update the requirements as the new url will download the correct v0.5.1 version.  It would also remove the R dependency so poretools could be removed from the temporary R blacklist. 

Any opinion form @bioconda/core people?

malloryfreeberg:
Thanks for investigating this. I'll update v0.5.1 to have the correct url/dependencies/etc. Is it worth it to also add a directory for v0.5.0. with its correct url/dependencies/etc?

bgruening:
@malloryfreeberg we do currently not enforce people to keep older versions as subdir, but more and more people doing this as it is easier to fix, update old recipes. I do appreciate keeping older versions as separate directory.

malloryfreeberg:
Thanks @bgruening. I will do this, for sure!

bgruening:
Thanks  @malloryfreeberg!

PR #2456, opened 2016-09-29
R part5

daler:

another round of R updates.

Note that we will have to pay attention to R version specifiers (see bioconductor-dupradar here)


daler:
@bgruening @johanneskoester  `r-phonr` is acting strangely -- conda-build keeps trying to build it and restarts. I think this is what was causing the timeout. I'm putting it back on the blacklist for now; nothing else appears to depend on it.

Not sure it belongs in bioconda anyway -- its description is "Tools for phoneticians and phonologists, including functions for normalization and plotting of vowels".

bgruening:
Its literature mining so can be counted as bioinformatics :-). Can we just skip the osx build?

daler:
Ah, true.

Locally it was doing the same thing for linux; if this was in fact the problem I'm not sure why osx but not linux timed out. Let's see if that package was the problem.

daler:
@bgruening let's keep moving . . . we can return to `r-ash`, `r-phonr`, and whatever else we find after getting the bulk of the work done.

bgruening:
Great, thanks! Yes lets keep moving!

PR #2457, opened 2016-09-29
GNU Wget 1.18

marcelm:

This recipe is based on the public domain recipe at
<https://github.com/conda/conda-recipes/tree/master/wget> authored by
Trent Nelson.

@bioconda/core This is a new recipe, I’m asking for review. This recipe should be moved over to conda-forge, but I don’t have time for that myself right now. We need a wget package somehow since wget is no longer part of the build environment and some bioconda recipes use it in their build scripts. See also PR #2390.

Note that there is also a package an PyPI named wget, which is *not* GNU Wget. If you `conda install wget`, you currently get that one.


recipes/wget/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+source:
+  fn: wget-1.18.tar.xz
+  url: http://ftp.gnu.org/gnu/wget/wget-1.18.tar.xz
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - python >=3
```
johanneskoester: Python as a build dependency? The source does look like pure C to me... Are you sure about this?

recipes/wget/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - zlib
+    - openssl
+  run:
+    - python >=3
+    - zlib
+    - openssl
+
+test:
+  commands:
+    - wget --help >& /dev/null
```
johanneskoester: Above line should be `wget --help &> /dev/null`

recipes/wget/meta.yaml
```diff
@@ -0,0 +1,29 @@
+package:
+  name: wget
```
johanneskoester: Should we rename to `gnu-wget`? Otherwise, we will have a name clash with the `wget` python library. Since the anaconda channel is top priority, we cannot shadow that package.

marcelm: Hm, as GNU Wget is about 15 years older and a few orders of magnitude more popular, it doesn’t feel right that this package should get the prefix and not the other one. But I agree it’s the easier way. I’ve renamed it.

recipes/wget/meta.yaml
marcelm: The configure script looks for Python, but you are right, on closer inspection it turns out Python is only needed for the test suite (which the recipe does not run).

PR #2464, opened 2016-09-30
Update FastTree(MP)

keuv-grvl:

My first goal was to make FastTreeMP (compiled with OpenMP) available through Bioconda.
I opted for updating the existing FastTree recipe to produce FastTree and FastTreeMP instead of a separate "FastTreeMP" recipe.

In the same time, I upgraded to from v2.1.8 to v2.1.9.

HTH

### DIFF
--- a/recipes/fasttree/2.1.8/build.sh
+++ b/recipes/fasttree/2.1.8/build.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+gcc -O3 -DUSE_DOUBLE -finline-functions -funroll-loops -Wall -o FastTree FastTree.c -lm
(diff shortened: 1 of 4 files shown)


bgruening:
Looks great! Thanks for using templates :)

bgruening:
@keuv-grvl thanks for your contribution! A Docker container was also created for you `docker pull quay.io/mulled/fasttree:2.1.9--0`

keuv-grvl:
I am happy to help :-)
Is the container creation automatic? Otherwise, do you have guidelines to do it?

bgruening:
Yes, every 24h these are created via travis. Here is the repo https://github.com/mulled/auto-mulled

We will try to get this setup as part of bioconda to make it more transparent, but we have a lot to do at the infrastructure side of bioconda. Help is appreciated :)

Issue #2166, opened 2016-08-18
missing libgomp.so.1


osallou:
Hi,
I tried to install salmon in a container with conda on a Debian.

```
conda install -y salmon
```

However it fails with missing library:  libgomp.so.1

```
osallou@tifenn/tmp $ sudo docker run --rm -it osallou/bioconda /bin/bash
root@466f69e85bc3:/# conda install -y salmon
Fetching package metadata: ........
Solving package specifications: .........

Package plan for installation in environment /root/miniconda3:

The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    icu-54.1                   |                0        11.3 MB
    sqlite-3.13.0              |                0         4.0 MB
    tbb-4.4_20150728           |                0         1.2 MB
    xz-5.2.2                   |                0         644 KB
    python-3.5.2               |                0        17.2 MB
    boost-1.60.0               |           py35_0        10.2 MB
    conda-env-2.5.1            |           py35_0          27 KB
    ruamel_yaml-0.11.14        |           py35_0         377 KB
    conda-4.1.11               |           py35_0         204 KB
    salmon-0.6.0               |      boost1.60_2         4.3 MB
    ------------------------------------------------------------
                                           Total:        49.3 MB

The following NEW packages will be INSTALLED:

    boost:       1.60.0-py35_0    
    icu:         54.1-0           
    ruamel_yaml: 0.11.14-py35_0   
    salmon:      0.6.0-boost1.60_2
    tbb:         4.4_20150728-0   

The following packages will be UPDATED:

    conda:       4.0.5-py35_0 --> 4.1.11-py35_0    
    conda-env:   2.4.5-py35_0 --> 2.5.1-py35_0     
    python:      3.5.1-0      --> 3.5.2-0          
    sqlite:      3.9.2-0      --> 3.13.0-0         
    xz:          5.0.5-1      --> 5.2.2-0          

Fetching packages ...
icu-54.1-0.tar 100% |################################| Time: 0:00:05   2.02 MB/s
...
salmon-0.6.0-b 100% |################################| Time: 0:00:01   3.35 MB/s
Extracting packages ...
[      COMPLETE      ]|###################################################| 100%
Unlinking packages ...
[      COMPLETE      ]|###################################################| 100%
Linking packages ...
[      COMPLETE      ]|###################################################| 100%
root@466f69e85bc3:/# salmon
[...]

PR #2472, opened 2016-10-01
add gcc and llvm to salmon

bgruening:

### DIFF
--- a/recipes/salmon/build.sh
+++ b/recipes/salmon/build.sh
@@ -16,5 +16,6 @@ elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
     sed -i 's/CFLAGS+=${STADEN_LIB}/CFLAGS+=${STADEN_LIB} CFLAGS+=-lz/' CMakeLists.txt
     cd build
     cmake -DCMAKE_INSTALL_PREFIX=${PREFIX} -DBOOST_ROOT=$PREFIX -DBoost_NO_SYSTEM_PATHS=ON -DBoost_DEBUG=ON ..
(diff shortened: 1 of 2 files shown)


bgruening:
```
/opt/conda/conda-bld/salmon_1475432655526/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/bin/autoconf: /opt/conda/conda-bld/salmon_1475432655526/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/bin/autom4te: /usr/bin/perl: bad interpreter: No such file or directory
```

@roryk @chapmanb any idea how we can fix this?

roryk:
Hi Björn,

Hm-- is the perl dependency necessary?

bgruening:
The error is complaining about `/usr/bin/perl: bad interpreter: No such file or directory` - so I think it is needed. But the real problem is that the interpreter is wrong it needs to be `/usr/bin/env perl`. Not sure where to change this.

bgruening:
xref: https://github.com/bioconda/bioconda-recipes/issues/2166

bgruening:
@roryk yes perl is necessary at build time for autoconf. The autoconf package should be fixed, but for the time being this should work.
Unfortunately, I got a new error which I do not understand. Maybe you can look at it :(

acaprez:
@bgruening - I think this should essentially be done now in #3752.

bgruening:
@acaprez thanks a bunch!!

PR #2477, opened 2016-10-02
Poretools fix

druvus:


recipes/poretools/0.5.0/meta.yaml
```diff
@@ -21,9 +21,9 @@ requirements:
         - h5py >=2.0
         - watchdog >=0.8.3
         - qcli >=0.1.1
-
```
bgruening: Do you want a version bump here as well?

druvus: No, version 0.5.0 didn't have a previous bioconda package. It was mislabelled as 0.5.1.

PR #2482, opened 2016-10-03
Add recipe mapdamage2

percyfal:

I have had some issues getting the tests to work. The test fails due to dependency on GLIBC_2.7, making it incompatible with the CentOS glibc version. Therefore, the tests are commented out for now.


recipes/mapdamage2/build.sh
```diff
@@ -0,0 +1,19 @@
+#! /bin/bash
+
+# seqtk is bundled with mapdamage. The path to seqtk is hardcoded in
+# the mapdamage script so simply adding a dependency on seqtk won't
```
bgruening: Can we patch this to change it to the conda path?

recipes/mapdamage2/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  skip: True # [osx]
+
+source:
+  fn: 2.0.6.tar.gz
+  url: https://github.com/ginolhac/mapDamage/archive/2.0.6.tar.gz
+  md5: f14e1ca52d68fd9c7e42e1f4e6f40f8c
+
+requirements:
+  build:
+    - python >=2.7,<3
```
bgruening: please depend on GCC and LLVM as described here: https://github.com/bioconda/bioconda-recipes/blob/master/GUIDELINES.md#cc

recipes/mapdamage2/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+    - zlib
+  run:
+    - python >=2.7,<3
+    - pysam
+    
+# Skip test for now due to error
+# Error: Could not import required module 'pysam':
+# 	- /lib64/libc.so.6: version `GLIBC_2.7' not found (required by /anaconda/envs/_build/lib/python2.7/site-packages/pysam/../../../libhts.so.1)
+#test:
```
bgruening: Submitting without tests is not allowed, please try to get this test working.
If you depend on GCC and include a libgcc as runtime dependency it might work.

recipes/mapdamage2/build.sh
percyfal: I have added two patches, one for the Makefile where the seqtk rule is replaced by a dummy output, one in composition.py in which the subprocess call now makes use of biocondas seqtk. I've been on a long hiatus and wasn't aware of the patch function, neat!

recipes/mapdamage2/meta.yaml
percyfal: I know, I was led astray by a similar comment: 

https://github.com/bioconda/bioconda-recipes/blob/7afde3c13e6fcf80ba0111611e1e630844c04a51/recipes/tbl2asn/meta.yaml#L19 

where at least the linux test is missing. However, I finally sorted out that the culprit was pysam>=0.9.0. Requiring an older version did the trick.

recipes/mapdamage2/meta.yaml
percyfal: gcc/libgcc dependency not needed; see comment on meta.yaml

bgruening:
Ha, this looks great now. Sorry for the trouble but I think it was worth it - nice recipe, nice contribution!

PR #2493, opened 2016-10-04
Fix IgBLAST’s run dependencies

marcelm:

### DIFF
--- a/recipes/igblast/meta.yaml
+++ b/recipes/igblast/meta.yaml
@@ -16,7 +16,7 @@ source:
   url: ftp://ftp.ncbi.nih.gov/blast/executables/igblast/release/1.5.0/ncbi-igblast-1.5.0-universal-macosx.tar.gz  # [osx]
 
 build:
-  number: 1
+  number: 2
 
 requirements:
   build:
@@ -25,6 +25,7 @@ requirements:
   run:
     - perl-threaded >=5.8.8
     - libgcc  # [linux]
+    - zlib
 
 test:
   commands:


marcelm:
@bgruening It seems to me that Travis does not actually build the updated package. The output does not contain any of the usual messages from the build script. Do you have an idea why that might be?

bgruening:
It seems it was passing the tests: https://travis-ci.org/bioconda/bioconda-recipes/jobs/164880610#L495

bgruening:
But this all is strange to me. Why does this makes needs zlib, if it compiles without zlib as build dependency? Have you included some pre-compiled binaries?

marcelm:
This recipe packages the IgBLAST binary distribution, so zlib is not needed in the build dependencies. The previous version of the recipe did that as well. I tried to compile from sources, but it is a huge package and failed somewhere in the middle.

> It seems it was passing the tests

Hm, ok, I think I understand now: I was expecting to see the log output from the build script in the Travis output, but it seems that is only included when something went wrong.

Ok, but then the question becomes: How did you get the error message [in your comment to PR #2390](https://github.com/bioconda/bioconda-recipes/pull/2390#issuecomment-251235928)? Why did this not fail on Travis?
```
/usr/local/bin/../share/igblast/bin/igblastn: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory
```

marcelm:
I’ll just merge this for now. The fix may be incomplete, but it is better than before.

PR #2496, opened 2016-10-04
new recipe for r-package sets, constructed using skeleton

bsennblad:

@bioconda/core

### DIFF
--- a/recipes/r-sets/meta.yaml
+++ b/recipes/r-sets/meta.yaml
@@ -41,6 +41,8 @@ requirements:
 
   run:
     - r-base
+    - libgcc              # [linux]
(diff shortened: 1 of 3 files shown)


recipes/r-sets/meta.yaml
```diff
@@ -0,0 +1,84 @@
...
+    - lib/R/lib/
+    - lib/
+
+# Suggests: proxy
+requirements:
+  build:
+    - r-base
+    - posix               # [win]
+    - {{native}}toolchain # [win]
+    - gcc                 # [not win]
```
bgruening: please add llvm for osx and libgcc for linux as runtime dependency

PR #2508, opened 2016-10-05
Gmap update

duxan:


yhoogstrate:
I have my concerns about this. I think the most elegant solution would be to pin the dependency in defuse's own conda repo: https://anaconda.org/dranew/defuse.

My worry is that when defuse dependends on version y, galaxy depends on version w and tool-x that is currently under development shall depend on version x while if I want to use it straight from the command line I would like to use the latest, version z. I think this PR is more or less fixing a deployment issue of defuse, and downgrades gmap for other use cases. I am curious how you think about this?

duxan:
You are quite right here. I rushed into switching back to original version and only after closely looking how it breaks deFuse found out it misses Perl (or some specific modules) to run with latest GMAP. 

I'll try to fix this and change this PR here, so please don't merge yet.

Thanks!

duxan:
Hi again,

It should be fixed now - latest version is default, 2015 version optional and Perl is added as runtime requirement.

Thanks!

bgruening:
You need to increase the build number, otherwise this will not be build.

duxan:
Tnx Björn, hope it's ok now.

bgruening:
@yhoogstrate @druvus is this one ready to go?

druvus:
gmap is compiling so I guess gcc/llvm/libgcc should be added.

yhoogstrate:
Ouch did i Just merge too quickly?

PR #2511, opened 2016-10-05
Add bwa and pandoc to runtime requirements

nsoranzo:

bwa is needed for scaffolding with long sequences.
pandoc is needed to produce '-stats.html' files from Markdown.

Ping @druvus.


druvus:
Thanks for fixing it!

PR #2516, opened 2016-10-06
updated vep to v86, updated dependecies, added some more plugins

matthdsm:

### DIFF
--- a/recipes/variant-effect-predictor/build.sh
+++ b/recipes/variant-effect-predictor/build.sh
@@ -31,7 +31,10 @@ perl vep_install.pl -a a --NO_HTSLIB
 # Remove test data
 rm -rf t/
 # Install plugins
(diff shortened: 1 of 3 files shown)


matthdsm:
As this is a somewhat bigger change to this recipe, I won't merge this right away. Could I get some code reviews from other committers/people using this recipe?

ping @bgruening, @chapmanb, @roryk

chapmanb:
Thanks for this update, looks good to me. In bcbio, I pinned VEP to the older version (83) for now as upgrades of VEP and snpEff are quite disruptive since they require re-downloads of databases. This is why we typically stick with a version rather than updating on every Ensembl release. It's good to have a compelling reason to re-download and prepare all those databases. Feel free to merge whenever this is ready and we can test, but will try to do an update in bcbio coinciding with a release to make the transition a bit easier.

matthdsm:
Hi Brad,

Thanks for the comment. This recipe includes some more plugins by default, which could be a nice addition to the vep annotation in bcbio. I'll comment furtheron that in the appropriate issue.

PR #2518, opened 2016-10-06
new recipe for r-package xlsxjars, constructed using skeleton

bsennblad:

### DIFF
--- a/recipes/r-xlsxjars/bld.bat
+++ b/recipes/r-xlsxjars/bld.bat
@@ -1,0 +1,8 @@
+"%R%" CMD INSTALL --build .
+if errorlevel 1 exit 1
+
+@rem Add more build steps here, if they are necessary.
+
+@rem See
+@rem http://docs.continuum.io/conda/build.html
+@rem for a list of environment variables that are set during the build process.
--- a/recipes/r-xlsxjars/build.sh
+++ b/recipes/r-xlsxjars/build.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# R refuses to build packages that mark themselves as Priority: Recommended
+mv DESCRIPTION DESCRIPTION.old
+grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
+
+$R CMD INSTALL --build .
+
+# Add more build steps here, if they are necessary.
+
+# See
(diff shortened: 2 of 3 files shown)


bsennblad:
@daler 
Ryan, 
Thanks for the suggestion. Unfortunately, that did not solve the problem. I added java-jdk as a build and run dependency in the r-xlsxjars meta.yaml, but still got the same error (see below). 
The problem occurs when loading rJava and fails to find libjvm.so. Is this not provided by java-jdk? Is there another (better/correct) way to add java as dependency? 
Is there anything else I can add or do, or could this be due to some issue with, e.g., rjava or java-jdk? I am a bit at a loss, so any suggestion is welcome.

> Error : .onLoad failed in loadNamespace() for 'rJava', details:
>   call: dyn.load(file, DLLpath = DLLpath, ...)
>   error: unable to load shared object '/opt/conda/conda-bld/r-xlsxjars_1475833877520/_b_env_placehold_placehold_placeho/lib/R/library/rJava/libs/rJava.so':
>   libjvm.so: cannot open shared object file: No such file or directory
> Error : package ‘rJava’ could not be loaded
> ERROR: lazy loading failed for package ‘xlsxjars’

(also @bioconda/core to get comments)

daler:
Hmm, I wonder if this is an issue with rJava being built in the `r` channel with a different java package. I have very little experience with java though so I'm afraid I won't be much help.

bsennblad:
I don't seem to be able to solve this at the moment. From internet searches on the error message, it seems to be related to problems in communicating the location of java-installations to R. I tested some simple quick-fixes, like trying to reset the JAVA_HOME path or linking the installed java-libs to /lib, to understand the problem better, but these did not work. I need to dig in more into how the bioconda system installs packages, e.g., where the files end up. However, as I currently are a bit short of time for this and also since bipcpnda appears to be going through some major changes now, I postpone this to a later time. I therefore close this pull-request (as unsolved) and remove the corresponding branch.

(@bioconda/core for info)

daler:
Thanks for reporting back. The changes in bioconda you refer to are unlikely to affect the problem you're seeing here. It's more related to conda itself. http://conda.pydata.org/docs/building/dirs.html has details on where things get installed.

PR #2521, opened 2016-10-06
WIP: added genesplicer recipe

matthdsm:

### DIFF
--- a/recipes/genesplicer/build.sh
+++ b/recipes/genesplicer/build.sh
@@ -1,7 +1,6 @@
-target=$PREFIX/share/
-mkdir -p $target
-mkdir -p $PREFIX/bin
-
-mv * $target && cd $target
-rm -rf bin
-cd sources/ && make && cp genesplicer $PREFIX/bin
+mkdir -p $PREFIX/bin $PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+cd sources/ && make
+rm -rf ../bin/*
+mv genesplicer ../ && cd ../
+cp -r * $PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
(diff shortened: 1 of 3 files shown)


matthdsm:
Hi,

I'm trying to get this package to install correctly, but I'm getting lost somewhere along the way.
Let me sketch the situation:

The source is a tarball which needs to be unpacked. The tarball contains a "bin" folder with executables that don't work, so they need to be built again.
The tarball also contains some data the executable needs to run. What I'd like to do is:
- [x]  `cd` to the `sources` dir and `make` the executable
- [x]  remove the entire "`bin`dir
- [ ] move all the other data to `/conda/install/dir/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
- [ ] add the executable to `/conda/install/dir/bin`

Any pointers as to how I should edit my build script to do all this?

Thanks a lot!

M

tomkinsc:
Take a look at some of the other recipes in the repo ([one example](https://github.com/bioconda/bioconda-recipes/blob/f5eb63e30a76fd13c28663786d219c9f7750267c/recipes/spades/build.sh)), but the build process is generally to:

1. `cd` to a particular source subdirectory, if necessary. Conda will automatically unpack source archives, and change the pwd to that of `$SRC_DIR`. 
2. `make`
3. `mkdir -p $PREFIX/bin` and the subdirectory `$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM`
4. Copy the compiled binaries and any dependencies to `$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM`
5. Symlink any binaries to expose on the path from `$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM` to `$PREFIX/bin/<binary name>`

bgruening:
Can you please add `libgcc` as run time dependency?

matthdsm:
done in #2529

PR #2526, opened 2016-10-06
add cummerbund 2.8.2

lecorguille:
- [X] I have read the guidelines above.
- [X] This PR updates an existing recipe.

To complete the dependencies based on conda package for Galaxy cufflinks 
https://github.com/galaxyproject/tools-devteam/blob/master/tool_collections/cufflinks/cuffdiff/cuffdiff_wrapper.xml


daler:
@lecorguille `bioconductor-gviz` and `bioconductor-rtracklayer` aren't yet built yet for R 3.3.1, may need to wait until we get to it. They're still on the blacklist.

 (cc @bgruening).

lecorguille:
@daler I'm wondering if a R library was developed at lest for for example (R (>= 2.7.0) )[https://github.com/Bioconductor-mirror/cummeRbund/blob/release-3.0/DESCRIPTION], was push on bioconductor during 3.0 compatible for (R 3.1)[https://www.bioconductor.org/news/bioc_3_0_release/]. And finally, if the current R stable version is R 3.2.5 and the last is R 3.3.1 ...
Which version must be set? Do I have to set a minimal version?
Or are we considere that there is a really really good retro-compatibility around R library?

daler:
`conda build` uses an environmental variable, `$CONDA_R`, to set which version of R to build for (along with others configured in `scripts/env_matrix.yml`). Many packages do have good backwards compatibility, but the trouble with the latest bioconductor update is that s4vectors (a dependency of many other packages) was R3.3 only.

Hence having to upgrade pretty much every R package in bioconda to support R3.3.

This recipe is passing so it's prob ok to merge. May want to try bumping the build number once the other packages are built for R 3.3.

bgruening:
Yeah, please wait if you can.

bgruening:
I assume you need this urgently, so here is a small present for your breakfast tomorrow :)

lecorguille:
@bgruening Thanks
See you soon in the tool-devteam repo around cufflinks tools suite :)

PR #2527, opened 2016-10-06
Add package examl

percyfal:

@bioconda/core


recipes/examl/build.sh
```diff
@@ -0,0 +1,8 @@
+#! /bin/bash
+set -x
+set -e
+
+cd parser && make -f Makefile.SSE3.gcc && mv parse-examl ${PREFIX}/bin
+cd ../examl && make -f Makefile.SSE3.gcc && mv examl ${PREFIX}/bin
+rm -f *.o && make -f Makefile.AVX.gcc && mv examl-AVX ${PREFIX}/bin
+rm -f *.o && make -f Makefile.OMP.AVX.gcc && mv examl-OMP-AVX ${PREFIX}/bin
```
daler: @percyfal are those optimized versions for different architectures? Will they be compiled correctly on a machine that doesn't have the corresponding architecture? (I have no idea what travis-ci is running)

percyfal: @daler well probably not. The installation instructions (from 2016) mention "recently purchased system". AVX has been around since 2011 so I'm hoping the travis system would work. Since I've been disconnected from work quite a while, I just noticed that the build deployment has changed. One evident change is that the output of the tests isn't shown in the logs, right? 

The osx SUBDAG=1 build currently fails, complaining about an unsupported clang option '-fmopen'. mpicc should use gcc which apparently seems to be out of date. Maybe adding a dependency on a newer gcc will help.

percyfal:
Building osx works locally but not on travis, despite adding libgcc runtime dependency. Skipping osx build for now.

recipes/examl/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+  skip: True # [osx]
+
+source:
+  fn: v3.0.17.tar.gz
+  url: https://github.com/stamatak/ExaML/archive/v3.0.17.tar.gz
+  md5: "bf1e2c3774646d3793fdc8a4c9145af7"
+
+requirements:
+  build:
+    - mpich
```
bgruening: You need to depend on `gcc` and `libgcc` isn't it?

percyfal: Nope, still fails for osx. Any other ideas or should I just skip osx for now?

percyfal: I'm giving this a last shot for the osx build. Unfortunately, I just realized that my erroneous push of another feature branch to master has resulted in two packages being built here - see PR #2690 for rpkmforgenes. Since I already pushed to my forked repo, rpkmforgenes has already been merged with master. I don't know if this is problem for the bioconda commit history once PR #2690 is merged. One option would be that I fork bioconda anew and resubmit these pull requests based on a correct bioconda history. What do you think?

PR #2543, opened 2016-10-07
Build poretools

druvus:

Comment poretools in R blacklist so it will be rebuilt.


bgruening:
@druvus feel free to help taking more of these packages from the blacklist :)

PR #2544, opened 2016-10-07
Added recipe for nudup NuGEN deduplication tool

lparsons:

### DIFF
--- a/recipes/nudup/bld.bat
+++ b/recipes/nudup/bld.bat
@@ -1,0 +1,3 @@
+md "bin"
+copy "nudup.py" "%PREFIX%\bin"
+if errorlevel 1 exit 1
--- a/recipes/nudup/bld.bat
+++ b/recipes/nudup/bld.bat
@@ -1,3 +1,0 @@
-md "bin"
-copy "nudup.py" "%PREFIX%\bin"
-if errorlevel 1 exit 1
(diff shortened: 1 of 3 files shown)


recipes/nudup/bld.bat
```diff
@@ -0,0 +1,3 @@
+md "bin"
+copy "nudup.py" "%PREFIX%\bin"
```
bgruening: We do not support windows so we can ignore these files :)

recipes/nudup/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  name: nudup
+  version: "2.2_post2016104"
+
+source:
+  git_rev: 740d9fe439dd8917605a56483a8796b377eb24c6
+  git_url: https://github.com/nugentechnologies/nudup.git
+
+build:
+  number: 0
+  skip: True # [win]
```
bgruening: [win or py3k]

lparsons: I wondered about that, but didn't see it in the docs. Thx.

recipes/nudup/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  skip: True # [py3k]
+
+requirements:
+  build:
+    - python
+    - setuptools
+
+  run:
+    - python
+    - samtools >=0.1.18
```
lparsons: I looked for that in the docs, but couldn't find it. Thanks for pointing it out.

lparsons:
@bioconda/core please review

PR #2554, opened 2016-10-08
Add libgcc to vsearch

bgruening:

### DIFF
--- a/recipes/vsearch/meta.yaml
+++ b/recipes/vsearch/meta.yaml
@@ -1,20 +1,29 @@
 package:
   name: vsearch
   version: "2.0.3"
+
 source:
-  git_url: https://github.com/torognes/vsearch.git
-  git_rev: v2.0.3
+  fn: v2.0.3.tar.gz
+  url: https://github.com/torognes/vsearch/archive/v2.0.3.tar.gz
+  md5: 2fa09ab54ba88b41c068d671d7193df9
+
+build:
+  number: 1
+
 requirements:
   build:
     - automake
     - autoconf
     - m4
-    - gcc
-build:
-  skip: False
+    - gcc  # [linux]
+    - llvm
+  run:
+    - libgcc  # [linux]
+
 test:
   commands:
     - vsearch --version > /dev/null
+
 about:
   home: https://github.com/torognes/vsearch
   license: "dual-licensed under GPLv3+ or BSD 2-clause"


colinbrislawn:
What's causing this failure? I'm not sure if I can see the error message.
(Sorry for my inexperience. I appreciate your help.)

colinbrislawn:
Based on comments to PR #2653, looks like autoreconf does need perl, but this build is failing because it tries to find perl using a static path.

Does that help at all? 
(I'm at a loss to address this, and would to get vsearch building again...)

druvus:
The issues lies in the hard-coded perl shebang path in autoreconf. Autoreconf is a helper script included in autoconf written in perl.

keuv-grvl:
The issue noticed by @druvus does exist in the `autoconf` package from bioconda (shebang is  `#!/usr/bin/perl` in `autoreconf`) but not in the package from conda-forge (shebangs is `#!/anaconda/bin/perl` in `autoreconf`).

Since the bioconda channel uses packages from conda-forge, might we drop `autoconf` from bioconda (and other duplicates)?

colinbrislawn:
@keuv-grvl Thanks for the feedback!

Just to make sure I understand, it's the version of `autoreconf` in bioconda that's causing this problem due to strange shebang? Would it be simpler to fix the bioconda package or remove it all together?

druvus:
Yes is due to the strange shebang. Autoconf change the shebang to use absolute path to the perl version used during building. 

I tried to fix the bioconda autoconf recipe in #2646 but I was getting weird error. I also tried manually using the image and it works if I preinstall perl-threaded but not if it is listed as dependency.

keuv-grvl:
@colinbrislawn Yes, the package from bioconda is causing the problem. It is not the version but the build strategy (see [here](https://github.com/bioconda/bioconda-recipes/blob/861ae24c9ab69d0fd999993b88cc2dd9823503ff/recipes/autoconf/build.sh) and [here]( https://github.com/conda-forge/autoconf-feedstock/blob/master/recipe/build.sh)) and environment as stated by @druvus.


I manually built vsearch by using the conda-forge autoconf package (`2.69-pl5.20.3.1_3`). So relying on conda-forge would be easier than maintaining 2 copies of the same tool.
Also, being general purpose software, conda-forge would be a better place for autoconf.

bgruening:
@keuv-grvl does this solve your problem: https://github.com/bioconda/bioconda-recipes/blob/fd57b3a52bf94a41e783424f1db3364f66be7bbe/recipes/sailfish/build.sh

colinbrislawn:
I'll try the sailfish workaround now! 

I like the idea of moving autoconf to the appropriate location, but I'm not the one to decide things like that. Maybe @bgruening would be :-)

bgruening:
I would merge a PR removing autoconf :)

colinbrislawn:
If we ~~nuke~~ omit autoconf from bioconda, would the builds just default to conda-forge, thus solving the problem? I'm pretty new to this and don't want to break anything!

daler:
I believe so. But it also means that anyone that ever used the autoconf from bioconda would not be able to use it again. Not good for reproducibility, but I figure since it's more or less broken as-is, it would be fine to remove.

@bgruening @johanneskoester thoughts on deleting the package from anaconda?

bgruening:
No, just removing it will not magically solve the problem. It is still on anaconda.org. As soon as the conda-forge version has a higher version then bioconda has we will get this package.

We could also could remove all packages from anaconda.org. As it is a build dependency I think we can remove them ... for runtime deps we can not do this without loosing reproducibility.

druvus:
The conda-forge version seems to correctly specify the absolute path to perl for the used environment. The only issue is that they have pinned the perl version but it might not be an issue as autoconf is just required during building.

druvus:
@bgruening @daler It seems I can get biocondas autoconf to pass the travis test if I add pkginfo as a dependency. I will just also let the osx test to start to see if I actually managed to repair the recipe.

keuv-grvl:
@bgruening I ran `conda install autoconf` then `conda install sailfish`, but the latter did not change shebang of `/anaconda/bin/autoreconf`.

Simply dropping autoconf from bioconda could be quite brutal. But I guess it will be a good long-term solution.

bgruening:
This one is done now ... it's green with the new autoconf package. Thanks to @druvus!

colinbrislawn:
> new autoconf package
👏

daler:
@bgruening looks like it's failing on master branch? 

https://travis-ci.org/bioconda/bioconda-recipes/jobs/173929221#L517

Subsequent PRs are failing too because of this, e.g., 

- https://travis-ci.org/bioconda/bioconda-recipes/jobs/173960296#L518 
- https://travis-ci.org/bioconda/bioconda-recipes/jobs/173965648#L517

cc @marcelm, @colinbrislawn

bgruening:
What? Why? This was passing on the PR  branch and autoconf is uploaded `2.69                 pl5.22.0_3` and should get installed :(

druvus:
weird, seems it only select anaconda/default package versions for all dependencies. Did something happen to the channel order?

johanneskoester:
The master branch passes again. I guess this was a temporal issue with anaconda? Or has it been fixed in another PR?

bgruening:
I fixed it and blacklisted it. Still there is a channel issue somewhere ... Sorry not in front of a computer to fix it.

PR #2566, opened 2016-10-08
Rebuild more r packages

bgruening:

### DIFF
--- a/recipes/bioconductor-genefilter/meta.yaml
+++ b/recipes/bioconductor-genefilter/meta.yaml
@@ -12,15 +12,14 @@ build:
     - lib/
 requirements:
   build:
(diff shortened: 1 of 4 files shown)


bgruening:
@daler not sure but it seems that we have a `loop` forever candidate here. Do you own a Mac or how have you figure out which package is the bad guy?

daler:
Won't have access to a mac until Tuesday. I had figured out which was looping by running locally and watching the log, but there was nothing about the recipe that caused it. Must be something in the R package tarball.

bgruening:
This one is also interesting.
```
  - llvm
  - r 3.3.1*
```

bgruening:
Finally! OSX is such a time sink.

PR #2569, opened 2016-10-09
remove libgd; rebuild gnuplot with conda-forge dependencies

croth1:


croth1:
@johanneskoester, @bgruening - do you know why the build log does not show what conda packages were installed into the build environment? 

https://travis-ci.org/bioconda/bioconda-recipes/jobs/166176190#L501 seems to directly start running the config script.

bgruening:
@croth1 the output is turned around, we are searching for the issue. But in the meantime look at the bottom: https://travis-ci.org/bioconda/bioconda-recipes/jobs/166176190#L1149

croth1:
```
    cloog:      0.18.0-0 anaconda   
    expat:      2.1.0-0  anaconda   
    fontconfig: 2.11.1-3 conda-forge
    freetype:   2.6.3-1  conda-forge
    gcc:        4.8.5-6  anaconda   
    gmp:        5.1.2-2  anaconda   
    isl:        0.12.2-0 anaconda   
    jpeg:       9b-0     conda-forge
    libgcc:     5.2.0-0  anaconda   
    libgd:      2.2.3-0  conda-forge
    libiconv:   1.14-3   conda-forge
    libpng:     1.6.22-0 anaconda   
    libtiff:    4.0.2-0  anaconda   
    libxml2:    2.9.2-0  anaconda   
    mpc:        1.0.1-0  anaconda   
    mpfr:       3.1.2-0  anaconda   
    ncurses:    5.9-8    anaconda   
    zlib:       1.2.8-3  anaconda
```

Has the `anaconda´ channel precedence over the conda-forge channel in our current build system? `libpng` and `libtiff` have conda-forge packages, yet in the build environment the anaconda packages are installed. 
This can put us into dependency hell if the anaconda package does not work as a drop-in replacement (e.g. differently configured version).

croth1:
@bioconda/core I am blocked here, would appreciate any help.

daler:
The order of the channels is defined [here](https://github.com/bioconda/bioconda-recipes/blob/master/config.yml#L7), does switching them help?

johanneskoester:
We can't put conda-forge above anaconda, because that will break the r channel, which was built without conda forge, right? Also people using conda forge will always have anaconda with a higher priority by default, unless they explicitly switch the order I think. Sounds again like channel dependencies would be a much better solution than these unintuitive implicit channel priorities.

croth1:
So currently there is no way to solve this problem?

johanneskoester:
As far as I know we add the conda -forge channel in the same order as suggested by the conda -forge team. Can you isolate the problem to one or a combination of their packages? If so, I think it has to be solved in conda-forge.

croth1:
@jakirkham, if you have some time, I would appreciate if you could have a quick look at https://github.com/bioconda/bioconda-recipes/pull/2569#issuecomment-252490246. Do you happen to know if there's a way to convince conda to use `libpng`, `libtiff`, etc from conda-forge instead of the official anaconda repositories here, without breaking the `r` channel?

jakirkham:
So the way we have it at conda-forge is basically conda-forge and then defaults. Please let me know if you have more questions.

croth1:
@jakirkham, thanks. I agree that starting to support r in conda-forge would be a great leap forward for us. We might need support from @mingwandroid, though. Packaging r from the scratch would be quite some work I guess. Unfortunately I am busy till the end of the month.

johanneskoester:
Well that just sounds like a specific workaround for a more general problem. Assuming we keep the channel priorities as they are now, I think the following simple rule would fix conda here: 'dependencies may never pull package versions from lower priority channels'. What do you think?

johanneskoester:
Better rule: 'when resolving dependencies, prefer priority over package version'.

johanneskoester:
@croth1 I just discovered that our channel order is `anaconda, conda-forge, r, bioconda`, and not `anaconda, r, conda-forge, bioconda` as I thought. This order is actually wrong in my opinion, because of the reasons outlined above. Can you change the order to the second case in config.yaml of your branch? Maybe this already fixes the dependency problem, because conda then already does what I want it to do.

croth1:
@johanneskoester, I am trying now. Let's see what's happening. On a different note: shouldn't bioconda have the highest priority instead of the lowest?

croth1:
I think the right way to solve this is to get the packages in the  r-channel into conda-forge and then drop the r-channel completely.

This will have several advantages

1) We will not get breakage from differently compiled libraries (e.g. ncurses)
2) We don't need any workarounds for broken packages in anaconda
3) We can fix and update r-packages without relying on continuum analytics

johanneskoester:
If conda forge wants to dedicate time to that, they can of course do it. On the other hand, the r channel was always reliable for me, so I don't see a particular reason for duplicating work. Anyway, looking at the error, the previous problems are gone now, right?

croth1:
Well not really. The problem is that it gets hard to predict which channel the dependent packages come from. The linux build apparently pulled in the working libgd from conda-forge and succeeded. The mac build pulled in the broken libgd from the anaconda channel and failed.

johanneskoester:
Ok. So apparently there is either a package causing the discrepancy in dependency resolution between osx and linux or a bug in conda that causes it. I think that is the next step that should be investigated now, shouldn't it?

croth1:
Yes, definitely. I'll try to track the problem down when I have more time. However I get the impression we are running into fundamental problems here that sooner or later will break our neck.

Including the conda-forge channel was an important step into the future, but now we end up mixing packages from channels that are not necessarily compatible. This is painful for us here at github, but it is even a bigger problem for the user. If I am not mistaken, they have to have exactly the same channel order as we have, otherwise they might end up with broken libraries.

I guess the problem is that conda-forge and anaconda did not develop the way I expected. I was assuming that conda-forge is a community effort with all available packages to which continuum will contribute. The anaconda channel then offers a subset of the conda-forge packages, ensuring compatibility between the channels.

At the moment I am not sure how this should be solved. I think a solution would be to move all required packages from the anaconda and r channel to conda-forge and then set the channel order to `bioconda` > `conda-forge` > `r` > `anaconda`. We would tell our users: "bioconda depends on conda-forge", forcing them to have the same channel ordering.

The fundamental problem here is that getting the r infrastructure to conda-forge will be quite some effort, as r has to be patched heavily. We would probably need some help from Continuum, otherwise this will require huge time investments from our side.

I'd love to hear your opinions or alternative solutions.

ping @jakirkham, @mingwandroid, @msarahan

mingwandroid:
> I guess the problem is that conda-forge and anaconda did not develop the way I expected. I was assuming that conda-forge is a community effort with all available packages to which continuum will contribute. The defaults channel then offers a subset of the conda-forge packages, ensuring compatibility between the channels.

I think this needs to be brought to the attention of @csoja, @toliphant and @pelson. I agree that *full* binary compatibility is a necessity for cooperative progress. How that is achieved is the tricky bit.

johanneskoester:
Exactly. For now, the best bioconda can do is to enforce a channel order that resembles a topological sort of the channel dependency graph. And ensure that conda uses that properly on all systems. In the future, we need a good strategy for the integration of the three main sources of packages : conda-forge, r and default. This strategy should I think either involve a policy about binary compatibility of duplicate packages, or substantial improvements to the dependency management of conda/conda-build, namely pinning beyond plain versions or something like channel dependencies.

johanneskoester:
Dear all, I have written a summary in an appropriate issue in the conda repos (conda/conda#3721). Please join the discussion over there. We really need to find a solution as soon as possible. Thank you!

PR #2577, opened 2016-10-09
updated sra-tools to version 2.8.0

croth1:


croth1:
```
      for name, source_section in extra_sources_sections.items():
          # Override the location to clone into
          source.WORK_DIR = main_work_dir + '/' + name
          os.makedirs(source.WORK_DIR)
```
The download script uses conda-build code that is now removed. It seems that the proper way to access the workdir now is using `conda_build.config`. However the config seems to be read-only now. 
Do need advice here. Ping @simonvh, @msarahan

msarahan:
Each recipe build now has its own build folder.  A global WORK_DIR no longer makes much sense.  As of conda-build 2.0, I would advise you to stay away from directly using any part of conda-build outside of the api module.  That module's interface is tested and we promise to keep it stable (at least we'll update the major version number if we change it.)

The newer way to do things is to create a Config object using that API, and then change that config object as you see fit.  You can also get MetaData objects by rendering your recipe, and these store a config object.  This object overrides config options passed to API functions, so if you pass around MetaData objects, be sure to change their config member directly.

The work_dir member of a config object is a derived property, and is not directly settable.  https://github.com/conda/conda-build/blob/master/conda_build/config.py#L348-L358

The settable thing instead is the build id, which determines the build folder, which is the folder in which the work_dir, test dir and build and test environments live.  https://github.com/conda/conda-build/blob/master/conda_build/config.py#L229-L234

If you have ideas on how to improve the way things work, I'm open to them.  This stuff is all pretty new.

simonvh:
Basically the script kind of hijacks the conda-build `provide` function to
download extra sources. I had a quick glance, but I think this
functionality is not provided in the API. So this needs a little more work.
I would be happy to have a more detailed look at it, but I'm wrapped up in
teaching a full-time course these two weeks, so not a lot of time.

croth1:
@msarahan, would it be possible to natively support multiple multiple sources in a meta.yaml? conda does not seem to support the download script by @stuarteberg anymore, and I don't see an easy way to fix that.

The only workaround I see at the moment is to download the extra sources in build.sh, but security-wise this is a bad idea. I would like to avoid that if possible.

msarahan:
@croth1 sure.  I won't be able to get to it myself for quite a while, though.  Probably looking at 1-2 months right now.  I'll gladly accept PRs that add this capability in the meantime.

croth1:
depends on: https://github.com/conda/conda-build/issues/1466

PR #2587, opened 2016-10-10
Bump and fix blast legacy

druvus:

### DIFF
--- a/recipes/blast-legacy/2.2.22/build.sh
+++ b/recipes/blast-legacy/2.2.22/build.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 cp -R bin $PREFIX
 cp -R data $PREFIX
 cp -R doc $PREFIX
--- a/recipes/blast-legacy/2.2.22/meta.yaml
+++ b/recipes/blast-legacy/2.2.22/meta.yaml
@@ -14,6 +14,13 @@ build:
   number: 1
   skip: True # [not linux64 or osx]
 
+requirements:
+  build:
+    - gcc   # [not osx]
+    - llvm # [osx]
+  run:
+    - libgcc # [not osx]
+
 test:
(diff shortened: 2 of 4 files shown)


druvus:
NCBI provide as-is pre-compiled binaries but they doesn't work for linux. Seems to be some compiler incompatibility both for the previous included and the newest version.  

A better option would be to compile blast from source using the NCBI toolkit (ftp://ftp.ncbi.nlm.nih.gov/blast/executables/legacy/2.2.26/ncbi.tar.gz). The build scripts are quite extensive so does anyone have experience of build blast from source with a minimal build script? 

This is also valid for other ncbi tools included in the ncbi toolkit, such as tbl2asn. 
@bioconda/core

tomkinsc:
It would be great to have tbl2asn built from source.

martin-mann:
I will have a look if there is some magic going on when building the NCBI toolkit from source..

martin-mann:
@druvus : have send a pull request to your branch, which compiles ncbi tools from source. It is not thoroughly tested yet and was not checked by travis. please have a look!

druvus:
The build fails due to that we don't have csh installed on the image. I am not using that shell so often so I don't know if we can get it installed by conda.

druvus:
Closing this PR as it seems to be a solution in #2795 .

jlanga:
#2795 refers to `blast+`, not `blast-legacy`. For a while NCBI developed both at the same time and they even share version numbers.

acaprez:
Sorry to bump an old thread, but I was looking into adding `blast-legacy` 2.2.26 using the pre-built binaries just now.  @druvus - what's the issue you mentioned with using the NCBI provided binaries on linux?  They seem to work on EL6 and EL7 - or does the problem only show up when running with real input?

Issue #2148, opened 2016-08-15
Freebayes parallel and vcflib issues


jrderuiter:
Freebayes parallel is currently not working because the required vcflib scripts (such as vcffirstheader) are not copied over when freebayes is installed. This can be solved in two possible ways:
- We can build freebayes' vcflib submodule as part of the freebayes build and copy over the required scripts to the bin directory.
- Or we can make the freebayes package depend on the vcflib package and use the scripts from that package.

In both cases we would need to patch the freebayes-parallel script to look for the vcf scripts in PATH. 

The second option has my preference, as this separates the build scripts/requirements of freebayes and vcflib into the two distinct packages. This option does however require updating the vcflib package to a newer revision than is currently tagged as release candidate on github, as the release candidate does not include the required scripts. What would be the appropriate way to tag this newer version? In my own anaconda account I tagged the version as '1.0.0_rc1_head', although I am not sure if this is the ideal name for the newer build

yesimon:
I actually wouldn't recommend making it depend on the vcflib package precisely because freebayes is vendoring it and is only promising for it to work with the specific vendored version. Looking at it briefly it seems like all references to vcflib scripts are relative instead of looking on the `$PATH` so checking out the submodule should work fine.

bgruening:
@jrderuiter can you work on option 1?

jrderuiter:
I'll try to have a go in the coming week or two!

jrderuiter:
See 2588.

jrderuiter:
See pull request #2588.

tiagoantao:
Is there any update on this? Currently this is still not working, I think

nsoranzo:
Fixed in https://github.com/bioconda/bioconda-recipes/pull/4422.

Slugger70:
This is still not working.. Currently, freebayes-parallel.sh goes looking for the vcflib scripts in a specific location (../vcflib/bin/xxx).. The last part of the build.sh file copies them to $PREFIX/bin and so freebayes-parallel can't find them.

A solution would be to change the path in freebayes-parallel or to actually put them in the path specified in the script.. I guess the prefered solution would be to change the freebayes-parallel shell script so that the correct path is used..

PR #2588, opened 2016-10-10
Added vcflib scripts to freebayes recipe.

jrderuiter:

This pull request is a first stab at fixing the issues mentioned in #2148. Essentially, the modified recipe now also builds the vcflib submodule and copies the required scripts + library from the vcflib directory for freebayes-parallel.

The new recipe works fine on linux (tested in a docker container), but I have been unable to get it working on my Mac. The main issue lies in failing to build vcflib, which I think requires the C++11 stdlib for clang. The error I am currently getting on OSX is as follows:

``` bash
ar: creating archive libvcflib.a
cp libvcflib.a lib
cd intervaltree && /Applications/Xcode.app/Contents/Developer/usr/bin/make && cp *.h* /Users/Julian/Anaconda/conda-bld/freebayes_1476131480742/work/vcflib/include/
g++ -Wall interval_tree_test.cpp -o interval_tree_test -std=c++0x
clang: warning: libstdc++ is deprecated; move to libc++ with a minimum deployment target of OS X 10.9
interval_tree_test.cpp:2:10: fatal error: 'thread' file not found
#include <thread>
         ^
1 error generated.
make[1]: *** [interval_tree_test] Error 1
make: *** [intervaltree] Error 2
```

Any tips are welcome, as I would like the package to keep support for OSX! (An alternative would be to make a separate freebayes-parallel package that drops OSX support, if need be.)

### DIFF
--- a/recipes/freebayes/build.sh
+++ b/recipes/freebayes/build.sh
@@ -59,6 +59,3 @@ cp scripts/generate_freebayes_region_scripts.sh $PREFIX/bin
 cp vcflib/bin/vcfstreamsort $PREFIX/bin
 cp vcflib/bin/vcfuniq $PREFIX/bin
 cp vcflib/scripts/vcffirstheader $PREFIX/bin
-
-# Copy library for vcflib.
-cp vcflib/tabixpp/htslib/libhts.so* $PREFIX/lib
--- a/recipes/freebayes/build.sh
+++ b/recipes/freebayes/build.sh
@@ -5,7 +5,7 @@ if [ "$(uname)" == "Darwin" ]; then
    sed -i.bak 's/LDFLAGS=-Wl,-s/LDFLAGS=/' vcflib/smithwaterman/Makefile
 fi
 
-# bamtools/cmake for zlib
+# Bamtools/cmake for zlib.
 export CPLUS_INCLUDE_PATH=${PREFIX}/include
 export LIBRARY_PATH=${PREFIX}/lib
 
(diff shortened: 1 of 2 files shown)


jrderuiter:
Note than I am now also copying over the libhts.so included with the vcflib build. Would it be cleaner to depend on libhts (if this is possible) or to include the built .so file?

I am slightly worried that including these executables/libraries from vcflib may result in conflicts down the line.

bgruening:
Depending on htslib would be cleaner, yes. Thanks for working on this!

jrderuiter:
I just added the dependency on htslib. I still haven't been able to get this working on OSX though, due to the above mentioned build error. The linux version should work fine.

bgruening:
@chambm any comments?

chambm:
`-std=c++0x` will not have <thread>. It was added in c++11. Figure out how to get the build system to pass `-std=c++11` or `-std=c++14` instead.

jrderuiter:
@chambm Any tips where to start?

Otherwise I would recommend merging this and adding OSX support later.

chambm:
Assuming [this](https://github.com/ekg/intervaltree/blob/master/Makefile) is the file that's being built (intervaltree/Makefile), the Makefile will have to be edited because it hardcodes the -std=c++0x flag. A quick sed command should take care of that!

bgruening:
@jrderuiter are you still working on this? If so can you rebase this branch?

jrderuiter:
I closed this pull request as I accidentially deleted the associated repo/branch. See #4422 for an updated pull request.

PR #2590, opened 2016-10-11
RcppGSL: 'Rcpp' Integration for 'GNU GSL' Vectors and Matrices

rasmusagren:

@bioconda/core


molpopgen:
Set the GSL dependency to - gsl {{CONDA_GSL}}*

See comments in #2729 for issues related to gsl/linux/bioconda.

rasmusagren:
@molpopgen Thanks, but that only helped partly :) It will still get GSL from the r channel rather than from conda-forge. I will exclude OS X for the time being and wait until there is an option to specify channel order.

PR #2595, opened 2016-10-11
R 15 or so

daler:

more r packages


bgruening:
The title is great ... it indeed should be something like 17 or so ;)

daler:
@bgruening looks like we have another looper?

bgruening:
@daler yes, this is really annoying. I commented out the OSX build in an other PR to get at least the Linux builds done :(

bgruening:
This is so sad :(

daler:
It seems that building packages for osx takes about an order of magnitude more time to debug than linux. Pretty frustrating.

bgruening:
@daler https://github.com/conda/conda-build/issues/1449 maybe this one?

daler:
@bgruening I saw that, but wasn't convinced it was the issue. I just opened this one: https://github.com/conda/conda-build/issues/1454.

daler:
@bgruening one difficulty here is that the build does not immediately fail.

For example, [this error](https://travis-ci.org/bioconda/bioconda-recipes/jobs/167058058#L1585) is not, in fact, because `bioconductor-annotate` doesn't exist. Rather, it's because `bioconductor-category` failed (mismatching md5sum -- seems bioarchive md5s don't match bioconductor?) and the build didn't stop at that point.

 I'm updating bioconda-utils to stop the build at first failure which should make debugging a lot easier.

bgruening:
@daler yeah md5sum are different. bioarchive will create tarballs out of svn, so the used zlib version or whatever can be different and the tarball is not the same.

daler:
@bgruening https://github.com/conda/conda-build/issues/1432, fixed in conda-build 2.0.5.

I'm re-writing parts of bioconda-utils to use conda_build.api (it's working now with 2.0.4), and will now test with 2.0.5.

then I'll include the new bioconda-utils tag in the `.travis.yml` file in this PR and see how it goes . . .

bgruening:
Nice, let me know if you need help or a review!

bgruening:
@daler even the linux builds seem to have problems finding dependencies that should be available.

daler:
You're right, and those tests should not be passing on linux.

daler:
@bgruening the only thing untested is the uploading to anaconda. I'm currently factoring out the upload code to a new module for better testing, and I created a new testing token for anaconda. Once uploading is tested, we should be ready to start using bioconda-utils + mulled-build.

bgruening:
I fixed the tests and released a new galaxy-lib version.

daler:
closing in favor of #2694

PR #2601, opened 2016-10-12
gvcftools 0.17.0 recipe

pvanheus:
* [*] I have read the guidelines above.


bgruening:
The description says its also a bunch of perl scripts? I guess you need perl as runtime dep?

pvanheus:
I fixed the description and added the perl dependency - there is a single perl script.

daler:
@pvanheus looks like the linux package does not build (can't find zlib), and has not been uploaded to anaconda despite being merged in.

@bgruening, any idea why this wasn't picked up as an error by the build system?

bgruening:
My only idea is that for 1-2 days Travis was unreliable because of migrations of the Travis build system. I don't think this was our fault.

pvanheus:
It has zlib as a build and runtime dependency, not sure why it should give a build error?

daler:
True, but the same thing goes for, e.g., samtools where the [`build.sh` has a bunch of fixes](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/samtools/build.sh) to allow it to find zlib.

PR #2604, opened 2016-10-12
inital push to troubleshoot synergy finder

mcornwell1957:

@bioconda/core

Hi all - 

First time creating a new package. I have been doing a lot of reading to try and solve all of my issues, and have made some progress, but am pretty stuck now. Overall, I wanted to add a new package [synergyfinder](https://cran.fhcrc.org/web/packages/synergyfinder/index.html). This is a standard CRAN package, so I was hoping it would be easy.

Upon skeleton building synergyfinder, I saw that it needed 3 other packages not available anywhere, [drc](https://cran.fhcrc.org/web/packages/drc/index.html), [kriging](https://cran.r-project.org/web/packages/kriging/kriging.pdf), and [nleqslv](https://cran.r-project.org/web/packages/nleqslv/index.html). So I took on building those as well.

After a lot of troubleshooting, I got kriging to build, and test succesfully. I got nleqslv to build, but not test successfully. And drc I created the recipe for it, but it is having a dependency issue.

I wasn't sure where to start, so let me know which errors, info, etc. we need to get this working!

Thank you very much,

Tosh

### DIFF
--- a/recipes/r-drc/bld.bat
+++ b/recipes/r-drc/bld.bat
@@ -1,0 +1,8 @@
+"%R%" CMD INSTALL --build .
+if errorlevel 1 exit 1
+
+@rem Add more build steps here, if they are necessary.
+
+@rem See
+@rem http://docs.continuum.io/conda/build.html
+@rem for a list of environment variables that are set during the build process.
--- a/recipes/r-drc/build.sh
+++ b/recipes/r-drc/build.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# R refuses to build packages that mark themselves as Priority: Recommended
+mv DESCRIPTION DESCRIPTION.old
+grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
+
+$R CMD INSTALL --build .
+
+# Add more build steps here, if they are necessary.
+
+# See
+# http://docs.continuum.io/conda/build.html
+# for a list of environment variables that are set during the build process.
--- a/recipes/r-drc/meta.yaml
+++ b/recipes/r-drc/meta.yaml
@@ -1,0 +1,95 @@
+{% set posix = 'm2-' if win else '' %}
+{% set native = 'm2w64-' if win else '' %}
+
+package:
+  name: r-drc
+  # Note that conda versions cannot contain -, so any -'s in the version have
+  # been replaced with _'s.
+  version: "3.0_1"
+
+source:
+  fn: drc_3.0-1.tar.gz
+  url:
+    - http://cran.r-project.org/src/contrib/drc_3.0-1.tar.gz
+    - http://cran.r-project.org/src/contrib/Archive/drc/drc_3.0-1.tar.gz
+
+
+  # You can add a hash for the file here, like md5 or sha1
+  # md5: 49448ba4863157652311cc5ea4fea3ea
+  # sha1: 3bcfbee008276084cbb37a2b453963c61176a322
+  # patches:
+   # List any patch files here
+   # - fix.patch
+
+build:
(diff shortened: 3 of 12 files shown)


bgruening:
Just a note until I see the error, you can use the skeleton build with a --recursion mode that will create all dependencies.

mcornwell1957:
@johanneskoester 
Hi Johannes (and others)
Please let me know what you need/what I can do to keep this process going,
Thank you very much!!

mcornwell1957:
all - I took another try at this, and discovered the --R flag in the conda build function. So I can build all dependencies with an R 3.2 base, but synergyfinder still expects packages that require an r-base 3.3.1. So is this a dead end? 

It seems like I can't get all packages on the same r-base, because if I downgrade the dependencies (r-drc, r-nleqslv, r-kriging) to r-base 3.2, then drugsynergy is unhappy, and if I keep them at r=3.3.1, then r-drc won't load.

Thank you very much, let me know what you think,

johanneskoester:
The problem is that several dependencies of r-drc are available for `r < 3.3` only in the R channel. @bgruening I thought you said that the R channel has moved to R 3.3.1 some months ago. This looks like we did our move to 3.3.1 a bit too early, didn't we? Any idea who we can bug about updating those packages? I don't want to start replicating the whole R channel here in bioconda.

bgruening:
@johanneskoester the `r` channel has published the new R version a few month back and many packages are migrated. I don't think this move was too early, R3.4 will be released soon and we failed to deliver the newest DESeq2 version since 6 month. 
If this does not work we need an other solution for R, without rebuilding everything.

Afaik, conda-forge is considering taking over all packages in `r`.

johanneskoester:
Woops, my fault. Actually I missed the 3.3.1 versions of the packages in the list. The packages from the R channel are not the problem @bgruening . Sorry!!

The real problem is r-plotrix from bioconda itself. It has a `skip: True` for OSX, but only since R 3.3.1. @bgruening git tells me you added the skip. Can you remember the reason? Do you think we could fix it?

bgruening:
@johanneskoester we have a lot of problems with R and the most recent conda versions. I think we can solve this if we find a stable conda version and get @daler new build-framework in.

So timeline imho would be:
 * get https://github.com/bioconda/bioconda-recipes/pull/2595 in and with this the new mulled-docker tests and a way way better build tool, thanks to @daler 
 * work on the r_blacklist and rebuild everything
 * work on all these small packages that made problems, like r-plotrix

johanneskoester:
Thanks for the update. Sorry if I appear to be uninformed. I just came back from vacation and I still try to walk through all those hundreds of emails. @daler @bgruening  let me know if you need further code reviews on the new bioconda-utils branch, but I fully trust you. So whenever you feel it is ready for merging, please go ahead.

@mcornwell1957 sounds like you need to be patient regarging the osx build. Can you live with adding a `skip: true # [osx]` to `r-drc` and synergyfinder for now?

mcornwell1957:
Sounds like I don't have a choice! haha. Ok, keep me updated and let me know if I can help.

Thank you for your help!

daler:
@mcornwell1957 sorry about the R mess. It is a rabbit hole that keeps getting deeper. And OSX makes it a lot harder and more time-consuming.

Do you have access to a Mac for testing? It would help if you could try getting https://github.com/bioconda/bioconda-recipes/pull/2595 to work on a Mac. For example `bioconductor-graph` needs to be fixed. 

You can also try this to more directly work on synergyfinder:

- create a new branch based on #2595 
- revert the changes in `sorted-r-blacklist` back to master
- comment-out the `skip: True` line in `r-plotrix` recipe
- try building `r-drc`

PR #2610, opened 2016-10-13
fix #2609

### LINKED ISSUE #2609: Issues with the simulate_travis script
Team Bioconda,

Thanks for the updated instructions for local builds.  This is great. However, I've encountered a few errors when running the `simulate_travis.py script`.

1.  PATH variables aren't passed. I'm on OSX and when I run simulate_travis i get an error about conda. It seems that the script calls out to scripts/travis-run.py and that this gets its PATH from the following line:

    ```export PATH=/anaconda/bin:$PATH```

    In my case the anaconda bin PATH is in my home directory. Can we autodect the anaconda locations?

2.  Can we specify the OS to build on the CLI? It would be nice to specify `linux` builds as well.

Thanks.
zach cp

daler:

xref #2609

### DIFF
--- a/simulate-travis.py
+++ b/simulate-travis.py
@@ -23,6 +23,21 @@ or modify the log level:
 
     simulate-travis.py --packages mypackagename --loglevel=debug
 
+Notes
+-----
+
+Any environmental variables will be passed to `scripts/travis-run.sh` and will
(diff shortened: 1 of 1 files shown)


bgruening:
Restarted the travis job.

PR #2618, opened 2016-10-14
recipe for AbundanceBin

keuv-grvl:

`./simulate-travis.py --loglevel=debug --packages abundancebin` successfully built the recipe.
A test file is distributed with the software. I made it smaller.


daler:
Would you please remove the test data? We're not aiming for functional tests, just that the program successfully runs. `command -v` locates it on the path but doesn't actually execute the program. Typically just calling the help of the tool works to find compiler or missing library issues.

A little more info is in the [guidelines](https://github.com/bioconda/bioconda-recipes/blob/master/GUIDELINES.md#tests).

keuv-grvl:
This functional test takes about 10 seconds, so I supposed it was OK.

A note about that might be useful in the guidelines.

daler:
Sure, I'll add that to the guidelines. It's more about the test file than the time -- if we had a small test file for every one of the 1500+ recipes it would add up in terms of storage space on github, clone time, and storage on continuumio's servers.

If this program doesn't output help like most command-line tools and there's no other good way to check that it can be executed, then go ahead and add the test file back in. But try to make it as small as possible (i.e. will it work with a single record with a 10-bp-long sequence?)

keuv-grvl:
There is no `-h` or `--help` in this software. Running it with no argument displays some help message but exit code is 255. `command -v` may be fine in this case.

Thank you for your help :smile:

daler:
Exit 255 is not that uncommon -- you can do something like in [recipes/rainbow/meta.yaml](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/rainbow/meta.yaml) or any of the UCSC tools. Try:

```yaml
test:
  commands:
    - "abundancebin &> /dev/null || [[ \"$?\" == 255 ]]"`
```

bgruening:
@keuv-grvl are you ok if I squash and merge this? I want to get rid of the removed test file.

keuv-grvl:
@bgruening OK for squash and merge

bgruening:
Sorry, forgot about this one. Please ping me next time :)

PR #2619, opened 2016-10-14
fixed bcftools plugin search path

rvalieris:

  added detect_binary_files_with_prefix build option
  to make conda automatically fix the path

bcftools plugin command has a default search path to look for plugins, and it was poiting to ``/anaconda/envs/_build/libexec/bcftools`` , this commit sets the ``detect_binary_files_with_prefix`` conda option to have conda automatically look for these paths on the binaries and update them at install time, this fixes the problem.

### DIFF
--- a/recipes/bcftools/1.3.1/build.sh
+++ b/recipes/bcftools/1.3.1/build.sh
@@ -2,5 +2,4 @@
 
 export CPPFLAGS="-I$PREFIX/include"
 export LDFLAGS="-L$PREFIX/lib"
-make plugins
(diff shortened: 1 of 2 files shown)


daler:
Good catch. Is there a quick command-line test you can add to make sure this doesn't happen in the future? E.g. does `bcftools plugin` fail before but works with this fix?

rvalieris:
bcftools plugin -lv shows the actual path that will be searched, yes I guess we could check the exit code. I am going to add a test.

PR #2630, opened 2016-10-18
Pandaseq fix

druvus:

The pandaseq recipe was passing the tests but failed to generate bioconda package. It seems it was /usr/bin/perl so I am trying to add perl-threaded as a build dep.


bgruening:
@druvus there is something in `autoconf` which has `/usr/bin/perl` hard-coded :(

druvus:
@bgruening thanks for merging. It seems this made autoconf to find the correct location of perl.

PR #2636, opened 2016-10-19
Feature/metaphlan2 rebuild

jerowe:

There is a missing annotation file that is necessary for metaphan2 to run. It is now added.


bgruening:
Thanks!

Issue #2404, opened 2016-09-26
R 3.3 migration


daler:
Opening this as a catch-call for all R 3.3-related issues.
### What's happening and why

We really want to have R and Bioconductor packages built for R 3.3. But it gets tricky.

Conda looks for a `$CONDA_R` environmental variable when building packages dependent on R. In bioconda, this is configured in [scripts/env_matrix.yml](https://github.com/bioconda/bioconda-recipes/blob/master/scripts/env_matrix.yml). When bioconda started we built everything for R 3.2 and did not anticipate the large number of R recipes that would be contributed. If we were to simply change `$CONDA_R` to "3.3" then our build system on travis-ci would time out because it would see that every single R recipe in bioconda would need to be updated.

To get around this we wrote and started using [bioconda-utils](https://github.com/bioconda/bioconda-utils) for our build system, which supports blacklists to force-disable the building of listed packages. Now we can put all R recipes on that blacklist (the aptly-named [`r-blacklist`](https://github.com/bioconda/bioconda-recipes/blob/master/r-blacklist)), change `$CONDA_R` to "3.3", and travis-ci will not timeout trying to build everything.

Now all we have to do is incrementally un-comment recipes in `r-blacklist` in topologically-sorted order, right?  But we run into more problems. CRAN and Bioconductor store only the most recent version of a package on their servers. So if we uncomment an R recipe from the blacklist, say, `bioconductor-ropls`, it is likely to fail because `bioconductor-ropls 1.4.2` [source url](http://bioconductor.org/packages/release/bioc/src/contrib/ropls_1.4.2.tar.gz)  no longer exists but [1.4.4](https://bioconductor.org/packages/release/bioc/src/contrib/ropls_1.4.4.tar.gz) does, at least at the time of this writing. So the `bioconductor-ropls` recipe needs to be fixed and updated accordingly. This process needs to be done for each of the 350+ R packages we have. 

Furthermore, much has changed in bioconda since the first R packages were added. For example we now require that any packages using gcc must depend on the conda package; we are now building on CentOS6; and we are now templating dependencies like Boost. This requires manually updating older packages.

While some of this migration can be automated, much can not (see our [bioconductor recipe docs](https://github.com/bioconda/bioconda-recipes/tree/master/scripts/bioconductor) for some autmation helpers). As a result, this migration will be an iterative process of uncommenting some recipes and then reading through the travis-ci logs to see what failed, fixing them and trying again.

[...]

PR #2644, opened 2016-10-19
Bump bioconductor-topgo build number

abretaud:

Bump the topgo build number as the 2.24.0 was not correctly pushed to bioconda channel

### DIFF
--- a/recipes/bioconductor-topgo/meta.yaml
+++ b/recipes/bioconductor-topgo/meta.yaml
@@ -6,7 +6,7 @@ source:
   url: https://bioarchive.galaxyproject.org/topGO_2.24.0.tar.gz
   md5: 633251cfa90f4e9f7346ca0b48dcc413
 build:
-  number: 0
+  number: 1
   rpaths:
     - lib/R/lib/
     - lib/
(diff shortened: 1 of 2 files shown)


lecorguille:
It is [blacklisted](https://github.com/bioconda/bioconda-recipes/blob/master/sorted-r-blacklist#L326)
See https://github.com/bioconda/bioconda-recipes/issues/2404

So until, you remove it from the list, it will not be builded

abretaud:
Thanks! I hadn't seen it (and it explains why the 2.24.0 version was not built before)

daler:
@lecorguille @abretaud dependencies need to be built for R3.3.1, too. `sorted-r-blacklist` is topologically sorted, so in general things at the top need to be built before things at the bottom. If you hit dependency issues, move up the list.

It looks like we're hitting that critical part in the DAG where s4vectors is built, so I think you'll be OK for this PR.

abretaud:
Ok, we'll see if it goes green
Any idea why the travis job is still waiting? I tried cancelling/restarting it, but still waiting (and I'm still impatient :smile: )

bgruening:
Only a few jobs can run at a time and currently Travis seems to be busy. Osx builds are taking very long at the moment. :-(

daler:
https://www.traviscistatus.com/ says they're still working through the OSX backlog.  Recently it's been taking hours for jobs to start :/

PR #2646, opened 2016-10-19
Fix hard-coded shebangs in autoconf

druvus:

This will fix the hard coded path in shebangs. Added an extra test to verify that the path is set correctly. The problem with the previous test was that the main autoconf script used as a test is a bash script while the most others are perl scripts.

### DIFF
--- a/recipes/autoconf/build.sh
+++ b/recipes/autoconf/build.sh
@@ -2,4 +2,12 @@
 
 ./configure --prefix=$PREFIX
 make
+
+sed -i.bak 's|/usr/bin/perl|/usr/bin/env perl|' bin/autom4te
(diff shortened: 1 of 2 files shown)


bgruening:
Nice!!!! There are a few PR open that suffer from this issue.

druvus:
Could someone from @bioconda/core review the PR as several recipes depends on autoconf.

bgruening:
Looks good to me! Thanks a lot!

druvus:
@bgruening sorry, it seems I forgot to add perl-threaded and got fooled by the test. Looking into this later tonight.

druvus:
I guess autoconf could be removed now when bioconda depends on conda-forge. So I should rather try to fix their recipe.

daler:
xref #2653, where autoconf is now in `build-fail-blacklist`

PR #2647, opened 2016-10-20
Adds the mm10 bsgenome from bioconductor.

jrderuiter:

This recipe adds the mm10 bsgenome from bioconductor. The recipe is essentially copied from the mm9 version, but adjusted to point to the proper files and include the proper description.


jrderuiter:
I am having many issues with building this package, as many of the bioconductor source links are no longer valid (resulting in 404s) for the current bioconductor release. Any ideas of how to address this? 

Updating recipes for the new releases is a lot of work, as every time I update a dependency that no longer works, I seem to encounter additional dependencies (of those packages) that cannot be built.

daler:
Has the organization of the bioconductor site always been like that? In the bioconductor-skeleton script I'm scraping the bioconductor site for the package URL, which uses the "release" URL. Is there somewhere it's indicates bioconductor version 3.3, 3.4 etc that we can use to fix the skeleton?

What we need is a script that can pull urls from bioconductor-* recipes, checks whether they have 3.* in their source URL, and if not, tries to figure out if one exists.

It's been taking so much time to rebuild all packages for R3.3.1 that another bioconductor release has come out in the meantime.

jrderuiter:
If you scrape the current version of bioconductor, then you will get the 'release' urls that do not contain a version number. However, scraping the 'legacy' bioconductors does give links with the version number. I just tried replacing release with the current version and that works, so the release link is just an alias to the current version. 

I don't know if this has always been the case though. I do think that the best way forward is to always include the bioconductor version and not use the release links.

We could write a script that checks for the existence of the package on the newest bioconductor and then works its way back to previous versions until it finds a working URL or a page that contains that package version and errors/warns if it can't find anything. That would be a naive approach at least.

jrderuiter:
That would focus just on fixing the urls of course, not updating packages.

daler:
Would you mind taking a stab at a script that does that? I could then incorporate that into the bioconductor-skeleton.

As for updating packages, see here for details on that: https://github.com/bioconda/bioconda-recipes/tree/master/scripts/bioconductor#updating-recipes. Basically it uses the bioconductor-skeleton with `--force` to update packages. So if bioconductor-skeleton uses the more stable 3.4 etc URLs then the updating will work better.

bgruening:
Can you rebase this PR?

jrderuiter:
Sure, I'll try later this weekend.

jrderuiter:
I tried rebasing, but the build fails on a conflict between bioconductor-bsgenome and r 3.3.1.

```
  - bioconductor-bsgenome >=1.33.5
  - r 3.3.1*
```

daler:
Yep, that's the case.

#2742 is the next PR for R that we're working on and it includes the bsgenome packages (should be ablt to add the mm10 one as well) but we're running into some issues apparently related to long package names. See that PR for details.

bgruening:
I'm going to close this one, we made some progress in rebuilding packages.

PR #2673, opened 2016-10-24
WIP: Exparna

HassanAmr:


HassanAmr:
Hi @bgruening,

Sorry fro making a pull request, but I've been stuck with this one for so long, and I need help.
When I run `./configure` I get an error that it can't find ViennnaRNA, and according to the following snippet from the file called configure:
```
if test "${with_RNA+set}" = set; then
  withval="$with_RNA"
  RNAPATHSET=1
else
  RNAPATHSET=0

fi;
if test  $RNAPATHSET = 1 ; then
        CPPFLAGS="$CPPFLAGS -I$with_RNA/include"
        CFLAGS="$CFLAGS -I$with_RNA/include"
        LDFLAGS="-L$with_RNA/lib $LDFLAGS"
fi
```
It will point to the right location if `with_RNA = $PREFIX`, and in the install instructions, it says I can simply give the option `--prefix=$PREFIX` to do so. But I don't think it is pointing to `$PREFIX`.
So, what am I doing wrong here?

martin-mann:
It might be that you have to use viennarna 1.8.5 as dependency. At least it was developed with this version.

I will check the ExpaRNA source asap if it runs also with the new Vienna package.

bgruening:
Thanks @martin-mann, changes it.
Btw. tomorrow and day after we will have a bioconda hackathon, feel free to join: https://github.com/bioconda/bioconda-recipes/issues/2277

martin-mann:
I released a new ExpaRNA version which fixes the problems you are still facing and also the VRNA <2 dependency. Now compiles with 2.2.10 without a problem.

BUT: when directly taking it from git you FIRST have to run `bash autotools-init.sh` to generate the autotools files like `configure` etc. Please update the recipe accordingly, otherwise it will directly fail again. ;)

recipes/exparna/meta.yaml
```diff
@@ -0,0 +1,27 @@
+package:
+  name: exparna
+  version: 1.0.1
+
+source:
+  git_url: https://github.com/BackofenLab/ExpaRNA.git
```
bgruening: please use https://github.com/BackofenLab/ExpaRNA/archive/1.0.1.tar.gz with checksum and test locally with our new simulate-travis.py file. See FAQs.md :)

HassanAmr: How can I merge my updated master with this branch? git merge master/exparna?
Because this branch is a few days old.

martin-mann:
Uaks.. seems to be an issue with an absolute path in the `autotools-init.sh`

It assumes perl in `/usr/local/bin`, which is not the case with conda...

Will fix the exparna release

bgruening:
@martin-mann alternatively: https://github.com/bioconda/bioconda-recipes/blob/fd57b3a52bf94a41e783424f1db3364f66be7bbe/recipes/sailfish/build.sh

martin-mann:
@bgruening think i found the issue: the conda aclocal from the automake package (which is a perl script) has a hard coded `/usr/local/bin/perl` in its first line. Thus it is incompatible to the bioconda `perl-threaded`

Please replace automake install with apt-get stuff and it should work..

bgruening:
I know that the reason I posted the link above we need to replace the `/usr/local/bin/perl` with `/usr/bin/env perl` which is the correct solution. Changing the build environment is not an option, we share this with other projects at the moment.

martin-mann:
Mhh.. why not asking the conda people to do it right?

But most likely they will "just" use their perl package..

bgruening:
@druvus is working on a proper fix. Either fixing the conda-forge one or fixing the bioconda one :)

martin-mann:
One could write a "simple" script that applies the change to all perl files within a directory and set it up via a recipe. 

That way, one can put the script as a dependency and run it in the build script to clean up globally..

Just a thought ☺

martin-mann:
@bgruening öhm. It doesnt find the aclocal in the prefix.. could you please check locally? Thanks..

bgruening:
@martin-mann not sure if OSX is supported at all. `skip: True  # [osx]` maybe?

HassanAmr:
the OSX test failed because of timeout

HassanAmr:
thanks @martin-mann and @bgruening :)

PR #2675, opened 2016-10-24
Fix channel priorities for building and in the docs.

johanneskoester:

@bioconda/core: as you know, we depend on 3 channels since some time:

| channel | dependency |
| ---------- | --------------- |
| anaconda | - |
| r | anaconda |
| conda-forge | anaconda |

Unfortunately, those channels are not disjoint. Some packages are available in at least two of them. From time to time, it can happen that those packages are binary incompatible, and hence not interchangeable when pulling dependencies (e.g. PR #2569). Currently, conda does not have a good mechanism to solve such problems.
The best we can do at the moment is to enforce an order (i.e. prioritization) of channels, that causes the least problems. My feeling is that this is the order `anaconda > r > conda-forge`. Why? Because the R channel builds on several packages that are also in conda-forge, while this is quite unlikely vice versa. We need to ensure that an r-channel dependency does not pull a binary incompatible package from conda-forge. My hope is that this ordering will help here.

This is only a temporal workaround. A real solution would be one of the following:

* ensure binary compatibility between channels, 
* support pinning of dependencies beyond the plain version
* to support channel dependencies in the conda dependency solver

This is a quite pressing issue, and I would really appreciate your help on this. In order to not fragment the discussion, only post about the workaround here.
For the general problem, please join the discussion with people from the other channels here: #2569.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -5,7 +5,7 @@ blacklists:
 docker_image: "continuumio/conda_builder_linux:5.11-5.2-0-1"
 channels:
     - anaconda
-    - conda-forge
     - r
+    - conda-forge
     - bioconda
 docker_client_version: '1.20'
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -51,8 +51,8 @@ Step 2: Setup Bioconda
 
 After installing Miniconda, you can use the ``conda`` command to setup bioconda with::
 
-   conda config --add channels conda-forge
    conda config --add channels r
+   conda config --add channels conda-forge
    conda config --add channels bioconda
 
 The ``r`` channel is added to satisfy the `R language <https://www.r-project.org/>`_
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -51,9 +51,9 @@ Step 2: Setup Bioconda
 
 After installing Miniconda, you can use the ``conda`` command to setup bioconda with::
 
+   conda config --add channels conda-forge
    conda config --add channels r
    conda config --add channels bioconda
-   conda config --add channels conda-forge
 
 The ``r`` channel is added to satisfy the `R language <https://www.r-project.org/>`_
 dependencies of some packages as well as some non-R dependencies.
@@ -61,7 +61,9 @@ Even if you don't plan on installing R packages, the `r` channel is required
 for some Bioconda packages. Similarly, the ``conda-forge`` channel contains
 many dependencies that are not strictly for bioinformatics but that are
 required for Bioconda packages to work (see more at
-`https://conda-forge.github.io/_`)
+`https://conda-forge.github.io`_).
+Be sure to add the channels in the given order, because it defines how conda handles ambiguity between channels.
+See `http://conda.pydata.org/docs/channels.html`_ for more information.
 
 Step 3: Install packages or create environments
 -----------------------------------------------


daler:
@johanneskoester sounds like the only option at the moment. Looks good to me.

daler:
@johanneskoester, @bgruening -- I just realized that later channels actually get higher priority. If I'm following your logic correctly we want R to be higher priority than conda-forge. In the docs and in travis-setup.sh we use this to give `r` channel top priority followed by `conda-forge`.

```
conda config --add channels r
```

But if I do that and then check with `conda config --get`, I see the priorities seem to be reversed according to the comments:

```
--add channels 'defaults'   # lowest priority
--add channels 'r'
--add channels 'conda-forge'
--add channels 'bioconda'   # highest priority
```

So the order should be reversed I think to be:

```
conda config --add channels r
```

Which gives

```
--add channels 'defaults'   # lowest priority
--add channels 'bioconda'
--add channels 'conda-forge'
--add channels 'r'   # highest priority
```

This could help explain issues like https://github.com/bioconda/bioconda-recipes/pull/2729#issuecomment-257187874.

Also, shouldn't `bioconda` be top priority so that we have the option of overriding other channels?

molpopgen:
Related: it would be nice to have documentation on things like {{CONDA_GSL}}*; how are those values determined and where are they defined?  In #2729, I stumbled upon that solution by randomly grepping through existing recipes.   It seems odd to me that there is an anaconda/gsl, conda-forge/gsl both at 2.1, but there's a shell variable set to 1.16, and you only discover that reading the build logs.

For addressing the GSL issue, it seems that condo-forge should be preferred over anaconda.  Otherwise, the broken anaconda/gsl will be preferred, leading to build failures, because the shell variable thing doesn't seem documented anywhere. (Please correct me if I'm wrong on that, but I've looked around quite a bit in the last day or 2).

daler:
Yep, docs are severely lacking and have taken a back seat to R stuff and build infrastructure.

For your specific question, all recipes are built for all unique combinations of the env vars listed in `scripts/env_matrix.yml`. That's where [`CONDA_GSL` is set to 1.16](https://github.com/bioconda/bioconda-recipes/blob/master/scripts/env_matrix.yml#L10).

These are exported to the env within `bioconda-utils` in a not-so-obvious manner with a lot of moving parts:

- [`config.yml` indicates an `env_matrix`](https://github.com/bioconda/bioconda-recipes/blob/master/config.yml#L1) in which CONDA_GSL is defined
- when figuring out which recipes need to be built, the filtering step attaches each [unique env to a Target object](https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/utils.py#L511) (e.g., `CONDA_GSL=1.6; CONDA_PY=27, CONDA_R=3.3.1;` etc}
- that [env is provided to the build function](https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/build.py#L312) which is either
  - sent [directly to docker](https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/build.py#L81)
  - or used to [temporarily update os.environ](https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/build.py#L94) so that conda-build sees it.
- These environmental variables are then [used over in conda-build](https://github.com/conda/conda-build/blob/9ace97969e77a10afc3c03022f8a1d0f47c2a043/conda_build/jinja_context.py#L205) to fill in the templated variables via jinja2.

molpopgen:
Ok, so then the "bioconda approved" version of GSL is 1.16.  This is important to know--I'm not sure that 2.1 can be used as a drop-in replacement (I've not checked if the soname version number is different or not).

Does that CONDA_PY mean that it is not possible to build Python packages for 3.x?

daler:
In the linked env_matrix.yaml above, the CONDA_PY var lists 27, 34, and 35. We build for them all, unless a recipe explicitly skips a particular version.

johanneskoester:
@daler, good catch. I think at the moment it would be best to have an order that ensures that conda-forge packages do not replace packages from r and default, because of the fact that r depends only on default.

Hence, we need: `r > default`, `r > conda-forge` and therefore `r > default > conda-forge`. Then, we can also put bioconda on the top. This would mean we should suggest the following sequence:

```
conda config --append channels conda-forge
conda config --add channels r
conda config --get
--add channels 'conda-forge'   # lowest priority
--add channels 'defaults'
--add channels 'r'
--add channels 'bioconda'   # highest priority
```

What do you think?

johanneskoester:
Instead, we could also provide a .condarc for download. I also recently thought about using things like pre-link to check the used channel order and complain if it is not like we need it.

daler:
@johanneskoester this channel order would not work in a case where a defaults channel package is broken but a conda-forge package works. I seem to remember this being the case before, but I'm not sure how often it actually happens. This could be solved by adding the fixed package to the high-priority bioconda channel, and in the general case I agree the channel order `bioconda > r > defaults > conda-forge` makes the most sense.

I'll make those changes.

daler:
Providing a .condarc for download would be useful for folks with no customization. We should definitely show what the .condarc channels section looks like in the docs.

A pre-link.sh seems like it could work. I'm guessing that would be inserted on the bioconda-utils side at build-time rather than requiring it in every recipe in the repo?

johanneskoester:
Yes, that would be nice. Right, the channel order is not perfect, but it seems like the only solution at the moment. The real solution is either pinning beyond plain versions or channel dependencies.

johanneskoester:
@daler, what is the status of this? I see different, inconsistent channel orders in travis.yml, github.io, and config.yaml now...

daler:
@johanneskoester you're right. Not sure what happened, but fixing now (#2862).

daler:
We should probably have a single text file somewhere to store channel order. But it's not clear how best to do that. The correct channel order needs to make it into `docs/index.rst`, `README.md`, `scripts/travis-setup.sh`, and `config.yml`. That's four different formats/syntaxes, hence the manual editing so far. But it would be good to unify this somehow.

If the canonical order is set in config.yml then it'll be annoying to convert YAML into command-line commands in bash, ReST, and Markdown. Maybe instead we have a `channel-setup.sh` containing the `conda config --add` commands that can be included into docs/index.rst as `.. include::`, included in README.md as a link to the file, and can be run directly from `travis-setup.sh`. Then we can work out the best way to get it into bioconda-utils.

Any other ideas?

johanneskoester:
Sounds good. People could even curl it to setup bioconda.

PR #2668, opened 2016-10-21
Add MiModD package recipe

wm75:
* [x] I have read the guidelines above.
* [x] This PR adds a new recipe.
* [ ] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

PR #2676, opened 2016-10-24
Move MiModD package recipe to recipes folder

wm75:

fixes #2668 which added the recipe to the root folder by mistake


wm75:
@johanneskoester ok, took me a few attempts to have zlib.h discovered at build time, but this seems to work now. You may want to investigate the test results though because two workers report bioconda_utils.utils:Nothing to be done and I don't know what the difference to the other two is.

johanneskoester:
That's fine. We have 2 workers per architecture to speed up builds of multiple packages. In your case, those workers don't have anything to build.

johanneskoester:
Thanks!

wm75:
@johanneskoester I see. Thanks for the explanation and sorry again for the trouble with the original PR.

johanneskoester:
No problem. You are not the first ;-). We should have automatic lints about that.

PR #2681, opened 2016-10-24
Trying to fix the autoconf package

druvus:

I was looking into to fix the conda-forge recipe but it is dependent on perl so I will try to fix this one instead by adding perl-threaded.


bgruening:
@druvus this looks wired: 
```
autom4te_perllibdir='..'/lib AUTOM4TE_CFG='../lib/autom4te.cfg'         ../bin/autom4te -B '..'/lib -B '..'/lib         --language M4sh --cache '' --melt ./autoconf.as -o autoconf.in
/bin/sh: ../bin/autom4te: /opt/conda/conda-bld/autoconf_1477373884041/_b_env_placehold_placehold_placeh: bad interpreter: No such file or directory
```

any idea?

bgruening:
But as far as I see this package will never get used, as there is one in the anaconda channel :(

druvus:
Agree, I vote for removing the bioconda version.  The only difference with the conda-forge version is that it depend on perl instead of perl-threaded.

It is weird as it is building on the docker image if I pre-install perl-threaded but not if perl-threaded is listed as a dependency. Seems to be some problem with conversions of paths.

PR #2686, opened 2016-10-25
delly v0.7.6

tobiasrausch:

Just fetching the static delly binary now instead of re-building it, I hope that's fine (please review  @bioconda/core). Thanks.

### DIFF
--- a/recipes/delly/meta.yaml
+++ b/recipes/delly/meta.yaml
@@ -10,6 +10,7 @@ source:
 build:
   number: 0
   skip: True # [osx]
+  binary_relocation: False
 
 test:
   commands:
--- a/recipes/delly/meta.yaml
+++ b/recipes/delly/meta.yaml
@@ -13,7 +13,7 @@ build:
 
 test:
   commands:
-    - delly --help 2> /dev/null
+    - delly --help > /dev/null
 
 about:
   home: https://github.com/tobiasrausch/delly
--- a/recipes/delly/meta.yaml
+++ b/recipes/delly/meta.yaml
@@ -8,7 +8,7 @@ source:
   md5: 8efd63cb83f8f18428bbacf8b1b1945f
 
 build:
-  number: 2
+  number: 0
   skip: True # [osx]
 
 test:
--- a/recipes/delly/meta.yaml
+++ b/recipes/delly/meta.yaml
@@ -3,21 +3,17 @@ package:
   version: '0.7.6'
 
 source:
-  fn: delly_v0.7.6_linux_x86_64bit
-  url: https://github.com/tobiasrausch/delly/releases/download/v0.7.6/delly_v0.7.6_linux_x86_64bit
-  md5: 148ed6dde55f07491138226e53198a47
+  fn: delly_v0.7.6_CentOS5.4_x86_64bit
+  url: https://github.com/tobiasrausch/delly/releases/download/v0.7.6/delly_v0.7.6_CentOS5.4_x86_64bit
+  md5: 8efd63cb83f8f18428bbacf8b1b1945f
 
 build:
-  number: 1
+  number: 2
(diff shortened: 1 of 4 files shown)


recipes/delly/meta.yaml
```diff
@@ -1,15 +1,20 @@
...
+  md5: 148ed6dde55f07491138226e53198a47
 
 build:
   number: 1
   skip: True # [osx]
 
+requirements:
+  build:
+  run:
+
```
johanneskoester: Empty requirements can be left out.

recipes/delly/meta.yaml
```diff
@@ -1,15 +1,20 @@
...
 test:
   commands:
     - delly --help 2>&1 | grep delly
```
johanneskoester: Since you are the dev of delly: does it already exit with exitcode 0 upon --help? If that is the case, the grepping can be removed and we can just do the following:
`delly --help 2> /dev/null`

Also, providing help output at stdout if imho better, because of piping into e.g. less.

If not, that's all fine for this release, but I would suggest those changes for the next, just to be more standards compliant.

recipes/delly/meta.yaml
```diff
@@ -1,15 +1,20 @@
 package:
   name: delly
-  version: '0.7.2'
+  version: '0.7.6'
 
 source:
-  git_url: https://github.com/tobiasrausch/delly.git
-  git_rev: v0.7.2
+  fn: delly_v0.7.6_linux_x86_64bit
+  url: https://github.com/tobiasrausch/delly/releases/download/v0.7.6/delly_v0.7.6_linux_x86_64bit
```
johanneskoester: Has the binary been built on CentOS <= 6, or is it including a static copy of libc? If not, we would need to rebuild here, because that is the minimal libc version supported by bioconda.

tobiasrausch: Thanks, changed to CentOS 5.4 static binary.

recipes/delly/meta.yaml
tobiasrausch: Yes, exitcode is 0 upon --help and help message is on stdout.

recipes/delly/meta.yaml
```diff
@@ -3,21 +3,17 @@ package:
...
-  fn: delly_v0.7.6_linux_x86_64bit
-  url: https://github.com/tobiasrausch/delly/releases/download/v0.7.6/delly_v0.7.6_linux_x86_64bit
-  md5: 148ed6dde55f07491138226e53198a47
+  fn: delly_v0.7.6_CentOS5.4_x86_64bit
+  url: https://github.com/tobiasrausch/delly/releases/download/v0.7.6/delly_v0.7.6_CentOS5.4_x86_64bit
+  md5: 8efd63cb83f8f18428bbacf8b1b1945f
 
 build:
-  number: 1
+  number: 2
```
johanneskoester: Build number should be 0. You only need to increase it when you update a build that already exists on anaconda.

recipes/delly/meta.yaml
johanneskoester: Great, can you please simplify the test as suggested?

tobiasrausch: It's `delly --help > /dev/null` now, thanks.

tobiasrausch:
One check failed because of the patchelf issue:

https://github.com/conda/conda-build/issues/1411

recipes/delly/meta.yaml
johanneskoester: No problem ;-)!

johanneskoester:
Right. Since your binaries should not contain any prefixes that need to be fixed, you can disable patchelf entirely (see [here](http://conda.pydata.org/docs/building/meta-yaml.html#binary-relocation)).

PR #2690, opened 2016-10-25
Feature/rpkmforgenes

percyfal:

@bioconda/core New recipe


recipes/rpkmforgenes/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  name: "rpkmforgenes"
+  version: "1.0.1"
+
+build:
+  number: 0
+
+source:
+  git_url: https://github.com/danielramskold/S3_species-specific_sequencing
+  git_rev: 1.0.1
+  md5: foo
```
johanneskoester: md5 foo? This is quite unlikely, isn't it ;-)? We have recently changed our policy to enforce non-git urls and md5 sums for all recipes (because galaxy now provides automatic mirroring for all our sources). You can use the release tarball from github instead.

recipes/rpkmforgenes/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+requirements:
+  build:
+    - python >=2.7,<3
```
johanneskoester: This interferes with pinning (a python specific caveat). You can use `build->skip` instead.

recipes/rpkmforgenes/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - pysam
+    - numpy
+  run:
+    - python >=2.7,<3
+    - pysam
+    - numpy
+test:
+  requires:
```
johanneskoester: The requirements can be omitted. They are handled automatically.

recipes/rpkmforgenes/meta.yaml
percyfal: Doh! I mistakenly in magit pushed to master and then ended up with the wrong version in the feature branch. Should be fixed by now.

bgruening:
Look great now! @percyfal thanks for your contribution!

PR #2696, opened 2016-10-25
Add --ignore-installed

bgruening:

adds the --ignore-installed option to pip during docs build.


daler:
Thanks Bjoern!

bgruening:
Mh, seems this does not fix it. Maybe a `pip install auxlib` can help? I see you have a branch so I will not mess with it, let me know when you go to bed. An other idea would be to not use pip but conda for the build-docs setup.

PR #2700, opened 2016-10-26
update rMATS to 3.2.5

BioNinja:


recipes/rmats/meta.yaml
```diff
@@ -16,6 +16,7 @@ requirements:
     - python
     - numpy
     - scipy
+    - pysam
   run:
     - python
     - numpy
```
BioNinja: Yes, you are right. Is the way to add pysam uncorrected?

PR #2718, opened 2016-10-27
Mcl blastfix

druvus:

Enable blast parsing for mcl. Also added an extra test to check for one of the new tools.

### DIFF
--- a/recipes/mcl/build.sh
+++ b/recipes/mcl/build.sh
@@ -2,8 +2,6 @@
 mkdir -p $PREFIX/bin
 ./configure --prefix=$PREFIX --enable-blast
 make
-sed -i.bak 's|/usr/local/bin/perl -w|/usr/bin/env perl|' src/alien/oxygen/src/mcxdeblast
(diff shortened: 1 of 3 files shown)


recipes/mcl/build.sh
```diff
@@ -1,8 +1,9 @@
 #!/bin/bash
-
-./configure --prefix=$PREFIX
 mkdir -p $PREFIX/bin
-
+./configure --prefix=$PREFIX --enable-blast
+make
+sed -i.bak 's|/usr/local/bin/perl -w|/usr/bin/env perl|' src/alien/oxygen/src/mcxdeblast
+sed -i.bak 's|/usr/local/bin/perl|perl|' scripts/clxdo
```
bgruening: I haven't look at this but is it intended to be `perl` and not `/usr/bin/env perl`

druvus:
It is a couple of perl one-liners in clxdo. I missed to do a global substitute so I am changing to use a patch file to make the changes more transparent.

Issue #2475, opened 2016-10-01
errors with gcc and gfortran from defaults channel


daler:
I've been hitting some issues with building R packages that I think is related to the gcc package from defaults channel.

How to reproduce:

``` bash
docker run -it --net="host" -v `pwd`:/bioconda-recipes condaforge/linux-anvil /bin/bash -c \
"cd bioconda-recipes; conda build -c r -c bioconda recipes/r-ash"
```

fails with the following message:

```
gfortran   -fpic  -I/opt/conda/conda-bld/r-ash_1475351407071/_b_env_placehold_placehold_placehold_pl/include -L/opt/conda/conda-bld/r-ash_1475351407071/_b_env_placehold_placehold_placehold_pl/lib  -c ash.f -o ash.o
f951: internal compiler error: Illegal instruction
```

The information I've found makes me think it's an issue with how it was compiled, e.g.,
- https://trac.macports.org/ticket/41364
- https://bugzilla.redhat.com/show_bug.cgi?id=222371
- https://github.com/Homebrew/legacy-homebrew/issues/18761

Before I open an issue with conda and/or conda-forge, anyone else know another way of fixing this? ping @bioconda/core

bgruening:
I was also seeing it and have no clue how to fix it. A few ideas ... Does the GCC from r work? Is this a regression in the latest build and maybe a previous build works?

daler:
No luck. I don't see a gcc in r channel, and gcc 4.8.2 (the earliest available in defaults channel) has the same error.

bgruening:
You are right, libgcc is in `r`:

`libgcc                       4.8.4                         0  r ` 

I guess the we should raise this issue upstream. Something is broken here :(

daler:
xref: https://github.com/conda/conda-build/issues/1424

mingwandroid:
libgcc in `r` is at version 4.8.5:

`libgcc:     4.8.5-1   r`

.. but IMHO it shouldn't be there at all.

daler:
@bgruening this seems to be CPU architecture related. Locally on my machine it still fails, but on travis-ci it's working (https://travis-ci.org/bioconda/bioconda-recipes/builds/171435307). I have a Xeon X5482, so if you're running on something around the same age it might be why you have issues too.

Being able to build these fortran-compiled packages opens up a lot of the R package DAG that can be built!

epruesse:
outdated

PR #2727, opened 2016-10-28
try a package with fortran error

daler:

@bgruening gfortran is in gcc, but we specify llvm for osx. LLVM doesn't have gfortran, so we get errors like this:

https://travis-ci.org/bioconda/bioconda-recipes/jobs/171435317#L277

I don't know enough about gcc/llvm to know what kinds of issues having both would have. Any ideas for handling this?

xref #2475

### DIFF
--- a/recipes/r-multitaper/meta.yaml
+++ b/recipes/r-multitaper/meta.yaml
@@ -16,11 +16,10 @@ build:
 requirements:
   build:
     - r
-    - gcc #[linux]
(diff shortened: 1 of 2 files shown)


bgruening:
All what I can say is let's use gcc on OSX for those packages. Please read more about this here: https://github.com/conda-forge/staged-recipes/issues/1193

The goal is to have gcc packaged and used also on OSX, but not sure when this will happen.

PR #2728, opened 2016-10-28
Gsl

molpopgen:
* [ X] I have read the guidelines above.

This recipe is for GSL 2.2.  The current conda recipe for 2.1 gives "incompatible library" errors for both Linux and OS X when you try linking.  I figure a consistently-compiled version for bioconda could be handy.

I have also gone ahead and checked that installing GSL 2.2 on OS X when compiled with clang results in a library that can be linked to when using Conda/GCC on OS X.

### DIFF
--- a/recipes/gsl/build.sh
+++ b/recipes/gsl/build.sh
@@ -1,0 +1,13 @@
+if [[ "$(uname)" == "Darwin" ]]; then
+  # CC=clang otherwise:
+  # error: ambiguous instructions require an explicit suffix
+  # (could be 'filds', or 'fildl')
+  # https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66509
+  export CC=clang
+fi
+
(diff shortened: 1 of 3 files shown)


bgruening:
I guess this should more into conda-forge. @johanneskoester @daler ?

molpopgen:
@bgruening Argh--I didn't check there first.  They have 2.1.  I guess I need to figure out how to list a dependency from condo-forge in a yaml file for bioconda.  I've not been able to find an example of a dependency from another channel anywhere.

bgruening:
@molpopgen bioconda now depends on conda-forge. Just add this channel to your global channel list as described in our Guidelines. Or simply `conda build ./foo -c conda-forge`

bgruening:
I cancelled the builds, hope this is ok.

molpopgen:
@bgruening No problem.  I'm checking that conda-forge/gsl does indeed work.

PR #2729, opened 2016-10-28
Fwdpp

molpopgen:
- [X ] I have read the guidelines above.
- [X] This PR adds a new recipe.

This installs a c++11 template library + a binary that reports its version.  All the dependencies are just for compiling & running the test suite.

### DIFF
--- a/recipes/fwdpp/meta.yaml
+++ b/recipes/fwdpp/meta.yaml
@@ -1,0 +1,36 @@
+about:
+    home: 'https://www.github.com/molpopgen/fwdpp'
+    license: "https://raw.githubusercontent.com/molpopgen/fwdpp/master/LICENSE"
+    summary: "A C++ template library for forward-time population genetic simulation. "
+    doi: 10.1534/genetics.114.165019
+
+requirements:
+  build:
+    #Must build with GCC even on OS X
+    #b/c libsequence was built that way.
+    #O/W we end up with "missing symbol"
+    #errors.
+    #These requirements are solely 
+    #to get the examples/unit tests
+    #compiled.
+    - gcc  
+    - zlib
+    - gsl {{CONDA_GSL}}*
+    - libsequence
+    - boost >=1.60.0
+
+build:
+  number: 0
+package:
+  name: fwdpp
+  version: '0.5.3'
+test:
+  commands:
+    - export LD_LIBRARY_PATH=$PREFIX/lib
+    - if [ `uname` == 'Linux' ]; then make check; fi
+    - if [ `uname` == 'Darwin' ]; then for i in $(find testsuite -type f -perm +111|grep -v integration); do $i -r detailed; done; fi
+    - fwdppConfig --version | grep 0.5.3 > /dev/null
+source:
+  fn:  0.5.3.tar.gz 
+  url: http://github.com/molpopgen/fwdpp/archive/0.5.3.tar.gz
+  md5: 09a00eb8fa1d54c953b5cb4e52a85a1c
--- a/recipes/fwdpp/meta.yaml
+++ b/recipes/fwdpp/meta.yaml
@@ -26,9 +26,6 @@ package:
   version: '0.5.3'
 test:
   commands:
-    - export LD_LIBRARY_PATH=$PREFIX/lib
-    - if [ `uname` == 'Linux' ]; then make check; fi
-    - if [ `uname` == 'Darwin' ]; then for i in $(find testsuite -type f -perm +111|grep -v integration); do $i -r detailed; done; fi
     - fwdppConfig --version | grep 0.5.3 > /dev/null
 source:
   fn:  0.5.3.tar.gz 
--- a/recipes/fwdpp/build.sh
+++ b/recipes/fwdpp/build.sh
@@ -1,5 +1,11 @@
 #!/bin/bash
 CPPFLAGS="-I$PREFIX/include $CPPFLAGS" LDFLAGS="-L$PREFIX/lib $LDFLAGS" ./configure --prefix=$PREFIX
 make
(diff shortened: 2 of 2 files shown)


molpopgen:
@bgruening Related to #2728, the Travis build setup is attempting to pull the GSL dependency for this from anaconda, and not from conda-forge.  This causes ./configure to fail because conda/gsl is broken (but I do know that the conda-forge version works).

molpopgen:
@bioconda/core -- this is ready for review.

I noticed some odd things with the Linux build system:

* GSL is a build dependency.
* conda-forge GSL is version 2.1.  If "gsl" is listed as a build dependency, conda/gsl is installed instead of conda-forge/gsl on the Travis/Docker build setup.  The former is broken, leading to failures.
* If you list gsl {{CONDA_GSL}}*, then it seems like 1.16 is installed, but not clear from where.

This behavior all differs from what "conda build recipe" does on Linux, where it correctly pulls GSL from conda-forge. 

So, I think there's an issue somewhere....

bgruening:
Can you check your order of Conda channels and compare to the order we have agreed here. Channel order is important these days.

molpopgen:
Not sure what you mean.  Isn't it up to the docker/Travis setup to do that correctly?  The guidelines don't mention any way to affect this stuff via meta.yaml.

molpopgen:
@bioconda/core -- based on posts to other discussions, it appears that this is working as intended.  Please review for any changes needed.

recipes/fwdpp/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+    #These requirements are solely 
+    #to get the examples/unit tests
+    #compiled.
+    - gcc  
+    - zlib
+    - zlib
+    - gsl {{CONDA_GSL}}*
+    - libsequence
+    - boost >=1.60.0
+#  run:
```
bgruening: Just to be sure, there is no boost, libgcc or gsl need at runtime?

recipes/libsequence/build.sh
```diff
@@ -2,5 +2,7 @@
 
 CPPFLAGS="-I$PREFIX/include $CPPFLAGS" LDFLAGS="-L$PREFIX/lib $LDFLAGS" ./configure --prefix=$PREFIX
 make
+export LD_LIBRARY_PATH=-L$HOME/$PREFIX/lib
```
bgruening: This looks strange and is not build. Can you revert this change?

recipes/fwdpp/meta.yaml
```diff
@@ -0,0 +1,35 @@
+about:
+    home: 'https://www.github.com/molpopgen/fwdpp'
+    license: "https://raw.githubusercontent.com/molpopgen/fwdpp/master/LICENSE"
+    summary: "A C++ template library for forward-time population genetic simulation. (doi:  10.1534/genetics.114.165019)"
```
bgruening: can you add a new key with `doi`?

recipes/fwdpp/meta.yaml
molpopgen: Correct--see comments at the top.  This is a C++11 template lib.  The build dependencies are simply for testing.

recipes/libsequence/build.sh
molpopgen: I believe this is required for tests to run.  Without it, previous pushes to this PR failed.

recipes/fwdpp/meta.yaml
molpopgen: Yes.  I'm going to leave the LD_LIBRARY_PATH, though, for reasons mentioned above.  OS X/conda in particular doesn't "just find" the runtime libs, IME.

recipes/libsequence/build.sh
bgruening: I have two problems with it. I think the syntax is wrong, shouldn't this be `=-L$HOME/:$PREFIX/lib`? And if you don't increase the build number of a recipe this will not be rebuild so your recipe is not using this modified version as it was not build.

recipes/fwdpp/meta.yaml
bgruening: Cool! Thanks a great idea for a test! Thanks!

recipes/libsequence/build.sh
molpopgen: The syntax is based on following other recipes, which use $PREFIX, not $HOME/prefix.  Hopefully no one is installing runtime libs in ~/ :).  I've pushed new changes moving this stuff into the build: field in meta.yaml and upped the build no. to 1.

molpopgen: PS, if you get a second, you may cancel builds  10798 and 10799.

bgruening: @molpopgen You increased the wrong recipe. `libsequence` was never rebuild as far as I can tell. So these changes here do not have any effect. Please try to remove the libsequence changes. The build number from fwdpp can be at 0 as it is the first version of this recipe.

molpopgen: Oh--sorry.  This is really confusing--you are referring to files in a different PR....

molpopgen: @bgruening Even more confusing: recipes/libsequence/build.sh does not contain the lines that you are referring to.  Can you clarify what you're after here?  I updated this in fwdpp, though...

bgruening: According to github and this PR here you have modified three files one it `recipes/libsequence/build.sh`

bgruening:
@molpopgen this looks great now. Feel free to merge as soon as it turns green or I will merge it after some hours of sleep.

bgruening:
Great, this looks very clean now. Thanks a bunch!

PR #2734, opened 2016-10-29
Trying to build graphmap for osx using gcc

druvus:
Checking if I can get graphmap to build for osx using gcc.


bgruening:
Increasing the build number, otherwise it will not be tested.

bgruening:
:+1:

PR #2735, opened 2016-10-29
R21

bgruening:


bgruening:
@daler I guess this is a bug in one of the subprocess call from bioconda-utils? 

ERROR bioconda_utils.build:build(109): STDOUT+STDERR: bash: cannot set terminal process group (-1): Inappropriate ioctl for device

daler:
That's actually from the condaforge/linux-anvil container, and I think it's a side effect of how they're running `tini`. It shows up no matter what you run in it. It's harmless as far as I can tell. But guess we could have bioconda-utils strip that out of the stderr stream if you think it would be confusing.

daler:
@bgruening by the way try out the new `--quick` option in bioconda-utils for local testing. Speeds up filtering considerably and pays attention to newly un-blacklisted recipes so we can use it for all this R work.

bgruening:
@daler thanks that works nicely!

PR #2739, opened 2016-10-30
adding gsl and bioconductor-dirichletmultinomial

jerowe:
- [ ] I have read the guidelines above.

https://www.gnu.org/software/gsl/
http://bioconductor.org/packages/release/bioc/html/DirichletMultinomial.html


bgruening:
You can rerun travis if you are a commiter, you are isn't it?

bgruening:
Otherwise a general trick for travis is to close and reopen a PR :)

recipes/gsl/meta.yaml
```diff
@@ -0,0 +1,24 @@
+package:
```
bgruening: Do you need gsl 2.2? gsl is available in conda-forge and I think the correct way would be to upgrade this version.

recipes/bioconductor-dirichletmultinomial/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+    - lib/R/lib/
+    - lib/
+requirements:
+  build:
+    - bioconductor-biocgenerics
+    - bioconductor-iranges
+    - bioconductor-s4vectors
+    - r
+  run:
+    - bioconductor-biocgenerics
```
bgruening: This one is compiling, isn't it? So you need to add gcc/llvm?

jerowe:
I didn't realize gsl was in conda-forge. I removed it from the PR.

I suppose I could be smart about restarting travis, but I always just push a new commit. ;-)

jerowe:
@bgruening , any idea why this is failing?

daler:
@jerowe see https://github.com/bioconda/bioconda-recipes/pull/2729#issuecomment-257187874, @molpopgen is running into GSL issues as well . Maybe `{{CONDA_GSL}}` would work?

molpopgen:
Also see #2590.  I think the Linux builds are not pulling from conda forge.  GSL diagnoses the problem because conda version is broken, causing failures seen here.

jerowe:
@daler and @molpopgen , thanks for the input. I am trying now with CONDA_GSL.

jerowe:
FYI, when I build on my local docker image I get

gsl:                       1.16-1           r  

So it looks like it is being pulled from r.

jerowe:
It works with {{CONDA_GSL}}

Thanks!

bgruening:
Awesome!

PR #2742, opened 2016-10-30
r23

bgruening:


bgruening:
@daler do you have any idea why OSX is failing here? https://travis-ci.org/bioconda/bioconda-recipes/jobs/171764189

The only error I can spot is the 80 character limit.

daler:
Yep, looks like it. I guess that package name is too long since others seem ok?

bgruening:
It's building for Linux ...

daler:
True, and everything should be installing into /anaconda on either OS. Maybe there's something osx-specific in conda code? Does artificially shrinking the package name help?

bgruening:
It still fails even with a short package-name.

bgruening:
xref: https://github.com/conda/conda-build/issues/1484

bgruening:
@daler we are currently not building packages in parallel, we should at some point but for now do you think we can add `--no-build-id` into the build process until the rebuilding in conda-forge and Co is done?

daler:
I didn't even know parallel builds were an option, this solution seems worth a shot.

bgruening:
Since conda-build-2 we can build packages in parallel, thinking about building dags in parallel ....
Should I go ahead and hardcode `--no-build-id` in bioconda-utils?

bgruening:
@daler https://github.com/bioconda/bioconda-utils/pull/25

daler:
Go for it. I think we might be limited to a single travis-ci processor, so parallel might not help us.

PR #344, opened 2015-12-02
Feature/homer

percyfal:
Add homer, software for motif discovery and next-gen sequencing analysis.

Here is another unusual package where advice/input would be appreciated. I provide some details of the recipe rationale here for future reference.

 Homer is installed by downloading a script, configureHomer.pl, which then is invoked with parameter -install to install some ~130 scripts. Upon installation, the configure script _hard codes_ the path to the installation directory, which means building the scripts in the recipe is out of the question. My current solution is to place the install script in $PREFIX/share/homer and link it to $PREFIX/bin. Furthermore, in $PREFIX/share/homer, for all ~130 as-of-yet uninstalled scripts, the recipe creates a corresponding script wrapper which is linked to $PREFIX/bin. When run, the script produces an error message with information on how to actually install homer using configureHomer.pl. The links need to be present prior to running the configuration script.

That's the best I could think of now.

PR #2748, opened 2016-10-31
Homer, Software for motif discovery and next-gen sequencing analysis

BioNinja:

add Homer,
Software for motif discovery and next-gen sequencing analysis

### DIFF
--- a/recipes/homer/build.sh
+++ b/recipes/homer/build.sh
@@ -1,0 +1,169 @@
+#! /bin/bash
+
+# ls -1 bin/ | grep -v old | sed -e "s/\*/ \\\/g"
+binaries="\
+GenomeOntology.pl \
+HomerConfig.pm \
+SIMA.pl \
+Statistics.pm
+addData.pl \
+addDataHeader.pl \
+addGeneAnnotation.pl \
+addInternalData.pl \
+addOligos.pl \
+adjustPeakFile.pl \
+adjustRedunGroupFile.pl \
+analyzeChIP-Seq.pl \
+analyzeHiC \
+analyzeRNA.pl \
+analyzeRepeats.pl \
+annotateInteractions.pl \
+annotatePeaks.pl \
+annotateRelativePosition.pl \
+annotateTranscripts.pl \
+assignGeneWeights.pl \
+assignGenomeAnnotation \
+assignTSStoGene.pl \
+batchAnnotatePeaksHistogram.pl \
+batchFindMotifs.pl \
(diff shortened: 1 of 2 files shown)


percyfal:
@BioNinja @bioconda/core This pull request looks very similar to the one I submitted a year ago and that we then decided to leave out due to issues with the build script; see PR #344, in particular the comments. I just wanted to point this out as I think the current build script should be improved. I don't recall all the details right now as this was some time ago.

daler:
@BioNinja @percyfal agreed, I think this recipe at least needs tests to see if `findMotifs.pl` works after running `configureHomer.pl`.

Did you ever work out whether using `sed` on all the scripts after they're processed by `configureHomer.pl`?

BioNinja:
It's the same recipe with @percyfal，I've test findMotifs.pl,it looks find for me on oxs, if this does not work for others, I think we should remove for now，sorry for the mistakes，@daler

BioNinja:
I don't know why conda build works for homer now

percyfal:
It's not that it didn't work - it's that the build doesn't actually include the scripts, just the install script, which then is called the first time any homer program is called. Alternatively, one could run the install script during the build, use sed to modify the paths and only include the homer scripts themselves (if I recall correctly, homer downloads large database files which would make the package prohibitively large).

percyfal:
@BioNinja no worries, I had a look at your recipe as I was curious as to how you solved my initial issues :-D

daler:
No worries @BioNinja. When you say findMotifs.pl works on OSX, do you mean before or after running configureHomer.pl?

@percyfal do you know if the databases required for all tools? And is there an alternative method to download them?  If databases are not required, then one option would be to, in `build.sh`,

- allow configureHomer.pl to do its thing
- fix the paths when it's done
- delete the databases it downloaded
- provide an extra script, say, download_homer_database.pl, for downloading them later to the installed path or ideally to some user-defined path

This latter part might not be directly supported by Homer, but perhaps you could fix the scripts to use an environment variable like HOMER_DATABASES that is set by download_homer_database.pl.

BioNinja:
I've test findMotifs.pl after running configureHomer.pl. And it will copy all homer scripts to .../mininconda/bin directory. But I have to download homer database mannally. for example configureHomer.pl -install hg19. @daler

daler:
OK that's good to know. Can you try running configureHomer.pl within the build.sh script and see how it goes? Ideally a user should be able to download a package and start using it immediately rather than do a separate install step.

PR #2754, opened 2016-10-31
Fix bunwarpj command and remove invalid version directory

gregvonkuster:


recipes/fiji/20151122/build.sh
```diff
@@ -1,32 +0,0 @@
-#!/bin/bash
-set -eu -o pipefail
-
```
gregvonkuster: @bgruening my initial commit on this upgraded recipe had a typo in the subdirectory name - it was 20151122 and it should have been 20151222.  I fixed the type in a second commit, but I must have forgotten to delete the subdirectory with the bad name.  Sorry about the mess!

Also, sorry about forgetting the build number increase - I've fixed that now as well.

bgruening:
That explains a lot! Thanks!

PR #2759, opened 2016-11-01
update channel order, see #2675

daler:

Update the channel order in bioconda-recipes and point to the bioconda-utils commit with the same channel order


bgruening:
Superb!

PR #2765, opened 2016-11-01
Adding quorum

colinbrislawn:

This works fine with `conda build` but I should probably also test with `./simulate-travis.py`...


colinbrislawn:
I accidentally included an older recipe along with this PR. Doh!

This PR now just includes Quorum, as intended.

recipes/quorum/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+pushd $SRC_DIR
```
bgruening: Can be removed.

recipes/quorum/meta.yaml
```diff
@@ -0,0 +1,30 @@
+build:
+  number: 1
```
bgruening: 0

recipes/quorum/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+  build:
+    - gcc
+    - autoconf
+    - automake
+    - libtool
+    - pkg-config
+    - yaggo >=1.5.8
+    - jellyfish
+  run:
+    - libgcc
```
bgruening: libgcc  # [linux]

recipes/quorum/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+package:
+  name: quorum
+  version: "1.1.0"
+source:
+  fn: quorum-1.1.0.tar.gz
+  url: https://github.com/gmarcais/Quorum/releases/download/v1.1.0/quorum-1.1.0.tar.gz
+  md5: 14cc37310ff81b447148797afe551fe9
+requirements:
+  build:
+    - gcc
```
bgruening: gcc  # [linux]
llvm  # [osx]

recipes/quorum/build.sh
```diff
@@ -0,0 +1,9 @@
...
+
```
bgruening: Can you add: `sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoreconf` hopefully it will make it work.

colinbrislawn:
Thank you for helping me troubleshoot this build (and also the autoconf package). I really appreciate your guidance.

recipes/quorum/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+source:
+  fn: quorum-1.1.0.tar.gz
+  url: https://github.com/gmarcais/Quorum/releases/download/v1.1.0/quorum-1.1.0.tar.gz
+  md5: 14cc37310ff81b447148797afe551fe9
+requirements:
+  build:
+    - gcc # [linux]
+    - llvm # [osx]
+    - autoconf
+    - automake
```
bgruening: you also need perl-threaded as build dependency

colinbrislawn: Added!

colinbrislawn:
Looks like autoreconf is still not working on linux.
`autoreconf: failed to run aclocal: No such file or directory`
This is strange because `automake` installed fine.

Also, it looks like it's getting jellyfish from conda-forge, while it should get the newer version from bioconda.
```
configure: error: Package requirements (jellyfish-2.0) were not met:
...
    jellyfish-0.5.6            |           py35_0         228 KB  conda-forge
```

colinbrislawn:
Let's see if the new `autoconf` helps with this...

recipes/quorum/build.sh
```diff
@@ -0,0 +1,6 @@
+sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoreconf
```
bgruening: We fixed autoconf, can you try to remove this and rebase against latest master?

colinbrislawn:
@bgruening I've removed that sed command, as requested. 

My git skills are not that great, and It's possible that I may not have synced this up with master properly. Your guidance is really helpful.

bgruening:
Please try the following:

```
git checkout master
git remote add upstream https://github.com/bioconda/bioconda-recipes.git
git pull upstream master # getting latest changes from this repo, not your fork
git checkout adding-Quorum # go back to your branch
git rebase master # rebase against the lastest master
git push origin adding-Quorum -f # force push your rebase 
```

colinbrislawn:
Thank you for your feedback! I've rebased my branch and updated this PR. Let's see how this works...

colinbrislawn:
Still not working... I've [reached out to the quorum devs](https://github.com/gmarcais/Quorum/issues/1) because I think this may have something to do with linking to jellyfish.

Any ideas?

colinbrislawn:
My colleague and I can build this locally, but only once we set this as a local path:

```
export JELLYFISH2_0_CFLAGS=-I/home/me/anaconda2/pkgs/jellyfish-2.2.6-0/include/jellyfish-2.2.6
export JELLYFISH2_0_LIBS=/home/me/anaconda2/pkgs/jellyfish-2.2.6-0/lib/libjellyfish-2.0.so
```

We can set these paths in `build.sh`, but should we generate these using the program?

bgruening:
setting it to `xport JELLYFISH2_0_CFLAGS=-I$PREFIX/include/jellyfish-2.2.6` and the other is not enough?

colinbrislawn:
Ah, using `$PREFIX` works really well! Thank you.

I've got it working, but only with using `gcc` for both Linux and OSX. Is this OK, or should I try to get it working with `llvm` like a 'normal' OSX build?

bgruening:
Ideally, we should get it working with `llvm`. But we had already a few recipes where this was not possible and in this case `gcc` is ok on OSX:

colinbrislawn:
Thanks for the feedback. This should now build using `gcc` on both Linux and OSX.

Should I pursue a working `llvm` build or is this acceptable?

bgruening:
I'm fine with this. But I'm not a OSX users, let's ask our famous @bioconda/osx subgroup :)

acaprez:
Looks like building on OS X is a known problem with the package: https://github.com/gmarcais/Quorum/issues/2.  A bit of searching of that error leads down the path of Clang vs GCC C++ compatibility issues, which IMO are best left to the package devs to resolve.  So using GCC for both platforms seems reasonable to me.

colinbrislawn:
The lib file name is slightly different on OSX and linux.

```
linux: libjellyfish-2.0.so
macOS: libjellyfish-2.0.dylib
```

What's the recommended way to accommodate this in the `build.sh` file?

PR #2771, opened 2016-11-02
Update: biobambam, snpeff

chapmanb:

- biobambam fix for empty MC tags in bamsormadup
- snpeff: update to latest development version with support for
  canonical transcript specification

### DIFF
--- a/recipes/snpeff/meta.yaml
+++ b/recipes/snpeff/meta.yaml
@@ -5,15 +5,16 @@ about:
 
 package:
   name: snpeff
-  version: '4.3'
+  version: '4.3b'
 
 build:
-  number: 3
+  number: 0
   skip: False
 
 source:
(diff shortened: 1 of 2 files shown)


recipes/snpeff/meta.yaml
```diff
@@ -5,15 +5,15 @@ about:
...
 
 build:
-  number: 3
+  number: 0
   skip: False
 
 source:
-  fn: snpEff_v4_3_2.zip
-  url: http://downloads.sourceforge.net/project/snpeff/snpEff_v4_3_core.zip
+  fn: snpEff_v4_3b.zip
```
bgruening: can you please add a checksum (next time is probably enough)

bgruening: Thanks!

PR #2784, opened 2016-11-02
remove all bat files

bgruening:

We do not support Windows, so let's remove these files and don't confuse new contributors.


bgruening:
ping @johanneskoester @daler

colinbrislawn:
Do we state that windows is not supported? I we should do that...

(I know that these recipes are built by Travis-CI and Travis-CI only builds on Linux and OSX. However, python and conda are cross-platform so defining our scope is helpful, especially for new users.)

bgruening:
@colinbrislawn Oh good point! Do you mind updating the readme accordingly?

johanneskoester:
Never mind, I'll do that now!

johanneskoester:
See PR #2790.

PR #2793, opened 2016-11-03
Building autoconf with fixed shebang

druvus:

Removing autoconf from the blacklist and changing shebang replacements and dependencies. 

Builds cleanly on the local image so I am crossing my fingers now.

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1,2 +1,1 @@
-recipes/autoconf
 recipes/gvcftools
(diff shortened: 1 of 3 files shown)


bgruening:
Thanks a lot! Let's see if we can clean other PR a little bit.
@druvus any plan to get this into conda-forge as well?

druvus:
@bgruening Something seems to be weird here. I don't think it actually got tested even if I bumped build number and removed from blacklist. Currently only on a mobile so I can't look into it until later tonight.

daler:
@druvus @bgruening I just looked into this. It's not building because bioconda-utils filters out a recipe if the version is found in any of the configured channels (bioconda, conda-forge, r, defaults). Since the version 2.69 is the same as that in conda-forge, it is not built.

PR #2801, opened 2016-11-03
Quick

daler:


daler:
@bgruening @johanneskoester this PR uses https://github.com/bioconda/bioconda-utils/pull/29. It adds `--quick` and it checks only the bioconda channel for existing recipes.

## `--quick`
Filtering takes only <1 min rather than 5. Instead of all 1500+ recipes being sent to `filter_recipes`, only recipes that:
- are different from the master branch on github
- have been newly removed from blacklists

are sent for filtering.

The downside is that if a recipe was not edited or changed it will not be built. I think this would only matter if a recipe was deleted from anaconda, which hopefully won't be an issue.

## only bioconda channel
Only check the bioconda channel for existing recipes. This makes bioconda-utils actually behave as discussed in https://github.com/bioconda/bioconda-recipes/pull/2675#issuecomment-257527120, where we let bioconda override existing recipes in other channels. Allows building of, e.g., bioconda-specific versions of autoconf, xref #2793.

## autoconf
@druvus, I had to put `autoconf` back on the `build-fail-blacklist`. Seems like it still needs work, sorry about that.

daler:
@bgruening @johanneskoester aaaand this doesn't work. Removing in #2816.

daler:
@bgruening #2813 needs a re-run, I'll take care of it.

johanneskoester:
Hi, sorry for the late reply. Are you still working on this? If so, one question: what about recipes where we changed the underlying parts in the env_matrix, e.g. a new R or Boost version. Will they be rebuild and uploaded or not?

daler:
After removing the `--quick` from master branch, I'm putting this aside for now. While we save a few minutes per worker, I'm not sure it's worth the issues that it would cause, like the one you bring up: as currently implemented, using --quick would ignore cases when env_matrix changes. You need the full recipe rendering to figure that out (that's the slow step; it includes the jinja2 templating).

--quick is still useful for local testing when you know exactly what changed, I just don't think it's a good idea for using in all travis-ci builds.

johanneskoester:
Ok, thanks!

PR #2807, opened 2016-11-03
jemalloc: added tests

vodkatad:

### DIFF
--- a/recipes/jemalloc/meta.yaml
+++ b/recipes/jemalloc/meta.yaml
@@ -3,7 +3,7 @@ package:
   version: 3.6.0
 
 build:
-  number: 0
+  number: 1
   detect_binary_files_with_prefix: true
   skip: True # [osx]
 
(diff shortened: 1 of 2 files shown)


bgruening:
I really :heart: this test!

druvus:
I think jemalloc would build for osx using gcc. I was looking into it as a dependency for another recipe.

vodkatad:
Sorry I've got no osx available for testing that locally :\

bgruening:
You can simply remove the skip line and see what travis is saying :)

druvus:
You need to update the test to look like this to handle different suffix between linux and osx.

<pre>
test:
  files:
      test.c
  commands:
    - test -e ${PREFIX}/lib/libjemalloc.dylib # [osx]
    - test -e ${PREFIX}/lib/libjemalloc.so # [not osx]
    - gcc -I${PREFIX}/include/ test.c -o test.o
</pre>

I also like this test. It should be included as an example in the documentation.

bgruening:
Thanks @vodkatad @druvus! This looks great!

PR #2811, opened 2016-11-04
Rcas 1

dyusuf:

### DIFF
--- a/recipes/r-dt/meta.yaml
+++ b/recipes/r-dt/meta.yaml
@@ -2,14 +2,12 @@ package:
   name: r-dt
   # Note that conda versions cannot contain -, so any -'s in the version have
   # been replaced with _'s.
-  version: "0.1"
+  version: "0.2"
 
 source:
-  fn: DT_0.1.tar.gz
-  url:
-    - http://cran.r-project.org/src/contrib/DT_0.1.tar.gz
-    - http://cran.r-project.org/src/contrib/Archive/DT/DT_0.1.tar.gz
-
+  fn: DT_0.2.tar.gz
+  url: http://cran.r-project.org/src/contrib/DT_0.2.tar.gz
+  md5: 1ada2950daac6504fd30a09d944f845e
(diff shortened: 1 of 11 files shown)


dyusuf:
@bgruening  the failure is "quadmath.h No such file or directory" 

two possible reasons: 
1. include path is wrong
2. libquadmath-devel is not installed in condaforge/linux-anvil

I looked at the docker file, it is likely the 2nd reason. 

I wanted to test out the 2nd reason with the docker container locally, but I could not connect to the network from the container, may be because of the firewall issue. 

could you help out?

dyusuf:
@daler @bgruening 

I tried to run simulate-travis.py to test the recipe. 

when running ./simulate-travis.py --packages package_name --loglevel=debug

I got the error from scripts/travis-run.sh,

updating index in: /opt/conda/conda-bld/linux-64
updating index in: /opt/conda/conda-bld/noarch
**Could not find URL: file:///opt/host-conda-bld/linux-64/**

what could be the reason?

bgruening:
Is this under OSX?
Is libquadmath not included in GCC?

bgruening:
The error I can see is: https://travis-ci.org/bioconda/bioconda-recipes/jobs/173260716#L797
`Error : object ‘as_tibble’ is not exported by 'namespace:tibble'`

dyusuf:
regarding error related to libquadmath, it seems have gone after I included gcc as dependency of r-purrr. 

I am now trying to find the solution for the error related to tibble.

before doing that, I trying to make simulate-travis.py work in my system. well, I got the error Could not find URL: file:///opt/host-conda-bld/linux-64/

I am using ubuntu 14.04

bgruening:
Which Docker version? Seems to be Docker related, as if the mount-point is not correct.

dyusuf:
Docker version 1.12.3, build 6b644ec

bgruening:
Solved on chat.

PR #2819, opened 2016-11-04
Add recipe for trimal version 1.4.rev15

gregvonkuster:

### DIFF
--- a/recipes/trimal/1.4/meta.yaml
+++ b/recipes/trimal/1.4/meta.yaml
@@ -1,10 +1,11 @@
 package:
   name: trimal
-  version: "1.4.rev15"
+  version: "1.4.1"
 
 source:
-  git_url: https://github.com/scapella/trimal.git
-  git_rev: f7b4a27
+  fn: v1.4.1.tar.gz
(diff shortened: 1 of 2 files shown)


recipes/trimal/1.4/meta.yaml
```diff
@@ -0,0 +1,26 @@
+package:
+  name: trimal
+  version: "1.4.rev15"
+
+source:
+  git_url: https://github.com/scapella/trimal.git
```
bgruening: @gregvonkuster if possible please use the tarballs and a checksum. This makes it easier to backup the sources on depot.

gregvonkuster:
@bgruening I've changed the recipe to use the more standard url and md5, and I've eliminated the revision information from the version string.  I'm not familiar enough with this package to know if the new revisions are "backward reproducible". Do you think we need to be concerned about this?

bgruening:
I think this is fine. But I don't see your changes.

gregvonkuster:
@bgruening I'm asking the above question to the expert on this package, so I'll post the response as soon as I get it.

gregvonkuster:
Yep - I have merge conflicts because of your change.  I'll try to get things cleaned up.  :(

PR #2820, opened 2016-11-04
adds clustalo

brwnj:

New recipe... @bioconda/core

### DIFF
--- a/recipes/clustalo/meta.yaml
+++ b/recipes/clustalo/meta.yaml
@@ -1,0 +1,25 @@
+{% set version = "1.2.3" %}
+
+package:
+  name: "clustalo"
+  version: "{{ version }}"
+
+source:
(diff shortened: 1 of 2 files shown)


recipes/clustalo/meta.yaml
```diff
@@ -0,0 +1,25 @@
+{% set version = "1.2.3" %}
+
+package:
+  name: "clustalo"
+  version: "{{ version }}"
+
+source:
+  fn: clustalo-{{ version }}-Ubuntu-x86_64 # [linux]
```
bgruening: Is it possible to compile this tool from source?

bgruening: Do you know how these binaries are compiled?
The problem with recompiled binaries is that they are likely not usable on any older OS, like HPC environments.

That said, we do not have an strict rule in place so feel free to merge.

PR #2825, opened 2016-11-06
Autoconf

dyusuf:

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1,1 +1,1 @@
-recipes/autoconf
+#recipes/autoconf
--- a/recipes/autoconf/build.sh
+++ b/recipes/autoconf/build.sh
@@ -18,15 +18,15 @@ perl_inteprator=$(./configure --prefix=$PREFIX | grep 'checking for perl...' | s
 #create soft link and place its path in bin/Make
 ln -s $perl_inteprator temp_perl
 temp_link=$(pwd)/temp_perl
-sed -i "s|"$perl_inteprator"|"$temp_link"|g" bin/Makefile
+sed -i.bak "s|"$perl_inteprator"|"$temp_link"|g" bin/Makefile
 
 make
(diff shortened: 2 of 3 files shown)


dyusuf:
atuconf passed travis test on both GNU and BSD

while, then unexpectedly
OSError: [Errno 28] No space left on device :)

dyusuf:
restarted the build, the problem is gone.

druvus:
Thanks for the work around to fix this issue.

You should just use perl-threaded instead of perl as indicated in the guidelines.

dyusuf:
@druvus welcome. regarding perl vs per-threaded, I referenced the conda-forge recipe. but probably both are ok.

dyusuf:
@daler @bgruening while merging, "OSError: [Errno 28] No space left on device" appears again for the Linux build. are we running out of the space? https://travis-ci.org/bioconda/bioconda-recipes/builds/173783720

johanneskoester:
We still haven't even reached 50% of our anaconda space. But it looks like a no space left error on the travis machine. Does any of the new changes to bioconda-utils cause a lot of stuff to be written to disk? Maybe the on-the-fly container building?

bgruening:
@johanneskoester the containers roughly double the amount of space that is needed. It can be the issue but I see this happen very infrequently and non-predictable. The OSX images do not have this problem and often the build fails during upload to anaconda.org or during website generation.

johanneskoester:
Website generation is also creating lots of files. Can we clean up in between?

daler:
I bet the issue is bioconda-utils' docker usage. Docker-py defaults to `remove=True`. When I switched over to subprocess I copied the commands verbatim and forgot to add the `--rm`, so every recipe built has a container still sticking around. I'll fix that.

The docs might be contributing too. We might want to consider something different than sphinx for the autogenerated recipe docs.

PR #2829, opened 2016-11-07
Update viennarna recipe (2.3.0)

s-will:


bgruening:
For the Galaxy tools, do you know if the commandline interface has changed?

s-will:
Hi @bgruening, compared to 2.2.10 there are new features made available via the command line. Otherwise, the 2.2.10 arguments should still work.
Here are the relevant  changes:
* Introduced command files that subsume constraint definition files (RNAfold and RNAcofold)
* Add ID manipulation feature to RNAeval

PR #2832, opened 2016-11-07
recipe perl-inline-c (bioperl v1.7 dep) and dependencies

martin-mann:

Adds the following recipes that are dependencies for bioperl v1.7 (to be added separately)

- perl-inline-c
- perl-inline
- perl-io-all
- perl-pegex
- perl-yaml-libyaml


martin-mann:
@bgruening I restart our bioperl v1.7 efforts by splitting it up into individual dependencies and the main recipe... here comes one of the new dependencies (and its subdependencies).

bgruening:
Looks good to me!
@jerowe this might be interesting for you as well?!

PR #2835, opened 2016-11-07
Add recipe for trimAl version 1.4

gregvonkuster:

### DIFF
--- a/recipes/trimal/1.4/meta.yaml
+++ b/recipes/trimal/1.4/meta.yaml
@@ -1,0 +1,27 @@
+package:
+  name: trimal
+  version: "1.4"
+
+source:
+  fn: trimAl.tar.gz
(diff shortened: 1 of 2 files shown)


recipes/trimal/1.4/meta.yaml
```diff
@@ -0,0 +1,27 @@
+package:
+  name: trimal
+  version: "1.4"
+
+source:
+  fn: trimAl.tar.gz
+  url: https://github.com/scapella/trimal/archive/trimAl.tar.gz
```
bgruening: A branch is also not a fixed version, the developer can check in some new stuff and you will get it with a rebuild.
Why not this one: https://github.com/scapella/trimal/archive/v1.4.1.tar.gz -- to old?

gregvonkuster:
Ok, thanks, added in a separate PR.

PR #2844, opened 2016-11-07
Genomation

dyusuf:


dyusuf:
the update has not appeared in anaconda since 1 hour, the reason of lagging? because travis is too busy?

PR #2860, opened 2016-11-08
Jbrowse bug fix

abretaud:

The sed command was doing strange things with conda scripts, fixed now.
Also the mac osx build was skipped for no particular reason


abretaud:
Nevermind, can't build on osx because of perl-db-file

PR #2862, opened 2016-11-08
consistent channel order

daler:

@johanneskoester can you make sure this looks OK?

xref #2675


docs/index.rst
```diff
@@ -56,8 +56,9 @@ Step 2: Setup Bioconda
 
 After installing Miniconda, you can use the ``conda`` command to setup bioconda with::
 
-   conda config --add channels r
    conda config --add channels conda-forge
```
johanneskoester: Will this work? People will have defaults already in their condarc... Don't you need --append?

daler: When I was trying it, `--add` will add the channel as the highest priority even if it already exists in `.condarc`. So if you do the `--add` commands in this order, the same set of commands can be used no matter what the existing state of .condarc.

docs/index.rst
```diff
@@ -56,8 +56,9 @@ Step 2: Setup Bioconda
...
+   conda config --add defaults
```
johanneskoester: you mean conda config --add **channels** defaults

PR #2865, opened 2016-11-08
Add recipe for pasta version 0.2

gregvonkuster:

### DIFF
--- a/recipes/pasta/0.2/build.sh
+++ b/recipes/pasta/0.2/build.sh
@@ -15,17 +15,19 @@ cp -R ./* $out_dir
 unamestr=`uname`
 if [ $unamestr == 'Linux' ];
 then
-    wget https://github.com/smirarab/sate-tools-linux/archive/master.tar.gz
-    tar -xzf master.tar.gz
-    mv sate-tools-linux-master $pasta_tools_dev_dir/sate-tools-linux
+    wget https://github.com/smirarab/sate-tools-linux/archive/2924290.tar.gz
+    tar -xzf 2924290.tar.gz
+    mv sate-tools-linux-292429063ee62055b610b592356ff398f68c0a4a sate-tools-linux
+    mv sate-tools-linux $pasta_tools_dev_dir
     cd $out_dir
     $PYTHON setup.py develop
     cp -R $share_dir/sate-tools-linux/* $PREFIX/bin
 elif [ $unamestr == 'Darwin' ];
 then
-    wget https://github.com/smirarab/sate-tools-mac/archive/master.tar.gz
-    tar -xzf master.tar.gz
-    mv sate-tools-mac-master $pasta_tools_dev_dir/sate-tools-mac
+    wget https://github.com/smirarab/sate-tools-mac/archive/0712214.tar.gz
+    tar -xzf 0712214.tar.gz
+    mv sate-tools-mac-0712214e20152b2ec989fc102602afa53d3a7b1a sate-tools-mac
+    mv sate-tools-mac $pasta_tools_dev_dir
     cd $out_dir
     $PYTHON setup.py develop
     cp -R $share_dir/sate-tools-mac/* $PREFIX/bin
--- a/recipes/pasta/0.2/build.sh
+++ b/recipes/pasta/0.2/build.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+out_dir=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+pasta_tools_dev_dir=$PREFIX/share
+share_dir=$PREFIX/share
+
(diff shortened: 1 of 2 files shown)


recipes/pasta/0.2/build.sh
```diff
@@ -0,0 +1,49 @@
...
+mkdir -p $out_dir
+mkdir -p $PREFIX/bin
+mkdir -p $PREFIX/pasta
+
+cp -R ./* $out_dir
+
+unamestr=`uname`
+if [ $unamestr == 'Linux' ];
+then
+    wget https://github.com/smirarab/sate-tools-linux/archive/master.tar.gz
```
bgruening: This is not reproducible as it is the master branch.
Can we not have a separate package for this one?

gregvonkuster: Ah, I meant to change that - thanks for pointing this out.  Actually, these separate repositories of tools are required to compile the Pasta package, but I meant to use an actual rev instead of the master branch tip.  I'll fix this shortly.  Thanks again!

gregvonkuster:
@bgruening Is this line in the loss pointing out the cause of the travis error?
```
/opt/conda/conda-bld/work/pasta-6d52da14f48f55c10c26c3e6ad7bc2d5646feeaf/conda_build.sh: line 19: wget: command not found
```
Do conda recipes that use wget need to define gnu-wget as a dependency?  I've seen several recipes like mine that use wget but do not define it as a requirement.  Can you advise when you get a chance?

bgruening:
gnu-wget is a good idea. We changed the build container, recently and wget is now not included anymore in the container, so you need to specify it.

gregvonkuster:
Thanks - I'll fix this.  ;)

gregvonkuster:
Hmm...adding gnu-wget did not help.  Not sure why this is failing... :(

bgruening:
wget helped a lot imho, but this is a new error isn't it: RuntimeError: Setuptools downloading is disabled in conda build. Be sure to add all dependencies in the meta.yaml  url=https://files.pythonhosted.org/packages/source/c/certifi/certifi-2016.9.26.tar.gz#md5=baa81e951a29958563689d868ef1064dr

https://travis-ci.org/bioconda/bioconda-recipes/jobs/175057591#L1148

gregvonkuster:
Thanks @bgruening - I'll try not checking certificates with wget.  If that doesn't wok, I'll try adding a dependency on certifi.  ;)

bgruening:
Yeah, certifi is the way to go I think.

gregvonkuster:
would certifi be a dependency of the gnu-wget recipe or this recipe?

gregvonkuster:
@bgruening I've added certifi as a requirement - let's see what happens.  ;)

gregvonkuster:
This latest failure looks like a glibc version problem:
```
+ /bin/bash -x -e /opt/conda/conda-bld/test_tmp/run_test.sh
+ clustalw2 -help
+ grep CLUSTAL
clustalw2: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by clustalw2)
Using Anaconda Cloud api site https://api.anaconda.org
```
I don't see glibc being used as a dependency so I'm wondering how to handle this.  Conda search doen't return much either:
```
$ conda search glibc
Fetching package metadata .................
test-glibc                   0.1                           1  bioconda
```
Thanks for the help!  ;)

gregvonkuster:
@bgruening I believe the problem with this one is that the required tools retrieved from separate github repos are pre-compiled, and this causes problems when attempting to execute them on different platforms.  Here is the section of the recipe that pulls in these required tools from the github repo that is separate from the pasta source code repo:
```
    wget --no-check-certificate https://github.com/smirarab/sate-tools-linux/archive/2924290.tar.gz
    tar -xzf 2924290.tar.gz
    mv sate-tools-linux-292429063ee62055b610b592356ff398f68c0a4a sate-tools-linux
    mv sate-tools-linux $pasta_tools_dev_dir
    cd $out_dir
    $PYTHON setup.py develop
```
The sate-tools-linux repo contains these pre-compiled packages - for example: https://github.com/smirarab/sate-tools-linux/blob/master/clustalw2.  Obviously this will pose problems when attempting to execute the pasta tool on a platform with different libraries.  The Pasta installation instructions here https://github.com/smirarab/pasta/blob/master/README.md state this, and I have not found a way to actually compile these tools as part of the pasta build.
```
3. Clone the relevant "tools" directory (these are also forked from the SATe project). There are different repositories for linux and MAC. You can use git clone https://github.com/smirarab/sate-tools-linux.git for Linux or git clone https://github.com/smirarab/sate-tools-mac.git for MAC.
```
Do you have any advice on this one?  I could try to locate the source for these tools if they are publicly available, but I'm wondering if there  is a better way.

Thanks!!

gregvonkuster:
Fixed in a separate PR, so closing this one.

PR #2866, opened 2016-11-08
Version bump MSGF+.

glormph:

The version of MSGF+ in bioconda is pretty old. Can't blame anyone because it was pretty hard to find another version and they are spread all over the place. Problems with this PR:

 + Where do I write the version of MSGF+? The meta.yaml previously had 1.0.0 which cannot be the version for the application (which was v9979)
 + MSGF downloads are updated all the time but the URL is not. Do we have a place to host files?

### DIFF
--- a/recipes/msgf_plus/meta.yaml
+++ b/recipes/msgf_plus/meta.yaml
@@ -8,7 +8,7 @@ package:
   version: 2016.10.26
 
 build:
-  number: 1
+  number: 0
   skip: False
 
 source:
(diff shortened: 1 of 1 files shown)


bgruening:
@glormph you can store your tarballs here: https://github.com/galaxyproject/cargo-port
You could use the real version number, aka 9979 ... or what do you think about 1.9979?

glormph:
That would work, have the PR ready: https://github.com/galaxyproject/cargo-port/pull/104

MSGFPlus developers seem to have abandoned the numbered versioning and are now using a date, so I just follow that.

bgruening:
@glormph merged, it should be there in latest an hour.

PR #2854, opened 2016-11-08
Percolator again

glormph:
* [X] I have read the guidelines above.
* [X] This PR adds a new recipe.
* [ ] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

Finally, tried to do this in 3 branches already. :)

Issue #2853, opened 2016-11-08
Travis prefers defaults over conda-forge


glormph:
When building a recipe I needed boost from `conda-forge` (`defaults` doesn't include static libs). This worked fine in a local conda build run, but when pushing to travis the jobs failed because boost was installed from `defaults` anyway.

My guess is that my local conda has channels ordered with conda-forge before defaults (as specified in my `.condarc` file) and that it stops searching after it finds the first channel.

Should there be a standard order in bioconda, or a way to order or specify the channels before a build/run?

daler:
The recent changes were inconsistent across docs and travis-ci, which is probably what was causing this. See #2862 for the PR that fixes it.

However, we need to choose a single consistent channel order, and we decided that defaults should have higher priority than conda-forge. See #2675 for details.

@johanneskoester maybe we should keep track of cases like this, where the conda-forge should have higher priority, as well as other cases where defaults should have higher priority, to see if we can empirically arrive at the best channel order given the kinds of dependencies used in bioconda recipes.

glormph:
In my particular case I just specified a version not in defaults and that fixed it, but good to know. I assume the issue can be closed.

johanneskoester:
@glormph, can you refer to the problematic PR please?

glormph:
Sure, it was my [percolator again PR](https://github.com/bioconda/bioconda-recipes/pull/2854)

johanneskoester:
The hardcoding of boost is a bit problematic though. It will cause incompatibilities with other packages in bioconda that depend on a different boost version. That's why we have the global variable for that.

So, in your case, what was the problem with the boost version from default?

glormph:
The default boost did not install static libs. The conda-forge one does.

Another way around it would be installing boost from source. Since I only need filesystem and system libs it would maybe not take as much time as a full install (which would timeout Travis).

Not sure I understand the problem with the hardcoding. Do you mean the buildstring or the requirement?

johanneskoester:
I mean the requirement. Say two packages depend on boost. One on the version we consistently define globally, the other on a differnt one. They cannot be installed into the same environment. This is fine if there is a technical reason for depending on a particular version. But it is not ok, if the only reason to use a different version is to work around a more fundamental problem. 

But wait, the tool links statically against boost? In that case, you don't even need the runtime dependency, don't you?

glormph:
[...]

PR #2874, opened 2016-11-09
Removed Boost from runtime dep because it is statically linked

glormph:

Old PR for this: #2854 
Issue in which @johanneskoester told me to do this: #2853

### DIFF
--- a/recipes/percolator/meta.yaml
+++ b/recipes/percolator/meta.yaml
@@ -9,7 +9,7 @@ source:
 
 build:
   number: 1
-  string: boost_1.62_{{PKG_BUILDNUM}}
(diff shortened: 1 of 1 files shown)


recipes/percolator/meta.yaml
```diff
@@ -8,7 +8,7 @@ source:
   md5: de87c96643fed652e454ac492ce850b4
 
 build:
-  number: 0
+  number: 1
   string: boost_1.62_{{PKG_BUILDNUM}}
```
johanneskoester: Can you please also remove the underscore after boost? Just a convention.

glormph:
Thanks for the help @johanneskoester!

PR #2888, opened 2016-11-10
RelocaTE2.v2.0.0

JinfengChen:

### DIFF
--- a/recipes/relocate2/build.sh
+++ b/recipes/relocate2/build.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+export PATH="$PREFIX/bin:$PATH"
+export PYTHONPATH="$PREFIX/lib/python2.7/site-packages:$PYTHONPATH"
+mkdir $PREFIX/scripts
+cp ./scripts/*.p* $PREFIX/scripts
+cp ./scripts/*.p* $PREFIX/bin
+cp ./README.md $PREFIX
+cp ./test_data.tar.gz $PREFIX
+echo "bwa=$PREFIX/bin/bwa" >> $PREFIX/CONFIG
+echo "bowtie2=$PREFIX/bin/bowtie2" >> $PREFIX/CONFIG
+echo "bowtie2_build=$PREFIX/bin/bowtie2-build" >> $PREFIX/CONFIG
+echo "blat=$PREFIX/bin/blat" >> $PREFIX/CONFIG
+echo "samtools=$PREFIX/bin/samtools" >> $PREFIX/CONFIG
+echo "bedtools=$PREFIX/bin/bedtools" >> $PREFIX/CONFIG
+echo "seqtk=$PREFIX/bin/seqtk" >> $PREFIX/CONFIG
+
--- a/recipes/relocate2/build.sh
+++ b/recipes/relocate2/build.sh
@@ -1,10 +1,12 @@
 #!/bin/bash
 
-mkdir $PREFIX/scripts
-cp ./scripts/*.p* $PREFIX/scripts
-cp ./scripts/*.p* $PREFIX/bin
-cp ./README.md $PREFIX
-cp ./test_data.tar.gz $PREFIX
+mkdir -p $PREFIX/scripts
+mkdir -p $PREFIX/bin
+cp $SRC_DIR/scripts/*.p* $PREFIX/scripts
+cp $SRC_DIR/scripts/*.p* $PREFIX/bin
+cp $SRC_DIR/scripts/*.p* $PREFIX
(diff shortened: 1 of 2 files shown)


recipes/relocate2/build.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+export PATH="$PREFIX/bin:$PATH"
```
bgruening: This should not be needed.

recipes/relocate2/build.sh
```diff
@@ -0,0 +1,17 @@
...
+export PYTHONPATH="$PREFIX/lib/python2.7/site-packages:$PYTHONPATH"
```
bgruening: This also not ... is this not working by default?

recipes/relocate2/meta.yaml
```diff
@@ -0,0 +1,29 @@
+package:
+  name: relocate2
+  version: 2.0.0
+
+source:
+  git_url: https://github.com/stajichlab/RelocaTE2.git
```
bgruening: Please use a tarball and a checksum.
This will always pull the latest github and if we rebuild the recipe it will not be the same.

recipes/relocate2/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - samtools >=1.3.1
+    - bowtie2 >=2.2.8
+    - bedtools >=2.25.0
+    - bwa >=0.6.2
+    - seqtk >=1.2
+    - pysam >=0.9.1
+    - blat
+  run:
+    - samtools 
+    - bowtie2
```
bgruening: You probably want to have the `>=` here as well.

recipes/relocate2/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+    - bedtools
+    - bwa
+    - seqtk
+    - pysam
+    - blat
+
+#test:
+#  commands:
```
bgruening: A test is strictly needed in bioconda.

recipes/relocate2/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+mkdir $PREFIX/scripts
+cp ./scripts/*.p* $PREFIX/scripts
+cp ./scripts/*.p* $PREFIX/bin
+cp ./README.md $PREFIX
+cp ./test_data.tar.gz $PREFIX
```
bgruening: How big is this and is this really needed for a package?

recipes/relocate2/meta.yaml
JinfengChen: I have problem when building the package offline. The testing failed and says "run_test.sh: line 3: relocaTE2.py: command not found".  I am not sure I understand how the testing works. But it seems building and installation are fine. Can you give me some suggestion how I should modify the testing?
Thanks.

Jinfeng

recipes/relocate2/build.sh
```diff
@@ -0,0 +1,15 @@
...
+echo "bwa=$PREFIX/bin/bwa" >> $PREFIX/CONFIG
+echo "bowtie2=$PREFIX/bin/bowtie2" >> $PREFIX/CONFIG
+echo "bowtie2_build=$PREFIX/bin/bowtie2-build" >> $PREFIX/CONFIG
+echo "blat=$PREFIX/bin/blat" >> $PREFIX/CONFIG
```
bgruening: I guess the problem is here $PREFIX will be a full path and this will not be the same on all systems.
Is this file really needed?

recipes/relocate2/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+mkdir $PREFIX/scripts
+cp ./scripts/*.p* $PREFIX/scripts
+cp ./scripts/*.p* $PREFIX/bin
+cp ./README.md $PREFIX
```
bgruening: Not needed I guess.

recipes/relocate2/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+mkdir $PREFIX/scripts
+cp ./scripts/*.p* $PREFIX/scripts
```
bgruening: Is $PREFIX/scripts needed? Can you just copy Python files to bin.

JinfengChen:
I tried using bin. Still get errors. The problem is in '_t_env'. None of
the compiled tools or Python scripts installed or copied into it.  '_b_env'
is fine.
Thanks.

Jinfeng

2016年11月10日星期四，Björn Grüning <notifications@github.com> 写道：

PR #2907, opened 2016-11-11
Generalized Additive Models

rasmusagren:

### DIFF
--- a/recipes/r-gam/bld.bat
+++ b/recipes/r-gam/bld.bat
@@ -1,0 +1,8 @@
+"%R%" CMD INSTALL --build .
+if errorlevel 1 exit 1
+
+@rem Add more build steps here, if they are necessary.
+
+@rem See
+@rem http://docs.continuum.io/conda/build.html
+@rem for a list of environment variables that are set during the build process.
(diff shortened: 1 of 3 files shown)


recipes/r-gam/bld.bat
```diff
@@ -0,0 +1,8 @@
+"%R%" CMD INSTALL --build .
+if errorlevel 1 exit 1
+
```
bgruening: This file can be removed

recipes/r-gam/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+  # Note that conda versions cannot contain -, so any -'s in the version have
+  # been replaced with _'s.
+  version: "1.14"
+
+source:
+  fn: gam_1.14.tar.gz
+  url:
+    - http://cran.r-project.org/src/contrib/gam_1.14.tar.gz
+    - http://cran.r-project.org/src/contrib/Archive/gam/gam_1.14.tar.gz
+
```
bgruening: Can we add a checksum here?

recipes/r-gam/bld.bat
rasmusagren: Why is that? Should R packages never include a bat file? This was generated with conda skeleton.

bgruening: bat is used for windows builds, but bioconda is not building for the Win platform.

druvus:
@rasmusagren I was also working to include r-gam and got it working in #2923.

rasmusagren:
@druvus Great, then I remove this one.

PR #2892, opened 2016-11-10
r-sartools. New version 1.3.2.

loraine-gueguen:
* [x] I have read the guidelines above.
* [ ] This PR adds a new recipe.
* [x] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

Recipe for last r-sartools version.

PR #2914, opened 2016-11-11
Build setup improvements

daler:

`simulate-travis.py` can now be used to set up channel priorities and to install dependencies. This should help simplify setup for new contributors and makes sure everyone is testing locally using the same environment. The updated `travis-setup.sh` reflects this, where the following commands are used:

```bash
./simulate-travis.py --set-channel-order
./simulate-travis.py --install-requirements
```

These commands can be used by contributors to set the channel order according to config.yml and install the requirements for the bioconda-utils version indicated in .travis.yml (including bioconda-utils itself).

This PR also uses an updated version of bioconda-utils (xref https://github.com/bioconda/bioconda-utils/pull/30) which fixes the channel order issues we've been having. One problem was that channel order provided on the command line was the reverse of channel order added to `.condarc`. Thanks @druvus in https://github.com/bioconda/bioconda-recipes/pull/2892#issuecomment-260016776 for identifying the issue. Another was that the `mulled-build` channel arguments were not being sent correctly (thanks @acaprez for pointing this out, https://github.com/bioconda/bioconda-utils/issues/31).

@johanneskoester @bgruening can I get a review on this and https://github.com/bioconda/bioconda-utils/pull/30 please?


bgruening:
Tested and works for me locally, let's merge it now and test it a little bit over the weekend.

PR #2923, opened 2016-11-12
R deps

druvus:

More R dependencies for maaslin

### DIFF
--- a/recipes/r-gam/meta.yaml
+++ b/recipes/r-gam/meta.yaml
@@ -17,12 +17,12 @@ build:
 
 requirements:
   build:
-    - r-base
+    - r
     - r-foreach
     - gcc   
 
   run:
-    - r-base
+    - r
     - r-foreach
     - libgcc    # [not osx]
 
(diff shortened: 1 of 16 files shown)


recipes/r-gam/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+build:
+  # This is required to make R link correctly on Linux.
+  rpaths:
+    - lib/R/lib/
+    - lib/
+
+requirements:
+  build:
+    - r-base
```
bgruening: I think we agreed to use `r` as base dependency.

bgruening:
@druvus if you are pain approved and have some time I would appreciate some help with this one: https://github.com/bioconda/bioconda-recipes/tree/r29

recipes/r-gam/meta.yaml
druvus: thanks, I will replace it.

druvus:
@bgruening I was looking into r29 but I am not sure what is going on there. I will look into it again during the week.

bgruening:
Let's try this one :)

PR #2928, opened 2016-11-13
File does not exists if there is no package in a PR

bgruening:


bgruening:
@johanneskoester here we go. If there is no package and bioconda-utils was not running there is no such file.

bgruening:
@daler @johanneskoester I see this now: https://travis-ci.org/bioconda/bioconda-recipes/jobs/175473800#L494

The wired thing is that all tests end with SUCCESS, but the summary has an error. I will study this and try a little bit more before reverting. Let me know if you have ideas.

johanneskoester:
And it says it is filtering 100% but still wants to build recipes...

bgruening:
@johanneskoester it's working now!!!!
However, it does take longer than expected. I would argue that it is worth it but we should watch it over the next days and maybe revert the push.
The testing with busybox is also working now.

johanneskoester:
Great. Is the upload speed the problem?

bgruening:
I assume it is download of all R dependencies, but actually after seeing some PR I think it is actually generating the website ... haven't noticed this yet, but this seems to take a fair amount of time.

johanneskoester:
Indeed, I also noticed that recently. It does a lot of unnecessary stuff, since most packages don't change...

PR #2937, opened 2016-11-14
perl-gd : fontconfig dep (runtime lib load error)

martin-mann:

when using perl-gd, the libfontconfig.so lib cannot be found


martin-mann:
Whatthef... the new libgd package 2.2.3-2 was just compiled and released....

Osx build fails now with:
```
conda.exceptions.PaddingError: Placeholder of length '32' too 
   short in package bioconda::libgd-2.2.3-2.
The package must be rebuilt with conda-build > 2.0
```

@bgruening any idea

recipes/perl-gd/meta.yaml
```diff
@@ -14,15 +14,13 @@ requirements:
     - llvm # [osx]
     - perl-threaded >=5.22.0
     - libgd >=2.2.3
-    - fontconfig # for libfontcontig.so
```
bgruening: Martin these `#` are not comments.
`#` is part of the template language. So maybe these things are interpreted as some variable?

martin-mann: I dont think so, but we can try if it fails again. Last fail was most likely due to an ancient dep build...

bgruening: Ha, good to know!

martin-mann:
The osx prefix problem persidts even with rebuild libgd package...

bgruening:
Let's see ...

PR #2942, opened 2016-11-14
Libgd rebuild with fontconfig usage

martin-mann:

### DIFF
--- a/recipes/libgd/meta.yaml
+++ b/recipes/libgd/meta.yaml
@@ -14,6 +14,7 @@ requirements:
     - jpeg >=9* # use recent libjpeg
     - libpng
     - libtiff
+    - xz
     - fontconfig >=2.11 # for libfontconfig.so
     
   run:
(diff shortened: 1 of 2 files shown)


recipes/libgd/meta.yaml
```diff
@@ -14,7 +14,7 @@ requirements:
     - jpeg >=9* # use recent libjpeg
     - libpng
     - libtiff
-    - fontconfig # for libfontconfig.so
+    - fontconfig >=2.11 # for libfontconfig.so
     
   run:
     - libgcc # [not osx]
```
bgruening: This for example `# [not osx]` is code, it is no a comment.

bgruening: Mh ... http://jinja.pocoo.org/docs/dev/templates/#line-statements
```
Since Jinja 2.2, line-based comments are available as well. For example, if the line-comment prefix is configured to be ##, everything from ## to the end of the line is ignored (excluding the newline sign):
```

martin-mann: You jjssed the `are configured for #` so hash is not used per default and as given [here in conda spec](http://conda.pydata.org/docs/building/meta-yaml.html#preprocessing-selectors) they are used in meta.yaml as comments 😎

PR #2958, opened 2016-11-15
blacklist mysqlclient

bgruening:

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1,1 +1,2 @@
 recipes/vsearch
+recipes/mysqlclient


johanneskoester:
@bgruening can you explain the failed test? I can't see the real error message at first sight.

bgruening:
I can't :( and I'm don't have time currently to look into it. 
It seems that all tests do pass with "SUCCESSFUL" but the summary is failing for some reason. Maybe a mixup with some error codes?

johanneskoester:
No problem. Having that blacklist for such workarounds is a good idea :-). But I think we should add the corresponding PR as a comment in the future.

martin-mann:
I encountered similar failures in "normal" PRs yesterday too... 

Is there a way to get a travis build with full output rather than this condensed "all fine; not successful" output?

Have there been any changes of late within the build infrastructure? Are you using "conda standard script" (whatever that means.. I have no idea about conda.. just trying to help Björn and myself to get a few packages online) so we could ask at conda-build?

PR #2961, opened 2016-11-15
Update Kaiju to 1.4.4, add OSX build.

acaprez:

ping @bioconda/core since this is a non-standard OS X build.

Building this with `clang` requires a minimum `MACOSX_DEPLOYMENT_TARGET` of 10.9, which I believe we don't want to do for compatibility reasons.  Building with GCC on OS X works fine, but I wanted to confirm that's ok to do.

### DIFF
--- a/recipes/kaiju/1.0/build.sh
+++ b/recipes/kaiju/1.0/build.sh
@@ -19,3 +19,4 @@ cp makeDB.sh $PREFIX/bin
 cp mergeOutputs $PREFIX/bin
 cp mkbwt $PREFIX/bin
 cp mkfmi $PREFIX/bin
+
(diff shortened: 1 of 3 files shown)


johanneskoester:
That's fine. But are you sure it really uses the gcc dependency for building? I ask because there is a gcc alias for clang on OSX and I have seen before that although gcc was specified as a dependency and should come first in the $PATH, clang was used.

acaprez:
I can confirm it does on my local OSX VM.  If needed, I can tweak the build script to be more explicit.  Also, I edited my first comment - it will work with `MACOSX_DEPLOYMENT_TARGET=10.9`.  So I guess my question is which way is preferable - 1) use GCC -or- 2) use `clang` and set `MACOSX_DEPLOYMENT_TARGET=10.9`?

bgruening:
GCC is completely fine with me.

johanneskoester:
I agree. Use gcc.

PR #2968, opened 2016-11-16
Feature/trinontate

jerowe:

@bioconda/core .

This one is rather urgent, and probably my screw up so I'm sorry!

In the meta.yaml file the version is listed as 3.2.2. It should be 3.0.1, because there is no such thing as a version 3.2.2. So there is an empty bioconda trinotate 3.2.2 package floating around, which should be killed.

```
 anaconda show bioconda/trinotate
```
```
Using Anaconda Cloud api site https://api.anaconda.org
Name:    trinotate
Summary: Trinotate is a comprehensive annotation suite designed for automatic functional annotation of transcriptomes, particularly de novo assembled transcriptomes, from model or non-model organisms
Access:  public
Package Types:  conda
Versions:
   + 3.2.2
   + 3.0.1

To install this package with conda run:
     conda install --channel https://conda.anaconda.org/bioconda trinotate
```

I'm rebuilding this package as 3.0.1, which is correct. I don't even know how they both exist. Could someone with write access delete the 3.2.2?


johanneskoester:
I deleted the wrong version on anaconda some days ago, and it does not show up in the web interface any more: https://anaconda.org/bioconda/trinotate/files. But aparently I need to use the CLI for removing, because some kind of folder was kept. I fixed it now. No need to trigger a rebuild though. This was already done last time.

PR #2974, opened 2016-11-16
SortMeRNA support for osx

wasade:

Builds work on OSX now. The Linux host I tested on failed, although it failed for the existing recipe as well so my guess is there's an independent environment issue on that host.

### DIFF
--- a/recipes/sortmerna/2.0/meta.yaml
+++ b/recipes/sortmerna/2.0/meta.yaml
@@ -12,10 +12,11 @@ build:
 
 requirements:
   build:
-    - gcc # [linux]
+    - gcc # [not osx]
+    - llvm # [osx]
     - zlib
   run:
-    - libgcc
+    - libgcc # [not osx]
     - zlib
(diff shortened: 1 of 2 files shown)


recipes/sortmerna/2.0/meta.yaml
```diff
@@ -9,11 +9,10 @@ source:
...
   number: 1
-  skip: True # [not linux64 or osx]
 
 requirements:
   build:
-    - gcc
+    - gcc # [linux]
     - zlib
   run:
     - libgcc
```
bgruening: libgcc needs also to be garded only for linux and for OSX your need to add llvm # [osx] (ideally). Please see your guidelines.

wasade: huh. but it works?

wasade: note, i adapted from the bowtie2 recipe

PR #2976, opened 2016-11-16
Add biocondcutor-alpine and its dependency r-speedglm

upendrak:
* [y ] I have read the guidelines above.
* [y ] This PR adds a new recipe.


r-speedglm/bld.bat
```diff
@@ -0,0 +1,8 @@
+"%R%" CMD INSTALL --build .
+if errorlevel 1 exit 1
+
+@rem Add more build steps here, if they are necessary.
```
bgruening: This file can be removed as we do not build for the Win platform.

r-speedglm/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+  # Note that conda versions cannot contain -, so any -'s in the version have
+  # been replaced with _'s.
+  version: "0.3_1"
+
+source:
+  fn: speedglm_0.3-1.tar.gz
+  url:
+    - http://cran.r-project.org/src/contrib/speedglm_0.3-1.tar.gz
+    - http://cran.r-project.org/src/contrib/Archive/speedglm/speedglm_0.3-1.tar.gz
+
```
bgruening: Can you add a checksum here.

r-speedglm/bld.bat
upendrak: Ok got it.

r-speedglm/meta.yaml
upendrak: MD5 (meta.yaml) = 7502ed87a6175f0c94d2fd799370bd39

bgruening: No from this tarball which you want to download/compile. -> http://cran.r-project.org/src/contrib/speedglm_0.3-1.tar.gz

bgruening:
@upendrak for picard it would be better to create a separate PR. Leave it like it is this time. But separate PR makes it easier to debug if something goes wrong.

upendrak:
Sorry about that last accidental commit. Currently in the process of removing it

bgruening:
Thanks @upendrak!!!

upendrak:
I still don't see the docker images on quay.io. Do i need to wait longer?

bgruening:
It should be there now. It still takes a small amount of time after merging, as travis needs to build the conda package again and the Docker image + pushing to the registry.

upendrak:
Got it. I understand the whole process now. Sorry for the trouble..

PR #2979, opened 2016-11-17
Move dnapi recipe into a proper folder.

johanneskoester:

The dnapi recipe has been put directly into the recipes dir, not into a properly named subfolder (see PR #2975). Apart from being simply wrong, this breaks our CI.

@bgruening, I will see if I can add a check for errors like that to our travis setup.


bgruening:
Maybe adding this to bioconda-utils?

johanneskoester:
I thought about that, but it seems to be more travis specific at the moment. In principle, bioconda-utils can be run in completely different layouts. Also, it won't be affected by the recipe at all.

bgruening:
I was more thinking in the direction to tell the user as early as possible, during local testing, that something went wrong. But I guess simulate-travis.py is good as well.

johanneskoester:
That is certainly a good point. We can maybe reconsider this later. The advantage of the current solution is that we can even fail before anaconda is deployed, see PR #2980.

PR #2973, opened 2016-11-16
Update cityhash to 0.1.7

mvdbeek:
* [x] I have read the guidelines above.
* [ ] This PR adds a new recipe.
* [x] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

Issue #2957, opened 2016-11-15
Perl-osx-nov PR merge: linux package build failed


martin-raden:
@bgruening Somehow the `perl-osx-nov/xmlbuilder` merge builds failed even after it went through for the branch build

[https://travis-ci.org/bioconda/bioconda-recipes/builds/175984864](https://travis-ci.org/bioconda/bioconda-recipes/builds/175984864)

What to do?

bgruening:
"fixed" it by put it on the blacklist: https://github.com/bioconda/bioconda-recipes/pull/2958

@daler any idea why all tests pass with "SUCCESSFUL" but the summary actually fails?
https://travis-ci.org/bioconda/bioconda-recipes/jobs/175984865#L495

martin-mann:
As said, also happend for the following merge of `xmlbuilder`...

daler:
@bgruening no idea -- I'll block out some time later today to look at this.

daler:
Hopefully this will fix it: https://github.com/bioconda/bioconda-utils/pull/34

martin-mann:
so after merging this, should I push a new build (number) of the failed package? (and remove it from the blacklist)

epruesse:
outdated

PR #2981, opened 2016-11-17
Update bioconda-utils

johanneskoester:

Update to a new bioconda-utils version that should fix the issues seen in #2957 and #2973.


bgruening:
What do you think about fixing this to master? This saves us from these additional PR and our contributors from potential rebases.

Is there a use-case where master of utils should not be used here?

Ping @daler

johanneskoester:
Yes, we could do that. I would activate branch protection first.

daler:
Yeah, probably a good idea to minimize travis builds whenever we can. I just turned on branch protection on bioconda-utils master branch. I'm making a small change in bioconda-utils to address #2982, so I'll test out the branch protection and then point recipes to bioconda-utils master branch.

johanneskoester:
Thanks!

PR #2990, opened 2016-11-17
Add smhasher

mvdbeek:

### DIFF
--- a/recipes/smhasher/meta.yaml
+++ b/recipes/smhasher/meta.yaml
@@ -10,7 +10,8 @@ source:
 requirements:
   build:
     - python
-    - gcc
+    - gcc [linux]
+    - llvm [osx]
 
   run:
     - python
--- a/recipes/smhasher/meta.yaml
+++ b/recipes/smhasher/meta.yaml
@@ -10,33 +10,20 @@ source:
 requirements:
   build:
     - python
-    - gcc [linux]
-    - llvm [osx]
+    - gcc   # [linux]
+    - llvm  # [osx]
 
   run:
     - python
-    - libgcc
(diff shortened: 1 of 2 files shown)


recipes/smhasher/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+  url: https://pypi.python.org/packages/5f/b2/421c060300f1449646787273ecd7df558f6f8e4caee3d6fbe8b193cd60d4/smhasher-0.150.1.tar.gz
+  md5: 4fc61f36b141e137dd2d71ecd0167866
+
+requirements:
+  build:
+    - python
+
+  run:
+    - python
+    - libgcc
```
bgruening: @mvdbeek this is only needed if you compile something afaik, do you really need it?

bgruening: Yes there is. Conda-forge is still using the build-in compiler.

mvdbeek: Ahh, that explains it. So bets practice is to include gcc then, right?

bgruening: Yes this is bioconda best practice at the moment.

recipes/smhasher/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+requirements:
+  build:
+    - python
+    - gcc [linux]
+    - llvm [osx]
+
+  run:
+    - python
+    - libgcc
```
bgruening: libgcc only on linux see here: https://github.com/bioconda/bioconda-recipes/blob/master/GUIDELINES.md#cc

PR #2997, opened 2016-11-18
Update mothur to 1.38.1.1

mdehollander:
* [ x ] I have read the guidelines above.

### DIFF
--- a/recipes/mothur/makefile.patch
+++ b/recipes/mothur/makefile.patch
@@ -1,0 +1,12 @@
+--- Makefile
++++ Makefile
+@@ -12,8 +12,8 @@
+ OPTIMIZE ?= yes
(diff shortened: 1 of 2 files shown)


acaprez:
I did the original Makefile patching, so I took a look.  They apparently split the makefile into two separate files in this version.  I attached an updated `Build-Flags.patch` file here that tweaks both of them.  It built successfully in my Linux VM.  I wasn't able to test OSX as I'm away from my work laptop, but hopefully it should build there without issue as well.
[Build-Flags.patch.txt](https://github.com/bioconda/bioconda-recipes/files/601190/Build-Flags.patch.txt)

druvus:
Merged in another PR.

PR #3000, opened 2016-11-18
Use sed from git

nsoranzo:

Sandboxing patch has been upstreamed.


bgruening:
PR number 3000!!!!

bgruening:
autoconf and rsync is missing? Why rsync, Oo.

nsoranzo:
> autoconf and rsync is missing? Why rsync, Oo.

@bgruening I'm adding autoconf and automake to the build requirements. I'll try to remove rsync from sed's `bootstrap.conf`, then `./bootstrap` should fall back to wget.

nsoranzo:
automake has the same shebang problem of texinfo: https://travis-ci.org/bioconda/bioconda-recipes/jobs/177031115#L534

bgruening:
`gnu_wget` is missing and rsync is still missing :(

nsoranzo:
rsync shouldn't be needed when I'll add gnu_wget (likely on Monday).

nsoranzo:
gnu-wget added, now 3 builds out of 4 are green, no clue what's happening under OSX in https://travis-ci.org/bioconda/bioconda-recipes/jobs/177191097#L2870

nsoranzo:
@bgruening I don't think the problem is the perl requirement, it should be installed anyway as it is a run requirement for texinfo. Actually it made fail the Linux build too!

bgruening:
:( then I have no idea? Can we skip the man page generation?

nsoranzo:
@bgruening I think man page creation is failing because it is trying to run `sed/sed` which is not properly compiled/linked.
I disabled man page creation and now I get https://travis-ci.org/bioconda/bioconda-recipes/jobs/177610710#L2967

bgruening:
@nsoranzo can you try this: https://github.com/bioconda/bioconda-recipes/blob/f5eb63e30a76fd13c28663786d219c9f7750267c/recipes/python-bioext/build.sh#L4

nsoranzo:
Green finally!

bgruening:
Yeah! Thanks!

PR #3008, opened 2016-11-19
Add mysql-python/

druvus:

It exist a mysql-python recipe at anaconda but it is not built for osx. It seems that they are not planning to build it according this [open issue](https://github.com/ContinuumIO/anaconda-issues/issues/254).

It is needed to build ete2 and other dependencies for osx.  ping @bioconda/core


bgruening:
@druvus do you want to try conda-forge?

druvus:
You're right. I will give it a try.

PR #3012, opened 2016-11-20
ViennaRNA: activate python2 interface (for linux)

dmaticzka:

This adds the python2 interface. This is enabled by the patch for the autotools configure that now also searches for "python-config" in addition to "python2-config.

I added the python runtime dependency to enable builds for python 2 and 3. Can we have this behaviour without the actual dependency? It's not required at runtime.

The python2 interface does not work with osx, same as for python3...

@s-will @bgruening What'ya'think? This would introduce different versions of ViennaRNA depending on the python lib to be used. This seems a bit awkward to me, but we can't compile both interfaces at the same time. (I think).


dmaticzka:
trying to test python import only for linux would make conda crash on osx, disabling the import tests fixed that...

s-will:
Hi Daniel, I discussed the problem quickly with @raumzeit. We think that having two package versions for ViennaRNA is a little awkward. As long as we don't build split packages, one should rather get both interfaces in the same package.
The main obstacle in conda is that one cannot define both python versions as build dependencies (since they cannot be installed into the same env.)
I am testing a workaround (aka hack), see PR #3021

PR #3020, opened 2016-11-21
Bump IgBLAST to 1.7.0

marcelm:

Keep IgBLAST 1.4.0 in a subdirectory

### DIFF
--- a/recipes/igblast/build.sh
+++ b/recipes/igblast/build.sh
@@ -4,33 +4,47 @@ set -euo pipefail
 IGBLAST_ADDRESS=ftp://ftp.ncbi.nih.gov/blast/executables/igblast/release
 SHARE_DIR=$PREFIX/share/igblast
 
-wget $IGBLAST_ADDRESS/edit_imgt_file.pl
-# replace the first line with /usr/bin/env perl instead of the hard-coded /opt
-sed -i.backup '1 s/^.*$/#!\/usr\/bin\/env perl/g' edit_imgt_file.pl
-chmod +x edit_imgt_file.pl
-mv edit_imgt_file.pl bin/
-
 mkdir -p $PREFIX/bin
+
+# This is going to contain the igblastn and igblastp binaries.
+# Only wrappers are installed into $PREFIX/bin/ .
 mkdir -p $SHARE_DIR/bin
 
 if [ $(uname) == Linux ]; then
-    # Distributed IgBLAST 1.5.0 binaries set an RPATH that conda build complains build.
-    for FILE in makeblastdb igblastn igblastp; do
-        patchelf --remove-rpath bin/$FILE
-    done
+    # If on Linux, compile the tool ourselves because the distributed binaries
+    # link against libbz2.so, and the usual conda bzip2 package does not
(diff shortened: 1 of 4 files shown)


marcelm:
@bioconda/core This recipe fails and I don’t know why. The log contains a lot of stuff, but nothing helpful. A local `simulate-travis.py` run finishes successfully.

daler:
Looks like a mulled-build test fail. @bgruening any ideas? Can we get more debug info from mulled-build?

bgruening:
@marcelm I think if you remove the /dev/null from your test you will get an error message: `igblastn -h > /dev/null 2>&1`
Also please note that you have not specified the gcc/llvm compiler in your updated recipe.

bgruening:
`[Nov 21 20:51:20] SERR /usr/local/bin/../share/igblast/bin/igblastn: error while loading shared libraries: libbz2.so.1: cannot open shared object file: No such file or directory`

marcelm:
This did not work because the conda bzip2 package does not include `libbz2.so`, only `libbz2.a` (for static linking).

This recipe packages the binary version of IgBLAST, so changing compile-time options is not possible.

Any ideas?

jlanga:
HI @marcelm. I had the same issue some days ago with `blast+`. There is a [PR at conda-forge](https://github.com/conda-forge/staged-recipes/pull/1977) that should will fix this.

marcelm:
Ah, this is great, then I’ll just have to wait! :-)

marcelm:
Current status: The conda-forge PR with the bzip2 recipe at conda-forge/staged-recipes#1977 was closed without being merged. Instead, conda-forge/staged-recipes#2467 has been merged, which includes a bzip2 recipe, but without Linux shared libs.

johanneskoester:
Can the work from the first PR be easily included into the new bzip2 feedstock?

marcelm:
I hope so, I’ll make a PR.

johanneskoester:
Thanks!

marcelm:
I opened [a PR](https://github.com/conda-forge/bzip2-feedstock/pull/1). However, there is another problem: The channel order [recommended in the bioconda manual](https://bioconda.github.io/#set-up-channels) lists conda-forge *after* defaults, so actually the version from conda-forge isn’t used (defaults has 1.0.6-3 and conda-forge has 1.0.6-0). Any suggestions of how best to solve this?

johanneskoester:
The only way to solve this is to finally get rid of the R channel. Then, we can have conda-forge with a higher priority than defaults. I hope to be able to make some more progress on that soon. Any help would be appreciated.

marcelm:
A workaround would be to increase the build number to 4 and have this recipe depend on 1.0.6-4 or later. Not the preferred solution of course.

johanneskoester:
Yes, that would work. I don't know if the conda-forge people would object against this.

marcelm:
Hi @bioconda/core I have finally fixed this recipe by not using the prebuilt binaries on Linux, but instead compiling from sources. This circumvents the issue with `libbz2.so` not being available in the `bzip2` package from the defaults channel. Since the changes are larger than for a normal version bump, I’d like to get a review.

PR #3021, opened 2016-11-21
Update ViennaRNA: build Python 2 and 3 bindings

s-will:

### DIFF
--- a/recipes/viennarna/build.sh
+++ b/recipes/viennarna/build.sh
@@ -1,7 +1,6 @@
-#!/bin/sh
-# ViennaRNA does not support Python 3 for MacOSX.
+#!/bin/bash
 
-## choose extra configure options depending on the operating system
+## Choose extra configure options depending on the operating system
 ## (mac or linux)
 ##
 if [ `uname` == Darwin ] ; then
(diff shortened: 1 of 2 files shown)


s-will:
Building interfaces for Python 2 and 3 seems to require a hack, since conda does not allow to cleanly define two different versions of the same package as build dependencies. (Other solutions like defining packages python27 and python35 are ruled out here, since they look too invasive)

Vienna requires Python 2 and 3 at run time to generate Swig bindings for both. Since conda does not allow installation of both python versions in the same environment, we perform this workaround: in the build script(!), install Python 2 in a different environment and set the path accordingly.

s-will:
I had to navigate around a few more unexpected obstacles, but finally this worked (i.e. not only in principle.)
@bgruening, are there immediate improvements that I missed? Who could help/would be interested to find a cleaner solution in the long run?
@dmaticzka, could this fully replace your version or should we merge anything?
@raumzeit, I could not include the gsl dependency (only tested locally yet); let's look at this offline.
@RobertKleinkauf, I hope this will solve your problem with packaging antarna.

dmaticzka:
Thanx alot for that solution! This fully replaces my variant, I just closed the PR.

s-will:
Hi @bgruening, I wonder why there is no new package (2.3.1, build 1) on the channel by now, although the PR was merged >4h ago. Is the system confused? -- maybe by the parallel pull requests for build 1? At least I am confused;)

bgruening:
The tarball was not reachable for some time. It is back now, but we have a huge backlog in Travis now :(

bgruening:
@s-will https://quay.io/repository/biocontainers/viennarna?tab=tags
Everything should be online.

PR #3024, opened 2016-11-21
Recommend against redirecting test output to /dev/null

marcelm:

Hiding test output is not that useful: The output may be less cluttered, but when the recipe builds, then no one looks at the Travis log anyway. If it does not build, on the other hand, then crucial error messages may get hidden, as in PR #3020.

### DIFF
--- a/GUIDELINES.md
+++ b/GUIDELINES.md
@@ -164,8 +164,8 @@ An example of this is [fastq-screen](recipes/fastq-screen).
 
 For command-line tools, running the program with no arguments, checking the
 programs version (e.g. with `-v`) or checking the command-line help is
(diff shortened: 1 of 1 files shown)


bgruening:
ping @bioconda/core 
I'm in favor of this change.

gregvonkuster:
This is good to know - I've been wondering about this as well.  I have some new recipe coming up and I will no longer redirect to /dev/null.

johanneskoester:
You are right. With the current way bioconda-utils works, it should be hidden anyway if the test succeeds.

bgruening:
Thanks @marcelm for enhancing also the docs - very much appreciated!

PR #3028, opened 2016-11-22
Add version 0.2 of plant_tribes_gene_family_classifier

gregvonkuster:

### DIFF
--- a/recipes/plant_tribes_gene_family_classifier/0.2/build.sh
+++ b/recipes/plant_tribes_gene_family_classifier/0.2/build.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+mkdir -p $PREFIX/bin
+mkdir -p $PREFIX/config
+mkdir -p $PREFIX/data
+mkdir -p $PREFIX/pipelines
+mkdir -p $PREFIX/scripts
+
+cp ./pipelines/GeneFamilyClassifier $PREFIX/bin
+ln -s $PREFIX/bin/GeneFamilyClassifier $PREFIX/pipelines/GeneFamilyClassifier
+cp ./config/* $PREFIX/config
+cp ./data/* $PREFIX/data
+cp ./scripts/* $PREFIX/scripts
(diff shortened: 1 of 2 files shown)


recipes/plant_tribes_gene_family_classifier/0.2/build.sh
```diff
@@ -0,0 +1,13 @@
...
+mkdir -p $PREFIX/bin
+mkdir -p $PREFIX/config
+mkdir -p $PREFIX/data
+mkdir -p $PREFIX/pipelines
+mkdir -p $PREFIX/scripts
+
+cp ./pipelines/GeneFamilyClassifier $PREFIX/bin
+ln -s $PREFIX/bin/GeneFamilyClassifier $PREFIX/pipelines/GeneFamilyClassifier
+cp ./config/* $PREFIX/config
+cp ./data/* $PREFIX/data
```
bgruening: How large are these data files?

gregvonkuster: There is just 1 file named data.info which looks like this:
```
$ cat data/data.info 
# All PlantTribes scaffold data directories should be placed in this loacation - "PlanTribes/data"

# Available scaffold data sets: 
22 plant genomes (Angiosperms clusters, version 1.0) = 22Gv1.0
22 plant genomes (Angiosperms clusters, version 1.1) = 22Gv1.1
```
I'm just being careful here to make sure this directory exists with this small file in it due to the files instructions which state:
```
All PlantTribes scaffold data directories should be placed in this loacation (sic)
```
I want to make sure that the AssemblyPostProcesser script will not have problems if this directory does not exist under some scenarios.  The script requires a certain directory structure relative to the location of the script (e.g., I know that the config directory must exist at the same level as the script's directory).

gregvonkuster: Thanks a bunch @bgruening ;)

PR #3030, opened 2016-11-22
added recipe for dmd

pirovc:

@bioconda/core

### DIFF
--- a/recipes/dmd/meta.yaml
+++ b/recipes/dmd/meta.yaml
@@ -1,0 +1,23 @@
+package:
+  name: dmd
+  version: "2.065.0"
+
+source:
+  fn: dmd.2.065.0.zip
+  url: http://downloads.dlang.org/releases/2014/dmd.2.065.0.zip
+  md5: a17a699a7e4715658393819e9dc1814a
+
+build:
+  number: 0
+  skip: True # [osx or linux32]  
+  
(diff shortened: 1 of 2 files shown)


recipes/dmd/meta.yaml
```diff
@@ -0,0 +1,23 @@
+package:
+  name: dmd
+  version: "2.065.0"
+
+source:
+  fn: dmd.2.065.0.zip
+  url: http://downloads.dlang.org/releases/2014/dmd.2.065.0.zip
```
bgruening: Is it possible to compile these tools and not use the precompiled ones?
Is there already a D compiler in conda?

pirovc: I believe so, but I never did it and I do not have the time right now to do so. I am adding this specific version of the D so I can satisfy a dependency for another tool. 

I could not find a D compiler in conda.

bgruening:
@johanneskoester what do you think? I feel more this should be in conda-forge.

kyleabeauchamp:
agree

johanneskoester:
Yes. Can you please move this over to conda-forge/staged-recipes? The reason is that since we depend on conda-forge we try to put all general purpose stuff in there and use bioconda only for bioinformatics-related packages.

PR #3032, opened 2016-11-22
added recipe for motus

pirovc:

@bioconda/core


recipes/motus/meta.yaml
```diff
@@ -0,0 +1,25 @@
+package:
+  name: motus
+  version: "1.0"
+
+source:
+  fn: mOTUs.Linux64bits.tar.gz
+  url: http://www.bork.embl.de/software/mOTU/share/mOTUs.Linux64bits.tar.gz
```
bgruening: This indicates something is compiled, correct? Can this be compiled in the build.sh script?

bgruening: Oh, even worse, the binaries are includes in the perl file as binary blob - as text - this is ... omg.

bgruening: It the test checking that the binaries are actually working?

bgruening: Sure, you can have as many commands as possible in the test section.
For you second question, this is certainly possible but it would be $PREFIX/opt/ and a link to $PREFIX/bin.

An other idea would be to run this tool once in the build.sh script and create the binaries in this step. Then we can remove the binary blob from the pl script.

pirovc: Thanks for the help. I changed the build to copy the script to $PREFIX/opt/ with a link to $PREFIX/bin. I also removed the blob from the perl script. I am not checking the binaries on the meta.yaml file because motus tool reference them directly (to the $PREFIX/opt/ folder) and I am not copying them to the bin. However, I am testing if they were generated in the build.sh script.

daler:
@pirovc #3080 now fixed -- you should be good to rebase and try again.

pirovc:
I believe this version is now working fine.

pirovc:
@bioconda/core @bgruening is anything holding this recipe back? I waiting for it to be merged so I can go on with another package that uses it as a dependency. Please let me know if I need to change anything or if I can do something to make it happen.

bgruening:
Sorry just missed this one. LGTM.

PR #3038, opened 2016-11-23
WIP: reago

HassanAmr:


HassanAmr:
Hi @bgruening,
Sorry for creating a WIP pull request, but I don't know why the patch is not working since it works locally (without conda), and I wanted to ask your help in the matter.

HassanAmr:
Is there a better way to discuss workflows without doing pull requests?

bgruening:
There seems to be a problem with the setup.py file?

/opt/conda/conda-bld/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/bin/python: can't open file 'setup.py': [Errno 2] No such file or directory

HassanAmr:
Yeah, I checked the working directory of conda build, the patch doesn't generate it.
but when I run `patch < reago.patch`, the file generates just fine.

bgruening:
You could simply put this setup.py in this repository and copy the file from RECIPE_DIR to SRC_DIR isn't it?

HassanAmr:
oh, ok. That will work. Should I then delete it?

bgruening:
what do you want to delete? Just push to your branch and it will update this PR. The patch can be removed, yes.

HassanAmr:
No, I mean should I delete setup.py after running it?

bgruening:
No, it will not end up in $PREFIX. So no need for removing it.

HassanAmr:
Ok, thanks.
What should I put for the License?
Also, how can I discuss the other WIP workflows without a PR, or should I just do the pull requests for the them as well?

recipes/reago/meta.yaml
```diff
@@ -13,9 +13,19 @@ requirements:
   build:
     - python
     - setuptools
+    - os
```
bgruening: No. This is not working. Please test this locally. os, sys, time, operator are all in the standard lib of pyhon. Networkx is a module that needs to be installed.

HassanAmr: Sorry for uploading it faulty (I was switching machines), it does indeed fail locally too.
So, I only need to include Networkx as a dependency, and leave the rest out?

bgruening:
@HassanAmr looks good! Please try to upstream this patch :)

HassanAmr:
thanks :) Will do

PR #3041, opened 2016-11-23
Update gsl pin to version 2.1.

johanneskoester:

@bioconda/core this PR pins GSL to version 2.1 and should update all dependent packages. 
The move is necessary, because Continuum apparently decided to remove the GSL 1.16 package from the default channel. This should not happen IMHO, because it defeats reproducibility, e.g. when people use conda environments to freeze a data analysis for future use or publication.

This makes me even more willing to get rid of the R and default channel as soon as possible.


johanneskoester:
Conda-forge has just added 1.16, which should fix our problem. Since many of the tools fail to build against 2.1, I will wait with the update.

PR #3045, opened 2016-11-23
remove scikit-allel as it is available upstream via conda-forge

alimanfoo:

The package scikit-allel is now available upstream via the conda-forge channel. Removing it from bioconda so I don't have to maintain in two places.


alimanfoo:
@bioconda/core could you review, this PR is intended to remove a package from bioconda.

bgruening:
@alimanfoo is the same exact version also available via conda-forge?

alimanfoo:
No, only versions since 0.21.2 are available via conda-forge.

bgruening:
Ok, in this case we leave them all in bioconda but remove the recipe to redirect people to the newer conda-forge one!

Thanks @alimanfoo!

alimanfoo:
Thank you. Just to check, this means that in future if someone does ``conda install scikit-allel`` then it will install from conda-forge even if bioconda is a higher priority channel?

PR #3047, opened 2016-11-23
Add angsd recipe

percyfal:

ping @bioconda/core


recipes/angsd/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  license: GPLv3, MIT
+  summary: "ANGSD: Analysis of next generation Sequencing Data"
+
+package:
+  name: angsd
+  version: {{ version }}
+
+requirements:
+  build:
+    - htslib
```
bgruening: You should add gcc/llvm as described in our Guideline to the build requirements and libggc as runtime dep.

recipes/angsd/build.sh
```diff
@@ -0,0 +1,21 @@
...
+misc/realSFS \
+misc/smartCount \
+misc/supersim \
+angsd \
+"
+BINDIR=$PREFIX/bin
+mkdir -p $BINDIR
+
+
+make LDFLAGS="-L=${PREFIX}/lib" INC="-I=${PREFIX}/include"
```
johanneskoester: Could you use standard env vars like `CPLUS_INCLUDE` or `C_INCLUDE`. That could be a way to avoid the makefile patches.

bgruening:
@percyfal looks good to me, if you add gcc/llvm explicitly.

recipes/angsd/build.sh
percyfal: Done; I wasn't aware of these variables, and what worked in the end was CPLUS_INCLUDE_PATH and C_INCLUDE_PATH.

recipes/angsd/meta.yaml
percyfal: Yep, despite knowing about it, I forget it every time...

recipes/angsd/build.sh
```diff
@@ -0,0 +1,20 @@
...
+misc/NGSadmix \
+misc/realSFS \
+misc/smartCount \
+misc/supersim \
+angsd \
+"
+BINDIR=$PREFIX/bin
+mkdir -p $BINDIR
+
+make LDFLAGS="-L=${PREFIX}/lib" C_INCLUDE_PATH="-I=${PREFIX}/include" CPLUS_INCLUDE_PATH="-I=${PREFIX}/include"
```
johanneskoester: Most likely you only need either C_INCLUDE_PATH or CPLUS_INCLUDE_PATH. Further, you only specify the path in them, not the flag. Usually we define the env vars before running make, but it might work like this as well.

percyfal: I can rewrite it for consistency. In any case, the tests needs to be restarted - it just seems to have been cancelled.

PR #3048, opened 2016-11-23
pyBigWig-0.3.1

dpryan79:

### DIFF
--- a/recipes/pybigwig/meta.yaml
+++ b/recipes/pybigwig/meta.yaml
@@ -15,11 +15,11 @@ requirements:
   build:
     - python
     - curl
-    - numpy >=1.7.0
+    - numpy x.x
   run:
     - python
     - curl
-    - numpy >=1.7.0
+    - numpy x.x
 test:
   imports:
     - pyBigWig
--- a/recipes/pybigwig/meta.yaml
+++ b/recipes/pybigwig/meta.yaml
@@ -24,12 +24,6 @@ test:
   imports:
     - pyBigWig
 
-  commands:
-    - nosetests pyBigWigTest -s -v
-
-  requires:
-    - nose
-
 about:
   home: https://github.com/dpryan79/pyBigWig
   license: MIT
--- a/recipes/pybigwig/meta.yaml
+++ b/recipes/pybigwig/meta.yaml
@@ -1,13 +1,11 @@
 package:
   name: pybigwig
-  version: '0.3.0'
+  version: '0.3.1'
 
 source:
-  #fn: pyBigWig-0.3.0.tar.gz
-  #sha256: 03cebed08bc6d8b373e16f2dbe1c853c8d347f3d7ea284faedcd4c15ccf2bc36
-  #url: https://pypi.python.org/packages/59/23/5d77c02ea726c66963f5a38a7aa2a0620285f9f5ada441d09174153603e1/pyBigWig-0.3.0.tar.gz
-  git_url: https://github.com/dpryan79/pyBigWig
-  git_tag: small_fixes
+  fn: pyBigWig-0.3.1.tar.gz
+  sha256: 1103dc64eb20bf02e72f2909471a38070bf000eab7206e0c2dd3e4f62b683254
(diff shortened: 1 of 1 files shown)


recipes/pybigwig/meta.yaml
```diff
@@ -1,34 +1,30 @@
...
+  url: https://pypi.python.org/packages/c2/fc/a25347e31f546331ad5ceff73997d4976844fff5448f4d7007eaa49a29b2/pyBigWig-0.3.1.tar.gz
 
 build:
   number: 0
   skip: False
 
 requirements:
   build:
-    - python
+    - python x.x
```
bgruening: Is this still needed?

dpryan79: For some unknown reason it appears so.

bgruening:
@johanneskoester do you have more details about `x.x` vs. the normal python behavior? I'm busy with students today but could look into it tm.

dpryan79:
I'm playing around inside the busybox right now and, at least if `python x.x` isn't used, then `python -c "import numpy"` also segfaults. This hints at a more general problem.

johanneskoester:
As far as I know, Python is always pinned by default, regardless of whether x.x is added or not. It is a different story for numpy. There, x.x is needed when a package links agains the C interface of numpy. @dpryan79 can you show us the build string/full package name when x.x is added and when not?

dpryan79:
`pybigwig-0.3.1-np110py34_0` in both cases, which is why this is so odd. After playing around with simulate-travis.py, it turns out that if you `python x.x` then the import test in the `meta.yaml` is skipped and the test with mulled is effectively skipped. I don't know why that's done, but at least I know to close the PR until I figure out what's up with python 3.4, which is going to be tough since `python -c "import numpy"` is sufficient to cause a crash (at least when I try locally with simulate-travis).

dpryan79:
A (likely final) update on this. The cause is likely a broken interaction between python 3.4 (Linux only) from continuum and the container where the test is run. This python version segfaults for some reason that I can't discern (it's compiled without much in the way of debug symbols, so gdb is little help) if one tries to import numpy (i.e., it seems to have nothing to do with pyBigWig, since if I compile with numpy support then `python -c "import pyBigWig"` works, though `python -c "import numpy"` still segfaults). If I copy a different copy of python3.4 (not from continuum or conda-forge) into the container then it has no problems importing numpy.

If you're curious, the backtrace is:

    #0  0x00007feb0c795b52 in __vdsosym () from /lib64/libc.so
    #1  0x00007feb12a54ef8 in ?? ()
    #2  0x00007feb0c742e88 in ?? ()
    #3  0x00007feb0c7e00bf in ?? () from /lib64/libc.so
    #4  0x00007feb0c7e00aa in __gmt () from /lib64/libc.so
    #5  0x000000000116e7e8 in ?? ()
    #6  0x0000000000000000 in ?? ()

So there's some sort of issue when loading a shared object. Since this only happens inside of the docker container, my only guess is that it has some obscure problem with a system library (afterall, the same binaries and libraries work outside of the docker container).

Presumably other recipes will start running into this problem sooner or later.

daler:
Wow that sounds like a nightmare of a debugging session. Thanks so much for reporting back. If you have a reproducible test case we can open an issue with conda-forge (it's their docker container that we're using), maybe they have some ideas.

dpryan79:
I've created a `python34_numpy` branch that demonstrates the segfault. To save the hassle of finding it, the `meta.yaml` file (the only part of the recipe) is below. One can presumably remove curl, which is a hold-over from the pyBigWig recipe.

    package:
      name: python34_numpy
      version: '0.0.0'

    source:
      path: ./

    build:
      number: 0
      skip: True # [not py34]

    requirements:
      build:
        - python
        - numpy
        - curl
      run:
        - python
        - numpy
        - curl

    test:
      imports:
        - numpy

    about:
      home: https://github.com/dpryan79/pyBigWig
      license: MIT
      summary: This is just a demonstration of numpy segfaulting

PR #3050, opened 2016-11-24
Add pannenkoek

druvus:

### DIFF
--- a/recipes/pannenkoek/meta.yaml
+++ b/recipes/pannenkoek/meta.yaml
@@ -1,0 +1,46 @@
+package:
+  name: pannenkoek
+  version: "0.1.6"
+
+source:
+  fn: pannenkoek-0.1.6.tar.gz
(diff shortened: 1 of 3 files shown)


recipes/pannenkoek/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - qiime
+    - rpy2
+    - numpy
+    - matplotlib
+    - argparse
```
bgruening: Is argparse really needed? I guess it is annotated in the setup.py, but unless we do want to support python2.6 this should be removed from here and the setup.py

druvus: Good catch. I will fix it.

bgruening: I created a PR upstream, but maybe this can be patched out in the meantime?

druvus:
@bgruening seems that pannenkoeken is now maintained as koeken. Closing until this old version is needed.

PR #3058, opened 2016-11-26
Use one subdag by default and remove it entirely later on

bgruening:

@bioconda/core I was in contact with Travis the last weeks and could get an extended build-time of 2h. Thanks to Travis!

I think we can just use one subdag, aka only one DAG from now on. My hope is that this will increase our turnaround as we use less concurrent builds for one PR. Most of our PRs only have a few packages included and for large restructurings like R rebuilds I hope 2h is enough.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,12 @@ sudo: required
 services:
 - docker
 install: scripts/travis-setup.sh
-script: travis_wait 75 scripts/travis-run.sh
+script: travis_wait 120 scripts/travis-run.sh
 env:
   global:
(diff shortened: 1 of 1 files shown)


daler:
That's great Bjoern, thanks for coordinating!

johanneskoester:
Amazing, this is great! Let us keep the functionality though (aka just use one dag but don't remove the corresponding code). We might need it in the future in order to reactivate the weekly tests.

bgruening:
Thanks, let's try this the next weeks!

PR #3074, opened 2016-11-28
trying to build dsrc for osx

cokelaer:


recipes/dsrc/meta.yaml
```diff
@@ -1,5 +1,5 @@
 package:
-  name: dscr
+  name: dsrc
   version: 2014.12.17
```
bgruening: Can you please also add a checksum, like sha256? Thanks.

cokelaer: Hi, I've merged the pull request unfortunately. So, I'll create another one with the md5 is that's okay (instead of sha256)

PR #3075, opened 2016-11-29
nudup seems to be broken

bgruening:

@lparsons not sure why but nudup seems to be broken

https://travis-ci.org/bioconda/bioconda-recipes/jobs/179697505


lparsons:
@bgruening Hmmm... Seems like some sort of download error? Perhaps an issue with Github at the time the test was run? Will take a closer look when I get to a computer.

lparsons:
I don't see why this would fail, except that perhaps it was due to the change in the way the file is retrieved? Doesn't seem like this should be in the blacklist though, since it does seem to build in at least some cases. Can you trigger the travis build again to test?

daler:
#3080 fixed, I'll trigger a rebuild to see if that helps.

bgruening:
It should not be needed. This was the first time I saw the error and thought the package is broken on master. I will close this. nudup is now online @lparsons! biocontainer as well.

lparsons:
No problem, thanks.

PR #3078, opened 2016-11-30
Add MaxBin 2

colinbrislawn:

### DIFF
--- a/recipes/maxbin2/build.sh
+++ b/recipes/maxbin2/build.sh
@@ -1,9 +1,5 @@
 #!/bin/bash
 set -euo pipefail
 cp makefile.new src
-
-if [ `uname` == Darwin ]; then
-    export CXXFLAGS="${CXXFLAGS} -std=c++0x -stdlib=libc++"
-fi
 cd src && make -f makefile.new
 cp MaxBin $PREFIX/bin/
--- a/recipes/maxbin2/build.sh
+++ b/recipes/maxbin2/build.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 set -euo pipefail
 cp makefile.new src
+
+if [ `uname` == Darwin ]; then
+    export CXXFLAGS="${CXXFLAGS} -std=c++0x -stdlib=libc++"
+fi
 cd src && make -f makefile.new
 cp MaxBin $PREFIX/bin/
--- a/recipes/maxbin2/build.sh
+++ b/recipes/maxbin2/build.sh
@@ -2,4 +2,4 @@
 set -euo pipefail
 cp makefile.new src
 cd src && make -f makefile.new
-cp src/MaxBin $PREFIX/bin/
+cp MaxBin $PREFIX/bin/
--- a/recipes/maxbin2/build.sh
+++ b/recipes/maxbin2/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+set -euo pipefail
+make
--- a/recipes/maxbin2/build.sh
+++ b/recipes/maxbin2/build.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
 set -euo pipefail
-cd src && make
+cp makefile.new src
+cd src && make -f makefile.new
 cp src/MaxBin $PREFIX/bin/
--- a/recipes/maxbin2/build.sh
+++ b/recipes/maxbin2/build.sh
@@ -1,3 +1,4 @@
 #!/bin/bash
 set -euo pipefail
 cd src && make
+cp src/MaxBin $PREFIX/bin/
--- a/recipes/maxbin2/build.sh
+++ b/recipes/maxbin2/build.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 set -euo pipefail
-make
+cd src && make
--- a/recipes/maxbin2/meta.yaml
+++ b/recipes/maxbin2/meta.yaml
@@ -16,7 +16,7 @@ source:
     - makefile.patch
 requirements:
   build:
-    - gcc
+    - gcc >=4.8
   run:
     - libgcc
     - fraggenescan >=1.30
--- a/recipes/maxbin2/meta.yaml
+++ b/recipes/maxbin2/meta.yaml
@@ -12,6 +12,8 @@ source:
   fn: MaxBin-{{ version }}.tar.gz
   url: http://downloads.jbei.org/data/microbial_communities/MaxBin/getfile.php?MaxBin-{{ version }}.tar.gz
   md5: {{ md5 }}
+  patches:
+    - makefile.patch
 requirements:
   build:
     - gcc # [linux]
--- a/recipes/maxbin2/meta.yaml
+++ b/recipes/maxbin2/meta.yaml
@@ -1,13 +1,17 @@
+{% set name = "maxbin" %}
+{% set version = "2.2.1" %}
+{% set md5 = "3748e8f65b5bf18f8bd341210724cc5c" %}
+
 build:
   number: 0
 
 package:
-  name: maxbin2
-  version: "2.2.1"
+  name: {{ name }}
+  version: {{ version }}
 source:
-  fn: MaxBin-2.2.1.tar.gz
-  url: http://downloads.jbei.org/data/microbial_communities/MaxBin/getfile.php?MaxBin-2.2.1.tar.gz
-  md5: 3748e8f65b5bf18f8bd341210724cc5c
+  fn: MaxBin-{{ version }}.tar.gz
+  url: http://downloads.jbei.org/data/microbial_communities/MaxBin/getfile.php?MaxBin-{{ version }}.tar.gz
+  md5: {{ md5 }}
 requirements:
   build:
     - gcc # [linux]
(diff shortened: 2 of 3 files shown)


colinbrislawn:
When I build this package localy with `conda build .` it fails with this error. 
```
g++ -c -Wall -g -std=c++0x EManager.cpp -o EManager.o -I ./
In file included from EManager.cpp:1:
./EManager.h:12:10: fatal error: 'thread' file not found
#include <thread>
         ^
1 error generated.
make: *** [EManager.o] Error 1
```

When I manually move into the working directory and run `make`, it compiles successfully with some warnings. Strange...

The `#include <thread>` error seems to be related to C++11. Any ideas?

colinbrislawn:
I've noticed that @keuv-grvl has crafted excellent recipes for related projects. Do you have any advice for me on this build?

Thanks!

recipes/maxbin2/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+    - fraggenescan >=1.30
+    - hmmer
+    - idba
+test:
+  commands:
+    - MaxBin > /dev/null
+about:
+  home: "http://downloads.jbei.org/data/microbial_communities/MaxBin/MaxBin.html"
+  license: "BSD 3-clause"
+  summary: "MaxBin is software for binning assembled metagenomic sequences based on an Expectation-Maximization algorithm."
```
keuv-grvl: I find useful to add an `extra` category with the recipe maintainer(s) and the associated DOI pointing to the original article (if available). Here is an example: [abawaca](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/abawaca/meta.yaml#L29-L32)

keuv-grvl:
Now I remember why I crafted the FragGeneScan recipe!

`<thread>` is quite new and required GCC 4.6+. Moreover the `makefile` states `LDFLAGS:=-pthread`. According to my c++-enthusiast colleague, it should be `-lpthread`. He helped me rewrite the makefile. You can find it [here](https://gist.github.com/keuv-grvl/92f0e22e769dec533a8e69344777cc7c).

Also you may test your recipes using [`./simulate-travis.py`](https://github.com/bioconda/bioconda-recipes/blob/master/FAQs.md#test-locally) to avoid interfering with your OS.

recipes/maxbin2/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+set -euo pipefail
+cd src && make
```
colinbrislawn: I modeled cd src && make after the docker build file. I like your idea of copying the executable to the right location.

recipes/maxbin2/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+  run:
+    - libgcc # [linux]
+    - perl
+    - bowtie2
+    - fraggenescan >=1.30
+    - hmmer
+    - idba
+test:
+  commands:
+    - MaxBin > /dev/null
```
colinbrislawn: I did notice that the return code was 255. Usually I would test for installation using 
`MaxBin --help` or `MaxBin --version`, but the program only accepts real inputs. Any ideas?

recipes/maxbin2/meta.yaml
colinbrislawn: Added!

recipes/maxbin2/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+  fn: MaxBin-2.2.1.tar.gz
+  url: http://downloads.jbei.org/data/microbial_communities/MaxBin/getfile.php?MaxBin-2.2.1.tar.gz
+  md5: 3748e8f65b5bf18f8bd341210724cc5c
+requirements:
+  build:
+    - gcc # [linux]
+    - llvm # [osx]
+  run:
+    - libgcc # [linux]
+    - perl
```
colinbrislawn: I don't think my collaborators need perl, but I'll check. Hopefully we can drop this.

colinbrislawn:
Thank you for all your feedback. This is exactly the kind of careful review that this recipe needed. 

I've addressed as many suggestions as possible and added your approach to program versioning. 

What should we do about the makefile? Patch it, or perhaps submit a PR to maxbin?

daler:
@colinbrislawn some build system issues were fixed over in #3080 if you want to rebase and give this another shot.

recipes/maxbin2/meta.yaml
keuv-grvl: :tada:

recipes/maxbin2/meta.yaml
keuv-grvl: I had this problem with #2618 and we solved it by testing if the return code was 255.
Providing reads as input would make a functional test but this is not the bioconda's aim (cf. also #2618)

recipes/maxbin2/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+about:
+  home: "http://downloads.jbei.org/data/microbial_communities/MaxBin/MaxBin.html"
+  license: "BSD 3-clause"
+  summary: "MaxBin is software for binning assembled metagenomic sequences based on an Expectation-Maximization algorithm."
+
+extra:
+  recipe-maintainers:
+    - colinbrislawn
+    - keuv-grvl
+  doi: 10.1186/2049-2618-2-26
```
keuv-grvl: You could also add MaxBin 2.0 paper : [10.1093/bioinformatics/btv638](https://dx.doi.org/10.1093/bioinformatics/btv638)
``` yaml
doi:
  - 10.1186/2049-2618-2-26
  - 10.1093/bioinformatics/btv638
```

recipes/maxbin2/build.sh
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
 set -euo pipefail
 cd src && make
+cp src/MaxBin $PREFIX/bin/
```
keuv-grvl: Your script is already located in `src` at this point so `src/MaxBin` does not exist but `./MaxBin` does

colinbrislawn:
@keuv-grvl  Done! Thanks for all your feedback.

@daler I'll rebase this afternoon. Thank you for supporting this dev platform!

keuv-grvl:
I just try to run MaxBin (manually installed) using the `run_MaxBin.pl` script.

For the complete pipeline, the following files are required in `$PREFIX/bin/` :
- `run_MaxBin.pl`
- `_getabund.pl`
- `_getmarker.pl`
- `_sepReads.pl`
- `heatmap.r`
- `marker.hmm`
- `bacar_marker.hmm`
- `src/MaxBin`

The last one is compile from sources exactly like in your `build.sh`. The `run_MaxBin.pl` search for it in a `src` folder.

So if you plan to use the complete pipeline with `run_MaxBin.pl`, you will need to compile the sources then copy some files.

Hope this helps :)

colinbrislawn:
@keuv-grvl That does help! I'll make sure to copy files to all needed locations.

I've delayed this rebase while working on other projects, but also because I ran into an issue incorporating the `maxbin.makefile` that you and @Celforyon made. How can I get this into the `src` directory so I can use it during the build?

keuv-grvl:
I like to use patches because they are managed by Conda.

First, you build your patch:
``` bash 
cd /path/to/bioconda-recipe/recipes/maxbin/
wget https://gist.githubusercontent.com/keuv-grvl/92f0e22e769dec533a8e69344777cc7c/raw/01719d5e63118deee3bbfa0c75d56c36917f76d7/maxbin.makefile
diff -u /dev/null maxbin.makefile > makefile.patch # this compares an empty file to maxbin.makefile
# you can test your patch as follows:
patch < makefile.patch
```

Then you add it to your `meta.yaml`:
``` 
source:
  fn: ...
  url: ...
  md5: ...
  patches:
    - makefile.patch
```

Patches are applied before calling `build.sh`, so `maxbin.makefile` will exist.
In your `build.sh`:
``` bash
cp maxbin.makefile src/
cd src/
make -f maxbin.makefile
```

colinbrislawn:
Thanks to help from @keuv-grvl and @ajaypanyala!

Once this building, I think we are ready to merge.

keuv-grvl:
Travis failed for OSX [here (L360-L362)](https://travis-ci.org/bioconda/bioconda-recipes/jobs/194634556#L360-L362). It seems to come from the C++11 support. This requires additional parameters (see [here](http://stackoverflow.com/questions/14115314/how-to-compile-c11-with-clang-3-2-on-osx-lion)). But I bet these parameters are OSX specific, so you may need to set `CXX_FLAGS` only for OSX (something like [this](https://github.com/bioconda/bioconda-recipes/blob/013af80cae89e5c403cb101e622d6f5156b30bee/recipes/pbbam/build.sh#L5-L7)).

Otherwise, you can simply skip the build for OSX.

recipes/maxbin2/meta.yaml
```diff
@@ -1,4 +1,4 @@
-{% set name = "maxbin" %}
```
bgruening: This will change the entire package name, is this intended?

colinbrislawn: Oh no! Good catch.

colinbrislawn:
Hey @ajaypanyala, could you review this PR, then Squash and Merge this PR?
Thanks!

PR #3089, opened 2016-12-02
[WIP] add recipe for MaxEntScan

matthdsm:

@bioconda/core 
Hi, 
I'm going to need some help with this. The included scripts depend on some reference data that has to be in the same directory as the executable. How do I best get around this? In it's current form, the data just gets copied to the bin directory, but I seriously doubt that's the way to go.

Thanks!
M

### DIFF
--- a/recipes/maxentscan/build.sh
+++ b/recipes/maxentscan/build.sh
@@ -2,6 +2,9 @@ target=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
 mkdir -p $target
 mkdir -p $PREFIX/bin
 
+# add perl shebang
+sed -i.bak "1i #!/usr/bin/env perl" *.pl
+
 #change hardcoded directory of splicemodels
 sed -i.bak "s@\"splicemodels/\"@\"$target/splicemodels/\"@" *.pl
 
--- a/recipes/maxentscan/build.sh
+++ b/recipes/maxentscan/build.sh
@@ -6,11 +6,14 @@ mkdir -p $PREFIX/bin
 sed -i.bak "1i #!/usr/bin/env perl" *.pl
 
 #change hardcoded directory of splicemodels
-sed -i.bak "s@\"splicemodels/\"@\"$target/splicemodels/\"@" *.pl
+sed -i.bak "s@\"splicemodels/\"@\"$target/splicemodels/\"@" score3.pl
+sed -i.bak "s@'me2x5'@'$target/me2x5'@" score5.pl
+sed -i.bak "s@'splicemodels/splice5sequences'@'$target/splicemodels/splice5sequences'@" score5.pl
 
 cp score3.pl $target/maxentscan_score3.pl
 cp score5.pl $target/maxentscan_score5.pl
 cp -r splicemodels $target
+cp me2x5 $target/me2x5
 
 chmod 0755 $target/*.pl
 ln -s $target/*.pl $PREFIX/bin
--- a/recipes/maxentscan/build.sh
+++ b/recipes/maxentscan/build.sh
@@ -1,0 +1,5 @@
+mkdir -p $PREFIX/bin
+cp score3.pl $PREFIX/bin/maxentscan_score3.pl
+cp score5.pl $PREFIX/bin/maxentscan_score5.pl
+cp -r splicemodels $PREFIX/bin
+chmod 0755 $PREFIX/*
--- a/recipes/maxentscan/build.sh
+++ b/recipes/maxentscan/build.sh
@@ -1,5 +1,15 @@
+target=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+mkdir -p $target
 mkdir -p $PREFIX/bin
-cp score3.pl $PREFIX/bin/maxentscan_score3.pl
-cp score5.pl $PREFIX/bin/maxentscan_score5.pl
-cp -r splicemodels $PREFIX/bin
-chmod 0755 $PREFIX/*
+
+#change hardcoded directory of splicemodels
+sed -i.bak "s@\"splicemodels/\"@\"$target/splicemodels/\"@" *.pl
+
+cp score3.pl $target/maxentscan_score3.pl
+cp score5.pl $target/maxentscan_score5.pl
+cp -r splicemodels $target
+
+chmod 0755 $target/*.pl
+ln -s $target/*.pl $PREFIX/bin
+
+
--- a/recipes/maxentscan/meta.yaml
+++ b/recipes/maxentscan/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: maxentscan
-  version: "2004.04.21"
+  version: "0_2004.04.21"
 source:
   fn: MaxEntScan.tar.gz
   url: http://genes.mit.edu/burgelab/maxent/download/fordownload.tar.gz
--- a/recipes/maxentscan/meta.yaml
+++ b/recipes/maxentscan/meta.yaml
@@ -2,7 +2,7 @@ package:
   name: maxentscan
   version: "1.0"
 source:
-  fn: fordownload.tar.gz
+  fn: MaxEntScan-1.0.tar.gz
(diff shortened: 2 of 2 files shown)


recipes/maxentscan/meta.yaml
```diff
@@ -0,0 +1,21 @@
+package:
+  name: maxentscan
+  version: "1.0"
+source:
+  fn: fordownload.tar.gz
+  url: http://genes.mit.edu/burgelab/maxent/download/fordownload.tar.gz
```
bgruening: Checksum is missing.

bgruening:
How big are the reference models? I would store them under $PREFIX/share/ and symlink them into $PREFIX/bin if they needs to be next to the scripts.

johanneskoester:
I'd vote for not polluting $PREFIX/bin. Maybe you can add a patch to the recipe that fixes the path to `../share/maxentscan`?

matthdsm:
Hi @bgruening 
The reference set used is only 4mb, so size shouldn't be much of an issue, but as @johanneskoester said, I'd like to avoid polluting the `bin` directory.

Any pointers as to how I should write a patch and how to move the data to the `share` dir? I'm still a bit at a loss regarding the available variables etc.

Thanks!
M

recipes/maxentscan/meta.yaml
matthdsm: added checksum!

johanneskoester:
The target location for share would be `$PREFIX/share/maxentscan`. You can simply create that directory and then use cp or mv.

The patch is just a usual patch created with the diff command. Basically, you modify the file with the hardcoded path to your needs, then you use diff to create the patch. Finally you add the patch to the recipe dir and refer to it in the meta.yaml as outlined in the conda manual.

matthdsm:
I pushed some modifications to the build script. These should fix the issues and avoid pollution of the bin directory. If you guys think it's okay, please go ahead and merge.

Thanks for the pointers!
M

recipes/maxentscan/meta.yaml
```diff
@@ -0,0 +1,23 @@
+package:
+  name: maxentscan
+  version: "1.0"
+source:
+  fn: fordownload.tar.gz
```
johanneskoester: Can you rename to a nicer filename? E.g. maxentscan-1.0.tar.gz.

recipes/maxentscan/build.sh
```diff
@@ -0,0 +1,15 @@
+target=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+mkdir -p $target
+mkdir -p $PREFIX/bin
+
+#change hardcoded directory of splicemodels
+sed -i.bak "s@\"splicemodels/\"@\"$target/splicemodels/\"@" *.pl
+
+cp score3.pl $target/maxentscan_score3.pl
+cp score5.pl $target/maxentscan_score5.pl
```
johanneskoester: If you change the name, which is reasonable, you should add that to extra->notes (see e.g. [here](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/gatk/meta.yaml)).

recipes/maxentscan/build.sh
```diff
@@ -0,0 +1,15 @@
...
+cp -r splicemodels $target
+
+chmod 0755 $target/*.pl
+ln -s $target/*.pl $PREFIX/bin
```
johanneskoester: What are those perl files that are linked into bin? Do they follow the same naming scheme as above? Are they all standalone executables?

recipes/maxentscan/meta.yaml
```diff
@@ -0,0 +1,23 @@
...
+    number: 0
+    skip: True # [osx]
+requirements:
+  build:
+    - perl-threaded
+  run:
+    - perl-threaded
+test:
+  commands:
+    - perl maxentscan_score3.pl
```
johanneskoester: The tests should work without adding perl. If not, the shebang line of the scripts has to be fixed.

recipes/maxentscan/meta.yaml
matthdsm: Hi Johannes,

Sadly, I cannot, as I have no say, nor any relation to the people who wrote the scripts. I am merely trying to make it easily installable using conda :/

johanneskoester: Actually conda renames it for you if you specify another filename in fn. That is the purpose of fn.

recipes/maxentscan/build.sh
matthdsm: The linked perl files are the actual scripts. They should be able to act as standalone executables. Is this the wrong way to do it? The linking and naming etc was heavily influenced (copied) by the recipe for variant-effect-predictor.

matthdsm: yes they do, the actual files are renamed during the copy to the target directory
```
cp score3.pl $target/maxentscan_score3.pl
cp score5.pl $target/maxentscan_score5.pl
```

matthdsm:
if pushed fixes for the requested changes. Again, if you thinks this'll work, go ahead and merge!

Thanks for the discussion @johanneskoester !

johanneskoester:
Looks like the md5 of the tarball is wrong. Actually I am now also worried about the download link, because it does not contain any version information. How do the authors handle updates? Also, where did you get the version number from? I can't find it in the FTP. Is it mentioned in the source code?

matthdsm:
Hi Johannes,
To be honest, there is no real version number, so I just named it 1.0. Should I handle this differently? This is some pretty stale code, and hasn't changed in at least 10 years.

I fixed the checksum, should be OK now

johanneskoester:
Well, that is a good question... @bioconda/core any opinion about this?

matthdsm:
perhaps I could go for something like [this](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/vcf2db/meta.yaml)? And as version I'll just use the last modified date in the ftp, so that would be `2004.04.21`

johanneskoester:
Yes, that's better I think.

matthdsm:
there still seems to be something wrong with the tests, I'm checking it out right now

bgruening:
Does is makes sense to prefix is with 0 ... something like `0_2004.04.21`. Maybe this tool gets a version and than we need to overwrite our version easily.

johanneskoester:
I agree.

matthdsm:
fixed! 
Any idea why the checks keep on failing?

johanneskoester:
I think this is the error: https://travis-ci.org/bioconda/bioconda-recipes/jobs/181651482#L552.

matthdsm:
hi @bgruening, @johanneskoester 
Any idea where the build goes wrong? I can't really make much of the travis logs, and the recipe builds cleanly in a local docker based on continuumio/miniconda

Thanks
M

johanneskoester:
After the regular build, we also test by building a busybox-based container image and running the test commands. This works by an external tool called mulled. My guess is that this tool does not see that your tests need source_files, and therefore the commands fail. @bgruening do you agree? Would you say this can be easily fixed in mulled?
Apart from that, so far our policy was that tests don't require additional files. We usually just check if a tool runs, e.g. by displaying the help or version info.

In summary, I think you have two options here:
1. Change the tests such that no additional files are needed.
2. Wait for a fix in mulled.

matthdsm:
I suppose I'll have to wait for that fix then.. The scripts don't give a proper exit status when I run them without test data. I thought test data was an officially supported function as of conda 2.0?

johanneskoester:
That might be, but as I said, we have to wait for mulled to adapt to that. Let's see what @bgruening  says.

daler:
@matthdsm what error code do the scripts exit with when no data are provided? If it's not 127 (i.e. command not found) then you can explicitly test for the error code it does provide. That will at least test that it has been installed. For example the UCSC tools exit 255, and many tools exit 1. 

```bash
programName || [[ $? == 255]]
```

matthdsm:
Allright @daler, thanks for the tip!

bgruening:
SRC_DIR is not available during mulled build, because it is testing only the installed binaries. No way to have access to the SRC_DIR. I would vote to include a test that is not dependent on the SRC and not using any additional file.

PR #3094, opened 2016-12-02
Recipe for MetaProb 2

keuv-grvl:

`./simulate-travis.py` is not working on my computer, so I could not test locally before this PR :-(

### DIFF
--- a/recipes/metaprob/meta.yaml
+++ b/recipes/metaprob/meta.yaml
@@ -1,0 +1,54 @@
+{% set name = "MetaProb" %}
+{% set version = "2" %}
+{% set sha256 = "ada049a1a9fde3ddd9854230af257ee495518a45f0ce2504272f7bcdc060717b" %}
+
+package:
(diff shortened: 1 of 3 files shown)


recipes/metaprob/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+  url: https://bitbucket.org/samu661/metaprob/downloads/{{ name }}_v{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+  md5: {{ md5 }}
+  patches:
+    - makefile.patch
+
+build:
+  number: 0
+  skip: True  # [ osx ]
+
```
druvus: I think you should add: `string: boost{{CONDA_BOOST}}_{{PKG_BUILDNUM}}`

@bgruening could you confirm?

bgruening: Oh yes, can confirm :)
@keuv-grvl can you add this please?

PR #3095, opened 2016-12-02
Recipe for LatentStrainAnalysis

keuv-grvl:

### DIFF
--- a/recipes/latentstrainanalysis/meta.yaml
+++ b/recipes/latentstrainanalysis/meta.yaml
@@ -44,7 +44,7 @@ test:
     - ReadPartitioning.sh | grep "Usage:"
 
 about:
-  home: https://github.com/douglasgscofield/LatentStrainAnalysis
+  home: https://github.com/brian-cleary/LatentStrainAnalysis
   license: "MIT"
   summary: "Partitioning and analysis methods for large, complex sequence datasets"
   description: |
--- a/recipes/latentstrainanalysis/meta.yaml
+++ b/recipes/latentstrainanalysis/meta.yaml
@@ -11,10 +11,11 @@ source:
 
 build:
   number: 0
+  skip: True # [py3k]
 
 requirements:
   build:
-    - python 2.7*
+    - python
     - numpy
     - scipy
     - gensim
@@ -22,7 +23,7 @@ requirements:
     - pyro4
 
   run:
-    - python 2.7*
+    - python
     - numpy
     - scipy
     - gensim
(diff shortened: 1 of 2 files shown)


recipes/latentstrainanalysis/meta.yaml
```diff
@@ -0,0 +1,52 @@
+{% set name = "LatentStrainAnalysis" %}
+{% set version = "1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  git_rev: 4db66e9
+  git_url: https://github.com/douglasgscofield/{{ name }}.git
```
bgruening: Can you please use a tarball and a checksum instead of git_url

recipes/latentstrainanalysis/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - python 2.7*
```
bgruening: only `python`
and add `skip: True # [py3k]` into the build section

recipes/latentstrainanalysis/meta.yaml
keuv-grvl: There is no release in this repo. Downloading the `master` branch (via `https://github.com/douglasgscofield/LatentStrainAnalysis/archive/master.zip`) could fail if there are major changes on `master`. Is there any other solution?

druvus: Any special reason you are using Douglas' fork? I think it would be better to select the original repo. In that case you can use `https://github.com/brian-cleary/LatentStrainAnalysis/archive/7f1846849a4fe374897de70b55bd65662ba18912.ta.gz` as download link. 

You can always build the download link like that: gitrepo/projectname/archive/commit_id.tar.gz

bgruening: ping @keuv-grvl

recipes/latentstrainanalysis/meta.yaml
```diff
@@ -0,0 +1,60 @@
+{% set name = "LatentStrainAnalysis" %}
+{% set version = "1" %}
```
druvus: Is this an arbiratary version number or stated in the code? Maybe better to set it to `0.0.1` in case they will make a release below 1 later.

recipes/latentstrainanalysis/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  commands:
+    - command -v HashCounting.sh
+    - command -v KmerSVDClustering.sh
+    - command -v ReadPartitioning.sh
+    - HashCounting.sh | grep "Usage:"
+    - KmerSVDClustering.sh | grep "Usage:"
+    - ReadPartitioning.sh | grep "Usage:"
+
+about:
+  home: https://github.com/douglasgscofield/LatentStrainAnalysis
```
druvus: This should be updated to `https://github.com/brian-cleary/LatentStrainAnalysis`

druvus:
thanks @keuv-grvl !

PR #3098, opened 2016-12-02
Recipe for Metabat 0.32.4

keuv-grvl:

Compilation from sources was really painful. I chose to use the static binaries.

### DIFF
--- a/recipes/metabat/meta.yaml
+++ b/recipes/metabat/meta.yaml
@@ -1,0 +1,63 @@
+{% set name = "metabat" %}
+{% set version = "0.32.4" %}
+{% set sha256linux = "6db36f8d43e4e7d246c4f5028979952d8f29ed5fe69ae8175fb4f2be98cb9a22" %}
+{% set sha256osx = "d17fcee3891f2f478379edff30c5eee98dd8d76b2d876859f12087828abb142a" %}
+
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+build:
+  number: 0
+
+source:
+  # Linux binary
+  url: https://bitbucket.org/berkeleylab/{{ name }}/downloads/{{ name }}-static-binary-linux-x64_v{{ version }}.tar.gz  # [linux]
+  sha256linux: {{ sha256linux }}  # [linux]
+
+  # OSX binary
+  url: https://bitbucket.org/berkeleylab/{{ name }}/downloads/{{ name }}-static-binary-osx-elcapitan-x64_v{{ version }}.tar.gz  # [osx]
+  sha256: {{ sha256osx }}  # [osx]
+
+
+requirements:
+  build:
+
+  run:
+    - libgcc  # [not osx]
+    - perl
+    - boost
+    - zlib
+
+test:
+  commands:
+    - metabat --help 2> /dev/null
+    - jgi_summarize_bam_contig_depths 2> /dev/null
+    - contigOverlaps 2> /dev/null
+    - aggregateContigOverlapsByBin.pl 2> /dev/null || [ "$?" == "255" ]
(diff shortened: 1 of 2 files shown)


recipes/metabat/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  name: {{ name }}
+  version: {{ version }}
+
+build:
+  number: 0
+
+source:
+  # Linux binary
+  url: https://bitbucket.org/berkeleylab/{{ name }}/downloads/{{ name }}-static-binary-linux-x64_v{{ version }}.tar.gz  # [linux]
+  sha256linux: {{ sha256linux }}  # [linux]
```
bgruening: linux?

recipes/metabat/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+  # OSX binary
+  url: https://bitbucket.org/berkeleylab/{{ name }}/downloads/{{ name }}-static-binary-osx-elcapitan-x64_v{{ version }}.tar.gz  # [osx]
+  sha256: {{ sha256osx }}  # [osx]
```
bgruening: Can you compile this from source and not use the binaries?

recipes/metabat/meta.yaml
keuv-grvl: Bad reuse of an old variable...

recipes/metabat/meta.yaml
keuv-grvl: Compiling from sources requires the Samtools sources and Boost.
The default makefile automatically download and build samtools (1.2) on its own. This won't be maintainable for long term. A proper solution would be to use samtools as a independent lib but this would involve major changes in the software architecture.
Boost is searched in folder like `/usr/lib`, but a simple patch could fix that.
Regarding these, I found that using static binaries is much more simple and maintainable (but depends on software authors).

bgruening: The problem with these binaries is, that they are probably not compiled on an old CentOS and therefor will not work on clusters and HPC, which tend to be very conservative.

keuv-grvl: Good point!
This software uses SCons as construction tool and I don't know it. I'll try to make something better than using the binaries

druvus: @keuv-grvl Scons exist as an anaconda package so you build it according their instructions. I also have an interest in those metagenomic software you started to add so let me know if you would like some help to get them fixed.

keuv-grvl: @druvus I tried using SCons in a first draft of this recipe, but it could not find libboost. I'll keep investigating. The SConstruct file is pretty obscure for the moment.

keuv-grvl: OK, I think I got it.
The compilation procedure is pretty ugly! Briefly, it downloads and compiles `samtools 1.2` to get `libbam.a` and `libhts.a` and some `*.h` at a specific location. Then the 3 executables (`contigOverlaps`, `jgi_summarize_bam_contig_depths` and `metabat`) are compiled are we're done.
Problems are :
- `boost 1.60.0` installs `icu 56` and we need `58` to compile the sources
- the SConsctruc file will be hard to patch, I'm thinking of rewriting a new Makefile (not hard if `libbam.a` and `libhts.a` are available)

HTSlib is already available via bioconda. `libbam.a` could be generate from Samtools, so I'll propose a PR for samtools which also build `libbam.a`

druvus: Could you set the icu version (see discussion and recipe in #3720)? 

I think`libbam.a`was included in older samtools (<1.0).

johanneskoester:

PR #3103, opened 2016-12-03
Add recipe for estscan version 3.0

gregvonkuster:

### DIFF
--- a/recipes/estscan/3.0/meta.yaml
+++ b/recipes/estscan/3.0/meta.yaml
@@ -13,7 +13,6 @@ build:
 
 requirements:
   build:
-    - libgfortran
     - gcc
   run:
     - blast-legacy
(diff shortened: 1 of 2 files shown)


recipes/estscan/3.0/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  url: https://sourceforge.net/projects/estscan/files/estscan/3.0/estscan-3.0.tar.gz
+  md5: 1252c862b41b03b59687f5ba148b77cd
+
+build:
+  number: 0
+  skip: True #[osx]
+
+requirements:
+  build:
+    - libgfortran
```
bgruening: Have you tried without libgfortran. I think this is not needed if you use GCC.

gregvonkuster:
@bgruening I've commented out fortran components compiled with F77 since this recipe does not use them.  This has allowed me to eliminate the fortran requirement,

bgruening:
LGTM!

gregvonkuster:
Thanks @bgruening ;)

PR #3108, opened 2016-12-04
Fix recipe requirements for version 0.2 of plant_tribes_assembly_post…

gregvonkuster:

…_processor

### DIFF
--- a/recipes/plant_tribes_assembly_post_processor/0.2/meta.yaml
+++ b/recipes/plant_tribes_assembly_post_processor/0.2/meta.yaml
@@ -21,6 +21,9 @@ requirements:
 test:
   commands:
     - AssemblyPostProcesser 2>&1 | grep Usage
+    - estscan -h 2>&1 | grep Available
+    - gt -version 2>&1 | grep GenomeTools
+    - 'TransDecoder.Predict -h 2>&1 | grep USAGE'
 
 about:
   home: 'https://github.com/dePamphilis/PlantTribes'
--- a/recipes/plant_tribes_assembly_post_processor/0.2/meta.yaml
+++ b/recipes/plant_tribes_assembly_post_processor/0.2/meta.yaml
@@ -21,9 +21,6 @@ requirements:
 test:
   commands:
     - AssemblyPostProcesser 2>&1 | grep Usage
-    - estscan -h 2>&1 | grep Available
-    - gt -version 2>&1 | grep GenomeTools
-    - 'TransDecoder.Predict -h 2>&1 | grep USAGE'
 
 about:
   home: 'https://github.com/dePamphilis/PlantTribes'
--- a/recipes/plant_tribes_assembly_post_processor/0.2/meta.yaml
+++ b/recipes/plant_tribes_assembly_post_processor/0.2/meta.yaml
@@ -8,15 +8,15 @@ source:
   md5: 06c7fdb7eaf091a8fe9f826866a1a206
 
 build:
-  number: 0
+  number: 1
   # Requires genometools which is not supported on osx
   skip: True # [osx]
 
 requirements:
   run:
-    - perl-estscan2 >=2,<3
-    - transdecoder >=2,<3
-    - genometools-genometools >=1,<2
+    - estscan
+    - transdecoder
+    - genometools-genometools
 
 test:
   commands:


bgruening:
When the requirements was wrong in the first version, why did the tests passed? And why does the test pass now? Should the test be extended?

gregvonkuster:
@bgruening there are several underlying dependencies for this tool, and ESTscan is the one that I've had to correct.  The tests in this recipe only test for the basic features of the  PlantTribes AssemblyPostProcesser perl script, but do not test for the more advanced features that use the 3rd party dependencies like ESTScan.  If I enhance the tests to test for these underlying dependencies, I'll have to add test datasets which I understand we are restricting to the Galaxy functional test framework.  I'll make sure that these advanced dependencies are included into the Galaxy tool's functional tests as much as is possible.

bgruening:
Yes, correct we don't want to test with test-data. I was just wondering if you can import a module or `which` the existence of a binary, `find` a file that needs to be there ...

If this is not possible, please feel free to merge.

gregvonkuster:
@bgruening  I am not able to merge into this repository, so can you please do it?  I don't necessarily need to be authorized to merge unless you feel I should. ;)

Regarding the additional tests you're wondering about, this Perl module calls these 3rd party dependencies, sending command line parameters that require data files for processing.  I suppose I could investigate a bit and find out if a command line can be used that would work for a simple test that does not require a dataset.  Another option would be to simply include the tests that are defined in the recipe for each of the dependencies.  What is the standard for this scenario?

gregvonkuster:
@bgruening I've added tests to cover accessibility of the underlying dependencies here https://github.com/bioconda/bioconda-recipes/pull/3108/commits/a16ae0dc2e72b9bf685ec8a6ebf380e67c8b161b

gregvonkuster:
@bgruening I'm starting to wonder if these travis timeout issues are due to my recipes having run requirements that are not included in the build section.  Is there a different timeout setting for run vs build?  If so, could my tests be timing out because the amount of time my recipes require at runtime is too much?

gregvonkuster:
@bgruening my multiple failing recipes have been blocking me for quite some time now.  I'm able to add many tests that all pass locally, but nothing works with travis tests.  I've reverted the test on this recipe to the original test that passed yesterday, and I'm hoping that it still passes now.  If so, can you please merge this recipe for me?  I don't have write access on this repository.  If you agree to grant me write privileges, I'll be responsible for merging my own PRs, and hopefully will bug you a lot less.  ;)

bgruening:
Actually, with your tests your nearly passed everything, but the container test. We install everything in a busybox and run your tests again. This is failing here with:

```
[Dec  5 18:56:52] STEP Run image [quay.io/biocontainers/plant_tribes_assembly_post_processor:0.2--1] with command [[AssemblyPostProcesser 2>&1 | grep Usage && estscan -h 2>&1 | grep Available && gt -version 2>&1 | grep GenomeTools && TransDecoder.Predict -h 2>&1 | grep USAGE]]
[Dec  5 18:56:52] DEBU Creating container [step-69fae522c9]
[Dec  5 18:56:52] DEBU Created container [a39afcda471c step-69fae522c9], starting it
[Dec  5 18:56:52] DEBU Container [a39afcda471c step-69fae522c9] started, waiting for completion
[Dec  5 18:56:52] SOUT #  Example Usage:
[Dec  5 18:56:52] SOUT Available options (default value in braces[]):
[Dec  5 18:56:52] ERRO Task processing failed: Unexpected exit code [1] of container [a39afcda471c step-69fae522c9], container preserved
```

https://travis-ci.org/bioconda/bioconda-recipes/jobs/181380509#L1781

gregvonkuster:
@bgruening Can you please let me know what I need to do to get this recipe merged as soon as possible?  I'm at the point where I am going to have to figure out how to create my own channel, which I would rather not do.  I feel that if the standard is set that developers all create their own channels, it will cause problems because other developers will have no idea the channels required to retrieve packages.

gregvonkuster:
I'm using my own channel for this recipe, so closing.

bgruening:
I don't know this tool, but this is what the tool produced " Available options (default value in braces[]):" and it returned an error code of 1.

PR #3115, opened 2016-12-05
update locarna recipe (1.8.12)

s-will:

Try a fresh start :)


bgruening:
Seems that libgcc is not available as `la` file: https://travis-ci.org/bioconda/bioconda-recipes/jobs/181306952

Has something changed recently in this regard? Do you need libtool? Can you not compile the static version?

s-will:
no this didn't change recently - also building statically does not necessarily help. But I have an idea what to test (since I did quite strange things concerning stdlibc++ in order to make this run ...)

PR #3133, opened 2016-12-07
SearchGui copy the jar file, lib, and resources to cwd

jj-umn:

 SearchGui updates files relative to the path of the jar file.
Thus we need to copy the jar file, lib, and resources to the
working directory when this script is first invoked in a job.


bgruening:
@jj-umn can you do this in your wrapper and not in the conda package? What is the advantage. Moreover, I we should make Marc aware of this. I thought this was fixed already ... :(

bgruening:
There is something wrong with your last merge. It seems you have too much packages in this PR.

PR #3138, opened 2016-12-07
Minor perl-btlib recipe cleanup

gregvonkuster:


recipes/perl-btlib/0.19/build.sh
```diff
@@ -11,7 +11,6 @@ elif [ -f Makefile.PL ]; then
     # Make sure this goes in site
     perl Makefile.PL INSTALLDIRS=site
     make
-    make test
```
bgruening: Why is this needed? I think it is actually good to have it, it increase the testing of the package isn't it?

PR #3139, opened 2016-12-07
Remove unnecessary run dependencies for estscan version 3.0 recipe

gregvonkuster:

The installation documentation here https://sourceforge.net/projects/estscan/files/documentation/Feb-2007/ is confusing in that it seems to state that these run-time dependencies are required just to execute this C version of estscan.  But these additional packages are only needed to use estscan features that this recipe does not cover (e.g. generating new scores matrices files).  In any case, neither of the perl-estscan components are needed here.


bgruening:
You need to increase the build number :)

gregvonkuster:
Ah, done.  Thanks!

PR #3145, opened 2016-12-07
Rnasnp 1.2 updated

zwanli:


bgruening:
@bioconda/core  can you look over this one?

johanneskoester:
Looks good. But @bgruening I am suprised that mulled works here. The test case uses files from the recipe directory...

bgruening:
I wonder if the test is actually working? But the way to set ENV vars is ok?

johanneskoester:
Yes. That's fine. I did the same recently for the new openjdk package on conda-forge.

johanneskoester:
Does mulled run the test with bash -e? If not, it could be that it fails at the second command, but goes on until the third, which passes again.

bgruening:
Mulled does nothing fancy, it takes the command and passes it to the container. Isn't `set +e` supposed to fail if the second command fails?

johanneskoester:
You are right, I somehow did not see that... I'd think it should fail then...

zwanli:
I added the `echo` command because RNAsnp exits with code 160, even though it outputs the right results. I tried the tool with many different input data and every time I get the right results but it exits with code 160. You can see here [L790](https://travis-ci.org/bioconda/bioconda-recipes/jobs/182228220#L790) that after you removed the final echo it outputs the right results but the test fails.

johanneskoester:
You can check for the exit code to be 160. It is in the variable `$?`.

johanneskoester:
@bgruening, actually, `set +e` turns of the behavior to exit on error.

bgruening:
@zwanli that's fine and to be clear this recipe is of very high quality - so thanks a bunch!
We were a little bit puzzled, because we implemented a second test (next to the usual conda test) in bioconda that puts the entire package into a container and runs the command in this container. At this time no conda package manager is part of the container and actually your two test files are also not part of the container, so this test should fail - but it wasn't :)

This was at least making me nervous :)

No matter how you change the command, until you need the additional file the test will fail in the second container based test.

Following idea to make it work. Your test files are small, can you create them on the fly as part of the test command? Something like `echo '>my_fasta_file\n acgtcggtatt' > foo.fa && echo 'snp' > bar.snp && RNAsnp -arg -arg`?

bgruening:
@johanneskoester where do you found this?
Afaik: `set -e causes the shell to exit if any subcommand or pipeline returns a non-zero status.`

johanneskoester:
@bgruening, exactly, `set -e` activates the desired behavior, like you say. `set +e` deactivates it. See here: http://redsymbol.net/articles/unofficial-bash-strict-mode/

bgruening:
Oh, I did not see it was `+`, this explains a lot!

zwanli:
Thank you @bgruening  :)
I have been trying to find out why it exits with 160. I looked inside the tool's source code. I searched for exit code 160 meaning but I couldn't find anything.
The only solution that worked for me is to add `set +e`, use `;` after calling RNAsnp, and then use `echo 'test'` to get exit code 0.
Now I'm creating the test files on the fly.

bgruening:
Awesome! Really really nice!

recipes/rnasnp/meta.yaml
```diff
@@ -8,18 +8,19 @@ source:
...
     - llvm  # [osx]
+
   run:
     - libgcc    # [not osx]
 
 test:
-  commands:
-    - RNAsnp -h >/dev/null
+    commands:
+        - set +e && printf "A3G" > snp.txt && printf ">gi|56682960|ref|NM_000146.3| Homo sapiens ferritin, light polypeptide (FTL), mRNA\nGCAGTTCGGCGGTCCCGCGGGTCTGTCTCTTGCTTCAACAGTGTTTGGACGGAACAGATCCGGGGACTCTCTTCCAGCCTCCGACCGCCCTCCGATTTCCTCTCCGCTTGCAACCTCCGGGACCATCTTCTCGGCCATCTCCTGCTTCTGGGACCTGCCAGCACCGTTTTTGTGGTTAGCTCCTTCTTGCCAACCAACCA" > seq.fa && RNAsnp -f seq.fa -s snp.txt; echo "RNAsnp exite_code " $?
```
johanneskoester: Actually I have some doubts here. Imagine that the tool fails with a linkage error. This will yield some other exit code, but the test will still proceed.

zwanli:
Thank you @bgruening and @johanneskoester :)

PR #3148, opened 2016-12-08
Sparse assembler

mictadlo:


druvus:
Seems you ned to fix md5.

mictadlo:
I change md5 sum, but it still has failed and I do understand why.

tetianakh:
@mictadlo it fails because it can't find the `./install.sh` script, which you execute in the `recipes/masurca/build.sh`.

PR #3157, opened 2016-12-08
update viral-ngs 1.12.2 -> 1.13.2

tomkinsc:


tomkinsc:
@bioconda/core: It looks like the [Travis job](https://travis-ci.org/bioconda/bioconda-recipes/jobs/182318452) for this is being killed because the log file size hits 4MB before reaching the recipe test section. I'm not quite sure how to rectify this.

johanneskoester:
Some time in the last days, the debug logging has been activated (@daler?). Can we revert that step? Has the issue that led to that move been resolved?

daler:
@johanneskoester @tomkinsc this is actually the mulled output that's creating 4MB of log, since  `--loglevel=info` is still set in `.travis.yml`. @bgruening is there a good way to control mulled-build output?

tomkinsc:
If there is a way to disable mulled-build, that would be an option as well since we distribute a custom Docker image for viral-ngs. @bgruening?

daler:
@tomkinsc you can temporarily delete `--mulled-test` from this line in `.travis.yml`: https://github.com/bioconda/bioconda-recipes/blob/master/.travis.yml#L15

tomkinsc:
Thanks, @daler. How should I avoid merging that change in?

daler:
@tomkinsc I think the way to do it would be a chaser PR that does nothing but add `--mulled-build` back in. Seems hacky though, and even if you distribute a separate container it means there's not one in biobuilds. Do you know what's causing the mulled-build error (e.g., from running locally)?

tomkinsc:
A chaser PR does feel hacky. In general, I think we should have an option to opt-out of being included in mulled (for cases like this). Not sure why mulled-build is failing, only that it's hitting Travis's 4MB log limit.

bgruening:
Sorry to be so late to the game. This PR is building also 2h isn't it? The problem is not the 4MB the question is rather why does this take 2h?

Looking at this log https://api.travis-ci.org/jobs/182318209/log.txt I see the following:

* the vast majority of output is produced by downloading the packages, this we can repress I think
* two packages where build: viral-ngs & estscan (this can be changed by rerunning, I guess this one was fresh in master at this point)

The reason this PR is not running through is that the test is not compatible with the mulled testing.

```
[Dec  8 19:26:23] STEP Run image [quay.io/biocontainers/viral-ngs:1.13.0--py34_0] with command [[export CONDA_DEFAULT_ENV_BAK=$CONDA_DEFAULT_ENV && export CONDA_DEFAULT_ENV=$(echo $CONDA_DEFAULT_ENV | perl -lape s/_build$/_test/g) && assembly.py -h &> /dev/null && broad_utils.py -h &> /dev/null && illumina.py -h &> /dev/null && interhost.py -h &> /dev/null && intrahost.py -h &> /dev/null && metagenomics.py -h &> /dev/null && ncbi.py -h &> /dev/null && read_utils.py -h &> /dev/null && reports.py -h &> /dev/null && taxon_filter.py -h &> /dev/null && cd $PREFIX/opt/$PKG_NAME-$PKG_VERSION && py.test -v test/unit/test_tools_novoalign.py && cd $PREFIX/opt/$PKG_NAME-$PKG_VERSION && py.test -v test/unit/test_tools_picard.py && cd $PREFIX/opt/$PKG_NAME-$PKG_VERSION && py.test -v test/unit/test_tools_samtools.py && cd $PREFIX/opt/$PKG_NAME-$PKG_VERSION && py.test -v test/unit/test_util_misc.py && cd $PREFIX/opt/$PKG_NAME-$PKG_VERSION && py.test -v test/unit/test_util_vcf.py && cd $PREFIX/opt/$PKG_NAME-$PKG_VERSION && py.test -v test/unit/test_snake.py && export CONDA_DEFAULT_ENV=$CONDA_DEFAULT_ENV_BAK]]
```

Such things like PREFIX are not available when conda is not installed, so this test will fail because there is no /opt/...

Can you symlink `py.test` to bin to have it in your PATH? I think this will turn green as soon as you have tests that does not rely on the conda-build infrastructure.

Why this is running so long ... I have no clue.

johanneskoester:
@bgruening, I don't think py.test should be in bin. It is only for testing purposes, right? What about the following:
1. The problematic tests are commented out for now, and the PR is merged if that works out.
2. Mulled is extended, such that it looks for a selector `# [not mulled]`. If that is found, mulled ignores the particular test case.
3. Once that selector is implemented, the tests are activated again.

tomkinsc:
I agree with Johannes' plan. Given the choice of including conda-dependent tests and building this for quay.io, I would choose to avoid building this for quay.io. There should be an option to skip building for mulled. The expectation is that command written as part of the `meta.yaml` test block will be executed in the context of a `conda build` process, so it's reasonable to assume conda will be available for testing a conda package. Building by default for mulled is great, but a skip selector should be available; `# [not mulled]` sounds great. I've commented out the py.test-based test commands for now.

bgruening:
I don't see a point in skipping it for quay.io. We can skip tests, especially the complicated ones, but these tests guarantee actually that this package is running everywhere.

So I think running the tests by default and having at least one test that needs to run in a container should be the recommended way.

@tomkinsc what are your arguments against the quay.io builds? They are way more optimized than hand-crafted ones and as an additional benefit that do have the exact some binaries as the conda package, obviously :)

bgruening:
I guess the filtering of the tests can be done here: https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/pkg_test.py#L25

Skipping of tests that are labelled as `# [not mulled]`

johanneskoester:
Well, looks like it is a bit more complicated. It seems like conda-build checks for the validity of the selector. Unknown selectors are not supported...

tomkinsc:
Ugh. Rather than skipping individual tests, how about checking the yaml file for a key, `build_mulled:(true|false)`?

johanneskoester:
That could work, if we put the key under extra. However, then we would skip the entire mulled build, instead of just blacklisting certain tests...

bgruening:
@tomkinsc can we please leave these builds in. If you want to run these tests with $PREFIX you can also run them directly in the build.sh file, like `make check`.
One thing the mulled-builds are testing is actually that the build & test environment do not infer with each other. If for complicated tests this does not matter we can put them into build.sh?!

johanneskoester:
Right, thanks for the reminder @bgruening. This is indeed an important feature of mulled. It guarantees that nothing is leaked in from the build environment during the tests. This solves issues we had before, where certain files had been forgotten, and then were missing later when people tried to run the package.

tomkinsc:
I'd prefer to have the option to skip building for mulled entirely, since the resulting Docker container wouldn't be fully functional anyway (one of the things our custom Dockerfile does for this package is set internal mountpoints for licensed gatk and novoalign). 

Should I move the tests to build.sh as @bgruening suggests, @johanneskoester? I feel like tests should be in the test section, and not the build script, but whatever. The test-specific skip selector would be the best solution; annoying that conda checks for 'valid' selectors...

bgruening:
> I'd prefer to have the option to skip building for mulled entirely, since the resulting Docker container wouldn't be fully functional anyway (one of the things our custom Dockerfile does for this package is set internal mountpoints for licensed gatk and novoalign).

This is not true, you can mount in the licence stuff in the other containers as well. Just creating a extra path only because GATK can not fix their license is kind of a mood argument - I'm so tired of this GATK stuff :(

But you are missing the real argument, the builds are an additional benefit, what counts for bioconda is the extra testing layer. BioConda packages are ensured to work on busybox like operating systems and do not have any requirements on the OS (besides libc). For this and a few other ugly things in conda-test we have integrated the mulled tests.

Yes we are loosing some flexibility in writing tests (many of those can probably be handled by a smarter mulled version) but what we gain ensures high quality.

> I feel like tests should be in the test section, and not the build script, but whatever. 

What about `make check` and `make test`?

Lets see if these are now passing, I'm still not sure why this takes so long :(

johanneskoester:
Actually, I can understand both your positions. However, I think optionally deactivating at least the upload to quay.io would be desirable. We simply can't foresee if there will be any technical reasons why a mulled-busybox-container might not work as desired. If that is the case (like it seems to be here), I can completely understand that @tomkinsc does not want to have a non-working copy of viral-ngs floating around. That could be bad for the reputation. Of course, it would be desirable if those licensing issues would go away, but waiting or working on that should not prevent @tomkinsc from being able to properly publish his stuff.

Deactivating mulled could work like this:
```
extra:
  mulled: false
```
We can think about whether this should only deactivate the upload or also deactivate the tests.
```

bgruening:
@johanneskoester there is no technical reason why the conda package should work and the container not - besides mulled bugs. Its exactly the same package just delivered with an other technology.

Documentation can certainly be improved for these special cases, but this should also be part of the package in general. For example this package should have a big disclaimer in the extra section pointing to the licensing issue.

tomkinsc:
@bgruening: The GATK license certainly causes headaches within our group. I've added a disclaimer to the notes section of the newer version of this package. Workarounds are also in progress for passing in GATK and Novoalign as command line arguments. This package is a pipeline with numerous dependencies, and for various reasons there is an assumption that it will be run from an active conda environment, so our custom Docker image also provides a wrapper to activate the environment before running commands. Slightly unusual, but in support of various integrations. 

Mulled is a great resource, but it would be nice to have the option to opt-out for the odd edge cases like this. Compatibility of this package will hopefully improve in the future.

Issue #3130, opened 2016-12-07
No travis log output when build times out


marcelm:
I am not sure whether this is correct, but I believe that the Travis log output is incomplete when a build times out.

I am trying to find out why #3085 times out all the time. [The log output](https://travis-ci.org/bioconda/bioconda-recipes/jobs/181182806) of one of these attempts is not helpful because it stops right before the recipe build starts. It seems that the error log is only printed when the build script recognizes that something has failed. When the script itself gets stopped because of a timeout, it does not print anything.

If my assumption is correct, I suggest that the behavior be changed. The log should really be updated live if possible.

gregvonkuster:
+1 I've been hitting this issue repeatedly with various recipes lately.

bgruening:
My best bet is that you are missing some runtime dependency in your recipe. Can you try to add gcc as runtime dep?

marcelm:
I’ve fixed the runtime deps already. This report is about the missing debug output, which made it very hard to find the problem.

daler:
@marcelm @bgruening it would be great to get logs to show up in real time but I think the way travis is set up might require some large changes in our infrastructure.

marcelm:
Hm, ok, wasn’t aware that this is a Travis thing. Too bad!

marcelm:
The actual problem is in `travis_wait`, which does the stdout buffering: travis-ci/travis-ci#5716

marcelm:
How much time do the jobs on Travis actually get? Is it two hours? Then giving `travis_wait` a timeout of 120 minutes is too much because the job has already been running for at least one or two minutes when it starts. So the entire job is killed "from the outside" and `travis_wait` never has a chance of reaching the timeout. As far as I can tell from its source code, it *would* print the output when a timeout is reached.

The obvious solution is to reduce the timeout by a few minutes. This won’t give us a 'live' update, but we would at least see the log output when a job that ran too long was killed.

marcelm:
Note also [the last lines of the log output](https://travis-ci.org/bioconda/bioconda-recipes/jobs/181182806#L477). It prints "Still running (119 of 120): scripts/travis-run.sh". It never reaches 120 minutes. Also, the very last line (`The job exceeded the maximum time limit` ...) is not the message that `travis_wait` would print.

daler:
[...]

PR #3170, opened 2016-12-11
Reduce travis_wait timeout to 110 minutes

marcelm:

Using a timeout of 120 minutes does not work since the total time the job is
allowed to run is already 120 minutes. At the point when travis_wait starts,
the job has already been running for one or two minutes.

See issue #3130


daler:
Thanks @marcelm!

Issue #3167, opened 2016-12-11
perl-threaded can be deprecated


bgruening:
@tomkinsc was able to get our custom perl modifications into conda-forge with https://github.com/conda-forge/perl-feedstock/pull/6

This means we can just use `perl` instead of `perl-threaded`. I see two possibilities from here on:

* bulk rebuild all perl-packages against `perl`
* change our `perl-threaded` package to depend on `perl`, some kind of meta-package and change our recommendation to use `perl` from now on.

The first one will take a while and a lot of effort, like the R migrations. The second is not tested, just a guess that this might work.

Any other ideas?

Thanks a lot to @tomkinsc who spend a lot of time in getting this package into conda-forge!

ping @bioconda/core

johanneskoester:
Thanks @tomkinsc! I'd prefer the bulk rebuild. We can do this gradually, one connected component of our internal perl dep-graph at a time. Maybe we can even combine this with moving perl packages over to conda-forge if they are not bioinformatics specific.

rvalieris:
there is also a perl packge, in addition to perl-threaded, on the bioconda channel, wouldn't that mean the bioconda version will have the highest priority and be installed instead of the conda-forge version ?

bgruening:
@rvalieris yes this is correct.
@johanneskoester do you see any disadvantages in the second option? I spend so many hours already in perl land and this is a mess and rebuilding takes probably ages. Please note that we haven't even finished the R rebuild by now :(

rvalieris:
so as I understand, for this to work without the need of rebuilding packages:
* conda-forge::perl needs to match the current bioconda::perl-threaded version (because perl modules are installed in versioned paths)
* prevent users from getting bioconda::perl by accident after the transition ( because bioconda::perl is not compiled with threads (and "thread-multi" is also included on the module path))

but wouldn't all modules need to be rebuild anyway when perl itself is updated ?

bgruening:
Sounds correct to me. A rebuild is imho only necessary if you increase the perl version and make it ABI/API incompatible.

johanneskoester:
I don't have a strong opinion on this. Feel free to choose option 2 if that is easier.

jerowe:
So just to be clear, what perl from what channel should we be building with? I know the channels get pulled in automatically, but I still want to know. It looks like the perl in conda-forge is now the functional equivalent of the perl-threaded. Is this correct?

bgruening:
[...]

Issue #3131, opened 2016-12-07
perl-bioperl 1.6.924-6 is broken


dakl:
Hi, 

I'm getting problems when installing perl-bioperl-1.6.924 (build number 6) using `conda install perl-bioperl`. Certain subpackages such as `Bio::PrimarySeqI` are missing. A quick look at anaconda could reveals that build 4 is 2.2 Mb but build 6 is only 5.7 kb. Manually downloading build 4 and installing is works fine. 

Is therefore seems to me that something is broken in the latest build of this version of bioperl. 

cheers

bgruening:
Hi @dakl,

I guess this was due to some work regarding a new bioperl release: https://github.com/bioconda/bioconda-recipes/pull/2909 and many more PR.

Any help here is appreciated!
Thanks,
Bjoern

matthdsm:
seconded, I'm having the same issue. Any news on a fix? Is it possible to revert (remove?) a build, so that version installed by default is at least functional?

Thanks
M

epruesse:
outdated

Issue #2917, opened 2016-11-11
Perl-bioperl-core-1.7.1 package broken.. please remove!!


martin-raden:
@bgruening I am crying... the bioperl 1.7 package (which kept me busy for more than a week) is not packed correctly.. even with successful travis build... (test was disabled due to cyclic deps...)

See file size: [perl-bioperl-core-1.7.1](https://anaconda.org/bioconda/perl-bioperl-core/files) = 14kb ...

Please remove the 1.7 package from the channel.

This causes no problems, since the `perl-bioperl` package is hard tagged with the old version (no 1.7 perl-bioperl package so far)

martin-mann:
Grmpf.. guess its this [local::lib](https://github.com/bioperl/bioperl-live/blob/master/INSTALL.md) quirk that is needed to get it installed in the target directory... so another dependency..

martin-mann:
Is there a way to incorporate this perl-local-lib package from a private channel?

[https://anaconda.org/dan_blanchard/perl-local-lib](https://anaconda.org/dan_blanchard/perl-local-lib)

Ahhh.. will pull conda `perl` as well... 😣

epruesse:
assuming this was fixed somehow

PR #3177, opened 2016-12-12
bumped vep version

matthdsm:

### DIFF
--- a/recipes/variant-effect-predictor/build.sh
+++ b/recipes/variant-effect-predictor/build.sh
@@ -31,10 +31,10 @@ perl vep_install.pl -a a --NO_HTSLIB
 # Remove test data
 rm -rf t/
 # Install plugins
-wget --no-check-certificate https://raw.githubusercontent.com/Ensembl/VEP_plugins/release/86/CADD.pm
-wget --no-check-certificate https://raw.githubusercontent.com/Ensembl/VEP_plugins/release/86/dbNSFP.pm
-wget --no-check-certificate https://raw.githubusercontent.com/Ensembl/VEP_plugins/release/86/MaxEntScan.pm
-wget --no-check-certificate https://raw.githubusercontent.com/Ensembl/VEP_plugins/release/86/GeneSplicer.pm
-wget --no-check-certificate https://raw.githubusercontent.com/Ensembl/VEP_plugins/release/86/dbscSNV.pm
+wget --no-check-certificate https://raw.githubusercontent.com/Ensembl/VEP_plugins/release/87/CADD.pm
+wget --no-check-certificate https://raw.githubusercontent.com/Ensembl/VEP_plugins/release/87/dbNSFP.pm
+wget --no-check-certificate https://raw.githubusercontent.com/Ensembl/VEP_plugins/release/87/MaxEntScan.pm
+wget --no-check-certificate https://raw.githubusercontent.com/Ensembl/VEP_plugins/release/87/GeneSplicer.pm
+wget --no-check-certificate https://raw.githubusercontent.com/Ensembl/VEP_plugins/release/87/dbscSNV.pm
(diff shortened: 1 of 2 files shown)


matthdsm:
Hi @bioconda/core 

Sorry for bothering you guys again, but I've no idea what goes wrong here. During the install , the vep installer tries to download and install the ensembl API. It seems there's an issue with curl during that step. I don't have a clue what could be the issue, since all that's changed in the installer script is the version number.. Any ideas?

johanneskoester:
Here is the error: https://travis-ci.org/bioconda/bioconda-recipes/jobs/183197781#L596
unzip is missing in the conda-forge container that we use now. However, it is available via conda, so you can add it as a build dependency.

matthdsm:
Well I'll be damned, I thought there was an issue with the curl request or something. adding unzip fixed it in a local build. I'll be pushing the changes shortly.

Thanks fot the help @johanneskoester!!

M

matthdsm:
I seems the build still fails, due to the broken bioperl package. What is the best way to force use of build version 4? When I try `conda install -c bioconda perl-bioperl=1.6.924-4`, I get a `PackageNotFoundError`.

Thanks
M

johanneskoester:
I think you need to use an underscore for the build version. So, build version 6 is broken? In what sense?

matthdsm:
There are a bunch of modules missing. There's been already several bug reports about this:
#3131 and #2917

johanneskoester:
@bgruening can you comment on the status of bioperl? How far are we away from fixing this?

bgruening:
bioperl-1.7 has still some wired bugs and dependency issues. Maybe I find time over christmas it seems to be problematic and I'm not a perl expert.

matthdsm:
Then what's the best way to pin the bioperl dependency on a previous build? This is something that needs to be done, since the current vep recipe is also broken because of the bioperl issue.

bgruening:
@johanneskoester do we want to remove the latest 2 bioperl 1.6 builds?

matthdsm:
👍 please do!

matthdsm:
I think everything should work now, but if I understand the travis logs correctly, there's something going wrong during the download from Ensembl. Can the travis build be retriggered? Or do you guys think there's another issue?

Thanks
M

rvalieris:
you can re-trigger the build my logging into the travis page(with github acc), I just triggered it now.

matthdsm:
didn't know that, thanks!

matthdsm:
ok, the local build works fine, and travis seems to be OK with the changes.
There's still one last issue left. During testing I noticed that the bio-db-hts library can't be found by vep (previous versions had this issue too)
Any perl masters around that have experience with this?
@bioconda/perl ?

rvalieris:
it seems bio-db-hts is installed in the ```x86_64-linux-thread-multi``` directory, but perl doesn't have that directory on ```@INC```, perl-threaded does.

I think one of the dependencies(or the dependencies of the dependencies) is pulling perl instead of perl-threaded.

bgruening:
To add more complexity it would be nice to deprecate perl-threaded and just stick to `perl`.
https://github.com/bioconda/bioconda-recipes/issues/3167#issuecomment-266816990

johanneskoester:
@bgruening if the bioperl builds are broken, I agree with removing them.

matthdsm:
Hi @bgruening ,

How do you propose I proceed from here? Should I stick with perl-threaded for the sake of continuity until you guys begin with the migration over to perl? Or do I pick the regular Perl? I don't mind changing it, as long as it doesn't screw up all the libraries and stuff. Is there already a mechanism in place to use the 'forged' version instead of the bioconda one?

M

matthdsm:
@bioconda/core 
If there aren't any objections, I'm going to merge this with perl-threaded intact. We can always update the vep version in a second build if we're sure it won't screw up library discovery.
Awaiting confirmation!
M

PR #3179, opened 2016-12-12
Update umis recipe to 0.4.0a.

roryk:


recipes/umis/meta.yaml
```diff
@@ -1,14 +1,14 @@
 package:
   name: umis
-  version: '0.3.1a0'
+  version: '0.4.0a'
 
 build:
-  number: 2
+  number: 1
```
bgruening: This should be 0 for a new version.

PR #3180, opened 2016-12-12
Check that conda-bld is not empty in simulate_travis

tetianakh:

Make sure that conda-bld directory is a valid conda channel
(i.e. it contains a package index) when running `simulate_travis.py`. 
This is done by adding the directory as a channel, updating the index, and installing a dummy package
if the update failed.

### DIFF
--- a/recipes/_dummy/meta.yaml
+++ b/recipes/_dummy/meta.yaml
@@ -1,0 +1,2 @@
+package:
+  name: _dummy
--- a/simulate-travis.py
+++ b/simulate-travis.py
@@ -59,6 +59,42 @@ args, extra = ap.parse_known_args()
 HERE = os.path.abspath(os.path.dirname(__file__))
 
 
+def _ensure_build_dir_not_empty():
+    """Make sure the build directory can serve as a valid channel.
+    If the directory is empty or contains no builds, the update
+    command will fail. In this case, a dummy package is installed to
+    to create conda index in the build directory.
+    """
+    import bioconda_utils.docker_utils
+    build_dir = bioconda_utils.docker_utils.get_host_conda_bld()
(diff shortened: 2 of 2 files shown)


recipes/_dummy/meta.yaml
```diff
@@ -0,0 +1,2 @@
+package:
+  name: _dummy
```
daler: It would be better to have this dummy recipe somewhere else (tmp dir?) -- otherwise the build system may see it as a package that needs to be in the bioconda channel.

simulate-travis.py
```diff
@@ -59,6 +59,42 @@
...
+    to create conda index in the build directory.
+    """
+    import bioconda_utils.docker_utils
+    build_dir = bioconda_utils.docker_utils.get_host_conda_bld()
+
+    ADD_CHANNEL_CMD = [
+        'conda', 'config', '--add', 'channels', 'file://%s' % build_dir]
+    REMOVE_CHANNEL_CMD = [
+        'conda', 'config', '--remove', 'channels', 'file://%s' % build_dir]
+    CONDA_UPDATE_DRY_RUN_CMD = ['conda', 'update', '--dry-run', '--all']
```
daler: I like this idea for checking all the moving parts. Should prob be using it more.

simulate-travis.py
```diff
@@ -59,6 +59,42 @@
...
+
+    with open(os.devnull, 'w') as DNULL:
```
daler: It would be better to report stderr/stdout for easier debugging. Maybe preface it with a log statement saying what's being done.

daler: OK, sounds good. Thanks.

PR #3193, opened 2016-12-13
reago

HassanAmr:

modified setup.py and added MANIFEST.in to include all needed files


recipes/reago/meta.yaml
```diff
@@ -27,5 +27,5 @@ test:
 
 about:
   home: https://github.com/chengyuan/reago-1.1
```
bgruening: You should add more tests to be sure it is working.
The CM files are not copied over isn't it?

HassanAmr: I'm still not sure about that part. Conda build counts 4 files, but I'm not sure if it is excluding the setup.py and the MANIFEST.in or not. And I can't tell for sure because it deletes the files locally since the test/build succeeds.

HassanAmr: Should I add more test cases in this case as well (when copying the files)?

HassanAmr:
Hi @bgruening ,
Is the genometools in bioconda the same as in the following link:

http://genometools.org/pub/nightly_builds/

Because the tools needs Readjoiner 1.2

bgruening:
No it is https://github.com/bioconda/bioconda-recipes/blob/master/recipes/genometools-genometools/meta.yaml so genometools-genometools

bgruening:
Mh, actually the test is failing.

HassanAmr:
Yes, there is a problem when splitting a string, I think. And a problem when calling cmsearch. I'll create a new patch for it.

HassanAmr:
Hi @bgruening,

Sorry, for adding $SRC_DIR again, I just want to test it out.
Any suggestions how I might perform the test without having to copy sample data files?
Or should I copy them, then delete them?

bgruening:
@HassanAmr can you look at http://conda.pydata.org/docs/building/meta-yaml.html#source-files

bgruening:
:)

PR #3199, opened 2016-12-14
RNA-STAR v2.4.0d (version used in Galaxy)

yhoogstrate:

### DIFF
--- a/recipes/star/2.4.0d/meta.yaml
+++ b/recipes/star/2.4.0d/meta.yaml
@@ -1,0 +1,23 @@
+{% set version = "2.4.0d" %}
+{% set sha256 = "41734876a92cb37085e35065aeca924b37fd0e9e5f65b39ac02c9d6461ad0788" %}
+
+package:
+  name: star
+  version: {{ version }}
+
+source:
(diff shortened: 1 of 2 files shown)


croth1:
Is this the same recipe as this? https://github.com/bioconda/bioconda-recipes/blob/f5eb63e30a76fd13c28663786d219c9f7750267c/recipes/star/meta.yaml

yhoogstrate:
More or less...  this one is specific for 2.4.0d and the template you show does not work for this version for at least two reasons:
 - The archive/package does not provide a packaged build for OS-X/Mac.
 - Several issues with linking `std::` and openMP break compilation using the provided Makefile

recipes/star/2.4.0d/meta.yaml
```diff
@@ -0,0 +1,23 @@
...
+  name: star
+  version: {{ version }}
+
+source:
+  fn: STAR_{{ version }}.zip
+  url: https://github.com/alexdobin/STAR/archive/STAR_{{ version }}.zip
+  sha256: {{ sha256 }}
+
+build:
+  number: 1
```
bgruening: This should be 0.

PR #3200, opened 2016-12-14
Entrez direct

jfear:

Updates to the latest version of Entrez direct to work with new https requirements.


daler:
@bgruening do we have a date set for when all new perl-dependent recipes should use conda-forge's perl instead of bioconda's perl-threaded?

daler:
So should this one have `perl-threaded` or `perl`?

bgruening:
It's find as it is :)

bgruening:
Thanks @jfear!

PR #3201, opened 2016-12-14
Metametamerge new recipe

pirovc:


pirovc:
Sorry, forgot to @bioconda/core

More info: this package is going to be a module of another pipeline and I need it as a dependency. It is a stable version. I'm uploading it separately because it can be used as standalone tool and it also makes it easier to maintain and update.

bgruening:
Thanks @pirovc!

PR #3206, opened 2016-12-14
Graphclust individual wrappers

mmiladi:


mmiladi:
@bgruening Could you please have a look at this? Tests for all perl wrapper except those which didn't use argparser are added and passing.

bgruening:
Already did, looks good!

mmiladi:
Good thanks! Though that was a PR on my repository and this one is PR on upstream. Then please you do the merge later. 

Osx Travis queue seems to be busy, it would be cool if we have a metaflag to not run osx test if it is not supported --> Reduces CO2 emission :-P

PR #3210, opened 2016-12-15
sparse assember new recipe

mictadlo:

### DIFF
--- a/recipes/sparseassembler/meta.yaml
+++ b/recipes/sparseassembler/meta.yaml
@@ -15,12 +15,12 @@ source:
 
 requirements:
   build:
-  - gcc [not osx]
+  - gcc # [not osx]
   - llvm  # [osx]
   - zlib
 
   run:
-  - libgcc [not osx]
+  - libgcc # [not osx]
   - zlib
 
 test:
--- a/recipes/sparseassembler/meta.yaml
+++ b/recipes/sparseassembler/meta.yaml
@@ -7,7 +7,7 @@ package:
 
 build:
   number: 0
-  script: "mkdir -p $PREFIX/bin; g++ -O3 -o SparseAssebmler *.cpp; mv * $PREFIX/bin"
+  script: "mkdir -p $PREFIX/bin; g++ -O3 -o SparseAssebmler *.cpp; mv SparseAssebmler $PREFIX/bin"
 
 source:
   git_url: https://github.com/yechengxi/SparseAssembler.git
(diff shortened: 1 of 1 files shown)


recipes/sparseassembler/meta.yaml
```diff
@@ -0,0 +1,34 @@
+{% set name = "SparseAssembler" %}
+{% set version = "20160205" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+build:
+  skip: True  # [osx]
```
bgruening: I'm not sure why you skip OSX if it seems to work.

recipes/sparseassembler/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+  number: 0
+  script: "mkdir -p $PREFIX/bin; chmod a+x *; mv * $PREFIX/bin"
+
+source:
+  fn: SparseAssembler
+  url: https://raw.githubusercontent.com/yechengxi/SparseAssembler/master/compiled/SparseAssembler
```
bgruening: Can you compile this tool. precomiled things are not recommended.

recipes/sparseassembler/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+  md5: c9ba3f2dbbd9146767c3bdad02118a85
+
+requirements:
+  build:
+  - gcc [not osx]
```
bgruening: If you use precompiled binaries ,why are you using gcc?

daler:
Looks like you're using a git repo, in which case you should be using `git_url` instead of `url` (http://conda.pydata.org/docs/building/meta-yaml.html#source-from-git).

Much better to use a release, but I don't see any source release in the repo. You might ask the author for a release tag. But please at least use a commit hash rather than master, since master branch can change at any time making the recipe non-reproducible.

mictadlo:
@daler @bgruening I changed it, but now I get `SparseAssembler: command not found`. What could I miss?

recipes/sparseassembler/meta.yaml
```diff
@@ -0,0 +1,33 @@
+{% set name = "SparseAssembler" %}
+{% set version = "20160205" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+build:
+  number: 0
+  script: "mkdir -p $PREFIX/bin; g++ -O3 -o SparseAssebmler *.cpp; mv SparseAssebmler $PREFIX/bin"
```
daler: Looks like typos here, s/ebm/emb/g. That will cause the error you're seeing on travis.

daler:
Getting there! Hopefully those typo fixes will do the trick.

mictadlo:
Thank you @bgruening  and @daler for your help

PR #3223, opened 2016-12-15
Pathoscope

mlbendall:

Requesting review from @bioconda/core since this is my first recipe. I think this recipe is pretty standard. It is a python package that only runs in 2.7 and depends on bowtie2 and samtools.

I did have some trouble running `simulate_travis.py --docker` (on OSX), was getting a "mounts denied" error. No problems without `--docker` and passed on travis-ci.org. I can create an issue for this problem if it persists.


johanneskoester:
Looks good!

PR #3231, opened 2016-12-16
pin hdf5 for kallisto plus tests

daler:

I think we need to have HDF5 pinned, but wanted to get feedback from @bioconda/core before doing this.

Currently if you install `kallisto` from bioconda, it'll pull in these packages:

```bash
conda create -n test-kallisto kallisto
# The following NEW packages will be INSTALLED:
# 
#    hdf5:     1.8.17-1
#    kallisto: 0.43.0-1 bioconda
#    libgcc:   5.2.0-0
#    zlib:     1.2.8-3
```

The following error happens only when actually using kallisto with example data, not when doing any minimal tests like checking help:

```
...
[   em] the Expectation-Maximization algorithm ran for 225 rounds
Warning! ***HDF5 library version mismatched error***
The HDF5 header files used to compile this application do not match
the version used by the HDF5 library to which this application is linked.
Data corruption or segmentation faults may occur if the application continues.
This can happen when an application was compiled by one version of HDF5 but
linked with a different version of static or shared HDF5 library.
You should recompile the application or check your shared library related
settings such as 'LD_LIBRARY_PATH'.
You can, at your own risk, disable this warning by setting the environment
variable 'HDF5_DISABLE_VERSION_CHECK' to a value of '1'.
Setting it to 2 or higher will suppress the warning messages totally.
Headers are 1.8.16, library is 1.8.17
```

Manually specifying the HDF5 version solves the problem;

```bash
conda create -n test-kallisto kallisto "hdf5==1.8.16"
```

So I'm pretty sure this means we should pin HDF5 in `scripts/env_matrix.yaml`.

I've done that in this PR. I've also added some very small test data; the test fails before pinning and passes after pinning so it should hopefully catch any regressions.


kyleabeauchamp:
I know that HDF5 versioning is a source of issues in a lot of packages in the defaults channel as well, so it would be good to make sure that our choice is consistent with the most commonly used version in defaults and conda-forge.

daler:
@kyleabeauchamp do you mean "most commonly downloaded" or "most commonly pinned in recipes"?

kyleabeauchamp:
Not sure, I just want to make sure we don't cause unnecessary clashes in the dependency DAG by pinning to an older version.

daler:
Ah. I'll have a grep around their repos to see what they're using.

kyleabeauchamp:
According to this search (https://github.com/conda/conda-recipes/search?utf8=%E2%9C%93&q=hdf5), it seems like there's not actually that much pinning elsewhere.  Might not be as bad as I imagined.

bgruening:
@daler yes we need to pin way more.
Please see this work if have started to pin the same/similar things than conda-forge: https://github.com/bioconda/bioconda-recipes/compare/env_matrix_ext?expand=1

daler:
@bgruening nice, I had missed that. What's the best way forward? 

- Should I merge this PR for now, or pull your changes, or change the `CONDA_HDF5` to what you have and then merge?
- Looks like in your branch you're not adding pinned packages to the build string likely because of the `*` and `|` in the pinned version specs. So should I remove the HDF5 version from the package name here?

johanneskoester:
Adding the pinned versions to the build string is necessary. Otherwise, updating the versions will not result in rebuilding the packages.

daler:
@johanneskoester in theory the env_matrix supports version specs like `CONDA_HDF5: "1.8.17|1.8.17.*"` or `CONDA_SQLITE: "3.13.*"` (to take a couple examples from Bjoern's branch). In those cases, it's not clear what to add to the build string. Maybe some combination of Jinja2 filters can be used to convert from env_matrix vars to valid build strings.

Anyway, it seems my initial reluctance to pin is unwarranted given that we may be moving towards much more pinning. I'll merge this for now so we at least have a working version.

johanneskoester:
@daler @bgruening that's true, but then we should probably disallow those. As far as I know they are equivalent to just omit .*. Or am I missing something?

PR #3236, opened 2016-12-17
Assemblyutility

mictadlo:

### DIFF
--- a/recipes/assemblyutility/meta.yaml
+++ b/recipes/assemblyutility/meta.yaml
@@ -29,6 +29,6 @@ test:
 
 about:
   home: https://github.com/yechengxi/AssemblyUtility
-  summary: Tools for DBG2OLC genoome assembler.
+  summary: Tools for DBG2OLC genoome assembler
 
 
--- a/recipes/assemblyutility/meta.yaml
+++ b/recipes/assemblyutility/meta.yaml
@@ -9,8 +9,9 @@ build:
   number: 0
 
 source:
-  git_url: https://github.com/yechengxi/AssemblyUtility.git
-  git_rev: 8e476764eb67ebb486e673ecbd8a0fa5d8eae22c
+  fn: {{ name|lower }}_{{ version }}.tar.gz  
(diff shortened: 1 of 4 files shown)


bgruening:
This one contains the other branch as well.
This is not so bad, but can you please remove the one commit to SparseAssembler. Mergings this PR will also merge the other PR of yours.

recipes/assemblyutility/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+build:
+  number: 0
+
+source:
+  git_url: https://github.com/yechengxi/AssemblyUtility.git
```
bgruening: You could use this tarball plus a checksum instead of git: https://github.com/yechengxi/AssemblyUtility/archive/8e476764eb67ebb486e673ecbd8a0fa5d8eae22c.tar.gz

mictadlo: Should I change, Sparc, SparseAssembler and DBG2OLC, too?

bgruening: Yes please, we recommend to use tarball and not git checkouts.
For your other error wich you see, please always create new branches from latest master, or rebase against latest master.

bgruening: After you have pushed something, the most save thing you can do is `git checkout master`, `git pull upstream master` and `git checkout -b new_branch_name` to create and change to a new branch.

recipes/sparseassembler/meta.yaml
```diff
@@ -25,7 +25,7 @@ requirements:
 
 test:
   commands:
-    - SparseAssembler | grep "Error! Genome size not given." > /dev/null
+    - SparseAssembler #| grep "Error! Genome size not given." > /dev/null
```
bgruening: This needs to be removed from this PR

recipes/assemblyutility/meta.yaml
mictadlo: Thank you.

PR #3246, opened 2016-12-19
pysam: avoid pinning to specific htslib versions

chapmanb:
The specific pins for htslib (and samtools, bcftools) cause resolution
errors now that there is a newer htslib (1.3.2, instead of 1.3.1). When
you install:

conda install htslib cnvkit

where the latest cnvkit relies on recent pysam (0.9.1.4) you'll get
htslib 1.3.2 and an old version of pysam and cnvkit prior to the version
pins. We relax these to these avoid confusing resolution issues and enable
both recent htslib and pysam.

### DIFF
--- a/recipes/pysam/meta.yaml
+++ b/recipes/pysam/meta.yaml
@@ -8,16 +8,16 @@ source:
     md5: a7e0e9cbc972618cde7aea54894067d6
 
 build:
-    number: 0
+    number: 1
     skip: False
 
 requirements:
     build:
         - gcc  # [linux]
         - llvm # [osx]
-        - htslib ==1.3.1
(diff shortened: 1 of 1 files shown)


bgruening:
I guess we need to pin htslib globally as we do with boost et all.
I started to work on more pinning here, just dropped the ball for the moment: https://github.com/bioconda/bioconda-recipes/tree/env_matrix_ext

chapmanb:
Björn -- thanks. Pinning to ranges instead of a specific version works great to me now. I know specific version pinning ends up in dependency nightmares so happy to try and avoid that whenever we can unless we have special infrastructure for it.

PR #3249, opened 2016-12-19
Metameta new recipe

pirovc:

@bioconda/core

### DIFF
--- a/recipes/metameta/meta.yaml
+++ b/recipes/metameta/meta.yaml
@@ -5,7 +5,7 @@ package:
 source:
   fn: v1.0.tar.gz
   url: https://github.com/pirovc/metameta/archive/v1.0.tar.gz
-  md5: 91786eda950d7440d26cbd6e4577ab99
+  md5: 0214502abbb8b6e9c4d49b74f2e1d58f
 
 build:
   number: 0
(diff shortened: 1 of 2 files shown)


daler:
@pirovc do you think it's worth adding a test to make sure snakemake can find all the installed configs and scripts?

pirovc:
Some includes are dynamic and depend on user choice. Still I could add a dummy configuration file and trigger a dry-run with the default options. Would that suffice?

daler:
I think that would improve the test; @johanneskoester any thoughts?

johanneskoester:
Sounds good!

pirovc:
@daler test added. If any file/include is missing or if snakemake is not running the test command is going to return an error.

recipes/metameta/meta.yaml
```diff
@@ -0,0 +1,26 @@
...
+  skip: True # [osx]  
+  
+requirements:
+  run:
+    - snakemake >=3.9.0
+
+test:
+  commands:
+    - touch "metameta_dummy_file_test"
+    - metameta --configfile $PREFIX/opt/metameta/scripts/metameta_dummy_file_test.yaml -np > /dev/null
```
johanneskoester: Please remove the piping into /dev/null (harder to debug if something goes wrong).

daler:
Thanks @pirovc!

PR #3256, opened 2016-12-19
Increase build number for perl-bioperl

nsoranzo:
This was missed in #2876.
Also:
- change `perl-threaded` requiments to `perl`.
- remove requirements provided by `perl-bioperl-core` >= 1.6.924-pl5.22.0_2

### DIFF
--- a/recipes/perl-bioperl/meta.yaml
+++ b/recipes/perl-bioperl/meta.yaml
@@ -3,24 +3,18 @@ package:
   version: "1.6.924"
 
 build:
-  number: 6
+  number: 7
 
 requirements:
   build:
-    - perl-threaded
+    - perl
     - perl-bioperl-core ==1.6.924
     - perl-bioperl-run
-    - perl-bio-asn1-entrezgene
-    - perl-bio-featureio
-    - perl-bio-samtools
 
   run:
-    - perl-threaded
+    - perl
     - perl-bioperl-core ==1.6.924
     - perl-bioperl-run
-    - perl-bio-asn1-entrezgene
-    - perl-bio-featureio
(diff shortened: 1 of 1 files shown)


nsoranzo:
I could use some help here. The tests fail because of this dependency chain:

`perl-bioperl -> perl-bioperl-core ==1.6.924 -> perl-gd -> libgd -> jpeg`

which is resolved by Conda in `perl-gd 2.56-4`, `libgd 2.2.3-1` and `jpeg 8d-2`. Unfortunately, `perl-gd 2.56-4` was built on 2016/11/19 using a more recent libgd, which requires `jpeg >=9.*` since `2.2.3-2` (built on 2016/11/14).

I can change the libgd requirement in perl-gd like this:
```
-    - libgd >=2.2.3
+    - libgd 2.2.3 4
```
which I think it is the only way to specify a minimum build number, but it's a bit restrictive since it is no more a minimum version, but an exact version-build requirement. Suggestions?

nsoranzo:
Simple way to reproduce the `perl-gd` issue: https://gist.github.com/nsoranzo/166c144b0e8dff845e6351290f7eebd5

nsoranzo:
I've rebased the PR, now the problem is very similar but with `libicui18n.so.56` instead of `jpeg`.

nsoranzo:
The build is now failing due to a `conda-build` bug fixed by https://github.com/conda/conda-build/pull/2089, need to wait for 2.1.16 release.

nsoranzo:
Took me more than 6 months, but it's finally green! Merging.

bgruening:
Wow, thanks a bunch @nsoranzo!

PR #3261, opened 2016-12-20
Add recipe for pasta version 0.2

gregvonkuster:

### DIFF
--- a/recipes/pasta/0.2/build.sh
+++ b/recipes/pasta/0.2/build.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+out_dir=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+pasta_tools_dev_dir=$PREFIX/share
+share_dir=$PREFIX/share
+
+mkdir -p $out_dir
+mkdir -p $PREFIX/bin
+mkdir -p $PREFIX/pasta
+
+cp -R ./* $out_dir
+
+unamestr=`uname`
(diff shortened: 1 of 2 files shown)


recipes/pasta/0.2/build.sh
```diff
@@ -0,0 +1,52 @@
...
+    tar -xzf 2924290.tar.gz
+    mv sate-tools-linux-292429063ee62055b610b592356ff398f68c0a4a sate-tools-linux
+    mv sate-tools-linux $pasta_tools_dev_dir
+    cd $out_dir
+    $PYTHON setup.py install
+    cp -R $share_dir/sate-tools-linux/* $PREFIX/bin
+elif [ $unamestr == 'Darwin' ];
+then
+    export DYLD_LIBRARY_PATH=$PREFIX/lib
+    wget --no-check-certificate https://github.com/smirarab/sate-tools-mac/archive/0712214.tar.gz
```
bgruening: This should ideally be an other package and a dependency on this one.

gregvonkuster:
I'm closing this for now, and will try to come up with a way to use Pasta that eliminates the requirement for these tools https://github.com/smirarab/sate-tools-mac/archive/0712214.tar.gz since they are pre-compiled with gcc 2.14, resulting in the following error on systems that do not have that version installed.  It is so unfortunate that Pasta's install process defined here https://github.com/smirarab/pasta/blob/master/README.md uses this approach.
```
/lib64/libc.so.6: version `GLIBC_2.14' not found (required by ./share/sate-tools-linux/clustalw2)
```

Issue #3271, opened 2016-12-21
Travis CI environment variables


sebastian-luna-valero:
Hi All,

We use Travis CI environment variables to exclude some tests. Specifically, we use `TRAVIS`, as listed here:
https://docs.travis-ci.com/user/environment-variables/

This variable should also be available when running Travis CI for `bioconda-recipes`, am I right?

For some reason, it does not seem to be the case and it causing our error here:
https://travis-ci.org/bioconda/bioconda-recipes/jobs/185740645#L3762

Best regards,
Sebastian

daler:
Thanks @sebastian-luna-valero, didn't realize others would be depending on those env vars but it makes sense. I'll add an issue to bioconda-utils.

That said, we generally don't run unit tests in recipes (see https://bioconda.github.io/guidelines.html#tests for more info) so in this particular case you might just keep the `cgat --help`.

sebastian-luna-valero:
You're right @daler , many thanks!

I will update the recipe accordingly.

bgruening:
Please feel free to reopen if this is still a problem after @daler has merged #45

daler:
@sebastian-luna-valero can you please try out https://github.com/bioconda/bioconda-utils/pull/49 to see if this fixes the issue?

PR #3278, opened 2016-12-21
Cgat scripts nosetests

sebastian-luna-valero:

After discussion here:
https://github.com/bioconda/bioconda-recipes/issues/3271

I have created a metapackage as explained here:
https://bioconda.github.io/guidelines.html#metapackages

I plan to provide a full package (i.e. instead of metapackage) after further testing.


bgruening:
Tests are missing and there is no build-script or a script specified, correct?

sebastian-luna-valero:
Yes @bgruening, this is a conda metapackage as explained here:
https://bioconda.github.io/guidelines.html#metapackages

A previous example of a conda metapackage in bioconda can be found here:
https://github.com/bioconda/bioconda-recipes/tree/master/recipes/hubward-all

There is a different way of creating conda metapackages:
http://conda.pydata.org/docs/commands/build/conda-metapackage.html

but I have been using that option before and actually found it worse to maintain.

bgruening:
@sebastian-luna-valero I know, but I think its worth to include one or two tests isn't it?

sebastian-luna-valero:
@bgruening here are the bioconda guidelines for meta packages:
https://bioconda.github.io/guidelines.html#metapackages

There is no source code to test in a meta package, am I wrong? What kind of tests do you propose in this case?

daler:
@sebastian-luna-valero @bgruening our tests are to make sure paths are correct, runtime deps are installed, etc. With a metapackage, no tests means we run the risk of having incompatible packages bundled together. If all runtime deps are properly specified in the recipes of the respective bundled packages, the dependency solver should identify any problems.

Honestly I don't know what the right way to test this would be other than writing a long `run_test.sh` that repeats every component command-line packages' test, and a `run_test.py` that imports every python dep. This seems like it would be a pain to maintain, which is why I omitted it in the original hubward-all metapackage.

So I guess we need to decide whether catching package incompatibilities is worth the maintenance cost of large tests. I think I'm OK with no tests for metapackages.

sebastian-luna-valero:
Thanks @daler !

I was naively assuming that if we are using only packages belonging to the 4 "official" channels (defaults, conda-forge, r and bioconda) each package would have been properly tested individually and therefore a collection of them into a single meta package should not be tested.

Please @bgruening let us know your thoughts.

bgruening:
I'm not sure, I would like to see some tests, just a few random imports, maybe one call. A recipe without any test leave a bad taste ...
You are both right, in theory everything should work ... but ... ;)

I will not blocking this, was just wondering if we not at least put a few imports in it.

daler:
OK, so in this case, how about a `cgat --help`, which should on its own import a bunch of stuff? In general a metapackage should just test the one or two things that have the most dependencies. I can add some text in the guidelines about this.

sebastian-luna-valero:
Thanks both!

Actually, now that I have converted this recipe to a metapackage, the CGAT code is not in and therefore it is not possible to run `cgat --help`. 

Please note that the use case for this metapackage is to help us with our Travis tests for our code. In our Travis configuration we:
1. download miniconda
2. install the metapackage with all our software dependencies
3. run our tests

Once we make a new release of our code and all looks green on Travis, I will create a standard conda recipe (i.e. not a metapackage) with the CGAT code in it. I will then include the `cgat --help` test.

Regarding the tests on this metapackage, I will add some imports and commands as a sanity check.

Feel free to merge when you think it is ready to go.

bgruening:
@sebastian-luna-valero just a general question for your use-case. If you need this just for testing would a simple requirements.txt file enough for your travis setup?

sebastian-luna-valero:
@bgruening there are two main reasons:

1. we used that in the past and had problems with dependency resolution.

2. our metapackage will contain both python and non-python dependencies.

bgruening:
I don't get the (2) as this is also possible with a simple requirements.txt file, but let's merge this now.

sebastian-luna-valero:
Thanks @bgruening 

I would like to streamline the installation of all our dependencies by just pulling packages available via conda so the installation procedure is exactly the same in different environments (i.e. Travis or elsewhere), that is why I am creating this metapackage.

On the other hand, I'd be interested to know how can you install non-python dependencies with requirements.txt on Travis. I have only found this: https://docs.travis-ci.com/user/languages/python/#Dependency-Management

Could you please point me to any documentation and/or example on how to do that?

PR #3280, opened 2016-12-22
Recipe crimson

bow:

Hi developers,

I'm creating this PR to add a new recipe for the [crimson](https://pypi.python.org/pypi/Crimson/0.3.0) package. This will be a dependency for another package whose PR is still in preparation. It's a fairly straightforward pure Python package that has been tested to run on Python 2.7, 3.4, and 3.5. Still, I may be overlooking something so any reviews are appreciated :).

Thanks in advance!


bgruening:
Looks great, thanks a lot!

PR #3287, opened 2016-12-22
Metameta all new recipe

pirovc:

@bioconda/core

Meta-package for the MetaMeta pipeline: install all necessary tools and dependencies (not strictly mandatory, that's why I created it as a separated package).


bgruening:
Can you add a small test, that just imparts one module or executes the kraken ...

pirovc:
Meta package is not needed any longer (switching to manual installation). I re-open it in case it becomes necessary again.

PR #3288, opened 2016-12-22
Minhash cpu version

joachimwolff:


recipes/sparse-neighbors-search/bld.bat
```diff
@@ -0,0 +1,2 @@
+"%PYTHON%" setup.py install
+if errorlevel 1 exit 1
```
bgruening: This file can be removed, we do not support Win builds.

recipes/sparse-neighbors-search/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  url: https://github.com/joachimwolff/minHashNearestNeighbors/archive/v0.2.3.tar.gz
+  md5: 49b3f9442b4caad82f8285c830296490
+
+build:
+  skip: True  # [not py27]
+  number: 0
+
+requirements:
+  build:
+    - python 2.7.12
```
bgruening: the version number is not needed here, as you skip the build in the build section

recipes/sparse-neighbors-search/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - setuptools
+    - gcc
+    - scipy
+    - numpy
+    - scikit-learn
+    - cython
+   
+
+  run:
+    - python 2.7.12
```
bgruening: here as well, just remove the version

PR #3294, opened 2016-12-22
Quast fix

lparsons:

### DIFF
--- a/recipes/quast/meta.yaml
+++ b/recipes/quast/meta.yaml
@@ -3,8 +3,7 @@ package:
   version: "4.3"
 
 build:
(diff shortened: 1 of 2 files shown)


recipes/quast/meta.yaml
```diff
@@ -16,17 +16,23 @@ requirements:
   build:
   - gcc
   - boost {{CONDA_BOOST}}*
+  - perl
```
bgruening: for some historical reasons (and yes we want to change this) we need to depend on perl-threaded

lparsons:
If this looks good, go ahead and merge (or I can)

PR #3309, opened 2016-12-23
Add featureCounts tool

upendrak:


upendrak:
Why does the Travis build failed this time? The `conda-build` command worked great

recipes/featurecounts/1.5.1/meta.yaml
```diff
@@ -0,0 +1,17 @@
+package:
+  name: featurecounts
+  version: "v1.5.1"
+
+source:
+  fn: subread-1.5.1-source.tar.gz
+  url: https://sourceforge.net/projects/subread/files/subread-1.5.1/subread-1.5.1-source.tar.gz/download
```
bgruening: the download at the end can be removed.

recipes/featurecounts/1.5.1/meta.yaml
```diff
@@ -0,0 +1,17 @@
...
+  md5: 3eb9a3ff0fc13401a1d76681549494f5
+
```
bgruening: you don't have any build and runtime dependencies defined, this does not work

bgruening:
The featurecounts tool is called subread and the recipe can be found here: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/subread/meta.yaml

recipes/featurecounts/1.5.1/meta.yaml
upendrak: Ok got it

recipes/featurecounts/1.5.1/meta.yaml
upendrak: Ok i will add it now..

bgruening:
@upendrak if you build your package locally with `conda build` you local environment can leak into the build environment, e.g. python and perl is installed and can be used. Here on travis is will not work, as we build in a build environment where no perl/python... is installed, so you need to define it.

upendrak:
Oh i see. Shall i fix it and then push the changes or you don't want me to do that since the subread is already present (that is old version though)

bgruening:
This was just to let you know what went wrong. I think we should close this PR as the package already exists.

upendrak:
Ok got it..

PR #3317, opened 2016-12-27
Add agg

druvus:

Doesn't work with llvm so using gcc for osx.

### DIFF
--- a/recipes/agg/meta.yaml
+++ b/recipes/agg/meta.yaml
@@ -16,9 +16,11 @@ source:
 requirements:
   build:
   - gcc  
+  - zlib
 
   run:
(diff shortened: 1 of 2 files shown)


recipes/agg/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  url: https://github.com/Illumina/agg/archive/v{{ version }}.tar.gz
+  md5: 76740cae2ba59de2d8b78b950b1eb499
+
+requirements:
+  build:
+  - gcc  
+  - zlib
+
+  run:
+  - libgcc  # [not osx]
```
bgruening: In this case you probably need libgcc on osx as well.

PR #3329, opened 2016-12-30
Update to v1.0.0

dariober:

### DIFF
--- a/recipes/asciigenome/meta.yaml
+++ b/recipes/asciigenome/meta.yaml
@@ -1,10 +1,10 @@
 package:
   name: asciigenome
-  version: '0.6.4'
+  version: '1.0.0'
 
 source:
-  fn: ASCIIGenome-0.6.4.zip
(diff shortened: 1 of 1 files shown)


recipes/asciigenome/meta.yaml
```diff
@@ -1,10 +1,10 @@
 package:
   name: asciigenome
-  version: '0.6.4'
+  version: '1.0.0'
 
 source:
-  fn: ASCIIGenome-0.6.4.zip
-  url: https://github.com/dariober/ASCIIGenome/releases/download/v0.1.0/ASCIIGenome-0.6.4.zip
+  fn: ASCIIGenome-1.0.0.zip
+  url: https://github.com/dariober/ASCIIGenome/releases/download/v1.0.0/ASCIIGenome-1.0.0.zip
```
bgruening: Can you please add an `md5:` checksum for this and use `ASCIIGenome-1.0.0.tar.gz`
Thanks!

dariober: Ok, I will from now on. Using .tar.gz instead of .zip requires me to change other settings elsewhere but it should be ok.

PR #3338, opened 2017-01-01
Add python-3.6

bgruening:

Build for python 3.6. ping @bioconda/core


kyleabeauchamp:
Presumably this will be a big chunk of work to get up and running fully?  Do you think things are "stable enough" on the continuum and conda-forge side of things to proceed here?

bgruening:
Not sure, no. But people starting to discuss dropping 34 in favor or 35/36 ... and we should also start discussing and preparing.

kyleabeauchamp:
FWIW, pysam has recently enabled 3.6 builds:

https://github.com/pysam-developers/pysam/pull/394

johanneskoester:
Also, since Miniconda ships with 3.6 now, there sometimes occur conflicts when trying to install Python 3.5 packages from bioconda into the root environment. I'll give this another try.

johanneskoester:
Yay, we need to fix a lot of lints :-).

daler:
Upon fixing a lint, do you want to bump the build number? I can start a lint-blacklist for this.

johanneskoester:
I think we can keep the build number unchanged. The lints are only about the way the recipe is written and won't affect the contents, right?

bgruening:
We have a few lints, like has_setuptools, that do effect the package slightly @johanneskoester do you have news about the conda-build version with new pinning support. If this is released soon we might want to do all of this together as we probably need to touch every package. R also needs to be bumped.

daler:
A recipe currently missing tests that has new ones added should probably be bumped to catch any issues. And same with any that are changed from perl-threaded to perl or java-jdk to openjdk.

I think we're going to have to move off of travis for these mega-updates.

johanneskoester:
Let's see whether this can be done in batches again. I would first fix the lints and then update to 3.6.

johanneskoester:
We could also have a look at this: https://buildkite.com

PR #3339, opened 2017-01-02
Bump build number for perl-gd

mvdbeek:

Bumped build in the hope that jpeg >=9 is being recognized as runtime dependency.
Seemed to work locally.
Ping @nsoranzo

### DIFF
--- a/recipes/perl-gd/meta.yaml
+++ b/recipes/perl-gd/meta.yaml
@@ -7,7 +7,7 @@ source:
   md5: c4b3afd98b2c4ce3c2e1027d101a8f1e
 build:
   ##ignore_prefix_files: True
-  number: 4
+  number: 5
   skip: True  # [osx]
 
 requirements:
@@ -18,12 +18,12 @@ requirements:
     - libgd >=2.2.3
     - perl-module-build
     - perl-extutils-makemaker
-    
+
   run:
     - libgcc # [not osx]
     - perl-threaded >=5.22.0
     - libgd >=2.2.3
-    
+
 test:
   # Perl 'use' tests
   imports:
@@ -31,7 +31,7 @@ test:
     - GD::Simple ## (no version defined)
     - GD::Polygon ## (no version defined)
     # - GD::Polyline ## (version 0.2 defined?!)
-    
+
 about:
   home: http://metacpan.org/pod/GD
   license: perl_5


nsoranzo:
I don't think that's going to fix https://github.com/bioconda/bioconda-recipes/pull/3256#issuecomment-268292534 because `perl-gd` was already built using the latest `gd`, but `conda create` does not pick it up for some reason. The only solution I found is to pin the `gd` version and build.

mvdbeek:
I agree with your diagnosis, but perhaps the fault is not at `conda create`, but at `conda-build`?
In that case it could be sufficient to rebuild. I've uploaded my package to anaconda, if you do `conda create -n perl_gd_5 perl-gd -c mvdbeek` you'll get the correct libjpeg.9.so. 
In any case this package, as installed by doing `conda create -n test perl-gd`, is broken, so we can hardly make things worse by bumping the version.

mvdbeek:
Thanks @bgruening ... unfortunately that didn't help, perl-gd is still pulling in the wrong libgd/jpeg dependency. Looking deeper into this it seems that `libtiff` is pulling in `jpeg=8d-2`, which I suppose makes the resolver decide on the older libgd.

mvdbeek:
Hmm, putting conda-forge first seems to solve the problem:
```
mvandenb@damagecase:~/src/tools-iuc/tools/yaha$ conda create -n testperlgd perl-gd -c conda-forge
Solving package specifications: ..........

Package plan for installation in environment /home/mvandenb/miniconda3/envs/testperlgd:

The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    jpeg-9b                    |                0         918 KB  conda-forge
    zlib-1.2.8                 |                3          87 KB  conda-forge
    perl-gd-2.56               |                5          82 KB  bioconda
    ------------------------------------------------------------
                                           Total:         1.1 MB

    expat:         2.1.0-2   conda-forge
    fontconfig:    2.12.1-1  conda-forge
    freetype:      2.6.3-1   conda-forge
    icu:           56.1-4    conda-forge
    jpeg:          9b-0      conda-forge
    libgcc:        5.2.0-0              
    libgd:         2.2.3-2   conda-forge
    libiconv:      1.14-4    conda-forge
    libpng:        1.6.26-0  conda-forge
    libtiff:       4.0.6-7   conda-forge
    libxml2:       2.9.4-3   conda-forge
    perl-gd:       2.56-5    bioconda   
    perl-threaded: 5.22.0-10 bioconda   
    xz:            5.2.2-0   conda-forge
    zlib:          1.2.8-3   conda-forge

Proceed ([y]/n)?
```

PR #3342, opened 2017-01-03
Clean up roprofile recipe

druvus:


druvus:
@bgruening I did a typo in the package name of the first version. The name should be without ".py" suffix. Maybe you could delete the "roprofile.py" package from anaconda to avoid confusion? 

Got patches added upstream so the previous patch is not needed anymore.

bgruening:
@druvus done. Happy new year!

druvus:
@bgruening thanks! Happy New Year!

PR #3349, opened 2017-01-03
perl conda-forge migration

bgruening:

This converts the perl-threaded package into a meta-package that installs the `perl` package.
Unfortunately, we also have a `perl` package in bioconda. So we need to remove this package from our channel to make the meta-package work and actually installing the threded-enabled perl package from conda-forge.

Next step would be to change our guidelines and change all perl-* packages over time to use `perl` instead of `perl-threaded` and than fix bioperl.

fixes: https://github.com/bioconda/bioconda-recipes/issues/3167
ping @bioconda/core @bioconda/perl


bgruening:
@jerowe maybe we want to also look at this and raise concerns :)

jerowe:
@bgreuning, I think I'm good. ;-) I've been building against the conda-forge perl and am not having problems so far.

It would be nice to move the perl-* packages from bioconda to conda-forge, but getting recipes into conda-forge can be a rather slow process.

bgruening:
@jerowe this PR makes some kind of strange move. We convert perl-threaded to a meta-package installing `perl` (hopfeully from conda-forge). You are a perl expert and I hope this works. If this is merged we can slowly migrate all perl-* packages from `perl-theraded` to `perl`.

jerowe:
I've been following the discussion. I see why its kind of odd, but hopefully its just a place holder. 

I tend to do my perl recipes in batches (when I need to get something released usually ;-) ). I don't mind incrementally rebuilding some recipes to use the conda-forge perl as I go along. Do you think that is a good idea, or it should it be a more organized effort?

bgruening:
> I have been building all recipes today using perl as opposed to perl-threaded, and can confirm that they are pulling from conda-forge.

Are you sure? Ok, that would solve all my problems but then it conflicts with my understanding of channel priorities.

jerowe:
I don't know why I keep thinking that perl-threaded is already pointing at the conda-forge perl, but I have been able to build recipes with perl (instead of perl-threaded) that have dependencies on perl-threaded, and so far they are all fine.

```
    libgcc:                 5.2.0-0             defaults 
    perl:                   5.20.3.1-2          conda-forge  
    perl-app-cpanminus:     1.7039-2            bioconda
...
   perl-threaded:          5.22.0-10           bioconda
```

bgruening:
@johanneskoester @bioconda/perl 
I have `5.22.0.1` in conda-forge, which should correspond to our `perl-5.22.0` and `perl-threaded-5.22.0`.

If you agree I will:

* [ ] remove the `perl` package completely from the bioconda channel
* [ ] merge this PR, so that perl-threaded will install the conda-forge `perl` version

johanneskoester:
Ok, let's do that.

bgruening:
Ok, done!
@jerowe any testing appreciated :)

bgruening:
https://github.com/bioconda/bioconda-utils/pull/56

recipes/perl-threaded/meta.yaml
```diff
@@ -1,21 +1,15 @@
...
-  number: 10
-  skip: False
+  number: 11
 
 requirements:
-  build:
-    - gcc   # [linux]
-    - llvm  # [osx]
+  run:
+    - perl
```
nsoranzo: Should this be:
```
  - perl 5.22.0
```
? Otherwise a package requiring the exact version 5.22.0 of perl-threaded may get any version of perl.

johanneskoester: Sounds reasonable for me. @bgruening?

bgruening: Yes, we can do this. I not able to do it today. @nsoranzo feel free to create a PR

nsoranzo: https://github.com/bioconda/bioconda-recipes/pull/3446

Issue #2770, opened 2016-11-02
Packages that need to be pinned


johanneskoester:
This issue is to collect problems with packages where a newer version of a dependency has been discovered to cause breakage (because of ABI changes).

| to pin | in package |
| ------- | ------------- |
| libjpeg | r-sartools |

lecorguille:
Question: how do you do that?

johanneskoester:
One adds a key-value pair to the [env matrix](https://github.com/bioconda/bioconda-recipes/blob/master/scripts/env_matrix.yml), e.g. `CONDA_LIBJPEG`. This becomes available as an env variable in the recipe. In the recipe, that variable is used to pin the version in the run requirements and in the build string, see [here](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/merfishtools/meta.yaml).

PR #3350, opened 2017-01-03
jpeg pinning

bgruening:

ping @bioconda/core 
Imho we need to start pinning more packages - let's start with jpeg

I did not changed the build-string, as I don't think this scales. We have many package with a bunch of dependencies that needs to be pinned. Changing the build-string is a manual process and error prone. If we increase the version of a library we should increase the build-number. This we can do step-by-step without the need of a blacklist.


bgruening:
Maybe related? https://github.com/bioconda/bioconda-recipes/issues/2770

johanneskoester:
Do we need the additional pins right now? I'm asking, because I hope that Mike will soon release a new pinning system that addresses all these issues directly in conda-build, see conda/conda-build#1585. We could think about pushing that effort a bit, by testing or contributing code or ideas.

bgruening:
I had the impression that this PR will take some time until it is finished. I'm fine with postponing our pinning work, but it feels like more and more packages do break here or as soon as conda-forge updates one of these requirements. There is also a new zlib upstream release ... I'm really scared :)

johanneskoester:
That's great news @msarahan! In this case, I'd vote for using that approach. I'm afraid we cannot really test it directly on Bioconda until there is a stable release though. Too many people rely on our build system to work properly. But maybe we can try it locally once merged into the conda-build master.

bgruening:
@msarahan awesome news! 
@johanneskoester we could re-compile some part of the stack locally or on some cluster at least for Linux builds?

johanneskoester:
@bgruening yes, we can do that.

PR #3352, opened 2017-01-04
Add recipes for rdfextras

bebatut:


recipes/rdfextras/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+requirements:
+  build:
+    - python
+    - setuptools
+    - rdflib >=3.2.1
+    - pyparsing <=1.5.7
+
+  run:
+    - python
+    - setuptools
```
bgruening: setuptools is not a runtime dependency, isn't it?

PR #3364, opened 2017-01-04
refactor java-jdk build script

rvalieris:

refactor java-jdk build script to makes sure only necessary files go into the package

I noticed recently that the java-jdk package had the tar.gz release in the root directory, I believe this is a mistake as it doubles the package size for no apparent reason, as it can be [seen here](https://anaconda.org/bioconda/java-jdk/files) the linux builds have double the size.

I also took this opportunity to explicit move files into the package instead of moving "everything left", to make it more clearer what is being included.

### DIFF
--- a/recipes/java-jdk/build.sh
+++ b/recipes/java-jdk/build.sh
@@ -22,7 +22,7 @@ mv $NAME/* .
 rm -r $NAME
 
 # clean up
-rm -rf release README readme.txt Welcome.html *jli.* demo sample *.zip
+rm -rf release README readme.txt Welcome.html *jli.* demo sample *.zip *.tar.gz ASSEMBLY_EXCEPTION THIRD_PARTY_README DISCLAIMER LICENSE
 
 if [[ `uname` == "Linux" ]]
 then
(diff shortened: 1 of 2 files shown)


bgruening:
@rvalieris should we deprecate this package in favor of https://github.com/conda-forge/openjdk-feedstock/blob/master/recipe/meta.yaml ?

We could make a last release similar to https://github.com/bioconda/bioconda-recipes/pull/3349

rvalieris:
oh !! I actually didn't noticed conda-forge already had openjdk, I guess I just searched for java and didn't found anything..

yes that makes sense,  we should favor using the conda-forge package. I will close this PR.

bgruening:
@rvalieris can you update the docs for Java stating that new recipes should use openjdk?

rvalieris:
you mean https://bioconda.github.io/guidelines.html ? I am not sure I can, the repo doesn't seem to accept PRs.

bgruening:
Yes this one: https://bioconda.github.io/guidelines.html#java
Code is here: https://github.com/bioconda/bioconda-utils/blob/master/docs/source/guidelines.rst

PR #3368, opened 2017-01-04
Kat updates

JenCabral:

As is, this does not build.

### DIFF
--- a/recipes/kat/build.sh
+++ b/recipes/kat/build.sh
@@ -25,7 +25,8 @@ sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoscan
 sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoupdate
 
 ./autogen.sh
-./configure --prefix=$PREFIX
-##--with-boost=${PREFIX}
+./configure --prefix=${PREFIX} \
+        --with-boost-libdir=${PREFIX}/lib \
+        --with-boost=${PREFIX}
 make
 make install
--- a/recipes/kat/build.sh
+++ b/recipes/kat/build.sh
@@ -2,9 +2,6 @@
 
 set -x -e
 
-export CC=${PREFIX}/bin/gcc
-export CXX=${PREFIX}/bin/g++
-
 export INCLUDE_PATH="${PREFIX}/include"
 export LIBRARY_PATH="${PREFIX}/lib"
 export LD_LIBRARY_PATH="${PREFIX}/lib"
@@ -28,6 +25,7 @@ sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoscan
 sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoupdate
 
 ./autogen.sh
-./configure --prefix=$PREFIX --with-boost=${PREFIX}
+./configure --prefix=$PREFIX
+##--with-boost=${PREFIX}
 make
 make install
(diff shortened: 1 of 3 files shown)


bgruening:
The issue was discussed here: https://github.com/TGAC/KAT/issues/16

jerowe:
One of the problems, at least with building the 2.0.8 version, is that there is a bug in the conda-forge automake. I have raised an issue addressing the problem [here](https://github.com/conda-forge/automake-feedstock/issues/10)

jerowe:
How are you guys pulling in these channels from defaults? Do i have my channel order wrong?

```
  autoconf:   2.69-pl5.22.0_3    bioconda

    automake:   1.15-0             defaults

    boost:      1.60.0-py35_0      defaults

    cloog:      0.18.0-0           defaults

    fontconfig: 2.12.1-0           defaults

    freetype:   2.5.5-1            defaults

    gcc:        4.8.5-7            defaults

    gmp:        6.1.0-0            defaults
```

jerowe:
FYI

```
WARNING conda.lock:touch(53): Failed to create lock, do not run conda in parallel processes [errno 36]
```

Is a bug that was introduced after conda 4.1.9. I have a bug report/PR up for it, but I'm not sure what the status is. Also, conda-env was merged with conda proper, so installing both isn't necessary.

bgruening:
The lock should not be the problem, it's just a warning and other recipes turn green even with this.
The autoconf issue can be fixed with this temporary: https://github.com/bioconda/bioconda-recipes/blob/27ebbe045174267682315e008b6e682f416b7ca8/recipes/snp-sites/build.sh#L2-L12

But the error has something to do with boost, isn't it?

jerowe:
I couldn't figure out if it was zlib or boost on my last build. The first time around I got the error about /usr/bin/perl, and jumped down that rabbit hole. I'm trying out some builds on my local conda-forge image now.

daler:
@jerowe, you can double-check your channel order here: https://bioconda.github.io/#set-up-channels

jerowe:
This should  be the build.sh

```
#!/bin/sh

set -x -e

export INCLUDE_PATH="${PREFIX}/include"
export LIBRARY_PATH="${PREFIX}/lib"
export LD_LIBRARY_PATH="${PREFIX}/lib"

export BOOST_INCLUDE_DIR=${PREFIX}/include
export BOOST_LIBRARY_DIR=${PREFIX}/lib
export LIBS='-lboost_system -lboost_program_options -lboost_filesystem -lboost_timer'

export CXXFLAGS="-DUSE_BOOST -I${BOOST_INCLUDE_DIR} -L${BOOST_LIBRARY_DIR}"
export LDFLAGS="-L${BOOST_LIBRARY_DIR} -lboost_filesystem -lboost_system"

# fix automake
sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/aclocal
sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/automake

# fix autoconf
sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autom4te
sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoheader
sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoreconf
sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/ifnames
sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoscan
sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoupdate

./autogen.sh
./configure --prefix=$PREFIX \
        --with-boost-libdir=${PREFIX}/lib \
        --with-boost=${PREFIX}
make
make install
```

I think I will go check out the original builds to see what happened there. We use kat, and I'd be surprised if it going without boost would be go unnoticed for long...

bgruening:
@jerowe thanks I updated this recipe with the `--boost` config options, let see ... :)

bgruening:
Ha! That was it. Thanks a bunch @jerowe!

bgruening:
@JenCabral I think with the tip of @jerowe it should be easy to update it now.

JenCabral:
This is great. Thanks @jerowe

PR #3378, opened 2017-01-05
HiCUP

daler:

Add HiCUP

### DIFF
--- a/recipes/hicup/build.sh
+++ b/recipes/hicup/build.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+set -eu -o pipefail
+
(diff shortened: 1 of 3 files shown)


bgruening:
Nice! This was on my todo list as well. With hicexplorer and hicbrowser we have a very nice pipeline :)

Thanks!

daler:
Great! My perl-fu is weak, so it took me a while to figure out how to get the symlinked scripts working. I think it was worth it though since it keeps `$PREFIX/bin` clean.

Didn't know about hicexplorer -- I'll probably be using that too, thanks.

PR #3379, opened 2017-01-05
Add OSX build for bioconductor-rhtslib.

acaprez:

Add OS X build for bioconductor-rhtslib.  xref: #3373

### DIFF
--- a/recipes/bioconductor-rhtslib/build.sh
+++ b/recipes/bioconductor-rhtslib/build.sh
@@ -4,6 +4,9 @@
 # "Priority: Recommended"
 mv DESCRIPTION DESCRIPTION.old
 grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
+if [ `uname` == "Darwin" ]; then
+        patch -p0 -i ${RECIPE_DIR}/rpath.patch
+fi
 #
 $R CMD INSTALL --build .
 #
--- a/recipes/bioconductor-rhtslib/build.sh
+++ b/recipes/bioconductor-rhtslib/build.sh
@@ -4,9 +4,6 @@
 # "Priority: Recommended"
 mv DESCRIPTION DESCRIPTION.old
 grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
-if [ `uname` == "Darwin" ]; then
-        patch -p0 -i ${RECIPE_DIR}/rpath.patch
-fi
 #
 $R CMD INSTALL --build .
 #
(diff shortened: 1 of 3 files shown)


bgruening:
@acaprez I think the problem is here: https://travis-ci.org/bioconda/bioconda-recipes/jobs/189332788#L517
Maybe libgfortran can help, or simply gcc on OSX?

acaprez:
@bgruening - it's actually an issue caused by the OS X SIP they added in El Capitan.  The `libgcc` package is getting pulled in as an R dependency, so `libgfortran` is there.  As part of SIP though, the `DYLD_LIBRARY_PATH` variable gets ignored/purged.  I disabled SIP on my dev VM a while back; that's why it build ok for me there.  The Travis CI VM's apparently still have it enabled though.

I was able to work around it by patching the `Makevars` file to pass an `rpath` option to the linker.  But in testing, I ran into an issue where `conda build` seems to ignore patch files if the `--dirty` flag is used.  Building on my local VM (with SIP re-enabled) failed using `simulate-travis.py` since it uses `--dirty`.  The same `conda build` command worked without it.  I went ahead and pushed the patch just now, but it's likely it still won't build since `--dirty` is used.

bgruening:
@acaprez sorry what is SIP? :)
Thanks for the detective work!

acaprez:
Sorry, System Integrity Protection: https://support.apple.com/en-us/HT204899.  Basically, root is no longer root - there are certain locations not even root can change.  It also includes other protections, one of which is to ignore `DYLD_LIBRARY_PATH` (Mac equivalent of `LD_LIBRARY_PATH`).  It's caused headaches with a lot of OS X packages, including Conda:  https://github.com/ContinuumIO/anaconda-issues/issues/498.

As I mentioned, I was able to get around it by patching initially, but it appears to be a bug in `conda build` to ignore patch files when `--no-build-id` is used.  Haven't had a chance to dig through open issues for `conda` to see if anyone else has reported it.

bgruening:
Uh is this what we have to fight from now on regularly? If so maybe we should use an older OSX image on Travis?

acaprez:
That might be a viable option, yeah.  The concern would obviously be that packages would build and test ok on Travis, but then not work on OS X 10.11+.  Basically any package that uses `DYLD_LIBRARY_PATH` at runtime would break.  Since `conda` does the relative rpath conversion at build time normally they shouldn't be, unless it's via custom `activate.d/deactivate.d` scripts or the like.

It should be possible in theory to patch packages to build using the current OS X image and not rely on `DYLD_LIBRARY_PATH`, but it is more effort.  In any event, I'd be interested in getting input from others familiar with OS X before any changes are made.

PR #3383, opened 2017-01-05
gtf_to_genes

jj-umn:
Python parser which caches all the genes / transcripts from a GTF file
and caches the data into python classes for high speed access.


recipes/gtf_to_genes/bld.bat
```diff
@@ -0,0 +1,8 @@
+"%PYTHON%" setup.py install 
+if errorlevel 1 exit 1
```
bgruening: @jj-umn please don't include these bat files. We do not compile for windows.
Let's merge it for now.

PR #3387, opened 2017-01-06
Stacks 1.43 and 1.44

lecorguille:


recipes/stacks/build.sh
```diff
@@ -6,5 +6,7 @@ export CPATH=${PREFIX}/include
  sed -i.bak "41 c extern int encoded_gtypes[4][4];" src/populations.cc
 
 ./configure --prefix=$PREFIX --enable-sparsehash --enable-bam --includedir=${PREFIX}/include/bam --libdir=${PREFIX}/lib
+cp Makefile Makefile.bk
+sed "s|^LIBS=.*$|LIBS=\"-I$PREFIX/include/ -L$PREFIX/lib/ -lz -lgomp\"|" Makefile.bk > Makefile
```
bgruening: you could use `sed -i.bak "s|^LIBS=.*$|LIBS=\"-I$PREFIX/include/ -L$PREFIX/lib/ -lz -lgomp\"|" Makefile`

lecorguille: Sure!
Actually, I didn't look at that file ... I just change the number version 😝 
I'm rather hurry about this PR. Are you OK without this modification?

One of my collegue want to launch jobs this weekend and I can't install Stacks since 1.42 on our HPC (thanks to CentOS 6.5)

I ❤️ Conda

lecorguille:
Thanks!

PR #3392, opened 2017-01-07
Enable osx build for ete2

druvus:

### DIFF
--- a/recipes/ete2/meta.yaml
+++ b/recipes/ete2/meta.yaml
@@ -8,8 +8,8 @@ source:
   md5: 87bf020c25c293e8e4e1b1e02a3dc2b1
 
 build:
-  number: 1
-  skip: true # [not py27 or osx]
+  number: 2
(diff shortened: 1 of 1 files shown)


bgruening:
Sometimes, it's so easy with OSX :)
Thanks!

recipes/ete2/meta.yaml
```diff
@@ -8,8 +8,8 @@ source:
   md5: 87bf020c25c293e8e4e1b1e02a3dc2b1
 
 build:
-  number: 1
-  skip: true # [not py27 or osx]
+  number: 2
```
bgruening: One thing I noticed. In such a case I don't think you need to increase the build number. A version with this build does not exists on OSX, so bioconda will create it. You just need to increase the number if such a release already exists in either OSX, or Linux.

druvus: Thanks, good to know!

PR #3399, opened 2017-01-08
Feature/fix rsem

jerowe:

@bgruening , can you take a look at this? There are a lot of things I ad to change with this recipe.


recipes/rsem/meta.yaml
```diff
@@ -1,27 +1,54 @@
...
-  version: 1.2.28
+  name: {{ name }} 
+  version: {{ version }}
+
 build:
   rpaths:
     - lib/R/lib/
     - lib/
-  number: 0
+  number: 2
```
bgruening: build number 1?

recipes/rsem/meta.yaml
```diff
@@ -1,27 +1,54 @@
...
+    - htslib 1.3
+    - zlib
     - r
+    - perl
+    - bioconductor-biobase
+    - perl-module-build
   run:
+    - libgcc
+    - zlib
+    - boost
```
bgruening: please pin boost and change the build-string

recipes/rsem/build.sh
```diff
@@ -34,6 +37,55 @@ rsem-tbam2gbam \
...
+export LD_LIBRARY_PATH="${PREFIX}/lib"
+
+export LDFLAGS="-L${PREFIX}/lib"
+export CPPFLAGS="-I${PREFIX}/include"
+
+########################################################
+### Prepare Samtools/Htslib 
+########################################################
+
+sed -i.bak 's/^CPPFLAGS =$//g' samtools-1.3/Makefile
```
bgruening: If you depend on samtools why do you compile this version as well?

recipes/rsem/build.sh
```diff
@@ -34,6 +37,55 @@ rsem-tbam2gbam \
...
+
+# Remove rdynamic which can cause build issues on OSX
+# https://sourceforge.net/p/samtools/mailman/message/34699333/
+sed -i.bak 's/ -rdynamic//g' samtools-1.3/Makefile
+sed -i.bak 's/ -rdynamic//g' samtools-1.3/htslib-1.3/configure
+
+########################################################
+### Build htslib 
+########################################################
+cd samtools-1.3/htslib-1.3
```
bgruening: Same here, the libs and includes should be in your $PREFIX as you depend on it, already.

recipes/rsem/Makefile.new
```diff
@@ -0,0 +1,158 @@
+CC = g++
+CFLAGS = -Wall -c -I.
```
bgruening: I guess both Makefiles, the new and the orig, are not needed in this PR as you have the patch?

recipes/rsem/build.sh
jerowe: RSEM needs to compile against specific samtools libraries, and for my first iteration I just wanted to get it working. I'm going to rework the `Makefile` it so I don't have to do this.

jerowe: Upon further inspection rsem requires some header files that are not in the conda package. I run make to get the headers, but use the bioconda package for the actual installed versions.

bgruening: That would be great! Thanks!

jerowe:
Merge?

bgruening:
Merge if you like! :)

PR #3401, opened 2017-01-08
canu added missing dependency

mictadlo:


recipes/canu/meta.yaml
```diff
@@ -9,6 +9,7 @@ source:
   md5: f0115f042c79cb5c1a247743236cd83b
 
 build:
+    number: 2
```
bgruening: build-numbers do start with 0, so the first iteration should be 1.

mictadlo: fixed

PR #3402, opened 2017-01-09
randfold

HassanAmr:


HassanAmr:
Hi @bgruening,
This package needs a tool named SQUID (by Sean Eddy: http://eddylab.org/software.html), but he doesn't suggest using it since he changes it at will. That's why I didn't do a separate package for it. Should I still put it in a separate package?

Both tools (randfold and SQUID) have the same license.

bgruening:
This is ok, I think!

PR #3404, opened 2017-01-09
Paraview depends on old autobahn version

ThomasWollmann:


bgruening:
There is no conda version < than 0.16.1 :(

bgruening:
So I guess we have two ways. Create a new autobahn package, or change VTK_USE_SYSTEM_AUTOBAHN:BOOL=ON to OFF.

ThomasWollmann:
I recommend to not create a package for the old version. We can use the new package, if they upgraded their dependencies.

PR #3415, opened 2017-01-09
update rseqc to 2.6.4

rvalieris:
this update explicitly pulls the dependencies
rather than bundling them inside the package

previous version of the package was bundling the dependencies (like pysam) instead of depending on the pysam package.


bgruening:
This is great!!!

PR #3416, opened 2017-01-09
cromwell 23 -> 24

tomkinsc:


tomkinsc:
@bioconda/core: What is the recommended strategy for debugging quay/mulled errors like [this](https://travis-ci.org/bioconda/bioconda-recipes/jobs/192811691)?

bgruening:
It seems your test does not work, so the grep does not find the string,
If you have simulate_travis running locally you have this container also locally and can just start it and try it inside the container.

Otherwise, I would remove the grep and see if you get any output on stderr/stdout

tomkinsc:
Thanks for the reminder about the `simulate-travis.py` script; I've been busy lately and out of the loop on some of the exciting new bioconda changes. I'll see what I can suss out.

tomkinsc:
It [appears to be](https://travis-ci.org/bioconda/bioconda-recipes/jobs/193066215#L3678) running correctly, so I'm not quite sure why the test was failing before. It builds for me locally, but fails with `simulate-travis.py`. Could it be related to the fact the test redirects stderr to stdout?

rvalieris:
it seems cromwell --help is returning exit code 1, this is interpreted as a failure, the "| grep" sidesteps this problem because now the exit code that matters is from grep.

tomkinsc:
Right, but even with grep the test still fails for some reason (stream redirection as part of the test process?).

rvalieris:
perhaps the tests are being run with `set -o pipefail` ?

tomkinsc:
@bioconda/core: Any insight here?

tomkinsc:
I'd love some help figuring out why this is failing.

daler:
@tomkinsc some explanation on how I arrived at adding `python` as run dep:

- I ran `simulate-travis.py` locally with `--loglevel=debug`
- the build succeeded but the mulled test did not
- looking at the output from mulled and recent `docker images | head` showed the hash of the docker container used by mulled-build
- I started that up in an interactive docker container, ran `cromwell`, and got the error `env: can't execute 'python': No such file or directory`. 
- So I added python as a runtime dep
- `simulate-travis.py` then succeeded locally

Let's see how it goes on travis . . .

daler:
@tomkinsc took a while with all the osx outages on travis, but that seems to have done the trick!

rvalieris:
Nice @daler,
any idea why that error didn't appeared on the travis log  ?

daler:
I'm not sure mulled-build passes all stdout/stderr through when it hits an error. @bgruening any ideas?

bgruening:
stderr should be printed I see it for example when zlib is missing. I think is this case the installation hangs ... Not really sure what we can do better here. But I will put it on my todo list.

PR #3420, opened 2017-01-10
Typo in determining options. msgf_plus should be able to run its edu.…

glormph:
…ucsd. modules to run as well

* [X ] This PR updates an existing recipe.

### DIFF
--- a/recipes/msgf_plus/msgf_plus.py
+++ b/recipes/msgf_plus/msgf_plus.py
@@ -71,7 +71,7 @@ def main():
     jar_dir = real_dirname(sys.argv[0])
     (mem_opts, prop_opts, pass_args) = jvm_opts(sys.argv[1:])
 
-    if pass_args != [] and pass_args[0].startswith('eu'):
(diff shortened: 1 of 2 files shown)


bgruening:
You need to increase the build number, otherwise nothing is tested.

recipes/msgf_plus/msgf_plus.py
```diff
@@ -71,7 +71,7 @@ def main():
     jar_dir = real_dirname(sys.argv[0])
     (mem_opts, prop_opts, pass_args) = jvm_opts(sys.argv[1:])
 
-    if pass_args != [] and pass_args[0].startswith('eu'):
+    if pass_args != [] and pass_args[0].startswith('edu'):
```
bgruening: edu and eu, or only edu?

glormph: Only edu, MSGFplus jar contains two executables that have to be run with `-cp` instead of `-jar`. Both of them are called `edu.ucsd.something_more`.

glormph:
It works! Thanks for the quick check Björn!

PR #3421, opened 2017-01-10
Add ATAC recipe

bebatut:


recipes/atac/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+build:
+  number: 0
+  skip: True # [py3k]
+
+requirements:
+  build:
+    - gcc # [not osx]
+    - llvm # [osx]
+    - python    
+    - perl-threaded
```
bgruening: From today on you can use `perl`.

PR #3422, opened 2017-01-10
Add recipe for PASTA version 0.2

gregvonkuster:

@bgruening The additional github repositories retrieved using wget in the build script (e.g., https://github.com/smirarab/sate-tools-linux/archive/a3e6f56.tar.gz) are necessary for this recipe.  They cannot be retrieved using a separate recipe and then required by this one because the Python pasta components wrap the various sate tools that are included in the separate repositories and the setup.py script requires them in a specified location (relative to the script) for proper installation.  I've corrected a problem with the pasta code (https://github.com/smirarab/pasta/issues/17) which was causing tests to fail in previous attempts for this recipe, so I'm hopeful that this will pass all tests and be acceptable.

### DIFF
--- a/recipes/pasta/0.2/build.sh
+++ b/recipes/pasta/0.2/build.sh
@@ -1,0 +1,56 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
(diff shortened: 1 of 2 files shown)


gregvonkuster:
Thanks @bgruening!!!  ;)

PR #3423, opened 2017-01-10
Added taxonlist.tsv

rasmusagren:


bgruening:
Thanks!

rasmusagren:
@bgruening I have a question regarding how to do if a package depends on standard command line tools such as wget, git or gzip. Should they be explicitly added? For example, some functionality of this tool requires wget > 1.16 but it's not required for installing. Should it then be added?

bgruening:
Yes. Everything needs to be added. Don't assume anything in the runtime and build time environment.

rasmusagren:
Ok, thanks. Maybe it could be good to remove wget/bzip2/git/unzip from the bioconda docker image to make it more transparent when you're missing a dependency like this?

PR #3425, opened 2017-01-10
deprecate java-jdk

rvalieris:

turns java-jdk into a metapackage, dependent on [conda-forge::openjdk](https://github.com/conda-forge/openjdk-feedstock), see also: https://github.com/bioconda/bioconda-recipes/pull/3364

ping @bgruening


bgruening:
I like this a lot! Thanks!

bgruening:
@bioconda/core ok?

chapmanb:
Thanks for working on this. It doesn't look like the conda-forge version has the fixes for fonts I put in place a while back:

https://github.com/bioconda/bioconda-recipes/commit/e42534a22eb9b2fb4c9316faddac7133f4f824b5#diff-2ffd9cd8a6121b9d5f198f795ffc19b6

https://github.com/conda-forge/openjdk-feedstock/blob/master/recipe/build.sh

I guess this is a little hacky but was needed to get java to run correctly on any kind of minimal system. I depend on this so would like to see it included before removing the bioconda version.

bgruening:
@chapmanb can you get your patches upstream?

For Java I think we do not need to remove any packages from the channel as we do not have a name conflict. The old version will stay.

johanneskoester:
That's my fault @chapmanb. I simplified the build script and thought this was not needed without properly checking, sorry. I have re-added it, see PR conda-forge/openjdk-feedstock#1.

chapmanb:
Johannes -- awesome, thanks so much for updating this and navigating all the conda-forge infrastructure. Wow, the number of checks/builds there is overwhelming or impressive, depending on how you look at it.

johanneskoester:
Yeah, I have some doubts if it scales in the end. Everything is much slower than here at bioconda. The build system seems also less mature, at least at the moment. I hope that the fix can be merged soon.

bgruening:
New package is available in conda-forge!
Thanks Johannes and @rvalieris!

PR #3429, opened 2017-01-10
Chicago

daler:

CHiCAGO bioconductor package plus dependency and example data package

### DIFF
--- a/recipes/bioconductor-chicago/build.sh
+++ b/recipes/bioconductor-chicago/build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+mv DESCRIPTION DESCRIPTION.old
+grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
+$R CMD INSTALL --build .
--- a/recipes/bioconductor-chicago/meta.yaml
+++ b/recipes/bioconductor-chicago/meta.yaml
@@ -1,0 +1,32 @@
+package:
+  name: bioconductor-chicago
+  version: 1.2.0
+source:
+  fn: Chicago_1.2.0.tar.gz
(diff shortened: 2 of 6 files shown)


bgruening:
ping @joachimwolff :)

daler:
@bgruening @joachimwolff I'd like to do chicagoTools as well, but it has no releases. Figured I'd give it a test on some real data first to see if it's useful before writing a recipe.

joachimwolff:
I have created a chicagoTools recipe: https://github.com/joachimwolff/bioconda-recipes/tree/chicagotools/recipes/chicagotools

joachimwolff:
The example data available is incomplete e.g. to run bam2chicago.sh a bam file is missing. I wrote to the authors of chicago and they send me one file. The only issue is that this file is 14 GB large and I do not know if it should be in the example data package or not.

daler:
Do you mean the chicagoTools-specific example data or the PCHiC data bioconductor package?  And is that 14GB file hosted somewhere? We can include a download script to get that data and put it in the right place that can be run by the user after installing the package.

joachimwolff:
It is not hosted anymore but I have it on my hard drive. I have to ask the authors if it is ok to make the file public available. 
The issue is the following: If you want to run bam2chicago.sh and you are looking in the PCHiC data dir at inst/extdata/hg19TestDesign all files that are needed are there, except a bam file. I asked the authors of chicago if they can send me a file bam file that can be used together with the provided files and they sent me that 14 GB bam file. 
@bgruening Are we having somewhere space to host it?

daler:
So the current solution is to manually send that file to anyone who realizes there's missing data (and then asks for it)? This should be resolved by the authors, not bioconda. In any case, I don't think this should block your chicagoTools recipe since we don't run functional tests on bioconda anyway. But it is good for potential users to know.

PR #3430, opened 2017-01-10
Updated RepeatMasker recipe (more tests, more usability)

pvanheus:
* Add more substantial test
* Use REPEATMASKER_DIR, REPEATMASKER_LIB_DIR and REPEATMASKER_MATRICES_DIR to set locations for
  the RepeatMasker installation, the Libraries directory and Matrices directory. Required
  to use own, externally supplied, Libraries (e.g. latest RepBase).
* Suppress complaint about age of library when the minimal library is being used.
* Harmonise use of variables in RepeatMasker code: use REPEATMASKER_LIB_DIR
  instead of REPEATMASKER_DIR/Libraries.


bgruening:
@johanneskoester @daler using the test-files as in this PR will skip the Docker tests. This is currently not supported via the Docker tests.
I think it is ok to merge this one and I should try to support this in Docker somehow, but such tests will not catch missing libgcc and other stuff.

@pvanheus alternatively, can you keep this single very simple `command` test in addition to the more elaborated test-file. This test would then be run in Docker as well and at least test the container and potential libgcc issues.

johanneskoester:
Yes, I'd also vote to keep the simple command test in addition.

pvanheus:
@johanneskoester @bgruening I've added the simple command test as requested.

bgruening:
Thanks @pvanheus!

PR #3435, opened 2017-01-11
copied gdk-pixbuf recipe from ostrokach

ThomasWollmann:
* [x] I have read the guidelines above.
* [x] This PR adds a new recipe.
* [ ] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

PR #3436, opened 2017-01-11
added recipe for openslide

ThomasWollmann:


ThomasWollmann:
Depends on https://github.com/bioconda/bioconda-recipes/pull/3435

ThomasWollmann:
Any ideas for tests?

bgruening:
Soory only on my mobile please look at tests that are using 'test ... .so'

ThomasWollmann:
"If you want to run the test suite, you will need PyYAML, python-requests,
xdelta3, cjpeg and djpeg (from libjpeg), a Git checkout of OpenSlide,
at least one installed font, and > 120 GB of disk space. " ... do we have 120GB disk space?

bgruening:
No we have not. It is also not necessary to run functional tests, we should only test the installation.
So usually if you have a binary, a simepl `my_program --help` is enough. For libraries you can use `test foo/bar/barfoo.so` and so on.
Sometime we only check if expected files are in the correct location with `which` and so on. But then local testing is very much needed and we trust the package author.

ThomasWollmann:
Ok I'll give it a try. However stil waiting for dependencies.

bgruening:
gdk-pixbuf is merged, but this one is also in a bad state.

ThomasWollmann:
continued in https://github.com/bioconda/bioconda-recipes/pull/3737

PR #3437, opened 2017-01-11
GraphicsMagick avoid shared-lib build

mmiladi:


mmiladi:
According to installation guideline, using shared-lib has many challenges. So the best is to avoid --enable-shared.
http://www.graphicsmagick.org/INSTALL-unix.html#optional-packages-options

       `if GraphicsMagick is built using shared libraries, then all delegate libraries must also be built as shared libraries.`

mmiladi:
@bgruening Could you please have a look an merge it

PR #3439, opened 2017-01-11
added recipe for itk

ThomasWollmann:


ThomasWollmann:
Failed cause of too long compile log. What is the standard way of solving this?

ThomasWollmann:
Maybe we can remove some libraries from the superbuild.

ThomasWollmann:
Still "The log length has exceeded the limit of 4 MB"

ThomasWollmann:
Seems like the build is using two versions of libgcc

bgruening:
@bioconda/core any idea?

johanneskoester:
I think this is the true error: https://travis-ci.org/bioconda/bioconda-recipes/jobs/191600859#L9862. Looks like it needs something from glibc 3.4*. However, we build on CentOS 6 with glibc 2.14*. If that is the case, I think the only solution would be a patch?

bgruening:
@ThomasWollmann can you please this PR up, so I work on it if I have time?
Something goes wrong when you rebase? How do you do this?

ThomasWollmann:
I'm using rebase in sourcetree

ThomasWollmann:
This time I merged instead of rebase to avoid a malicious rebase

ThomasWollmann:
Seems like the docker test is not evaluating "requires" of the test section correctly.

bgruening:
Can you create a simple command?

ThomasWollmann:
What kind of command do you prefer?

bgruening:
I enhanced our testing a little bit to allow the PREFIX variable in test commands.

ThomasWollmann:
Thanks a lot @bgruening !

bgruening:
Conda and BioContainer just arrived! https://quay.io/repository/biocontainers/itk?tab=tags

PR #3441, opened 2017-01-11
Kaiju - added wget as run dependency

rasmusagren:

### DIFF
--- a/recipes/kaiju/meta.yaml
+++ b/recipes/kaiju/meta.yaml
@@ -8,7 +8,7 @@ source:
   md5: dfead4e9e0477fb619f9174e74c384ad
 
 build:
-  number: 1
+  number: 2
 
 requirements:
   build:
@@ -16,6 +16,7 @@ requirements:
   run:
     - libgcc
     - perl-threaded
+    - gnu-wget >=1.16
 
 test:
   commands:


rasmusagren:
@bgruening Sorry for spamming you about this recipe, but I can't quite get it to work. It builds ok, but then generates the following error when you try to run it:

```
/pica/v10/b2013006/sw/apps/miniconda3/pkgs/kaiju-1.4.4-1/bin/convertNR: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.18' not found (required by /pica/v10/b2013006/sw/apps/miniconda3/pkgs/kaiju-1.4.4-1/bin/convertNR)
/pica/v10/b2013006/sw/apps/miniconda3/pkgs/kaiju-1.4.4-1/bin/convertNR: /usr/lib64/libstdc++.so.6: version `CXXABI_1.3.5' not found (required by /pica/v10/b2013006/sw/apps/miniconda3/pkgs/kaiju-1.4.4-1/bin/convertNR)
```

I guess this has to do with that Conda uses an older gcc version than what is required, but how could it then compile on the image used in travis? I've seen several threads about related issues, but I'm a bit lost when it comes to compiler issues so I thought it would be faster to ask. The tool is written in C++11 and the makefile looks like this https://github.com/bioinformatics-centre/kaiju/blob/master/src/makefile. Any help would be much appreciated!

bgruening:
Mh, usually this kind of error indicates that there is no libgcc installed. Can you make sure this is installed?

rasmusagren:
It's a run dependency in the recipe, that should be enough, right?

bgruening:
Yes.

bgruening:
What I don't understand is that the test in our Docker container is passing:
https://travis-ci.org/bioconda/bioconda-recipes/jobs/190955537#L495

This indicates that it should work, can you check the Docker container? My guess is that this has something to do with your installation.

rasmusagren:
That test works for me as well. The error is a bit tricky to test for in the recipe since it happens after the program has downloaded a 40 GB database.

rasmusagren:
I've found the problem. I should probably have mentioned this before, but I was using the conda directive in Snakemake when I got the error and it seems like it's using the wrong package versions. There is probably nothing wrong with this recipe. Thanks for your help!

bgruening:
ok, glad our very detailed tests are working, thanks!

rasmusagren:
Here is the explanation for the issue, just if someone has the same problem. https://bitbucket.org/snakemake/snakemake/issues/465/conda-and-symbolic-links

PR #3442, opened 2017-01-11
updated kmergenie recipe

rchikhi:


rchikhi:
i'll fix problems found by Travis if they occur..

rchikhi:
triggering travis rebuild

jerowe:
@rchikhi , instead of overwriting the existing recipe, could you create a new recipe for your version?

kmergenie/
   - old_version/
        meta.yaml
        build.sh
   - meta.yaml #these are the ones you update
   - build.sh  #these are the ones you update

Also, good idea on the test data ;-)

rchikhi:
well i thought it'd be appropriate to PR this version because as it stands right now, it actually doesn't work for me and other users. (kmergenie installs correctly but then doesn't run, because it looks for subprograms that are not in the right folder)

jerowe:
The github recipes are really a placeholder for the packages in the cloud. If we get rid of the recipe on github, it does not get rid of the recipe on the cloud. 

So we should either fix the current recipe or delete it (from the cloud). 

If you have a fix for the current version, please raise a PR and increase the build #.

rchikhi:
I'm not sure if I got your recommendation right, although my current strategy is to indeed fix the current recipe. By raising a PR, do you mean doing something else than fixing the current PR we're in?

rchikhi:
..travis isn't executing automatically on commits anymore?

johanneskoester:

PR #3447, opened 2017-01-11
Add recipe for falcon

bebatut:


recipes/falcon/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - pypeflow
+    - networkx
+  run:
+    - python
+    - setuptools
```
bgruening: setuptools is probably not needed

recipes/falcon/meta.yaml
```diff
@@ -0,0 +1,36 @@
+{% set name = "falcon" %}
```
druvus: It is a name collision with another falcon software in conda-forge.

bgruening: Can we call it `pacbio_falcon`?

bebatut: I changed thatn

Issue #2272, opened 2016-09-03
error in installing meme


onkarnath89:
PASS: ama1
PASS: ama2
PASS: ame1
PASS: centrimo1
PASS: create-priors1
PASS: create-priors2
**FAIL:** dreme1 (CRASH!)
**FAIL:** dreme2 (CRASH!)
PASS: fasta-center1
**FAIL:** fasta-dinucleotide-shuffle1 (CRASH!)
PASS: fimo1
PASS: fimo2
PASS: fimo3
PASS: fimo4
**FAIL:** glam2_1 (CRASH!)
FAIL: glam2scan1 (CRASH!)
PASS: gomo1
PASS: gomo2
PASS: mast1
PASS: mast2
PASS: mast3
PASS: mast4
PASS: mast5
PASS: mast6
PASS: mast7
PASS: mast8
PASS: mast9
PASS: mast10
PASS: mast11
PASS: mast12
PASS: mast13
PASS: mast14
PASS: mast15
PASS: mast16
PASS: mast17
PASS: mast18
PASS: mcast1
PASS: meme1
PASS: meme2
PASS: meme3
PASS: meme4
PASS: meme5
PASS: meme6
PASS: meme_psp
PASS: motif-in.meme.crp0.text
PASS: motif-in.meme.crp0.xml
PASS: motif-in.meme.crp0.html
PASS: motif-in.meme.lipocalin.text
PASS: motif-in.meme.lipocalin.xml
PASS: motif-in.meme.lipocalin.html
PASS: motif-in.meme.puf3p.text
PASS: motif-in.meme.puf3p.xml
PASS: motif-in.meme.puf3p.html
PASS: motif-in.meme.crp0.extdna.text
PASS: motif-in.meme.crp0.extdna.xml
PASS: motif-in.meme.crp0.extdna.html
SKIP: motif-in.dreme.klf1.text (dreme.txt missing)
SKIP: motif-in.dreme.klf1.xml (dreme.xml missing)
SKIP: motif-in.dreme.klf1.html (dreme.html missing)
SKIP: motif-in.dreme.klf1.extdna.text (dreme.txt missing)
SKIP: motif-in.dreme.klf1.extdna.xml (dreme.xml missing)
SKIP: motif-in.dreme.klf1.extdna.html (dreme.html missing)
SKIP: motif-in.dreme.puf3p.text (dreme.txt missing)
SKIP: motif-in.dreme.puf3p.xml (dreme.xml missing)
[...]

Issue #1840, opened 2016-06-27
DREME tool from MEME suite uses Python 2 print statement in Python 3 environment


jma1991:
The DREME tool from the MEME suite uses Python 2 print statements. This means when you try and run it in a Python 3 environment with Anaconda you get an error.

```
> conda install meme
> dreme
File "/home/jashmore/anaconda3/bin/dreme", line 759
    print "Finding secondary RE in left flank..."
```

rvalieris:
meme doesn't support python3, the new update now only builds on python2.
to make sure you are using python2 use conda environment: `conda create -n py2 python=2 meme`

rfarouni:
@rvalieris  But on the website, it says that Python3 is supported. 

http://meme-suite.org/doc/install.html?man_type=web#prereq_python

Are there any other Python 2 dependencies that prevent building the recipe for Python 3?

qicaibiology:
> hi @rfarouni, yeah looks like they added support recently, it wasn't there on the version I updated.
> 
> I can't look into this right now but feel free to work on it.

Hi @rvalieris I still have the same problem in MEME 5.0.5. my MEME works but when I type dreme it shows:
```
File "/home/caiqi/anaconda3/bin/dreme", line 765
    print "Finding secondary RE in left flank..."
                                                ^
SyntaxError: Missing parentheses in call to 'print'. Did you mean print("Finding secondary RE in left flank...")?
```

I work on anaconda3

PR #3449, opened 2017-01-11
update meme 4.11.2

rvalieris:
also fixed the build to python 2 only,
this fixes issues #2272 and #1840

as stated in the docs: http://meme-suite.org/doc/install.html, meme doesn't support python3 so I'm changing it to build on python 2 only.

unfortunately, I could not integrate the tests meme ships with because of this bug: https://github.com/conda/conda-build/issues/1453, the tests need to be executed from the build dir, however when build the shebangs strings exceed the linux maximum length and that makes the tests fail.

as a local workaround, the tests can be run if conda-build is executed with the `--prefix-length 1` option.

### DIFF
--- a/recipes/meme/build.sh
+++ b/recipes/meme/build.sh
@@ -1,15 +1,21 @@
 #!/bin/bash
-#
+set -e
 MEME_ETC_DIR=${PREFIX}/etc
-cpanm YAML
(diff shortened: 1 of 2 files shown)


bgruening:
LGTM @rvalieris!

PR #3454, opened 2017-01-12
Novoplasty

mictadlo:


mictadlo:
@bioconda/core Does this [error](https://travis-ci.org/bioconda/bioconda-recipes/jobs/191164571#L576) happens because of an empty line in the beginning of `novoplasty.pl`?

johanneskoester:
Looks like there is a shebang line missing in that perl script. Hence, it is interpreted as bash. I guess this should be fixed upstream.

@bgruening am I right that perl recipes should no longer depend on perl-threaded but perl now?

bgruening:
@johanneskoester yes. In practice we have a nice migration in place but best-practice is that we should only use `perl` from now on.

mictadlo:
Any `sed` command which works for Linux and Mac to remove the empty line on top of the file?

johanneskoester:
`tail -n+2`

mictadlo:
Thank you, but why Linux has failed?

mictadlo:
@bioconda/all I am not able to fix [it](https://travis-ci.org/bioconda/bioconda-recipes/jobs/192228592#L593). Any idea what did I miss?

ndierckx:
Hi, 
I am the author of NOVOPlasty, do you want me to do something to fix it?

bgruening:
@ndierckx I guess it would help if you could fix the shebang to `!#/usr/bin/env perl` in all your scripts.

ndierckx:
so I should change ` #!/usr/bin/perl -w` by `!#/usr/bin/env perl` ?

mictadlo:
* I think `!#/usr/bin/env perl` would be better. 
* Could you please also remove the empty line on at beginning of the file?

jerowe:
It might also be helpful to be more specific with the perl executable.

If you use perl-module-build and a [ Build.PL](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/metavelvet-sl), your script will be installed with the 'proper' perl.

In your build.sh add something like:

```
mkdir -p perl-build/lib
cp ${RECIPE_DIR}/Build.PL perl-build
find -name "*.pl" | xargs -I {} cp {} perl-build
find -name "*.pm" | xargs -I {} cp {} perl-build/lib
cd perl-build
perl ./Build.PL
./Build manifest
./Build install --installdirs site

cd ..
```

ndierckx:
I am not an informatician, so I have limited knowledge about the subject, but I don't use any build module. It is just one perl script without any modules, so I run it just like perl NOVOPlasty.pl
You recommend I change this?

I can change the shebang line to `!#/usr/bin/env perl`, but without the first blank line, I had users who couldn't run the script (no idea why the line was necessary), so I don't know if I should remove it.

jerowe:
> I am not an informatician

@ndierckx , you wrote a genome assembler. ;-)

I don't know why you would need a blank line. If you would like to remove it try this: `tail -n +2 "$FILE" > "$FILE.tmp" && mv "$FILE.tmp" "$FILE"`

The Build module is not strictly necessary, but it is best practice. This is because conda environments are build to be standalone. A user will install your assembler into a particular environment. It is better to point to THAT perl, rather than to an env perl. This way if they have multiple environments loaded, your script will always work. I hope that's clear, if not please feel free to ask. Again, its not required, just a good idea.

ndierckx:
Thanks for the suggestions, I will look into it!
I removed the blank line and changed the shebang line, maybe it will work now?

bgruening:
Thanks all!

PR #3467, opened 2017-01-13
Add new infernal version

bgruening:


mmiladi:
@bgruening 
The reason is the Perl path is hardcoded:
```
$ head -2 infernal-1.1.2/easel/devkit/sqc 
#! /usr/bin/perl
```
We need a workaround. Would that be to link `/usr/bin/perl` to the available perl binary?

PR #3494, opened 2017-01-16
Add MethylDackel-0.2.0, which is the next version of PileOMeth

dpryan79:

@bioconda/core: This only sort of adds a new recipe. I've finally gotten around to renaming PileOMeth, which I should have done long ago. So this is just the next release of that recipe. Let me know if there's some other preferred way to handle package renaming.


johanneskoester:
That's fine. People still might want to access the old versions under the old name.

bgruening:
Dackel? You can speck German ... ;)

PR #3501, opened 2017-01-17
Dexseq

dyusuf:


recipes/bioconductor-dexseq/meta.yaml
```diff
@@ -49,6 +49,7 @@ requirements:
     - r-rcolorbrewer
     - r-statmod
     - r-stringr
+    - python 2*
```
bgruening: can you just use `python` here and skip the build for [py3k or osx]. This will solve your issue.

dyusuf: thanks

PR #3503, opened 2017-01-17
Add netcdf Fortran library

annefou:
* [Y ] I have read the guidelines above.
* [ Y] This PR adds a new recipe.

### DIFF
--- a/recipes/netcdf-fortran/meta.yaml
+++ b/recipes/netcdf-fortran/meta.yaml
@@ -14,8 +14,10 @@ build:
 requirements:
     build:
         - gcc
+        - libgfortran
         - libnetcdf 
     run:
+        - libgfortran
         - libnetcdf 
 
 test:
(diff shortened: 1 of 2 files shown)


recipes/netcdf-fortran/meta.yaml
```diff
@@ -0,0 +1,26 @@
+package:
+    name: netcdf-fortran
+    version: 4.4.4
+
+source:
+    git_url: https://github.com/Unidata/netcdf-fortran
```
bgruening: Can you please use tarballs and not git urls.

recipes/netcdf-fortran/meta.yaml
```diff
@@ -0,0 +1,26 @@
...
+    git_tag: 4.4.4
+
+build:
+    number: 1
```
bgruening: this should be started with 0

PR #3506, opened 2017-01-17
update xsv to 0.10.3

rvalieris:

not sure if this will compile as the previous version of the recipe didn't had rust as a dependency.


rvalieris:
error: `/opt/rh/devtoolset-2/root/usr/bin/ld: cannot find Scrt1.o: No such file or directory`

seems like the gcc installed by conda can't find the Scr1.o file, worst is that I can't find any option on cargo/rust to override this.

PR #3516, opened 2017-01-18
Feature/perl recipes

jerowe:

Adding in some base level perl, and adding back in the 2.0.8 recipe of kat.


recipes/kat/2.0.8/build.sh
```diff
@@ -0,0 +1,42 @@
...
+
+# fix autoconf
+sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autom4te
+sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoheader
+sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoreconf
+sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/ifnames
+sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoscan
+sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/autoupdate
+
+<<<<<<< HEAD
```
bgruening: here is something wired, but the tests pass?

jerowe:
Good catch. I only noticed this because I got a conflict on my merge from upstream. I thought I cleaned it up, but obviously not.

recipes/kat/2.0.8/build.sh
```diff
@@ -0,0 +1,33 @@
...
+
+export BOOST_INCLUDE_DIR=${PREFIX}/include
+export BOOST_LIBRARY_DIR=${PREFIX}/lib
+export LIBS='-lboost_system -lboost_program_options -lboost_filesystem -lboost_timer'
+
+export CXXFLAGS="-DUSE_BOOST -I${BOOST_INCLUDE_DIR} -L${BOOST_LIBRARY_DIR}"
+export LDFLAGS="-L${BOOST_LIBRARY_DIR} -lboost_filesystem -lboost_system"
+
+# fix automake
+sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/aclocal
```
bgruening: @jerowe I think this is fixed upstream and we do not need this fix anymore. At least we should test it at some point :)

jerowe: @bgruening , you were too quick for me! I was just getting ready to test this. ;-) In any case it shouldn't hurt, right?

bgruening: it doesn't hurt and we have several packages with this hack. If it is not needed we can clean them up all at once.

PR #3531, opened 2017-01-19
update viral-ngs 1.13.4 -> 1.14.0

tomkinsc:


tomkinsc:
Looks like this build is encountering a "no space left on device" [error](https://travis-ci.org/bioconda/bioconda-recipes/jobs/193437854#L986). Do we have a way to bypass quay builds yet?

tomkinsc:
@bgruening: Do you have any suggestions for working around the "no space" issue? This is the package for our own software pipeline, with dependencies, so I can distribute in a separate channel if necessary (but I'd prefer to keep it in bioconda).

bgruening:
Mh ... no real solution. I can build it for Linux and push it for you. We skip the build here completely and merge it without building, just adding a large note that this package needs to be build by the core team.
But, I can only build on Linux for you :(

tomkinsc:
Hmm...that's far from ideal. I'd love to move to an automated process to push new builds to bioconda, but a manual build process is a definite block there unless I can have a (secondary) API key for pushes. At last check it _does_ build successfully for OSX. I'll note that this was not an issue before the move to rely on the quay/mulled/conda-forge container system.

This is an issue for metapackages too, right? Otherwise I could repackage dependencies into separate packages and include them all in the top level.

I'm not sure how to move forward here short of abandoning bioconda... :(

bgruening:
The problem is more that your package is too big. There is, and there will be, some limitation on the build-system no matter what you do. Travis has a build-time limit of 45min and some space limit of several GB. We could get the build-time limit up to 2h, but we still have the space limitations for the foreseeable future.

Your last package already used the mulled stuff and it works, it's just that you package is not getting bigger and bigger. I wrote travis and hope we can get more space for a build but no promises.

druvus:
viral-ngs is removed from bioconda so closing.

PR #3549, opened 2017-01-22
patch bowtie2 to work across filesystems

daler:

This PR patches the `bowtie2` Perl wrapper script which under some conditions failed with the following error: 

```
/anaconda/pkgs/bowtie2-2.3.0-py27_0/bin/bowtie2-align-s:
error while loading shared libraries:libtbbmalloc_proxy.so.2: cannot open shared object file: No such file or directory.
```

Specifically this was happening on a cluster where conda was installed on one filesystem and the environment containing bowtie2 was on another filesystem. The patch removes the part of the wrapper that attempts to resolve symlinks. 

This does not happen in previous versions of bowtie2, but according to the release notes there was substantial refactoring in 2.3.0 which could be why we're only seeing this now.

I think that removing the symlink resolution here is OK, since the bowtie2 perl wrapper installed in a conda environment is unlikely to be symlinked elsewhere. This seems to fix the problem and tests pass, but @bioconda/core does anyone foresee issues with this?

Also @bgruening would there be any way to test this sort of cross-filesystem issue from within mulled-build?

Thanks @wresch and @jfear for working this out.

### DIFF
--- a/recipes/bowtie2/bowtie2.patch
+++ b/recipes/bowtie2/bowtie2.patch
@@ -1,0 +1,17 @@
+--- bowtie2.orig        2017-01-20 19:41:25.706765000 -0500
++++ bowtie2     2017-01-20 16:23:38.574188000 -0500
+@@ -38,10 +38,10 @@
+  my ($vol,$script_path,$prog);
+  $prog = File::Spec->rel2abs( __FILE__ );
+
+-while (-f $prog && -l $prog){
+-    my (undef, $dir, undef) = File::Spec->splitpath($prog);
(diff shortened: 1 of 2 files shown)


bgruening:
@daler testing via NFS could be complicated.
Maybe we could test this in the build container with single docker mounts? Not sure. Keep in mind that the runtime container at the end, only contains the environment and no conda ... so we can not use the runtime container - in which the tests currently run.

+1 for this PR.

daler:
@bgruening you're right, I had forgotten about conda not even being installed in the busybox containers. I think if other packages hit this issue we can just handle it on a case-by-case basis.

PR #3563, opened 2017-01-23
Add Bowtie2-2.2.4 recipe

amcooksey:


upendrak:
@bgruening do you know what went wrong with this build? @amcooksey is my colleague here at CyVerse. Thanks..

bgruening:
Mh, the error seems to be: https://travis-ci.org/bioconda/bioconda-recipes/jobs/194634090#L669

bgruening:
And this one: https://travis-ci.org/bioconda/bioconda-recipes/jobs/194634090#L660

bgruening:
And all this is due to that error: https://travis-ci.org/bioconda/bioconda-recipes/jobs/194634090#L323

bgruening:
Thanks @amcooksey and @upendrak!

bgruening:
Docker Image here: https://quay.io/repository/biocontainers/bowtie2?tab=tags

upendrak:
@bgruening Thanks a lot. So the build number should be "0" for linux and skip=True for OSX? I just want to understand it properly.

bgruening:
If you have a new recipe or a new version of a recipe, we always start with build-number 0. Also if you omit the build-number it stats with 0.
I just skiped the OSX package because the recipe was broken for OSX - which I think is due to some recent travis changes. It is on my todo list, but I thought you are only interested in the containers so this is fine ...

upendrak:
Ok for now i will include skip=True for OSX so that Travis won't throw an error. I am teaching @amcooksey as well understanding the whole process better. Thanks for fixing the issue..

bgruening:
Oh that's not want I wanted to say! :)
It would be great if you can also build on OSX and find ways to build it there, but if there is no way in getting it run on OSX - you are allowed to skip it.

upendrak:
We tried to build the OSX but for some reason, both OSX and Linux urls have the same md5sum and so the build is throwing an error and so we have skipped OSX and it worked great. Next time, we will keep this in mind

PR #3565, opened 2017-01-24
SWEEP new recipe

mictadlo:


mictadlo:
@bioconda/all What does the following [error](https://travis-ci.org/bioconda/bioconda-recipes/jobs/194707400#L662) mean?:

```
Packages/dependencies missing in current linux-64 channels: 
  - sweep 20160812 py27pl5.22.0_0 -> samtools ==0.1.9
  - samtools ==0.1.9
```

druvus:
Your meta-file is asking for the exact version 0.1.9 of samtools that is not available in bioconda. Sweep is calling samtools/bcftools once using a specific command line so you could test using available 0.1.* versions and  then set the requirement to samtools <1.0 .

bgruening:
And maybe this is just a typo and you want to have 0.1.19?

recipes/sweep/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - perl-app-cpanminus
+    - perl-module-build
+
+  run:
+    - python
+    - pysam
+    - biopython
+    - perl
+    - perl-app-cpanminus
+    - perl-module-build
```
druvus: `perl-app-cpanminus` and `perl-module-build` are not needed as run requirements.

recipes/sweep/build.sh
```diff
@@ -0,0 +1,17 @@
...
+cd perl-build
+cp ${RECIPE_DIR}/Build.PL ./
+perl ./Build.PL
+perl ./Build manifest
+perl ./Build install --installdirs site
+
+cd ..
+## End build perl
+
+find . -name "*.py" | xargs -I {} mv {} perl-build
```
druvus: This would not work python files are copied into the temporary perl-build folder with no further action.

PR #3566, opened 2017-01-24
Add Cluster picker

JasperYH:


recipes/cluster-picker/meta.yaml
```diff
@@ -0,0 +1,26 @@
...
+package:
+  name: cluster-picker
+  version: 1.2.3
+
+build:
+  number: 0
+
+source:
+  fn:  ClusterPicker_1.2.3.jar
+  url: http://hiv.bio.ed.ac.uk/clusterPicker/ClusterPicker_1.2.3.jar
```
bgruening: can you add a checksum?

recipes/cluster-picker/run_test.sh
```diff
@@ -0,0 +1,7 @@
+#! /bin/bash
+
+echo -e ">seq1\nATTA\n>seq2\nATTA\n>seq3\nATTA\n>seq4\nATTA" > dummy.fasta
```
bgruening: You could add these 3 lines also to the commands-test section into meta.yaml. Cleanup is not needed.

recipes/cluster-picker/cluster-picker.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
bgruening: You might want to look at: https://bioconda.github.io/guidelines.html#java

There is a reference to a similar script in python that is maybe more readable and maintainable.

JasperYH:
Thanks, I'll update in another pull request shortly.

PR #3569, opened 2017-01-24
Recipe fsnviz

bow:

Hi developers,

This is a PR for the [fsnviz](https://pypi.python.org/pypi/FsnViz) package, for visualizing RNA-seq fusion detection results. It's a Python 3.5-only package that depends on Circos. Please let me know if anything needs to be changed.

Thanks in advance!


recipes/fsnviz/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+$PYTHON setup.py install 
+
+# Add more build steps here, if they are necessary.
+
+# See
+# http://docs.continuum.io/conda/build.html
+# for a list of environment variables that are set during the build process.
```
daler: Can you remove the comments from here? (https://bioconda.github.io/guidelines.html#comments-in-recipes)

bow: Ah, sorry, I thought I had that removed :/. Done in 20e4df72.

PR #3571, opened 2017-01-24
fastq-multx: updates version in place

brwnj:


bgruening:
A new version should always start with 0 as build number.
Thanks @brwnj

brwnj:
Thanks for updating!

druvus:
The issue seems to be that the binary is shipped with the latest release. You can fix it by adding `rm fastq-multx` before make in build.sh. Or get it fixed upstream in a new release.

PR #3574, opened 2017-01-24
Adding a dependency needed for python-irodsclient

crislawrence:

@bioconda/core - Please review....my first attempt at any conda recipe.  It is a dependency needed for python-irodsclient.  Thought I'd try this first.

### DIFF
--- a/recipes/prettytable/meta.yaml
+++ b/recipes/prettytable/meta.yaml
@@ -1,0 +1,23 @@
+package:
+  name: prettytable
+  version: "7"
+
+source:
+  fn: prettytable-0.7.tar.bz2
+  url: https://pypi.python.org/packages/c7/21/8911c1b4aab57b891a54bcb6b1fab2ef59e80a78f7f699dd3d162ddda87f/prettytable-0.7.tar.bz2
+  md5: 8425f158caaa32d35fee6617f3fcb80b
+
+requirements:
+  build:
+    - python
+    - setuptools
+
+  run:
+    - python
+
+about:
+  home: http://code.google.com/p/prettytable
+  license: BSD License
+  summary: 'A simple Python library for easily displaying tabular data in a visually appealing ASCII table format'
+  license_family: BSD
+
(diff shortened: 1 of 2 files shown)


recipes/prettytable/meta.yaml
```diff
@@ -0,0 +1,23 @@
+package:
+  name: prettytable
+  version: "7"
```
bgruening: This should be 0.7, right?

crislawrence: Hopefully 0.7.2.  Python-irodsclient won't work with lesser versions.

crislawrence: Travis-CI has been going for 1:45 min.  The first job finished in just under 10 min.  I really don't know what travis-ci is doing now.  It said that job 2.2 hadn't even started.  So I am at a loss to know what's happening.

bgruening:
Cris, good new and bad news. This package is already in conda-forge ... and bioconda can use the packages from conda-forge. So you can directly submit your irods one and depend on prettytable.
I will close this now, ok?

recipes/prettytable/meta.yaml
crislawrence: Actually, I guess the version is too old for me to use.  Let me see if I can create a recipe using the proper version.

bgruening: Wait, I send you a mail :)
Everything is fine here :) Just submit your second PR with the irods-client.

crislawrence: OK....I will use conda-forge and tomorrow, I'll do a PR for python-irodsclient

bgruening: To explain this a little bit. Conda has so called channels - like PPA on Ubuntu. You can mix channels and enable multiple ones. But Conda has a concept of channel priority so you need to enable the channels in a correct order. Every bioconda package has also conda-forge enable. So you can mix bioconda-packages with conda-forge ones.

PR #3588, opened 2017-01-25
Oops: Forgot matplotlib

zfrenchee:
@bgruening

### DIFF
--- a/recipes/manormfast/meta.yaml
+++ b/recipes/manormfast/meta.yaml
@@ -17,6 +17,7 @@ requirements:
     - numpy
     - scipy
     - statsmodels
+    - matplotlib
 
 test:
   imports:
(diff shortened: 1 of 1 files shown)


bgruening:
You need to bump the build number.

zfrenchee:
@bgruening done. Unfortunately, the pypi package number is still 0.1.2... that might cause some confusion...

recipes/manormfast/meta.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: manormfast
-  version: "0.1.2"
+  version: "0.1.3"
```
bgruening: Thats not the build number ;)

zfrenchee: ?!?! Which one is that?

zfrenchee: Note: I didn't forget matplotlib in the pypi package, just in this file. I don't know what other number there is to change.

PR #3594, opened 2017-01-26
udpate license info

osallou:

### DIFF
--- a/recipes/pash/meta.yaml
+++ b/recipes/pash/meta.yaml
@@ -23,7 +23,7 @@ requirements:
 
 about:
   home: http://www.bioinformatics.bbsrc.ac.uk/projects/bismark/
-  license: Unknown
+  license: GPL-3+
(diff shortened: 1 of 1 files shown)


recipes/pash/meta.yaml
```diff
@@ -23,7 +23,7 @@ requirements:
 
 about:
   home: http://www.bioinformatics.bbsrc.ac.uk/projects/bismark/
```
abretaud: It's the url which is wrong, it should be: https://genboree.org/pash-supp/
=> there, the licence is not specified

osallou: should contact author, if license is not specified, it is not sure we ware allowed to redistribute software....

abretaud: Yes, *we* should (you? me? hehe)

abretaud: I just sent an email to the author

johanneskoester:

PR #3601, opened 2017-01-26
working out ucsc update

daler:

Turns out there were a few changes in UCSC v343, so this PR is just working out what needs to be done in order to update everything. 

@bioconda/core: even though `libpng` is listed as a dependency and the `include.patch` adds `-I ${PREFIX}/include`, it seems that `png.h` is not able to be found and the build fails. Any ideas?

(example error: https://travis-ci.org/daler/bioconda-recipes/jobs/195573621#L786)


johanneskoester:

PR #3613, opened 2017-01-27
Add stellar

eseiler:


eseiler:
@bioconda/core

bgruening:
@eseiler can you please try to compile this tool from source.
The reason is that we want to make sure these binaries are relocatable even to very old environments that are common for HPC.

Issue #3224, opened 2016-12-16
Using more recent compilers


holtgrewe:
I have not found a solution to this but it might have occured earlier.

I want to compile Flexbar which needs seqan-library which needs a relatively recent compiler, e.g., GCC >= 4.9.1 or some Clang compiler.

How can I solve this issue?

bgruening:
@holtgrewe sequan-library is already available so you could simply depend on this, no need to compile it on your own - or do you a newer GCC also for flexbar?

holtgrewe:
@bgruening seqan-library 2.2 (which flexbar requires) is header-only and uses features only available in C++14. Thus, for compiling flexbar I need a modern compiler. Fun times!

I guess I would have to add a build requirement on GCC 4.9.1 then and this would have to go into the global anaconda channel or conda-forge?

bgruening:
This is a huge can of works. But yes you would need to get a newer GCC version in, have a look at this: https://github.com/conda/conda-recipes/pull/502 and similar PRs

holtgrewe:
@bgruening I think I don't need the bleeding edge version there but rather just copy one of the other recipes (and dependencies).

`anaconda search -t gcc-5` shows packages with name prefixes (e.g., `salford_systems/gcc-5`). Would I add gcc-5 and dependencies from this user's packages to bioconda-recipes then?

bgruening:
I don't think that bioconda wants to maintain GCC. It should be added to conda-forge or actually to anaconda probably as the PR I linked above.

We are discussion the compiler situation since one year, not sure if anything happend on this front recently.

holtgrewe:
OK, what I don't understand is why there is a gcc-5 in conda-recipes but I don't see it in conda install. Do you know?

bgruening:
No. conda-recipes are not as transparent as they should be. This is one reason conda-forge will take the entire low-level stack and maintain this in a transparent, community driven way.

holtgrewe:
OK, so you would welcome a PR with `gcc-5` with a copy from whatever is in conda-recipes into conda-forge? Or is there more stuff blocking this?

bgruening:
Imho, but I can be totally wrong, think you have a higher chance to get this in in conda-forge.

daler:
[...]

PR #3614, opened 2017-01-27
Add Sak

eseiler:


eseiler:
@bioconda/core

bgruening:
@eseiler usually we compile all tools and not using precompiled binaries. Can you try that seqan is already available as dependency.

eseiler:
@bgruening The current gcc versions don't support compiling the source (need at least >=4.9).
Also see
https://github.com/conda-forge/staged-recipes/pull/2089
and
https://github.com/bioconda/bioconda-recipes/issues/3224

bgruening:
Hi Enrico,

fair enough then. This would be the correct PR to work further on this issue: https://github.com/conda-forge/staged-recipes/pull/873

recipes/sak/build.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+BINARY_HOME=$PREFIX/bin
+PKG_HOME=$PREFIX/opt/$PKG_NAME-$PKG_VERSION
```
bgruening: Is there any reason you put it under opt and not simply copy the binaries into PREFIX/bin?

bgruening:
Thanks @eseiler!

PR #3617, opened 2017-01-27
Rphylip

bsennblad:
@bioconda/core
I have added recipes for the Rphylip R-package (r-rphylip) and, as this require the phylip programs to be installed, I also added a recipe for phylip. 
A biobuilds recipe already existed for phylip, but as biobuilds is not enquired by the bioconda recipes (according to the guidelines), I reimplemented biobuilds recipe with minor adaptations for bioconda. I have two main questions regarding the phylip recipe:
1) The  biobuilds recipe give the license as BSD; however, this is not explicit in the original phylip website. Instead, an explicit copyright is given (copied in the file LICENSE). (I first included 'license_file: LICENSE' in meta.yaml under 'about', but this appears not to be supprted?) 
Maybe this statement conforms to a BSD license? Please advice!
2) The biobuilds recipe defines wrappers for the phylip programs inside the build.sh file. Is it considered better to include these wrappers as separate files in the recipe-folder?
Thanks,
Bengt


recipes/phylip/LICENSE
```diff
@@ -0,0 +1,34 @@
+# <From the original phylip website>
+#
+# PHYLIP, the Phylogeny Inference Package, is a package of programs
```
bgruening: This needs to be in the unpacked tarball to be recognised for `license_file` afaik.

recipes/r-rphylip/meta.yaml
```diff
@@ -0,0 +1,67 @@
+{% set name = 'Rphylip' %}
+{% set version = '0.1-23' %}
+
+{% set posix = 'm2-' if win else '' %}
```
bgruening: everything referring to win can be excluded, we do not build for the Windows platform.

recipes/r-rphylip/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+build:
+  number: 0
+
+  rpaths:
+    - lib/R/lib/
+    - lib/
+
+requirements:
+  build:
+    - r-base >=3.3.1
```
bgruening: This pinning can be removed.

recipes/r-rphylip/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - r-ape >=3.0_10
+
+  run:
+    - r-base  >=3.3.1
```
bgruening: Same here, this r-base pinning can be removed.

recipes/phylip/tests/Makefile
```diff
@@ -0,0 +1,45 @@
+# Makefile, tebsts directory and samples by Catherine Letondal <letondal@pasteur.fr>
+
```
bgruening: We do not include test files in bioconda. Functional tests should be done upstream. Could you simply add something similar to `phylib --help` or --version?

recipes/phylip/build.sh
```diff
@@ -0,0 +1,95 @@
...
+SHARE_DIR="${PREFIX}/share/${PKG_NAME}-$PKG_VERSION-$PKG_BUILDNUM"
+for d in fonts java exe; do
+    install -d "${SHARE_DIR}/${d}"
+    install ${d}/* "${SHARE_DIR}/${d}"
+done
+pushd "${SHARE_DIR}/java"
+rm -f *.unx
+popd
+
+# Generate wrapper scripts in ${PREFIX}/bin like the Debian package does
```
bgruening: @bsennblad we have a python based wrapper script to encapsulate this. Look here for some Guidelines: https://bioconda.github.io/guidelines.html#java

recipes/phylip/LICENSE
bsennblad: OK. I remove the LICENSE file

recipes/phylip/tests/Makefile
bsennblad: I have now built the testing into the wrapper script.

recipes/phylip/build.sh
bsennblad: Added python-based wrapper scripts and removed bash variants.

recipes/r-rphylip/meta.yaml
bsennblad: Removed win thingies

recipes/r-rphylip/meta.yaml
bsennblad: Pnning removed. Originally the recipe didn't build unless this was in place, but now it works all right. I suspect that I managed to do an update in beteen.

recipes/phylip/build.sh
bsennblad: @bgruening 
I had alter the java-wrappers slightly from that suggested in the guideline (i.e., the peptide-shaker one) to get it to work with phylip. Here are the relevant lines from the code from the end of the wrappers drawgram.py and drawtree.py:
    # Trying to fix font problems phylip appears to require that fonts
    # are placed in the working directory Current workaround is to
    # temporarily link them there and then remove links. Far from
    # optimal solution, but I can't get anything else to work:(
    subprocess.call("ln -s {fd}/* .".format(fd=re.sub("java","fonts",jar_dir)), shell=True)

    # execute jar-file, record signal
    sig=subprocess.call(java_args)
    
    # remove links
    subprocess.call(["rm -f font[1-6] fontfile"], shell=True)

    # return signal from jar-file
    sys.exit(sig)

Otherwise, all seems to build properly now, so I'll probably merge soon.

Thanks for your input!

PR #3620, opened 2017-01-27
Try building on OS X against XQuartz.

acaprez:

### DIFF
--- a/recipes/r-mixomics/meta.yaml
+++ b/recipes/r-mixomics/meta.yaml
@@ -15,7 +15,6 @@ source:
 
 build:
   number: 0
-  skip: True  # [osx]
 
   # This is required to make R link correctly on Linux.
   rpaths:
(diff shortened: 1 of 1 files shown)


bgruening:
I'm not sure if this is relevant, but there is now also the entire X stack in conda and could be used.

acaprez:
Ah good to know, thanks.  The `r-rgl` package is what needs all the graphics libs.  The X11 packages in conda-forge cover some of the required libraries, but it still needs others like `libGL`, `libGLU`, `freetype`, etc.  XQuartz has everything; my guess would be that's what `r-rgl` was originally built against.  Building in my test VM with XQuartz installed worked fine, so that's why I thought I'd try it here.

acaprez:
ping @bioconda/core 

This builds ok using XQuartz.  Using the package would require having it installed outside of Conda then.  IMO that's reasonable since there are quite a few other scientific packages that need it as well, but wanted to see what others thought before merging.

johanneskoester:
I'm fine with that, although I'd like to reduce outside dependencies in the long run if the packages are available in conda-forge. If you do this for now, please add a note to `extra -> notes` and a `TODO` comment to the recipe that explains why libglu was disabled on osx.

PR #3625, opened 2017-01-27
Add recipe for plant_tribes_phylogenomics_analysis v 0.3

gregvonkuster:

### DIFF
--- a/recipes/plant_tribes_phylogenomics_analysis/0.3/build.sh
+++ b/recipes/plant_tribes_phylogenomics_analysis/0.3/build.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+mkdir -p $PREFIX/bin
+mkdir -p $PREFIX/config
+mkdir -p $PREFIX/data
+mkdir -p $PREFIX/pipelines
+mkdir -p $PREFIX/scripts
+
+cp ./pipelines/PhylogenomicsAnalysis $PREFIX/bin
+ln -s $PREFIX/bin/PhylogenomicsAnalysis $PREFIX/pipelines/PhylogenomicsAnalysis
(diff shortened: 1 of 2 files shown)


gregvonkuster:
@bgruening is this just a travis timeout error or is there something wrong with the recipe?  It's sometimes difficult to see exactly what the problem is in the error log.  Sorry for bugging you on this.

bgruening:
I don't see any error either. Can it be that the test is just not working?

https://travis-ci.org/bioconda/bioconda-recipes/jobs/195976067#L593
Can you remove the grep and let's see what the output is?

gregvonkuster:
@bgruening Thanks, somehow the test worked within my local conda environment.  I've tried several things, and eliminating the stderr redirect always causes the test to fail.  I've added some command line parameters that are required and the tests pass locally.  So lets see if this works with travis.  ;)

bgruening:
I'm running out of ideas, it seems that the command does not produce any output and the grep is not successful.

gregvonkuster:
Any idea what may be causing this?
```
ERROR bioconda_utils.build:build(129): STDOUT+STDERR: bash: cannot set terminal process group (-1): Inappropriate ioctl for device
bash: no job control in this shell
```

bgruening:
This is from the Docker container, it's just a warning you can see this in every PR ...
Usually you see an output in the logs, if you call a tool for testing like this: `samtools --help`

I'm not sure why you don't see anything.

bgruening:
Here we go! https://travis-ci.org/bioconda/bioconda-recipes/jobs/196154407#L597

gregvonkuster:
Hmmm....well, at least I now know what needs fixing.  Thanks @bgruening I'll open a separate PR when this is fixed on the backend.

PR #3626, opened 2017-01-27
r-data-table: fix a path length problem

lecorguille:
Try rebuilding to fix a path length problem:
```
PaddingError: Placeholder of length '80' too short in package bioconda::r-data.table-1.10.0-r3.3.1_0.
```

### DIFF
--- a/recipes/r-data.table/meta.yaml
+++ b/recipes/r-data.table/meta.yaml
@@ -13,7 +13,7 @@ source:
   md5: f0e08dd5ba1b3f46c59dd1574fe497c1
 
 build:
-
+  number: 1
   rpaths:
     - lib/R/lib/
     - lib/


bgruening:
@daler @johanneskoester it seems something in conda changes and the previous warnings are now errors. I see more and more PR of this sort and it seems to work. What I'm wondering is if we should relax the pinning to 3.2.* and use peoples rebuild already?

johanneskoester:
@bgruening you mean the pinning of `CONDA_R`? Wouldn't this cause an inconsistent collection of R packages? New PRs against 3.3.2, and all the rest against 3.3.1?

bgruening:
Yes `CONDA_R`. If I pin against `3.3.*`, conda can choose during install if I get `3.3.1` or `3.3.2` isn't it?

johanneskoester:
That sounds dangerous. There must be a reason that the R channel pins against the minor/patch version. Let me ask @asmeurer who recently offered to help with some advice regarding the R porting effort. @asmeurer, can you enlighten us on the decision to pin against the patch version in the R channel?

bgruening:
According to this comment I would assume it works: https://github.com/ContinuumIO/anaconda-issues/issues/1239#issuecomment-268478295

Similar to python 2.7.* releases.

johanneskoester:
I'm not sure R strictly follows semantic versioning. In the release notes of 3.3.2, there is at least one change that breaks the API of a function: https://cran.r-project.org/doc/manuals/r-release/NEWS.pdf

bgruening:
API should be fine, as long as ABI keeps stable, no?

asmeurer:
Because I had no idea how R handled ABI/API breaks at patch changes (as far as I had heard there have been problems), and because it is so easy to rebuild everything with every R release that there's little downside to doing it.

johanneskoester:
Thanks @asmeurer. Yes it is indeed easy, but it requires a lot of time when relying on CI. Ok, in the [FAQ](https://cran.r-project.org/bin/windows/base/rw-FAQ.html#What_0027s-the-best-way-to-upgrade_003f), they say that you have to install the new version, copy over your packages, and run `update.packages(checkBuilt=TRUE)`. This function will do the following (see [here](https://stat.ethz.ch/R-manual/R-devel/library/utils/html/update.packages.html)):

* install all packages with newer versions
* rebuild packages where the major.minor version changed.

Hence, at least the recommended R update process does not rebuild packages if only the patch version changed.

Therefore, I'd say we could risk to just pin to major.minor.
If we choose to do so, let's do it together with conda-forge though.
@ocefpaf do you agree?

asmeurer:
When I was building on VMs on my laptop it only took a day to rebuild everything. 

I've had at least one R user tell me that things have broken in R patch versions. I really don't know myself either way because I'm not a user of R. Someone who is an expert would have to say. Given the way R is generally used, it wouldn't surprise me too much if it was more lax about breaking changes. 

At any rate, I'd say the recommendation for R environments in general should be to pin the exact versions used for an analysis (say, in an `environment.yml`).

johanneskoester:
Ok, so we need to decide whether the simplified patch version upgrade is worth the risk of breaking individual packages.

At least, with the upcoming new pinning mechanism in conda-build, it should be possible overwrite the relaxed strict pin for an for individual packages that break.

> At any rate, I'd say the recommendation for R environments in general should be to pin the exact versions used for an analysis (say, in an environment.yml).
Absolutely. At least for scientific use, this should be the recommendation in any case, also for non-R.

ocefpaf:
> Therefore, I'd say we could risk to just pin to major.minor.
If we choose to do so, let's do it together with conda-forge though.
@ocefpaf do you agree?

Not sure to be honest I not a big R user. I would stick to @asmeurer experience here and keep the pinning as is. Also to keep our packages with the same build string as `defaults`.

johanneskoester:
defaults does not contain R packages. Just the R channel which we aim to replace at some point. But I can still understand your argument. I am still unsure... Maybe we should really ask one of the R devs about this, because the statements in the docs (see above) pretty clearly show that it would be ok to just pin major.minor.

ocefpaf:
Change `defaults` to continuum :wink:

asmeurer:
I thought they recently added the `r` channel to the default channels.

johanneskoester:
Really? I wasn't aware of that...

asmeurer:
See https://www.continuum.io/blog/developer-blog/announcing-general-availability-conda-4-3 (second bullet from the bottom).

PR #3649, opened 2017-01-30
added knime recipe

ThomasWollmann:


ThomasWollmann:
"OSError: [Errno 28] No space left on device". Seems like KNIME is too large for bioconda?

bgruening:
Likely but more for Travis and the VM. Please also take care of following the guidelines for java packages.

ThomasWollmann:
What shall we do about the filesize?

ThomasWollmann:
I removed the plugins for now, but I think KNIME is more useful if we ship it with all free plugins.

bgruening:
Is this really needed as Conda package? I mean KNIME should use Conda as resolver, but does this needs to be a Conda package. Or the other way around who would install it this way, especially without plugin?
I don't think there is anything we can do at them moment about the size. Packages should be small - we also don't ship databases inside of packages.

ThomasWollmann:
In my opinion it would make it easier to integrate KNIME as a command line tool into Galaxy.

PR #3652, opened 2017-01-30
SAVAGE recipe

jbaaijens:


johanneskoester:
The error is here: https://travis-ci.org/bioconda/bioconda-recipes/jobs/196552816#L843

The problem seems to be that one of the libs that are compiled into ViralQuasispecies has been compiled on a newer system. The only reason I can imagine is that you ship something precompiled. Any idea what that could be?

acaprez:
I believe it's because `libgcc` is missing from the run requirements.  At runtime it uses the OS-supplied `libstdc++.so.6` instead of the one from `libgcc` and causes the error.

bgruening:
Thanks @jbaaijens!

PR #3658, opened 2017-01-30
Change perl-threaded to perl

nsoranzo:

### DIFF
--- a/recipes/perl-module-build/meta.yaml
+++ b/recipes/perl-module-build/meta.yaml
@@ -17,11 +17,11 @@ build:
 requirements:
   build:
     - gcc
-    - perl >=5.22.0
+    - perl
 
   run:
     - libgcc
-    - perl >=5.22.0
+    - perl
 
 test:
   # Perl 'use' tests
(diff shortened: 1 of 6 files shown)


recipes/perl-module-build/meta.yaml
```diff
@@ -11,17 +11,17 @@ source:
...
 build:
-  number: 2
+  number: 3
   skip: False
 
 requirements:
   build:
     - gcc
-    - perl-threaded >=5.22.0
+    - perl
```
nsoranzo: @daler I hope removing the minimum version is fine with you, the [Travis build](https://travis-ci.org/bioconda/bioconda-recipes/jobs/196708452#L736) fails without it.

daler: should be fine, but @bioconda/perl any issues with removing a min version?

johanneskoester: I think that is ok, because conda-build will pin to CONDA_PERL automatically.

PR #3661, opened 2017-01-30
perl-want a new module for Perl

mictadlo:


recipes/perl-want/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+requirements:
+  build:
+    - perl
+    - perl-module-build
+
+  run:
+    - perl
+
+# no tests so far
```
druvus: Please add a test.
```
test:
  imports:
    - Want
```

druvus:
Added in another PR.

PR #3673, opened 2017-02-01
Add r-splitstackshape recipe

upendrak:


recipes/r-splitstackshape/meta.yaml
```diff
@@ -0,0 +1,46 @@
+{% set name = 'splitstackshape' %}
+{% set version = '1.4.2' %}
+
+{% set posix = 'm2-' if win else '' %}
+{% set native = 'm2w64-' if win else '' %}
```
bgruening: In general, as we do not build for windows you can remove all these lines that are specifically for windows.

upendrak: Ok got it. Do you want me to modify the PR?

PR #3677, opened 2017-02-01
peptide-shaker add version 1.15.0

jj-umn:


druvus:
It seems that you have the 1.15.0 version both in the sub and main folder. The idea is to create a subfolder for the old recipe, ie 1.14.6 in this case.

jj-umn:
@druvus Within galaxy, I need to reference  particular versions of searchgui and peptide-shaker.  I'd have to find out how galaxy builds an env without a version.  @bgruening Can you enlighten me?

bgruening:
@jj-umn what @druvus means it just the ordering of the recipes I think. The directory structure should be
```
1.14.6
1.x.y
meta.yml (latest)
build.sh (latest)
```

recipes/peptide-shaker/1.15.0/meta.yaml
```diff
@@ -0,0 +1,43 @@
+{% set name = "PeptideShaker" %}
+{% set version = "1.15.0" %}
```
bgruening: You probably wanted to move `1.14.6` into the `1.14.6` directory, isn't it?

PR #3681, opened 2017-02-02
deepmedic recipe

ThomasWollmann:

### DIFF
--- a/recipes/deepmedic/meta.yaml
+++ b/recipes/deepmedic/meta.yaml
@@ -15,8 +15,8 @@ build:
 requirements:
   build:
     - theano
-    - nibabel
     - nose
+    - nibabel
     - pp
     - six
     - scipy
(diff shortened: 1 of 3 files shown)


recipes/deepmedic/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - theano
+    - nose
+    - nibabel
+    - pp
+    - six
+    - scipy
+    - numpy
+  run:
+    - theano
+    - nose
```
bgruening: is nose really needed as run dependency

ThomasWollmann: probably not...

druvus:
Seems you were misisng the pp dependency. I have added it #3690 so I will restart the test once it is available.

ThomasWollmann:
Thanks @druvus!

bgruening:
mh, https://travis-ci.org/bioconda/bioconda-recipes/jobs/197942786#L746 Seems that a specific version of pp is needed. :(
You could either patch setup.py or create an older pp package.

druvus:
@ThomasWollmann you need to patch setup.py to get rid of `dependency_links` part.

ThomasWollmann:
@druvus is it enough to just remove it?

druvus:
I think so

druvus:
Thanks @ThomasWollmann

PR #3683, opened 2017-02-02
added omero python recipe

ThomasWollmann:


recipes/python-omero/build.sh
```diff
@@ -0,0 +1,4 @@
+cd lib/python
+
+CONDA_PYTHON=$(conda info --root)/bin/python
```
bgruening: Why is this so complicated?
Does a simple `${RECIPE_DIR}/setup.py install` not work?

recipes/python-omero/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  sha1: f94b71ab30d6239e6a4ba2d8b52eca9f262b800a
+  
+build:
+    number: 0
+    skip: True  # [py3k or osx]
+
+requirements:
+  build:
+    - python
+    - conda-build
```
bgruening: This is always available, isn't it?

ThomasWollmann:
This package is still very VIP.

druvus:
Seems that you have a SHA1 mismatch.

bgruening:
I think you also need python as build dependency.

PR #3687, opened 2017-02-03
Do not skip khmer build on macOS

marcelm:

It seems the khmer recipe was meant to work under macOS (there is an `llvm` requirement), but the `skip: True` was never removed.


marcelm:
@bgruening Did you have a specific reason for skipping the macOS build of khmer or was that just an oversight? I need the macOS version as a dependency of another recipe.

marcelm:
I’ll just merge it for now. It seems to be fine.

Issue #3638, opened 2017-01-28
integrate cargo-port URLs in the bioconductor skeleton builder


bgruening:
We have currently 2000 tarballs backuped on cargo-port. So I think it is a reliable service. We do had a few downtimes but I think it is overall working.

What do you think about adding the cargo-port URL by default to the meta.yaml as second or third link?
A link typically looks like this: https://depot.galaxyproject.org/software/bioconductor-affyio/bioconductor-affyio_1.42.0_src_all.tar.gz

I hope that rebuilding r-packages will be much more easier with this change in place.
ping @bioconda/r

johanneskoester:
Yes, let's do that!

druvus:
Good idea!

daler:
#3713 adds this. If I remember, bioaRchive tarball md5sums differed from bioconductor tarball md5sums. So if the url the recipe is built from one url and the package is built from another, it'll fail from md5 mismatch. But cargo-port mirrors the top url directly, right? So should we remove bioaRchive urls to avoid such md5 issues?

PR #3713, opened 2017-02-05
add cargoport url to bioconductor recipes

daler:

xref #3638 

Newly-created or updated bioconductor recipes will now have up to 3 URLs -- bioconductor, bioaRchive, and galaxy cargo-port.

Still need to add ability to retrofit older packages with cargo-port urls and migrate this to bioconda-utils. But it's working for new recipes.


bgruening:
@daler thanks for working on it! What do you think about adding this URL no matter what as second or third URL. As soon as we merge this package we will create this tarball on cargo-port and it will be available if we need to rebuild this package.

daler:
Good point, it should always be included.

johanneskoester:
Is this ready for merging?

bgruening:
@daler are you planning to work on this further? I have started the R rebuilds and this would come in handy I guess.

daler:
Sorry, this dropped off the radar. If I remember this was almost ready to go, will look at it tomorrow and merge in recent changes in master.

daler:
@bgruening @johanneskoester this should be a much-improved script. It probably belongs in bioconda-utils though, right?

Anyway, the newest changes are:

- cargoport url is added no matter what, as discussed above
- urls now explicitly include the bioconductor version. This should be WAY more stable.
- can specify a package version and bioconductor version at the command line. I'm now scraping the bioconductor release page to get a list of versions, and added some autodetection which tries older and older bioc versions until the package version is identified. For example the following now works nicely, identifying the most recent bioc release for `tximport==1.0.3` as 3.3 and adjusts the url accordingly: `scripts/bioconductor/bioconductor_skeleton.py tximport --force --pkg-version 1.0.3`

- an option for building the recipe in a versioned subdirectory for easer archiving of older recipes
- copied over the nice colored log output Johannes had added to bioconda-utils

bgruening:
Awesome! Thanks a lot! This will make migrations so much easier in a long run!

PR #3742, opened 2017-02-08
Add perl-config-simple

bogemad:


recipes/perl-config-simple/bld.bat
```diff
@@ -0,0 +1,29 @@
+:: If it has Build.PL use that, otherwise use Makefile.PL
```
bgruening: Can you please remove this file? It is not needed as we do not build for Win.

PR #3757, opened 2017-02-09
update rsem to 1.3.0

rvalieris:

a few extra changes:

* removed `rsem_perl_utils.pm` from the list of binaries, this is a module and it was already being installed in the perl lib path.
* added bioconductor-ebseq to the runtime dependencies, since the Makefile installs EBSeq.
* removed the Makefile patch, recent changes to the Makefile made the patch invalid, in the end I preferred to allow the makefile to compile its own version of htslib already included in the rsem source, this won't generate lib conflicts because the lib is being statically compiled into the rsem binaries.
* removed samtools,htslib build-time dependencies, since its already included in the rsem source.
* rsem makefile already includes the `--without-curses` flag as well


bgruening:
Nice cleanup! Thanks!

PR #3758, opened 2017-02-09
Maker a new recipe

mictadlo:

### DIFF
--- a/recipes/maker/build.sh
+++ b/recipes/maker/build.sh
@@ -8,8 +8,8 @@ cd src
 if [ -f Build.PL ]; then
     (echo yes; yes '') | perl Build.PL
     perl -i.bak -wpe 's[^#!.+][#!/usr/bin/env perl]' Build
-    perl ./Build
-    perl ./Build test
+    #perl ./Build
+    #perl ./Build test
     # Make sure this goes in site
     perl ./Build install --installdirs site
 elif [ -f Makefile.PL ]; then


mictadlo:
@bgruening @chapmanb @martin-mann

I have got [`Bio::Root::Version is not installed`](https://travis-ci.org/bioconda/bioconda-recipes/jobs/200134117#L1158), but I used `perl-bioperl`(https://github.com/bioconda/bioconda-recipes/pull/3758) in my `meta.yaml`.

Is `Bio::Root::Version` part of `perl-bioperl`?

Thank you in advance

Michal

chapmanb:
Michal;
It looks like the latest `perl-bioperl` package (perl-bioperl-1.6.924-6) is empty which likely causes the problem here. perl-bioperl-1.6.924-4 appears okay. I haven't kept up with all the latest perl changes to this package, but a quick fix might be to remove the problematic recent versions. Björn and Martin, what do you think?

mictadlo:
@chapmanb I have got

```
Package missing in current linux-64 channels: 
  - perl-bioperl ==1.6.924-4
```

[here](https://travis-ci.org/bioconda/bioconda-recipes/jobs/201338436#L592).

What did I miss?

chapmanb:
Michal -- the syntax for this is a little strange, I think you want:
```
- perl-bioperl ==1.6.924=4
```
with an `=` instead of the last `-`. Hope this helps pin it to a working version. I'd still like to visit removing the broken packages as well if Björn and Martin have permissions to do so.

mictadlo:
I have got `conda.exceptions.CondaValueError: Value error: Malformed version string 'perl-bioperl-1.6.924=4': invalid character(s).`

mictadlo:
With `perl-bioperl-1.6.924` I got `Bio::Root::Version is not installed`. How could we fix it?

chapmanb:
Michal;
You want to have `- perl-bioperl ==1.6.924=4` in both the build and run requirements. I don't believe the `==perl-bioperl-1.6.924` syntax is valid but conda must skip it until you stuck in the extra `=4`. Hope that fixes it for you.

mictadlo:
I used `- perl-bioperl ==1.6.924=4`, but I got the this [error](https://travis-ci.org/bioconda/bioconda-recipes/jobs/202066750#L552) `Value error: Malformed version string '1.6.924=4': invalid character(s).`

chapmanb:
Michal -- gah sorry about all the retries. I'm not sure what is going on. @bgruening could we remove the empty broken -5 and -6 versions of perl-bioperl to try and sort out this problem? I'm not sure why pinning to sub-versions isn't working but getting rid of those seems to be the best approach at this point.

rvalieris:
I believe this should work for pinning the version:
 ```
requirements:
  build:
   - perl-bioperl 1.6.924 4
```

**Comment by mictadlo** [comment]
Thank you. Perl problems have been solved. Unfortunately, there it failed to build, but I do not why. Any ideas?

**Comment by rvalieris** [comment]
add a line like this:
`perl -i.bak -wpe 's[^#!.+][#!/usr/bin/env perl]' Build`
 right after the `perl Build.PL` line.

but even with that I get this error later:
```
cc: error: unrecognized command line option ‘-fstack-protector-strong’
make: *** [MPI.o] Error 1
```

have you tried building this package locally ? use this command from the bioconda-recipes directory:
`CONDA_PERL="5.22.0" conda build recipes/maker`

that way you can see the errors faster than waiting for travis to build.

mictadlo:
Thank you. @rvalieris I got `OSError: [Errno 28] No space left on device` on travis ci. I am using MacOS and I didn't get the docker container running.

nsoranzo:
@rvalieris I have the same problem with `-fstack-protector-strong` in https://github.com/bioconda/bioconda-recipes/pull/3839#issuecomment-281176326 .

mictadlo:
still not luck. Any ideas?

mictadlo:
@jerowe @druvus  by any chance, do you know how to fix it?

jerowe:
@mictadlo, all things perl are pretty broken right now.

You could try to do something hacky where instead of using the conda gcc you are using the conda-forge gcc, but that is all I can think of. I'm not even sure where conda-forge is keeping its gcc, since I can't just run a `which gcc`.

Its unfortunate because it means perl has very minimal support in all machines that don't have gcc >4.9, which is probably most centos6 machines.

mictadlo:
Why does conda-forge not switch to centos7?

mictadlo:
Anything new on gcc >4.9 front?

bgruening:
Nothing new, but there is hope ;)
An other question regarding maker, the license is not ideal, have you checked if you are allowed to distribute the package?

mictadlo:
I thought it would be fine because of https://raw.githubusercontent.com/Homebrew/homebrew-science/master/maker.rb

cbrueffer:
maker 2.31.9 was added in f2a7aef3c74bbd066f945aaeb81a75db1c608368, I think this PR can be closed?

PR #3761, opened 2017-02-10
Increase savage build number in order to retrigger generation of the …

johanneskoester:
…corresponding docs.

The savage package (PR #3652) by @jbaaijens has been built and uploaded successfully, but the corresponding entry on bioconda.github.io has never been generated. I assume there was a temporal bug in bioconda-utils. @daler would this make sense, or do you think there is a different reason?

### DIFF
--- a/recipes/savage/meta.yaml
+++ b/recipes/savage/meta.yaml
@@ -5,7 +5,7 @@ package:
   version: {{ version }}
 
 build:
-  number: 0
+  number: 1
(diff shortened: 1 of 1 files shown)


bgruening:
@johanneskoester It seems to be this bug: https://travis-ci.org/bioconda/bioconda-utils/jobs/200062121#L904

johanneskoester:
I see. I guess it is not yet fixed?

bgruening:
I don't think so :(

johanneskoester:
Ok, I'll have a look.

johanneskoester:
Should be fixed with PR #3779.

PR #3767, opened 2017-02-10
Added in srst2 recipes for 0.1.4.6 and 0.2.0

ericenns:

### DIFF
--- a/recipes/srst2/0.1.4.6/meta.yaml
+++ b/recipes/srst2/0.1.4.6/meta.yaml
@@ -1,0 +1,38 @@
+package:
+  name: srst2
+  version: 0.1.4.6
+
+source:
+  fn: v0.1.6.tar.gz
+  url: https://github.com/katholt/srst2/archive/v0.1.6.tar.gz
+  sha256: 24cbd03933d4cc14bbfa7f8a073c9b0caafb13d55afb3bfa25f8b093fc5bc615
+  patches:
+    - use_setuptools.patch
+
+build:
+  number: 0
+  skip: True #[not py27]
+
+requirements:
+  build:
+    - python
+    - setuptools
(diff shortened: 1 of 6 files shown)


recipes/srst2/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - setuptools
+    - numpy >=1.7.1
+    - scipy >=0.12.0
+
+  run:
+    - python
+    - numpy >=1.7.1
+    - scipy >=0.12.0
+    - bowtie2 >=2.1.0
+    - samtools ==0.1.18
```
bgruening: do you need setuptools in your run section?

recipes/srst2/0.1.4.6/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - setuptools
+    - numpy >=1.7.1
+    - scipy >=0.12.0
+
+  run:
+    - python
+    - numpy >=1.7.1
+    - scipy >=0.12.0
+    - bowtie2 >=2.1.0
+    - samtools ==0.1.18
```
bgruening: Maybe this can be removed?

recipes/srst2/meta.yaml
ericenns: setuptools is just needed to be able to install the python package, from looking at other recipes it only appears to be included in the build section.

recipes/srst2/0.1.4.6/meta.yaml
ericenns: Not sure what your askin. If it is in regards to the 0.1.4.6 recipe, I do need it as we have a galaxy tool for 0.1.4.6 that we would like to continue using until we have time to validate 0.2 for our purposes.

PR #3779, opened 2017-02-13
[ci skip] Better error reporting when generating docs. Fix notes section of openms.

johanneskoester:

Since some time, the docs were not properly updated any more. This was because of an undetected error in the openms recipe (the notes section contained a list). This PR improves the error messages of our doc script (they now point to the failing recipe) and fixes the problem with openms.


bgruening:
Great! Thanks a lot!

PR #3792, opened 2017-02-14
Updated package bioconductor-genefilter version from 1.54.2 to 1.56.0

guillaumecharbonnier:

### DIFF
--- a/recipes/r-catools/meta.yaml
+++ b/recipes/r-catools/meta.yaml
@@ -21,7 +21,7 @@ source:
 build:
   # If this is a new build for the same version, increment the build
   # number. If you do not include this key, it defaults to 0.
-  # number: 1
+  number: 1
 
   # This is required to make R link correctly on Linux.
   rpaths:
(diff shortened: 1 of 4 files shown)


recipes/r-catools/meta.yaml
```diff
@@ -21,7 +21,7 @@ source:
 build:
   # If this is a new build for the same version, increment the build
   # number. If you do not include this key, it defaults to 0.
-  # number: 1
+  number: 1
```
bgruening: Any reason you need to rebuild these?

guillaumecharbonnier: Yes, I did that because the current build produces
"PaddingError: Placeholder of length '80' too s…"
Let me know if this is not the correct way to solve this issue as it is the first time I try solve it without updating the package version at the same time.

bgruening: Just wanted to double check!

PR #3806, opened 2017-02-15
patch krona script to fix symlink resolution under osx

tomkinsc:

The `-f` option has a different meaning under the OSX `readlink`, so the script will fail when `-f` is used on OSX. This patches a shell script to add a function to mimmic `readlink -f` functionality. It relies on perl, which is already a dependency of Krona. This patch is needed until the next version which should have a fix upstream. This PR also changes the `perl-threaded` dependency to `perl`.


odoublewen:
Hi @tomkinsc - 
This recipe seems to be broken for me, for a fresh `conda install krona` -- I've tried this on both Linux or Mac OS and get the same results.

The installation goes alright, but the ktUpdateTaxonomy.sh fails - it's looks for `taxonomy` in bin rather than opt.

I'm happy to dig in a see what can be done, but did you make this patch file?  If so maybe you would have a better solution?

Also I saw your note about a more permanent from upstream, and wondering what you know about KT developments...

Cheers!

```
$ ktUpdateTaxonomy.sh 
/Users/solberg/miniconda3/envs/krona/bin/ktUpdateTaxonomy.sh: line 186: cd: /Users/solberg/miniconda3/envs/krona/bin/taxonomy: No such file or directory

Update failed.
   Could not enter '/Users/solberg/miniconda3/envs/krona/bin/taxonomy'.
```

odoublewen:
I noticed that the updateTaxonomy.sh installed by conda was unpatched -- identical to the upstream distro copy of this file.

After I manually patched the file, everything works fine:
```
$ patch updateTaxonomy.sh ~/Downloads/updateTaxonomy.patch.txt 
patching file updateTaxonomy.sh
$ ktUpdateTaxonomy.sh
Fetching taxdump.tar.gz...
```

So... seems like I've determined that the patchfile is not being applied as part of the `conda install krona`   I'm not familiar with patchfiles in conda recipes -- any ideas?

Thanks

odoublewen:
More info:
I didn't notice that conda actually did not install this build.
It is falling back to build 0, apparently because per 5.22 is not available for this build.
```
$ conda search krona
krona                        2.5                  pl5.22.0_1  bioconda        
                             2.5                  pl5.22.0_2  bioconda        
                             2.5                           3  bioconda        
                             2.5                           4  bioconda        
                             2.6                  pl5.22.0_1  bioconda        
                             2.6                  pl5.22.0_2  bioconda        
                             2.6                           3  bioconda        
                             2.6                           4  bioconda        
                             2.6.1                         1  bioconda        
                             2.6.1                         2  bioconda        
                             2.7                           0  bioconda        
                             2.7                  pl5.22.0_1  bioconda        
```

If I try to request this build, I get:
```
$ conda install krona=2.7=pl5.22.0_1
Solving package specifications: 

PackageNotFoundError: Dependency missing in current osx-64 channels: 
  - krona 2.7 pl5.22.0_1 -> perl 5.22.0*

Close matches found; did you mean one of these?

    perl: perl-b, petl, pepr
```

odoublewen:
OK, this is all due to the fact that I did not have `conda-forge` in my channel list, so I couldn't get perl.  And then when I went to install krona, conda silently decided to install build 0 (which I did not notice for a while :/

Sorry for the chatter!  All well now.

PR #3810, opened 2017-02-16
Bump unicycler to 0.3.0

rrwick:

### DIFF
--- a/recipes/unicycler/meta.yaml
+++ b/recipes/unicycler/meta.yaml
@@ -18,11 +18,9 @@ requirements:
   build:
   - python
   - setuptools
-  - gcc 
 
   run:
   - python
-  - libgcc  
   - spades >=3.6.2
   - pilon
   - java-jdk


druvus:
@rrwick bioconda is using an older gcc version so I think you would need to patch the makefile to use `-std=c++1y`instead of `-std=c++14`.

druvus:
@bioconda/core any ideas of workaround how build the latest version of unicycler? Seems that seqan library doesn't support gcc  <4.9. This is a little bit outside my C knowledge.

bgruening:
@druvus I don't think there is a solution yet. We need to have a new compiler and there are various attempts in conda-forge, e.g. this one: https://github.com/conda-forge/staged-recipes/pull/873

I could imagine to put such a package in bioconda for a very limited time to enable such compilations, under a different namespace, maybe `gcc5`. But we should also help conda-forge to get this done sooner than later.

rrwick:
I've been poking around with this stuff today and came to the same conclusion. I can't make Unicycler build with anything less than GCC 4.9. 😦 

Until it can use GCC 5+, should Unicycler be removed from bioconda? I'd rather users not use bioconda to install Unicycler if that means they'll be stuck on an old version.

johanneskoester:

PR #3816, opened 2017-02-16
Update Metaphlan2 version

bebatut:


bebatut:
I removed the copy of the database files because `No space left` on travis test. Do you think we need to package the tool with the db?

druvus:
I think it would be better to skip it and instead provide a download script. The old conda file is approx 1Gb which is too big. I would vote that the 2.5.0 version would be rebuilt without the database.

recipes/metaphlan2/build.sh
```diff
@@ -8,4 +8,4 @@ utils/metaphlan2krona.py
 mkdir -p $PREFIX/bin
```
druvus: `strainphlan.py`should also be added as a binary for 2.6.0 recipe.

Maybe also some other of the utils scripts?

bebatut:
I will look at that tomorrow... 
Where do you think the data must be copy? By default to bin? We can customize that too...

druvus:
The default for metaphlan2 is the same directory as the script but you can set the db location as an argument. I think it would be better to have an option to download the database folder to a custom location based on users choice.

recipes/metaphlan2/2.5.0/download_metaphlan2_db.py
```diff
@@ -0,0 +1,66 @@
+#!/usr/bin/env python3
+
+import argparse
+import tarfile
+import os
+import urllib2
+import shutil
+
+
+METAPHLAN2_URL = 'https://bitbucket.org/biobakery/metaphlan2/get/2.6.0.tar.gz'
```
bgruening: 2.5.0?

recipes/metaphlan2/2.5.0/download_metaphlan2_db.py
```diff
@@ -0,0 +1,66 @@
...
+    Returns the name that the file is saved with.
+    """
+    print("Downloading %s" % url)
+    target = os.path.basename(url)
+    print("Saving to %s" % target)
+    open(target, 'wb').write(urllib2.urlopen(url).read())
+    return target
+
+
+def unpack_tar_archive(filen, wd=None):
```
bgruening: filen?

bgruening:
have you added a test that needs QT for plotting: https://travis-ci.org/bioconda/bioconda-recipes/jobs/203109810#L1209

If this matplotlib you can change the plotting backend ...

PR #3817, opened 2017-02-16
Pin to boost 1.61

johanneskoester:

The boost 1.60 package has some unfixed bugs. E.g., it apparently does not pin icu. This leads to problems when relying on functions of boost that require icu.
This PR is a try to update our boost pinning to the latest version in anaconda.


bgruening:
Great work @johanneskoester! How does this relate to the new pinning in conda-build 3? We are not waiting for this, or is this already available Oo :)

johanneskoester:
As far as I know it is not yet released although it is in master I think. This is still our old pinning. I just needed to do this know because I encountered a bug in the old boost version.

PR #3818, opened 2017-02-16
add tedna recipe

cmonjeau:


cmonjeau:
@abretaud

recipes/tedna/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+  number: 0
+
+source:
+  fn: tedna_1.2.2.tar.gz
+  url: https://urgi.versailles.inra.fr/content/download/3481/29402/file/tedna_1.2.2.tar.gz
+  md5: b60b9b974e349db692eea75aad15b5e9
+
+requirements:
+  build:
+  - google-sparsehash
```
abretaud: Dear @cmonjeau, can you add `bc` in the build dep list?
(it will solve the `/bin/sh: bc: command not found error`, maybe not the following one laule)

abretaud: oh, and the indentation is not ok here

recipes/tedna/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
```
abretaud: Add this line here:
```sed -i.bak 's|CFLAGS = |CFLAGS = -I'$PREFIX'/include/ |' makefile```

File: recipes/tedna/meta.yaml (hunk 1 above)
Comment (abretaud) [inline recipes/tedna/meta.yaml]: and add gcc

File: recipes/tedna/meta.yaml [hunk 3]
```diff
@@ -0,0 +1,27 @@
...
+
+  run:
+  - google-sparsehash
```
abretaud: and here libgcc

recipes/tedna/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+source:
+  fn: tedna_1.2.2.tar.gz
+  url: https://urgi.versailles.inra.fr/content/download/3481/29402/file/tedna_1.2.2.tar.gz
+  md5: b60b9b974e349db692eea75aad15b5e9
+
+requirements:
+  build:
+    - google-sparsehash
+    - gcc
+    - libgcc
```
bgruening: libgcc is here not needed. Also if you build for OSX please use llvm. More information are in our Guidelines. Thanks!

abretaud:
Compilation on osx fails with llvm, and we have no macosx to debug
Should we switch back to gcc on osx or disable osx build?

bgruening:
Switch back to GCC. Maybe @bioconda/osx can help as well.

abretaud:
Green!

PR #3822, opened 2017-02-16
Added abs_path patch to repeatmodeler

lparsons:


lparsons:
I'm not sure if this is something that has changed since old versions of Perl, but [`abs_path` is not exported by default](http://perldoc.perl.org/Cwd.html#abs_path-and-friends), so one has to request is explicitly. Once you request and explicitly export, no defaults are exported, thus the `getcwd`). I submitted feedback upstream, but I'm not optimistic it will be updated given the age of this software.

I've gone ahead and made the change to to the perl dependency.

rvalieris:
okay ! that makes sense, thanks for the explanation.

bgruening:
Thanks Lance!

PR #3825, opened 2017-02-17
moca v0.4

saketkc:

### DIFF
--- a/recipes/moca/meta.yaml
+++ b/recipes/moca/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: moca
-  version: "0.4.0"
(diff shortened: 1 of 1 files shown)


recipes/moca/meta.yaml
```diff
@@ -27,10 +28,10 @@ requirements:
     - cryptography >=1.7.1
     - cycler >=0.10.0
     - enum34 >=1.1.6
-    - functools32 >=3.2.3.post2
+    - functools32 >=3.2.3.post2 # [py27]
```
druvus: This is not needed when you are only building for py27

PR #3837, opened 2017-02-17
Add recipe for Du Novo tool

NickSto:

This adds a recipe for the tool [Du Novo](https://github.com/galaxyproject/dunovo).

It's my first recipe, so I'd appreciate comments, @bioconda/core. Not sure if I did everything right.

Travis-CI shows problems building on OS X, but I don't have an OS X machine to test on. I excluded `osx` in my `meta.yaml`, but I don't know if that's enough. 

Du Novo is a relatively simple tool based on Python and C. The main scripts are written in Python, and use the C code by compiling it into `.so`'s and importing with `ctypes`.

### DIFF
--- a/recipes/dunovo/build.sh
+++ b/recipes/dunovo/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+make
--- a/recipes/dunovo/build.sh
+++ b/recipes/dunovo/build.sh
@@ -1,3 +1,8 @@
 #!/bin/bash
 
 make
+mkdir $PREFIX/include/dunovo
+cp *.py *.so $PREFIX/include/dunovo
+ln -s ../include/dunovo/correct.py $PREFIX/bin
+ln -s ../include/dunovo/align_families.py $PREFIX/bin
+ln -s ../include/dunovo/dunovo.py $PREFIX/bin
--- a/recipes/dunovo/build.sh
+++ b/recipes/dunovo/build.sh
@@ -1,8 +1,5 @@
 #!/bin/bash
 
 make
-mkdir $PREFIX/include/dunovo
-cp *.py *.so $PREFIX/include/dunovo
-ln -s ../include/dunovo/correct.py $PREFIX/bin
-ln -s ../include/dunovo/align_families.py $PREFIX/bin
-ln -s ../include/dunovo/dunovo.py $PREFIX/bin
+mv *.so $PREFIX/lib
+mv *.py *.sh $PREFIX/bin
--- a/recipes/dunovo/meta.yaml
+++ b/recipes/dunovo/meta.yaml
@@ -19,10 +19,10 @@ requirements:
 
   run:
     - libgcc    # [not osx]
-    - python 2.7
+    - python >=2.7,<3
     - mafft 7.221
     - samtools 0.1.18
-    - bowtie2 2.0.2
+    - bowtie2 2.2.4
     - paste
     - gawk
 
--- a/recipes/dunovo/meta.yaml
+++ b/recipes/dunovo/meta.yaml
@@ -16,15 +16,24 @@ requirements:
     - make
     - gcc   # [not osx]
     - llvm  # [osx]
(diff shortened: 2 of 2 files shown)


recipes/dunovo/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+requirements:
+  skip: True  # [py3k or osx]
+  build:
+    - make
+    - gcc
+
+  run:
+    - libgcc
+    - python >=2.6,<3
+    - argparse  # [py26]
```
rvalieris: this means that argparse isn't a dependency if python version != 2.6, is this intended ?
keep in mind that since python3 was already skipped in the `skip:True` line, so you don't have to worry about it in the requirements, `python<3` is redundant.

recipes/dunovo/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+make
+mkdir $PREFIX/include/dunovo
+cp *.py *.so $PREFIX/include/dunovo
```
rvalieris: why `include/`, do you have this directory hardcoded somewhere ?
libraries like .so files should go in `lib/`, executable scripts can go directly at `bin/` or you can put them at `share/` or `lib/` and make links if there's a lot of them.

the `include` dir should contain only header files.

recipes/dunovo/meta.yaml
NickSto: Thanks for taking a look! I saw this trick in another package. Argparse was added to the standard library in 2.7, so if Python is 2.7, no dependency needs to be installed. For Python 2.6, it's not in the standard library, but it can be installed from PyPI, hence the dependency.

As for python<3, I just figured it's nice to be safe and explicit.

recipes/dunovo/build.sh
NickSto: So the Du Novo code expects the `.so` files to be in the same directory as the `.py` files that load them via `ctypes`. I wasn't sure where to stick this mix of files, so I followed what I saw another package do and put it in a special directory under `include`. I could go monkey with the Du Novo code to get it to check `lib` if it doesn't find them in its own directory.

recipes/dunovo/meta.yaml
rvalieris: cool, thats fine then.

recipes/dunovo/build.sh
rvalieris: it doesn't have to be include then ? if can you just change the directory to `$PREFIX/lib/denuvo/` and make the links to bin/ that will be fine.

NickSto: Hmm, okay, well that's good to hear. I'm not sure which path I'll take for now.

recipes/dunovo/meta.yaml
johanneskoester: When constraining python like that, conda will not pin to the Python version used for the build. This can cause various technical issues. Please remove the constraint.

NickSto: Oh, okay, I removed it. What exactly do you mean conda will not pin to the Python version used for the build?

recipes/dunovo/build.sh
NickSto: Okay, I changed it to put the `.so` files in `lib` and the executables in `bin`, and edited the Python scripts to find the `.so` files in `lib`.

NickSto:
Okay, wait, now Travis shows it failing on Linux, whereas it worked in the original version I PR'd (fd440d3). I'm not sure why it's failing or what change caused it.

bgruening:
Hi @NickSto! Sorry to be late here... I have fixed a few things. Maybe it works now.

NickSto:
Man, of course Bjorn swoops in and fixes everything. Don't know why I had the skip line in requirements, not build. Is that why it was failing?

bgruening:
@NickSto yes, it was the skip in the wrong section. It was not recognised and hence you still was building against python 3.

PR #3839, opened 2017-02-17
Change perl-threaded to perl

nsoranzo:
Remove unnecessary perl-extutils-makemaker build requirement.

### DIFF
--- a/recipes/perl-gd/build.sh
+++ b/recipes/perl-gd/build.sh
@@ -8,19 +8,18 @@ set -e -o pipefail
 # If it has Build.PL use that, otherwise use Makefile.PL
 if [ -f Build.PL ]; then
     perl Build.PL LIB=site \
-    	-options "JPEG,PNG,FT" \
-    	-lib_jpeg_path $PREFIX \
-    	-lib_ft_path $PREFIX \
-    	-lib_png_path $PREFIX \
-    	-lib_zlib_path $PREFIX \
-    	>&1;
-        
+        -options "JPEG,PNG,FT" \
+        -lib_jpeg_path $PREFIX \
+        -lib_ft_path $PREFIX \
+        -lib_png_path $PREFIX \
+        -lib_zlib_path $PREFIX
+
     sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' Build
-    ./Build 2>&1;
+    ./Build
     # disable non-portable test7
     sed -i.bak1 "s|IMAGE_TESTS => 7|IMAGE_TESTS => 6|1" ./t/GD.t
     sed -i.bak2 "s|tests => 11|tests => 10|1" ./t/GD.t
-    ./Build test 2>&1;
+    ./Build test
     # Make sure this goes in site
     ./Build install --installdirs site
 elif [ -f Makefile.PL ]; then
--- a/recipes/perl-gd/meta.yaml
+++ b/recipes/perl-gd/meta.yaml
@@ -7,22 +7,21 @@ source:
   md5: c4b3afd98b2c4ce3c2e1027d101a8f1e
 build:
   ##ignore_prefix_files: True
-  number: 5
+  number: 6
   skip: True  # [osx]
 
 requirements:
   build:
     - gcc # [linux]
     - llvm # [osx]
-    - perl-threaded >=5.22.0
     - libgd >=2.2.3
+    - perl
     - perl-module-build
-    - perl-extutils-makemaker
 
   run:
     - libgcc # [not osx]
-    - perl-threaded >=5.22.0
     - libgd >=2.2.3
+    - perl
 
 test:
   # Perl 'use' tests
--- a/recipes/perl-gd/meta.yaml
+++ b/recipes/perl-gd/meta.yaml
@@ -12,7 +12,10 @@ build:
 
 requirements:
   build:
-    - gcc # [linux]
+# Temporarily remove gcc build requirement because conda-forge Perl was compiled
+# with -fstack-protector-strong option and defaults gcc does not support it, see
+# https://github.com/bioconda/bioconda-recipes/pull/3839
+#    - gcc # [linux]
     - llvm # [osx]
     - libgd >=2.2.4
     - perl
--- a/recipes/perl-gd/meta.yaml
+++ b/recipes/perl-gd/meta.yaml
@@ -14,13 +14,13 @@ requirements:
   build:
     - gcc # [linux]
     - llvm # [osx]
-    - libgd >=2.2.3
+    - libgd >=2.2.4
     - perl
     - perl-module-build
 
   run:
     - libgcc # [not osx]
-    - libgd >=2.2.3
+    - libgd >=2.2.4
     - perl
 
 test:


nsoranzo:
@bgruening I get this error at https://travis-ci.org/bioconda/bioconda-recipes/jobs/202766860#L572 :
```
cc: error: unrecognized command line option ‘-fstack-protector-strong’
```
which I think is due to conda-forge `perl` having been compiled with gcc >= 4.9 . In fact, if I do:
```bash
(root) ~$ conda create -n perl_5_22_0 perl=5.22.0
Fetching package metadata .................
...
(root) ~$ source activate perl_5_22_0
(perl_5_22_0) ~$ perl --version | head -n 2

This is perl 5, version 22, subversion 0 (v5.22.0) built for x86_64-linux-thread-multi
(perl_5_22_0) ~$ perl -e 'use Config; print $Config{ccflags}'
-D_REENTRANT -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2(perl_5_22_0)
```

Please note that the `-fstack-protector-strong` option was introduced in gcc 4.9.0, so it is not supported by 4.8.5 (latest version available in `defaults` channel).

There's also another build failing for the same reason: https://github.com/bioconda/bioconda-recipes/pull/3758#issuecomment-280637466

bgruening:
Oh no .... this week is just getting worse and worse.

bgruening:
ping @bioconda/core the only reason I can see is that they updated the Docker image with a new compiler. Please note that conda-forge is not using the gcc conda package to compile packages, afaik.

johanneskoester:
Mhm, this is worrying. At some point there was a decision about using the continue gcc package for compilation. Now they use something else. If we can't even agree on a common compiler, how should we ensure compatible packages? We might end up reverting the conda-forge dependency again.

kyleabeauchamp:
I think it's fair for us to ask for advance notice on decisions like this.  IMHO bioconda, conda-forge, continuum, and others are all key stakeholders of the conda-forge build chain, so we should get some sort of deprecation notice for breakages.

johanneskoester:
So, question to c experts: does such a change mean we would need to rebuild every package that depends on a library build with a newer gcc?

mictadlo:
Could LLVM be update, too?

jerowe:
@johanneskoester , it depends. Some packages will be fine, some won't. It depends on the features that are being used during the compilation phase. I know there are some major changes >4.9. I remember having trouble getting older versions of perl compiled with gcc >5, but I think with perl 5.20.x we should be fine.

What is this that conda-forge isn't using a conda-package as a gcc compiler? That is very worrisome.

I have noticed some trouble with perl packages. Sometimes when building new packages, if a dependency is pulling in perl-threaded the perl5lib becomes lib/perl5/site_perl/5.22.0/ instead of lib/perl5/site_perl/5.20.0/. I've been working on a reasonable way to deal with this, hopefully that is at least conda-forge linter friendly. If there is a difference in compilers that will affect all the XS, Moose, and I think most of the Scalar-* packages as well.

I think we should start to consider pinning toolchains instead of individual variables. Thoughts?

Set A:
```
-pl5.20.3_4
-py2
-boost-1.60
-gcc-4.8
```

Set B:
```
-pl5.20.3_4
-py3
-boost-1.60
-gcc-4.8
```

Set C:
```
-pl5.20.3_4
-py2
-boost-1.61
-gcc-4.8
```

Set D:
```
-pl5.20.3_4
-py3
-boost-1.61
-gcc-4.8
```

Set E:
```
-pl5.20.3_4
-py3
-boost-1.61
-gcc-4.9 (or 5.x? )
```

etc, etc.

johanneskoester:
Yes, I think such pinning will be done semi-automatically by the new conda-build version. 
At least we should try to add gcc/libgcc via PRs as dependency to those packages in conda-forge that we use. This should increase their awareness of the issue. If I understood correctly, it was a pragmatic decision in their team that not everybody was happy with as well.

jerowe:
I just checked over a recent build I have over on conda-forge, and its at least pulling in the gcc compiler from the defaults channel. I suppose that doesn't actually mean that its using that compiler. It seems that its something that could be changed in the build.sh by explicitly specifying the path?

johanneskoester:
Yes. It depends a lot on the design of the makefile.

bgruening:
This indicates that the version in the Docker image is not > 4.9 isn't it? https://github.com/bioconda/bioconda-recipes/pull/3810

So from where do conda-forge get the newer version?

nsoranzo:
@bgruening The gcc in the Docker image is 4.8.2 20140120 (Red Hat 4.8.2-15), which contains a backport of `-fstack-protector-strong` since 4.8.0-6 (see https://rpmfind.net/linux/RPM/centos/7.3.1611/x86_64/Packages/gcc-4.8.5-11.el7.x86_64.html ).
This is using gcc-4.8.5-7 from `defaults` channel:
```
(root) ~$ conda create -n gcc_env gcc
Fetching package metadata ...............
Solving package specifications: ..........
...
(root) ~$ . ~/miniconda3/bin/activate gcc_env
(gcc_env) ~$ gcc --version
gcc (GCC) 4.8.5
Copyright (C) 2015 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

(gcc_env) ~$ echo 'int main(){}' > foo.c
(gcc_env) ~$ gcc foo.c
(gcc_env) ~$ gcc -fstack-protector-strong foo.c
gcc: error: unrecognized command line option ‘-fstack-protector-strong’
```
While this is the same, but run on the Docker image:
```
$ gcc --version
gcc (GCC) 4.8.2 20140120 (Red Hat 4.8.2-15)
Copyright (C) 2013 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ echo 'int main(){}' > foo.c
$ gcc foo.c
$ gcc -fstack-protector-strong foo.c
```

jerowe:
Ya. This is pretty bad.

```
BUILD START (revised): perl-module-metadata-1.000019-pl5.22.0_0
Package: perl-module-metadata-1.000019-pl5.22.0_0
source tree in: /anaconda/conda-bld/perl-module-metadata_1487763966071/work/Module-Metadata-1.000019
+ source /anaconda/bin/activate /anaconda/conda-bld/perl-module-metadata_1487763966071/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla
+ '[' -f Build.PL ']'
+ '[' -f Makefile.PL ']'
+ perl Makefile.PL INSTALLDIRS=site
perl: /lib64/libc.so.6: version `GLIBC_2.11' not found (required by perl)
Command failed: /bin/bash -x -e /anaconda/conda-bld/perl-module-metadata_1487763966071/work/Module-Metadata-1.000019/conda_build.sh
```

Building of packages that require gcc/libgcc in perl is going to fail, because perl itself is compiled with a newer version of gcc.

jerowe:
Perl 5.22.0
```
conda create -p perl-5.22 perl=5.22
Fetching package metadata .................
Solving package specifications: .

Package plan for installation in environment /root/perl-5.22:

    perl: 5.22.2.1-0 conda-forge

#
# To activate this environment, use:
# > source activate /root/perl-5.22
#
# To deactivate this environment, use:
# > source deactivate /root/perl-5.22
#

[root@8609b6952bd4 ~]# source activate /root/perl-5.22
(/root/perl-5.22) [root@8609b6952bd4 ~]# ldd `which perl`
/root/perl-5.22/bin/perl: /lib64/libc.so.6: version `GLIBC_2.11' not found (required by /root/perl-5.22/bin/perl)
```

Perl 5.20.3
```
[root@8609b6952bd4 ~]# conda create -p perl-5.20 perl=5.20                                                                                                                                                   
Fetching package metadata .................
Solving package specifications: .

Package plan for installation in environment /root/perl-5.20:

    perl: 5.20.3.1-2 conda-forge

[root@8609b6952bd4 ~]# ldd `which perl`
[root@8609b6952bd4 ~]# source activate /root/perl-5.20
(/root/perl-5.20) [root@8609b6952bd4 ~]# ldd `which perl`
        linux-vdso.so.1 =>  (0x00007ffceaa9b000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f9a190cf000)
        libnsl.so.1 => /lib64/libnsl.so.1 (0x00007f9a18eb6000)
        libdl.so.2 => /lib64/libdl.so.2 (0x00007f9a18cb2000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f9a18a2f000)
        libcrypt.so.1 => /lib64/libcrypt.so.1 (0x00007f9a187f6000)
        libutil.so.1 => /lib64/libutil.so.1 (0x00007f9a185f3000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f9a1829a000)
        /lib64/ld-linux-x86-64.so.2 (0x00005593a3dcf000)

```

Perl 5.20 is fine. Can we revert back?

gqmelo:
Hi,

> Building of packages that require gcc/libgcc in perl is going to fail, because perl itself is compiled with a newer version of gcc.

Actually the error ```perl: /lib64/libc.so.6: version `GLIBC_2.11' not found (required by perl)``` has nothing to do with ```gcc```. This error basically means that perl was compiled in a system with a newer libc and you are trying to run in a system with an older libc.
conda-forge uses CentOS 6 to build, which means conda-forge packages should work in systems with glibc >=2.12.

Are you using CentOS 5 or something else older than CentOS 6? Because then there is no guarantee that those conda-forge packages will work. Please note that some of them may work though depending on the glibc functions they use. That is probably the case with perl 5.20.

By the way, you can easily find glibc version of any distribution in https://distrowatch.com.

bgruening:
@gqmelo thanks for looking into it, but we are using the same Docker image than conda-forge. So something seems to be wrong here.

nsoranzo:
My package seems to build fine if I remove the `gcc` requirement, not sure if this is a good idea though. Suggestions?

gqmelo:
> but we are using the same Docker image than conda-forge. So something seems to be wrong here.

Ok, but where did you find the error ```version `GLIBC_2.11' not found (required by perl)```? I can't find it in the travis log.

Also about the ```-fstack-protector-strong``` flag, if you don't want to change the compiler, you can just replace it with ```-fstack-protector``` wherever the flag is used (you probably will have to patch a ```configure``` script or something)

nsoranzo:
`-fstack-protector-strong` appears in CFLAGS because `Module::Build` imports it from Perl's own `Config` module, which in turn reads it from the file `lib/perl5/5.22.0/x86_64-linux-thread-multi/Config_heavy.pl`

There is really no easy way to change it because it is not written in any `configure` or `Makefile`. And it is particularly problematic because any Perl module using `Module::Build` and `gcc` would have to be somehow "fixed".

gqmelo:
Ok, I see. I'm not familiar with perl, but if can't change those flags while compiling a module I think your options are:

- use the gcc available on the docker image
- use gcc >=4.9
- build your own perl package
- rebuild perl on conda-forge trying to remove the ```-fstack-protector-strong``` flag and any other flag that was backported by RedHat so the the Config module does not contain those flags

nsoranzo:
> I think your options are:
>
> - use the gcc available on the docker image

AFAIK that should be fine unless the Perl module uses Fortran or OpenMP.

> - use gcc >=4.9

There is no such a package on `defaults`, `conda-forge` or `bioconda` channels.

> - build your own perl package

We just removed our Bioconda `perl-threaded` package to use the conda-forge `perl`.

> - rebuild perl on conda-forge trying to remove the ```-fstack-protector-strong``` flag and any other flag that was backported by RedHat so the the Config module does not contain those flags

That would be easily obtained by adding a build requirement on `gcc` to the conda-forge `perl` recipe, but I think they would disagree, see https://github.com/conda-forge/conda-forge.github.io/issues/29.

gqmelo:
> That would be easily obtained by adding a build requirement on gcc to the conda-forge perl recipe, but I think they would disagree

Using gcc as a package in conda-forge is certainly not an option. But I meant checking where in perl compilation this flag is added. You can probably just avoid the flag using some configuration options, then submit a PR to perl feedstock and see if the maintainers find it acceptable.

jerowe:
@gqmelo, you're right. I do my local builds against the conda-forge image, and was trying to debug a failing perl package. I used the deprecated bioconda/biobuilder image, which is centos5. So please disregard my message about the glibc. There is still something going on with builds of perl packages, but you're right that this is not it. ;-)

johanneskoester:
I am not so sure about gcc. Actually, there are other feedstocks on conda-forge that use gcc, e.g. our new R feedstocks. I would say we should just try a PR adding gcc and see how they react.

nsoranzo:
@johanneskoester I'd do that but I think it may have higher chances if it comes from a Bioconda core member.

johanneskoester:
You are right. I've done it.

bgruening:
Thanks @nsoranzo!

PR #3840, opened 2017-02-18
Add Taco-0.7.0 recipe

upendrak:


upendrak:
@bgruening can you please let me where the problem is with the build? Thanks

bgruening:
It seems to be a precompiled binary, isn't it? T his binary is not able to run on old systems and complains about missing libraries. Can you compile it?

upendrak:
Yes it is a precompiled binary and i tested it thoroughly on linux and it worked great. Can we install the missing libraries and then rebuild it. I am getting all sorts of errors when i compile it.

bgruening:
Solved in https://github.com/bioconda/bioconda-recipes/pull/3851

PR #3845, opened 2017-02-18
Add python 3.6 to env matrix

kyleabeauchamp:

I know there is already an open PR (#3338) for this, but I figured it might be nice to see what the error log looks like with the latest master.

### DIFF
--- a/recipes/pybigwig/meta.yaml
+++ b/recipes/pybigwig/meta.yaml
@@ -9,7 +9,7 @@ source:
 
 build:
   number: 0
-  skip: False
+  skip: True # [py36]
 
 requirements:
   build:
--- a/scripts/env_matrix.yml
+++ b/scripts/env_matrix.yml
@@ -2,6 +2,7 @@ CONDA_PY:
   - 27
   - 34
   - 35
+  - 36  
 CONDA_BOOST: "1.61"
 CONDA_R: "3.3.1"
 CONDA_PERL: "5.22.0"


bgruening:
Things starting to look better as it seems. We have a few big rebuilds coming along the next month and I'm starting to worry about the organisation. Conda-build will release a new version with a new pinning mechanism soon and I guess we should switch to this version and align with the pinning schema of conda-forge. Giving that we need a huge r-rebuild and this python rebuild ... I guess we need to rebuild mostly all packages. Not sure how to organise/coordinate this efficiently.

@kyleabeauchamp you are trying to rebuild all python packages at once here, isn't it. I don't think this will work due to the travis timeout.

kyleabeauchamp:
SGTM, so it sounds like py36 will wait for conda-build and conda-forge upgrades.

I wonder if we could temporarily pay for a paid Travis account to work around the timeouts during the rebuild?  Otherwise some sort of alternative CI system might be necessary?  Or refactoring of the bioconda-utils.

kyleabeauchamp:
SGTM, so it sounds like py36 will wait for conda-build and conda-forge upgrades.

I wonder if we could temporarily pay for a paid Travis account to work around the timeouts during the rebuild?  Otherwise some sort of alternative CI system might be necessary?  Or refactoring of the bioconda-utils.

bgruening:
More interesting than the timeouts, are concurrent builds and surprisingly you can not get much more even if you pay afaik. Different CI system is for sure interesting, for example to separate out the Linux builds, but this again involves a lot of build-infrastructure work. Not sure anyone of the core team has currently the resources to drive this effort :(

The easiest way would be to build it locally on some cluster, but I kind of like to idea that everything is compiled on public CI's and not hidden - but this comes at it's price.

johanneskoester:
Open-source projects have 4 concurrent builds in Travis. We could ask for more. The Rust project, e.g., has been granted 15 concurrent builds. We could also offer them some money (i.e. a custom pro account), if they don't like to give it for free. If possible, I'd like to stick to one single CI. @bgruening should I ask them, or do you already have a running contact with them (because of the build time extension)?

kyleabeauchamp:
I'm happy to contribute somewhat a portion of the CI costs if that helps.

kyleabeauchamp:
I think we should think about having some sort of non-profit / OSS organization sponsor bioconda for the purposes of donation.  Either a host university or NumFocus.

bgruening:
@johanneskoester no contract here, please ask! I just remembered one from conda-forge that said more concurrent builds are hard to get.

@kyleabeauchamp conda-forge is thinking about joining NumFocus, not sure if we are big enough. But a host university sounds like a great idea.

kyleabeauchamp:
PS: I think Folding@Home is a good template for getting infrastructure donations etc:

https://folding.stanford.edu/home/donate/

johanneskoester:
These are good suggestions. I also have a grant that could be used, and I might get some money from a some us company that use bioconda. But before we do any of these steps, I want to get rid of the redundancy in our build system: in pull requests, every commit is built twice. And then again in master. What can we do? We can try to use the caching capabilities of Travis. I will look into it this week.

bgruening:
If you create a PR from your fork it only gets build once. So maybe we should encourage people to do that.
Moreover, I kind of feel that the entire recipe scanning is not good for build times, often recipes are rebuild over and over again just, because there are on master and not on anaconda.org. We should consider just building what is in the git diff between branches and for pinned versions we should have a separate system that filters all recipes and rebulids only the ones that are necessary.

johanneskoester:
I think it should be possible to get the same behavior for builds from the main repo. Indeed, the filtering is another thing to improve. I'll try to figure something out in the next weeks.

druvus:
py36 is now added

PR #3853, opened 2017-02-19
Add missing dependencies.

johanneskoester:

The pindel package was broken, because it fails to find libgomp.so in stripped down containers. That library is part of libgcc, hence, this PR adds gcc/libgcc as dependencies to fix pindel.


bgruening:
Thanks Johannes!

PR #3864, opened 2017-02-20
Mercat recipe

ajaypanyala:

### DIFF
--- a/recipes/mercat/meta.yaml
+++ b/recipes/mercat/meta.yaml
@@ -14,7 +14,7 @@ build:
   script: python setup.py install
 
 
-preserve_egg_dir: yes
+#preserve_egg_dir: yes
 
 requirements:
   build:
@@ -23,6 +23,7 @@ requirements:
 
   run:
     - python
+    - pyqt
     - scikit-bio
     - prodigal
     - trimmomatic
--- a/recipes/mercat/meta.yaml
+++ b/recipes/mercat/meta.yaml
@@ -1,0 +1,53 @@
+
+package:
+  name: mercat
+  version: "0.1"
+
+source:
+   git_url: https://github.com/pnnl/mercat.git
+#  git_rev: 0.1
+
+# build:
+#  number: 0.1
+
+build:
+  script: python setup.py install
+
+
+preserve_egg_dir: yes
+
+requirements:
+  build:
+    - python
+    - setuptools
+
+  run:
+    - python
+    - scikit-bio
+    - prodigal
+    - trimmomatic
+    - dask
+    - setuptools
+    - pandas
+    - numpy
+    - humanize
+    - plotly
+    - psutil
+    - joblib
+
+
+test:
+  commands:
+    - mercat --help
+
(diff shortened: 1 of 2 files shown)


daler:
@ajaypanyala please don't modify the build environment by using apt-get install. Any dependencies should be listed in the recipe. In this case, add `pyqt` to the requirements section of the recipe.

ajaypanyala:
Thank you Ryan! I changed it back.

-Ajay

recipes/mercat/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - python
+    - setuptools
+
+  run:
+    - python
+    - scikit-bio
+    - prodigal
+    - trimmomatic
+    - dask
+    - setuptools
```
bgruening: Is setup tools really needed as a runtime dependency?

recipes/mercat/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - pandas
+    - numpy
+    - humanize
+    - plotly
+    - psutil
+    - joblib
+
+
+#test:
```
bgruening: Please include a test.

druvus:
Please avoid usin git_url. A regular url with a checksum is recommended. You could do somethin like this for package, source and build sections:

```
{% set name = "mercat" %}
{% set version = "0.1" %}

package:
  name: {{ name|lower }}
  version: {{ version }}

build:
  skip: True  # [not py27]
  number: 0
  entry_points:
    - mercat=mercat.mercat:mercat_main

source:
  fn: {{ name|lower }}_{{ version }}.tar.gz
  url: https://github.com/pnnl/mercat/archive/0.1.tar.gz
  md5: bce67a7bcb58c8daed4a00b377dad097
```

recipes/mercat/meta.yaml
ajaypanyala: It's not needed. Will remove that. Thanks!

recipes/mercat/meta.yaml
ajaypanyala: A simple test passes on MacOSX build, but not for Linux. When trying to import python modules, I get the error  
> ImportError: libstdc++.so.6: cannot open shared object file
I tried specifying libgcc, gcc as runtime dependencies in the recipe, but I still see it.
[https://travis-ci.org/ajaypanyala/bioconda-recipes/jobs/203529745](https://travis-ci.org/ajaypanyala/bioconda-recipes/jobs/203529745)

druvus:
@ajaypanyala please take a look at the bioconda contributing guidelines and especially the [checklist](https://bioconda.github.io/guidelines.html#bioconda-recipe-checklist). It is probably a reason why a test is failing so skipping the test section and merging after @bgruening requested tests might not be the best strategy.

ajaypanyala:
@druvus @bgruening  Sorry about the merge. I will fix the test section, run travis build again and merge it once they pass. Thanks!

PR #3867, opened 2017-02-20
Adding the newest GMAP/GSNAP version

maciej-kandula:

This pull allows using the newest GMAP/GSNAP version which resolved some fatal bugs.
Note, this is my first pull request.

I ran the 'conda build' without any issues.
On my Debian machine './simulate-travis.py' returns "non-zero exit status 1", however. No idea what is wrong though as this is a copy of the old version, just with the file (and version entries) changed.


recipes/gmap/2016.09.23/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+env MAX_READLENGTH=500 ./configure --prefix=$PREFIX
```
yhoogstrate: Do you mind adding the following:

```
# Citation from the README:
#
# Note 3: GSNAP previously had a configure variable called
# MAX_READLENGTH (default 300) as a guide to the maximum read length.
# That variable is no longer needed, since GSNAP can align reads of
# arbitrary length.  (But, for longer reads, GMAP will probably be much
# faster.)
#
# However, whenever possible, based on the length of the read, GSNAP
# will use stack memory instead of heap memory for some algorithms.  To
# control this decision, there is a variable called
# MAX_STACK_READLENGTH, set like this
#
#    ./configure MAX_STACK_READLENGTH=<length>
```

recipes/gmap/2016.09.23/meta.yaml
```diff
@@ -0,0 +1,46 @@
+package:
+  name: gmap
+  version: "2016.09.23"
```
yhoogstrate: Is there a obvious reason for this version over *2017-02-15*?

recipes/gmap/2016.09.23/build.sh
maciej-kandula: Done, thanks for the hint

recipes/gmap/2016.09.23/meta.yaml
maciej-kandula: Version 2016-09-23 should work for most people. However, on some scarce occasions (when a fasta looks in a very particular way) it will crash. It is not possible to go around this using any of the available parameters.
2017-02-15 takes care of the above.

maciej-kandula: To clarify: 2016-09-23 was previously the "default" version in bioconda. I just moved it to a specific dir.

bgruening:
@maciej-kandula thanks a lot for this update!

PR #3876, opened 2017-02-21
Build lapack for osx

druvus:


bgruening:
@druvus I guess this package should be migrated into bioconda-legacy and we should use the conda-forge one.

druvus:
@bgruening you're right!

PR #3886, opened 2017-02-22
Feature/perl rebuild part2

jerowe:


perl-gcc-blacklist
```diff
@@ -0,0 +1,2 @@
+recipes/perl-time-hires
```
bgruening: This is because of the conda-forge discussion about the wrong GCC version used by perl?

perl-unable-to-find-blacklist
```diff
@@ -0,0 +1 @@
+recipes/perl-font-afm
```
bgruening: What is this about?

bgruening:
Great work  @jerowe!!!

jerowe:
Yes.

bgruening:
Then let's merge it. I think it is ok to compile these perl-gcc package without gcc explicitly specified. This should work and should be fixed as soon as we get a newer GCC package.

perl-unable-to-find-blacklist
jerowe: I meant to ask you the same thing. There are some recipes that I can run `conda build /bioconda/recipes/perl-font-afm`, and I'm fine, but running using the bioconda utils decides there are no files in the download, or possibly downloading has gone awry. I had a nice report on my work computer, and will get back to you with it tomorrow.

jerowe:
Did all of these pass? I thought there were still a few packages that needed some tweaks.

bgruening:
I hope they did, it was green and I have checked a few of them from bioconda and biocontainers.

jerowe:
They did. I ran a local check too

PR #3892, opened 2017-02-22
kraken: Enable OS X builds via gcc dependency

yesimon:


yhoogstrate:
What happens if you run: `export CXX=$(which -a python | tail -n+2)`?

recipes/kraken/meta.yaml
```diff
@@ -18,8 +17,10 @@ build:
 
 requirements:
   build:
+    - libgcc
```
bgruening: The other way around gcc as build dep and libgcc as runtime dep.

druvus:
Closing, kraken was built for osx in #5234.

Issue #3418, opened 2017-01-10
perl-threaded is gone


bgruening:
Hi BioConda!

From now on please use `perl` and not `perl-threaded` for your perl packages.

https://github.com/bioconda/bioconda-utils/pull/56

Thanks,
Bjoern

PR #3900, opened 2017-02-23
ericscript: perl-threaded -> perl

roryk:
as instructed in #3418.

### DIFF
--- a/recipes/ericscript/meta.yaml
+++ b/recipes/ericscript/meta.yaml
@@ -8,14 +8,14 @@ source:
   md5: "a871d2c1f92c407ee26a6cabdc870dc6"
 
 build:
-  number: 0
+  number: 1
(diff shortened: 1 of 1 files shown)


bgruening:
SAMtools not found? https://travis-ci.org/bioconda/bioconda-recipes/jobs/204679836#L1897

roryk:
Thanks, I've tried about a bunch of things to fix this by patching ericscript.pl to change the PATH and I can't get this to not throw this error when running the simulate-travis script. I think I might be missing something obvious. Any clues?

chapmanb:
It looks like this was a permission error on a file from samtools that was only triggered when running with mulled. I updated samtools 0.1.19 to fix the permissions and it now builds cleanly. Thanks Rory for working on this.

Issue #3889, opened 2017-02-22
Travis CI: Error: no such patch: /opt/recipe/


CicoZhang:
Hey, everyone. I am adding an old package to bioconda. I have applied a patch, but only for OSX. It's been twice already that I got this error from Travis for the linux build. What does this error mean? Conda build works swimmingly locally.  I am not sure whether this error concerns the patch.

![image](https://cloud.githubusercontent.com/assets/11710225/23215014/1f4a1156-f911-11e6-92e9-d7c9840e3f43.png)

https://travis-ci.org/CicoZhang/bioconda-recipes/jobs/204195323

Thank you in advance.

CicoZhang:
Interestingly, OSX has passed the test already, though I am using Ubuntu.
![image](https://cloud.githubusercontent.com/assets/11710225/23217323/238299b6-f919-11e6-934f-b84df156eade.png)

bgruening:
@CicoZhang this simply means that your Linux build is broken. I might not get the question, sorry :(

CicoZhang:
Hi, @bgruening , conda build the package locally perfectly, but fail to pass on Travis (Linux fails, OSX passed).

I get the error like this:

```
Error: no such patch: /opt/recipe/
ERROR bioconda_utils.build:build_recipes(375): BIOCONDA BUILD SUMMARY: of 1 recipes, 1 failed and 0 were skipped. Details of recipes and environments follow.
ERROR bioconda_utils.build:build_recipes(386): BIOCONDA BUILD SUMMARY: FAILED recipe ogdf-Sakura-0.tar.bz2, environment CONDA_R=3.3.1;CONDA_GSL=1.16;CONDA_PY=27;CONDA_HDF5=1.8.17;CONDA_BOOST=1.61;CONDA_GMP=5.1;CONDA_NPY=110;CONDA_PERL=5.22.0;CONDA_NCURSES=5.9
/home/travis/build.sh: line 158:  3465 Terminated              travis_jigger $! $timeout $cmd 
```

The whole log can be found [here](https://travis-ci.org/CicoZhang/bioconda-recipes/jobs/204195323)

Are there any 'obvious' mistakes here :) ?
![image](https://cloud.githubusercontent.com/assets/11710225/23218229/03a9b978-f91c-11e6-851e-861062fa2889.png)

daler:
[...]

PR #3901, opened 2017-02-23
OGDF: 201207

CicoZhang:

[OGDF](http://www.ogdf.net/doku.php) is a self-contained C++ class library for the automatic layout of diagrams. OGDF offers sophisticated algorithms and data structures to use within your own applications or scientific projects. The library is available under the GNU General Public License.


CicoZhang:
@daler I have changed the style of version number, but it still fails to pass the linux platform. Conda-build works well locally (I am using Ubuntu 16.10) #3889

acaprez:
I've run into this issue as well.  I'm fairly sure it's a bug in the way `conda-build` handles the `patches:` section of the `meta.yaml`.  Basically, if there's a single entry in the list that only applies to either Linux or OS X via preprocessing selectors, `conda-build` doesn't parse the "empty" list for the other OS properly and looks for a nonexistent patch file.  The workaround I found was to use the preprocessing selector for both the patch file line and the `patches:` line itself.  See for example the `meta.yaml` for [bioconductor-rhdf5](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bioconductor-rhdf5/meta.yaml).

CicoZhang:
@acaprez Cool. Thank you for the workaround for this naughty bug. :)  

@daler @bgruening  My problem is solved. It has passed.

bgruening:
Cool! Then let's merge it.

PR #3933, opened 2017-02-26
Linting

daler:

This PR integrates the changes from https://github.com/bioconda/bioconda-utils/pull/57 into bioconda-recipes.

The idea is to reduce the workload of reviewing PRs by automatically flagging common problems. By default, all linting functions defined in `bioconda_utils.lint_functions.registry` are run for all newly-changed recipes in a PR. If any of these fail, the entire build fails and the travis log message indicates what the problems were, along with a link to the docs where fixes are described for each problem.

In some cases we may want to skip a linting test, so there is a mechanism for this. When testing on travis, we can add special text in the commit message, for example `[lint skip in_other_channel for recipes/packagename]` will skip the "in_other_channel" lint test for the recipe "recipes/packagename". A similar mechanism exists for testing locally with `simulate-travis.py`.

See the linting docs for more, which will be at https://bioconda.github.io/linting.html once the changes are merged with bioconda-utils master. 

This PR includes a demo. I've bumped the build number of `argparse`. Since changes to this recipe are within the commit range of this PR, it is linted. However, the recipe is missing tests so the build should fail quickly upon failing the `missing_tests` lint test. In a subsequent commit, I'll include the text `[lint skip missing_tests for recipes/argparse]` in the commit message, which will skip that test for this recipe and allow it to build (hopefully successfully).


johanneskoester:
Great work! This will be a huge improvement to our workflow.

PR #3940, opened 2017-02-27
Bump MultiGPS to version 0.73

gregvonkuster:


gregvonkuster:
@bgruening No idea why this is timing out since, with the exception of the downloaded jar file, this recipe is the same as the previous versions.  Not sure what to try next...

gregvonkuster:
@bgruening Thanks for helping with this.  Any idea why the job log won't display?  Yesterday the browser would fetch indefinitely, today I'm seeing this error.
```
There was an error while trying to fetch the log.
```

daler:
@gregvonkuster that's likely due to the S3 outage yesterday; I restarted the test.

gregvonkuster:
Thanks a bunch @daler ;)

gregvonkuster:
Hmm..same problem - travis fails, but the job log doesn't load:
```
There was an error while loading data. 
```

daler:
@gregvonkuster seems to be lingering effects . . . https://github.com/bioconda/bioconda-recipes/pull/3966#issuecomment-283377433

gregvonkuster:
@daler Ah, thanks for the info!  ;)

bgruening:
More infos here: https://www.traviscistatus.com/incidents/hx7cnxbch9xf

gregvonkuster:
@bgruening it looks like travis is now working so that this can be re-tested?

gregvonkuster:
Is this still a travis issue or something else?  This is a very simple recipe that upgrades 2 earlier working recipes where the only change is the URL to a new .jar file.  I can now see the job log, but it just states the timout.  Am I missing something?

bgruening:
I will look into it this weekend.

gregvonkuster:
Ah, so sorry to do this to you, but thanks a bunch @bgruening .  I'll keep a lookout on this one.

bgruening:
For yet unknown reason, gcc is defined as run-time dep ... :(
So one of your deps is installing gcc ...

gregvonkuster:
Interesting...  Thanks for finding the problem - I'm not sure which of the dependencies needs it - I still see nothing in the test log.  This seems like it will be a problem for some recipes that have dependencies that require gcc.  I guess in some cases libgcc will need to be a run dependency of downstream packages.  This seems kind of strange.  Maybe the libgcc runtime requirement is missing at the dependency level.  Hmm...

bgruening:
Ok, here we go. I spend too much time one that and still don't understand it. For some reason conda have choosen an older meme version and this meme version was "broken" as it was depending on gcc as runtime dependency.

gregvonkuster:
Thanks so much @bgruening - it is good to know this about meme.  Your help on this is so much appreciated.  You're the best!  ;)

Issue #3947, opened 2017-02-27
Replace openmpi by the one on conda-forge?


bebatut:
Hi,

I saw there is a recipe for `openmpi` here, and also one on conda-forge. And the one on conda-forge seems more up-to-date.
Do you think it makes sense to remove the one here and use only the one in conda-forge?
@jerowe What do you think?

Currently, only 2 recipes are using it here (`abyss`  and `oases`).

Bérénice

keuv-grvl:
OpenMPI is a general purpose software, so it would fit better at conda-forge. Moreover, there are 7 contributors on https://github.com/conda-forge/openmpi-feedstock, so it's a big help for us.

jerowe:
Lets get rid of it then. I probably stole the recipe from conda-forge anyways. ;-)

bgruening:
Agree, please create a PR that removes the recipe, I will move it to the bioconda-legacy channel afterwards.

bgruening:
The package is now migrated to bioconda-legacy.

bebatut:
Thanks @bgruening

epruesse:
This caused a regression for `conda install sina`, which requires `arb-bio`, which requires `phyml=3.2.0`, which requires `openmpi=2`, which is no longer available.

epruesse:
@bioconda/core Anyone have an idea of how to fix this? The current versions of `sina`, `arb-bio`, and `phyml` are effectively no longer installable, breaking reproducibility.

kyleabeauchamp:
Presumably we could have the conda-forge folks createa  openmpi=2 recipe?

bgruening:
The openmpi story should be much better these days. Feel free to reopen if needed.

PR #3963, opened 2017-02-28
Remove OpenMPI recipe

bebatut:

This PR removes a recipe for OpenMPI as discussed in #3947, to use the one in `conda-forge`


bgruening:
@johanneskoester is this an intended behavior? If so can we force-merge such PR?

johanneskoester:
Not really. So, it correctly detects a change with openmpi. Then it tries to build it, but since the recipe is not there, it fails. Hence, it exists with status 1.
I guess it would be correct to check whether a change detected with git-range is a deletion. In that case, one should exit with 0. I could fix that now, but given the refactorings of @daler, it would cause conflicts. @daler do you want to build that into your PR (a quick `os.path.exists` on the results of `modified_recipes` should be enough)? Otherwise, I'd like to wait until you have merged it.

daler:
Sure -- I'll add that in while addressing your other comments on that PR.

johanneskoester:
Thanks, @daler!!

bgruening:
Everything is normal, nothing to see here ... ;)
Sorry, I will force merge this as I was told this is blocking some packages.

PR #3968, opened 2017-02-28
add transposome recipe

cmonjeau:


cmonjeau:
@abretaud and red :)

It appears that it requires at least gcc 4.9 because of the -fstack-protector-strong used by perl libs.

bgruening:
Just remove gcc entirely ... :)

PR #3970, opened 2017-02-28
add sequana recipes

cokelaer:


cokelaer:
close since biokit dep is currently only available for py27 and sequana is for py35 only

PR #3973, opened 2017-02-28
Early skip

daler:

xref https://github.com/bioconda/bioconda-recipes/commit/1f42e58b08b8493e8209bc9ec0c194bf0987b4a2#commitcomment-21063753

### DIFF
--- a/scripts/travis-run.sh
+++ b/scripts/travis-run.sh
@@ -1,6 +1,16 @@
 #!/bin/bash
 set -euo pipefail
 
+if [[ $TRAVIS_BRANCH != "master" && "$TRAVIS_PULL_REQUEST" = "false" && $TRAVIS_REPO_SLUG = "bioconda/bioconda-recipes"]]
+then
+    echo ""
+    echo "Tests are skipped for pushes to the main bioconda-recipes repo."
+    echo "If you have opened a pull request, please see the full tests for that PR."
+    echo "See https://bioconda.github.io/build-system.html for details"
+    echo ""
+    exit 0
+fi
+
 # determine recipes to build
 if [[ $TRAVIS == "false" ]]
 then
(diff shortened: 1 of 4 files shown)


scripts/travis-run.sh
```diff
@@ -1,6 +1,16 @@
...
+if [[ $TRAVIS_BRANCH != "master" && $TRAVIS_PULL_REQUEST == "false" && $TRAVIS_REPO_SLUG == "bioconda/bioconda-recipes" ]]
+then
+    echo ""
+    echo "Tests are skipped for pushes to the main bioconda-recipes repo."
+    echo "If you have opened a pull request, please see the full tests for that PR."
+    echo "See https://bioconda.github.io/build-system.html for details"
+    echo ""
+    exit 0
+fi
+
```
johanneskoester: So, if we exit with 1, we could move this to travis-setup and save even more time, right?

daler: Ah, good point! But will a failure cause problems with merging?

johanneskoester: It currently seems to be impossible to hide the status of the build from github. So we either have X or ✓ and there is no way around.

johanneskoester: Yes, that's right. We can't merge if one fails. I wonder if all this is worth it. Are there really people that patiently wait until their branch is build in the fork until they issue a PR?

daler: I do sometimes :). Actually I think the big use-case for this is testing on OSX for people without access to a Mac.

If we exit 0 on the branch push, I think the only disadvantage is that people might get confused when the full test on the PR fails but the branch push passes. That can be alleviated with informative messages and documentation.

johanneskoester: You are right. Ok, let's do it like this. But additionally, could you add the same check to travis-setup? This way, we also skip the deployment of miniconda, and get a minimal build time.

PR #3980, opened 2017-03-01
Added seqtk-fork, which is a fork of seqtk.

ndaniel:


ndaniel:
Hi kyleabeauchamp!

The author of original seqtk-fork is needed by [FusionCatcher](https://github.com/ndaniel/fusioncatcher & https://sourceforge.net/projects/fusioncatcher/files/ ), which I am planning to bring to bioconda due to demand from the users (see: https://github.com/ndaniel/fusioncatcher/issues/20 ).

The author of original seqtk has never accepted by changes to seqtk and therefore I had no choice than to fork it myself and use it that way.

kyleabeauchamp:
Sounds good to me, maybe mention the fusioncatcher detail in the `summary` section in `meta.yaml` so that it's obvious to future users?

ndaniel:
@kyleabeauchamp 

Excellent suggestion! I will add it the summary.

ndaniel:
Added also FusionCatcher tool.

ndaniel:
@kyleabeauchamp

The requested info has been added to `meta.yaml`.

recipes/fusioncatcher/meta.yaml
```diff
@@ -0,0 +1,46 @@
+package:
+  name: fusioncatcher
+  version: "0.99.7b"
+
+source:
+  fn: fusioncatcher_v0.99.7b.zip
+  url: https://sourceforge.net/projects/fusioncatcher/files/fusioncatcher_v0.99.7b.zip
+  sha256:
```
yhoogstrate: `a062fa3b8069c943ccf78e2b2d27f0cc7fcd73b119592ce96c3320653f4ce24b`

bgruening:
I hate when this happens. @ndaniel are you referring to https://github.com/lh3/seqtk/pull/38 ? Seems like @lh3 just implemented it in a different way?

ndaniel:
@bgruening 

This [lh3/seqtk#38](https://github.com/lh3/seqtk/pull/38) proves nicely my point and the need for this fork of seqtk! So  [lh3/seqtk#38](https://github.com/lh3/seqtk/pull/38) has the title **added -B/-E to trimfq for keeping first/last INT bp** and @lh3 very clearly stated [here](https://github.com/lh3/seqtk/pull/38) in his final comment when closing the pull that "-E" is still not implemented in the official seqtk. The option "-E" for seqtk is needed by FusionCatcher as shown [here](https://github.com/ndaniel/fusioncatcher/blob/master/bin/fusioncatcher.py#L6997).

So [seqtk-fork](https://github.com/ndaniel/seqtk) is needed by [FusionCatcher](https://github.com/ndaniel/fusioncatcher) as shown [here](https://github.com/ndaniel/fusioncatcher/blob/master/DEPENDENCIES#L11) in dependencies needed by  [FusionCatcher](https://github.com/ndaniel/fusioncatcher). 

[FusionCatcher](https://github.com/ndaniel/fusioncatcher) needs [seqtk-fork](https://github.com/ndaniel/seqtk) in bioconda in order get [FusionCatcher](https://github.com/ndaniel/fusioncatcher) in bioconda.

So every user who is using [FusionCatcher](https://github.com/ndaniel/fusioncatcher) is actually using [seqtk-fork](https://github.com/ndaniel/seqtk)! For example, [FusionCatcher v0.99.6a](https://sourceforge.net/projects/fusioncatcher/files/fusioncatcher_v0.99.6a.zip) was downloaded over 1706 times since April 1, 2016 as shown [here](https://sourceforge.net/projects/fusioncatcher/files/fusioncatcher_v0.99.6a.zip/stats/timeline?dates=2016-04-01+to+2017-03-03). Also [FusionCatcher](https://github.com/ndaniel/fusioncatcher), which needs/uses [seqtk-fork](https://github.com/ndaniel/seqtk), has been used in [32 scientific articles](https://github.com/ndaniel/fusioncatcher/blob/master/doc/manual.md#3---fusioncatcher-in-scientific-articles). [FusionCatcher](https://github.com/ndaniel/fusioncatcher) has been used by teen cancer survivor Elana Simon to study her own rare disease as shown [here](http://www.independent.co.uk/news/science/teen-cancer-survivor-elana-simon-helps-scientists-study-her-own-rare-disease-9160379.html), who was also given as example of precision medicine at White House for this (see: [here](http://www.nydailynews.com/news/national/teen-cancer-science-project-honored-president-obama-article-1.1807789)).

yhoogstrate:
Before merging this let's try merging the -E patch one More Time and please Report the usage statisticus of your fork then. This way of forking is confusing for  users and packaging and not beneficial to the Community (Have had similar situations and i know the frustration).

kyleabeauchamp:
What happens if the user installs both seqtk and seqtk-fork?  What happens to the executables---overwritten?

ndaniel:
@kyleabeauchamp 
I do not know. The authors of conda could answer probably those questions better than me and also these:
- what happens if the user installs both Perl and Perk-forks (see: https://anaconda.org/bioconda/perl-forks )? 
-What happens to the executable--overwritten?

ndaniel:
@yhoogstrate 

You wrote: 
> This way of forking is confusing for users and packaging and not beneficial to the Community...

I would say that forking and supporting forking is beneficial for the Community. 
FusionCatcher has been used successfully used by the Community and it has been very beneficial for the Community. Please see:
- http://www.nydailynews.com/news/national/teen-cancer-science-project-honored-president-obama-article-1.1807789
- http://www.usnews.com/news/the-next-generation-of-stem/articles/2015/05/12/white-house-recognizes-new-york-teens-cancer-research
- http://www.independent.co.uk/news/science/teen-cancer-survivor-elana-simon-helps-scientists-study-her-own-rare-disease-9160379.html

bgruening:
@ndaniel you are closing your PR to soon :)
To be clear I value the possibility to create forks and create a healthy community. But the referenced PR was given up to early. You never responded to hist closing, correct. Fight for your option which you need!

The problem with these kind of forks is, that the main project is well maintained and you fork is not. Which I can understand, you intention was not to `fork` it and maintain it, you just wanted this one option.

The perl example you posted was a bad one, fork here is referring to forking a process, like in threading ... so no valid. But we do support forks if there is a valid reason. I'm not even opposed to merge this PR but a little bit more poking and trying to get this upstream would be great! I know it's a pain, we are dealing with this on a daily basis to get your packaging patches online. But it's worth it imho, the community benefits in the end if you get your patches upstream, more then from your unmaintained fork.

yhoogstrate:
Hi again Daniel,

I am really happy to read this news and I am proud of your results. I am also really glad that you not only use but also contribute to free scientific software (which I wish many more people should be doing). However, I don't see how this news is relevant to the technical implementation (i.e. whether or not to fork) we are discussing here.

Before this becomes an advertisement for FusionCatcher I shall only further respond to technical details. Unfortunately you needed to make a fork of seqtk rather than that your patches were incorporated or even fully reviewed. Creating a fork at itself is not a problem if it is only used for FusionCatcher with its built-in bootstrap method.

But now we are trying to move towards bioconda and here forks shall cause confusion when people start searching for `seqtk` and end-up with multiple results, named `seqtk-fork` and `seqtk`. It is just not intuitive and people don't know which one to choose, especially seqtk users that have nothing to do with fusioncathcer. Also from a maintainance perspective it is unnecessary to maintain software twice. When I think forking is really valuable is when projects end-up comepletely dead but I would not say this is the case here. That's why I am tyring to help you here and suggest to ask the seqtk author to reconsider reviewing and accepting your feature. I am even willing to spend time reviewing that code too. The solution would not only benefit you but also all other seqtk users and bioconda users. If it turns out this is not going to work out for seqtk, we can think of other options, but ending up with two nearly identical named forks of seqtk in bioconda does not seem like the best solution to me.

Youri

ndaniel:
@bgruening 

> To be clear I value the possibility to create forks and create a healthy community. 
> But the referenced PR was given up to early. 

I think that this is the best way forward. If someone else wants to take this further up it is great!

> You never responded to hist closing, correct. Fight for your option which you need!

No need to fight when I get answers which contain the word *hate*.

>The problem with these kind of forks is, that the main project is well maintained and you fork is not. 

First this is a problem of Bioconda. Second my fork is maintained well!

> Which I can understand, you intention was not to fork it and maintain it, you just wanted this one option.

>The perl example you posted was a bad one, fork here is referring to forking a process, like in threading ... so no valid. 

Ok. I hear you. So my fork of seqk just does **extra** three things which are:
- added option "-E" which does not exist in original seqtk
- added option "-B" which does not exist in original seqtk
- fixed a bug regarding the reads which have length zero which exists in seqtk and it is amplified in cases when options -E or -B are used.

I did a pool request for all these three on September 20, 2014 and finally only one (that is -E) of these three options was implemented on April 6, 2016, after almost two years (whilst -B was ignored)! So, merging the remaining two things in seqtk would take few more years considering how the things are going now and in the end the merging into the official seqtk does not depend ultimately on me.  

So the fork of seqtk behaves like any other version of seqtk from which is derived. The fork could be easily seen as a version r82b instead of official r82 of seqtk. I think that it is not fair to submit my fork of seqtk as a very specific version of the official seqtk because if it has some bugs than someone will take the blame.

>But we do support forks if there is a valid reason. 

When a reason becomes a valid reason in this case? What level of support it is needed to make the reason valid?

> I'm not even opposed to merge this PR but a little bit more poking and trying to get this upstream would be great! I know it's a pain, we are dealing with this on a daily basis to get your packaging patches online. But it's worth it imho, the community benefits in the end if you get your patches upstream, more then from your unmaintained fork.

ndaniel:
@yhoogstrate 


>I am really happy to read this news and I am proud of your results. I am also really glad that you not only use but also contribute to free scientific software (which I wish many more people should be doing). However, I don't see how this news is relevant to the technical implementation (i.e. whether or not to fork) we are discussing here.

I thought that I explained this in my previous posts. So, my patches to official seqtk were not accepted because they are not *popular*. Therefore I had to fork seqtk and use the forked seqtk. Mentioning everything about FusionCatcher had the point to show that actually my fork of seqtk is very popular because it is used by FusionCatcher and FusionCatcher is very popular. Here is a quote from my previous post which shows this " My point here is that the usage of "-E" is very popular and it is implemented only and only in seqtk-fork and not in official seqtk.". Another point here is that merging my patches to seqtk does not depend on me! If they will never get merged what is the way forward then? Does bioconda give a way forward here?

> Before this becomes an advertisement for FusionCatcher I shall only further respond to technical details. Unfortunately you needed to make a fork of seqtk rather than that your patches were incorporated or even fully reviewed. Creating a fork at itself is not a problem if it is only used for FusionCatcher with its built-in bootstrap method.

It looks to me that this is bioconda's problem and not seqtk's or FusionCatcher's.

> But now we are trying to move towards bioconda and here forks shall cause confusion when people start searching for seqtk and end-up with multiple results, named seqtk-fork and seqtk.

It looks to me that this is bioconda's problem and not seqtk's or FusionCatcher's.
So my fork of seqk just does extra three things which are:

added option "-E" which does not exist in original seqtk
added option "-B" which does not exist in original seqtk
fixed a bug regarding the reads which have length zero which exists in seqtk and it is amplified in cases when options -E or -B are used.
I did a pool request for all these three on September 20, 2014 and finally only one (that is -E) of these three options was implemented on April 6, 2016, after almost two years afterwards (whilst -B was ignored because it is not popular enough even that no fact was given to support this)! So, merging the remaining two things in seqtk would take few more years considering how the things are going now and in the end the merging into the official seqtk does not depend ultimately on me.

How bioconda can help here?

> It is just not intuitive and people don't know which one to choose, especially seqtk users that have nothing to do with fusioncathcer. 

I totally agree with you. How bioconda can help here?

>Also from a maintainance perspective it is unnecessary to maintain software twice.

I do not see how this is relevant with seqtk-fork. I mean that one can make the same typee of argument that if a program like STAR has 10 versions then one can claim that the same software needs to maintained ten times. Right?

> When I think forking is really valuable is when projects end-up completely dead but I would not say this is the case here. 

This is correct. The project does not have to be dead in order to make forking useful.
The starting assumption is that forking is good and beneficial for community. 

>That's why I am tyring to help you here and suggest to ask the seqtk author to reconsider reviewing and accepting your feature. 

I have asked already this in  September 20, 2014 and the authors kind of implemented 50% of my request after almost two years. What if my patches will never be accepted by seqtk? How bioconda can help here? Does bioconda support forks?

>I am even willing to spend time reviewing that code too. The solution would not only benefit you but also all other seqtk users and bioconda users. If it turns out this is not going to work out for seqtk, we can think of other options, but ending up with two nearly identical named forks of seqtk in bioconda does not seem like the best solution to me.

I would like to hear first of the options how bioconda can help with forks!

bgruening:
>> You never responded to his closing, correct. Fight for your option which you need!

>I do not have to fight when I get answers which contain the word *hate*. I guess that conda community does not support this kind of answers either.

This comment was referring to your original PR against upstream seqtk. Not sure why this discussion started to be that agressive. Sorry mate!

>>The problem with these kind of forks is, that the main project is well maintained and you fork is not. 

>First this is a problem of Bioconda.

Thats why we discussing it here.

>Second my fork is maintained well!

The latest bugfixes are not pushed back to your fork from upstream right?

>>The perl example you posted was a bad one, fork here is referring to forking a process, like in threading ... so no valid. 

>Ok. I hear you. 

Thanks :)

>Then I have another question. How different versions of the same program/tools are supported then? Aren't there conflicts?

Yes this can happen. But you don't have the same program. Given I want to create a workflow with FusionCatcher and a program that depends on the upstream seqtk I can not create one environment that satisfies this.

> I did a pool request for all these three on September 20, 2014 and finally only one (that is -E) of these three options was implemented on April 6, 2016, after almost two years afterwards (whilst -B was ignored because it is not popular enough even that no fact was given to support this)! So, merging the remaining two things in seqtk would take few more years considering how the things are going now and in the end the merging into the official seqtk does not depend ultimately on me.  

Now we start a real discussion! Thanks!
As I stated several times, I'm not strictly opposed to this and if you really don't want to try it again or even answer to his `closing` action to give him solid feedback, you can still rename your tool completely. As graphicsmagick/imagemagick did for example.

>>But we do support forks if there is a valid reason. 

>When a reason becomes a valid reason in this case? What level of support it is needed to make the reason valid?

Maintenance? Changing the name to not cause confusion?

>> I'm not even opposed to merge this PR but a little bit more poking and trying to get this upstream would be great! I know it's a pain, we are dealing with this on a daily basis to get your packaging patches online. But it's worth it imho, the community benefits in the end if you get your patches upstream, more then from your unmaintained fork.

> Your statement that my fork of seqtk is unmaintained is not correct! 

Sorry for this, but your last changes date back to 2015, the latest upstream changes are not included ...

> Community benefits also from other things than getting patches upstream.

Sure, and we are trying to discuss here if this is the case here as well. Not good?

> How bioconda is supporting forked tools? https://github.com/bioconda/bioconda-recipes/issues/4002

Yes. But now I'm a little bit more careful and say, we do offer also forks from other projects. Not sure if supporting is misleading.

bgruening:
@ndaniel let's get back to real technical discussions. Would it be possible to create an entire new namespace for your fork? seqtknd for example? So that all binaries and libs are called like that? This would enable parallel installation into the same environment for example.

ndaniel:
> This comment was referring to your original PR against upstream seqtk. Not sure why this discussion started to be that aggressive. Sorry mate!

To me it looks very unwelcoming and it is irrelevant for me why it got this way. I have been on bioconda for two days and it looks very unwelcoming.

> Yes this can happen. But you don't have the same program. Given I want to create a workflow with FusionCatcher and a program that depends on the upstream seqtk I can not create one environment that satisfies this.

The answer is yes and no in the same time. The executable has the same name but the package name is different. I can create also workflow with program XYZ and a program which depends on an earlier version of seqtk and again one will not be able to create a environment tat satisfies (because each one will require a different version of seqtk). Right?

> As I stated several times, I'm not strictly opposed to this and...
 
This is irrelevant. I made very clear that is offending to use the word **hate** in an answer because some one who just came two days ago on bioconda has built a PR for a forked tool which does not fit the guidelines which are in somebody's head only. As far as I understand there is no guideline for PR for forked tools.


> you can still rename your tool completely. As graphicsmagick/imagemagick did for example.

It looks like this is the way forward. This is why I closed the issue.

>> Your statement that my fork of seqtk is unmaintained is not correct!
> Sorry for this, but your last changes date back to 2015, the latest upstream changes are not included ...

Again this is very unwelcoming. If I change the name of the fork from seqtk-fork to mumbojumbo (and one would not know this) then suddenly it will be very well maintained? Right?

ndaniel:
>@ndaniel let's get back to real technical discussions. Would it be possible to create an entire new namespace for your fork? seqtknd for example? So that all binaries and libs are called like that? This would enable parallel installation into the same environment for example.

I will try to do that.

bgruening:
> I will try to do that.

Cool!

> "I hate when this happens"

Can you accept a sorry. As I wrote several times this was a misunderstanding and I fully take the blame. Thanks.

daler:
Hi all, just catching up on this.

In this PR, the recipe [copies the seqtk-fork binary](https://github.com/bioconda/bioconda-recipes/pull/3980/files#diff-63f7d5b2c8d2b20824b4cd2338450e31R8) to `$PREFIX/bin/seqtk`. So does the recipe for [`seqtk`](https://github.com/bioconda/bioconda-recipes/blob/f81e377e50d6de6cbe4216fd9ec2037206bd7dbc/recipes/seqtk/build.sh#L8). The issue here is if seqtk (upstream) was already installed, it's `seqtk` executable would get silently overwritten. Conversely if seqtk is installed after fusioncatcher it would overwrite seqtk-fork's `seqtk` executable.

Several options:

1. You could vendor your fork into fusioncatcher. For example [RSEM does this with samtools](https://github.com/deweylab/RSEM/tree/master). To prevent executable naming collisions you would have to rename the executable or possibly keep the executable in fusioncatcher's installation directory and access it via, e.g., distutils.pkg_resources. Looks like you have most of the infrastructure for doing this already in fusioncatcher's bootstrap.py.

2. I'd encourage you to try again getting it into upstream. The author's last comment on that PR was more of a "not now, thanks" rather than a "no! never!". Worst thing that can happen is that your changes still don't get into upstream. Not a big deal. But the advantage of getting it into upstream means that you don't have to maintain your own branch with bugfixes from upstream or maintain a separate bioconda recipe.

3. Change this recipe to create an executable with a different name. I'd prefer the executable and package name `fusioncatcher-seqtk` which would make it clear from the name alone that they're interdependent. `seqtk-fork` seems a bit too generic and requires looking up the differences between them before making a decision on which to install.

For options 1 and 3, the important part is allowing fusioncatcher to be installed in an environment that also contains seqtk upstream.

ndaniel:
@daler 

Thanks for the constructive feedback!

I did a pull request with the latest additions to seqtk, as shown here: https://github.com/lh3/seqtk/pull/90 From here this does not depend on me anymore (until I hear again from the author of seqtk).

Until my patches to seqtk will be merged (if they will ever be merged?) into the official seqtk I will proceed further with a forked seqtk which is [fcseqtk](https://github.com/ndaniel/fcseqtk) (where even the executable is named fcseqtk) which I am planning to get into bioconda. **fc** from fcseqtk comes from FusionCatcher.

daler:
@ndaniel sounds good. I'd say go ahead and update this PR or create a new one, and then whatever the next seqtk release is that has your changes you can switch to that.

For the conflicting version issue, there is no solution for having tools requiring different pinned versions in the same environment. This comes up all the time with Python3-only and Python2-only packages.  For differing version numbers of the same package, the conda solver would complain about incompatible versions and nothing would get installed.  But you can have multiple environments on a system, so the solution in practice is to activate one environment, use one tool, then activate the other environment and use the other tool. It's always a pain to do this, so we try to encourage recipes to specify minimum versions rather than pinned versions.

The conda solver only knows about package names and versions. It does not know the details of what executables are copied where. So we have to be careful about overwriting existing files that the solver doesn't know about.

ndaniel:
Thanks!

This might be an issue for FusionCatcher. Each version of FusionCatcher, by design, uses very specific versions of Bowtie, STAR, and Bowtie2 software (due to specific software/tools having bugs which affect FusionCatcher or removing command line options which FusionCatcher is using, etc.). FusionCatcher's own installer is installing the correct versions of Bowtie/STAR/Bowtie2 into its FusionCatcher's tools directory.

To me it looks like if FusionCatcher is installed by BioConda and afterwards the user might install using bioconda some other tool which upgrades STAR/Bowtie/Bowtie2 and therefore will break FusionCatcher.
Also packaging STAR/Bowtie/Bowtie2 into FusionCatcher looks to me that it brings no added value to FusionCatcher and it is just an overhead (e.g. one has to write a custom scripts in `build.sh` for installing Bowtie/STAR/Bowtie2 inside FusionCatcher's tree structure in bioconda because the official RP for STAR/Bowtie/Bowtie2 cannot be used because they do no guarantee the specific versions being available?).

> But you can have multiple environments on a system...

Let's assume that a user installs FusionCatcher in a very specific environment created just for FusionCatcher. How the user know what tools he/she can still install in this environment which do not interfere with FusionCatcher's tools. For example, installs installs the tool XYZ which depends on a different version of Bowtie (and this is not very obvious from tool's description). This will break FusionCatcher and it will not be obvious why this has happened because looking superficially at it it will look that tool XYZ and FusionCatcher have nothing in common. This would generate a lot of bug reports for FusionCatcher.

The way forward here would be to have FusionCatcher's installion process (e.g. `build.sh`) to call other RP from bioconda and tell those to install the software tools needed by FusionCatcher in `$PREFIX/lib/fusioncatcher/tools` and not in `$PREFIX/bin`. Is this possible?

> So we have to be careful about overwriting existing files that the solver doesn't know about.

Keeping records like this (executable-name, executable-version, executable-CRC) might help a lot here and would help solver figure out what is going on and what has changed.

daler:
You can pin dependency versions in the fusioncatcher recipe to pin them to a specific, conda-installed version. Then if the user tries to update, say, bowtie2 to a newer version then the conda solver will raise an error because that would not be compatible with the pinned environment.

Ideally you shouldn't even need the installation script anymore since conda will handle the dependencies and build the environment in an isolated location.

PR #3995, opened 2017-03-02
r-plasmidprofiler: pandoc as requirement, required versions

JenCabral:


recipes/r-plasmidprofiler/meta.yaml
```diff
@@ -17,38 +17,40 @@ build:
 
 requirements:
   build:
-    - r
-    - r-ape
-    - r-dplyr
-    - r-gdata
-    - r-ggdendro
+    - r 3.3.1
```
bgruening: Is pinning the R version really needed?

JenCabral: We have tested these package combinations and found it to be stable with these versions. We want to pin them to ensure the package continues to work in the future

PR #4003, opened 2017-03-03
Adding Sniffles (SV variant detection from long reads)

holtgrewe:

### DIFF
--- a/recipes/sniffles/meta.yaml
+++ b/recipes/sniffles/meta.yaml
@@ -1,0 +1,36 @@
+package:
+  name: sniffles
+  version: '1.0.3'
+
+source:
+  fn: Sniffles-v1.0.3.tar.gz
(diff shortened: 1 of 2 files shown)


recipes/sniffles/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+  md5: "0eaffafb89f7fd00593f7e2eb29d8505"
+
+build:
+  # No OpenMP on Os X...
+  skip: True  #[osx]
+  number: 0
+
+requirements:
+  build:
+    - gcc	>=4.8.2   # [linux]
```
bgruening: Can you remove this exact pinning of gcc or is this strictly needed?

holtgrewe: Fixed

PR #4015, opened 2017-03-03
Add compiler spec, OS X build for beagle-lib.

acaprez:


bgruening:
Urgs, sorry this test does not work with the new busybox testing as no conda is installed in this test containers. Can you come up with a test that does not depend on the build environment?

acaprez:
Yep, I changed it to check for the shared library via `test -f`, similar to other library-only packages.  Thanks for the help debugging this!

PR #4042, opened 2017-03-06
Hotfix

daler:

Default for travis-run and simulate-travis is now to check git diff against HEAD.

Also makes whitespace changes to delly and soapdenovo-trans to trigger builds that were missed from the bioconda-utils bugs.

Note that this temporarily uses the hotfix2 branch of bioconda-utils as a double-check on that PR. Once that PR and this PR are merged I'll submit a second PR to point to bioconda-utils master.

cc @acaprez @johanneskoester


daler:
@johanneskoester looks like delly is still having boost issues?

johanneskoester:
@daler I don't really understand why I did not see the error in my pull request. However, I hope to fix it in PR #4045.

PR #4047, opened 2017-03-07
Pin the old mothur-1.36.1 to old boost version

johanneskoester:

According to @bgruening it does not work with the newer version. Hence, I hardcoded the latest working version into this recipe. This is reasonable since it is just an old version of the recipe that won't be updated.

ping @bgruening @bebatut @shiltemann


johanneskoester:
@daler, this is weird. I have increased the build number, but still bioconda-utils tells me nothing to be done: https://travis-ci.org/bioconda/bioconda-recipes/jobs/208500204#L474

PR #4064, opened 2017-03-08
Using older CPU architecture for minialign

holtgrewe:

@ocxtal I am using an older platform string here. The `-march=native` makes the built package use features that some not bleeding edge CPUs (e.g., our cluster with `Intel(R) Xeon(R) CPU E5-2650 v2 @ 2.60GHz` CPUs) do not support. This is a *proposal* only.

@bgruening @johanneskoester is there a policy of backwards compatibility in terms of CPU instruction set?

### DIFF
--- a/recipes/minialign/build.sh
+++ b/recipes/minialign/build.sh
@@ -4,6 +4,7 @@ export LDFLAGS="-L$PREFIX/lib"
 export CPATH=${PREFIX}/include
 
 mkdir -p $PREFIX/bin
+sed -i -e 's/march=native/march=corei7/g' Makefile
 make
 make install PREFIX=$PREFIX
 
--- a/recipes/minialign/build.sh
+++ b/recipes/minialign/build.sh
@@ -4,7 +4,7 @@ export LDFLAGS="-L$PREFIX/lib"
 export CPATH=${PREFIX}/include
 
 mkdir -p $PREFIX/bin
-sed -i -e 's/march=native/march=corei7/g' Makefile
+sed -i -e 's/-march=native/-msse4.2 -mpopcnt/g' Makefile
 make
 make install PREFIX=$PREFIX
 
--- a/recipes/minialign/meta.yaml
+++ b/recipes/minialign/meta.yaml
@@ -6,7 +6,7 @@ package:
   version: {{ version }}
 
 build:
-  number: 0
+  number: 1
 
 source:
   fn: {{ name|lower }}_{{ version }}.tar.gz


kyleabeauchamp:
I work with some E5-2630 v2s, so presumably this helps me as well.

ocxtal:
Hi there,

I guess that the precompiled binary, which is compiled on an AVX2-enabled machine and installed with bioconda, fails to run on the Xeon E5-2650 v2 (Ivy Bridge generation) processor with some `illegal instruction` message. If it is the case, the actual reason is the unsupported AVX2 instructions, which is introduced from the Haswell (2013~) generation.

Considering the objective of the bioconda framework, it seems better if a single binary run on broader range of machines even if it cannot utilize the new AVX2 instructions. It is also a good choice for users of some clusters which is composed of machines with different generations with sharing the same `bin` directory over a network filesystem.

Generally, the AVX2-enabled binary runs 20~30% faster than the non-AVX2 one on Haswell and Skylake processors. The difference is slightly large but not fatal, I propose that the precompiled binary in the bioconda repository support for older processors. (It is time to build your local binary if you prefer the better performance to the handiness of the precompiled one.) And I also suggest that the supported microarchitecture should be Nehalem and later, which provide SSE4.2 and POPCNT instructions both of which are required in the dynamic programming routines.

Summary of my proposal:
* support for Nehalem and later processors by the bioconda precompiled-binary repository
* change `-march=native` to `-msse4.2 -mpopcnt` in the `build.sh` script

And here is a patch:
https://github.com/ocxtal/bioconda-recipes/commit/a828c0321f19dde364768a54dbea7c8112f202ec

Regards,

Hajime Suzuki

holtgrewe:
@ocxtal Thanks a lot for this broader background. As a side note, Haswell was not available from the large vendors for server hardware (at least for us) until 2014.

@bioconda/core I think this is something where a general policy ruling would be good for the project: what is the oldest CPU architecture that you want to support and what are the time lines (e.g. in ten years, requiring AVX2 is probably OK).

johanneskoester:
@holtgrewe indeed, I was about to say the same.
@jakirkham and @msarahan, since I could not find it documented: may I ask whether continuum and conda-forge ensure a certain compiler optimization target?

ocxtal:
I'm sorry for my confusing expression, but my proposal on the supported architecture only focus on the minialign program. I think it is fair for minialign to limit its supported architecture to the Nehalem and later, but having no idea about other programs.

By the way, I also agree to determine a policy about compiler optimization flags. There are a lot of programs and libraries emerging which explicitly utilize the latest instructions (for example, the [parasail library](https://github.com/jeffdaily/parasail) will replace the current widely-used [SSW library](https://github.com/mengyao/Complete-Striped-Smith-Waterman-Library), the former supports for SSE2, AVX2, and AVX-512 but the latter supports only for SSE2). This is just what I have noticed and felt about this problem.

Thank you,

Hajime Suzuki

johanneskoester:
Thanks, @msarahan! So, basically when using the new compiler packages the problem would be solved. Will it become available in the next few weeks?

msarahan:
I would hesitate to call it solved until the runtime specification of variants exists and is in general use.

The compiler packages should be available soon, though.  We're working out the kinks - currently investigating a segfault that occurs when legacy xz (from defaults) is used at the same time as our compiler package.  Fun times.

The other side of rolling out the new compiler packages is that we'll also need to recompile all of the packages that use C++, because the new compiler packages will switch to the new ABI.  We're still working out how to most efficiently do that internally.

johanneskoester:
Ok, so that means in turn that bioconda and conda-forge need to recompile at the same time, right? Even if packages are pinned to a minor version, they won't work together, right?

I guess one option would be that all three channels (or 4 including R), upload their rebuilds into a separate channel. Once everything on all channels has been rebuilt, the packages can be copied over. Does that make sense, or is there a simpler solution?

msarahan:
yes, I recommend doing this concurrently with rollout of conda-build 3, so that any package with a hash is immediately recognizable as a "new" C++ ABI package.

johanneskoester:
@msarahan to clarify:
1. upgrade to conda-build 3
2. Depend on new compiler package
3. rebuild packages with conda-build 3

However, what happens in this example:
- conda-forge has a package A, that has been already rebuild
- bioconda has a package B that depends on A, but it has not yet been rebuild, and only pins agains a version.
I guess this would lead to a problem with binary incompatibility, because B will pull the new build of A, right?

Hence, would the following procedure help?
1. upgrade to conda-build 3
2. rebuild with conda-build 3
3. Depend on new compiler package
4. rebuild with conda-build 3

holtgrewe:
OK, @ocxtal I'll just merge this for now? We don't need to solve the big problem to get this ticket closed, I guess ;)

ocxtal:
I also think so 👍

Hajime Suzuki

PR #4091, opened 2017-03-09
Hmmer2

zachcp:

This pull request installs hmmer2 by appending a "2" to all executables. This allows the simultaneous installation of hmmer2 and hmmer3.  This change is desirable because the API/ABI of hmmer2/3 is not identical. Currently there is no way to use both HMMER2 and HMMER3 at the same time. The impetus to include this package in this manner is to facilitate the packaging of the Antismash program that utilizes this convention. Because it will not clash with the HMMMER program I believe it will be okay to use in this way but I will file an issue asking for suggestions/criticism.

Thanks,
zach cp

### DIFF
--- a/recipes/hmmer2/Makefile.in.patch
+++ b/recipes/hmmer2/Makefile.in.patch
@@ -1,11 +1,11 @@
---- Makefile.in.bak	2017-03-09 11:49:34.000000000 -0500
-+++ Makefile.in	2017-03-09 12:06:52.000000000 -0500
+--- Makefile.in
++++ Makefile.in
 @@ -112,7 +112,7 @@
  	mkdir -p ${BINDIR}
  	-mkdir -p ${MANDIR}/man${MANSUFFIX}
  	for file in $(PROGS) $(PVMPROGS); do\
--	   cp src/$$file $(BINDIR)/$$file;\
-+	   cp src/$$file $(BINDIR)/$${file//2};\
+-	   cp src/$$file $(BINDIR)/;\
++	   cp src/$$file $(BINDIR)/"$${file}2";\
  	done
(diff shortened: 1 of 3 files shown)


peterjc:
This looks good in principle - I have a Galaxy tool where I'd explicitly like to use HMMER2 and assume the binary is available as ``hmmsearch2``, so ought to be able to use your BioConda package 👍

johanneskoester:
I must say that I am not entirely convinced. Since conda supports isolated environments, there is not really a point of doing this in order to be able to use both versions. It is also very unusual in the conda ecosystem. E.g. we don't do that for the different python versions and so on.

peterjc:
I'm not sure Python is a good counter example given Python 3 introduced the convention ``python3``, ``pip3`` etc for the default version of Python 3, rather than Python 2.

johanneskoester:
Yes, it did, but not in Conda. This was done to have both versions on systems that don't have a package system that is able to install multiple versions into isolated environments. In conda, this is not needed. And hence, the conda python package also provides python 3 simply as `python`.

zachcp:
my primary use case was to support Antismash which uses both hmmer2 and hmmer (hmmer3) at the same time.  Basically if both binaries are required on the same system they are essentially two different programs. Its a fundamental problem. Maybe a better comparison would be bowtie vs bowtie2.

johanneskoester:
Indeed, that's a different story, sorry I should've read your post more carefully. So, if antismash will depend on both, we indeed would have to do something like this. In that case however, this new package has to be named hmmer2. Otherwise conda will refuse to install both.

zachcp:
thanks for your  time/review @johanneskoester . I changed the name to hmmer2 and will pull if/when the builds pass.

PR #4099, opened 2017-03-10
Update sina build

mdehollander:

### DIFF
--- a/recipes/sina/meta.yaml
+++ b/recipes/sina/meta.yaml
@@ -16,8 +16,8 @@ source:
   url: ftp://ftp.arb-silva.de/SINA/builds/2017/build-1163/sina-1.3.0-r24332.ubuntu1404-amd64.tgz
 
 run:
-    - boost {{CONDA_BOOST}}*
- 
+    - boost # [linux]
+
 test:
     commands:
         - sina --help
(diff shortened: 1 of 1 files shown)


mdehollander:
How to have libboost_program_options.so.1.54.0 included during runtime?

acaprez:
Hmm, it looks like the tarball is a binary build for Ubuntu 14.04?  The boost linking issues aside, that won't work in the EL6-based condaforge image.  The glibc version there is older and will cause a version error at runtime.  You might be able to some some rpath manipulation magic to get it to find the boost libs, but that's iffy IMHO and it still wouldn't run because of the glibc issue.  The only reasonable option here really is to build from source IMO.

mdehollander:
@acaprez, I think you are right, thanks for the explanation. I will see if I can work with the developer @epruesse to make a package from source.

epruesse:
I agree, building from source is preferable. There are some issues to be fixed first though.

For the time being: there are centos5 and centos6 builds at the same location, maybe those will help.

@acaprez How would you satisfy a dependency on openmotif within conda? SINA depends on ARB which in turn requires openmotif to build fully. Even if openmotif is present on the build hosts (I would be surprised if it is), it won't be available on many client installations.

acaprez:
Ideally both those would be added as conda packages first; then it wouldn't be an issue.  Unfortunately, getting packages which require GUI libraries to work in conda can be quite challenging.  Conda-forge did add a lot of X11 packages fairly recently though, so it might not be too bad at this point.

mdehollander:
It seems ARB and its dependencies (https://github.com/epruesse/SINA/issues/8) are now in conda, so it should be possible to build a package from source

epruesse:
@mdehollander Not quite, epruesse/SINA#8 isn't done yet. `openmotif` is just now complete for linux, hopefully osx will build soon. After that I can work on getting ARB into bioconda, and then I can start messing with SINA...

PR #4100, opened 2017-03-10
Added select sequnces recipe

eggzilla:


bgruening:
@eggzilla this name seems a little bit to generic isn't it?

eggzilla:
That is true, however the naming is derived from a RNAz script with the name selectSequences.pl. I could go for SelectAlignmentSequences or RNAcastingshow?

PR #4104, opened 2017-03-10
Anaquin

sebastian-luna-valero:

Followed installation instructions:
https://s3.amazonaws.com/sequins/software/Anaquin_CompileFromSource_v1.pdf

### DIFF
--- a/recipes/anaquin/build.sh
+++ b/recipes/anaquin/build.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+sed -i.bak 's/^CPPFLAGS =$//g' Makefile
+sed -i.bak 's/^LDFLAGS  =$//g' Makefile
+
+# varfilter.py in install fails because we don't install Python
+sed -i.bak 's#misc/varfilter.py##g' Makefile
+
+# Remove rdynamic which can cause build issues on OSX
+# https://sourceforge.net/p/samtools/mailman/message/34699333/
+sed -i.bak 's/ -rdynamic//g' Makefile
+
+export CPPFLAGS="-I$PREFIX/include"
+export LDFLAGS="-L$PREFIX/lib"
+
+./configure --prefix=$PREFIX --enable-libcurl --without-curses
+make install prefix=$PREFIX LIBS+=-lcrypto LIBS+=-lcurl
--- a/recipes/anaquin/build.sh
+++ b/recipes/anaquin/build.sh
@@ -1,17 +1,36 @@
 #!/bin/bash
+set -eu -o pipefail
 
-sed -i.bak 's/^CPPFLAGS =$//g' Makefile
-sed -i.bak 's/^LDFLAGS  =$//g' Makefile
+# download and copy required dependencies
+# klib
+wget https://github.com/attractivechaos/klib/archive/spawn-final.tar.gz
+tar xzf spawn-final.tar.gz
+cp -r $SRC_DIR/klib-spawn-final $PREFIX/lib/klib/
 
-# varfilter.py in install fails because we don't install Python
-sed -i.bak 's#misc/varfilter.py##g' Makefile
+# vcflib
+wget https://github.com/student-t/vcflib-forAnaquin/archive/v1.0.0-rc1.tar.gz
+tar xzf v1.0.0-rc1.tar.gz
+cp -r $SRC_DIR/vcflib-forAnaquin-1.0.0-rc1/src $PREFIX/lib/vcflib/
 
-# Remove rdynamic which can cause build issues on OSX
-# https://sourceforge.net/p/samtools/mailman/message/34699333/
-sed -i.bak 's/ -rdynamic//g' Makefile
+# clean Makefile
+sed -i'' -e 's/^BOOST.*//g' Makefile
+sed -i'' -e 's/^EIGEN.*//g' Makefile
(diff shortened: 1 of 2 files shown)


sebastian-luna-valero:
Hi @bioconda/core 

I would appreciate another pair of eyes looking into this new recipe.

Best regards,
Sebastian

recipes/anaquin/build.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+set -eu -o pipefail
+
+# download and copy required dependencies
+# klib
+wget https://github.com/attractivechaos/klib/archive/spawn-final.tar.gz
```
bgruening: If you can it would be great to put this into a dependency, or at least get a fixed version of it.

recipes/anaquin/build.sh
```diff
@@ -0,0 +1,36 @@
...
+tar xzf spawn-final.tar.gz
+cp -r $SRC_DIR/klib-spawn-final $PREFIX/lib/klib/
+
+# vcflib
+wget https://github.com/student-t/vcflib-forAnaquin/archive/v1.0.0-rc1.tar.gz
+tar xzf v1.0.0-rc1.tar.gz
```
bgruening: Same here.

recipes/anaquin/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  fn: {{version}}.tar.gz
+  url: https://github.com/student-t/Anaquin/archive/{{version}}.tar.gz
+  md5: c989314e15198a49b6731d82e23b3272
+
+requirements:
+  build:
+  - gcc  # [not osx]
+  - llvm # [osx]
+  - boost {{CONDA_BOOST}}*
+  - eigen 3.3.1
```
bgruening: We need to pin this globally as well I think. But we can do this later.

recipes/anaquin/build.sh
sebastian-luna-valero: This is a pure library (i.e. no binaries):
https://github.com/attractivechaos/klib

Please note that I am downloading a github released tarball (i.e. a fixed version).

Moreover, it has not got updates since 2013 so I understood this is a very one off dependency. Do you think it is still good to create a bioconda package for it?

recipes/anaquin/build.sh
sebastian-luna-valero: vcflib is also available in bioconda:
https://github.com/bioconda/bioconda-recipes/tree/master/recipes/vcflib

but it only provides the binaries (which is sensible):
https://github.com/bioconda/bioconda-recipes/blob/master/recipes/vcflib/build.sh

Also, the author of anaquin modified the source code to work with it, as explained here:
https://s3.amazonaws.com/sequins/software/Anaquin_CompileFromSource_v1.pdf

so not sure what to do here?

recipes/anaquin/meta.yaml
sebastian-luna-valero: Thanks.

recipes/anaquin/build.sh
student-t: Hi. I'm the author of the package. vcflib must be modified otherwise it just exit() the program whenever it encounters an invalid VCF file. I made it so that it throw exception, thus I can catch it to do something.

student-t: I think I can embed the VCF parsing code from vcflib directly into Anaquin if this is the easiest solution.

recipes/anaquin/build.sh
student-t: Anaquin only requires random generator from klib, and it's a header file. I can directly embed the file into Anaquin (so no external dependency) if that's the easiest solution.

student-t: klib is a header-only library, so you can't compile it directly.

recipes/anaquin/build.sh
bgruening: @student-t offering a release tarball with everything included would be great. We keeping backups of all these release tarballs to make it a little bit more reproducible. Thanks!

jerowe: @student-t, You could split each 'wget' call into a separate recipe, and just add them to the build/run deps.

student-t: OK. I will modify the source code and minimise the external dependencies.

sebastian-luna-valero:
Thanks @student-t @bgruening @jerowe for your comments!

I will be waiting for @student-t to update the source code of Anaquin and then adapt the recipe accordingly. Please ping me when that is the case.

student-t:
@sebastian-luna-valero I added klib to the library and fixed up the references for vcflib. I think that's enough for the build?

sebastian-luna-valero:
Thanks @student-t 

Could you please tag/release the new version of the code:
https://github.com/student-t/Anaquin/releases

So we can target that on the `recipe`?
https://github.com/bioconda/bioconda-recipes/blob/071b7fdc4bb74f21a92eb8327846748b7bea3b4d/recipes/anaquin/meta.yaml#L14

Best regards,
Sebastian

student-t:
I have some bug fixes I want to address. I will create a tag in a day or two. Thanks!

sebastian-luna-valero:
@student-t could you please confirm whether a new tag/release is ready for this recipe to go ahead?

johanneskoester:

PR #4130, opened 2017-03-14
Prokka

Slugger70:

Something goes wrong with bioperl during docker build in Travis.

### DIFF
--- a/recipes/prokka/build.sh
+++ b/recipes/prokka/build.sh
@@ -1,10 +1,11 @@
 #!/bin/sh
 set -e
 
-sleep 5
-
 ./bin/prokka --setupdb
 mkdir -p "${PREFIX}/bin" "${PREFIX}/db" "${PREFIX}/share/doc/prokka"
 mv bin/* "${PREFIX}/bin/"
 mv db/* "${PREFIX}/db/"
 mv doc/* "${PREFIX}/share/doc/prokka/"
+
+prokka --listdb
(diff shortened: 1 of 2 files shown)


bgruening:
@jerowe do you have any idea what is going on here? Do we need finally a bioperl 1.7?

jerowe:
Let me take a look.

jerowe:
The new logs ar eso pretty!

jerowe:
It looks like it wants a particular version of makeblastdb

https://travis-ci.org/bioconda/bioconda-recipes/jobs/210862803#L409

I don't know if you can hold off on perl things, but I am working a lot on the perl conda skeleton creator. My goal is to be able to build Moose and BioPerl in one go from the skeletons. I'm nearly there, but it will take some more time.

jerowe:
That reminds me, are we using the gcc available on the conda-forge docker image, or do we need to explicitly add gcc in the build deps?

If we are not using the conda-forge gcc any perl modules that require cpanm to install a perl module that has a Makefile will not work.

bgruening:
For perl is was ok to use the conda-forge GCC in the hope that we get a new packaged GCC soon.

bgruening:
@jerowe the makeblast error is indeed a OSX problem. The Linux problem is here: https://travis-ci.org/bioconda/bioconda-recipes/jobs/210862802#L1254

PR #4131, opened 2017-03-14
Adding recipe for bpipe

holtgrewe:
Pinning Java is required here as the defaults java is broken (at least
for me locally, there is a problem finding libnet.so through java NIO)

### DIFF
--- a/recipes/bpipe/meta.yaml
+++ b/recipes/bpipe/meta.yaml
@@ -1,0 +1,28 @@
+{% set name = "bpipe" %}
+{% set version = "0.9.9.2" %}
+
+about:
+  home: 'http://docs.bpipe.org/'
+  license: BSD-3-clause
+  summary: Bpipe - a tool for running and managing bioinformatics pipelines
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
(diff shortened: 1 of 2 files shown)


recipes/bpipe/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+  url: http://download.bpipe.org/versions/{{ name }}-{{ version }}.tar.gz
+  sha256: '29986da8dfb89d9789c25d3bf7e3486acfd99eb1fbac887e7f3aebc8e07ee471'
+
+build:
+  number: 0
+
+requirements:
+  run:
+  # The Zulu OpenJDK from conda-forge is pinned as the defaults one is broken (libnio.so...)
+  - openjdk ==8.0.112
```
bgruening: Can you pin it like this: `openjdk 8.0* zulu8*` this gives us more flexibility.

holtgrewe: OK

daler: @bgruening should we always do this? That is, should we add it to the linting documentation or the java guidelines?

bgruening: I actually had hoped that `default` and `conda-forge` will fix this and can agree on one package. But if this is not the case we should pin ours to conda-forge. So yes adding it guidelines makes sense.
For the linting part I guess it would make sense to pin this with your pinning schema, isn't it?

PR #4118, opened 2017-03-12
Plastid 0.4.7

lparsons:
* [x] I have read the guidelines above.
* [x] This PR adds a new recipe.
* [ ] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

PR #4141, opened 2017-03-14
Fix pysam shared library rpaths for OS X.

acaprez:

xref:  #4118


kyleabeauchamp:
Is this something we should propose upstream?  I ask because the next pysam release is being worked on this weekend: https://github.com/pysam-developers/pysam/issues/413

acaprez:
If that's something they'd be willing to support sure, that would be great.  It wouldn't be terribly involved I believe.  Basically, just check for `CONDA_BUILD=1` in the environment on OS X and tweak the `ext.extra_link_args` list appropriately.

PR #4144, opened 2017-03-15
Upgrading optitype to commit

holtgrewe:
Sadly, there is no alternative until a proper release >1.2 is done
upstream.


holtgrewe:
Hm, the build system does not like this either for some reason. OK, I'll add this to our local repository and wait until upstream fixes this

PR #4148, opened 2017-03-15
new recipe

hardingnj:

Adds IBDseq tool of Browning
(excuse typo in branch name)

### DIFF
--- a/recipes/ibdseq/ibdseq
+++ b/recipes/ibdseq/ibdseq
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+# Wraps ibdseq.jar
+#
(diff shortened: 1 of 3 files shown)


recipes/ibdseq/ibdseq
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+# Wraps ibdseq.jar
+#
+# This is copied from https://github.com/chapmanb/gatk
+#
+set -o pipefail
```
bgruening: There is a python version of this wrapper that we slightly prefer over the bash one. Maybe you can use this for the next release.

PR #2657, opened 2016-10-20
nextflow 0.21.3 -> 0.23.4

odoublewen:
- [x] I have read the guidelines above.
- [ ] This PR adds a new recipe.
- [x] This PR updates an existing recipe.
- [ ] This PR does something else (explain below).

PR #4158, opened 2017-03-16
updated recipe for nextflow 0.23.4

odoublewen:

This PR replaces PR #2657

### DIFF
--- a/recipes/nextflow/0.21.3/build.sh
+++ b/recipes/nextflow/0.21.3/build.sh
@@ -1,0 +1,3 @@
+mkdir -p $PREFIX/bin
+chmod 755 nextflow
+cp nextflow $PREFIX/bin
--- a/recipes/nextflow/0.21.3/meta.yaml
+++ b/recipes/nextflow/0.21.3/meta.yaml
@@ -1,18 +1,18 @@
 package:
   name: nextflow
-  version: "0.23.4"
+  version: "0.21.3"
 
 build:
   number: 0
 
 source:
   fn: nextflow
-  url: https://github.com/nextflow-io/nextflow/releases/download/v0.23.4/nextflow
-  md5: ace2aebfc7b18be1f4605ff803ccbc9c
+  url: https://github.com/nextflow-io/nextflow/releases/download/v0.21.3/nextflow
+  md5: 241e4bd48eb9ef6af31e22002ef8b0eb
 
 requirements:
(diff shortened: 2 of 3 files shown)


bgruening:
@odoublewen do you know what this is?
https://travis-ci.org/bioconda/bioconda-recipes/jobs/211616089#L1627

Is wget a requirement?

odoublewen:
Yes, good point, this looks like an undocumented requirement (either curl or wget, see below).

However, this isn't the direct cause of the error.  The docker container actually does have wget -- see the error message is providing usage help page for wget.  The actual cause of the error is the line above, where it say `mktemp: invalid argument`

Here are the relevant lines from the installer script:

```
function install() {
    local tmpfile=$(make_temp)
    get "http://www.nextflow.io/releases/latest/nextflow" "$tmpfile" "$1" || exit $?
...
```
and:
```
function make_temp() {
    local base=${NXF_TEMP:=$PWD}
    if [ "$(uname)" = 'Darwin' ]; then mktemp ${base}/nxf-tmp.XXXXX || exit $?
    else mktemp -t nxf-tmp.XXXXX -p ${base} || exit $?
    fi
}
```
and:
```
function get() {
    if command -v wget &>/dev/null; then
        GET="wget -q $1 -O $2"
    elif command -v curl &>/dev/null; then
        GET="curl -fsSL $1 -o $2"
    else
        echo_red "Error: cannot find 'curl' nor 'wget' utility --  please install one of them"
        exit 1
    fi
...
```

I'll ask the author, @pditommaso

odoublewen:
Fundamentally, this is because the busybox container has a mktemp that doesn't have the -p option (it is more like the darwin mktemp)

odoublewen:
The author [says](https://github.com/nextflow-io/nextflow/issues/305#issuecomment-287092648) that the package requires Bash (this was undocumented) and GNU mktemp.

Easiest solution would be if we could run the CI tests with a fatter docker image, like debian jessie.

bgruening:
@odoublewen but this is exactly the error we want to catch. You can depend on `coreutils` to satisfy the dependencies.

odoublewen:
Of course.  Makes total sense.  Thanks for reminding me of `coreutils`

I think this recipe PR is ready to be merged.

bgruening:
Nice one! Thanks @odoublewen!

PR #4159, opened 2017-03-16
Remove deprecated doc infrastructure.

johanneskoester:

@daler am I right that, since docs are generated from bioconda-utils, we can now remove the doc infrastructure in bioconda-recipes?


daler:
That's right, and thanks for cleaning this up!

PR #4162, opened 2017-03-16
newer version proBAMconvert

vladie0:
My first attempt to modify the conda package for proBAMconvert, i'm still pretty amateur at this.
I've updated to a newer proBAMconvert version, compatible with pysam 0.10.0, also, the previous version did have some errors in galaxy which should be fixed now.

Also, I'm no quite sure but should the usage of python2 not be a requirement ? proBAMconvert is not made compatible with python 3.

Also, is there away to adjust the galaxy xml file to use a fork of the bioconda recipe so I could test it out before requesting a pull?

Thank you for all your work, I'm still in the progress of figuring Conda out !

### DIFF
--- a/recipes/probamconvert/meta.yaml
+++ b/recipes/probamconvert/meta.yaml
@@ -19,7 +19,7 @@ requirements:
   run:
     - python
     - cogent
-    - pysam > 0.9
+    - pysam >=0.9
     - mysql-python
     - lxml
     - numpy
--- a/recipes/probamconvert/meta.yaml
+++ b/recipes/probamconvert/meta.yaml
@@ -8,8 +8,8 @@ about:
   license: Apache License
 
 source:
-  url: https://github.com/Biobix/proBAMconvert/archive/0504665126486b4d4a03ce863c447891dc051ee7.tar.gz
-  md5: 7646cbbc70333517850453c678ccbe6c
+  url: https://github.com/Biobix/proBAMconvert/archive/6e9080412b500bcb52aa0576afc50575a25fb4f3.tar.gz
+  md5: d3cbf80f8ce900c85b26a90c9cd220c1
 
 build:
   number: 0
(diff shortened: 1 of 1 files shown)


recipes/probamconvert/meta.yaml
```diff
@@ -19,7 +19,7 @@ requirements:
   run:
     - python
     - cogent
-    - pysam ==0.9
+    - pysam > 0.9
```
bgruening: This needs to be an `>=0.9` without a space.

bgruening:
If you drop the py3k from the skipping build step it will be build for python3. I just saw a few print statements in your code without brackets and assumed it is python2 only. Sorry for that!

vladie0:
It is indeed intented for python2 usage only. 
Comming back to my previous version, is there a way to adjust the galaxy configuration xml to use a forked bioconda recipe ?

bgruening:
What means forked? If you have a version in your own channel you can add this channel to the Galaxy config.

bgruening:
Thanks for the update!

jegalle:
Hi Bjoern

First of all thanks for the help with this conda package, it is actually the first time we'll push something to conda. (but definitely not the last time :) )

So vladie's pull request is merged, but when I try to install probamconvert trough conda:

`conda install -c bioconda probamconvert`

I still get the version from before the pull request, in  probamconvert-1.0.0-py27_0.json I still have "pysam ==0.9". 

Is there something else that needs to be done before it will install the new version?

Thanks in advance!

Jeroen

bgruening:
Oh, I see. This is my fault. I haven't see that one piece was missing in this PR.
You always need to increase the version number or build number to trigger a new build.

Ideally we need real releases from proBAMconvert and use a real version number here.

See also my PR where I fixed it: https://github.com/bioconda/bioconda-recipes/pull/4177
Should be available in a few minutes.

jegalle:
Ok, thanks! We'll keep that in mind for future releases.

PR #4191, opened 2017-03-17
Update setup.py

ThomasWollmann:


bgruening:
@ThomasWollmann what ever was broken, can we add a test for this?

ThomasWollmann:
Not all modules were imported, therefore I added automatic module discovery, since nobody wants to list all the modules manually. I'll add checks for important modules.

PR #4204, opened 2017-03-18
remove jpeg from bioconda

bgruening:


bgruening:
I hope this does not break to many recipes, but it should not be here.

PR #4208, opened 2017-03-18
Salmon 0.8.2

rob-p:

Update to Salmon 0.8.2, and build from source (which should work on both linux and macOS now)


bgruening:
Cool, thanks @rob-p!

rob-p:
Sure;  I haven't done a bioconda PR in a while, so let me know if I missed anything!

bgruening:
Looks good to me, let's see what travis say. There is no need to remove the old recipe. Some people want to keep it to maintain also old versions.

bgruening:
It's failing on OSX with: `dyld: Symbol not found: __ZNK5boost15program_options22error_with_option_name23substitute_placeholdersERKNSt3__112basic_stringIcNS2_11char_traitsIcEENS2_9allocatorIcEEEE`

rob-p:
Interesting.  All the Travis CI tests passed in my branch (including the OSX one).  Is there a reason they might be different.

bgruening:
Not we you have the latest master on your branch as well. Wired. But we can restart the travis test.

rob-p:
[Here](https://travis-ci.org/rob-p/bioconda-recipes) are my tests, any idea what might be different here?

rob-p:
Ok, now I can repro the issue --- I'll see if I can figure out what the hell is going on!

rob-p:
@bgruening --- do you know if the boost libraries on OS X are built with clang or with gcc?  Also, do you know if they're built with c++11 flags?  To use the vendored boost, it should be built with the same compiler (clang++ on OS X) and with c++11 support.

bgruening:
I don't know for sure, but it looks like that the host compiler was used: https://github.com/conda/conda-recipes/blob/master/boost/meta.yaml
At least no conda package was used as compiler.

rob-p:
@bgruening --- Sweet; it looks like requiring a specific version (1.63.0) of boost on macOS solves the problem.  All the builds seem to pass now.

bgruening:
@rob-p thanks a bunch!

rob-p:
@bgruening Sure; happy to have Salmon in bioconda!

bgruening:
And here is your rkt and Docker container, kindly provided by bioconda/biocontainers :)
https://quay.io/repository/biocontainers/salmon?tab=tags

rob-p:
awesome!

chapmanb:
Rob;
Thanks so much for working on this. I'd switched over in Salmon 0.8.1 to use the pre-compiled binaries (https://github.com/bioconda/bioconda-recipes/pull/4163) as a way to avoid the `icu 56.*` dependency. That seems to be a contentious package and causes other bioconda packages like `lxml` to break because they require `icu 58.*`. Do you think using the binaries is an option? Or is there a way to avoid this strict dependence on icu 56? Thanks again.

rob-p:
@chapmanb, Hi Brad,

  I actually am not sure if there is a strict dependence on this version of icu.  I can try and push a build with icu 58.* and see if it passes all of the tests.  If building from source with icu 58.* works, would that be OK?  The main reason I wanted to build from source is for Mac.  My pre-compiled linux build should be widely usable, but I only have macOS 10.12 machines to compile on, so it would be nice to have Conda build from source for older versions of macOS.  Having source builds for linux is also nice, because I imagine conda has a newer (better) compiler than the ancient one with which I'm forced to build my pre-compiled linux binaries.

--Rob

chapmanb:
Rob -- thanks for looking at this. I couldn't get icu 58 to work because of an incompatibility with autoconf; although I couldn't manage to track down the original source of the incompatibility. If you can do that would be perfect. It might be easier to have a salmon distribution that does not require autoconf to skip trying to sort that out. Hope this helps some and thanks again.

rob-p:
btw, @bgruening / @chapmanb, do either of you know how to force travis to re-build the package?  When I push my changes again, it says:

```
Filtering 2 of 2 (100.0%) recipes/salmon
14:12:36 BIOCONDA INFO Nothing to be done.
+set +x
/home/travis/build.sh: line 162:  3535 Terminated              travis_jigger $! $timeout $cmd
```

even though I bumped the build number.

chapmanb:
It sounds like you have it right -- upping the build number should be enough. Where is the PR and I could look and see if I spot anything?

rob-p:
I got it to trigger --- https://travis-ci.org/rob-p/bioconda-recipes/jobs/212700774.  Unfortunately, now I'm seeing a version conflict that I was not before (perhaps the one you mention)?  That's strange though, because it built fine locally on my linux machine before I did the push.

chapmanb:
That's similar to what I got, although it was with autoconf and icu. I could never figure out what actually caused the underlying incompatibility since it didn't seem like autoconf (or gcc) depends on icu. I don't really understand what is going on with icu but it seems like anyway we can drop/remove the dependency would help a ton as it causes all kidns of weirdness.

rob-p:
So, I tried removing icu from the dependencies.  I get a failed build (locally).  The problem seems to be that the only available boost is compiled with a specific dependency on icu 54.*.  Now, I don't rely on any unicode functionality in Salmon, so I could easily use a boost variant compiled without ICU support, but none seems available.  Perhaps that could be done upstream and then the icu dependency could be removed?

daler:
@chapmanb @rob-p sometimes the version conflicts are incorrectly reported, which gets super confusing: https://github.com/conda/conda/issues/3622. Not sure if that's what's happening here.

chapmanb:
Ryan;
I think that is what's going on since trying to track down the underlying incompatibility with the libraries didn't lead me anywhere. I'm not sure what to do about this now other than work around the issue. For myself, bcbio does have a lot of dependencies and complex relationships so happy to just have everything working cleanly in one environment.

PR #4213, opened 2017-03-19
Upgrade dDocent to v2.2.15

jpuritz:


recipes/ddocent/meta.yaml
```diff
@@ -1,10 +1,10 @@
 package:
   name: ddocent
-  version: '2.2.13'
+  version: '2.2.15'
```
bgruening: You could use variables to make this easier :)

PR #4214, opened 2017-03-19
Jemalloc 4.5.0

rob-p:

I'll note that I was not able to get the test example to work in continuous integration under linux. I always get an error that `gcc` was not found.  This is despite the fact that I list it as a requirement under the `test:` section.  So, I've omitted this test.  Any idea how to fix that?


rob-p:
I moved the old recipies into a `3.6.0` subfolder --- yet this seems to be what is causing the build to fail above.  Should I have handled this in a different way?

rob-p:
I'm going to merge this, but please let me know what I can do to run the test program in CI under linux.

PR #4226, opened 2017-03-21
obitools cleanup

pbordron:

I clean the requirement of the obitools recipes. Some are only required in the build part whereas some others are only required in the run part.


bgruening:
Thanks!

PR #4228, opened 2017-03-21
Specified versions for snippy dependencies

Slugger70:

### DIFF
--- a/recipes/snippy/meta.yaml
+++ b/recipes/snippy/meta.yaml
@@ -12,7 +12,7 @@ build:
 
 requirements:
   run:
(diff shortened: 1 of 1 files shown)


recipes/snippy/meta.yaml
```diff
@@ -8,21 +8,21 @@ source:
...
 requirements:
   run:
-    - perl-threaded
-    - perl-bioperl
-    - perl-list-moreutils
-    - bwa
-    - samtools
-    - parallel
-    - freebayes
+    - perl-threaded >=5.6
```
bgruening: Can you change this to Perl.

PR #4229, opened 2017-03-21
Update: htslib with libhts.so.1 symlink

chapmanb:
Provides back-compatibility. Fixes chapmanb/bcbio-nextgen#1873


kyleabeauchamp:
Are you sure this solves it?  Those people are talking about using HTSLib 1.4 with pysam, which does not yet support 1.4 and bundles / links to a previous version of HTSLib.

chapmanb:
Kyle;
I did a quick test and some general things were okay, but you're right this is a bad fix and we need to wait for full compatibility:

https://github.com/pysam-developers/pysam/issues/413

My fault, I'll restrict to 1.3 for now in bcbio usage rather than pushing this.

PR #4231, opened 2017-03-21
GarNet

zfrenchee:

[GarNet](https://github.com/fraenkel-lab/GarNet) is a tool for multi-omics analysis from the Fraenkel Lab at MIT. Additional docs [here](https://fraenkel-lab.github.io/GarNet/html/index.html).

### DIFF
--- a/recipes/garnet/bld.bat
+++ b/recipes/garnet/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install 
+if errorlevel 1 exit 1
+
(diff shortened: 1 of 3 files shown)


zfrenchee:
@bgruening This package is only for python 3.5 and up. I think that's why it's causing an error. How should I fix this?

bgruening:
You can skip `skip: True  # [py2k or py34]` in your build section if you like.

zfrenchee:
Thanks @bgruening !

PR #2906, opened 2016-11-11
perl-bioperl-core 1.7.1

martin-mann:
* [x] I have read the guidelines above.
* [ ] This PR adds a new recipe.
* [x] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

PR #4241, opened 2017-03-22
Pindel pin deps

johanneskoester:

Add htslib to the list of pinned libraries. Further, I have changed our travis setup such that linting only happens when recipes are filtered based on commits instead of the env matrix. Otherwise, any change in the env matrix leads to linting all recipes in the repository.


johanneskoester:
@bioconda/perl there is a perl-related error. Have you seen this before? Any idea why the bioperl recipe is triggered at all?

martin-mann:
mhh.. no idea why bioperl is triggered..

but seems to be more a travis problem rather than a perl issue?!

I triggered a restart of the build (without much hope).. 

Had faced similar error messages some time ago but for a C++ recipe... but can not recall what (if) solved them. :/

bgruening:
We need to revert this PR: https://github.com/bioconda/bioconda-recipes/pull/2906/files
This is still a regression and the new bioperl broke a lot so we removed it from the channel. I don't have access to a real computer, maybe some of you can revert it?

@johanneskoester I still don't understand why this commit shows up in testing as this recipe was not touched.

johanneskoester:
Because this PR changes the environment matrix. Doing this triggers filtering all recipes based on presence in the channel. Since the package has been deleted from the channel, bioconda-utils will try to build it.

johanneskoester:
Ok, @bgruening, doing this now. Thanks!

PR #4251, opened 2017-03-23
Abricate version 0.5

Slugger70:

This will create a recipe for the recently released abricate v0.5


bgruening:
@druvus can you have a look on this please?

bgruening:
@Slugger70: @druvus has still a dependency on perl and a few more tests included, maybe you can add this as well?

recipes/abricate/meta.yaml
```diff
@@ -17,14 +17,16 @@ requirements:
   - zlib
   run:
   - libgcc # [linux]
-  - perl-threaded
```
druvus: You should probably keep perl as a dependency

druvus:
I added the extra tests to catch missed dependencies needed by helper scripts. Would probably be good to have them but not necessary.

druvus:
The main issue I saw with the 0.3 version of the recipe is the unspecific copying into $PREFIX. This way it is cluttered by unnecessary files (README.md & LICENSE) and two new folders (db & scripts). The script assumes by default (if not set in the prompt) that the database is located `../db` in relation to the script. 

@jerowe @bgruening do you have some recommendation how we could package this in a better way?

bgruening:
ping @Slugger70

tseemann:
@druvus 
if you want to change default `--db` it will need to be patched in `abricate` and `abricate-get_db`.  These are the only 2 scripts now in version 5.0

tseemann:
@Slugger70 maybe add `abricate --check` next time to the test suite.

bgruening:
Let's add it now and either increase the build number to trigger a test or just add it for next time.

tseemann:
Abricate 0.6 was released  - https://github.com/tseemann/abricate/releases

*  need --setupdb before it will work

@Slugger70

PR #4268, opened 2017-03-24
updated version of phylip recipe

bsennblad:
New version of phylip-recipe to build 1. 
(Reason for update: By default, phylip only save a maximum number of phylogenetic trees determined by the hard-coded variable maxtrees. This update applies a patch that changes maxtrees to 10000 in several programs -- previous value of 100 in several programs was way to low for biologically relevant analyses.)


bgruening:
Top @bsennblad!

PR #4270, opened 2017-03-25
nextflow 0.24.1

pditommaso:


pditommaso:
@bioconda/core I've update nextflow recipes to version `0.24.1` however I'm wondering why it contains a folder [0.21.3](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/nextflow/0.21.3) (apparently not used) and if it's safe to delete it?

bgruening:
@pditommaso we keep these folders to also maintain older versions.

pditommaso:
@bgruening I see. So when updating an recipe the previous version should be moved into a sub-folder ?

bgruening:
We do not have a strict guideline here, but yes this is what a few people do to keep multiple version accessible in parallel (without using the git history).

recipes/nextflow/build.sh
```diff
@@ -1,3 +1,4 @@
 mkdir -p $PREFIX/bin
 chmod 755 nextflow
 cp nextflow $PREFIX/bin
+$PREFIX/bin/nextflow -download
```
bgruening: what does this do?

pditommaso: The installed "nextflow" binary is a mere wrapper script that launches a Java application. The `-download` option pulls the required JAR dependencies in the home directory, otherwise they will downloaded the very first time the user run the `nextflow` command. This is needed to pull the application dependencies so that they are included in the nextflow biocontainer created by this recipe (otherwise they will downloaded each time the containers is run making it useless).

bgruening: Is the downloaded JAR versioned?

pditommaso: Yes

bgruening:
Thanks @pditommaso!

pditommaso:
welcome

pditommaso:
Are bioconda recipes supposed to be automatically published as BioContainers? If so why the version provided by this PR is [not available](https://quay.io/repository/biocontainers/nextflow?tab=tags)?

PR #4292, opened 2017-03-29
R gdtools

funnell:

New contributor, please review.

### DIFF
--- a/recipes/r-gdtools/meta.yaml
+++ b/recipes/r-gdtools/meta.yaml
@@ -1,0 +1,79 @@
+{% set name = 'gdtools' %}
+{% set version = '0.1.4' %}
+
+{% set posix = 'm2-' if win else '' %}
+{% set native = 'm2w64-' if win else '' %}
+
+package:
+  name: r-{{ name|lower }}
+  version: {{ version|replace("-", "_") }}
+
+source:
+  fn: {{ name }}_{{ version }}.tar.gz
+  url:
+    - https://cran.r-project.org/src/contrib/{{ name }}_{{ version }}.tar.gz
+    - https://cran.r-project.org/src/contrib/Archive/{{ name }}/{{ name }}_{{ version }}.tar.gz
+
+  md5: 62c2946391c7d1558e32025a6799cd0e
+
+build:
+  # If this is a new build for the same version, increment the build number.
+  number: 0
+
+  # This is required to make R link correctly on Linux.
+  rpaths:
+    - lib/R/lib/
+    - lib/
+
+# Suggests: htmltools, testthat, fontquiver (>= 0.2.0)
+requirements:
+  build:
+    - r-base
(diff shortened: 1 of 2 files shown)


recipes/r-gdtools/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+  run:
+    - r-base
+    - r-rcpp
+    - r-withr
+    - xorg-libxrender
+    - xorg-libxext
+    - xorg-libxdmcp
+    - xorg-libxau
+    - cairo [osx]
```
bgruening: why cairo only on OSX?

recipes/r-gdtools/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+    - lib/R/lib/
+    - lib/
+
+# Suggests: htmltools, testthat, fontquiver (>= 0.2.0)
+requirements:
+  build:
+    - r-base
+    - r-rcpp
+    - r-withr
+    - posix                # [win]
```
bgruening: posix and the native-toolchain is not needed

recipes/r-gdtools/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+    - xorg-libxext
+    - xorg-libxdmcp
+    - xorg-libxau
+    - cairo [osx]
+
+test:
+  commands:
+    # You can put additional test commands to be run here.
+    - $R -e "library('gdtools')"  # [not win]
+    - "\"%R%\" -e \"library('gdtools')\""  # [win]
```
bgruening: win specific lines can also be removed

recipes/r-gdtools/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+  license: GPL-3 | file LICENSE
+  summary: Useful tools for writing vector graphics devices.
+  license_family: GPL3
+
+# The original CRAN metadata for this package was:
+
+# Package: gdtools
+# Version: 0.1.4
+# License: GPL-3 | file LICENSE
+# Title: Utilities for Graphical Rendering
```
bgruening: Can you clean this up and remove not useful comments. Thanks!

recipes/r-gdtools/meta.yaml
funnell: Travis builds seemed to work on linux without it. I only added cairo so that the macOS builds would finish.

bgruening: Mh, but does the tool needs cairo? I suppose for drawing under some environments?

funnell: Looking at this issue (https://github.com/davidgohel/gdtools/issues/24), it seems that cairo is needed on macOS.

bgruening:
Thanks @funnell!

PR #4313, opened 2017-03-30
Default to /usr/local base dir

pvanheus:


recipes/repeatmasker/meta.yaml
```diff
@@ -8,7 +8,7 @@ source:
     patches:
       - patch1
 build:
-  number: 7
+  number: 9
```
nsoranzo: Why skip 8?

PR #4317, opened 2017-03-30
Add bzip2 dep, add test, set boost <= 1.58 on rmblast

pvanheus:

### DIFF
--- a/recipes/bzip2dynamic/meta.yaml
+++ b/recipes/bzip2dynamic/meta.yaml
@@ -2,7 +2,7 @@
 
 
 package:
-  name: bzip2
+  name: bzip2dynamic
   version: {{ version }}
 
 source:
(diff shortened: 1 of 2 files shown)


recipes/bzip2dynamic/meta.yaml
```diff
@@ -0,0 +1,39 @@
+{% set version = "1.0.6" %}
+
+
+package:
+  name: bzip2
+  version: {{ version }}
```
bgruening: the name should always reflect the foldername, but this package already exists in conda-forge. So no need to include it here.

pvanheus:
Ah that's an error. I'll push a new commit when I'm at work.

pvanheus:
Workaround while we wait for https://github.com/conda-forge/bzip2-feedstock/issues/1 to be merged - make a new package bzip2dynamic.

PR #4318, opened 2017-03-30
Expansionhunter update to 2.0.8

maarten-k:


recipes/expansionhunter/meta.yaml
```diff
@@ -8,22 +8,23 @@ package:
   version: {{ version }}
 
 build:
-  number: 0
+  number: 1
```
druvus: Set build number to 0 for new versions.

maarten-k: fixed. (Travis build failed so I changed this field to trigger a new build.)

bgruening:
Thanks!

PR #4320, opened 2017-03-31
added bazel recipe

ThomasWollmann:

### DIFF
--- a/recipes/bazel/build.sh
+++ b/recipes/bazel/build.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
-bash ./compile.sh
+chmod +x ./compile.sh
+chmod +x ./tools/build_rules/gensrcjar.sh
+
+./compile.sh
 
 cp output/bazel $PREFIX/bin/
--- a/recipes/bazel/build.sh
+++ b/recipes/bazel/build.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-scl enable devtoolset-4 bash
-
 bash ./compile.sh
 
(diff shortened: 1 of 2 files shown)


ThomasWollmann:
@bgruening do you know how to enforce .sh execution via bash? Seems like chmod does not work.

bgruening:
Can it be that this file is generated on the fly and had no chmod`ing before?

ThomasWollmann:
maybe ... what is the solution?

bgruening:
patching the compile.sh script?

ThomasWollmann:
Seems a strange solution, since on other systems it compiles flawlessly.

bgruening:
@ThomasWollmann what are these other systems? You compile here on a CentOS-6 and a OSX. Both normal systems I would say.
I would need to look deeper into it, maybe the error is very hidden.

ThomasWollmann:
http://blog.abysm.org/2016/06/building-tensorflow-centos-6/

recipes/bazel/build.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-chmod u+x **/*.sh
+scl enable devtoolset-4 bash
```
bgruening: This is not allowed. If you want to have this, remove the dependency on gcc from your recipe.

ThomasWollmann: I guess this will cause problems with glic.

ThomasWollmann:
https://anaconda.org/conda-forge/bazel

bgruening:
:)

PR #4323, opened 2017-03-31
Add jellyfish 1 as run requirement for kraken

nsoranzo:
It is required for `kraken-build`.
Also add `perl` as build requirement.
Change `perl-threaded` to `perl`.


yesimon:
So this isn't a bug, because `jellyfish v1` is not a runtime requirement of kraken. That's why there's a `kraken-all` metapackage that also asks for `jellyfish v1`. Thjis was on purpose to make it easier to install kraken and jellyfish v2 in the same conda environment. However since nobody has complained, we might as well keep it, but please delete the `kraken-all` metapackage.

PR #4324, opened 2017-03-31
rmblast update

pvanheus:
* add makeblastdb test
* add bzip2 dependency


tseemann:
@pvanheus you added `makeblastdb` as a test. that will fail and return 1. Maybe use `makeblastdb -help` ?

pvanheus:
@tseemann the raw `makeblastdb` was just an attempt to see what it was doing - to see its output. The original test was `makeblastdb 2>&1 |grep nucl >/dev/null` which returns 0 as expect in my testing but doesn't work on Travis.

pvanheus:
The problem is now clear - this is building against the bzip2 package from before https://github.com/conda-forge/bzip2-feedstock/pull/1 was merged. Not sure if it is possible to specify that bzip2 1.0.6  build number > 0 is needed. bzip2 >=1.0.6--1 doesn't work.

bgruening:
`1.0.6_1` will work imho.

recipes/rmblast/meta.yaml
```diff
@@ -10,21 +10,26 @@ source:
...
 build:
-  number: 0
+  number: 1
 
 requirements:
   build:
-    - boost # [linux]
+    - boost <=1.58 # [linux]
     - gcc # [linux]
+    - bzip2 >=1.0.6 1 # [linux]
```
nsoranzo: If you specify a build number, then the version must be exact:
```
    - bzip2 1.0.6 1 # [linux]
```

bgruening:
Everything looks fine, the correct version of bzip2 is installed. It seem that makeblastdb does not find the shared-object from bzip? @pvanheus do you know if makeblastdb is compiled as well in this PR or is this some kind of static binray that is shipped?

pvanheus:
Closing this one so that I can do a cleaner one.

PR #4339, opened 2017-04-03
Add recipe for pprint

bebatut:

### DIFF
--- a/recipes/pprint/meta.yaml
+++ b/recipes/pprint/meta.yaml
@@ -1,0 +1,30 @@
+package:
+  name: pprint
+  version: "0.1"
+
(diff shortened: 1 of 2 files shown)


recipes/pprint/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+  run:
+    - python
+
+test:
+  imports:
+    - helowrld
+
+about:
+  home: https://github.com/hamadi15/test
```
bgruening: Is this spam?

bebatut: Wrong package :(

Issue #1015, opened 2016-03-09
Making urls sustainable


bgruening:
To fix old recipes and to be prepared to rebuild recipes, for example to include zlib or extend builds for OS-X the `url` need to be sustainable.

The Galaxy community has developed several solutions to tackle this problem in several ways.
For example: 

[Bioarchive](https://bioarchive.galaxyproject.org/#/pkg/Biobase/): This one traverses the bioc svn repo and recreates all tarballs that are needed. Some scripts from @daler are using this service already.

[https://github.com/galaxyproject/cargo-port](Cargo-Port): Cargo-Port is a simple tsv file which everyone can extend. After checking the checksum and merging the PR the file will be downloaded, verified and stored.

Both services are storing data in one central place: https://depot.galaxyproject.org

I have talked to the Galaxy team, which is maintaining this storage server, and we agreed that it would be great if bioconda could directly use this storag-backend to make a backup of all needed files.

My idea would be to add a new travis-hook to:
* automatically download the specified URL
* verify it's integrity
* upload it to https://depot.galaxyproject.org
* add a new link to the recipe and commit it

The new recipe should look like this:

```
fn: limma_3.26.7.tar.gz
url:
  - http://bioconductor.org/packages/release/bioc/src/contrib/limma_3.26.7.tar.gz
  - https://depot.galaxyproject.org/bioc/src/limma_3.26.7.tar.gz
```

Problems that needs to be solved.
 * We currently do not have a nice way to fix old recipes that are hidden in git. Here the proposed solution of subdirectories would help a lot.
* It's not guaranteed that we will find every old tarball - bioarchive could help here at least for bioc

Any comments more then welcome!

xref: https://github.com/bioconda/bioconda-recipes/issues/1012

daler:
cargo-port looks fantastic, please extend our thanks to the Galaxy team for storage.

Just to clarify your proposed solution -- I would write this example limma recipe using only the currently-working bioconductor url; upon committing (more specifically, upon merging into master) the travis hook 
would add the second URL as well? Does conda support multiple urls like that?

bgruening:
@daler exactly. I thought that conda support multiple URL's I have seen this in wild. Not sure if it's working. If this is not the case we need to replace the given URL.

johanneskoester:
[...]

PR #4349, opened 2017-04-03
Synergyfinder

mcornwell1957:

This is my second try I uploading this package (the one was last march! #2604 )

I think everything is working well, the only change I had to do was rebuild r-plotrix with R3.3.2 IF THIS WAS WRONG PLEASE LET ME KNOW please do revert those changes.

It was basically the same issue as before, where r-plotrix was still on the 3.3.1 build, but all other dependencies plus the three additional packages I added that synergyfinder requires, all are on the R3.3.2 build.

Does this all sound good?

As far as testing goes, I have DOCKER, but I get a bug when I try and test with travis (both with and without the docker flag)
```
macintoshs-imac:bioconda-recipes mgc31$ ./simulate-travis.py --docker
+ bioconda-utils lint recipes config.yml --git-range HEAD
scripts/travis-run.sh: line 79: bioconda-utils: command not found

```
But I was able to successfully build them all locally

Please let me know if this all works and if there is anything else I need to do. This is my first (hopefully successful package and I want to do it right. Thank you!


daler:
Thanks @mcornwell1957. Many things have changed in the past year, not surprising that building again solve things.

The error you're getting about bioconda-utils is because the linting is not run within the docker container. You'll need to run `simulate-travis.py --install-requirements` locally for this to work.

Can you please:
1. add hashes to your recipes, i.e. https://bioconda.github.io/linting.html#missing-hash
2. remove extra comments in the recipes
3. remove the windows-specific parts of the recipes (i.e. the `set posix`, `set native`; `[win]`, `[not win]`)

recipes/r-nleqslv/meta.yaml
```diff
@@ -13,46 +10,28 @@ source:
...
 
-
 requirements:
   build:
     - r-base
-    - posix                # [win]
-    - {{native}}toolchain  # [win]
-    - gcc                  # [not win]
+    - posix
+    - {{native}}toolchain
```
daler: can remove the posix and native toolchain lines, since they were selected just for windows

mcornwell1957:
Hi @daler How did I do? - sorry for the commits and the git messiness, I think I got that figured out.

Thank you very much for your help - I think I may be doing this a lot, so I want to make sure I am doing this right, so please, let me know what else I can do and if there is anything else I should be doing

Q: with the two paths for R URL, I see that one if for the current package and the other digs into the archive. The archive url does not work (obviously because it was not pushed into archive yet) So does this work in the way that it will attempt the first archive, and then on failure, it tries the second if for example it gets updated on R but we still want our desired version?

daler:
@mcornwell1957 getting there! Still need to remove the windows-specific `"\"%R%\" -e \"library('plotrix')\""` which is what's causing the errors on travis. Actually, might as well clean up the comments in the build.sh files too.

For the URLs, conda tries them in order. The galaxy depot automatically caches tarballs upon successful merge into the master branch, so that's more of a future-proofing step. This is unfortunately not well-documented, but for example see here: https://github.com/bioconda/bioconda-recipes/issues/1015. For the bioconductor recipe you'll at least need the URL to the tarball from the bioconductor site to get it to build.

mcornwell1957:
Yep, I see, so in your original message when you said remove the [win] specific parts, you meant the full lines (obviously, I just wasn't sure, that's on me)

I have now removed everything win related and removed more from the build.sh files (I wasn't sure if we should remove the comments towards the top, and I am sure we left the shebang line)

I apologize for the several commits again, I keep trying to fetch the upstream and merge before committing and it is screwing things up a little, but I get it after I do it a few times.

Let me know what else I need to do! Again, thank you very much

daler:
Looking good -- let's see how the tests do.

daler:
Looks good! Feel free to merge when you're ready.

mcornwell1957:
Sorry - I am not sure how I can merge without write access? I am looking at this: https://help.github.com/articles/merging-a-pull-request/ and I am not sure how I can. 

Could you let me know what I need to do? I am still a bit of a git amateur..

daler:
Post on #1, we can add you to the team, and then you'll see a big green "merge pull request" button here. In the meantime, I've merged this already.

PR #4355, opened 2017-04-04
Update rmblast as per #4324

pvanheus:

This is a cleaned up version of the commits that went into PR #4324. It:

* adds the bzip2 dependency for rmblast (for makeblastdb)
* adds code to the build.sh to link bzip2.so.1.0 to bzip2.so.1 if necessary
* locks boost to <= 1.58


bgruening:
:)

PR #4358, opened 2017-04-04
Add gclib

davebx:

### DIFF
--- a/recipes/gclib/meta.yaml
+++ b/recipes/gclib/meta.yaml
@@ -9,7 +9,7 @@ build:
 source:
   fn: gclib-0.0.1.tar.gz
   url: https://github.com/gpertea/gclib/archive/907a9d6484b92ea0c357baf2da5d85bc6e6b82b1.tar.gz
-  md5: ab29af6ef38e1d386d45c411f5e40f68
+  md5: 13891db103fab3457beb621ab43f8843
 
 requirements:
   build:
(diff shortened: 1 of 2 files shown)


recipes/gclib/meta.yaml
```diff
@@ -0,0 +1,20 @@
+package:
+  name: gclib
+  version: "907a9d"
+
+build:
+    number: 0
+    skip: False
+
+source:
+  git_url: https://github.com/gpertea/gclib.git
```
bgruening: On github you can use this URL: `https://github.com/gpertea/gclib/archive/907a9d6484b92ea0c357baf2da5d85bc6e6b82b1.tar.gz`
+ a checksum and everything looks great!

recipes/gclib/meta.yaml
```diff
@@ -0,0 +1,20 @@
+package:
+  name: gclib
+  version: "907a9d"
```
bgruening: I'm not sure Conda can handle this in a long run. Is the next hash smaller or greater than this one?
Can you use 0.0.0.1 or something.

recipes/gclib/meta.yaml
davebx: This is the syntax they had in https://conda.io/docs/building/meta-yaml.html#source-from-git

bgruening: @davebx this is correct. At bioconda we do backup all tarballs on Galaxy depot, so we have a mirror of everything. This backup is much easier to do with real tarballs and not git urls. Hope this makes sense.

bgruening:
@davebx please see here for more information: https://bioconda.github.io/guidelines.html#c-c

PR #4342, opened 2017-04-03
Add r-rubic recipe.

jrderuiter:
* [x] I have read the guidelines above.
* [x] This PR adds a new recipe.
* [ ] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

Adds recipe for the RUBIC R package (https://github.com/NKI-CCB/RUBIC).

PR #4365, opened 2017-04-05
Fix rubic dependencies

jrderuiter:

Updates the recipe for r-rubic, which was added in #4342. The update essentially fixes the data.table dependency to 1.9.4 (which is also added as a recipe), as this is the only version of data.table for which is known that this package works. I have submitted an issue to the data.table repository to see if the package can be made to work with newer versions of data.table.

The recipe also includes a small patch to make the plotting compatible with ggplot versions newer than 1.0.1. This fix should be included in newer versions of RUBIC, so we can drop the patch then.


jrderuiter:
The package data.table 1.9.4 is failing the linting, as this package is already present in the r channel. However, the r channel only has versions for R <=3.2.2, so we can not use their package for bioconda. It's unlikely they'll ever provide a build of data.table 1.9.4 for R 3.3.1.

Personally, I would therefore still like to include this recipe in bioconda (skipping the linting). Is this acceptable?

bgruening:
@jrderuiter do you have time to help us migrating all the r packages from the R channel to conda-forge, including data.table? This would be awesome and also solve your problem here (midterm).

jrderuiter:
@bgruening I might have time to help with that somewhere in the near future. Is there any specific timeline for that transition? I don't have any experience with submitting new recipes to conda-forge yet.

Can we still include this in the mean time or should we wait for the migration?

bgruening:
@johanneskoester is currently working on it and I will join this week hopefully. So we might get this working soon. You can start with your package for example?
I would prefer to wait, but if you need it urgently we can get it in I think.

jrderuiter:
@bgruening I'll have a look at submitting this recipe to conda-forge (as this indeed makes more sense). We can leave this pull-request open in the mean time. Thanks for your suggestions!

johanneskoester:
The R migration is tracked in this issue: conda-forge/staged-recipes#2009.

PR #4366, opened 2017-04-05
Build additional object files used by GffCompare

davebx:

GffCompare links with GFastaIndex.o, among others, which gclib's Makefile doesn't build. With this change, that is no longer the case. Obviously GffCompare's Makefile will also need to be patched to reflect this modification.


bgruening:
Thanks Dave!

PR #4369, opened 2017-04-05
Also build a shared library.

davebx:


bgruening:
@davebx you need to always increase the build number. Also you could test the so files like here: https://github.com/conda-forge/libtiff-feedstock/blob/master/recipe/meta.yaml#L43

davebx:
@bgruening Right, forgot about the build number. Is it worthwhile to test for the existence of the other files that get installed?

bgruening:
Not all, but if you want to be sure there is a *so file I would add at least one test.

PR #4373, opened 2017-04-05
Aliencertificatewrapper

eggzilla:

### DIFF
--- a/recipes/rnalien/build.sh
+++ b/recipes/rnalien/build.sh
@@ -7,5 +7,8 @@ export CPPFLAGS="-I${PREFIX}/include"
 stack setup
 stack update
 stack install --extra-include-dirs ${PREFIX}/include --local-bin-path ${PREFIX}/bin
+mv ${PREFIX}/bin/RNAlien ${PREFIX}/bin/RNAlien-bin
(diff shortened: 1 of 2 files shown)


recipes/rnalien/build.sh
```diff
@@ -7,5 +7,8 @@ export CPPFLAGS="-I${PREFIX}/include"
 stack setup
 stack update
 stack install --extra-include-dirs ${PREFIX}/include --local-bin-path ${PREFIX}/bin
+mv ${PREFIX}/bin/RNAlien ${PREFIX}/bin/RNAlien-bin
```
bgruening: You could just create such a bash script:
```
#!/bin/bash
export EMBOSS_ACDROOT=../share/EMBOSS/acd/
export EMBOSS_DATA=../share/EMBOSS/data
export PLPLOT_LIB=../share/EMBOSS
_water $@
```

and copy it from RECIPE_DIR into the $PREFIX.

eggzilla: I think I will stay with the variant of writing the wrapper-script via echo for the moment :-D
Hope this is ok with you?

bgruening: ok

PR #4402, opened 2017-04-11
Bump buildno (fix libintl.so.9 dependency)

epruesse:
Rebuilding the pkg should take care of the (now missing) libintl.so.9.


epruesse:
This may have to be done for all packages linking libintl.so.9...

epruesse:
Looks like the recipe wasn't in good shape: missing_hash, uses_git_url

PR #4403, opened 2017-04-12
biopython: 1.69

saketkc:


bgruening:
Mh, we got a segmentation fault :(

daler:
@saketkc @peterjc after some debugging it looks like `Bio.SVDSuperImposer` is causing the segfault, likely because of missing dependencies.

peterjc:
OK, that's interesting. The recipe does list ``numpy`` as a build requirement so that puzzles me. Any more details?

daler:
The conda-build command runs fine, but it was the mulled-build test that was segfaulting. That runs in a minimal docker container. I ran that container interactively and tried running every `import` test. Turns out `python -c "import Bio.SVDSuperimposer"` triggered the segfault.

I just looked at the source for that module, it's not doing anything crazy. So in that same container I tried `python -c "import numpy"` and that segfaulted. Numpy is the culprit.

@dpryan79 I think I remember you fighting with numpy segfaults a while ago, do you have any suggestions to try?

daler:
@dpryan79 never mind, I found that PR: https://github.com/bioconda/bioconda-recipes/pull/3048#issuecomment-262927140

peterjc:
@saketkc that license change will likely fail, under Biopython 1.69 the ``LICENSE`` file is just a symlink to ``LICENSE.rst``

See more detailed explanation in my pull request #4404 which would fix Biopython 1.68 first.

saketkc:
@peterjc Just fixed!

saketkc:
I have skipped building on `python 3.4` for now. I will go ahead and merge this for now if there are no objections.

daler:
@saketkc go for it. I don't know of a clear path for fixing the 3.4 issue so it could be a while if we wait for that to be resolved.

saketkc:
@peterjc Are you happy with the `license` related changes?

dpryan79:
@daler: My only "fix" has been to skip testing on python3.4.

peterjc:
No. If this is going to be a release for Biopython 1.69, then we need:

```
license_file: LICENSE.rst
```

(We renamed the file, it was ``LICENSE`` up to and including Biopython 1.68, but from Biopython 1.69 onwards it is ``LICENSE.rst`` instead).

However, separate from this, my preference is to fix the Biopython 1.68 recipe first (otherwise the old recipe with the incorrect license will persist in the bioconda archive), i.e. merge #4404 and release it, then apply this pull request.

PR #4410, opened 2017-04-12
Build DWGsim for OS X

karel-brinda:

### DIFF
--- a/recipes/dwgsim/build.sh
+++ b/recipes/dwgsim/build.sh
@@ -25,13 +25,13 @@ cp dwgsim $PREFIX/bin
 cp dwgsim_eval $PREFIX/bin
 cp scripts/dwgsim_eval_plot.py $PREFIX/bin
 
-cd scripts
-
-mkdir -p perl-build
-cp ${RECIPE_DIR}/Build.PL perl-build
-cp *.pl perl-build
-
-cd perl-build
-perl ./Build.PL
-./Build manifest
-./Build install --installdirs site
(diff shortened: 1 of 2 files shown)


bgruening:
Are the patches still needed in this package?
ping @jerowe

karel-brinda:
I think it would be better to make the formula work without any patches because they are version-specific. I will experiment a little bit with it. I think we can pass different parameters to Make directly from the command line.

I am also wondering what is better to use for compilation: either the Samtools conda package (currently used), or the original git submodule. I think that it would be better to use the submodule as the author of the program did all the testing against that specific version. As far as I remember, conda does not clone repositories recursively; therefore, the submodule cloning step would have to be added to `build.sh`.

@bgruening @jerowe What's your opinion?

bgruening:
If you don't need this, which your changes here indicate, I'm all in for removing the complicated patches. Maybe you can extend the tests a little bit to make sure it works reliably.

You can use submodules with conda - you just need to use `git_url:`, which we recommend and the linter will complain, but your can deactivate this linting step. However, we try to build against shared htslib or sametools libraries, as this enables us to share environments with other packages. 
For example if you use build-in htslib and ship this with your package, other packages depending on htslib will overwrite it - so this is probably not going to work. This is also the reason we pin such libraries, so that all (most?) of the bioconda packages are using one library version.

karel-brinda:
I solved it here:
https://github.com/bioconda/bioconda-recipes/pull/4424

Issue #4250, opened 2017-03-22
quiime caused ImportError: No module named PyQt4


mictadlo:
Hi @bgruening,
```
$ convert_fastaqual_fastq.py -f RSnSeQ_110fruit.fastq -c fastq_to_fastaqual -F
Traceback (most recent call last):
  File "/mnt/apps/miniconda2/bin/convert_fastaqual_fastq.py", line 4, in <module>
    __import__('pkg_resources').run_script('qiime==1.9.1', 'convert_fastaqual_fastq.py')
  File "/mnt/apps/miniconda2/lib/python2.7/site-packages/setuptools-27.2.0-py2.7.egg/pkg_resources/__init__.py", line 744, in run_script
  File "/mnt/apps/miniconda2/lib/python2.7/site-packages/setuptools-27.2.0-py2.7.egg/pkg_resources/__init__.py", line 1499, in run_script
  File "/mnt/apps/miniconda2/lib/python2.7/site-packages/qiime-1.9.1-py2.7.egg-info/scripts/convert_fastaqual_fastq.py", line 14, in <module>
    from qiime.util import make_option, create_dir,\
  File "/mnt/apps/miniconda2/lib/python2.7/site-packages/qiime/util.py", line 49, in <module>
    from skbio.util import remove_files, create_dir
  File "/mnt/apps/miniconda2/lib/python2.7/site-packages/skbio/__init__.py", line 15, in <module>
    import skbio.io
  File "/mnt/apps/miniconda2/lib/python2.7/site-packages/skbio/io/__init__.py", line 309, in <module>
    import_module('skbio.io.clustal')
  File "/mnt/apps/miniconda2/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/mnt/apps/miniconda2/lib/python2.7/site-packages/skbio/io/clustal.py", line 123, in <module>
    from skbio.alignment import Alignment
  File "/mnt/apps/miniconda2/lib/python2.7/site-packages/skbio/alignment/__init__.py", line 230, in <module>
    from ._alignment import Alignment, SequenceCollection, StockholmAlignment
  File "/mnt/apps/miniconda2/lib/python2.7/site-packages/skbio/alignment/_alignment.py", line 21, in <module>
    from skbio.stats.distance import DistanceMatrix
  File "/mnt/apps/miniconda2/lib/python2.7/site-packages/skbio/stats/distance/__init__.py", line 193, in <module>
[...]

PR #4411, opened 2017-04-12
Qiime

pbordron:

The current qiime recipe fails because the matplotlib requirement in the recipe is overwrited by the matplotlib requirement of the scikit-bio recipe (scikit-bio installs the last matplotlib version which is the version 2.0.0 when redacting this message)
Moving the matplotlib requirement after the scikit-bio requirement solves the issue.

Following additional libs are required to pass the travis tests. The related issues are detailled in parenthesis:
    - libgcc (ImportError: libstdc++.so.6:: cannot open shared object file: No such file or directory)
    - glib (ImportError: libgthread-2.0.so.0: cannot open shared object file: No such file or directory)
    - xorg-libxext (ImportError: libXext.so.6: cannot open shared object file: No such file or directory)
    - xorg-libxdmcp (ImportError: libXdmcp.so.6: cannot open shared object file: No such file or directory)
    - xorg-libsm (ImportError: libSM.so.6: cannot open shared object file: No such file or directory)
    - xorg-libxrender (ImportError: libXrender.so.1: cannot open shared object file: No such file or directory)
    - xorg-libxau (ImportError: libXau.so.6: cannot open shared object file: No such file or directory)

The import tests are completed by the command tests (they are thoses that lead me to the initial matplotlib issue).

@bgruening: is it the normal conda behavior to overwrite constraints on requirements like the matplotlib issues solved by this PR? A safe behavior will be: (1) make the (recursive) union of all constraints of the requirement, (2) check the constraints consistency to tell if the installation is possible, and (3) when requirement contraints are satisfied, install a right version of each requirement.

### DIFF
--- a/recipes/qiime/1.9.1/meta.yaml
+++ b/recipes/qiime/1.9.1/meta.yaml
@@ -6,9 +6,11 @@ source:
   fn: qiime-1.9.1.tar.gz
   url: https://pypi.python.org/packages/source/q/qiime/qiime-1.9.1.tar.gz
   md5: 32d4db2b2c888dbbfc6aa33f02725526
+  sha1: 2ab4ac9beb745ed3eca3243312cc9ce14e8a20d8
+  sha256: c12dcdae1fe44762a49668276d7e93f487d87a0dc2ff8ca696ac0eff9d92a8a8
 
 build:
-  number: 0
+  number: 1
   skip: True # [not py27]
 
 requirements:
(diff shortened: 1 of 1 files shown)


pbordron:
Also solve the issue https://github.com/bioconda/bioconda-recipes/issues/4250 which is a side effect of the matplotlib issue solved here.

bgruening:
@pbordron awesome PR! This pretty much sounds like a bug to me. Can you test this with the latest conda-build version and if this is still broken can you post this issue at conda-build.

Thanks!

Issue #1581, opened 2016-05-17
License description consistency


hardingnj:
Currently the licence descriptions are a bit all over the place. Is it work spending some effort bringing them into line? I'm not sure whether down the line we may want to filter by license type etc?

Happy to do the work if people think it's worthwhile

`grep -r "license:" . | perl -lne '$_ =~ m/license: (.+)$/; print $1;' | sort | uniq -c`

> ```
>  10 ''
>   4 academic
>   1 Academic and Research License
>   1 Academic Free License 3.0
>   2 Academic Free License (AFL)
>   1 Academic Free License version 3.0
>   1 Academic use, commerically restricted (http://www.stats.ox.ac.uk/~marchini/software/gwas/gwas.html#licence)
>   3 Affero GPL
>   2 AGPL
>   2 AGPL-3
>   1 AGPL + file LICENSE
>   1 AGPLv3
>   1 "Apache 2"
>   2 apache_2_0
>  11 Apache 2.0
>  12 Apache License 2.0
>   1 Apache License V2.0
>   4 Apache License, Version 2.0
>   9 Apache Software License
>   3 Apache v2
>   1 "Apache v2.0"
>   2 Apache v2.0
>   1 Artistic
>   1 "Artistic"
>   1 artistic_2
>  64 Artistic-2.0
>   1 'Artistic-2.0 | file LICENSE'
>   2 'Artistic-2.0 + file LICENSE'
>   1 'Artistic-2.0 | GPL-2 + file LICENSE'
>   3 Artistic License
>  10 Artistic License 2.0
>   1 Biopython License Agreement
>   2 Boost Software License
>  36 BSD
>   1 BSD_2_clause + file LICENSE
>   1 BSD 3-clause
>   1 BSD-3-clause
>   1 BSD-3-Clause
>   1 BSD_3_clause + file LICENCE
>   4 BSD_3_clause + file LICENSE
>  52 BSD License
>   1 BSD New
>   1 CC BY-NC 2.0
>   1 CC BY-SA 3.0
>   1 CCPL (Creative Commons Public License)
>   2 CeCILL
>   1 CeCILL FSLA
>   1 Commercial (requires license for multithreaded use)
>   1 Copyright © 2009-2014,2015 by Thomas E. Dickey
>   1 'Creative Commons Attribution-NonCommercial-ShareAlike 3.0 Unported'
>   7 Custom
>   2 Custom OSS
[...]

PR #4423, opened 2017-04-13
GNU licenses like GPL, AGPL, LGPL are not public domain

peterjc:

I've not found any BioConda documentation defining the ``license_family:`` field (see also #1581), but none of the GNU licenses (i.e. any versions of the GPL and LGPL) can be considered as public domain.

It makes sense to me to class any version of the GPL as ``license_family: GPL`` as in this LGPL example:

```
about:
  home: https://pypi.python.org/pypi/xmlbuilder/1.0
  license: LGPL v3
  summary: 'pythonic way to crate xml/(x)html files'
  license_family: LGPL
```

### DIFF
--- a/recipes/biomaj/meta.yaml
+++ b/recipes/biomaj/meta.yaml
@@ -49,5 +49,5 @@ about:
   home: http://biomaj.genouest.org
   license: GNU Affero General Public License v3 or later (AGPLv3+)
   summary: 'BioMAJ'
-  license_family: Public-Domain
+  license_family: AGPL
   summary: Automates the update cycle and the supervision of the locally mirrored databank repository
--- a/recipes/biomaj/meta.yaml
+++ b/recipes/biomaj/meta.yaml
@@ -9,6 +9,9 @@ source:
   patches:
     - deps.patch
 
+build:
+  number: 1
+
 requirements:
   build:
     - python
--- a/recipes/biomaj/meta.yaml
+++ b/recipes/biomaj/meta.yaml
@@ -9,9 +9,6 @@ source:
   patches:
     - deps.patch
 
-build:
-  number: 1
-
 requirements:
   build:
     - python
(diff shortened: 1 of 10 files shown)


peterjc:
Does this also need the build number bumped?

daler:
Thanks @peterjc. These need a build number bump in order to trigger a rebuild and re-upload to anaconda.org. The bioconda homepage for each will be updated at least nightly with the master branch of bioconda-recipes, so no build number bump needed for that.

Since these are more than just cosmetic changes to the recipes, build number bump is probably best.

peterjc:
Given the builds fail (and from getting to know BioConda it is not unusual for a previously working recipe to bit-rot due to external changes), is it worth committing this without the build-number bump?

That way the license will be clearer next time someone updates the recipes concerned, but it does not push an update to the Anaconda package right now.

daler:
I agree with your point about the next time it needs to be updated. I rarely confirm the existing license in a recipe is correct when updating it, so these updates are valuable to have. I'd be OK with only changing the licences and not the build numbers.

peterjc:
Travis skipped everything on Linux, but decided to test ``biomaj`` on the Mac, and failed - something unreleated to my change:

```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 1781: ordinal not in range(128)
```

I can reproduce this locally,

```
$ LC_ALL=C python3 -c "print(len(open('README.md').read()))"
  File "<string>", line 1, in <module>
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 1781: ordinal not in range(128)
```

This is due to an ellipse ``…`` on line 52 of their ``README.md`` file, issue logged https://github.com/genouest/biomaj/issues/86

Can I merge this pull request anyway (updating the license metadata)?

peterjc:
The little locale/unicode issue in ``biomaj`` has been fixed upstream, and is not relevant to this pull request anyway.

Can I merge this as it is to update the license metadata ready for next time any of these recipes are worked on?

bgruening:
Thanks Peter!

PR #4428, opened 2017-04-13
Stringmlst

Takadonet:

### DIFF
--- a/recipes/stringmlst/meta.yaml
+++ b/recipes/stringmlst/meta.yaml
@@ -1,0 +1,30 @@
+package:
+  name: stringmlst
+  version: "0.3.6.1"
+
+source:
+  url: https://github.com/jordanlab/stringMLST/archive/0.3.6.1.tar.gz
+  md5: 13d8eb71e17abe0294989fef05071776
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - pyfaidx ==0.4.8.1    
+
+  run:
+    - bwa ==0.7.13
+    - bedtools ==2.24
+    - samtools ==1.3
+    - pyfaidx ==0.4.8.1
+    - python
+
+test:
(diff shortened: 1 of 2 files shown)


recipes/stringmlst/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+  number: 0
+  skip: True  # [osx]
+    
+
+requirements:
+  build:
+    - pyfaidx ==0.4.8.1    
+
+  run:
+    - bwa ==0.7.13
```
bgruening: Is the strict pinning necessary?

recipes/stringmlst/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - pyfaidx ==0.4.8.1
+    - python
+
+test:
+  commands:
+    - stringMLST.py  -v
+
+about:
+  home: https://github.com/jordanlab/stringMLST
+  license: Custom
```
bgruening: What means custom?

Takadonet: The current license in the repo is a custom one written by the authors. I poke them and they decided to take https://creativecommons.org/licenses/by-nc-sa/4.0/ ! So that is awesome but they indicate that the changes will only take effect April 21.... So will not merge till then.

recipes/stringmlst/meta.yaml
Takadonet: These versions were indicated that needed to be installed to ensure that the tool works.

druvus: I think the versions could be relaxed. Their installation example are using higher version numbers.

Takadonet: got rid of specific versions of requirements.

PR #4431, opened 2017-04-13
turns out we must build with GCC on OS X

molpopgen:

### DIFF
--- a/recipes/fwdpy11/build.sh
+++ b/recipes/fwdpy11/build.sh
@@ -4,5 +4,6 @@
 export CFLAGS="-I${PREFIX}/include"
 export CXXFLAGS="-I${PREFIX}/include"
 export LDFLAGS="-L${PREFIX}/lib"
-
+export CC=gcc
+export CXX=g++
 $PYTHON setup.py install
--- a/recipes/fwdpy11/build.sh
+++ b/recipes/fwdpy11/build.sh
@@ -1,9 +1,8 @@
 #!/bin/bash
 
-
 export CFLAGS="-I${PREFIX}/include"
 export CXXFLAGS="-I${PREFIX}/include"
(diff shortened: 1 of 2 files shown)


molpopgen:
@bioconda/core. Ready to review.  Turns out gcc is needed on osx.  It seems that it is impossible (for me at least) to come up with a build setup that passes all tests on all combos of osx/xcode/conda.

molpopgen:
Gonna hold off on this one.  Sorted out the llvm builds, but need to figure out how to automate it in setup.py.

johanneskoester:
For me building with gcc on osx would be fine. Continuum does the same for R recipes, so why not?

bgruening:
Given the new gcc package which is expected after the new conda-build release, we might also switch to GCC on all platforms.

johanneskoester:
Indeed.

molpopgen:
I'm all in favor of GCC builds for anything Anaconda related.  However, there are some more general issues, mainly for homebrew users.  This project is a C++11/pybind11-based Python package relying on exception translation from C++ to Python working properly.  With OSX/llvm, it is critical to get the `-stdlib` and `-mmacosx-version-min` handled correctly, else the exceptions abort rather than pass on to Python.  With GCC, it is clear that none of this is relevant.  However, the 'brew GCC packages seem to be misbehaving in several respects, and thus it seems LLVM is required for that user base.

I'm about to push changes to the main package assuming an LLVM build on OS X, allow a --gcc option to setup.py to override the darwin-specific stuff, and use that to get a bioconda/gcc/osx build sorted.

But yes, the sooner we can move everything to GCC on OSX, the happier I'll be.

Thanks.

molpopgen:
@bioconda/core ok--this one is ready to roll.

bgruening:
Feel free to merge looks good!

PR #4450, opened 2017-04-18
Re-add ucsc-liftover, updated to 332

nsoranzo:
Double-quote env variables in Bash scripts.

### DIFF
--- a/recipes/ucsc-liftover/build.sh
+++ b/recipes/ucsc-liftover/build.sh
@@ -6,8 +6,7 @@ mkdir -p "$BINDIR"
 (cd kent/src/lib && make)
 (cd kent/src/jkOwnLib && make)
 (cd kent/src/hg/lib && make)
-(cd kent/src/utils/stringify && make)
-(cd {program_source_dir} && make)
+(cd kent/src/hg/liftOver && make)
 mkdir -p "$PREFIX/bin"
-cp bin/{program} "$PREFIX/bin"
-chmod +x "$PREFIX/bin/{program}"
+cp bin/liftOver "$PREFIX/bin"
+chmod +x "$PREFIX/bin/liftOver"
--- a/recipes/ucsc-liftover/build.sh
+++ b/recipes/ucsc-liftover/build.sh
@@ -1,12 +1,15 @@
 #!/bin/bash
-
-export MACHTYPE=x86_64
-export BINDIR=$(pwd)/bin
-mkdir -p "$BINDIR"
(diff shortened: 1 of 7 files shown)


nsoranzo:
@bioconda/core Any clue why the build https://travis-ci.org/bioconda/bioconda-recipes/jobs/223113639#L611 is failing with:
```
/bin/sh: libpng-config: command not found
/bin/sh: libpng-config: command not found
/bin/sh: mysql_config: command not found
/bin/sh: mysql_config: command not found
```

johanneskoester:
No idea, it should be there because of your dependencies. Does it work locally with plain conda build?

bgruening:
Just a guess. Are these scripts with a invalid shebang?

nsoranzo:
@johanneskoester Yes, it works locally with `conda build`, but fails with `./simulate-travis.py` with the same error.
@bgruening They both have `#!/bin/sh` as shebang.

nsoranzo:
@johanneskoester @bgruening I suspect this is due to the conda build environment being created twice, first with the correct build requirements, then with others, see https://travis-ci.org/bioconda/bioconda-recipes/jobs/223113639#L832 

I have absolutely no idea why that's happening.

bgruening:
have you seen this one? `pngwrite.c:7:87: fatal error: png.h: No such file or directory`

johanneskoester:
If it cannot find the libpng header, maybe try adding `CPP_INCLUDE_PATH=$PREFIX/inlcude` or `C_INCLUDE_PATH=$PREFIX/include`?

nsoranzo:
That's not the issue unfortunately, libpng-config is not found, which in turn creates the error. I have been debugging locally and it looks like that all requirements are not in the Conda environment.

bgruening:
@nsoranzo it looks like that the build.sh script is executed, even if the `build` section exists. If you look at the OSX build, which just downloads a pre-compiled binary, this tries also to use the build.sh script as it seems.
This can be a conda bug. If we move the section into the build.sh script maybe this will work?

nsoranzo:
@bgruening Thanks for the tip, that fixed the OS X build, but not the Linux one (as I expected). I added a few debugging commands to clarify my previous comments, you should now clearly see that the correct Conda build environment is overwritten but one containing readline, sqlite, python, mercurial, setuptools,wheel and pip.

nsoranzo:
After another rebase, this is finally green thanks to some upstream Conda  bugfix.

PR #4455, opened 2017-04-18
Add recipe for velvet-sc

bebatut:

### DIFF
--- a/recipes/velvet-sc/meta.yaml
+++ b/recipes/velvet-sc/meta.yaml
@@ -17,10 +17,10 @@ package:
 
 requirements:
   build:
-    - gcc   # [not osx]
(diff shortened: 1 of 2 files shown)


recipes/velvet-sc/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  number: 0
+  skip: False
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+requirements:
+  build:
+    - gcc
```
bgruening: osx, llvm?

bebatut: It was failing with llvm for OSX and work with gcc

PR #4456, opened 2017-04-18
New Recipe Sourmash

Takadonet:

### DIFF
--- a/recipes/sourmash/meta.yaml
+++ b/recipes/sourmash/meta.yaml
@@ -20,7 +20,6 @@ requirements:
     - setuptools
     - screed >=0.9
     - pyyaml >=3.11
-    - libgcc
 
   run:
     - python
--- a/recipes/sourmash/meta.yaml
+++ b/recipes/sourmash/meta.yaml
@@ -26,6 +26,7 @@ requirements:
     - setuptools
     - screed >=0.9
     - pyyaml >=3.11
+    - libgcc
 
   run:
     - python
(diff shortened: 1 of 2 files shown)


recipes/sourmash/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - sourmash = sourmash_lib.__main__:main
+  number: 0
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - screed >=0.9
+    - pyyaml >=3.11
+    - libgcc
```
bgruening: libgcc is not needed as build-time dependencies, isn't it?

bgruening: Do you need gcc here?

Takadonet: Was getting run time  error 
"SERR ImportError: libstdc++.so.6: cannot open shared object file: No such file or directory"

bgruening: In build? Are you sure? In runtime, this can be but build-time deps?

Takadonet: Here is the failed build before adding libgcc . 

https://travis-ci.org/phac-nml/bioconda-recipes/jobs/223166549#L1859

Error near the bottom.

bgruening: Yes, this is only for run-time deps not for build-time.

Takadonet: removed and test pass.

bgruening:
@Takadonet thanks!

PR #4500, opened 2017-04-22
Added recipe: xxmotif-1.6

croth1:

### DIFF
--- a/recipes/xxmotif/meta.yaml
+++ b/recipes/xxmotif/meta.yaml
@@ -1,0 +1,32 @@
+{% set version = "1.6" %}
+
+package:
+  name: xxmotif
+  version: {{ version }}
+
+source:
+  fn: xxmotif-{{ version }}.tar.gz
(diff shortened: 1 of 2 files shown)


recipes/xxmotif/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+  url: https://github.com/soedinglab/xxmotif/archive/{{ version }}.tar.gz
+  sha256: 235eb8ad562c4757a549751bb290f443ff70c29a142edbff0f03440f097b61cb
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - cmake
+    - gcc
```
bgruening: Is llvm on OSX possible?
libgcc is not needed as build dependency

croth1: Unfortunately this code is very old and fragile. For reasons that elude me it apparently uses private gcc implementation details. I haven't managed to build it with llvm yet and I would rather not touch the code if I can avoid.

bgruening: Fair enough!

PR #4505, opened 2017-04-23
use openjdk

bgruening:


druvus:
Current fastqc recipe is now using openjdk

PR #4521, opened 2017-04-24
Update ProPhyle version

karel-brinda:

### DIFF
--- a/recipes/prophyle/meta.yaml
+++ b/recipes/prophyle/meta.yaml
@@ -1,18 +1,19 @@
-{% set version = "0.1.0.33" %}
+{% set version = "0.1.0.34" %}
+{% set sha256 = "5b272d14dada63483344a06a2b27558476226442814c5373ad7bfe73ed00ba3c" %}
 
 package:
   name: prophyle
   version: {{ version }}
 
-  #source:
-  # fn: prophyle-{{ version }}.tar.gz
-  #url:
-  #sha256:
+source:
+  fn: prophyle-{{ version }}.tar.gz
+  url: https://pypi.io/packages/source/p/prophyle/prophyle-{{ version }}.tar.gz
+  sha256: {{ sha256 }}
 
 build:
   number: 0
   skip: True # [py27]
-  script: pip install --upgrade git+https://github.com/karel-brinda/prophyle
+  script: python setup.py install
 
 requirements:
   build:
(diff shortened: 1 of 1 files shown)


recipes/prophyle/meta.yaml
```diff
@@ -10,7 +10,7 @@ package:
   #sha256:
 
 build:
-  number: 0
+  number: 1
```
druvus: Build number should restart at 0 for new versions.

recipes/prophyle/meta.yaml
```diff
@@ -10,7 +10,7 @@ package:
...
   skip: True # [py27]
   script: pip install --upgrade git+https://github.com/karel-brinda/prophyle
```
bgruening: Can we please change this recipe to use the `source` section and not github. Thanks.

karel-brinda: @bgruening Hi Björn, thank you for your comment. ProPhyle is still in a testing phase and does not have any Github releases, yet. This is why the source section is currently not used. Once a stable version is ready, we plan to create a standard Github release and use the source section of the receipt properly.

However, we might use the source section already now – with the PyPI releases (https://pypi.python.org/pypi?:action=display&name=prophyle).  Do you know how to derive a PyPI package address from the version (to be able to update the receipt just by incrementing a single Jinja variable)?

bgruening: I'm not sure I understood the question, but this URL you can use: https://pypi.python.org/packages/bc/59/d5bf13ee886b86ecaaf8e86cd6cf98adc1cf65709de02130f61ae78d2041/prophyle-0.1.0.34.tar.gz
md5=29d6ee0a1c41694342464c31b3826140

karel-brinda: This is the url I was looking for:
https://pypi.io/packages/source/p/prophyle/prophyle-0.1.0.34.tar.gz
It does not contain any hashes.

bgruening: This one is even better. For your information, you can also use github with a URL similar to this:https://github.com/karel-brinda/prophyle/archive/3a4e3ab0be3e10ad5f15b69bce0749cb812c360d.tar.gz The hash is your latest commit.

karel-brinda: Wow, I did not now this. Thank you!

bgruening:
```
RuntimeError: Setuptools downloading is disabled in conda build. Be sure to add all dependencies in the meta.yaml  url=https://pypi.python.org/simple/psutil/r
```
psutil is missing as dependency

PR #4535, opened 2017-04-26
Remove libxml2 recipe, which is present in default and conda-forge

blankclemens:
Since this package duplicates the ones in `default` and `conda-forge`, contains no special magic related to `bioconda` and is additionally not fully working, I'd like to remove it.

I'd kindly ask to check if this might break something.

Concerning the "not fully worrking":

When installing `lxml` I cannot import `etree` from it with the bioconda version:

```
from lxml import etree                                                                             │
ImportError: libiconv.so.2: cannot open shared object file: No such file or directory
```


bgruening:
Thanks!

bgruening:
Transferred to bioconda-legacy.

PR #4537, opened 2017-04-26
Tries to solve dependency problem: pysam 0.10.0 + htslib

yhoogstrate:
Loading pysam 0.10.0 with htslib-1.4 this returns the following error:
```
  File "/home/<user>/anaconda2/envs/pysam0100/lib/python2.7/site-packages/pysam/__init__.py", line 5, in <module>
    from pysam.libchtslib import *
ImportError: libhts.so.1: cannot open shared object file: No such file or directory
```
This makes sense since pysam 0.10.0 was built against htslib 1.3.

Is there a possiblity to rebuild pysam 0.10.0 also with htslib 1.4 (e.g. by increasing the built nr.) or does it require to pin to htslib 1.3 or 1.4 explicitly? Could you please comment on this @chapmanb @kyleabeauchamp ?

### DIFF
--- a/recipes/pysam/0.10.0/meta.yaml
+++ b/recipes/pysam/0.10.0/meta.yaml
@@ -18,9 +18,9 @@ requirements:
     build:
         - gcc  # [linux]
         - llvm # [osx]
-        - htslib >=1.3,<1.4
-        - samtools >=1.3,<1.4
-        - bcftools >=1.3,<1.4
+        - htslib ==1.3
+        - samtools ==1.3
+        - bcftools ==1.3
         - cython
         - python
         - setuptools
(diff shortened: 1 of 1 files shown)


chapmanb:
Older pysam versions like 0.10.0 don't work correctly with htslib 1.4, since it needed some fixes to be compatible. So I wouldn't try to build with this.

I'm not sure I understand the PR since the previous version pinned to be any 1.3.x htslib and already excluded 1.4. pysam is compatible with any 1.3.x so we don't need to restrict to 1.3.

Could you provide more details about what is happening that you'd like to fix? In practice, I'd use pysam 0.11+ if you want to get up to date and 0.10.0 if you need to maintain htslib 1.3 compatbility.

yhoogstrate:
Thanks @chapmanb,

I am working on tool that uses pysam 0.10.0 and during the build process it installs pysam 0.10.0 with htslib 1.4. You are right that this is not a mistake in the recipe so I'll close this PR and I will try to fix it elsewhere.

Thanks!

PR #4541, opened 2017-04-27
cmapPy

oena:


druvus:
@oena Bioconda does't build for windows so you need to remove the windows specific bld.bat file. You could also remove all comments from the meta.yaml file.

oena:
@druvus ah right, my bad. Thanks!

recipes/cmappy/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - cmapPy
+    - cmapPy.clue_api_client
+    - cmapPy.pandasGEXpress
+
+  commands:
+    - gctx2gct --help
+    - gct2gctx --help
+    - concat_gctoo --help
+
+  requires:
```
bgruening: You can remove this.

recipes/cmappy/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  url: https://pypi.python.org/packages/83/9d/2fb659a7d1ffb1460edb1cae5e7ff7c7ff744f53a73276b12ff0be48416a/cmapPy-1.0.8.tar.gz
+  md5: be801cf4ab097c002742c8edacb2aba6
+
+build:
+  entry_points:
+    - gctx2gct=cmapPy.pandasGEXpress.gctx2gct:main
+    - gct2gctx=cmapPy.pandasGEXpress.gct2gctx:main
+    - concat_gctoo=cmapPy.pandasGEXpress.concat_gctoo:main
+
+  number: 1
```
bgruening: This can start with 0.

recipes/cmappy/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  version: "1.0.9"
+
+source:
+  fn: cmapPy-1.0.9.tar.gz
+  url: https://pypi.python.org/packages/91/8d/fa4f925ce1ab47b95e091ff595e905b959bd8f628578e5b10898db275a39/cmapPy-1.0.9.tar.gz
+  md5: 27cabdf06940d60592dabd5850a8f25b
+
+
+build:
+  skip: True
```
bgruening: This will now skip every build, so you will not get any package. What do you wanted to do?

oena: @bgruening the package is only Python 2.7 compatible, so I only want to build for Python 2.7. The bioconda documentation seems to say I should put in this argument; what am I missing?

https://bioconda.github.io/guidelines.html#python

bgruening: Please put this in `skip: true  # [py3k]`.
And remove this once from the URL which you included a minute ago :)

oena:
@bgruening I keep getting error messages that "Error: bad character '=' in package name dependency python==2.7*" and so on. What am I missing regarding how to specify appropriate versions of requirements?

bgruening:
@oena late here, I just fixed it real quick. Will look at it tomorrow more if it fails, ok?

oena:
@bgruening the same error seems to be occurring still :( Any thoughts on what's going on?

bgruening:
@oena I get mixed messages here. The program is saying "Python3 is not supported" and the other time it is saying that "Python2" is not supported ;) Any idea?

oena:
@bgruening that's my fault, it should be supported only in Python 2-- I think I got confused about the selectors, it was developed on Py 2.7 so I wanted to try building on that specifically. Travis CI doesn't seem to work on either though

bgruening:
I have at least an idea what can cause the issue. py36 is not known to Conda, we are using currently an older version and it seems it is not skipping the py36 build which keeps crashing.
Let me think about an other solution.

bgruening:
@oena  @druvus thanks a lot!

oena:
Aha, didn't know about the osx dependencies. Thank you @druvus and @bgruening for all your help!

PR #4542, opened 2017-04-27
stringMLST

ar0ch:

This updates @Takadonet's recipe to the latest version of stringMLST, and includes the correct licensing information (previous tag was still under our old license). Also places a hard dependency on samtools/htslib >=1.0, which was relaxed during the initial submission to no-requirement.

 @bioconda/core If you could review, that'd be wonderful

Thanks,
Aroon

### DIFF
--- a/recipes/stringmlst/meta.yaml
+++ b/recipes/stringmlst/meta.yaml
@@ -1,10 +1,10 @@
 package:
   name: stringmlst
-  version: "0.3.6.1"
+  version: "0.3.7"
 
 source:
-  url: https://github.com/jordanlab/stringMLST/archive/0.3.6.1.tar.gz
-  md5: 13d8eb71e17abe0294989fef05071776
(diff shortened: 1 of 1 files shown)


tomkinsc:
Looks good, maybe add a `license_file` key:value pair under `license` to point to one of the license files in the source tarball.

druvus:
@ar0ch @tomkinsc Skip is set to True so the recipe wasn't tested or built.

ar0ch:
@druvus Hmm, the original accepted recipe has that value set.  :-/ 

I'll fix that and submit a new PR

ar0ch:
The output from Travis doesn't appear to change when the [`Skip` flag is unset](https://travis-ci.org/ar0ch/bioconda-recipes/jobs/226489171)

The recipe does build fine when manually run through `conda build`

druvus:
@ar0ch I fixed it in #4586

ar0ch:
Thanks @druvus!

PR #4544, opened 2017-04-27
Fix deprecated Perl syntax in ESTScan and fix dependencies and tests

gregvonkuster:


bgruening:
Can you please add 'perl' as dependency as well?

gregvonkuster:
Sure, but I thought since the perl-btlib dependency requires perl that it would be "passed through" to this recipe.  Is this not correct?

bgruening:
This is true, but we might build packages for multiple perl versions in the future.

gregvonkuster:
Thanks @bgruening ;)

PR #4556, opened 2017-04-28
Remove recipe r-robustbase due to "in_other_channels"

blankclemens:
Updating the package fails due "in_other_channels" linting error. Since the recipe seem to not differ from the one in `r` I would assume it's safe to remove it - but I'd like somebody to check who knows about some possible special R thingy going on here.


bgruening:
The r-channel is currently migrated to conda-forge, so yes we should remove this recipe here. The question is if we want to do it now or later? E.g. the r-channel is hard to update by us, so if you need an update maybe it's better to keep this recipe until it is in conda-forge.

blankclemens:
Thanks for the clarification. So it might not make sense to remove them one by one, but at some point all/most of the r packages?

I do not need the update, but while rebuilding I thought why not update this as well - and then run into this problem ;) - the update is in the r and default channel already anyways.

bgruening:
https://anaconda.org/bioconda-legacy/r-robustbase 
Thanks.

PR #4559, opened 2017-04-28
REL: microversion release of biom-format

wasade:

This PR adds an entry for a new microversion release of BIOM-Format. `build.sh` and `meta.yaml` were copied from the `2.1.5` recipe and updated as necessary.

### DIFF
--- a/recipes/biom-format/2.1.6/build.sh
+++ b/recipes/biom-format/2.1.6/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+$PYTHON setup.py install
+
+# Add more build steps here, if they are necessary.
+
+# See
+# http://docs.continuum.io/conda/build.html
+# for a list of environment variables that are set during the build process.
--- a/recipes/biom-format/2.1.6/build.sh
+++ b/recipes/biom-format/2.1.6/build.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+CC=${PREFIX}/bin/gcc
+CXX=${PREFIX}/bin/g++
+
 $PYTHON setup.py install
 
 # Add more build steps here, if they are necessary.
--- a/recipes/biom-format/2.1.6/meta.yaml
+++ b/recipes/biom-format/2.1.6/meta.yaml
@@ -1,0 +1,61 @@
+package:
+  name: biom-format
+  version: "2.1.6"
+
+source:
+  git_url: https://github.com/biocore/biom-format.git
+  git_rev: 2.1.6
+
+build:
+  # noarch_python: True
+  preserve_egg_dir: True
+  entry_points:
+    # Put any entry points (scripts to be generated automatically) here. The
+    # syntax is module:function.  For example
+    #
+    # - biom-format = biom-format:main
+    #
+    # Would create an entry point called biom-format that calls biom-format.main()
+
+    - biom=biom.cli:cli
+
+  number: 3
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - pyqi ==0.3.2  # [py2k]
+    - click
+    - h5py
+    - numpy >=1.3.0
+    - future >=0.15.0
+    - scipy >=0.13.0
+    - pandas
+
+  run:
+    - python
+    - setuptools
+    - click
+    - h5py
+    - pyqi ==0.3.2  # [py2k]
+    - numpy >=1.3.0
+    - future >=0.15.0
+    - scipy >=0.13.0
+    - pandas
+
+test:
+  # Python imports
+  imports:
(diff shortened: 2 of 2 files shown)


wasade:
I'm not sure how to resolve the build issue encountered. One of the dependencies, `click`, is attempting to execute `locale` through a subprocess and failing. The issue I'm seeing appears to be the same as in this [thread](https://github.com/garbas/pypi2nix/issues/110), and it looks like the [resolution](https://github.com/garbas/pypi2nix/issues/110#issuecomment-290660209) is to pass in arguments to docker on run. I'm not sure how to modify what arguments are passed to docker in this build system. Any recommendations?

bgruening:
Hi @wasade! I think we have this package already. Can you please look at: https://github.com/bioconda/bioconda-recipes/tree/master/recipes/biom-format

wasade:
@bgruening, this is for a microversion release pushed out last week. `2.1.6` is not presently represented in bioconda

bgruening:
I see, than please start the build-number with 0.

bgruening:
@wasade can you try removing the `python -c "import biom.cli"` if this works. It really unfortunate that click is using the `locale` binary.

wasade:
Done, I think :)

wasade:
`locale` issue persists. Is there a work around for bioconda? Any idea why this doesn't impact other packages which use click (like biom-format 2.1.5)?

bgruening:
@wasade how urgent is this? The problem is that click now depends on the `locale` binary on python3, which is not in our busybox Docker container. I assume that the old version of biom-format had an old click version without this problem. I'm not sure how to solve this yet. We can deploy the Conda package, but will not have a Docker container in the worst case.

bgruening:
Reading this: http://click.pocoo.org/5/python3/
Makes me think we should maybe drop python3?

wasade:
@bgruening, it's not critical in that biom-format 2.1.6 is available from pypi and the biocore channel. We did not explicitly change a minimum click requirement for this release, so I'm guessing it's just sourcing whatever is the latest. Py3 is a hard requirement for biom-format as many of the packages which depend on it are py3 only.

bgruening:
Hi @wasade!
Look at this one: https://github.com/bioconda/bioconda-recipes/pull/4612
It passed and has click defined? It is something special with biom-format?

ewels:
To weigh in on this - MultiQC is also suffering from the same problem. See the change from @chapmanb [here](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/multiqc/meta.yaml#L13-L16) and the [associated travis failure](https://travis-ci.org/bioconda/bioconda-recipes/jobs/191266605).

ewels:
Also, see one of the files in the PR that you mentioned @bgruening - [`run_test.sh`](https://github.com/bioconda/bioconda-recipes/pull/4612/files#diff-e1e5114a3d9e1895b8a937e348a8361e):

```bash
#!/bin/bash

# Click tests for python3 unicode handling using the `locale`
# executable, which does not seem to be present in busybox.
# The workaround is to bypass the `locale` test by setting the
# LANG env variable directly (from looking at Click's source).
LANG=en_US.UTF-8 fsnviz --help
```

So I guess this is why their tests are not failing..

bgruening:
Ha! Interesting! Also I'm not a fan of this we try to offer the possibility to specify a bigger test container - say CentOS, Ubuntu - to prevent this error. Ideally we should probably fix click.

wasade:
Thanks, @ewels. Do you guys advise adding in the same `skip` that's in place with MultiQC?

ewels:
Well, it's not ideal as it skips the python 3 build, so your package won't be available to Python3 users. But it does at least mean that the build succeeds so that the package is available to _some_ users.. :)

You can see this if you do `conda search multiqc`. Before this error started appearing and the `skip` was added, MultiQC version 0.9 was available in several versions:
```
0.9                      py35_0  bioconda
0.9                      py27_0  bioconda
0.9                      py34_0  bioconda
```

But the new v1.0 is only available for py2.7:
```
1.0                      py27_0  bioconda
```

Note that v0.9 had exactly the same `click` code as v1.0, so something has changed elsewhere between these two releases. Now that v1.0 is released for py2.7 users, I will prepare a second PR to play around without the `skip` to see if I can get it working for everyone.

Phil

wasade:
Ah, that is prohibitive for our users as many consumers are Py3 only. 

From what I can tell, its the build system with bioconda which changed seeing as biom 2.1.5 is in and works (with the same click dependency)?

druvus:
Added in #6400 so I am closing this.

wasade:
Fantastic, thank you!

PR #4563, opened 2017-04-28
add sequana recipes

cokelaer:

### DIFF
--- a/recipes/sequana/meta.yaml
+++ b/recipes/sequana/meta.yaml
@@ -34,7 +34,7 @@ requirements:
     - setuptools
     - easydev >=0.9.34
     - reports >=0.3.0
-    - matplotlib >=2.0.0
+    - matplotlib 
     - pyvcf
     - pandas
     - cutadapt >=1.9.1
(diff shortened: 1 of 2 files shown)


recipes/sequana/meta.yaml
```diff
@@ -1,21 +1,17 @@
...
-
 package:
-  name: {{ name }}
-  version: {{ version }}
+  name: sequana
+  version: "0.2.0"
 
 source:
-  fn: {{ name }}-{{ version }}.tar.gz
-  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
```
bgruening: Any reason why you reverted the nicer template syntax from the previous version?

cokelaer:
well, finally, a working recipes but with no imports... let us merge and see if we can install it with conda.

bgruening:
@cokelaer merging a PR without tests is not good. Sure it will pass, without tests but it is for sure broken :(

cokelaer:
I understand but it looks like the qt4/qt5 is unsolvable for now. I cannot reproduce the error locally so I am a bit desperate to get something to test further. It looks like the import fails when importing biokit, which relies on matplotlib /qt . Within a fresh conda environement biokit and its matplotlib/qt dependencies work. So, I was expecting sequana (this recipes) to behave in the same way. I would appreciate any help but this looks like it is not obvious at all. Any help appreciated of course. I also found difficult to find information on how to use the docker locally to reproduce what is going on on Travis CI so if you have web page with further information, I am also happy to have a look.

PR #4572, opened 2017-05-01
Bump Picopore to v1.1.4

scottgigante:
Not sure why I keep swapping between tar.gz and zip... probably to do with what dist I use to build+push.

* [Y] I have read the guidelines above.
* [N] This PR adds a new recipe.
* [Y] This PR updates an existing recipe.
* [N] This PR does something else (explain below).


druvus:
thanks @scottgigante

PR #4576, opened 2017-05-02
Gmap zlib

bow:

Two things:
* zlib is apparently still not working in 2014-12-23.
* Fixes a wrong duplicate test (the second tests should be `--check` instead of `--version` again)


bgruening:
Cool thanks!

PR #4588, opened 2017-05-03
Bulk update support

johanneskoester:
Consider env matrix updates only when on a branch called "bulk/*".
    
Together with PR #96, this allows us to perform bulk updates in an iterative way,
without using blacklist. The workflow is as follows:
    
1. Create a PR with an updated env matrix, with a target branch called "bulk/*".
2. The PR will be tested, but bioconda-utils will only build recipes that have changed in addition to the env matrix.
3. Once the PR is merged, bioconda-utils will build the changes induced by the new env matrix. Whenever a recipe has been successfully built, the package will be immediately uploaded to anaconda and biocontainers (mulled).
4. If errors occur, they can be fixed in a PR against the bulk branch. In such a case, we start again at step 1. Importantly, since successfully built packages have been already uploaded, we won't do redundant work and the list of packages will become shorter with each iteration.
    
If a timeout occurs during above process, once can simply restart the build job of the bulk branch, and bioconda-utils will go on with the still missing recipes.
    
For these changes to work, PR #96 has to be merged first.

@bioconda/core please check this out. I think it will solve our bulk update issues, so that we can finally update to Python 3.6 and R 3.4 (for R 3.4 we need to take care of conda-forge first).

### DIFF
--- a/scripts/travis-run.sh
+++ b/scripts/travis-run.sh
@@ -24,7 +24,7 @@ SKIP_LINTING=false
 # determine recipes to build. If building locally, build anything that changed
 # since master. If on travis, only build the commit range included in the push
 # or the pull request.
-if [[ $TRAVIS = "true" ]]
+if [[ $TRAVIS == "true" ]]
 then
     RANGE="$TRAVIS_BRANCH HEAD"
     if [ $TRAVIS_PULL_REQUEST == "false" ]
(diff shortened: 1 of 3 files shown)


scripts/travis-run.sh
```diff
@@ -34,22 +34,32 @@ then
...
     git diff --exit-code --name-only $RANGE scripts/env_matrix.yml
     ENV_CHANGE=$?
     set -e
-    if [ $ENV_CHANGE -eq 1 ] || [ $TRAVIS_EVENT_TYPE = "cron" ]
+
+    if [[ $ENV_CHANGE -eq 1 && $TRAVIS_BRANCH == "master" && $TRAVIS_PULL_REQUEST != "false" ]]
+    then
+        echo ""
+	echo "Env matrix was modified but this pull request does not target a bulk/* branch."
+	echo "Please target a branch named bulk/<reasonable-name>. Rebuild and PR against that branch until all builds succeed."
```
bgruening: What is meant with `rebuild`?

johanneskoester: If Travis times out because there are too many packages to be build, you can restart the build job until it has finally processed all recipes.

johanneskoester:
One protected branch bulk is fine. But instead of recreating it, I would rather keep it present and in sync with master. Otherwise we would have to recreate the branch protection whenever we create the branch again. However the issue is fine. I will also move all build system configuration in a separate secured repo, so that only bioconda/core can modify it.

johanneskoester:
I hope to finish all these modifications until tomorrow. Then, I can test them with the Python rebuild over the weekend.

scripts/travis-run.sh
bgruening: Maybe something like that: `PR against that branch and trigger a rebuild in travis until all builds succeed.`?

johanneskoester:
xref #4592.

scripts/travis-run.sh
johanneskoester: Yeah, I will rephrase it a bit.

bgruening:
Ok, I will finish the R migration also tomorrow I hope!

bgruening:
Cool! Thanks!

PR #4589, opened 2017-05-03
VirMet

ozagordi:


bgruening:
You need to increase the build number if you want to have a new package.

ozagordi:
Thanks. I had forgotten a dependency in the previous meta file, so it's not a different version of the tool, just of the recipe. Do I have to change it also in this case?

bgruening:
Yes, for changes like this and to ensure reproducibility we enforce a build number increase. Otherwise we do not build the package and not push it to anaconda.org.

ozagordi:
I see, but then the package will not be in sync with the releases on GitHub. Will v1.1.1 be enough?

bgruening:
I'm talking about the https://conda.io/docs/building/meta-yaml.html#build-number-and-string.
So the version number is still in sync with the github release, but we increase the build-number.

ozagordi:
Ok, thanks. I had missed that.

bgruening:
Great! Thanks!

PR #4462, opened 2017-04-19
Add legacy biopython version 1.67

mvdbeek:
* [x] I have read the guidelines above.
* [ ] This PR adds a new recipe.
* [x] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

PR #4590, opened 2017-05-03
BioConda recipes for Biopython 1.65 and 1.66

peterjc:

We now have BioConda packages for Biopython 1.67 onwards, with 1.67 added as a legacy package in #4462 to help with Galaxy tool dependencies.

Older versions of Biopython 1.65 and 1.66 are still being used by one or more of the IUC maintained Galaxy Tools, and therefore it is expedient to have them available on BioConda as well.

See https://github.com/galaxyproject/tools-iuc/

```
$ grep biopython tools/*/*.xml
tools/jbrowse/macros.xml:      <requirement type="package" version="1.66">biopython</requirement>
tools/kraken_taxonomy_report/kraken_taxonomy_report.xml:        <requirement type="package" version="1.66">biopython</requirement>
tools/ncbi_entrez_eutils/macros.xml:      <requirement type="package" version="1.68">biopython</requirement>
tools/ncbi_entrez_eutils/tool_dependencies.xml:  <package name="biopython" version="1.66">
tools/ncbi_entrez_eutils/tool_dependencies.xml:    <repository name="package_biopython_1_66" owner="biopython"/>
tools/progressivemauve/macros.xml:      <requirement type="package" version="1.65">biopython</requirement>
tools/progressivemauve/tool_dependencies.xml:  <package name="biopython" version="1.65">
tools/progressivemauve/tool_dependencies.xml:    <repository name="package_biopython_1_65" owner="biopython"/>
tools/transtermhp/macros.xml:      <requirement type="package" version="1.65">biopython</requirement>
tools/transtermhp/tool_dependencies.xml:  <package name="biopython" version="1.65">
tools/transtermhp/tool_dependencies.xml:    <repository name="package_biopython_1_65" owner="biopython"/>
```

### DIFF
--- a/recipes/biopython/1.65/build.sh
+++ b/recipes/biopython/1.65/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py install 
--- a/recipes/biopython/1.65/meta.yaml
+++ b/recipes/biopython/1.65/meta.yaml
@@ -1,0 +1,113 @@
+{% set name = "biopython" %}
+{% set version = "1.65" %}
+{% set sha256 = "6d591523ba4d07a505978f6e1d7fac57e335d6d62fb5b0bcb8c40bdde5c8998e" %}
+
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
+  url: https://pypi.python.org/packages/4e/77/8590d61dcda439d83f378106954e748db1a71e565335168a966642133ef8/biopython-1.65.tar.gz
+  sha256: {{ sha256 }}
+
+about:
+  home: http://www.biopython.org/
+  license: Biopython License Agreement
+  license_file: LICENSE
+  summary: 'Freely available tools for computational molecular biology.'
+
+build:
+  number: 0
+  skip: True # [py34]
+
+requirements:
+  build:
+    - python
+    - numpy x.x
+    - reportlab
+    - mmtf-python
+
+  run:
+    - python
+    - numpy x.x
+    - reportlab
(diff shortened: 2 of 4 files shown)


bgruening:
@peterjc have you checked the `default` channel? There are packages from 1.60 onward. And Galaxy can use these packages as well.

peterjc:
I hadn't checked the default channel. Is there no point adding it here then?

I noticed this apparent omission while trying to help a colleague debug a JBrowse failure within Galaxy, where it could not find Biopython. Something was wrong with the dependencies and this seemed a likely cause.

bgruening:
If you are happy with the biopython package from default, there is no point in adding them here yes. The only drawback is that submitting something to default is very hard or impossible.

peterjc:
Where are the default channel packages listed? The do not appear under https://anaconda.org/search?q=type%3Aconda+biopython

bgruening:
Second entry, it's the one under anaconda.

peterjc:
Got it. They may be a better way to see this, but https://anaconda.org/anaconda/biopython/files shows cross-platform binaries for Biopython 1.60 though to 1.68 (current latest release) split over several pages. Thanks.

peterjc:
The TravisCI results are green, so I could merge this - despite it adding little value given there are binary packages on the default channel which ought to work OK?

bgruening:
Yes, you can merge it now. Please note that Conda has something that is called "channel priorities" and in Galaxy BioConda has a higher priority than default (https://github.com/galaxyproject/galaxy/blob/dev/config/galaxy.ini.sample#L213) this means you will get the BioConda version.

peterjc:
Thanks for the training :)

bgruening:
Always!

PR #4591, opened 2017-05-03
Add some common exclusions to the .gitignore

peterjc:

Having accidentally committed a ``meta.yaml~`` file created using emacs, I realised that the BioConda ``.gitignore`` could be expanded a little.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,19 @@
 cached_bioconductor_tarballs
 .web_cache
+
+#Ignore backup files from some Unix editors,
+*~
+*.swp
+*.bak
+
+#Ignore patches and any original files created by patch command
+*.diff
+*.patch
+*.orig
+*.rej
+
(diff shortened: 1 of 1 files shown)


peterjc:
@bgruening is one approval enough for me to merge this (assuming the automated tests pass)?

bgruening:
Yes. For packages you can even merge it on your own if it is green. For those changes please wait for an approval.

bgruening:
Thanks Peter!

acaprez:
Hmm, one of the extensions added here was `.patch`.  There are many existing recipes with patch files named that way.  Given that, I think it would be better to not ignore it.

peterjc:
Good point - BioConda is unusual in that respect, and I hadn't noticed this usage.

(This doesn't stop you adding a patch file, it means you have to do so a little more explicitly)

PR #4609, opened 2017-05-04
change snpeff version to fix conda version sorting

rvalieris:

snpeff current version is "4.3k" but conda is sorting it below "4.3", this PR changes the version to "4.3.0k", that seems to do the trick (see openssl versions).

### DIFF
--- a/recipes/snpeff/meta.yaml
+++ b/recipes/snpeff/meta.yaml
@@ -1,3 +1,8 @@
+{% set snpeff_ver = "v4_3k" %}
+# NOTE: if the version contains a trailing letter, use the <d>.<d>.1<l> format
+{% set version = "4.3.1k" %}
+{% set sha256 = "396fadbefb4994838713c9a3e96393d16e7ec833254a962834e7cbd699b9f35e" %}
+
 about:
   home: 'http://snpeff.sourceforge.net/'
   license: "LGPLv3"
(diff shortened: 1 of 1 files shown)


bgruening:
ping @jj-umn and @nsoranzo

nsoranzo:
According to the Conda documentation, `version` should follow [PEP 440](https://www.python.org/dev/peps/pep-0440/) (and not use the dash `-`), which means that both `4.3k` and `4.3.0k` are invalid. So I don't know how they are actually sorted...

rvalieris:
yep, I saw the PEP440 reference, but [openssl](https://anaconda.org/anaconda/openssl) is using this format and the package is shipped by anaconda itself.

If we could add the "k" by following the PEP440 that would be ideal but as far as I skimmed I don't think so.

nsoranzo:
> yep, I saw the PEP440 reference, but openssl is using this format and the package is shipped by anaconda itself.

`openssl` anaconda packages are just following the upstream release scheme, see https://github.com/openssl/openssl/releases .

I think the problem for `snpeff` is that `4.3` is shorter than `4.3k` and this somehow messes up Conda ordering, in fact `4.3b`, `4.3g`, `4.3i`, `4.3k` are correctly ordered in `conda search`.

rvalieris:
yes I think you are correct

I found the actual code that tests the version order:
```python
from conda.models.version import VersionOrder
>>> VersionOrder('4.3') > VersionOrder('4.3k')
True
>>> VersionOrder('4.3') > VersionOrder('4.3.0k')
True
>>> VersionOrder('4.3') > VersionOrder('4.3.1k')
False
```

conda thinks a letter suffix indicates a "pre-release" version, so this commit won't work... any suggestions ?

rvalieris:
so here is my proposal, since the developer of snpeff [doesn't use 3 digits in the version string](https://sourceforge.net/projects/snpeff/files/), we can modify the version string like this:
```
<d>.<d>    -> no change
<d>.<d><l> -> <d>.<d>.1<l>
```

this will guarantee any version ending with a letter gets sorted after  `<d>.<d>` as the author intended.

ping @bioconda/core

johanneskoester:
What about `<d>.<d>.<l>`?

rvalieris:
`<d>.<d>.<l>` is interpreted as the same as `<d>.<d>.0<l>`

```python
>>> VersionOrder('4.3') > VersionOrder('4.3.k')
True
>>> VersionOrder('4.3.0k') == VersionOrder('4.3.k')
True
```

johanneskoester:
Ok, I see. The solution is fine, but can you please add a not to `extra -> notes`, and a comment above the version key?

bgruening:
Perfect, thanks for the detailed work here!

PR #4611, opened 2017-05-04
remove awscli

rvalieris:

awscli is now available on conda-forge: https://github.com/conda-forge/awscli-feedstock

ping @bioconda/core


bgruening:
Thanks!

rvalieris:
could we remove the package from the channel as well ? so that the conda-forge version can be installed without problems.

johanneskoester:
@rvalieris, yes, but ideally the exact same version as on bioconda should be available in conda-forge then. Otherwise, environments pinning our version can't work anymore.

PR #4616, opened 2017-05-05
graphmap bump version

mictadlo:

### DIFF
--- a/recipes/graphmap/build.sh
+++ b/recipes/graphmap/build.sh
@@ -10,8 +10,8 @@ mkdir -p $PREFIX/bin
 cd $SRC_DIR/codebase
 wget -O seqlib.tar.gz https://github.com/isovic/seqlib/archive/55737bb85100af27c8ae761681e169792dcce115.tar.gz
 tar -xvf seqlib.tar.gz -C seqlib --strip-components 1
-wget -O argparser.tar.gz https://github.com/isovic/argparser/archive/c25e642b7a00d44fb25b9712f404bfa16ab1468e.tar.gz
-tar -xvf argparser.tar.gz -C argparser --strip-components 1
+wget -O argparser.tar.gz https://github.com/isovic/argparser/archive/72af9764acefbcc92ff76e3aba8a83d9a3e33671.tar.gz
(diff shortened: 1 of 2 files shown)


mictadlo:
HI @bgruening and @druvus,
During update graphmap I got 
[tar: argparser: Cannot chdir: No such file or directory](https://travis-ci.org/bioconda/bioconda-recipes/jobs/228951904#L2862). Any Idea why?

Thank you in advance.

Michal

druvus:
@mictadlo @bgruening We need to fix the version for graphmap. The original `0.22.1`is ranked higher than the later `0.4.0`. Should we change to suggested date variant or fix it in some other way?

druvus:
@bgruening Do you have any suggestion how to fix it? The issue is that the author of the package changed version naming 0.22 vs 0.3.0.

bgruening:
Puh, I just invent here something ... maybe `0.22.1.4`, so we can switch easily back if the maintainer fixes this problem?
Either way, we need to add a big comment :( and fill an upstream bug report.

druvus:
The version that is causing the problem is ancient. They have released the following versions since then: `0.22->0.3.0->0.3.1->0.3.2->0.4.0->0.4.1->0.5.0->0.5.1`

@bgruening is it possible to move a single version to bioconda-legacy? I know it is not good to remove a version but the fast changes for nanopore sequencing probably make it hard to use tho old 0.22 version.  So it would be better to keep the current version naming of the authors
. 
@mictadlo I checked your graphmap commit you are using and it is actually a downgrade.

bgruening:
@druvus we can not move a single version, sorry. But we can remove it ...

druvus:
I would vote to close this PR as it is an older version than the newest present in Bioconda, i.e. not a bump. We should instead try to get the latest 0.5.2 version into Bioconda.

I also think it would be best to remove the 0.22 version as it confuse most people that want the latest version.

kyleabeauchamp:
I have a PR open for the latest but it needs additional fixes to get working

druvus:
Graphmap is bumped to latest version 0.5.2 in #4691 so I am closing this PR.

PR #4621, opened 2017-05-05
Add bioconductor-bsgenome.hsapiens.ucsc.hg38 recipe

bow:

This PR adds a new recipe for the bioconductor BSgenome.Hsapiens.UCSC.hg38 package.


bow:
@bioconda/r I'm sorry to have to ping you all, but I've been stuck building this particular bioconductor package. I hope you don't mind having a look.

As you can see from [the Travis log](https://travis-ci.org/bioconda/bioconda-recipes/jobs/232804093#L979), the build fails because it could not find the R executable required for building. I can't figure out why this is the case, though.

Here's what I've tried (my local box is Linux and I've tried these using bioconda-utils on the v0.10.0 tag):

* `conda build recipes/bioconductor-bsgenome.hsapiens.ucsc.hg38`, with the latest conda-build (version 2.1.14) and also the one used by the Docker container (version 2.0.7). In both cases, the package could be built without errors.

* `DISABLE_BIOCONDA_UTILS_BUILD_GIT_RANGE_CHECK=true ./simulate-travis.py --packages bioconductor-bsgenome.hsapiens.ucsc.hg38 --docker --loglevel=debug`, which fails with the same error in Travis.

* Various combinations of specifying build dependencies in the `meta.yml` and also playing around with `build.sh` (I removed the `grep` command since I didn't notice anything to grep out of the file). All resulted in the same error as Travis.

* Running `simulate-travis.py` with `--disable-docker`. This got the package built, but the test failed saying it couldn't find the `mulled-test` directory. I suppose this is unrelated.

Just to be sure that older packages can still be built, I also tried doing the same on bsgenome hg19 (same command as above, only using `hg19` instead of `hg38`. I increased the build number in its `meta.yml` as well.

To my surprise, I also got the same error there as when I tried building `hg38`: the R executable can not be found.

When I checked the actual path where the R executable is supposed to be found (inside the Docker container, just by doing `ls` in `build.sh`), the R executable is indeed missing.

I suppose something is off in the Docker container (why aren't the build dependencies found?), something that was present when the hg19 package was built, but has since been removed or changed. I can't figure out what this is, unfortunately.

Any ideas on what I should check next?

dyusuf:
I guess it might be related to the conflict of R versions.

the current bioconductor is 3.5 and requires R 3.4. http://bioconductor.org/news/bioc_3_5_release/

Well, current bioconda build allows 3.3.1. And there is not yet the conda pkg for R  3.4* in channel r or conda-forge.

may be, wait for R 3.4 to appear in the conda channels?

@bioconda/core any inputs?

bow:
Hmm..that may be related. But if this is the case, shouldn't the R executable at least be installed and/or conda-build exits with some dependency error message?

dyusuf:
yes, the valid question while I don't have the answer. 

To see if the R version does play effect, one thing you could try is to make a test build with simulate-travis.py (without disabling docker) by using a package from CRAN which should be compatible with 3.3*. Let's see whether the same error can be reproduced.

So far, the testset only includes the bioconductor pkgs that clearly require 3.4.

bow:
Hmm...this is interesting. So I did the test again with `r-dplyr`, which is pulled from CRAN and is supposed to be compatible with other R versions (at least no specific R version is specified in its `meta.yaml`). Here's the command:

```
DISABLE_BIOCONDA_UTILS_BUILD_GIT_RANGE_CHECK=true ./simulate-travis.py --packages r-dplyr --docker --loglevel=debug
```

The build succeeded, but then I got the same mulled-build error as when I ran the bsgenome.hg38 recipe with `--disable-docker`:

```
...
FileNotFoundError: [Errno 2] No such file or directory: 'mulled-build'
```

So far it does look like it's related to the R version.

But then I tried changing bsgenome hg38's `meta.yaml` to use the 3.3 version (with this URL: https://bioconductor.org/packages/3.3/data/annotation/src/contrib/BSgenome.Hsapiens.UCSC.hg38_1.4.1.tar.gz and the correct `sha1` checksum) and then trying the build again with this command:

```
./simulate-travis.py --docker --loglevel=debug
```

... and it still errors out with a missing R executable :(.

daler:
@bow can you try changing `r` to `r-base` in the build and run requirements to see if that helps?

bow:
@daler Still the same error, unfortunately (shown below, I just realized I haven't pased the actual error line):

```
+ source /opt/conda/bin/activate /opt/conda/conda-bld/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho
+ /opt/conda/conda-bld/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/bin/R CMD INSTALL --build .
/opt/conda/conda-bld/work/BSgenome.Hsapiens.UCSC.hg38/conda_build.sh: line 3: /opt/conda/conda-bld/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/bin/R: No such file or directory
```

It is still pulling the `r` package from conda even when I only specified `r-base`, since `r` is a dependency of `bioconductor-bsgenome`.

daler:
Hmm. @bgruening do you think a change from `r` to `r-base` for all dependencies would solve this? Or is it something else?

johanneskoester:
I don't get this. `r` is just a metapackage pulling in `r-base`. And we see in the travis log that it indeed installs `r-base` before building the package. So why is it not present when the build is started??

daler:
I'm stumped as well :/. Will dig deeper when I get a chance.

johanneskoester:
Could it be that it somehow uses a leaked build environment from another recipe? But there is no other build in this travis instance...

bow:
So after playing around more with this, I could finally get it past the build stage by manually changing the `DEFAULT_CONDA_BUILD_VERSION` variable to 2.1.15 (the latest version released 5 days ago) in bioconda-utils' `docker_utils.py` (still off the v0.10.0 tag).

It's still not completed successfully (I get the mulled-build errors now), but I guess then the cause is related to `conda-build`'s version. Not sure why still (since I played with building other non-bsgenome bioconductor packages on 2.0.7 and they all work fine).

druvus:
@bow the package was added in #5865.

PR #4637, opened 2017-05-08
Virmet

ozagordi:


daler:
py36 will soon be available (see the `bulk` branch) but I suspect we need to be using a newer version of conda to recognize the `# [py36]` selector. There are issues with newer conda versions preventing us from upgrading, so for now the way you have it here is the way to go.

ozagordi:
Thanks!

PR #4638, opened 2017-05-08
Update samtools/htslib to 1.4.1.

dpryan79:
Samtools may fail due to https://github.com/samtools/samtools/issues/679, we'll see.


dpryan79:
This will have to wait for the release tarballs to get fixed (otherwise `autoreconf` will need to get run to fix things).

PR #4661, opened 2017-05-10
Biopet add r dependencies for reports

rhpvorderman:

For the reports biopet depends on R. These dependencies are now added to the package.


recipes/biopet/0.7.0/meta.yaml
```diff
@@ -15,6 +15,9 @@ requirements:
   run:
     - openjdk
     - python
+    - r
```
bgruening: Can you use `r-base` here and in the other places?

rhpvorderman:
I see travis is failing because of disk size: `OSError: [Errno 28] No space left on device`. Should I be more specific than r-essentials?

bgruening:
Yes, please.

rhpvorderman:
Done! Thanks for reviewing!

PR #4668, opened 2017-05-10
add getopt

bgruening:

### DIFF
--- a/recipes/bioconductor-diffbind/meta.yaml
+++ b/recipes/bioconductor-diffbind/meta.yaml
@@ -6,7 +6,7 @@ source:
   url: http://bioconductor.statistik.tu-dortmund.de/packages/3.3/bioc/src/contrib/DiffBind_2.0.9.tar.gz
   md5: 9d00cf8d809f138ce2a3b1e5620bbf51
 build:
-  number: 0
+  number: 1
   rpaths:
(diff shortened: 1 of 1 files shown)


notestaff:
@bgruening 
What does it mean that the CRON job failed, even after the normal travis-ci tests passed and the pull request was successfully merged?  I'm seeing the same thing with my last commit, and it also seems to have happened with other commits.  Is it ok to ignore the CRON error?

bgruening:
Which cron error? Can you give me a link?

notestaff:
I'm re-running it now,  at
https://travis-ci.org/bioconda/bioconda-recipes/builds/232734919
but it already seems to take more time than normal.  The CRON error was that it timed out, not that the tests failed.  The pull request I merged is
https://github.com/bioconda/bioconda-recipes/pull/4701

bgruening:
but I see your package here: https://anaconda.org/bioconda/gap2seq and here https://quay.io/repository/biocontainers/gap2seq?tab=tags so everything seems to be fine. The cron job just runs to pick up packages that we missed for some reason and is just a safeguard atm. Do you miss any packages?

notestaff:
No, I'm able to install them and they work, I just wanted to understand if
the CRON error indicates there's some problem.  Sounds like it doesn't.

PR #4670, opened 2017-05-11
Pairix recipe

nvictus:

Adding a new recipe for [pairix](https://github.com/4dn-dcic/pairix) by @SooLee and @carlvitzthum. Please review, @bioconda/core.  Thanks.

### DIFF
--- a/recipes/pairix/meta.yaml
+++ b/recipes/pairix/meta.yaml
@@ -14,14 +14,15 @@ build:
   
 requirements:
   build:
-    - gcc
+    - gcc  # [linux]
+    - llvm # [osx]
     - zlib
     - python
     - setuptools
 
   run:
-    - python
     - htslib
+    - python
 
 test:
   imports:
--- a/recipes/pairix/meta.yaml
+++ b/recipes/pairix/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "0.1.5" %}
+{% set version = "0.1.6" %}
 
 package:
   name: pairix
@@ -7,7 +7,7 @@ package:
 source:
   fn: {{ version }}.tar.gz
   url: https://github.com/4dn-dcic/pairix/archive/{{ version }}.tar.gz
-  md5: 05ac0be28a85e4d6bdca744ddec8824e
+  md5: 4e8f9fa5478e2f2c594177025f308854
 
 build:
   number: 0
@@ -28,7 +28,7 @@ test:
   imports:
     - pypairix
   commands:
-    - which pairix
+    - pairix --help
 
 about:
   home: https://github.com/4dn-dcic/pairix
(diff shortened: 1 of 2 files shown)


recipes/pairix/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+requirements:
+  build:
+    - gcc
+    - zlib
+    - python
+    - setuptools
+
+  run:
+    - python
+    - htslib
```
bgruening: Mh, is zlib really needed as build dependency if you don't have it in the run dependency?

recipes/pairix/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+  fn: {{ version }}.tar.gz
+  url: https://github.com/4dn-dcic/pairix/archive/{{ version }}.tar.gz
+  md5: 05ac0be28a85e4d6bdca744ddec8824e
+
+build:
+  number: 0
+  
+requirements:
+  build:
+    - gcc
```
bgruening: Can you use `gcc  # [linux]` and `llvm  # [osx]`

recipes/pairix/meta.yaml
bgruening: Why is htslib needed and not in the build deps?

recipes/pairix/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+  run:
+    - python
+    - htslib
+
+test:
+  imports:
+    - pypairix
+  commands:
+    - which pairix
```
bgruening: Is something like `parix --help` possible?

bgruening:
Thanks @nvictus?

recipes/pairix/meta.yaml
nvictus: Will do!

recipes/pairix/meta.yaml
nvictus: Pairix is meant to work with bgzip, which is provided by htslib. Pairix builds a copy of the bgzip executable, but I figured it is better to use the official htslib version and overwrite it.

recipes/pairix/meta.yaml
nvictus: I couldn't find a benign option that returns an exit code of 0. @SooLee, does one exist or would it be possible to add a `--help` option that exits with 0?

Originally I tried to pull in sample files and run a test using:
```
  source_files:
    - samples/test_4dn.pairs.gz
    - samples/test_4dn.pairs.gz.px2
  commands:
    - pairix -l samples/test_4dn.pairs.gz
```
but never got it to work. The build system couldn't find the source files.

daler:
@nvictus you can test for other exit codes, e.g., https://github.com/bioconda/bioconda-recipes/blob/master/recipes/asciigenome/meta.yaml#L19

nvictus:
Thanks, @daler! They ended up adding a `--help` option that exits with 0 in the end, but great to know.

bgruening:
Cool, thanks!

PR #4671, opened 2017-05-11
Fixes for openbabel 2.4.1

nsoranzo:

This can be of interest for @mwojcikowski.

### DIFF
--- a/recipes/openbabel/2.4.1/meta.yaml
+++ b/recipes/openbabel/2.4.1/meta.yaml
@@ -12,9 +12,10 @@ source:
         - fix_library_path_search.diff
         - fix_data_path.diff
         - fix_babel_test_paths.diff
+        - fix_libinchi_cmakelists.diff
 
 build:
-    number: 0
+    number: 1
     detect_binary_files_with_prefix: true
 
 requirements:
@@ -40,7 +41,6 @@ test:
     - openbabel
     - pybel
   commands:
-    - python run_test.py
     - obabel --help
     - obabel -:'c1ccccc1' --gen3d -omol2
     - obabel -:'c1ccccc1' --gen3d -oinchi
--- a/recipes/openbabel/2.4.1/meta.yaml
+++ b/recipes/openbabel/2.4.1/meta.yaml
@@ -13,6 +13,7 @@ source:
         - fix_data_path.diff
         - fix_babel_test_paths.diff
         - fix_libinchi_cmakelists.diff
+        - 372.patch
 
 build:
     number: 1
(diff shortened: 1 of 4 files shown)


recipes/openbabel/2.4.1/meta.yaml
```diff
@@ -35,12 +37,15 @@ requirements:
         - zlib
         - libxml2
         - cairo   # [linux]
+        - xorg-libxau
```
mwojcikowski: Why do you need any xorg packages for OB? Are you trying to build OB-GUI? I don't think conda is a good place for it.

mwojcikowski:
Other than the `xorg-*` packages everything looks fine to me.

recipes/openbabel/2.4.1/meta.yaml
nsoranzo: During testing using Bioconda `./simulate_travis.py`, it fails due to errors like:
```
[May 10 16:20:25] SERR ==============================
[May 10 16:20:25] SERR *** Open Babel Error  in openLib
[May 10 16:20:25] SERR /usr/local/lib/openbabel/2.4.1/daltonformat.so did not load properly.
[May 10 16:20:25] SERR Error: libXrender.so.1: cannot open shared object file: No such file or directory
```
After adding `xorg-libxrender` to the run requirements, I got errors for missing `libXext.so.6` and so on until I added all these.

Also, if you download the present bioconda package https://anaconda.org/bioconda/openbabel/2.4.1/download/linux-64/openbabel-2.4.1-py35_0.tar.bz2 and run e.g. `ldd lib/openbabel/2.4.1/daltonformat.so`, you'll see that all these xorg libraries are already required.

nsoranzo: See e.g. https://travis-ci.org/bioconda/bioconda-recipes/jobs/230372715 , where openbabel is a run requirement for rDock (and actually the reason for this PR).

mwojcikowski: I would try to get rid of those rather by addition of `-DBUILD_GUI=OFF`, than including them. Can you point me to the travis job building OB packages?

mwojcikowski: Our upstream channel does not require `xorg-*`, see https://travis-ci.org/openbabel/conda-openbabel/jobs/216283344#L368 

If we explicitly disable GUI it will not link libraries against it. Apparently bioconda's docker image has wxWidgets and the GUI is compiled by default.

mwojcikowski: I was wrong, X packages are linked via `libcairo.so` and are present upstream. Although I think such dependency is out of conda's scope. 

Those packages are available only in `conda-forge` repo. Is it now a "standard" dependency for bioconda channel?

bgruening: Yes, BioConda is depending now on conda-forge and the rule is to put general purpose packages there.
Thanks for your review  @mwojcikowski!

PR #4676, opened 2017-05-11
Adds SCAN.UPC recipe.

kurtwheeler:

This PR is a WIP because I can't test this recipe. Both `simulate-travis.py` actually running [travis](https://travis-ci.org/kurtwheeler/bioconda-recipes/jobs/231194721) throw this error:
```
  File "/home/kurt/miniconda3/bin/bioconda-utils", line 11, in <module>
    load_entry_point('bioconda-utils==0.10.0', 'console_scripts', 'bioconda-utils')()
  File "/home/kurt/miniconda3/lib/python3.5/site-packages/bioconda_utils/cli.py", line 351, in main
    argh.dispatch_commands([build, dag, dependent, lint])
  File "/home/kurt/miniconda3/lib/python3.5/site-packages/argh-0.26.2-py3.5.egg/argh/dispatching.py", line 328, in dispatch_commands
  File "/home/kurt/miniconda3/lib/python3.5/site-packages/argh-0.26.2-py3.5.egg/argh/dispatching.py", line 174, in dispatch
  File "/home/kurt/miniconda3/lib/python3.5/site-packages/argh-0.26.2-py3.5.egg/argh/dispatching.py", line 277, in _execute_command
  File "/home/kurt/miniconda3/lib/python3.5/site-packages/argh-0.26.2-py3.5.egg/argh/dispatching.py", line 260, in _call
  File "/home/kurt/miniconda3/lib/python3.5/site-packages/bioconda_utils/cli.py", line 285, in build
    disable_travis_env_vars=disable_travis_env_vars,
  File "/home/kurt/miniconda3/lib/python3.5/site-packages/bioconda_utils/build.py", line 255, in build_recipes
    for i in utils.newly_unblacklisted(orig_config, recipe_folder)
TypeError: newly_unblacklisted() missing 1 required positional argument: 'git_range'
```
Which makes sense because https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/build.py#L263 calls `utils.newly_unblacklisted` with only 2 arguments but that function requires three arguments: https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/utils.py#L484.

It looks like https://github.com/bioconda/bioconda-utils/pull/95 may be fixing it, but really it's just removing the `--quick` option.

I'm not sure how to get past that error, but I was able to build the recipe locally, upload it to my own conda channel, and use it.


kurtwheeler:
@bgruening and @johanneskoester, do either of you know how to fix the build error I seem to have run into? You can take a look at the job itself for more context but the pertinent part is:

```
  - r-base >=3.3.2
```

Conda seems to be saying that `r-base` is in conflict with itself. This doesn't make much sense to me, especially since many other packages rely on `r-base`. In fact I am inclined to think that this may even be a Conda bug because I wouldn't think it'd ever be valid for Conda to say there was a specification conflict with only a single package. I'm considering opening an issue with conda/conda-build about this, but I'll wait to hear what you say about it first.

I've gotten this error for a while on my own machine, but it's taken a while to get this PR clean enough to the point where it's output in Travis. However when I was playing around with this recipe I managed to get the package to build, but then when trying to install it I got a similar error except like:

```
UnsatisfiableError: The following specifications were found to be in conflict:
  - bioconductor-scan.upc
```

daler:
I've had similar problems before and this is apparently a known issue with conda: https://github.com/conda/conda/issues/3622. Makes it hard to track down the underlying problem, but `bioconductor-scan.upc` seems as good a lead as any.

kurtwheeler:
@daler Thanks for pointing that out. It looks like it's not an issue in the newest version of conda and conda-build. I've gotten more meaningful output on my machine whereas previously I was stuck.

kurtwheeler:
@bgruening and @johanneskoester I've gotten the tests passing. This is ready for a maintainer to take a look at.

recipes/bioconductor-oligo/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+build:
+  script: R CMD INSTALL --build .
+  number: 0
+  skip: True # [osx]
+  rpaths:
+    - lib/R/lib/
+    - lib/
+requirements:
+  build:
+    - 'r-base >=3.2.2'
```
bgruening: It's not a good idea to pin it here. But I guess you need to to work around this bug?

recipes/bioconductor-scan.upc/LICENSE
```diff
@@ -0,0 +1,8 @@
+MIT License
+Copyright (c) 2012, Stephen R. Piccolo, Andrea H. Bild, and W. Evan Johnson
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
```
bgruening: This file can be removed.

kurtwheeler: Can it? The license text below states `The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.`

bgruening: Yes, this is just a recipe. If you want to ship the license with the package you should use: https://conda.io/docs/building/meta-yaml.html#license-file

kurtwheeler: Ah okay, I've done that now. So I should leave this file in the recipe right?

bgruening: No you can remove it. `license_file:` will copy it from the tarball.

kurtwheeler: Ah I understand now. Thanks.

johanneskoester:
Note that Travis also has a restart button ;-)

kurtwheeler:
I don't think I have permission to click it for the bioconda-recipes travis.

kurtwheeler:
@johanneskoester or @bgruening, I think this is ready to be merged.

bgruening:
@kurtwheeler looks good to me!

PR #4680, opened 2017-05-12
Update pysam recipe

kyleabeauchamp:


kyleabeauchamp:
cc @bioconda/core thoughts on why this recipe is failing due to a failure on a previous version?  https://travis-ci.org/bioconda/bioconda-recipes/jobs/231955784#L1030

I'm trying to build 0.11.2 but the build is failing on 0.9.1.4

bgruening:
@kyleabeauchamp I guess this a bug in the rebuild logic of the py36 changes recently. You can put it on a blacklist if you want or remove the old recipe from git.

johanneskoester:
+1 for removing the recipe. bioconda-utils should only consider changed recipes here, but it looks like it is considering all old versions as well if the primary recipe changed.

PR #4687, opened 2017-05-15
added weeder2 recipe

simonvh:

### DIFF
--- a/recipes/weeder/build.sh
+++ b/recipes/weeder/build.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+TGT="$PREFIX/share/weeder2"
+[ -d "$TGT" ] || mkdir -p "$TGT"
+[ -d "${PREFIX}/bin" ] || mkdir -p "${PREFIX}/bin"
+
+cd "${SRC_DIR}"
+g++ weeder2.cpp -o weeder2 -O3
+cp -r FreqFiles $TGT/
+cp weeder2 $TGT/
+
+cp $RECIPE_DIR/weeder2.py $TGT/
+ln -s $TGT/weeder2.py $PREFIX/bin/weeder2
+chmod 0755 "${PREFIX}/bin/weeder2"
--- a/recipes/weeder/build.sh
+++ b/recipes/weeder/build.sh
@@ -1,12 +1,11 @@
 #!/bin/bash
 
 TGT="$PREFIX/share/weeder2"
-[ -d "$TGT" ] || mkdir -p "$TGT"
-[ -d "${PREFIX}/bin" ] || mkdir -p "${PREFIX}/bin"
+mkdir -p "$TGT"
+mkdir -p "${PREFIX}/bin"
 
 cd "${SRC_DIR}"
 g++ weeder2.cpp -o weeder2 -O3
-cp -r FreqFiles $TGT/
 cp weeder2 $TGT/
 
 cp $RECIPE_DIR/weeder2.py $TGT/
--- a/recipes/weeder/meta.yaml
+++ b/recipes/weeder/meta.yaml
@@ -4,7 +4,7 @@ package:
 
 source:
   fn: weeder2.0.tar.gz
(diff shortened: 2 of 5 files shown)


recipes/weeder/build.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+TGT="$PREFIX/share/weeder2"
+[ -d "$TGT" ] || mkdir -p "$TGT"
```
bgruening: if you do `mkdir -p` you don't need this if condition

recipes/weeder/meta.yaml
```diff
@@ -0,0 +1,32 @@
+package:
+  name: weeder
+  version: "2.0"
+
+source:
+  fn: weeder2.0.tar.gz
+  url:  http://159.149.160.51/modtools/downloads/weeder2.0.tar.gz
```
bgruening: That does not look like a stable URL. Bioconda will backup the tarball for you, but you should try to provide a stable URL at first hand.

recipes/weeder/build.sh
```diff
@@ -0,0 +1,14 @@
...
+[ -d "${PREFIX}/bin" ] || mkdir -p "${PREFIX}/bin"
+
+cd "${SRC_DIR}"
+g++ weeder2.cpp -o weeder2 -O3
+cp -r FreqFiles $TGT/
```
bgruening: How big are these `FreqFiles`

recipes/weeder/meta.yaml
simonvh: I don't like it either, but this is the official site...

recipes/weeder/build.sh
simonvh: 285M

bgruening: Thats big. Could you use the post installation hooks to download this DB? here is one small example: https://github.com/bioconda/bioconda-recipes/blob/2d8dcb197920588ddc6c27195fe9a2d41ddc3c1f/recipes/augustus/3.2.2/build.sh#L29

recipes/weeder/meta.yaml
bgruening: You can get a stable API here: https://github.com/galaxyproject/cargo-port

recipes/weeder/build.sh
simonvh: The files are included in the source download, but this would be with the aim of reducing the size of the conda package, right?

I don't understand the relevance of your example. Do you mean adding a `post-link.sh`, like https://github.com/bioconda/bioconda-recipes/blob/master/recipes/krona/post-link.sh?

bgruening: Yes, reducing the conda package size is the aim. This example also works - the idea is just that you trigger the download post-installation or give hints how a user can get the datasets.

simonvh:
Ok, @bgruening thanks for the suggestions. For my part it's ready to merge.

recipes/weeder/post-link.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Directory to store FreqFiles
+TGT="$PREFIX/share/weeder2"
+
+# Cargo port url
+CARGO_URL=http://depot.galaxyproject.org/software/
+
+# Download and unpack
+TARBALL=${PKG_NAME}_${PKG_VERSION}_src_all.tar.gz
+curl -L ${CARGO_URL}/${PKG_NAME}/$TARBALL -o $TGT/$TARBALL
```
bgruening: @simonvh have you tested this locally? Is `PKG_NAME` available at install time? This would be great, was not aware of this.

simonvh: Yes, see [here](https://conda.io/docs/building/build-scripts.html). There's a small set of env variables available: `PREFIX`,  `PKG_NAME`, `PKG_VERSION` and `PKG_BUILDNUM`.

bgruening:
Awesome! Thanks a lot!

PR #4689, opened 2017-05-15
Remove retrying because conda-forge has a better one

ostrokach:
Installing retrying from bioconda gives a warning:

```
Package bioconda::retrying-1.3.3-py_1 uses a pre-link script. Pre-link scripts are potentially dangerous discouraged.
This is because pre-link scripts have the ability to change the package contents in the
package cache, and therefore modify the underlying files for already-created conda
environments. Future versions of conda may deprecate and ignore pre-link scripts.
```

This PR removes for retrying because it is uses a prelink script and it's easier to use the package from conda-forge than to update this one.


bgruening:
Thanks @ostrokach! Does cona-forge have the exact same version as well?

ostrokach:
Yup! Only bioconda's build version is higher so it took precedence.

bgruening:
Ok, moved it to bioconda-legacy.

PR #4697, opened 2017-05-15
umi_tools: updated to version 0.4.4

croth1:

Should also fix an issue when running umi_tools in an environment without write access. 

xref: https://github.com/CGATOxford/UMI-tools/issues/126

### DIFF
--- a/recipes/umi_tools/meta.yaml
+++ b/recipes/umi_tools/meta.yaml
@@ -1,11 +1,11 @@
-{% set version = "0.4.3" %}
+{% set version = "0.4.4" %}
 
 package:
   name: "umi_tools"
   version: {{ version }}
 source:
   url: https://pypi.io/packages/source/u/umi_tools/umi_tools-{{ version }}.tar.gz
-  sha256: 69bea4125591df11a44c8d95f6c9ede29f4b0e09f016d45b0ace5f5c5a278ff7
+  sha256: feb1903c515f3d5a081ea4b3a74ab2d4904b687d7fc1858e0bf2d8b60a413586
 build:
   number: 0
 
@@ -28,6 +28,7 @@ requirements:
 test:
   commands:
     - umi_tools --help
+    - test -f $SP_DIR/umi_tools/_dedup_umi.c
 about:
   home: https://github.com/CGATOxford/UMI-tools
   license: MIT License
--- a/recipes/umi_tools/build.sh
+++ b/recipes/umi_tools/build.sh
@@ -1,3 +1,6 @@
 #!/bin/bash
 
 $PYTHON setup.py install
+
+# this will create a c file for the extension modules
+umi_tools dedup --help %> /dev/null


croth1:
Any idea where this might come from? Is it a conda-build problem?

> /opt/conda/conda-bld/recipe_1494910573193/work/umi_tools-0.4.4/conda_build.sh: /opt/conda/conda-bld/recipe_1494910573193/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/bin/umi_tools: /opt/conda/conda-bld/recipe_1494910573193/_b_env_placehold_placehold_placehold: bad interpreter: No such file or directory

bgruening:
@croth1 this usually indicates that some intepreter given in the shebang is not found. This can be for example /usr/bin/perl, which need to be /usr/bin/env perl ....

croth1:
@bgruening, in this case the command line scripts are installed by python itself and thus should have a proper interpreter. The error message hints towards the interpreter actually being cut off by a maximum length.

croth1:
@bioconda/core, any idea what's the problem here? Is this a problem with an older conda version?

bgruening:
@croth1 even when the scripts are installed by python if you have `#!/usr/bin/python` as your first line it will result in an error as above.

croth1:
setuptools is magical, it does that automatically 👍 

This is what I get when I install umi_tools in the sb27 environment:
```
#!/Users/ch/miniconda/envs/sb27/bin/python
# EASY-INSTALL-ENTRY-SCRIPT: 'umi-tools==0.4.4','console_scripts','umi_tools'
__requires__ = 'umi-tools==0.4.4'
import sys
from pkg_resources import load_entry_point

if __name__ == '__main__':
    sys.exit(
        load_entry_point('umi-tools==0.4.4', 'console_scripts', 'umi_tools')()
    )
```

bgruening:
And if you relocate this binary it will not work ;) 
Just saying, one thing I would try is that you have a correct shebang :)

recipes/umi_tools/meta.yaml
```diff
@@ -28,6 +28,7 @@ requirements:
 test:
   commands:
     - umi_tools --help
+    - test -f $SP_DIR/umi_tools/_dedup_umi.c
```
bgruening: @croth1 this SP_DIR is also not available during the tests I'm afraid.

croth1: well the confusing thing is: it does work on my mac. But I have conda `conda 4.3.18`, maybe bioconda's version is older?

croth1:
```
source /Users/ch/miniconda/bin/activate /Users/ch/miniconda/conda-bld/umi_tools_1496828419265/_t_env
+ /bin/bash -x -e /Users/ch/miniconda/conda-bld/umi_tools_1496828419265/test_tmp/run_test.sh
+ umi_tools --help

umi_tools.py - Tools for UMI analyses
===============================================

:Author: Tom Smith & Ian Sudbury, CGAT
:Release: $Id$
:Date: |today|
:Tags: Genomics UMI

There are 3 tools:
  - extract
  - group
  - dedup

To get help on a specific tool, type:

    umi_tools <tool> --help

To use a specific tool, type::

    umi_tools <tool> [tool options] [tool arguments]

+ test -f /Users/ch/miniconda/conda-bld/umi_tools_1496828419265/_t_env/lib/python2.7/site-packages/umi_tools/_dedup_umi.c
```

This is what I get here, everything looks fine.

bgruening: True, but we are also testing in a Docker container without conda installed, so this magic variable does not exists there. But your tests are failing earlier.
Can you try with these versions: https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/bioconda_utils-requirements.txt#L5

croth1: hmhm... builds on mac - cannot test linux for now

PR #4699, opened 2017-05-15
Remove sed duplicated in conda-forge

nsoranzo:
Newer sed 4.4 is now available in conda-forge.


bgruening:
Let's remove it from here, however I will leave the package in bioconda as the same version is not in conda-forge. People that needs the new version should pin it with `>=4.4`.

nsoranzo:
Agreed!

PR #4710, opened 2017-05-16
uploaded secapr illumina processing pipeline

tobiashofmann88:
@marcelm

### DIFF
--- a/recipes/secapr/meta.yaml
+++ b/recipes/secapr/meta.yaml
@@ -9,7 +9,7 @@ about:
 
 source:
   url: https://github.com/AntonelliLab/seqcap_processor/archive/v1.0.1.tar.gz
-  sha: 7d8a2bdc145fc704c90839c4a066c598d6d1a3514848ecd0399255bc36b19fb4
+  sha256: 7d8a2bdc145fc704c90839c4a066c598d6d1a3514848ecd0399255bc36b19fb4
 
 build:
   skip: True  # [not py27]
--- a/recipes/secapr/meta.yaml
+++ b/recipes/secapr/meta.yaml
@@ -1,0 +1,45 @@
+# build with: conda build --python 2.7 .
+
+package:
+  name: secapr
+  version: '1.0.0'
+
+about:
+  home: 'https://github.com/AntonelliLab/seqcap_processor'
+  license: none
+  summary: 'Process sequence-capture fastq files into alignments for phylogenetic analyses'
+
+source:
+  url: https://github.com/AntonelliLab/seqcap_processor/archive/v1.0.0.tar.gz
+
+build:
+  skip: True  # [not py27]
+  number: 0
+  script: $PYTHON setup.py install
+
+requirements:
+  build:
+    - python
+    - biopython
+  run:
+    - python
+    - biopython
+    - emboss
+    - bcftools
+    - samtools
+    - picard
+    - seqtk
+    - bwa
+    - lastz
+    - mafft
+    - muscle
+    - cogent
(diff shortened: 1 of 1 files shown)


marcelm:
The recipe still refers to version 1.0.0. Is it possible you forgot to commit or push the most changes you made?

Also, if you click on 'Details' above, you can see that the 'recipe linter' complains about a missing hash. You’ll need to add this to the recipe. To do so, download the tar.gz file manually and then run the program `sha256sum` on the downloaded file, perhaps like this:
```
wget https://github.com/AntonelliLab/seqcap_processor/archive/v1.0.1.tar.gz
sha256sum v1.0.1.tar.gz
```
Then add a `sha256` key to the `source` section of the recipe:
```
source:
  url: ...
  sha256: 1234567890abcdef
```
Replace `1234567890abcdef` with the actual checksum.

marcelm:
@bioconda/core This recipe fails to build and I have no idea why. (The Travis log has always been convoluted, but it seems to me this has reached an entirely new level now!) @tobiashofmann88 and I would appreciate some help.

rvalieris:
it seems the logs are exceeding the 4MB limit and travis killed the job.

funny thing is that most of the logs consist of conda-install progress bars, @bioconda/core couldn't we silence these progress bars by passing `--quiet` to conda install ?

marcelm:
Thanks a lot! Yes, I saw that message, but thought it must have been mistaken because I wouldn’t expect around 700 lines to fill 4 MB. The package has quite a few dependencies, but not an excessive amount, shouldn’t there be other packages with the same problem?

bgruening:
We are building here 3 packages for 3 python versions, that's a little bit much. This will be fixed with the new bioconda-utils release and we will add a --quite for the install of packages.
Until than you could skip the py34 build in a first run and remove it later in a follow up PR.

marcelm:
Hm, that should not be the reason because this package is only built for Python 2.7.

Isn’t it also strange that it takes 39 minutes to build? There is no compilation. Building the package locally takes less than 1 minute (granted with all the dependencies downloaded already).

bgruening:
Oh, true. It is still building the package 3 times as it seems. Try searching for `secapr-1.0.1-py27_0.tar.bz` ...

marcelm:
You are right! Those three `BIOCONDA INFO BUILD SUCCESS` lines are actually identical. That explains also the time. Anything I can do to help? I’m not familiar with the bioconda-utils internals.

bgruening:
@marcelm it's obviously some bug, I need to look closer into it and check a few other PR from the last days.

bgruening:
Seems to be a general problem and you are just unlucky, because you download that much data :(

marcelm:
Thanks a lot! It’s not urgent.

marcelm:
@bioconda/core The build passes now! Thanks to whoever fixed this. This is a new recipe and it would be great if someone could have a look and merge it if it is fine.

johanneskoester:
Looks good to me!

johanneskoester:
Thanks!

marcelm:
Thanks a lot!

johanneskoester:
There was an error when building the osx version on master. The error is unrelated to your package and caused by a misconfiguration of our build system that I fixed now. However, the osx version of your package will appear with the next cron build tomorrow morning if that is ok.

marcelm:
@tobiashofmann88 has been patient so far, it should be fine. Thanks for letting us know.

tobiashofmann88:
Nice thanks a lot for your help!

PR #4733, opened 2017-05-18
perl-bio-db-hts update

matthdsm:
version bump 2.4 > 2.7

### DIFF
--- a/recipes/perl-bio-db-hts/meta.yaml
+++ b/recipes/perl-bio-db-hts/meta.yaml
@@ -1,10 +1,11 @@
 package:
   name: perl-bio-db-hts
-  version: '2.4'
+  version: '2.7'
 
 source:
-  fn: Bio-DB-HTS-2.4.tar.gz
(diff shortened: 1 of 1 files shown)


matthdsm:
@bioconda/perl 
Any idea what went wrong here? I'm not able to see the travis logs.

Thanks for the help
M

bgruening:
@matthdsm the error is this one `cc: error: unrecognized command line option ‘-fstack-protector-strong’`

matthdsm:
Hi @bgruening 
I've had this error before, but this according to StackOverflow, this could be caused by using an old gcc version. The current version in de default conda channel is already a bit outdated, and doesn't support this flag. How do you suppose this could be fixed?

Thanks
M

matthdsm:
@bioconda/core 
Anyone any idea?

Thanks
M

johanneskoester:
Soon, conda will receive a new and updated gcc package, that hopefully solves this. This will come along with conda-build 3, which also introduces much improved pinning mechanisms.

PR #4703, opened 2017-05-16
update cgat-scripts-nosetests to version 0.0.4

sebastian-luna-valero:
* [X] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [ ] This PR adds a new recipe.
* [X] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

PR #4739, opened 2017-05-19
Update quast to version 4.5

jvhaarst:

### DIFF
--- a/recipes/quast/meta.yaml
+++ b/recipes/quast/meta.yaml
@@ -17,6 +17,9 @@ requirements:
   - boost {{CONDA_BOOST}}*
   - perl
   - python
+  - libgcc
+  - matplotlib
+  - openjdk
   - joblib
   - simplejson
 
--- a/recipes/quast/meta.yaml
+++ b/recipes/quast/meta.yaml
@@ -1,15 +1,19 @@
++{% set name = "quast" %}
++{% set version = "4.5" %}
++{% set hash = "873ed8cd69e2d92fb87d75a37ecb75ea" %}
+
 package:
-  name: quast
-  version: "4.5"
+  name: {{ name|lower }}
+  version: {{ version }}
 
 build:
   number: 1
   string: boost{{CONDA_BOOST}}_{{PKG_BUILDNUM}}
 
 source:
-  fn: quast-4.4.tar.gz
-  url: http://downloads.sourceforge.net/project/quast/quast-4.5.tar.gz
-  md5: 873ed8cd69e2d92fb87d75a37ecb75ea
+  fn: {{ name|lower }}-{{ version }}.tar.gz
+  url: http://downloads.sourceforge.net/project/{{ name|lower }}/{{ name|lower }}-{{ version }}.tar.gz
+  md5: {{ hash }}
 
 requirements:
   build:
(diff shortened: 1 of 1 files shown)


jvhaarst:
The build on linux fails with : `OSError: [Errno 28] No space left on device`
That's not something that can be fixed in the recipe.

sebastian-luna-valero:
Please have a look at line 680 of your output:
https://travis-ci.org/bioconda/bioconda-recipes/jobs/233997330#L680

```
* Compiling GAGE *
ERROR! Java compiler not found (javac)! Please install it or compile GAGE java classes manually (/opt/conda/conda-bld/recipe_1495199699563/work/quast-4.5/quast_libs/gage/*.java)!
```

Although I agree `Errno 28` is not an intuitive error message to get.

jvhaarst:
Now it doesn't give the javac error, but fails before that because the docker image runs out of space:
https://travis-ci.org/bioconda/bioconda-recipes/jobs/234805582#L680

The only way to fix this is to not run the tests, I guess.

sebastian-luna-valero:
Thanks!

I am hitting a similar issue in https://github.com/bioconda/bioconda-recipes/pull/4703

I have pinged @bioconda/core to have a look, since I am not sure what could be causing the error:  `[Errno 28] No space left on device`

There must be something particular to our recipes, otherwise more people would be complaining about it.

By the way, I am not familiar with `quast` but it is building dependencies like `gnuplot`, `bwa`, `bedtools`, etc:
https://travis-ci.org/bioconda/bioconda-recipes/jobs/234805582#L640

I think they should be dependencies specified in the recipe, not built from source.

jvhaarst:
I'm not the maintainer of quast, but that's indeed a bit weird.
They probably want to be in control of the versions, so that is why they decided to do it this way.

johanneskoester:
Indeed, the optimal solution would be to modify the build such that external dependencies are used. Nevertheless, the space issues could maybe be solved by some modification in bioconda-utils, see #4703.

daler:
@jvhaarst, https://github.com/bioconda/bioconda-utils/pull/114 should have fixed the out-of-space issue (at least it solved this error for another PR).

jvhaarst:
It looks like everything went fine, so my work seems done.
Thank you all for the help with my first bioconda PR.

recipes/quast/meta.yaml
```diff
@@ -1,44 +1,51 @@
...
+  md5: {{ hash }}
 
 requirements:
   build:
   - gcc
   - boost {{CONDA_BOOST}}*
-  - perl-threaded
+  - perl
   - python
+  - libgcc
```
johanneskoester: libgcc does not need to be a build requirement. Only a runtime requirement.

recipes/quast/meta.yaml
```diff
@@ -1,44 +1,51 @@
+{% set name = "quast" %}
+{% set version = "4.5" %}
+{% set hash = "873ed8cd69e2d92fb87d75a37ecb75ea" %}
```
johanneskoester: Name and hash don't need to be variables. Only the version, because it occurs at multiple places and changes with a new release. However, this is not crucial so you can also keep it like this if you prefer this style.

jvhaarst: The hash also changes at a new release, so I guess it's better to keep it there.
The name is indeed not necessary, but it also doesn't conflict right now, so I'll leave it for now.

recipes/quast/meta.yaml
jvhaarst: Fixed.

recipes/quast/meta.yaml
johanneskoester: Yes, as I said, you can keep it like that. Indeed, the hash changes, but it occurs only at one place in the file so one could also simply modify it there (and have less clutter in the syntax).

PR #4754, opened 2017-05-21
Add netReg to Bioconda

dirmeier:

### DIFF
--- a/recipes/netreg/build.sh
+++ b/recipes/netreg/build.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# conda on macOS will fall back to libstdc++
+# which lacks a ton of standard C++11 headers
+if [[ $(uname) == Darwin ]]; then
+	export CXXFLAGS="${CXXFLAGS} -stdlib=libc++"
+fi
+
+mkdir -p build
+cd build
+
+cmake -DCMAKE_INSTALL_PREFIX="${PREFIX}" -DBOOST_ROOT="${PREFIX}" ..
+
+make VERBOSE=1
+make install
--- a/recipes/netreg/meta.yaml
+++ b/recipes/netreg/meta.yaml
@@ -1,0 +1,38 @@
+about:
+  home: https://github.com/dirmeier/netReg
+  license: GPL3
+  summary: "netReg fits linear regression models using network-penalization."
+
+build:
+  number: 0
+
+package:
+  name: netreg
+  version: '1.0.0'
+
+requirements:
+  build:
+    - gcc # [linux]
+    - llvm # [osx]
+    - libcxx # [osx]
+    - cmake >=3.8
+    - boost ==1.63.0
+    - armadillo >=7.800.1
+    - openblas >=0.2.19
+    - zlib
+  run:
+    - libgcc # [linux]
+    - libcxx # [osx]
+    - boost ==1.63.0
+    - armadillo >=7.800.1
+    - openblas >=0.2.19
+    - zlib
+
+source:
+  fn: netreg-1.0.0.tar.gz
+  sha256: 72d5cfcfd291ff6bd7f31782b3fab0077e93cd242fc102e6a22f68f7d364c349
+  url: https://github.com/dirmeier/netReg/archive/v1.0.0.tar.gz
+
+test:
+   commands:
+     - netReg -h


dirmeier:
Travis fails, because it pulls dependencies from different channels. This is due to the fact that defaults has higher priority that conda-forge. I can workaround this locally as I get set my priorities as I want. However, I don't know how I can fix that for travis. 

A solution is to add armadillo/boost/openblas to bioconda. While this seems much of a overhead, I think, in the long-run it is needed anyway.(?)

bgruening:
@dirmeier I'm not sure this is due to the channel order. Other boost packages do also compile. Maybe they broke it recently, but can it be that you for some reason the bzip2 is not picked up?

dirmeier:
Yeah, I am not too sure. It works on my machines where I have Mint/OSX. The only notable difference here is that I have my priorities `conda-forge` > `defaults`.  Maybe when `libbzip2` is not compiled form the same channel as `boost` this might result in linker errors. 

This post suggests similar problems: [conda-forge](https://github.com/conda-forge/boost-feedstock/pull/15#issuecomment-251734905)

I'll look into it.

SoapZA:
@bgruening ok, this is definitely a channel issue. The installed openblas has the following libs:
```
lrwxr-xr-x    1 travis  wheel        30 Jun  3 17:46 libopenblas.a -> libopenblas_nehalemp-r0.2.19.a
lrwxr-xr-x    1 travis  wheel        34 Jun  3 17:46 libopenblas.dylib -> libopenblas_nehalemp-r0.2.19.dylib
-rw-rw-r--    2 travis  wheel  23565376 Jan 25 03:48 libopenblas_nehalemp-r0.2.19.a
-rwxrwxr-x    2 travis  wheel  17331752 Jan 25 03:48 libopenblas_nehalemp-r0.2.19.dylib
```
The build errors out due to
```
dyld: Library not loaded: @rpath/libopenblasp-r0.2.19.dylib
  Referenced from: /anaconda/conda-bld/netreg_1496509853404/_t_env/lib/libarmadillo.7.80.1.dylib
  Reason: image not found
```
Notice that `libopenblasp-r0.2.19.dylib` is not installed, because this dependency comes from conda-forge and not defaults.

dirmeier:
Thanks for looking into this. So I guess we need to add openblas/armadillo here?

bgruening:
@SoapZA I have changed the channel order in the bulk branch, can you PR against the bulk branch and see if this solved your problem?

SoapZA:
@bgruening I've rebased the now 1 commit on top of bulk, but travis doesn't like pulling from bulk it seems. Am I missing something?

SoapZA:
@bgruening Trying to merge into `master` or `bulk` leads to Travis failures. In the former case it is apparently due to massive linting issues, and in the latter other case due to travis not being able to work against `bulk`.

bgruening:
Mh, don't try to mess with the buld branch, start frech from branch of from bulk and add you recipe. It's just that I have changed the channel order so this might help you here as well.
Otherwise, try help me and click the travis rerun button on the bulk branch (and maybe fix some packages) and then we are done in a few days with rebuilding all R packages :)

SoapZA:
@bgruening that's what I did. Anyhow, all of that failed, so I just rebased on `master` and cherry-picked your branch priority commit, and voila, it is indeed the channel priorities causing the issue!

We can proceed in a number of ways:
1. Add armadillo to bioconda. This will solve the issue in the short-term but problems like this can re-occur.
2. Permanently change the channel priorities, making **conda-forge** preferred over **defaults**. Given that 99% of packages in **defaults** are also in **conda-forge**, **conda-forge** is for all intents and purposes a superset of **defaults**. This will be more robust in the long-term and will likely also make support easier, as **conda-forge** packages are very responsive. Unfortunately, this will likely cause some fallout in the short-term, due to ABI issues stemming from linking to **defaults** packages.

The core idea here is that as you increase priority in the channel order list, you go down the dependency chain, that is, if A depends on B, then A has to come from a channel with equal or higher priority than B, as A is built against B and carries information about B. If B comes from a channel with a higher priority, then A comes from a channel that does not know about the channel of B and hence will likely fail. To be honest, this sounds like a deep architectural issue of conda in general and the link posted by Simon in https://github.com/bioconda/bioconda-recipes/pull/4754#issuecomment-303407775 states this too. @johanneskoester any ideas/comments?

bgruening:
We going to change the channel priority as soon as we merge the bulk branch. Still 470 packages to rebuild.

dirmeier:
Hey Björn @bgruening, any updates here?
Cheers, Simon

PR #4773, opened 2017-05-22
Fixed vcftools dependencies

vinjana:
…cf-840. Removed Perl code installation from vcftools. Removed perl as direct dependency in vcftools.

The package vcftools-0.1.14 builds some C code but by default also installs some Perl libraries. At the same time these same perl libraries are separately provided by the perl-vcftools-vcf-953 package and installed as dependency for vcftools.

Furthermore, also for vcftools-0.1.11 perl-vcftools-vcf-953 is listed as dependency in the meta.yaml, while the actual version packaged with vcftools (840) is installed with vcftools itself.

This PR does

- add perl-vcftools-vcf-840 and make vcftools-0.1.11 depend on it
- remove the Perl code installation from the vcftools-0.1.14 and vcftools-0.1.11 builds.


recipes/vcftools/build.sh
```diff
@@ -3,5 +3,6 @@ set -euo pipefail
 export CPPFLAGS="-I$PREFIX/include"
 export LDFLAGS="-L$PREFIX/lib"
 ./configure --prefix=$PREFIX
+sed -i -e 's/SUBDIRS = cpp perl/SUBDIRS = cpp/' ./src/Makefile
```
marcelm: For macOS, this needs to be `sed -i.bak`

recipes/vcftools/0.1.11/build.sh
```diff
@@ -1,4 +1,7 @@
 #!/bin/bash
 
+export CFLAGS="-I$PREFIX/include"
+export LDFLAGS="-L$PREFIX/lib"
+sed -i -e 's/DIRS = cpp perl/DIRS = cpp/' ./Makefile
```
marcelm: For macOS, this needs to be `sed -i.bak` (that is, the `-i` needs a backup extension)

recipes/perl-vcftools-vcf/840/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -eu
```
marcelm: Should probably be `set -euo pipefail` for consistency (even though there are no pipes, others might use this as a template for their own build script)

recipes/perl-vcftools-vcf/840/meta.yaml
```diff
@@ -0,0 +1,28 @@
+package:
+  name: perl-vcftools-vcf
+  version: '0.840'
+
+source:
+  fn: vcftools-0.1.11.tar.gz
+  url: http://downloads.sourceforge.net/project/vcftools/vcftools_0.1.11.tar.gz
+  sha256: 81b98c8bc93e366e2f0b14eb78e2736ae1c790176939eef88dbf0546ab952aab
+
+
```
marcelm: Remove one of the empty lines

marcelm:
I’m not sure why a review was requested from me - I have no experience with packaging anything that involves Perl. I pointed out what I saw, but cannot help more.

P.S: In your PR description, please replace `[x ]` with `[x]` to make it render correctly.

recipes/vcftools/build.sh
vinjana: o.k.

vinjana:
Sorry @marcelm, this was my first PR into bioconda and I am also not so proficient with github. I made all requested changes and will try getting an additional reviewer. Thanks!

marcelm:
No worries! These changes will not fix your Travis failures I believe.

vinjana:
True. Actually Travis continues to give me a lot of headache -- not only with this build.

marcelm:
The error is here: https://travis-ci.org/bioconda/bioconda-recipes/jobs/234866431#L755
```
sed -i.bak -e 's/CFLAGS = /CFLAGS = -I/opt/conda/conda-bld/recipe_1495467737945/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh\/include /' ./cpp/Makefile

sed: -e expression #1, char 25: unknown option to `s'
```

recipes/vcftools/0.1.11/build.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash
-
+set -euo pipefail
+sed -i.bak -e 's/DIRS = cpp perl/DIRS = cpp/' ./Makefile
+sed -i.bak -e "s/CFLAGS = /CFLAGS = -I$PREFIX\/include /" ./cpp/Makefile
```
marcelm: You need to use a different delimiter than `/` because the `$PREFIX` contains slashes as well. `'s|CFLAGS = |CFLAGS = ...|`, for example.

recipes/vcftools/0.1.11/build.sh
```diff
@@ -1,4 +1,8 @@
 #!/bin/bash
-
+set -euo pipefail
+sed -i.bak -e 's/DIRS = cpp perl/DIRS = cpp/' ./Makefile
```
marcelm: Prefixing a path with `./` is not necessary. Just write `Makefile` only. (Same for other file names below.)

vinjana: I followed the convention of the file from bioconda that I used as template.

vinjana:
O.k. now I am at a point where I was before already with another PR candidate: One package depends on another, but cannot find it during the build it travis. I have no clue how to fix this :-(

recipes/vcftools/meta.yaml
```diff
@@ -8,20 +8,18 @@ source:
...
 build:
-  number: 4
+  number: 5
 
 requirements:
   build:
     - gcc  # [not osx]
     - llvm  # [osx]
     - zlib
-    - perl-threaded
```
bgruening: would be better to explicitly state `perl` as dependency.

recipes/vcftools/0.1.11/meta.yaml
```diff
@@ -8,18 +8,16 @@ source:
   sha256: 81b98c8bc93e366e2f0b14eb78e2736ae1c790176939eef88dbf0546ab952aab
 
 build:
-  number: 0
+  number: 1
 
 requirements:
   build:
-    - perl-threaded
```
bgruening: `perl`

recipes/perl-vcftools-vcf/0.840/run_test.pl
```diff
@@ -0,0 +1 @@
+use Vcf;
```
bgruening: can we drop this file and simply use imports like here https://github.com/bioconda/bioconda-recipes/blob/master/recipes/perl-gd/meta.yaml#L31

vinjana: I tried, but with "imports" simply "run_test.py" is called which apparently then tries to "import Vcf" in Python. 

Is there an explicit way to tell the interpreter to try the a default "run_test.pl"? Is that documented somewhere?

jerowe: I noticed this has been happening lately too.

I changed my tests to -

test:
  commands:
    - perl -e "use SomeLib"

But its not awesome because then it doesn't check for versioning.

bgruening: Oh, is this broken in general? If so can you open an issue at conda-build with a minimal example?

jerowe: I think so, but let me confirm. If its a problem I will raise an issue.

vinjana: Consider conda/conda-build#1967. So the issue probably is fixed in conda-build 3.0.0.
How to proceed now with the pull request? Use the following?

```yaml
test:
  commands:
      - perl -e "use Vcf;"
```

vinjana:
Consider conda/conda-build#1967. The problem that `test: import: - Vcf` doesn't work seems to be related to conda-build and is probably fixed in conda-build 3.0.0.

My question is, How to proceed now with this pull request? I updated the test to use the following, like jerowe did with hers:

```yaml
test:
  commands:
      - perl -e "use Vcf;"
```

Is this o.k. and can be merged?

.idea/libraries/R_User_Library.xml
```diff
@@ -0,0 +1,8 @@
+<component name="libraryTable">
+  <library name="R User Library">
```
bgruening: Can you remove this file please.

Issue #4772, opened 2017-05-22
Advice for recipe downloading multiple files?


peterjc:
I've skimmed https://bioconda.github.io/faqs.html and other pages, but have yet to see an example ``meta.yaml`` which defines multiple independent entries under ``source:``. Searching the git repository with grep I could finds lots of examples with multiple URLs, but they all seem to using conditionals like ``# [linux64]``, ``# [linux32]``, ``# [osx]`` etc.

I would like to create an ``effectiveT3`` BioConda recipe to replace the ``tool_dependencies.xml`` used on my Galaxy wrapper:

https://github.com/peterjc/pico_galaxy/blob/master/tools/effectiveT3/tool_dependencies.xml
http://toolshed.g2.bx.psu.edu/view/peterjc/effectivet3

In essence this downloads FIVE separate JAR files, and puts them into a folder structure:

* ``$EFFECTIVET3/TTSS_GUI-1.0.1.jar``
* ``$EFFECTIVET3/module/TTSS_ANIMAL-1.0.1.jar``
* ``$EFFECTIVET3/module/TTSS_PLANT-1.0.1.jar``
* ``$EFFECTIVET3/module/TTSS_STD-1.0.1.jar``
* ``$EFFECTIVET3/module/TTSS_STD-2.0.2.jar``

This is a JAVA package, so I have read https://bioconda.github.io/guidelines.html#java which explains where to put the JAR files, recommends adding a shell script, and suggests https://github.com/bioconda/bioconda-recipes/tree/master/recipes/peptide-shaker as a model.

I could put the primary file ``TTSS_GUI-1.0.1.jar`` into ``meta.yaml``, but can I put in multiple entries under ``source:`` each with their own checksum?

Or, should I could put the primary file ``TTSS_GUI-1.0.1.jar`` into ``meta.yaml``, and have the ``build.sh`` itself download the four child JAR files itself (and verify their checksums)?

peterjc:
Is there any easy way to get ``shasum`` installed as a build dependency? e.g. https://travis-ci.org/bioconda/bioconda-recipes/jobs/234836961

```
/opt/conda/conda-bld/recipe_1495462269594/work/conda_build.sh: line 17: shasum: command not found
```

daler:
@peterjc WRT your original question, I think your latter solution is the best (choose a single source; download the rest in `build.sh`). WRT to `shasum`, try `sha256sum` instead, should be installed in `condaforge/linux-anvil`.

[...]

PR #4774, opened 2017-05-22
clinod recipe, using batchman from SNNS

peterjc:

This would close ##4675 by packaging the command line NoD predictor (clinod).

I have tried to follows BioConda JAVA best practise https://bioconda.github.io/guidelines.html#java by adding the wrapper script ``clinod.py`` (put under bin as ``clinod`` with no extension) following the peptide-shaker example by @bgruening.

Note my ``build.sh`` downloads, verifies and uses batchman from SNNS pre-compiled for 64bit Linux. If and when that is packaged for (Bio)Conda we could use that.

Cross reference #4772 for how best to deal with additional downloads.


peterjc:
Hmm, this would need to be marked as 64-bit Linux only (pending a proper SNNS package).

peterjc:
How do we mark a recipe as Linux only?

This now works but it uses the pre-compiled 64-bit Linux binary for SNNS provided by the authors, which means the entire recipe only works on 64-bit Linux.

Right now packaging SNNS is not going smoothly (#4776), so it this acceptable as a stop-gap?

daler:
@peterjc, you can use the preprocessing selectors with a `build:skip` key:

```yaml
build:
  skip: True  # [not linux]
```

some examples of other recipes doing this are gap2seq, repeatmasker, exonerate, and mothur.

As long as the binary runs successfully in the mulled-build test, I think it should be fine for now.

peterjc:
@daler are you happy for this to be merged now with the Linux only restriction added?

daler:
Looks good! Everything's nicely documented in meta.yaml and build.sh so it's clear what's happening.

daler:
@bgruening @johanneskoester second opinion?

johanneskoester:
If you ask me, I think it would be cleaner to fix the SNNS pull request first. A fix could as well be to use the precompiled binary in there. Then, this recipe can simply depend on that.

peterjc:
I'm unclear where BioConda sits on the pragmatic vs ideal spectrum for build recipes.

@johanneskoester I agree packaging SNNS would be ideal, so if you can help fix #4776 that would be great. Right now I'm out of ideas for that, and the ``clinod`` package as-is seems to work despite the dependency hack.

johanneskoester:
Well can't you just use the precompiled binary in the snns recipe?

peterjc:
@johanneskoester that's a nicer hack - OK - I'll do that and then update the ``clinod`` package.

johanneskoester:
Thanks! And sorry for the additional work!

peterjc:
This recipe itself now ought to work on macOS, if and when the ``snns`` package is available (which is currently Linux only). Do I need this here:

```
build:
  skip: True  # [not linux]
```

peterjc:
Right now the second test is failing,

```
[Jun  2 09:57:17] INFO Invoke Task [test]
[Jun  2 09:57:17] STEP Run image [quay.io/biocontainers/clinod:1.3--py27_0] with command [[clinod | grep -i "^NucleOlar localization sequence Detector" && clinod -in=four_human_proteins.fasta -f=MEDIUM_TAB -nonols -clean_sequence]]
[Jun  2 09:57:17] DEBU Creating container [step-61aebd615b]
[Jun  2 09:57:17] DEBU Created container [0334ceddc40b step-61aebd615b], starting it
[Jun  2 09:57:18] DEBU Container [0334ceddc40b step-61aebd615b] started, waiting for completion
[Jun  2 09:57:18] SOUT NucleOlar localization sequence Detector v. 1.3b (13 May 2011)
[Jun  2 09:57:18] SERR Input file does not exist or is empty! Given data: four_human_proteins.fasta
[Jun  2 09:57:18] ERRO Task processing failed: Unexpected exit code [1] of container [0334ceddc40b step-61aebd615b], container preserved
involucro -v=2 -f /anaconda/lib/python3.5/site-packages/galaxy_lib-17.5.9-py3.5.egg/galaxy/tools/deps/mulled/invfile.lua -set CHANNELS='bioconda,file://anaconda/conda-bld,r,defaults,conda-forge' -set TEST='clinod | grep -i "^NucleOlar localization sequence Detector" && clinod -in=four_human_proteins.fasta -f=MEDIUM_TAB -nonols -clean_sequence' -set TARGETS='clinod=1.3=py27_0' -set REPO='quay.io/biocontainers/clinod:1.3--py27_0' -set BINDS='build/dist:/usr/local/,/anaconda/conda-bld:/anaconda/conda-bld' build-and-test
```

Running the equivalent command locally this is consistent with the file not being found:

```
$ ./clinod -in=four_human_proteins.fasta -f=MEDIUM_TAB -nonols -clean_sequence
#ID	Start	End	NOLS
sp|Q9NSY1|BMP2K_HUMAN	965	998	SQQQKVKQRSLQKLSSRQRRTKQDMSKSNGKRHH
sp|Q9NSY1|BMP2K_HUMAN	1000	1035	TPTSTKKTLKPTYRTPERARRHKKVGRRDSQSSNEF
sp|P06213|INSR_HUMAN	286	307	CQDLHHKCKNSRRQGCHQYVIH
```

vs:

```
$ ./clinod -in=does_not_exist.fasta -f=MEDIUM_TAB -nonols -clean_sequence
Input file does not exist or is empty! Given data: does_not_exist.fasta
```

Am I missing something simple in my test definition, which current is as follows (ignoring the comments):

```yaml
test:
  files:
    - four_human_proteins.fasta
  commands:
    - clinod | grep -i "^NucleOlar localization sequence Detector"
    - clinod -in=four_human_proteins.fasta -f=MEDIUM_TAB -nonols -clean_sequence
```

johanneskoester:
We currently don't support test files (the second, more strict test happens in a busybox container). It is sufficient to simply test if the binaries can be properly invoked, e.g. with --help or --version.

peterjc:
@johanneskoester If BioConda does not support test files, could the BioConda linting catch this? There are other examples in the existing recipes which I copied - including when writing https://github.com/bioconda/bioconda-recipes/blob/master/recipes/effectivet3/meta.yaml where it seemed to work.

In this case just testing the binary starts is not really enough - that won't catch a runtime failure if it can't find ``batchman``.

In an earlier version of the pull request I used echo to pipe in a test FASTA sequence, let me try that again...

johanneskoester:
You can check for the existence of batchman or even run it in a separate test.
Those recipes you name have probably been build before we had the additional layer of tests.

@bgruening would it make sense to allow test files in mulled, and remove them from the container again after they have been used for the test?

bgruening:
@peterjc please note that you also can run some tests in your build.sh file and all things you can echo will also work.

@johanneskoester I think this will only work if the test files are part of the package, at the time mulled is testing the repo the test files (locate in the recipe) are somewhere else. Not impossible, but not easy I guess.

peterjc:
This seems to work with an echo trick in the ``meta.yaml`` test.

Is this OK to merge, or should I move that test into ``build.sh`` instead?

recipes/clinod/build.sh
```diff
@@ -0,0 +1,14 @@
...
+mkdir -p $outdir
+mkdir -p $PREFIX/bin
+cp -R * $outdir/
+cp $RECIPE_DIR/clinod.py $outdir/clinod
+ls -l $outdir
+ln -s $outdir/clinod $PREFIX/bin
+chmod 0755 "${PREFIX}/bin/clinod"
+
+# clinod expects the batchman binary from SNNS in its folder:
+ln -s "`which batchman`" $outdir/batchman
```
bgruening: I'm actually surprised that this works, it is an absolute symlink which should be on all systems different isn't it?
I would expect a relative to work.

peterjc: Other than the test file thing, this is the bit of the recipe I am least confident in.

How would you suggest I work out where Conda would install ``batchman`` from the ``snns`` package (which I have deliberately listed as both a build and runtime dependency)?

I'm probably making a number of wrong assumptions in how the Conda ecosystem works - while this passes the tests it may not work "in the wild". I guess we'd find out once the package goes live (although I'm not immediately ready to test this with my Galaxy wrapper for Effective T3).

bgruening: You can look at the package. https://anaconda.org/bioconda/snns/files
It is stored in $PREFIX/bin. I guess this is all fine as Conda will fix all these hardcoded paths for us.

peterjc: So you're suggesting this:

```bash
ln -s $PREFIX/bin/batchman $outdir/batchman
```

Perhaps supplemented with a test like to verify the new symlink works:

```bash
$outdir/batchman -h 2>&1 | grep SNNS
```

bgruening: Yes to the first. The second will not work as you don't have $outdir defined in the runtime stage.
Eseentially, you have 3 stages. 
* Build stage, all in build.sh
* test/run-time stage your tests in meta.yaml - conda is still here
* And the BioConda specific one: The mulled stage: No conda no nothing just your package in a busybox.

All 3 are isolated.

bgruening:
Thanks @peterjc!

peterjc:
Thanks all for the feedback and guidance

Issue #4675, opened 2017-05-11
Command line NoD predictor (clinod)


peterjc:
Command line NoD predictor is a tool for predicting nucleolar localization sequences (NoLSs) in a FASTA file of proteins using a neural network. There is also a webtool version at http://www.compbio.dundee.ac.uk/www-nod/

Would a recipe based on https://github.com/peterjc/pico_galaxy/blob/master/tools/clinod/tool_dependencies.xml be welcomed? Note clinod requires the SNNS Batch Interpreter v1.0 executable, which I would propose to download precompiled.

bgruening:
Sure! :)

bgruening:
ping @peterjc

peterjc:
Done, belatedly closing this issue - thanks for the reminder:

https://anaconda.org/bioconda/clinod
https://github.com/bioconda/bioconda-recipes/tree/master/recipes/clinod

PR #4776, opened 2017-05-22
Stuttgart Neural Network Simulator (SNNS)

peterjc:

The licensing for SNNS is unclear (*update - that's been fixed*), but I would like to package it for clinod - see #4675 / #4774

### DIFF
--- a/recipes/snns/build.sh
+++ b/recipes/snns/build.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+set -eu -o pipefail
+
+# In principle we should be able to build this from source,
+# possiblly deliberatly using our verion of flex to first
+# regenerate the parsers using something like this:
+#
+# $ touch kernel/sources/*.l tools/sources/*.l
+# $ ./configure --enable-global --prefix $PREFIX
+# $ make install
+#
+# I failed to get this to work, so have compromised with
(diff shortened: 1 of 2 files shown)


peterjc:
Oh, good news http://www.ra.cs.uni-tuebingen.de/SNNS/licence.html says "The Licensing Terms for SNNS have significantly changed from Verion 4.2 to Version 4.3. SNNS is now available under the LGPL Version 2." even though the tar-bar seems not to mention this 👍

peterjc:
The provided configure script does not like macOS,

```
checking build system type... configure: error: cannot guess build type; you must specify one
```

How can I flag the recipe as Linux only?

peterjc:
Finally, this recipe is working.

As discussed on #4774 in line with @johanneskoester's suggestion this uses the provided 64-bit Linux binaries (since I was unable to get their code to compile, something wrong with finding the flex libraries).

They do not provide macOS binaries, and their configuration scripts do not support it.

Note I have not included a test for the provided GUI tool.

peterjc:
I'm not 100% sure from reading their documentation if the flex and/or bison tools are needed at runtime. There are references to using these tools if rebuilding a model. I've never used their tools directly myself though.

bgruening:
Thanks for checking!

peterjc:
Cheers.

PR #4785, opened 2017-05-23
Update gdk-pixbuf/meta.yaml

ThomasWollmann:

### DIFF
--- a/recipes/gdk-pixbuf/build.sh
+++ b/recipes/gdk-pixbuf/build.sh
@@ -4,14 +4,7 @@ ln -s $PREFIX/lib $PREFIX/lib64
 
 export PKG_CONFIG_PATH="$PREFIX/lib/pkgconfig:$PREFIX/share/pkgconfig:/usr/lib64/pkgconfig:/usr/share/pkgconfig"
 export ACLOCAL_FLAGS="-I$PREFIX/share/aclocal"
-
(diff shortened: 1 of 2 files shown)


ThomasWollmann:
Why does it try to compile it with Python 3 on OSX?

bgruening:
Not sure, can you try to skip py3k?

ThomasWollmann:
Now stuck with the good old: GLIB 2.37.6 or better is required.

ThomasWollmann:
I switched to a more recent GLIB version. However in OSX it can't find it. @bgruening is there an additional env variable I have to set for OSX like LD_LIBRARY_PATH during runtime?

johanneskoester:
Maybe there is a dedicated argument in the configure script?
Further, the common style is to add a `# ` in front of the selectors.

A general question: is there a specific reason for the update? I am asking because in the long run this recipe should be moved over to conda-forge, because it is not bio-related.

ThomasWollmann:
@johanneskoester  I'm not aware of a special argument.
I have a coworkers who use mac and analyse whole slide images with openslide. gdk-pixbuf is a dependency of openslide.

johanneskoester:
Ok. If we manage to fix this I'm fine with keeping it for now until it is finally moved over to conda-forge (which will still allow openslide to depend on it of course). @bioconda/osx maybe you can help with the osx problem?

acaprez:
I was able to build it ok on OS X by tweaking a couple of things:
1)  Add `${PREFIX}/lib` to the rpath at link-time by changing `LDFLAGS` to:
`LDFLAGS="-L$PREFIX/lib -Wl,-rpath ${PREFIX}/lib"`
2)  Fix the version of `gettext` used to 0.19.7 for both the build and run requirements, i.e.
`gettext ==0.19.7 # [osx]`
This is needed because `glib` was apparently built against 0.19.7, and they incremented the version in the SONAME for `libintl` from 8 to 9 between 0.19.7 and 0.19.8.

PR #4793, opened 2017-05-23
Update: ensembl-vep, perl-bio-db-hts with htslib

chapmanb:
Upgrades to htslib 1.4 and current perl in conda-forge. Avoids using
package gcc for compilation because of fstack-protector-strong error,
instead using gcc in condaforge/linux-anvil container.

Fixes #4703
Fixes chapmanb/bcbio-nextgen#1911

### DIFF
--- a/recipes/ensembl-vep/meta.yaml
+++ b/recipes/ensembl-vep/meta.yaml
@@ -8,21 +8,21 @@ source:
   md5: 2c088ead3a3a5245a7209e9b087666af
 
 build:
-  number: 1
+  number: 2
   skip: True # [osx]
 
 requirements:
   build:
     - curl
+    - perl
     - perl-archive-zip
     - perl-lwp-simple
-    - perl-threaded >=5.2*
     - unzip
-    - zlib
   run:
-    - htslib 1.3.*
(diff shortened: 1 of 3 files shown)


recipes/ensembl-vep/meta.yaml
```diff
@@ -8,21 +8,21 @@ source:
...
+    - perl
     - perl-archive-zip
     - perl-lwp-simple
-    - perl-threaded >=5.2*
     - unzip
-    - zlib
   run:
-    - htslib 1.3.*
+    - perl
+    - htslib {{CONDA_HTSLIB}}*
```
johanneskoester: @chapmanb you need to also modify the build string such that it contains the pinned htslib version (see other recipes). Otherwise, we won't be able to detect that a rebuild is necessary when the env matrix is updated.

chapmanb: Thanks for the heads up, and sorry for missing this. I have a new PR fixing the recipe build strings.

johanneskoester: No problem, thanks for the quick fix!

PR #4790, opened 2017-05-23
Package Effective T3 v1.0.1 for BioConda

peterjc:
* [X] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [X] This PR adds a new recipe.
* [ ] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

Based on the install recipe https://github.com/peterjc/pico_galaxy/blob/master/tools/effectiveT3/tool_dependencies.xml used in my Galaxy wrapper http://toolshed.g2.bx.psu.edu/view/peterjc/effectivet3 which was not setup as a separate package.

As per #4772 the main JAR file is in the ``meta.yaml`` file, and the child files are fetched and verified from ``build.sh``.

Note: It appears to be BioConda convention to use all lower case names (I was using ``effectiveT3`` in Galaxy).

PR #4799, opened 2017-05-24
Update for Genepop 4.6, works on Mac

peterjc:

Original recipe by @tiagoantao 

It looks like @bgruening restricted this to exclude Mac, not sure why - it compiles for me locally and my tests passed.

I have also made the compilation requirements explicit in line with evolving BioConda best practises, and added a minimal test.

Note the authors seem to reuse their filenames and URL, meaning older releases are not cached. I have added a checksum, but this is a concern as sources for the old releases are no longer available.

### DIFF
--- a/recipes/genepop/build.sh
+++ b/recipes/genepop/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-
+cd Cpp
 g++ -DNO_MODULES -o Genepop GenepopS.cpp -O3
 
 mkdir -p $PREFIX/bin
--- a/recipes/genepop/meta.yaml
+++ b/recipes/genepop/meta.yaml
@@ -27,4 +27,4 @@ about:
 test:
   commands:
     # Expected output verion and prompt for a file or press ENTER to quit:
-    - echo -e '\n' | Genepop
+    - bash -c "echo -e '\n' | Genepop"
--- a/recipes/genepop/meta.yaml
+++ b/recipes/genepop/meta.yaml
@@ -24,7 +24,8 @@ about:
   license: CeCILL
   summary: The Genepop population genetics package.
 
-test:
-  commands:
-    # Expected output verion and prompt for a file or press ENTER to quit:
-    - bash -c "echo -e '\n' | Genepop"
+#Potential test commented out as could not get it to work in framework...
+#test:
+#  commands:
+#    # Expected output verion and prompt for a file or press ENTER to quit:
+#    - bash -c "echo -e '\n' | Genepop"
--- a/recipes/genepop/meta.yaml
+++ b/recipes/genepop/meta.yaml
@@ -1,15 +1,24 @@
 package:
   name: genepop
-  version: "4.5.1"
+  version: "4.6"
+
 build:
   number: 0
-  skip: True # [osx]
+
 source:
+  # Sadly they reuse the same URL and filename for each release
+  # so the checksum is our fall back for coping with this.
   fn: sources.tar.gz
   url: http://kimura.univ-montp2.fr/~rousset/sources.tar.gz
+  sha256: 9a87c9f5401d9de5a9a675e8c4db64ae6f63e83e7924d678c7da12a2210c9475
+
 requirements:
   build:
+    - gcc # [linux]
+    - llvm # [osx]
   run:
+    - libgcc # [linux]
+
 about:
   home: http://kimura.univ-montp2.fr/~rousset/Genepop.htm
   license: CeCILL
--- a/recipes/genepop/meta.yaml
+++ b/recipes/genepop/meta.yaml
@@ -24,8 +24,7 @@ about:
   license: CeCILL
   summary: The Genepop population genetics package.
 
-#Potential test commented out as could not get it to work in framework...
-#test:
-#  commands:
-#    # Expected output verion and prompt for a file or press ENTER to quit:
-#    - bash -c "echo -e '\n' | Genepop"
+test:
+  commands:
+    # Expected output is the version and an error message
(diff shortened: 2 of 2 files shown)


bgruening:
@peterjc at this time we wasn't building for OSX by deafult, we had a whitelist of package to build or was it a black-list? Either way I removed it and included the skip: True in all packages that was on this list. The idea was and still is to get them onto OSX if people care, as you do :) So plan worked out :)

peterjc:
@bgruening that makes sense, safer to assume the recipes at that point would be Linux only by default.

peterjc:
It looks like some magic may be needed for the test command,

```
subprocess.CalledProcessError: Command '['mulled-build', 'build-and-test', 'genepop=4.6--0', '-n', 'biocontainers', '--test', "echo -e '\\n' | Genepop", '--extra-channels', 'file://anaconda/conda-bld,r,defaults,conda-forge']' returned non-zero exit status 1
```

In place of:

```
echo -e '\n' | Genepop
```

I am trying:

```
bash -c "echo -e '\n' | Genepop"
```

peterjc:
@bgruening @tiagoantao is this OK now?

I tried but failed to get a new test to pass - e.g. https://travis-ci.org/bioconda/bioconda-recipes/jobs/235594244

tiagoantao:
As long as it works on Linux (the only thing that I was supporting anyways) its fine by me.

peterjc:
Thanks @tiagoantao I've tested GenePop 4.6 compiled by hand this way on the Mac and Linux (via your Biopython wrapper tests), so the basics are OK.

Can a BioConda guru (@bgruening, @daler?) explain why the test fails and/or approve this as is without the attempted test?

bgruening:
@peterjc can you try this: `echo 'random_test.txt' | Genepop | grep 'Genepop version'`

peterjc:
That's clever - the exact output on screen seems to depend on the terminal state, sometimes on the Mac I get:

```
Genepop version 4.6
From set_eof_check_EOLtype(): Cannot open file!
Check input file name random_test.txt
From parseFile(): Cannot open random_test.txt although it could previously be opened.
I exit.
```

Sometimes:

```
Genepop version 4.6
From parseFile(): Cannot open random_test.txt although it could previously be opened.
I exit.
```

I will try this in the recipe now...

peterjc:
Mac passed :)

Linux failed - in a way which seems very similar to some of my other open pull requests where unless the logging it misleading it seems the test command framework itself is broken?

```
...
11:24:56 BIOCONDA ERROR COMMAND FAILED: mulled-build build-and-test genepop=4.6--0 -n biocontainers --test echo 'does_not_exist.txt' | Genepop | grep 'Genepop version' --extra-channels file://anaconda/conda-bld,r,defaults,conda-forge
...
[May 29 11:24:56] ERRO Task processing failed: Invalid bind specification [], has to be of the form: source:dest
involucro -v=2 -f /anaconda/lib/python3.5/site-packages/galaxy_lib-17.5.9-py3.5.egg/galaxy/tools/deps/mulled/invfile.lua -set CHANNELS='bioconda,file://anaconda/conda-bld,r,defaults,conda-forge' -set TEST='echo 'does_not_exist.txt' | Genepop | grep 'Genepop version'' -set TARGETS='genepop=4.6=0' -set REPO='quay.io/biocontainers/genepop:4.6--0' -set BINDS='build/dist:/usr/local/,/anaconda/conda-bld:/anaconda/conda-bld' build-and-test
11:24:56 BIOCONDA ERROR TEST FAILED: recipes/genepop, ...
```

Note the first and second lines quoted above do not show the test command in any way - if run exactly as written it would clearly fail:

```
$ mulled-build build-and-test genepop=4.6--0 -n biocontainers --test $TESTCMD --extra-channels file://anaconda/conda-bld,r,defaults,conda-forge
```

Cross reference https://github.com/bioconda/bioconda-recipes/pull/4790#issuecomment-303776317

bgruening:
Thanks Peter!

peterjc:
Thanks @bgruening - Could you explain why that particular combination of quotes worked under Linux (and Mac) as I don't understand and it would likely help me with other recipes?

bgruening:
Sure! It's a little bit more complicated. On Linux we run very advanced tests to ensure that packages run in most-if-not-all environments. To do so, we use a tool called mulled/involucro that converts a conda package into a Docker container (https://github.com/galaxyproject/galaxy-lib/tree/master/galaxy/tools/deps/mulled).

Taking as Docker container busybox, we can really make sure that the packages works everywhere, as we don't have conda, or any library installed. The code is constructed here: https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/pkg_test.py#L128

The disadvantages is that we can not use any advanced testing feature from conda, like the test-depends etc ... as we don't have conda in this test environment. An other disadvantage is that we need to pass all tests via a commandline call to mulled/involucor ... and here we need to quote tests as well ...

Does this make sense?

peterjc:
Thanks @bgruening - those details are useful, especially https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/pkg_test.py#L128 which shows exactly where that command is built up.

I still don't understand which style of quoting is necessary for this to work though - should I in general avoid single quotes as part of the sample test commands in favour of double quotes?

bgruening:
I think what happens here: https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/pkg_test.py#L133 and later in sp.run() is that the string is single-quoted so you need to use double-quotes in your tests.

peterjc:
It does seem that using single-quotes within a test command is problematic for the test framework (under Linux at least). That is helpful to know - should we lint for this?

PR #4811, opened 2017-05-24
MultiQC for py3

ewels:

Trying to get MultiQC to build for python3 with click. Can't get test builds to work locally or on travis with branch, so trying in a PR instead..


ewels:
Hmm, not sure why the first three commits there are showing up. They've already been merged.. :\

ewels:
ok, so as expected, getting [an error](https://travis-ci.org/bioconda/bioconda-recipes/jobs/235927260#L677) for py3 on linux due to `click` looking for the `locale` command.

ewels:
Brilliant stuff - you fixed it @bgruening! What was the secret? `libgcc`?

bgruening:
@ewels I didn't fixed it, sorry :(
I just figured out that the command `multiqc --version` is triggering the click error.
After this I saw that `libgcc` was missing, this is due to the fact that conda-forge is not putting this into their run-time dependencies, which they should.

bgruening:
@ewels just for your information: http://click.pocoo.org/5/python3

```
At the moment, it is strongly recommended is to use Python 2 for Click utilities unless Python 3 is a hard requirement.
```
Maybe it is really better to skip python3 for click packages until 3.7 is our with a proper fix.

ewels:
Ah ok, that's a shame. Thanks for the link, I've read it before - I still struggle to agree with it though, as I and many others have been running MultiQC + click on Python3 for over a year now with no problems, including from bioconda with the previous build system. This packaging issue is the first problem I've run into.

bgruening:
@ewels we could patch click, so that it only issues a warning and fails gracefully if local is not present.

ewels:
Yeah, I was just thinking about this.. So now that `skip:false` is there, it is building for Python3 right? So what is it that will be broken? Just the Python 3 docker image on quay.io?

ewels:
(I just tried installing from bioconda into a test py3 environment locally and it seems to work fine)

bgruening:
The python3 docker images is very likely broken and all systems that do not have a utf-8 locale, as far as I understood it.

ewels:
Ok - well at least it's now available for some people I guess, probably better than not being available for any py3 users. Thanks for your work on this!

Phil

PR #4814, opened 2017-05-25
Fix runtime requirements for khmer

luizirber:
sourmash won't work with khmer < 2.1


druvus:
Thanks @luizirber. I updated khmer but forgot to add the requirement here.

druvus:
Only 2.1rc1 was included in bioconda so I am updating it in #4820

druvus:
thanks @luizirber it should be fixed now

PR #4815, opened 2017-05-25
Last

npavlovikj:
* [] This PR updates an existing recipe.


npavlovikj:
@bioconda/core, for the older "last" recipes, "--mulled-test" fails for Python 3.6. When I don't use "--mulled-test", the builds are successful. Also, the new "last" recipe uses only Python 2.7 and it builds fine, but because the older ones don't have this restriction and fail for Python 3.6, the whole build fails. I would appreciate if you can look into this, and help me fix this issue. Another option is to just add "skip: True # [py3k]" in the old recipes for now.

recipes/last/meta.yaml
```diff
@@ -1,34 +1,33 @@
...
-  patches:
-    - maf-convert.23patch
+  fn: last-852.zip
+  url: http://last.cbrc.jp/last-852.zip
+  md5: 96b8fdadab2b0ed77f8e55ce4c91687f
 
 build:
-  number: 0
-  skip: False
+  number: 1
```
bgruening: if it is a new version you can start with 0

recipes/last/meta.yaml
```diff
@@ -1,34 +1,33 @@
...
+  skip: True # [py3k]
```
bgruening: Is there any reason to skip py3k? Is it not supported, if so are the older version also not supported for py3k?

PR #4823, opened 2017-05-25
Handle case of a new branch with a single commit on forks.

### LINKED ISSUE #4553: Travis always failing on first commit in branch for forks
Hi,

I think I've found a bug related to testing travis builds on forks. Basically, travis will hit a bug during `travis-run.sh` at the `bioconda-utils lint recipies` call on the first  commit since the `--git-range` argument won't get formatted properly.

## example

Here's an commented example (starting from the base of `bioconda-recipie` repo): 

```bash
$ git checkout -b test_travis
$ echo "change 1" > change.txt
$ git add change.txt
$ git commit -m "change 1"
$ git push --set-upstream origin test_travis # my fork is set as origin
```

The travis build for this commit will fail due to the git range argument being parsed incorrectly ([example](https://travis-ci.org/ivirshup/bioconda-recipes/jobs/226702865)). The error will be:

```
+bioconda-utils lint recipes config.yml --git-range
usage: bioconda-utils lint [-h] [--packages PACKAGES [PACKAGES ...]]
                           [--cache CACHE] [--list-funcs]
                           [--only ONLY [ONLY ...]]
                           [--exclude EXCLUDE [EXCLUDE ...]] [--push-status]
                           [--user USER] [--commit COMMIT] [--push-comment]
                           [--pull-request PULL_REQUEST] [--repo REPO]
                           [--git-range GIT_RANGE [GIT_RANGE ...]]
                           [--full-report] [--loglevel LOGLEVEL]
                           recipe_folder config
bioconda-utils lint: error: argument --git-range: expected at least one argument
/home/travis/.travis/job_stages: line 153:  4130 Terminated              travis_jigger $! $timeout $cmd
```

Now running:

```bash
$ echo "change 2" >> change.txt
$ git add change.txt
$ git commit -m "change 2"
$ git push
```

Travis will succeed in building this ([example](https://travis-ci.org/ivirshup/bioconda-recipes/jobs/226703027)).

## what's up?

I think what's happening is that at the first commit `$TRAVIS_COMMIT_RANGE` is null. Since `$TRAVIS_PULL_REQUEST` is `"false"`, `$RANGE` will be set to `${TRAVIS_COMMIT_RANGE/.../ }`, which in this case is `""`.

## possible resolution

I've managed to get an initial commit to actually build by removing lines 30-33 ([relevant build](https://travis-ci.org/ivirshup/bioconda-recipes/jobs/226720003), though there is a bunch of logging as well). As I have no idea if this will break any other processes – or any idea what those processes are – I'm waiting on opening a pull request until someone more knowledgable can provide input.

## things I think are related

* #4038
* 59432bbc72cc957cc48cb4f7e2ccb2938d6a070a

acaprez:

Fixes #4553.  As @ivirshup nicely explained there, when pushing a new branch with a single commit to your fork, the build will always fail as the `TRAVIS_COMMIT_RANGE` variable is empty for that case.  This change checks for that, and sets the range to pick up the last commit.  When multiple commits are made locally to a new branch and then pushed, `TRAVIS_COMMIT_RANGE` has the proper value, so we should only ever need to look at the last commit when it's empty.

Ping @bioconda/core since this is a change to a build script.

### DIFF
--- a/scripts/travis-run.sh
+++ b/scripts/travis-run.sh
@@ -29,7 +29,12 @@ then
     RANGE="$TRAVIS_BRANCH HEAD"
     if [ $TRAVIS_PULL_REQUEST == "false" ]
     then
-        RANGE="${TRAVIS_COMMIT_RANGE/.../ }"
+        if [ -z "$TRAVIS_COMMIT_RANGE" ]
+        then
+            RANGE="HEAD~1 HEAD"
+        else
+            RANGE="${TRAVIS_COMMIT_RANGE/.../ }"
+        fi
     fi
 
     # If the environment vars changed (e.g., boost, R, perl) then there's no


daler:
Thanks @acaprez. +1 in general, but I think this currently will fail because the new `[ -z ]` test comes after `set -u` so if unset bash complains even when you're trying to check if unset. Options are to either wrap it in a `set +u; set -u` or move the test up to the top where other defaults are set.

bgruening:
:+1: for moving the test up!
Thanks @acaprez for tackling this!

acaprez:
@daler - hmm, I don't quite see the issue.  The `TRAVIS_COMMIT_RANGE` variable is always set within the Travis environment - it's just an empty string on the initial commit.  (See for example one my [test builds](https://travis-ci.org/acaprez/recipes/jobs/235841436) ).  So the `-z` test should always work.  When running locally via `simulate_travis.py`, the test is never executed since it's within the top level `if [[ $TRAVIS == "true" ]]` block.

daler:
Ah, thanks, you're right, the var exists but is empty on travis. Looks good then!

PR #4827, opened 2017-05-25
Update Segway

rcwchan:

Note: I also ran into #4553

### DIFF
--- a/recipes/segway/build.sh
+++ b/recipes/segway/build.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 export HDF5_DIR=$PREFIX
-$PYTHON setup.py install
+python setup.py install
 
 # Add more build steps here, if they are necessary.
 
--- a/recipes/segway/meta.yaml
+++ b/recipes/segway/meta.yaml
@@ -26,7 +26,7 @@ requirements:
     - optbuild
     - optplus
     - pytables >=3.0
-    - numpy
+    - numpy >=1.1.0
     - forked-path
     - hdf5
     - hgtools
--- a/recipes/segway/meta.yaml
+++ b/recipes/segway/meta.yaml
@@ -26,7 +26,7 @@ requirements:
     - optbuild
     - optplus
     - pytables >=3.0
-    - numpy >=1.1.0
+    - numpy >=1.11
     - forked-path
     - hdf5
     - hgtools
@@ -41,7 +41,7 @@ requirements:
     - optbuild
     - optplus
     - pytables >=3.0
-    - numpy
+    - numpy >=1.11
     - forked-path
     - hdf5
     - colorbrewer
--- a/recipes/segway/meta.yaml
+++ b/recipes/segway/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: segway
-  version: "1.4.3"
+  version: "1.4.4"
 
 source:
-  fn: segway-1.4.3.tar.gz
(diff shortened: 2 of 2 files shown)


rcwchan:
Not sure why the Mac OSX test passes, but the Linux one doesn't...

rcwchan:
Not sure why this is failing when the requirements for Segway have not changed between 1.4 and 1.4.3

rcwchan:
So I had to change `$PYTHON$ to `python` to get `setup.py` to run, not sure why. Now I get this error:

```
  File "setup.py", line 9, in <module>
    from segway import __version__
  File "/opt/conda/conda-bld/recipe_1495815468344/work/segway-1.4.3/segway/__init__.py", line 2, in <module>
    from version import __version__
ImportError: No module named 'version'

```

rcwchan:
@ericr86 The build worked correctly using my modified temporary segway build, so version.py is indeed the issue. I just gave it permissions 755 so that it was the same as the other py files. Could you please update the pypi host?

Thanks!

EricR86:
The permissions have been updated on the latest commit.

rcwchan:
Segway bitbucket commit though, not pypi host?

rcwchan:
`version.py`'s permissions have been updated (triggering a new Segway release on pypi--thanks @EricR86), but the Travis CI build is still failing with the same error. @bgruening I'm stumped as to why, would you mind please taking a look at this?

simonvh:
Have you tried changing:

```
from version import __version__
```

to

```
from segway.version import __version__
```
?

EricR86:
the segway/__init__.py module file does the import of version. In that directory (segway) you can import version and anything from it no problem. Is there a way to check what is currently on the python path? Like say printing out sys.path?

Also why do we have to use the "python" command explicitly for this to work and we seemingly cannot use the $PYTHON environment variable set by the build machine? What is the difference? Does one set up a different python path?

simonvh:
In Python 3 you can't do this and you get exactly this error. However, given that you run python explicitly, might that be Python 3 by any chance?

rcwchan:
@simonvh This is possible (I am not sure exactly how the build machine works), though the build/test is run only for py27 according to the `meta.yaml` file? Is there a way to force the build machine to invoke python2.7 instead?

Perhaps this is related to not being able to run python through `$PYTHON`?

Thanks so much for all the help!

simonvh:
Why can't you use the $PYTHON variable? What's the error there?

rcwchan:
@simonvh The build machine gives the following message when using $PYTHON:

`/opt/conda/conda-bld/recipe_1495747542684/work/segway-1.4.3/conda_build.sh: line 5: /opt/conda/conda-bld/recipe_1495747542684/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/bin/python: No such file or directory`

(Changeset d5489cb: https://travis-ci.org/bioconda/bioconda-recipes/jobs/236165216)

simonvh:
So I have tried to look into this, but I don't have any clues. To me it looks like a conda issue. It might be worth posting an issue [over there](https://github.com/conda/conda/issues). @bioconda/core any other ideas?

johanneskoester:
We are using an old version of conda/conda-build because of other unsolved bugs that occurred in a later version (see bioconda/bioconda-utils#123). Until we use the latest version, we cannot really report bugs there. Or did you try with the latest conda-version locally?

johanneskoester:
Actually, our conda has been updated since you last tried. Let me restart your build...

rcwchan:
Thanks so much for all the help, @simonvh and @johanneskoester!!

PR #4828, opened 2017-05-25
flowtools 1.5.0

cristhomas:


cristhomas:
@bioconda/core 
I'm at a loss here. the 2 issues are:
1/ r-bh exists in another channel. 
so, I added it to bioconda a while back and I'm trying to update the version. why won't it let me?

2/ for r-rrcov:
  - r-base 3.3.1*
  - r-robustbase >=0.92_1 -> r-base 3.3.2*

This one idk how to fix. I ran conda build locally, and it pulls r-base 3.3.2 so the issue didn't come up.

johanneskoester:
We want to avoid duplication. Since r-bh is available in the R channel as well (hence the lint telling you that it is in another channel), you can remove it from bioconda. Also note that we will soon update to a newer r-base, which will involve a lot of rebuilding anyway.

cristhomas:
so what should I do then? r-bh is available, but the error that prompted me to update it said that it was not up to date, so the problem would still be there if I removed r-bh from bioconda, wouldn't it? And would it be better for me to wait until you guys have updated the r-base? I didn't do anything special other than update conda build to have it use r-3.2.2.

johanneskoester:
Updating the R infrastructure will solve your problem, since we can get a newer version of r-bh then (via conda-forge). So, if this is not urgent I suggest to wait.

cristhomas:
OK. Not super urgent, so I'll wait and update. Are you guys moving up to R 3.4.0?

johanneskoester:
That would be ideal. Depends a bit on how fast we can achieve that move on conda-forge.

cristhomas:
ok, thanks for getting back to me!

bgruening:
@cristhomas we are done and have rebuild everything against R 3.3.2. Can you please rebase against master?

cristhomas:
oh wow, sorry, just seeing this now. Yes, I'll work on it.

PR #4835, opened 2017-05-26
New: gatk4 wrapper, Update: gatk-framework

chapmanb:
- gatk4 -- initial wrapper for gatk4 alpha release. This is BSD licensed
  and fully open source. Uses the Broad team's gatk-launch wrapper
  script instead of a custom bioconda wrapper. Only supports local
  execution for now.
- gatk-framework -- Fix problem where we detect java in `/bin/java` even
  if JAVA_HOME is unset. Thanks to Tanner Koomar.
  https://groups.google.com/d/msg/biovalidation/-HzLgUnK2eA/ugmqnPGjAwAJ

### DIFF
--- a/recipes/gatk4/meta.yaml
+++ b/recipes/gatk4/meta.yaml
@@ -1,0 +1,32 @@
+{% set version="4.0a1.2.7.2" %}
+
+about:
+  home: https://www.broadinstitute.org/gatk/
+  license: BSD
+  summary: Genome Analysis Toolkit (GATK4), currently in alpha
+
+package:
+  name: gatk4
+  version: {{ version }}
+
+build:
+  number: 0
+  skip: False
+
+source:
+  fn: gatk-package-{{ version }}-local.jar
+  url: https://github.com/broadinstitute/gatk-protected/releases/download/1.0.0.0-alpha1.2.7.2/gatk-protected-local.jar
+  md5: 632eddb1413774a294c6f3a7388aed85
+
+requirements:
+  build:
+    - gnu-wget
+  run:
+    - java-jdk >=8,<9
+    - python
+
+test:
+  commands:
+    - gatk-launch -h
+    - gatk-launch --list
+    - gatk-launch HaplotypeCaller --help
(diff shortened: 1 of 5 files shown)


recipes/gatk4/meta.yaml
```diff
@@ -0,0 +1,32 @@
+{% set version="4.0a1.2.7.2" %}
+
+about:
+  home: https://www.broadinstitute.org/gatk/
+  license: BSD
+  summary: Genome Analysis Toolkit (GATK4), currently in alpha
+
+package:
+  name: gatk4
```
johanneskoester: @chapmanb why did you call this gatk4? Why not gatk? The gatk-launch script will conflict with future versions anyway.

recipes/gatk4/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+source:
+  fn: gatk-package-{{ version }}-local.jar
+  url: https://github.com/broadinstitute/gatk-protected/releases/download/1.0.0.0-alpha1.2.7.2/gatk-protected-local.jar
+  md5: 632eddb1413774a294c6f3a7388aed85
+
+requirements:
+  build:
+    - gnu-wget
+  run:
+    - java-jdk >=8,<9
```
johanneskoester: This should be the new openjdk package that we stored in conda-forge.

johanneskoester: The same holds for the gatk-framework recipe above.

chapmanb:
Thanks so much for the review. I'll update and push new packages. I also needed to work on the gatk-launch wrapper to allow passing in JVM specific options so can do that in one go and submit a PR upstream to the Broad folks. Apologies, I always get what I'm supposed to do with java-jdk mixed up.

I kept this as a separate package for now since it's alpha and quite different from GATK3 -- open source, different wrapper, different command lines. Most people who just do `conda install gatk` will still want 3.7 for now. We could swap over to gatk once it's validated and tested as the recommended version but wanted to give a handle for people to start working with it alongside GATK 3.

johanneskoester:
Now that you say it, that makes total sense!

PR #4838, opened 2017-05-26
update protrac to 2.3.1

rvalieris:

### DIFF
--- a/recipes/protrac/build.sh
+++ b/recipes/protrac/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 mkdir -p $PREFIX/bin
-chmod +x *pl
-sed -i -e 's/\r/\n/g' proTRAC_2.1.pl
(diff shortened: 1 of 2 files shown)


recipes/protrac/build.sh
```diff
@@ -1,4 +1,6 @@
 #!/bin/bash
+export LC_ALL=C
+export LANG=C
 mkdir -p $PREFIX/bin
 chmod +x proTRAC_$PKG_VERSION.pl
 sed -i -e 's/\r/\n/g' proTRAC_$PKG_VERSION.pl
```
bgruening: I think you also have to use `sed -i.bak` on OSX

rvalieris:
`/usn/bin/penl^M: bad interpreter: No such file or directory`
seems like the `s/\r/\n/g` is replacing all r in the file...

PR #4858, opened 2017-05-29
Add jsdom npm package

blankclemens:

This installs a the jsdom package through npm globally (`-g` option). I'm not sure if this is the way to go, but this allows to use npm packages as conda dependencies. To avoid name conflicts I suggest (and did so here) to prefix these npm packages with `npm-`.

To use such a package in node, e.g. through `require('jsdom')` you have two options:

  - symlink the package locally with `npm link jsdom`
  - tell node where to look for the global packages, e.g. through `export NODE_PATH=$(npm root --quiet -g)`


bgruening:
npm is just the package manager afaik, can we prefix it with `node-`?
@bioconda/core as this is the first node package (I'm aware of), reviews are appreciated.

recipes/npm-jsdom/build.sh
```diff
@@ -0,0 +1,3 @@
+#! /bin/bash
+
+npm install -g jsdom@9.12.0
```
bgruening: Is there no switch to tell npm to install the package into the $PREFIX.

bgruening:
Can we follow the conda-forge schema here: https://github.com/conda-forge/configurable-http-proxy-feedstock/blob/master/recipe/meta.yaml

And add the node version to our matrix?

johanneskoester:
Adding `CONDA_NODEJS` to our matrix is fine. However, I think this recipe should rather go into conda-forge, since it is not bio-related.

blankclemens:
That's a fair point. I did open a PR on conda-forge, anyone has experience on how long it normally takes to get merged?

https://github.com/conda-forge/staged-recipes/pull/3049

blankclemens:
Added to conda-forge: https://github.com/conda-forge/staged-recipes/pull/3049

PR #4865, opened 2017-05-30
Add the bisulfite aligner bison

dpryan79:


bgruening:
Without looking at it more deeply the docker tests do not allow these kind of imput data, and we don't want to put test data into bioconda. A simple `bison --version` is enough to test that the binary is working.

dpryan79:
@bgruening That's fine in the final recipe, but in the interim for MPI programs you have to actually test that the MPI implementation works and provides the required features. As an example, I found that the OSX openmpi implementation doesn't support MPI_THREAD_FUNNELED (i.e., it's useless) and the Linux one doesn't support MPI_THREAD_MULTIPLE. That restricts what can effectively be used. Sure, the tests for that shouldn't go in the final code, but since we're testing stuff on Travis anyway, we might as well test this there too and then remove the tests before merging.

bgruening:
@dpryan79 make sense. But keep in mind that you can add this also to the build.sh file and test such things there ... like `make test`.

PR #4882, opened 2017-06-02
Import smallgenomeutilities 0.1

SoapZA:

### DIFF
--- a/recipes/smallgenomeutilities/0.1/build.sh
+++ b/recipes/smallgenomeutilities/0.1/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+"${PYTHON}" setup.py install
--- a/recipes/smallgenomeutilities/0.1/meta.yaml
+++ b/recipes/smallgenomeutilities/0.1/meta.yaml
@@ -1,0 +1,39 @@
+package:
+  name: smallgenomeutilities
+  version: "0.1"
+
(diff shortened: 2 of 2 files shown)


SoapZA:
@bgruening @johanneskoester 
Any ideas what this conflict is about? I'm not explicitly requesting py3.6. What's the plan with enabling 3.6 for bioconda anyways?
```
  - numpy 1.10*
  - python 3.6*
```

bgruening:
@SoapZA 3.6 is enabled and this just happens because we build against 3.6 already. The error message is confusing but this is a known bug of conda-build there is one other dependency that is not yet available for 3.6 in your dependencies ...
@johanneskoester has done the migration to 3.6. for all packages soon, so either you wait a little bit or try to find this one dependency which is not yet available for 3.6.

johanneskoester:
Python 3.6 migration is almost finished (one or two packages missing). The problem you see here is cause by the numpy pinning in master, which is still 1.10 (and numpy 1.10 is not available for python 3.6). I plan to merge the migration branch into master today. This should solve your problem.

SoapZA:
Thanks guys for the dedicated effort 👍 , I will just wait and fix my other PRs in the mean time

johanneskoester:
You're welcome ;-)! You should be able to go on after PR #4888 is merged.

PR #4883, opened 2017-06-02
Add genologics recipe

vladsaveliev:

Python API for Clarity Genologics LIMS

### DIFF
--- a/recipes/genologics/build.sh
+++ b/recipes/genologics/build.sh
@@ -1,2 +1,2 @@
-$PYTHON setup.py install --single-version-externally-managed --root=/
+$PYTHON setup.py install --single-version-externally-managed
 chmod -R o+r $PREFIX/lib/python*/site-packages/genologics*
--- a/recipes/genologics/build.sh
+++ b/recipes/genologics/build.sh
@@ -1,2 +1,4 @@
-$PYTHON setup.py install --single-version-externally-managed
+#!/usr/bin/env bash
+
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
 chmod -R o+r $PREFIX/lib/python*/site-packages/genologics*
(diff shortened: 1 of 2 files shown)


recipes/genologics/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+build:
+  number: 0
+
+requirements:
+  build:
+    - python
+    - requests
+
+  run:
+    - python 2.6.*|2.7.*  # [py2k]
```
bgruening: can you please remove this here and skip the build for py3k in your build section with `skip: True  # [py3k]`

vladsaveliev: The module is supported by python 3. That's why the following 2 lines - I wanted to specify version ranges for both py2 and py3:
```
    - python 2.6.*|2.7.*  # [py2k]
    - python >=3.4  # [py3k]
```
I also tried one line `python 2.6.*|2.7.*|>=3.4`, but it didn't work properly. Is there a standard way to specify python version range more specifically? I didn't find any examples, probably the way to do is just not to specify it? I can just remove the version ranges.

bgruening: In BioConda we build automatically against py27,py35, py36, see here: https://github.com/bioconda/bioconda-recipes/blob/bulk/scripts/env_matrix.yml#L1

So this is not needed :)

vladsaveliev: So python versions outside of those 3 are not supported anyway?

bgruening: No. Is this a problem for you?

vladsaveliev: Not at all, removed version ranges :)

bgruening: Cool, thanks!

recipes/genologics/build.sh
```diff
@@ -0,0 +1,2 @@
+$PYTHON setup.py install --single-version-externally-managed --root=/
```
bgruening: Sorry, just see this now. Can you please remove the `--root=/`. I don't think this is needed.

vladsaveliev: Removed, thanks!

vladsaveliev:
Thank you for merging and help @bgruening

PR #4895, opened 2017-06-05
update bioconductor_skeleton to auto-find data packages

daler:

URL pattern is different for data packages (e.g. `org.Hs.eg.db`); this update allows the auto-finding to work with these as well.


daler:
ping @bgruening; I added this to `bulk` as well

bgruening:
Ok, thanks for improving this!

PR #4898, opened 2017-06-06
All platform designs

kurtwheeler:


kurtwheeler:
@bgruening @johanneskoester I'm getting

```
conda.CondaMultiError: Placeholder of length '80' too short in package /opt/conda/conda-bld/recipe_1496762523369/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/R/library/affyio/libs/affyio.so.
```

from Travis. Normally when that happens locally conda just rebuilds it with a different placeholder or something. Either way [affyio](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/bioconductor-affyio) is in bioconda-recipes already and I can't change what version Travis tries to use so I don't think I can rebuild it if I want to.

How can this be fixed?

bgruening:
@kurtwheeler you can increase the build number in the affyio package.
@kurtwheeler @johanneskoester these packages are really big and we don't have so much space. Are these packages really needed in BioConda? It is primarily data, right?

kurtwheeler:
@bgruening whether these packages belong in BioConda or not is really up for you to say. I'm not exactly sure what the scope of BioConda is for sure, this just seemed like the right place to put recipes for packages host on bioconductor.org. I'm happy to handle them myself. I can put the recipes in the lab I work in's github organization and I've already created a conda channel for the lab. Let me know if that's what I should do.

I can explain why I have need of these though. The [SCAN.UPC](https://bioconductor.org/packages/3.5/bioc/html/SCAN.UPC.html) package I recently added makes use of these packages to interpret and process gene expression data that is collected by various manufacturers and models of microarrays. Each of these packages corresponds to a different microarray design.

bgruening:
@kurtwheeler sorry if this was misleading, I do see why this data is useful. I'm just a little bit nervous about our quote. We do have 100GB for bioconda and last time I checked this is adding ~50MB per packages (is this true?). So the question is if we want to distribute data via a package manager in general as this can grow quickly ...
ping @bioconda/core

daler:
We do have some data packages already like org.Hs.eg.db, and I find it very useful to be able to put everything in an environment.yml rather than install separately. In those org.db cases, the AnnotationHub package now makes them obsolete (AnnotationHub is an interface for downloading annotation data locally).

I don't think platform data are included in AnnotationHub, and I'm not aware of something similar for platform data, in which case it would be useful to include them here. I suppose it would depend on how much ContinuumIO is willing to increase our quota.

johanneskoester:
I'm fine with including them. I hope to find some time to negotiate an extension of our quota as soon as possible.

bgruening:
Just as an example https://bioconductor.org/packages/3.5/data/annotation/src/contrib/pd.drogene.1.1.st_3.12.0.tar.gz is 47MB in size. Let's assume in average they are 20MB (it seems to be far more). It means 40MB for Linux and OSX, merging it into master means building for 3.3.1, rebuilding it for 3.3.2 means additional 40MB, makes 80MB per package. In this PR are 170 packages, if we are generous this means 10GB of space. This is the very friendly calculation :(

We have encouraged people to __not__ put data into their package, e.g. reference tables, small databases are we going to change this?

Oh and looking at the recent stats I see: `98.4 GB of 100.0 GB` are used. Urgs, so not even sure we can merge this one.

johanneskoester:
Ok, I did not expect this size. I still think it would not be bad to have for convenience, but we cannot merge this at the moment. We first would need to have much more space on anaconda. Let's see what they give us, and then reconsider. @kurtwheeler I am sorry that we cannot immediately consider this. It looks like you have put a lot of work into this :-(.

kurtwheeler:
Hey that's quite alright @johanneskoester . I've already got these packages uploaded to anaconda.org/kurtwheeler so I'm able to use them. This just seems like the right place for them to live long term.

I also used a set of scripts to generate them, which cut down on how much work they were. However those scripts are general enough that I can run them for most of bioconductor. They're smart enough to recursively build recipes for dependencies that are on bioconductor or cran, but fail if the dependencies aren't on one of those two sites. This PR was kind of a test to see what happens when I open a PR with hundreds of recipes in it because I think I could generate thousands of recipes with them.

I'm not sure how many of those are already in bioconda though so the PR wouldn't necessarily include thousands, but I'm curious what might happen if I opened a PR with 2-10x this number of recipes. (I could also write a script to open 2-10x this number of PRs if you didn't want one giant PR). I'm curious what the maintainers of this repo's thoughts are on that kind of thing.

daler:
@kurtwheeler would you mind sharing the scripts? We have https://github.com/bioconda/bioconda-recipes/tree/master/scripts/bioconductor but it would be nice to improve them, e.g. to build recursively.

johanneskoester:

PR #4900, opened 2017-06-06
Preserve the original JAR filename: clinod-1.3.jar

peterjc:

The original recipe accidentally used the Galaxy cache's filename ``clinod_1.3_src_all.jar`` rather than the original filename, ``clinod-1.3.jar``.

This restores the original JAR filename.

### DIFF
--- a/recipes/clinod/clinod.py
+++ b/recipes/clinod/clinod.py
@@ -13,7 +13,7 @@ import shutil
 from os import access
 from os import getenv
 from os import X_OK
-jar_file = 'clinod_1.3_src_all.jar'
+jar_file = 'clinod-1.3.jar'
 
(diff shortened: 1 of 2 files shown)


bgruening:
But the old was also working? Do we need a new package for that or is this is cosmetic?

peterjc:
This broke my first attempt at updating my Galaxy wrapper to use clinod when installed via BioConda, so I figure it might break other wrapper code which looks for the JAR by name.

bgruening:
fair enough, also I think accessing the jar file dynamically is quite hard :)

peterjc:
Yes - this is one reason why I quite like the BioConda policy to provide a wrapper for Java tools which are released just as a JAR file.

PR #4902, opened 2017-06-06
Increase build number for affyio to rebuild with conda > 2.0

kurtwheeler:

In https://github.com/bioconda/bioconda-recipes/pull/4898 I am getting:

```
conda.CondaMultiError: Placeholder of length '80' too short in package /opt/conda/conda-bld/recipe_1496762523369/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/R/library/affyio/libs/affyio.so.
```

I was advised that increasing the build number would cause the package to be rebuilt. However doing so within that PR wasn't sufficient because it doesn't rebuild affyio first causing a failure before it happens. Therefore I'm going to do it in this PR so it is available for that one.

### DIFF
--- a/recipes/bioconductor-affyio/meta.yaml
+++ b/recipes/bioconductor-affyio/meta.yaml
@@ -6,7 +6,7 @@ source:
   url: https://bioarchive.galaxyproject.org/affyio_1.42.0.tar.gz
   md5: aa52c3eacb3401eaf7bdaefea62e6c06
 build:
-  number: 0
+  number: 1
   rpaths:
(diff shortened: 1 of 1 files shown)


bgruening:
We have rebuild all R packages against R 3.2.2 and with that against the longer path length. So I don't think this PR is needed anymore. I will restart #4898 for you.

PR #4917, opened 2017-06-07
MIRA 4.0.2 is GPLv2+ and also available on Mac OS X

peterjc:

Following existing recipe approach and using the pre-compiled binaries from MIRA author Bastien Chevreux for either 64 bit Linux or Mac OS X (Darwin).

### DIFF
--- a/recipes/mira/meta.yaml
+++ b/recipes/mira/meta.yaml
@@ -6,18 +6,17 @@ about:
 
 build:
   number: 2
-  skip: True # [not linux64 and not osx]
+  # TODO - this ought to work on osx as well
+  skip: True # [not linux64]
 
 package:
   name: mira
   version: '4.0.2'
 
 # No build requirements as using the author's binaries.
-# However, these were built expecting libgcc (on Mac).
 requirements:
   build:
   run:
-  - libgcc
 
 source:
   fn: mira_4.0.2.tar.bz2
--- a/recipes/mira/meta.yaml
+++ b/recipes/mira/meta.yaml
@@ -1,25 +1,33 @@
 about:
   home: https://sourceforge.net/p/mira-assembler/wiki/Home/
-  license: GPLv2
+  license: 'GNU General Public License v2 or later (GPLv2+)'
+  license_file: LICENCE
   summary: 'MIRA is a whole genome shotgun and EST sequence assembler for Sanger, 454, Solexa (Illumina), IonTorrent data and PacBio (the later at the moment only CCS and error-corrected CLR reads)'
 
 build:
-  number: 1
-  skip: True # [osx]
+  number: 2
+  skip: True # [not linux64 and not osx]
 
 package:
   name: mira
   version: '4.0.2'
 
+# No build requirements as using the author's binaries.
+# However, these were built expecting libgcc (on Mac).
 requirements:
   build:
   run:
+  - libgcc
 
 source:
   fn: mira_4.0.2.tar.bz2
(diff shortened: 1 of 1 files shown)


recipes/mira/meta.yaml
```diff
@@ -1,11 +1,12 @@
...
-  license: ''
+  license: 'GNU General Public License v2 or later (GPLv2+)'
+  license_file: LICENCE
   summary: 'MIRA is a whole genome shotgun and EST sequence assembler for Sanger, 454, Solexa (Illumina), IonTorrent data and PacBio (the later at the moment only CCS and error-corrected CLR reads)'
 
 build:
-  number: 1
-  skip: True # [osx]
+  number: 2
+  skip: True # [not linux64 or osx]
```
bgruening: You are still skipping here OSX, this is not what you want isn't it?
Just remove this line.

peterjc: Well spotted - I was making assumptions about the operator precedence, and should probably have written ``# [not linux64 and not osx]`` here?

If currently BioConda only supports Mac OS X and 64 bit Linux removing the conditional would be fine for now.

Presumably we don't do 32 bit Linux, nor anything interesting like ARM or ...?

bgruening: Yes, we only do linux64 and osx at the moment.

jerowe:
sorry, this has been sitting for awhile.

Do we have to use static binaries? Those seem to be troublesome more often than not.

As for the libgcc*dylib, I have no idea. I have no experience with OSX. Maybe @bioconda/osx will have some ideas?

peterjc:
The Mac recipe fails with:

```
Fixing linking of /usr/lib/libSystem.B.dylib in /anaconda/conda-bld/mira_1497001237721/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/bin/mira
Fixing linking of @executable_path/../lib/libgcc_s.1.dylib in /anaconda/conda-bld/mira_1497001237721/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/bin/mira
Error: Could not find libgcc_s.1.dylib
```

I though including ``libgcc`` as a run time dependency would have solved this, but it did not.

We could go down the route of compiling MIRA, but from knowledge of multiple queries on the MIRA mailing list of the potential pitfalls I was not inclined to try this myself when the author's binaries have previously just worked.

I could refocus this pull request on just Linux if that is expedient (fixing license, checksums, etc)?

jerowe:
@peterjc, there are a lot of assemblers with complex steps I have never been able to get working on osx. With the biocontainers project it is not that big of a deal anymore. If someone really wants to run their software from a mac they can use the biocontainer.

jerowe:
I contributed the original recipe, and I would have tried to compile it from source. You may of course have better luck. ;-)

cbrueffer:
This needs some updating due to 5f052908d6f9d4412836cf10e652f8ec06788ea0 ; sorry, was unaware of this PR.

peterjc:
@cbrueffer do you agree the license is GPL v2 or later, not just GPL v2?

cbrueffer:
You're right, I stopped reading the docs after "MIRA has been put under the GPL version 2.".

peterjc:
Rebased, if the Mac changes still fail I will remove those, and just focus on the license clarification.

PR #4924, opened 2017-06-08
add singularity

bgruening:

### DIFF
--- a/recipes/singularity/build.sh
+++ b/recipes/singularity/build.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+set -eu
+
+./configure --prefix=$PREFIX --sysconfdir=$PREFIX
+make
+make install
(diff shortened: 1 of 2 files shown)


bgruening:
@jmchilton its tricky to surprise you with a cake ;)

johanneskoester:
Wouldn't this be better suited in conda-forge?

bgruening:
Not sure, but we need a container for this as well.

johanneskoester:
Well, it is a general purpose package. So in the spirit of trying to reduce our storage usage, I think this should at least be submitted to conda-forge in parallel, so that we can remove it from bioconda as soon as it has landed there.

bgruening:
Don't worry it is only 1.4 MB. I will prepare a conda-forge one after our presentation at ISMB.

johanneskoester:
Ok! Thanks.

jakirkham:
FWIW we're adding it to conda-forge in PR ( https://github.com/conda-forge/staged-recipes/pull/4838 ). :)

PR #4937, opened 2017-06-09
added recipes for transrate assembly-evaluation tool

notestaff:

### DIFF
--- a/recipes/transrate/build.sh
+++ b/recipes/transrate/build.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+set -ef -o pipefail
+
+gem source -r https://rubygems.org/
+gem build transrate.gemspec
+gem env gemdir
+gem install --install-dir=$PREFIX/lib/ruby/gems/2.1.0 --bindir=$PREFIX/bin transrate
--- a/recipes/transrate/build.sh
+++ b/recipes/transrate/build.sh
@@ -2,4 +2,7 @@
 
 set -ef -o pipefail
 
+wget https://raw.githubusercontent.com/blahah/transrate/master/LICENSE
+
 gem install --install-dir=$PREFIX/lib/ruby/gems/2.1.0 --bindir=$PREFIX/bin transrate
+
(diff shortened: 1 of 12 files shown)


notestaff:
@bgruening @tomkinsc What should I do when local conda-build succeeds but the travis build fails?  It fails at the test stage, without printing any errors: https://travis-ci.org/bioconda/bioconda-recipes/jobs/241333602 .  Is there a systematic way to debug this?   Thanks a lot!

recipes/transrate/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - zlib
+    - transrate-tools
+    - salmon
+    - blast
+    - snap-aligner
+    - ruby >=2.1.7,<=2.1.7
+
+test:
+  commands:
+     - transrate 2>&1 | grep "by Richard Smith-Unna" > /dev/null
```
bgruening: can you remove the `> /dev/null` I think then we will the see error :)

recipes/transrate/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -ef -o pipefail
+
+wget https://raw.githubusercontent.com/blahah/transrate/master/LICENSE
```
bgruening: I don't think you need this, just remove the license_file as well.

bgruening:
We offer a simulate_travis script that tries to run the same setup as here and help you with debugging.

notestaff:
@bgruening thanks a lot for fixing the ruby recipe!

notestaff:
@blahah I've been trying to make a bioconda recipe for transrate -- this branch is the work so far.  Some issues: The transrate gem on rubygems.org is 1.0.1 but the current version of transrate is 1.0.3 . The dependence on snap-aligner is on 1.0dev.96 , but would 1.0dev.97 work? bioconda has a build for 1.0dev.97 . Would blast+ 2.2.31 work? bioconda has a build for that, but not for 2.2.29 . In bioconda the guideline is to have a recipe for building things from source, so just fetching binaries would not work.

blahah:
@notestaff we can resolve all this, but it will have to wait for a transrate developer to be available. I'll update here when that happens.

notestaff:
thanks!

PR #4950, opened 2017-06-12
Py3 lofreq

ozagordi:
* [x] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [ ] This PR adds a new recipe.
* [ ] This PR updates an existing recipe.
* [x] This PR does something else (explain below).

The existing recipe for `lofreq` only works with python 2.7. This PR adds a recipe for python 3.5 that can be imported by projects requiring python 3.x support.

PR #4955, opened 2017-06-13
MinVar

ozagordi:


ozagordi:
Hi
I apologise for the multiple commits and builds. This now works, provided that one has installed `lofreq` independently. The version on bioconda of `lofreq` only works with python 2.7, there is a related PR #4950.

recipes/minvar/meta.yaml
```diff
@@ -0,0 +1,45 @@
+package:
+  name: minvar 
+  version: "1.2a3"
+
+build:
+  number: 0
+  skip: True # [not py35]
```
bgruening: is this skip really needed? builds for py36 seems to pass

recipes/minvar/licence.md
```diff
@@ -0,0 +1,146 @@
+SOFTWARE LICENSE AGREEMENT (hereinafter referred to as AGREEMENT)
+=================================================================
+
+This is a legal agreement between you (hereinafter referred to as the LICENSEE)
+and the University of Zurich, Rämistrasse 71, 8006 Zurich (hereinafter referred
+to as UNIVERSITY OF ZURICH) pertaining to the right to use the software product
+“MinVar” (hereinafter referred to as SOFTWARE). Using the SOFTWARE indicates
+your acceptance of these terms and conclusion of the Agreement between the
```
bgruening: usually we do not ship license files, is there any reason for it?

ozagordi: Right. Licence is custom and there is a licence file field, I thought to fill it but then I forgot. Is it ok if I remove `licence.md` and delete the `licence_file` field?

recipes/minvar/meta.yaml
ozagordi: Ok, then I will skip py27 only.

recipes/minvar/licence.md
bgruening: This sounds alright, but normally this license file is part of the tarball/package and license_file with copy it by them self. So no need to put it into the repo.

recipes/minvar/meta.yaml
bgruening: Because of the lofreq dependency? Can you wait until this PR is merged?

ozagordi: Sure. At least at the beginning, though, I will still skip py27. I haven't written minvar with support for it. What is your policy for tools that depend on something that is not (yet) on bioconda?

ozagordi:
@bgruening, besides py27 (python 2 is not supported) I had to skip py36, build failed otherwise.

bgruening:
Any idea why py3k failed? Is this on our side?

bgruening:
Can you add this reason as a comment to the skip section.

bgruening:
Build number should start with 0. Thanks!

ozagordi:
Hi @bgruening.
py36 fails because setuptools_scm_git_archive is not available. I added this in a comment and reset the build to 0.

Thanks!

bgruening:
@ozagordi sorry for the slow turnaround, this usually is faster but a lot of us are currently on holiday.
One last question, why do you need scm_git, are you downloading packages that should be a dependency?

ozagordi:
No, `setuptools_scm_git_archive` is a plugin that allows the automatic parsing of the release version from GitHub. As a developer I don't need to manually change the value of a variable every time that I make a new release, the user runs `minvar --version` and the version is always correct.

bgruening:
Thanks @ozagordi! :)

PR #4957, opened 2017-06-13
BEHST

davidechicco:


recipes/BEHST/build.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+#
```
bgruening: please clean this up.

recipes/BEHST/build.sh
```diff
@@ -0,0 +1,45 @@
...
+#$ -cwd
+#$ -S /bin/bash
+#
+set -o nounset -o pipefail -o errexit
+set -o xtrace
+
+echo "Running build.sh"
```
bgruening: can you please also remove all debug output

recipes/BEHST/build.sh
```diff
@@ -0,0 +1,45 @@
...
+mkdir ../temp/
+
+
+sh project.sh ../data/pressto_LUNG_enhancers.bed DEFAULT_EQ DEFAULT_ET
+
+# sudo yum -y install python
+# 
+# sudo yum -y install python-devel
+# 
+# sudo yum -y install epel-release
```
bgruening: cleanups :)

recipes/BEHST/meta.yaml
```diff
@@ -0,0 +1,59 @@
+package:
+  name: behst
+  version: 29efb4c3a8e333a2824430ef32d8bf01
```
bgruening: Is there any real version number available? I fear that this will confuse conda in any future version if you change the hash. E.g. the hash could be "smaller" than an older one.

recipes/BEHST/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+source:
+  fn: 326f13179c55.zip
+  url: https://bitbucket.org/hoffmanlab/behst/get/326f13179c55.zip
+  md5: 59b4e4ede8d1362ccc2de312d8f39421
+
+build:
+  number: 1
+  skip: True # [not py27]
```
bgruening: This does probably not work, you only can use one skip: True. But you can write something like [not py27 or osx]

recipes/BEHST/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+  skip: True # [osx]
+    
+requirements:
+  build:
+    - mercurial
```
bgruening: not needed

recipes/BEHST/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - pandas
+    - zlib
+    - libpng
+    - bedtools
+    - pybedtools
+    - r-base
+    - curl
+    - r-gprofiler
+  run:
+    - mercurial
```
bgruening: not needed

recipes/BEHST/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - zlib
+    - libpng
+    - bedtools
+    - pybedtools
+    - r-base
+    - curl
+    - r-gprofiler
+    
+test:
+  requires:
```
bgruening: all the requirements are not needed, as these are your runtime dependencies
test-reqs are only useful if you need to have deps that are not in your runtime.

recipes/BEHST/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - zlib
+    - libpng
+    - bedtools
+    - pybedtools
+    - r-base
+    - curl
+    - r-gprofiler
+    
+  files:
+    - test.sh
```
bgruening: Can we include more tests?

recipes/BEHST/build.sh
```diff
@@ -0,0 +1,45 @@
...
+echo $(ls)
+
+cd ./bin/
+
+mkdir ../results/
+
+mkdir ../temp/
+
+
+sh project.sh ../data/pressto_LUNG_enhancers.bed DEFAULT_EQ DEFAULT_ET
```
bgruening: What does this do internally? How does it know that it should store everything into $PREFIX?

recipes/BEHST/build.sh
davidechicco: You mean removing the first two lines? Okay

recipes/BEHST/build.sh
davidechicco: Okay

recipes/BEHST/build.sh
davidechicco: I cannot understand this comment

recipes/BEHST/build.sh
davidechicco: Okay

recipes/BEHST/meta.yaml
davidechicco: My package does not have a version number  ( https://bitbucket.org/hoffmanlab/behst ). And I don't know how to set it...

recipes/BEHST/meta.yaml
davidechicco: Okay

recipes/BEHST/meta.yaml
davidechicco: You're wrong, it's needed

recipes/BEHST/meta.yaml
davidechicco: It's was needed on Travis.ci

recipes/BEHST/meta.yaml
davidechicco: Okay I'll remove them

recipes/BEHST/build.sh
bgruening: No, all but the first line :)

recipes/BEHST/build.sh
bgruening: Everything need to end up in $PREFIX. https://conda.io/docs/build_tutorials/postgis.html?highlight=prefix#build-script

recipes/BEHST/meta.yaml
bgruening: Create a tag, called 0.1.

recipes/BEHST/meta.yaml
bgruening: Before you used the URL.

recipes/BEHST/build.sh
davidechicco: Mmmhhh... so? 
I'm not understanding...

recipes/BEHST/build.sh
davidechicco: I read that page but I cannot understand... what is inside $PREFIX? Thanks

recipes/BEHST/meta.yaml
davidechicco: I created a tag called `tag_for_conda2`, but it did not work :-(

recipes/BEHST/meta.yaml
davidechicco: Mmmhhh... okay I'll try

recipes/BEHST/meta.yaml
davidechicco: Yes sure

davidechicco:
Did everything work fine? What's going to happen now?

Thanks

recipes/BEHST/build.sh
nehaljwani: https://conda.io/docs/building/environment-vars.html

davidechicco: I edited `build.sh` this way:
```
PREFIX=$HOME

cd $PREFIX/bin/

mkdir $PREFIX/../results/

mkdir $PREFIX/../temp/
```
Is it correct?

davidechicco: The previous edit did not work.
I changed to this:

PREFIX=.
```

cd $PREFIX/bin/

mkdir $PREFIX/../results/

mkdir $PREFIX/../temp/
```

The recipe passed the tests on Travis.ci. I hope it's fine now.

recipes/BEHST/meta.yaml
daler: The goal is to have users be able to get the most up-to-date version of the package, which in turn means conda needs to be able to figure out what the latest version is. [PEP440](http://legacy.python.org/dev/peps/pep-0440/) is generally a good guide to follow if you haven't decided on versions yet; `0.1` would be a better version than `tag_for_conda2`.  Then you would modify the source URL to reflect the new tag and use that tag as your version.

recipes/BEHST/build.sh
```diff
@@ -0,0 +1,17 @@
...
+PREFIX=.
+
+cd $PREFIX/bin/
+
+mkdir $PREFIX/../results/
+
+mkdir $PREFIX/../temp/
+
+
+sh project.sh $PREFIX/../data/pressto_LUNG_enhancers.bed DEFAULT_EQ DEFAULT_ET
```
daler: `$PREFIX` is the install path, I'm not actually sure where `$PREFIX/..` ends up being when conda installs the package. It's possible it would work on travis-ci, but I think would be broken when someone installs it into their environment.

To avoid cluttering with extra nonstandard dirs `/results` and `/temp`, these should go in `/share`.  Specifically `$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM`. See https://bioconda.github.io/guidelines.html#fsh-section and, probably more useful in this case, https://bioconda.github.io/guidelines.html#java.

daler: Actually now that I read the source, this line is doing a functional test on example data. Is that correct? For bioconda tests we're only concerned with whether things made it onto the path correctly; we don't want to consume community resources on unit tests (see https://bioconda.github.io/guidelines.html#tests).

Also, we have limited space in the bioconda channel. Continuumio is already being generous with our quota, but we're close to the limit. Data should not be included in the package. If you want to provide a mechanism for getting example data for testing or for users to acquire, see for example [this recipe](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/chromhmm) or [this recipe](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/spectacle).

davidechicco: My package on Bitbucket contains data files in the /data/ folder.

What should I do?

Thanks

daler: You can selectively copy only executables to $PREFIX. That [chromhmm recipe](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/chromhmm) I mentioned does that -- it ignores the data directory, only copies over the jar file, and adds a script for downloading the data later.

daler:
@davidechicco it seems that `project.sh` is the main entrypoint, yet that script has hard-coded paths to data files which are expected to be up a directory and over. This makes it challenging to install in an arbitrary environment. It will either take a bit of work in a `download_behst_data.sh` script (fiddling with readlink etc) or some work upstream to fix `project.sh` to not rely on hard-coded paths. Seeing as you're an author of BEHST it would be easiest to handle an optional argument specifying the data directory.

We're aiming for something like this:

```bash
/bin/
  project.sh # symlink to ../share/behst-0.1-0/bin/project.sh
/share/
  /behst-0.1-0
    /bin
      project.sh
    /data
        # data copied over here by a download_behst_data.sh script
    /{results,temp}
        # can be created by download_behst_data.sh script, build.sh, or project.sh. The latter would be cleanest
```

The trick is for the data download script to be able to figure out the path for `/share/behst-0.1-0/bin/project.sh` given the symlink `/bin/project.sh` which should be doable with `readlink`.

davidechicco:
Hey I made a new pull request here: [https://github.com/bioconda/bioconda-recipes/pull/5047](https://github.com/bioconda/bioconda-recipes/pull/5047)

recipes/BEHST/build.sh
```diff
@@ -0,0 +1,13 @@
...
+#
+set -o nounset -o pipefail -o errexit
+set -o xtrace
+
+mkdir -p $PREFIX/bin
+
+cp $SRC_DIR/bin/* $PREFIX/bin
+cd $PREFIX/bin
+
+./download_behst_data.sh
```
daler: This is still downloading data into the package, which we can't support.

recipes/BEHST/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+  version: 0.4
+
+source:
+  fn: a2a25d973909.zip
+  url: https://bitbucket.org/hoffmanlab/behst/get/a2a25d973909.zip
+  sha256: cf4cc7179e6311f72d350c4fd408930b367b568f4ca77469474c4e5044ffd015
+
+build:
+  number: 0
+  skip: True # [not py27 or osx]
```
daler: I don't see any dependencies that are py27 only, is there another reason why you're restricting the version here?

recipes/BEHST/test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+#$ -cwd
+#$ -S /bin/bash
+#
+set -o nounset -o pipefail -o errexit
+set -o xtrace
+
+cd $PREFIX/bin
+
+./project.sh --test
```
daler: A user will not be changing to the bin directory first, so this is not a good test of whether it will actually work in practice.

epruesse:
@davidechicco Can this be closed?

druvus:
@davidechicco I am closing this old PR. Please feel free to reopen if behst version 0.7 still should be added.

PR #4972, opened 2017-06-15
R rebuild against R 3.3.2

bgruening:

This PR has all changes that were necessary to rebuild all* R packages against R 3.3.2. Moreover, we migrated all packages from the r-channel to conda-forge (build against 3.2.2.).
This means we can drop the `r`-channel completely from you build-infrastructure. Which is done here: https://github.com/bioconda/bioconda-utils/pull/122

Please note that if you want to run against R3.3.1 you still need the r-channel. 

Additionally, this means we changed the channel priority to prefer conda-forge over defaults.
Overall, a large change and we want to make you aware of these and we hope it does not break too many packages. We do think it is worth the effort to drop the r-channel.

ping @bioconda/all 

\* a few packages are still on the sorted-r-blacklist and I would appreciate some help from R experts or OSX experts.


rvalieris:
r-ggally is on the blacklist but apparently it built on 3.3.2 successfully (https://anaconda.org/bioconda/r-ggally/files), am I missing something ?

bgruening:
@rvalieris can you remove it from the blacklist?

rvalieris:
sure, I just wanted to double check :)

Issue #4592, opened 2017-05-03
Update pinned versions


johanneskoester:
Please post here to propose an update of the pinned versions (e.g. Python, R, Perl, boost, ...) or add a new pinning.

[**Status**](https://github.com/bioconda/bioconda-recipes/commits/bulk)

### Todo
* [x] Python 3.6
* [x] R 3.3.2

bgruening:
@bioconda/osx we are currently migration to Python 3.6. This is happening on the `bulk` branch.
Look at this beauty: https://anaconda.org/bioconda/repo/history

However, there are a few (;)) packages that are failing and it would be great to get your help here.
For example look here: https://travis-ci.org/bioconda/bioconda-recipes/jobs/229428332 grep some failure and submit the change to the bulk branch. 

Thanks!
Bjoern

johanneskoester:
@bioconda/osx, I am currently trying to fix an issue with conda-build on osx. I hope this this will solve many of the failures we see on osx. I'll keep you posted.

johanneskoester:
Ok, the issue was that travis sometimes mixes up the order of the log messages on osx. If that seems to happen in your case, reloading with F5 should be sufficient to get the right order.

johanneskoester:
Mhm, actually it still does not make any sense... @daler, could you have a look, maybe on a local mac machine? What is wrong there???

johanneskoester:
It looks like after htseq has been built for the first time, the build folder is always the htseq folder and it sometimes tries to rebuild htseq and sometimes tries to build the current recipe but cannot find the correct files... This is really a mystery to me.

johanneskoester:
The build issues have been fixed.

johanneskoester:
@bioconda/core osx migration to py36 is done! On linux some recipes remain but I hope to be done in the next few days.

bgruening:
@bioconda/r @bioconda/core we have started the migration to R 3.3.2.
Please look at the bulk branch and help fixing recipes as much as you can, ever fixed package helps!

Just click here on bulk and look for a failing one:
https://travis-ci.org/bioconda/bioconda-recipes/branches

If you want to get familiar with conda-forge we would also like to move all r-* package from bioconda into conda-forge. I have script to assists you with this!

Help welcome!

bgruening:
[...]

PR #4990, opened 2017-06-19
Bump cutadapt to 1.14

marcelm:

### DIFF
--- a/recipes/cutadapt/meta.yaml
+++ b/recipes/cutadapt/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: cutadapt
-  version: '1.13'
+  version: '1.14'
 
 about:
   home: https://cutadapt.readthedocs.io/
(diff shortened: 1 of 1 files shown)


marcelm:
@bioconda/core I’m getting `$MACOSX_DEPLOYMENT_TARGET mismatch: now "10.7" but "10.9" during configure` for the macOS builds on this version bump. #4592 seems related. Is this going to be fixed soon? I’m a bit reluctant to just set the variable to 10.9 in the build script as some other recipes seem to do. I’d prefer if this were fixed for all recipes.

rvalieris:
yes its the same issue but we (bioconda) haven't discussed it yet.

the gist of it is that the defaults channel and the conda-build tools use 10.7 by default, but conda-forge have decided (https://github.com/conda-forge/toolchain-feedstock/issues/7, https://github.com/conda-forge/toolchain-feedstock/pull/9) to use 10.9.

since we now give priority to the conda-forge channel over the defaults we should also be using 10.9 I think.

@bioconda/core, @bioconda/osx comments please.

bgruening:
@rvalieris I don't think we have much choice, aren't we.
Could you please outline how we should do the migration? Should we simply export `MACOSX_VERSION_MIN="10.9"` during our build on travis or should we also depend on toolchain?

rvalieris:
the variable `MACOSX_DEPLOYMENT_TARGET` should be enough, but maybe we also need to change something it the travis config, from what I understood they made the change because of travis.

rvalieris:
macosx env target has been fixed here https://github.com/bioconda/bioconda-recipes/pull/5039.

marcelm:
Perfect, thanks!

PR #4991, opened 2017-06-19
test passed locally

eschen42:

I needed r-batch; otherwise, `library(batch)` failed. I don't understand why; I am confused by this because my last effort (w4mclassfilter) did not require that I add r-batch to meta.yml.

My R package depends on reshape2 and sqldf.


bgruening:
R 3.3.2 has dropped many libraries from the standard lib. And the difference is that we now build against R3.3.2 and not R.3.3.1. Thanks to R ;(

PR #4993, opened 2017-06-19
clean blacklist

rvalieris:

I going to try to clean the blacklist a bit here. xref: https://github.com/bioconda/bioconda-recipes/pull/4972


bgruening:
Thanks for working on this @rvalieris!!!

sorted-r-blacklist
```diff
@@ -58,20 +58,16 @@ recipes/bioconductor-bubbletree
 # TODO: bioconductor-cexor needs specific version of genomicranges?
 #recipes/bioconductor-cexor
 
-
+# conflicts with recent Rcpp version
+recipes/r-histonehmm
```
bgruening: can we pin an older version, or do we need to have an older version available for R3.3.2 before :(

bgruening:
This is fantastic! Feel free to merge, I will drink a juice and celebrate you now!

sorted-r-blacklist
rvalieris: the main problem is that I have no idea which version of Rcpp it needs, the dev site is not explicit either..

rvalieris:
not so fast ! there are a few more recipes I want to give it a go

bgruening:
You could do this in a second PR isn't it? Tests are taking longer and longer :)
Your call! :)

rvalieris:
oh ! I didnt noticed, just this one more then I will merge.

bgruening:
Thanks @rvalieris!

PR #4996, opened 2017-06-19
Bioconductor dexseq OS X

acaprez:


recipes/bioconductor-dexseq/pythonpath.patch
```diff
@@ -0,0 +1,16 @@
+--- inst/python_scripts/dexseq_count.py	2016-10-17 15:19:41.000000000 -0500
++++ inst/python_scripts/dexseq_count.py.new	2017-06-16 14:14:38.000000000 -0500
+@@ -1,3 +1,5 @@
++#!/opt/anaconda1anaconda2anaconda3/bin/python
```
bgruening: This should be /usr/bin/env python isn't it?

recipes/bioconductor-dexseq/pythonpath.patch
```diff
@@ -0,0 +1,16 @@
...
++
+ import sys, itertools, optparse, warnings
+ 
+ optParser = optparse.OptionParser( 
+--- inst/python_scripts/dexseq_prepare_annotation.py	2016-10-17 15:19:41.000000000 -0500
++++ inst/python_scripts/dexseq_prepare_annotation.py.new	2017-06-16 14:14:55.000000000 -0500
+@@ -1,3 +1,5 @@
++#!/opt/anaconda1anaconda2anaconda3/bin/python
```
bgruening: here as well ...

bgruening:
Thanks @acaprez for working on this!

recipes/bioconductor-dexseq/pythonpath.patch
acaprez: `condo-build` recognizes the `/opt/anaconda1anaconda2anaconda3` string as special and will replace it with the full prefix at install time ([relevant docs section](https://conda.io/docs/building/meta-yaml.html#making-packages-relocatable), second bullet point).

I prefer to use that instead of `/usr/bin/env python` to guarantee the correct interpreter is used.    In theory, `/usr/bin/env python` could point to a different `python`, depending on what else has been done to the `PATH` after activating the environment.

bgruening: I see it more from the other side, if people want to modify things on there own they hopefully know what they are doing and we prevent them from doing it.
I think the majority of packages are using /usr/bin/env and maybe we should make this best practice?

acaprez: Well my reasoning was that since all scripts (`python`, `R`, `perl`, etc.) that are installed as part of normal package installations are treated that way by `conda-build`, it's good to match that behavior.  IMHO, using `/usr/bin/env foo` is contrary to the idea of creating reproducible, portable installs as it potentially allows local variations in a user's environment to alter or break functionality.  That being said, I'm happy to follow whatever the Core team would like to establish as best practice.

bgruening: As this is not a rule now please feel free to leave it like that. We should probably discuss this in an issue and find evidence how conda-forge is doing it.
Thanks for all your efforts!

acaprez: Sure, sounds good - thanks.

PR #5009, opened 2017-06-20
Update of the Blast-legacy/2.2.22 recipe

arbernard:

The main modifications made on the recipe are:

- Update of the download URL since the old one was a dead link
- Management of the BLASTMAT environment variable that need to be set in some case (Example: [Qiime assign taxonomy](https://github.com/galaxyproject/tools-iuc/pull/1365))


bgruening:
@arbernard is there any test which we can run to test this?
The simplest I could imagine is to `echo BLASTMAT` and check if it is set.

bgruening:
The echo would be only useful if you grep in it and fail if the ENV are not present, I added a check to the test, let's see if this turns green as well.

PR #5013, opened 2017-06-21
Update snakemake, try noarch

kyleabeauchamp:

cc @johanneskoester   I'll let you decide to merge since I haven't traditionally touched this recipe, but AFAIK snakemake would be a good candidate for a switch to noarch...

### DIFF
--- a/recipes/snakemake/meta.yaml
+++ b/recipes/snakemake/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "3.13.0" %}
+{% set version = "3.13.2" %}
 
 package:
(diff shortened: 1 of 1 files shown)


kyleabeauchamp:
Is the hashing preference something we want to encourage?  The guidelines are md5-focused (https://bioconda.github.io/guidelines.html).

johanneskoester:
Yes, we should. So, my rule is, we should use sha256 whenever provided. If only md5 is available, it is better to use md5 than to download the package and create sha256 yourself.

PR #5024, opened 2017-06-21
Add MrBayes

epruesse:

### DIFF
--- a/recipes/mrbayes/build.sh
+++ b/recipes/mrbayes/build.sh
@@ -9,12 +9,13 @@ autoconf
     --prefix=$PREFIX \
     --disable-debug \
     --with-beagle=$PREFIX \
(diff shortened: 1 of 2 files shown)


epruesse:
This seems to work now. The issue wasn't really MrBayes itself, but idiosyncrasies of the bioconda build chain. The Linux side is built with root, has no HOME set, no ssh or rsh installed, is run in a conda environment with the conda command unavailable (and not installable, as that's only for the root environment) and hides all output if things go right. Also, `readline` seems to be pulled in in incompatible versions unless explicitly pinned (that's probably conda and/or conda-forge, depending on how you view the issue).

bgruening:
Looks good, this is a complicated package, thanks a bunch for working on this!

PR #5052, opened 2017-06-26
Update htslib to latest 1.5 release

chapmanb:
Bumps CONDA_HTSLIB and re-builds packages that reference it


recipes/ensembl-vep/meta.yaml
```diff
@@ -10,7 +10,7 @@ source:
   md5: ad9059806b061ac11bb3e7d9f51ca30d
 
 build:
-  number: 0
+  number: 1
   skip: True # [osx]
```
bgruening: @chapmanb if such a recipe has a build-string containing CONDA_HTSLIB you don't need to increase the build number, in fact it should be resetted.
A few recipes do not use the build-string ... this seems to be wrong :(

chapmanb: Ooops sorry I'm still learning my way around this magic. I just unset the ones that don't need to be bumped and fixes the one that was missing a build string. Let me know if anything else looks off.

PR #5057, opened 2017-06-27
Nglview 0.6.4

hainm:


daler:
@hainm looks like you might need `libgcc` (https://bioconda.github.io/guidelines.html#c-c)

hainm:
thanks @daler. `nglview` is purely python and javascript that do not require compiler. 

The root the problem is from `zeromq` package from conda-forge channel. 
Do you still think I need to add `libgcc` to nglview too?

daler:
The regular conda test passes, but the much stricter mulled-build test fails. conda-forge doesn't include that test. So it's possible that simply adding libgcc it would help.

PR #5063, opened 2017-06-27
nextflow 0.25.1

pditommaso:

### DIFF
--- a/recipes/nextflow/meta.yaml
+++ b/recipes/nextflow/meta.yaml
@@ -1,23 +1,25 @@
 package:
   name: nextflow
-  version: "0.24.2"
+  version: "0.25.1"
 
 build:
   number: 0
 
 source:
   fn: nextflow
-  url: https://www.nextflow.io/releases/v0.24.2/nextflow
-  md5: 688550f3789d83cd656dc3a55fa80489
+  url: https://www.nextflow.io/releases/v0.25.1/nextflow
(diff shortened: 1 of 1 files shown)


bgruening:
look good, thanks @pditommaso

pditommaso:
@bioconda/core I have problem with the biocontainer created by this recipe. The nextflow installer executed during the build process downloads some JAR libraries that are stored in the path `$HOME/.nextflow`. However in the resulting container none of those files exist thus the container is useless. Do you any idea what's the problem? is the home path (which I suppose is `/root`) stripped during the container build.

bgruening:
@pditommaso everything what is not in $PRRFIX ist stripped away. This also holds true for Conda. Does the conda recipe work? PLease store the JAR file under $PREFIX.

pditommaso:
I see. Yes the recipe works, but the container downloads the required deps each time is launched. To store the JARs under `$PREFIX` I need to define a custom env var. How can I do that? Thanks.

bgruening:
Only during container startup with `-e` I guess. Need to leave, and will be back in the evening, Sorry.

pditommaso:
No hurry. Anyhow I know `-e` but it's not a solution. I would require to add a env variable in a permanent way. Is that possible?

pditommaso:
Just to close this, is there any way to define a custom variable in a Conda recipe / Biocontainer ?

Issue #4848, opened 2017-05-28
gnuplot 5.0.4 errors


jerowe:
```
(/scratch/gencore/software/gnuplot/5.0.4) [gencore@login-0-4 ~]$ :gnuplot
gnuplot: error while loading shared libraries: libjpeg.so.8: cannot open shared object file: No such file or directory
```

I'll be looking into this today.

bgruening:
My guess is that is was not strictly pinned against this older libjpeg version and you get now the version 9 during install. Pinning it would solve it probably.

rvalieris:
this is now fixed on https://github.com/bioconda/bioconda-recipes/issues/5065

PR #5065, opened 2017-06-27
pin jpeg lib on gnuplot recipe

### LINKED ISSUE #2568: gnuplot is missing libjpeg
gnuplot: error while loading shared libraries: libjpeg.so.8: cannot open shared object file: No such file or directory

rvalieris:

fixes #2568 and #4848 

I know gnuplot 5.0.5 is available on conda-forge but this should keep the latest bioconda version working, in case people are still using it.


rvalieris:
@bioconda/core 
I think the test is failing here because the mulled builds don't support the `test::files` directive, is it possible to include these files in the test or should I workaround it ? I could generate the test data on-the-fly I guess.

bgruening:
On the fly would be great. Supporting test files needs a little bit work and I'm a little bit busy currently.

johanneskoester:
I guess you want the test file because a simple gnuplot -h would not discover a certain error?

rvalieris:
ok, I will try to make the test file on-the-fly !
@johanneskoester , the test file was added by someone else, I just want to keep the tests as is, if possible.

croth1:
FYI, we have a properly pinned gnuplot in conda-forge. We might want to move this recipe to bioconda-legacy to avoid double maintenance costs.

johanneskoester:
Yes. Definitely. This is our policy for all non bio-related packages. Can you create a PR that removes the recipe?

rvalieris:
I am aware that gnuplot was ported to conda-forge, however I though the policy was to not remove a package from bioconda unless a equivalent version exists on conda-forge, in this case conda-forge gnuplot have version 5.0.5, and bioconda have 5.0.4, this is why I decided make a new rebuild with the correct pinning.
the recipe itself can be removed now though, I will make a PR later.

johanneskoester:
Yes, that's fine.

PR #5071, opened 2017-06-28
Kraken ea

buchkovm:

### DIFF
--- a/recipes/kraken-ea/build.sh
+++ b/recipes/kraken-ea/build.sh
@@ -4,18 +4,18 @@
 #export LDFLAGS="-L$PREFIX/lib"
 #export CPATH=${PREFIX}/include
 
-mkdir -p "$PREFIX/libexec" "$PREFIX/bin"
+mkdir -p "$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM" "$PREFIX/bin"
 export CXXFLAGS="-std=c++0x -Wall -fopenmp  -DGZSTREAM_NAMESPACE=gz -g -I."
 export CFLAGS="-I$PREFIX/include"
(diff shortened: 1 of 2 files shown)


recipes/kraken-ea/build.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+#export CPPFLAGS="-I$PREFIX/include"
+#export LDFLAGS="-L$PREFIX/lib"
+#export CPATH=${PREFIX}/include
+
+mkdir -p "$PREFIX/libexec" "$PREFIX/bin"
```
daler: I've not seen `libexec` before and I don't see it in the [FSH](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard). Would it be possible to use `$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM` instead?

recipes/kraken-ea/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - libexec/kraken-translate
+
+requirements:
+  build:
+    - perl
+    - zlib
+
+  run:
+    - zlib
+    - jellyfish 1.*
```
daler: I think the `*` notation is still reserved for special things like numpy. Does `jellyfish<2` work?

buchkovm:
The above requests have been successfully incorporated into the build script.

druvus:
What is special with this fork compared to the original kraken?

buchkovm:
This fork has been modified to split the reads into different fastq files based on the taxonomic classification of the read. For example you could end up with one file containing bacterial reads and another containing human reads. The fork is being added in preparation to being the dependency for another tool that utilizes this custom functionality.

PR #5074, opened 2017-06-29
Update MEME suite to version 4.12.0

imk1:


imk1:
This might not work because this requires perl >= 5.24, which is not available on main anaconda channels.

PR #5088, opened 2017-07-03
Update lofreq to 2.1.3.1 with python 3 support

ozagordi:


ozagordi:
Hi @bgruening.
By merging this, I would be able to add lofreq as dependency of MinVar (PR #4955).

Thanks.

bgruening:
@ozagordi the samtools dependency is not needed anymore? In the old recipe was samtools compiled for linux?

ozagordi:
In the old version samtools was not listed as a dependency, but `build.sh` would download, compile and link it (only in the linux version). Same story for htslib.

bgruening:
@ozagordi and this changed? Because you have not defined samtools as dependency, correct?

ozagordi:
This version uses compiled binaries provided by the maintainer for both linux and osx, so there is no need to define samtools as dependency.

bgruening:
@ozagordi ok, just double checking :)

PR #5105, opened 2017-07-04
added perl-bio-eutilities

PatrickRWright:

### DIFF
--- a/recipes/perl-bio-eutilities/build.sh
+++ b/recipes/perl-bio-eutilities/build.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# If it has Build.PL use that, otherwise use Makefile.PL
+if [ -f Build.PL ]; then
+    perl Build.PL
+    ./Build
+    ./Build test
+    # Make sure this goes in site
+    ./Build install --installdirs site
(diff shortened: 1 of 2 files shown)


jerowe:
@PatrickRWright , do the versions need to be hard coded?

PatrickRWright:
Hi, since I am new I am also not exactly sure if the answer is "yes", but  for now my plan was to build something that stably runs for a long time even if other dependencies change. What would you suggestion be?

jerowe:
The suggestion is not unless it has to be. ;-) BioPerl can be problematic. Could you try with just hard coding the bioperl version, and leave the others blank?

PatrickRWright:
Ok. I will do that. I will comment out the versions and leave them as comments as a reminder what definately worked. I hope such comments are fine. Thank you for the suggestions.

jerowe:
One can never have too many comments. ;-)

jerowe:
FYI - if you are relying on particular versions for a major project you could build conda env with your hard requirements. It is what my department does projects in order to maintain reproducibility. All it requires is a anaconda cloud account.

PatrickRWright:
Ok, thanks. It all appears to be working. Can you merge this?

PR #5117, opened 2017-07-05
Dr. Disco v0.8.0 => v0.8.2

yhoogstrate:

### DIFF
--- a/recipes/dr-disco/0.8.0/build.sh
+++ b/recipes/dr-disco/0.8.0/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+sed -i.bak -E 's/setup_requires.+//' setup.py
+sed -i.bak -E 's/install_requires.+//' setup.py
+$PYTHON setup.py install
--- a/recipes/dr-disco/0.8.0/meta.yaml
+++ b/recipes/dr-disco/0.8.0/meta.yaml
@@ -1,0 +1,51 @@
+package:
+  name: dr-disco
+  version: "0.8.0"
+
+source:
+  fn: v0.8.0.tar.gz
+  url:  https://github.com/yhoogstrate/dr-disco/archive/v0.8.0.tar.gz
+  sha256: 64f9c553389869d563de9423fffd6da4e4eb9df734580a313e7015dfc0559f82
+
+build:
+  preserve_egg_dir: true
+  number: 0
+  skip: true  # [not py27]
+
+requirements:
+  build:
+    - python
+    - click
+    - htseq ==0.6.1.post1
+    - htslib ==1.3
+    - pysam ==0.10.0
+    - fuma ==3.0.5
+    - numpy
+    - scipy
+
+  run:
+    - python
+    - click
+    - htseq ==0.6.1.post1
+    - htslib ==1.3
+    - pysam ==0.10.0
+    - fuma ==3.0.5
+    - numpy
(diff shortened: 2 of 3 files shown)


jerowe:
This recipe is fine as is, but instead of overwriting the existing recipe could you update it?

The structure should be:
```
recipes/dr-disco/0.8.0
recipes/dr-disco/meta.yaml
recipes/dr-disco/build.sh
```

yhoogstrate:
Thank you for the great service @jerowe 

> Sometimes I like to be a rebel and make a directory for all versions. ;-)

Do you then still need to symlink the latest version into the 'top' directory?

jerowe:
Thanks. ;-) No, that's not necessary. Conda has ways of figuring out the most recent version, and the bioconda build system doesn't really care which directory they sit in.

Is this ready to merge?

yhoogstrate:
yes it is, thanks @jerowe

jerowe:
You haven't added back in the directories. Were my instructions clear, or do you need me to make a PR?

bgruening:
Looks good @yhoogstrate! Thanks for the review @jerowe!

mbargull:
> This recipe is fine as is, but instead of overwriting the existing recipe could you update it?

Hi @jerowe, were there any specific reasons for your recommendation?
Acc. to the guidelines, this would usually be done if builds of the older versions hadn't been available before, but not when issuing a normal update. Expert from the guidelines (http://bioconda.github.io/guidelines.html#versions):
> In general, recipes can be updated in-place. The older package[s] will continue to be hosted and available on anaconda.org while the recipe will reflect just the most recent package.
>
>However, if an older version of a packages is required but has not yet had a package built, create a subdirectory of the recipe named after the old version and put the recipe there.

jerowe:
Well, I was definitely wrong! I really thought that was the guidelines. 

Thanks for correcting me and sorry for the incorrect instructions!

jerowe:
Also, I didn't realize how my previous message came across until I looked at it. I wasn't trying to be a jerk, sometimes I just type before I think. ;-)

mbargull:
@jerowe, no worries! I totally see the appeal to have previous version easily available and won't deny my first reaction/recommendation might have been quite like yours ;). (BTW: IMHO none of your comments seemed harsh to me as in all of them you tried to or offered help -- that's far from being a jerk :D)

daler:
@mbargull @jerowe if I remember correctly we originally decided on that policy based on the assumption that if an older recipe needed to be looked at for some reason, digging through the git history was preferable to maintaining ever-growing recipe directories. I'm sure we could reconsider though if you think maintaining older recipes is useful.

yhoogstrate:
Is it possible to add the bioconda-recipes git revision to the results of `conda info <package>`?

daler:
Hmm, that's an interesting idea. Pretty sure conda-info package reads the channel repodata hosted on anaconda.org, so I don't think it's something we can control.

However since the recipe is included in the package tarball itself, then we could build a tool that takes a package and the git repo and figures out which commit the recipe came from.

karel-brinda:
I think that the question whether to have version subdirectories is quite important and should be more widely discussed. It depends on what is the final product of Bioconda – whether only the binaries, or the binaries AND the recipes.

I believed for quite a long time that the product are also the recipes, and that they are intended to be useful also out of the Conda system itself.

PR #5119, opened 2017-07-05
Methylcoder

ArneKr:


recipes/methylcoder/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  name: methylcoder
+  version: "1.0"
+
+build:
+  number: 0
+  skip: True # [not py27]
+
+source:
+  fn: brentp-methylcode-59ac293.tar.gz
+  url: https://github.com/brentp/methylcode/tarball/master
```
bgruening: Please download a specific commit as tarball - the master branch can change.

bgruening:
I also changed the python pinning. The skip in the build section should be enough.

ArneKr:
@bgruening I can´t believe that is missed that :(. I saw that somebody added methylcoder a while back to bio.tools and noticed that it had no corresponding package. Guess it would be better to just refer to bwa-meth on the bio.tools page instead of adding an obsolete package here.

bgruening:
@ArneKr yeah, lets add some note to bio.tools instead :)
Can we close this PR?

PR #5121, opened 2017-07-06
OpenMS update to 2.2

blankclemens:

I'll give it a try with Travis, if it exceeds build time I'll switch the skip flag.


recipes/openms/meta.yaml
```diff
@@ -1,18 +1,19 @@
...
+  url: https://github.com/OpenMS/OpenMS/releases/download/Release2.2.0/OpenMS-{{ version }}-src.tar.gz
+  sha256: 61e7d24890abb6f462e2f026efb0dc88b84639b228d96e919c8074d3b11d286e
   
 build:
-    number: 1
-    string: boost1.62_{{PKG_BUILDNUM}}
-    skip: True  ## This package exceeds the travis build time, see the extra section.
+    number: 0
+    string: boost{{ boost }}_{{PKG_BUILDNUM}}
+    skip: False  ## This package exceeds the travis build time, see the extra section.
```
bgruening: @blankclemens if this still exceeds build time, we are allowed to upload it directly, in this case let me know and we do this on Monday.

recipes/openms/meta.yaml
```diff
@@ -22,16 +23,17 @@ requirements:
     - glib
     - autoconf
     - automake
+    - doxygen
```
blankclemens: I thought it would be nice for a standalone installation if it fits the build time. On the other hand I agree there's probably not a real use case - I'll remove it with the next commit.

bgruening:
Cool thanks :)

PR #5124, opened 2017-07-06
Bump WhatsHap to version 0.14

tobiasmarschall:

### DIFF
--- a/recipes/whatshap/meta.yaml
+++ b/recipes/whatshap/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: whatshap
-  version: "0.13"
+  version: "0.14.1"
 
 about:
   home: https://whatshap.readthedocs.io/
@@ -8,25 +8,23 @@ about:
   summary: 'phase genomic variants using DNA sequencing reads (haplotype assembly)'
 
 source:
-  fn: whatshap-0.13.tar.gz
-  url: https://pypi.python.org/packages/e2/f3/c71cb4bc3ba3b1234ee0c5fe729c60440164e07437695ca09c61750e0e11/whatshap-0.13.tar.gz
-  md5: 6c9c8b7d6138b60a5a070f0de59f576f
+  fn: whatshap-0.14.1.tar.gz
+  url: https://pypi.python.org/packages/ec/d4/d6c19b716926895ee3def919ab23194212172ed5cf89ccb91c2949fb258d/whatshap-0.14.1.tar.gz
+  md5: edaaeada862844285de1fee9d76da93d
 
 requirements:
-  # To do: Pinnng of the setuptools version is a workaround. Try to remove it.
   build:
     - gcc
     - python
     - setuptools
-    - pysam <0.9.0
+    - pysam >0.11.0
     - pyvcf
     - pyfaidx
     - xopen
   run:
     - libgcc
     - python
-    - setuptools
-    - pysam <0.9.0
+    - pysam >0.11.0
     - pyvcf
     - pyfaidx
     - xopen


tobiasmarschall:
@marcelm @johanneskoester Hi guys, sorry to bother you. I tried my best, but don't get what's going wrong here. I'm not even sure which of the error messages is the relevant one. Maybe this one?
```
RuntimeError: Setuptools downloading is disabled in conda build. Be sure to add all dependencies in the meta.yaml  url=https://files.pythonhosted.org/packages/source/c/certifi/certifi-2016.9.26.tar.gz#md5=baa81e951a29958563689d868ef1064dr
```
Any clues?

Best,
Tobias

johanneskoester:
https://travis-ci.org/bioconda/bioconda-recipes/jobs/251073536#L1165
Is there a new dependency that does not occur in the meta.yaml?

tobiasmarschall:
No. I just double checked again that we did not introduce any new dependencies since the previous release. Also, I've never heard of this ```certifi``` package that seems to be the culprit. Any advice how to debug this?

druvus:
I have also seen this issue for some packages. I haven't managed to track it down yet but it seems to have started when the channel order were altered and it goes away if I use the old channel order. I am currently a little bit short of time but I will try to spend some time on the issue.

marcelm:
I don’t have a solution, but I have analyzed the problem further.

The error message occurs when some build script tries to download `certifi`. That is a Python package containing SSL certificates.

WhatsHap has a `setuptools` build dependency. With the `bioconda` and `conda-forge` channels enabled, I can run `conda create -n testenv setuptools` to simulate what happens when the setuptools dependency gets installed during the build. Conda gives me this output (truncated):
```

    ca-certificates: 2017.4.17-0      conda-forge
    certifi:         2017.4.17-py36_0 conda-forge
...
    setuptools:      33.1.1-py36_0    conda-forge
...
```
That is, it uses *certifi* from `conda-forge` in version 2017.4.17-py36_0.

[In the `setup.py` of the `setuptools` package](https://github.com/pypa/setuptools/blob/master/setup.py#L66), 'certifi' is listed in the `extras_require` section, and it is pinned to version 2016.9.26. That is the exact version that is mentioned in the "Setuptools downloading is disabled" error message that Travis gives.

I don’t have an idea of *how* these versions actually conflict with each other.

The [setuptools recipe on conda-forge](https://github.com/conda-forge/setuptools-feedstock/blob/master/recipe/meta.yaml) lists `certifi` as a dependency without any version number.

johanneskoester:
@jakirkham please see the post of @marcelm above. It seems like setuptools from conda-forge wants to download a particular version of certifi (which is of course disabled by conda). Have you seen this before? Would it make sense to either relax the pinning within the setuptools setup.py, or synchronize the pinning with the certify package in the recipe dependencies?

jakirkham:
Why is a post-release of `setuptools` getting [installed]( https://travis-ci.org/bioconda/bioconda-recipes/jobs/251073536#L1557-L1559 )? 😕 In any event, it looks like many packages are getting downloaded outside of `conda` that shouldn't be. FWICT this [line]( https://github.com/tobiasmarschall/bioconda-recipes/blob/31ff1c8be45db9148148e5e6cde0be7633439d32/recipes/whatshap/meta.yaml#L34 ) is responsible.

In `conda-forge`, we block this `setuptools` behavior by adding the arguments `--single-version-externally-managed --record record.txt` so things are not install by `setuptools` (given we already provide them 😉). Would recommend doing the same here and expect that will fix the issue.

johanneskoester:
Thanks for the hint. Indeed, we should put this into our lints. @daler, can we already lint build.sh? For the future, I think it would make sense to provide a PR for conda skeleton pypi about this...

daler:
I was hitting this exact issue recently, and the `--single-version-externally-managed --record record.txt` is what eventually solved it. In my case it seemed to be related to dependencies on packages that had included entry points in the meta.yaml (specifically, gffutils depending on pyfaidx).

@johanneskoester we are not currently linting `build.sh` but all the pieces are already in place to do so. I'll add a lint for it.

+1 for updating `conda skeleton pypi` to use this.

johanneskoester:
Thank you @daler! A note: the lint should only fire if the recipe depends on setuptools. The argument does not work in case of distutils.

marcelm:
Thanks, very interesting!

I suggest to also get [the conda docs](https://conda.io/docs/building/meta-yaml.html#script) updated.

WhatsHap also depends on packages that define `entry_points`, including pyfaidx. Weird, I don’t see how that could lead to this problem. I still don’t understand why setup.py tries to pull in those dependencies that it shouldn’t, but at this point I am ready to accept the solution!

I’ve verified locally (using `simulate-travis.py`) that adding `--single-version-externally-managed --record=record.txt` fixes the issue.

daler:
@johanneskoester will do.

@marcelm I feel your pain. I spent way too long trying to figure out how to solve it, and still don't understand why it fixes it. Ah well, at least we have a solution!

jakirkham:
> Weird, I don’t see how that could lead to this problem. I still don’t understand why setup.py tries to pull in those dependencies that it shouldn’t, but at this point I am ready to accept the solution!

So the default behavior of `setuptools` is to pull in all of the dependencies or at least check they are satisfied. Hence this behavior is seen in the log where it notes some dependencies are already there. The fact it doesn't think a working `setuptools` is present is why this is going off the rails. Though, of course, installing dependencies via `setuptools` when using a package manager (e.g. `yum`, `apt-get`, `conda`, etc.) is ridiculous. Hence `setuptools` devs added the `--single-version-externally-managed` option. This is actually what they [recommend]( http://setuptools.readthedocs.io/en/latest/setuptools.html#what-your-users-should-know ) (see "Creating System Packages") to do for packagers. They [force us to use `--root` or `--record`]( http://setuptools.readthedocs.io/en/latest/setuptools.html#install-command ) to note the files installed so they can be properly removed, which IMHO is unnecessary, but fine that is how things are done. All in all this is really just the way we should be installing `setuptools` reliant packages with `conda`. Would add this does not work for `distutills` only packages, but that shouldn't be a surprise as they don't have a way of requiring dependencies for the build/install.

jakirkham:
Would also add without these options, `setuptools` creates `egg`s of packages installed, which is a bit suboptimal from a `conda` point of view. For instance the `egg`s can hard code the version in the `egg` name. If this path starts showing up in your packages, then they can break if the wrong version `egg` is present at run time (compared to build time). We had an issue like this with `conda-smithy` that showed up in some peculiar cases and took a while to fully address.

marcelm:
@jakirkham Thanks for the explanation; so really this option should have been used from the beginning.

jakirkham:
Of course. Correct.

Though again it doesn't work on `distutils`-based packages only on ones using `setuptools`. So I'd be careful about enforcing this across the board. In `conda-forge`, we made a change a long time ago that required `setuptools` be specified as a `build` dependency to have it installed at all. If something similar was (or is) done here at bioconda, then it should be easier to catch this distinction in the beginning.

tobiasmarschall:
I've created a new pull request #5140 that adds ``--single-version-externally-managed --record=record.txt``. Now all tests pass.

bgruening:
Thanks a lot all, especially @jakirkham for your help here!
This was merged and fixed in #5140

PR #5132, opened 2017-07-07
metameta v1.1

pirovc:


pirovc:
@bioconda/core I need some help here. I'm updating the tool to v1.1 (and fixing a dependency on the v1.0). I can't figure out what's wrong from the travis log (PermissionError: [Errno 13] Permission denied: 'locks'). Any ideas?

rvalieris:
is this "locks" a file created by the program ? looks like it got a "permission denied" error trying to delete the file.

johanneskoester:
Mhm, I don't think so... @bgruening it looks like this happens in the cleanup after mulled. Any idea?

pirovc:
@rvalieris This tool is a Snakemake pipeline and afaik it doesn't create such file. Is there a way to debug this problem? Travis log says that the problem is still on the version 1.0.

pirovc:
@bioconda/core simulate-travis.py locally works fine. I removed the old version (1.0) and just kept the update (1.1) and it seems to pass the tests. So it's maybe some kind of problem of having two builds at once?

Is it ok to merge this update the way it is (I need the new version online for a paper submission) and after it's merged I request the pull again to try to solve the problem and have version 1.0 back?

Thanks for the help.

bgruening:
Recipe looks good to me. I can't see any obvious error.
Thanks @pirovc!

PR #5149, opened 2017-07-09
Genonets

ArneKr:

### DIFF
--- a/recipes/genonets/bld.bat
+++ b/recipes/genonets/bld.bat
@@ -1,0 +1,8 @@
+"%PYTHON%" setup.py install 
+if errorlevel 1 exit 1
+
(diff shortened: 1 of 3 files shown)


daler:
@ArneKr can you change 

```bash
$PYTHON setup.py install
````

to 
```bash
$PYTHON setup.py install --single-version-externally-managed --report=report.txt
```
in `build.sh`?

This is because setuptools is a build dependency, and this will avoid problems with packages that depend on `genonets`. We're in the process of adding a lint function to catch this. See https://github.com/bioconda/bioconda-recipes/pull/5124#issuecomment-313715598.

daler:
Thanks!

PR #5157, opened 2017-07-09
[WIP] dependencies for SuperLearner

daler:

### DIFF
--- a/recipes/r-arm/build.sh
+++ b/recipes/r-arm/build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+mv DESCRIPTION DESCRIPTION.old
+grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
+$R CMD INSTALL --build .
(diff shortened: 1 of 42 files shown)


daler:
It probably does. I have these and another 18 deps. I was originally thinking it would take a while to push everything through conda-forge.

daler:
Since I spent the time on the bioconda versions, I just pushed them to verify they build on our system, then will close this and start the conda-forge ones.

bgruening:
I would estimate approximately a week. I have used this script: https://github.com/bgruening/conda_r_skeleton_helper to create a lot of r-packages and it worked quite well. One PR is enough for all of those, but windows can be an issue. They want at least that we try to compile them for windows as well and comment into the recipe if we skip it.

daler:
Nice, thanks for the script! Doesn't seem to be anything too crazy for windows in this batch. A bunch need compiling, but none with additional external libs. I was hoping for collaborators to use these this week, but we have workarounds. This will be a good exercise for me for working with conda-forge and will avoid duplication and migration in the future.

PR #5165, opened 2017-07-10
removed dependency on configureHomer after install

simonvh:

@cokelaer @BioNinja Please have a look. As far as I've tested this should remove the need to run configureHomer after the install. The hard-coded paths are now automatically fixed by conda-build.


cokelaer:
I did not start this recipes so @BioNinja may be more knowledgeable but looks good to me. At least this is an improvment. This is not related to your change but on linux, the final executable were partly missing and we will need to check that at some point.

BioNinja:
it looks nice. @simonvh , does it works on macs

cokelaer:
@simonvh @bioninja let us squash and merge. I do not know the policy / good practice: I'm guessing that simon can squash and merge the changes. We then need to check if the executable are properly installed both for mac and linux.

simonvh:
Yes, I'll merge. I can't test osx as I don't have access to a mac. Let me know if you encounter (or hear of) any issues.

cokelaer:
Okay, so FYI, your changes improve the recipes since we do not need to call another script but the previous problem persists: executables in ./share/homer/bin are all empty. I'll open an issue.

PR #5173, opened 2017-07-11
scg (single cell genotyper)

dlaehnemann:

This is my first contribution, so feedback from @bioconda/core more than welcome!

Also, there seems to be a weird issue that only appears during the `mulled-build` linux-64 tests (in the biocontainer?) on Travis, where calling `scg --version` returns 1 and the copying of testing files to the temporary testing folder seem to not work. You can see the failure here: https://travis-ci.org/dlaehnemann/bioconda-recipes/jobs/252117209

To resolve this for now, as I'm not sure how to go about fixing the `mulled-build`, I restricted install tests for linux-64 to running `scg --help`. But I guess this calls for an issue to wherever the `mulled build` is maintained? Any pointers?


bgruening:
@dlaehnemann the error is due to the fact that the source files are not part of the container. You could put them into $PREFIX and call your tool with `--config_file $PREFIX/examples/config.yaml`

recipes/scg/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+package:
+    name: {{ name|lower }}
+    version: {{ version }}
+
+source:
+    url: https://bitbucket.org/aroth85/scg/downloads/{{ name }}-{{ version }}.tar.gz
+    md5: 6f0fb91266de66d6a9d73fb741f49e13
+
+build:
+    script: $PYTHON setup.py install
```
bgruening: can you add `--single-version-externally-managed --record=record.txt` to the end of setup.py

recipes/scg/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+        - scipy >=0.15
+        - pyyaml
+
+    run:
+        - python
+        - pandas >=0.16
+        - numpy >=1.9.2
+        - scipy >=0.15
+        - pyyaml
+        - libgcc
```
bgruening: is libgcc really needed? You are don't have gcc as build-time dependency. Are you compiling anything?

dlaehnemann: It's what I added after getting this Travis build error for linux-64 (which the addition resolved, exposing the other `mulled-build` errors):
https://travis-ci.org/dlaehnemann/bioconda-recipes/jobs/251981055

recipes/scg/meta.yaml
dlaehnemann: I had that in an earlier version, but that gave the libstdc++ error that I resolved by adding the libgcc runtime dependency below:
https://travis-ci.org/dlaehnemann/bioconda-recipes/jobs/251981055

dlaehnemann: Could try it again, though, if it is useful. What exactly does it do?

dlaehnemann:
The biggest test file is 4.4kb, so nothing too big. The $PREFIX solution requires the test files to be part of the recipe, doesn't it? I.e. manually adding them to the recipe folder... Or would that reference the source directory, directly?

recipes/scg/meta.yaml
bgruening: Ok, than this is a conda-forge error. The do not have libgcc defind every time.

recipes/scg/meta.yaml
bgruening: This should not be related to the libstdc++ version.

bgruening:
@dlaehnemann yes than it would need to be added to the recipe. But such tests are not strictly needed in Bioconda.

dlaehnemann:
Jep, I figured that such tests are not required in Bioconda, with `scg --help` being enough. Just figured I'd keep them for OSX, as they work there and figured I'd mention it, in case someone has a clue to the underlying problem elsewhere (for filing an issue).

I now tried adding the `--single-version-externally-managed --record=record.txt` again and ran `./simulate-travis.py`, where the `docker run` failed giving me:

```
+ /opt/conda/conda-bld/scg_1499781124531/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/bin/python setup.py install --single-version-externally-managed --record=record.txt
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: option --single-version-externally-managed not recognized
```

Am now trying with only `--record=record.txt`...

bgruening:
Mh, you are not using setuptools? Are you using distribute or similar old python installation tools? `--single-version-externally-managed` should be known by standard python setuptools based installations.

dlaehnemann:
Well, it's python 2.7 if that could be a problem. And setuptools is not mentioned as an explicit dependency, should it? Or should it in 2.7? With just `--record=record.txt` the `./simulate-travis.py` succeeds.

bgruening:
Oh yes distutils, I think they have problems with externally-managed. https://bitbucket.org/aroth85/scg/src/a23238507d52990474641e8d9293d90acf473d25/setup.py?at=default&fileviewer=file-view-default#setup.py-1

Just remove it, sorry for the noise.

dlaehnemann:
Done. I left the `--record=record.txt` in for now, but can easily remove it. In any case, thanks for the thorough feedback!

bgruening:
Thanks!

PR #5192, opened 2017-07-12
bringing back metameta v1.0

pirovc:

As stated in the PR #5132 I needed the new version of the metameta tool (1.1) on bioconda but the folder with the version 1.0 was giving a strange error on travis ("permission denied" error trying to delete the "locks" file). I kept just the new version and passed the tests and now I'm bringing the old version back to keep it there.

### DIFF
--- a/recipes/metameta/1.0/build.sh
+++ b/recipes/metameta/1.0/build.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+mkdir -p $PREFIX/bin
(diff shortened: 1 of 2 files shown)


pirovc:
Looks like the errors were happening because the recipe was creating files during the test command. I manage to test the tools without avoiding file creation and it's working fine now. Not sure if it's a problem related to snakemake (because the errors were always related to snakemake folders) or a general conda thing.

PR #5198, opened 2017-07-12
Add run requirements and test for plot-vcfstats

nsoranzo:

### DIFF
--- a/recipes/bcftools/meta.yaml
+++ b/recipes/bcftools/meta.yaml
@@ -10,7 +10,7 @@ about:
     VCF and BCF and streams will work in most, but not all situations.
 
 build:
-  number: 1
+  number: 2
 package:
   name: bcftools
   version: {{ version }}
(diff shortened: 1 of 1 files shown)


bgruening:
Thanks @nsoranzo!

chapmanb:
Nicola and Björn;
This recent addition of texlive-core as a bcftools dependency is causing issues resolving perl with other packages (chapmanb/bcbio-nextgen#2004). It looks like the most recent texlive-core with perl 5.22 support also depends on icu 58, which causes problems if you have icu 56:
```
texlive-core 20170520 pl5.22.2.1_0
----------------------------------
dependencies:
    cairo 1.14.*
    freetype 2.7
    gmp 6.1.*
    icu 58.*
    libpng >=1.6.28,<1.7
    mpfr 3.1.*
    perl 5.22.2.1
    pixman 0.34.*
    zlib 1.2.8
```
As a result you get an older perl version, breaking compatibility with any other up to date perl libraries.

Independently of all the pinning issues, tex is a pretty heavy weight dependency (24Mb by itself plus the large dependency tree). Could we reconsider including this as a required package with bcftools? For something as widely used and popular as bcftools, it seems like we're dragging along a lot for less used functionality. I'd vote for having a more heavy weight `bcftools-plot` package that imports bcftools plus the additional dependencies. Would that work?

kyleabeauchamp:
Hmm, yeah, texlive is a giant dependency to require for bcftools usage...

nsoranzo:
The pinning of texlive-core were updated after the dep was added to bcftools, sorry it's breaking bcbio. We can create a subpackage containing just plot-vcfstats with its own requirements, but I'm on holiday, so feel free to comment out the texlive-core requirement for now.

chapmanb:
Nicola -- thanks so much, I commented these out for now and we can revisit a better fix when you're back from holiday. Hope you're enjoying.

PR #5199, opened 2017-07-12
Monovar

dlaehnemann:

This is some python software by someone else (https://bitbucket.org/hamimzafar/monovar) that I cleaned up quite a bit, got it a fixed version tag and then added a complete set of test files, as some were lacking in the original repo. The result is installation through copying and symlinking files--and as that feels a bit hacky and this is only my 2nd contribution, I'll shout out to @bioconda/core again.


recipes/monovar/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+package:
+    name: {{ name|lower }}
+    version: {{ version }}
+
+source:
+    url: https://bitbucket.org/dlaehnemann/monovar/get/{{ version }}.tar.gz
+    md5: 90b338e4f4021f9808f4fbc222adc2ed
+
+build:
+    script: cp -r ${SRC_DIR}/src/ $PREFIX/lib/monovar && ln -s $PREFIX/lib/monovar/monovar.py $PREFIX/bin/monovar
```
bgruening: can you use a separate build.sh script for this, I think this is more clear

recipes/monovar/meta.yaml
```diff
@@ -0,0 +1,50 @@
+{% set name = "MonoVar" %}
+# MonoVar doesn't have versions, so I'm just freezing the commit 7b47571d85a377c425feea86f6e591ee832fe342 as 'bioconda'
+{% set version = "bioconda" %}
```
bgruening: I don't think this is a good idea. Can you just invent a version, like 0.0.1?

recipes/monovar/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+        - python
+        - numpy
+        - scipy
+        - pysam ==0.8.3
+
+    run:
+        - python
+        - numpy
+        - scipy
+        - pysam ==0.8.3
```
bgruening: Does pysam needs such a strict pinning?

recipes/monovar/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+        - examples/mpileup.ref.fa
+        - examples/mpileup.ref.fa.fai
+        - examples/data_files/mpileup.1.bam
+        - examples/data_files/mpileup.1.bam.bai
+        - examples/data_files/mpileup.2.bam
+        - examples/data_files/mpileup.2.bam.bai
+        - examples/data_files/mpileup.3.bam
+        - examples/data_files/mpileup.3.bam.bai
+    commands:
+        - samtools mpileup -BQ0 -d10000 -f examples/mpileup.ref.fa -q 40 -b examples/filenames.txt | monovar -p 0.002 -a 0.2 -t 0.05 -m 2 -f examples/mpileup.ref.fa -b examples/filenames.txt -o output.vcf
```
bgruening: can you do something like `monovar 2>&1 | grep "specific error message"`

recipes/monovar/meta.yaml
dlaehnemann: Sure. My thought behind this was, that this is not any real version and in case the creator wants to start versioning, I don't want to mess that up. But then I guess that 0.0.1 should always be smaller!

recipes/monovar/meta.yaml
dlaehnemann: Sure.

recipes/monovar/meta.yaml
dlaehnemann: Let me check.

recipes/monovar/meta.yaml
dlaehnemann: I can try. This would also get around the `mulled-build` failing due to the tests on actual files that don't seem to end up anywhere in the docker container -- or at least not where the test command expects them. Or am I missing something about setting up tests on actual `source_files` for the docker container?

dlaehnemann:
@bgruening thanks for the review, I changed everything accordingly. However, [the OSX build](https://travis-ci.org/bioconda/bioconda-recipes/jobs/252990232) hung up on a `TimeoutError: [Errno 60] Operation timed out`, seeming like a connection problem. Can you restart that build? I don't seem to have that power...

dlaehnemann:
Ping: Any chance of getting the OSX build to restart, to see if the TimeoutError goes away? Or should I rebase all of this on a more current version of master?

bgruening:
restarted

bgruening:
Thanks!

dlaehnemann:
Well, thanks for the review!

PR #5207, opened 2017-07-13
Update meta.yml to fix incompatibility issues

davidliwei:
Fix incompatibility issues of vispr and new panda versions >=0.20. Require panda versions to be <0.20.


bgruening:
@davidliwei I think you need to bump the build version so that we get a new package.

PR #5217, opened 2017-07-14
update the latest tRNAscan version

bgruening:

 ping @mictadlo 
I think this is the latest tRNAscan version and not the 1.23, which has a higher version number - it is confusing :(


bgruening:
fix in https://github.com/bioconda/bioconda-recipes/pull/5275

Issue #5134, opened 2017-07-07
`simulate_travis.py --install-requirements` downgrades miniconda root environment


HenningTimm:
As mentioned here https://github.com/bioconda/bioconda-recipes/issues/3808, miniconda now ships with Python 3.6. This creates some issues when using the `simulate_travis.py` script.

I installed a brand new version of miniconda, and installed conda-build and anaconda.
When following the [tutorial on how to contribute a recipe](https://bioconda.github.io/contribute-a-recipe.html), running

    ./simulate-travis.py --set-channel-order
    ./simulate-travis.py --install-requirements

downgrades the root environment of miniconda from 3.6 to 3.5 using packages from conda-forge without any questions asked. 

    timm@neuromancer:~/repos/bioconda-recipes$ ./simulate-travis.py --install-requirements
    Fetching package metadata .............
    Solving package specifications: .
    
    Package plan for installation in environment /home/timm/miniconda3:
    
    The following NEW packages will be INSTALLED:
    
        argh:               0.26.2-py35_1      conda-forge
        bkcharts:           0.2-py35_0         conda-forge
        bzip2:              1.0.6-1            conda-forge
        certifi:            2017.4.17-py35_0   conda-forge
        colorlog:           2.10.0-py35_0      conda-forge
        galaxy-lib:         17.5.9-py35_1      bioconda   
        gmp:                6.1.2-0            conda-forge
        imageio:            2.2.0-py35_0       conda-forge
        involucro:          1.1.2-0            bioconda   
        lzo:                2.06-0                        
        mccabe:             0.6.1-py35_0       conda-forge
        ncurses:            5.9-10             conda-forge
        pandoc:             1.19.2-0           conda-forge
        pyaml:              15.8.2-py35_0      conda-forge
        pydotplus:          2.0.2-py35_0       conda-forge
        pygithub:           1.29-py35_0        conda-forge
        sphinx_rtd_theme:   0.1.9-py35_0       conda-forge
        webencodings:       0.5-py35_0         conda-forge
        whoosh:             2.7.4-py35_0       conda-forge
    
    The following packages will be UPDATED:
    
        _license:           1.1-py36_1                     --> 1.1-py35_1                    
        anaconda:           4.4.0-np112py36_0              --> custom-py35_0                 
        anaconda-navigator: 1.6.2-py36_0                   --> 1.6.3-py35_0                  
        anaconda-project:   0.6.0-py36_0                   --> 0.6.0-py35_0                  
[...]

PR #5227, opened 2017-07-15
support an isolated bioconda-utils conda installation

daler:

## Summary
`./simulate-travis.py --bootstrap /tmp/miniconda` makes it much more pleasant to work on bioconda recipes, and `simulate-travis.py` now more closely mimics exactly what is run on travis-ci.

## Rationale
Previously, contributors were forced to change their root conda environment to run tests locally. This is because `bioconda-utils` requires `conda-build`, which in turn will only work in the root environment. This was frustrating if the root environment was used for something other than bioconda development, and was brought up in #5134, https://github.com/bioconda/bioconda-utils/pull/145#issuecomment-312010030, and #5355.

## This PR
This PR adds the ability to install a separate version of conda, isolated from whatever may already be on the machine, and keeps track of that location during subsequent calls of `simulate-travis.py` -- all while being as minimally invasive as possible. I've changed the travis-ci setup to use this command as well, so local testing will now be a better simulation of the travis environment.

To start using it, first perform the setup:

```bash
./simulate-travis.py --bootstrap /tmp/miniconda 
```

That's it -- then use `simulate-travis.py` as before. You can even remove bioconda-utils and conda-build from your root environment. I've added some additional features as well, see below. 

## Other changes
- Simplified the steps to get started on recipes. All you need to start developing bioconda recipes is Python (2 or 3) and the PyYAML package. Optionally, docker. Everything else is handled automatically. This also simplifies setup on travis-ci.
- `~/.config/bioconda/config.yml` persistently stores the location of the bootstrapped installation so you don't have to specify each time
- Previously, mulled-build tests did not run locally when running `simulate-travis.py`. mulled-build will now run by default, but there is an argument to disable it
- `--git-range` argument to `simulate-travis.py` specifies a git range to be used for both linting and building
- `--skip-linting` argument for . . . skipping linting
- Documentation in bioconda-utils has been updated to reflect these changes.


daler:
@bioconda/core can I get a review on this please? That last commit ead08ce builds a recipe using the new framework to test it, and that works. But it would be helpful just to get another set of eyes on this. Thanks.

simulate-travis.py
```diff
@@ -101,30 +250,37 @@ def _remote_or_local(fn, branch='master', remote=False):
...
 
-    sp.run(
-        [
-            'pip', 'install',
-            'git+https://github.com/bioconda/bioconda-utils.git@{0}'.format(env['BIOCONDA_UTILS_TAG'])
-        ],
-        check=True)
+if args.install_alternative_conda:
+    _install_alternative_conda(args.install_alternative_conda, overwrite=args.overwrite)
+    sys.exit(0)
```
bgruening: We could output here a small message, like conda environment successfully created at ...

simulate-travis.py
```diff
@@ -101,30 +250,37 @@ def _remote_or_local(fn, branch='master', remote=False):
...
+
+if args.set_channel_order:
+    _set_channel_order()
     sys.exit(0)
 
+if args.bootstrap:
+    _install_alternative_conda(args.bootstrap, overwrite=args.overwrite)
+    _set_channel_order()
+    _install_requirements()
+    sys.exit(0)
```
bgruening: Maybe here as well, to tell the user what was done.

simulate-travis.py
```diff
@@ -156,10 +312,18 @@ def _remote_or_local(fn, branch='master', remote=False):
         ('--docker' not in env['BIOCONDA_UTILS_BUILD_ARGS'])
     ):
         env['DOCKER_ARG'] = '--docker'
+        if not args.disable_mulled:
+            env['DOCKER_ARG'] += ' --mulled-test'
 
     # Override env with whatever's in the shell environment
     env.update(os.environ)
+
+    # Only at the very end do we want to modify the path:
```
bgruening: s/do//

daler:
Thanks for reviewing Bjoern!

PR #4888, opened 2017-06-03
Finished Python 3.6 migration

johanneskoester:
* [x] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [ ] This PR adds a new recipe.
* [ ] This PR updates an existing recipe.
* [x] This PR does something else (explain below).

In the past weeks, we have migrated all existing recipes to Python 3.6 (if possible). This PR merges the required changes into master. It marks the successful completion of the migration process.

PR #4803, opened 2017-05-24
MultiQC

ewels:
* [x] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [ ] This PR adds a new recipe.
* [x] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

PR #5249, opened 2017-07-18
MultiQC: v1.1 Release.

ewels:

### DIFF
--- a/recipes/multiqc/meta.yaml
+++ b/recipes/multiqc/meta.yaml
@@ -1,14 +1,14 @@
 package:
   name: multiqc
-  version: "1.0"
+  version: "1.1"
 
 source:
-  fn: v1.0.tar.gz
-  url: https://github.com/ewels/MultiQC/archive/v1.0.tar.gz
-  md5: 0ba67f2fcb20c6d42929069a0865731f
+  fn: v1.1.tar.gz
+  url: https://github.com/ewels/MultiQC/archive/v1.1.tar.gz
+  md5: e7d1b452ba5f89d568c4e68665ccd9f5
 
 build:
-  number: 4
+  number: 0
   preserve_egg_dir: True
   # click needs to be patched in order to properly detect unicode in py3k
   skip: true  # [py3k]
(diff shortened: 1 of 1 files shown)


ewels:
Tests passing, so will merge 👍

micknudsen:
Is there any particular reason why there no longer is a Python 3.6 version?

ewels:
Hi @micknudsen,

Damn, no I hadn't noticed this sorry. We had problems with Python 3 in previous releases due to the `click` dependency (see #4803, #4811 and others). I had thought that @bgruening fixed it in the latter version (possibly with broken containers), but it looks like the `skip` was added again by @johanneskoester in [#4888](https://github.com/bioconda/bioconda-recipes/pull/4888/files#diff-b2ec8fee278f656d88d113e6dfee721a).

The TL;DR; is that the `click` author seems to really dislike python 3 for some reason (😞 ) and that it has difficulties picking up the encoding type in some environments. When this happens [the build fails](https://travis-ci.org/bioconda/bioconda-recipes/jobs/235927260#L677). Whenever I install it manually (`python setup.py install` or `pip install`) it works fine, presumably because I have defined a locale on my system:

```bash
export LC_ALL=C.UTF-8
export LANG=C.UTF-8
```

Further reading from the click author [here](http://click.pocoo.org/5/python3/#python-3-surrogate-handling).

This is obviously pretty frustrating. @johanneskoester / @bgruening - is this a fair description of the problem? Any new ideas of what we can do to fix this, other than waiting for the click author to change something in his package (which I suspect will never happen)..? Could we set a locale as above in the build environment or something?

Phil

bgruening:
@ewels you can try to use:

```
extra:
    container:
         extended-base: true
```

Other than that, I think your summary is pretty complete. There is a change that Python3.7 will fix this to a degree that click developers are happy, but in the meantime it would be nice if they can just issue a warning and not an error.

ewels:
Thanks @bgruening! Out of curiousity, do you know what is it that this container has that the busybox is missing?

bgruening:
Busybox has no locales installed, it is really a tiny box :)

PR #5290, opened 2017-07-20
add imseq

dkoppstein:


bgruening:
@dkoppstein can you compile this package from source? Or do you need a new GCC compiler?

dkoppstein:
@bgruening is it preferable to compile from source rather than using the precompiled binary? If so, perhaps we should update the guidelines. I just took the lazy approach ;-)

bgruening:
@dkoppstein yes it is, but I saw you need seqan so it can be it is not possible atm with our old compiler. But please try.

dkoppstein:
@bgruening oddly, I am able to build this using `conda build` and `simulate-travis.py` on my local machine using the updated commit in the PR, but it is failing on travis for some reason. Do you have any thoughts why this might be? Imseq ships with a minimal seqan so that shouldn't be the issue here.

dkoppstein:
Thanks @bgruening for the helping hand!

bgruening:
@dkoppstein sure thing sorry could't check if it went green.

PR #5302, opened 2017-07-20
add behst

daler:

New PR for BEHST, much simplified after upstream updates.


davidechicco:
Thanks @daler 
So, how should I arrange my tests for the Bioconda recipe now?

You mentioned I should avoid functional tests, but only check if the executable files exist in the global $PATH setting (that is, without absolute paths). Is this correct?
For example, if I understand correctly, my tests should check the existence of the global `hiC_parser.py` file, and not of the `./bin/hiC_parser.py` file. Correct?

To check the existence of the executable files, you mentioned I should replace the `[ ! -f $exeFileName ]` with the `hash` command in the general bash script. Now my new general wrapper script is `./behst.py` in Python 2.7; which Python command should I use to check the existence of a file in a global sense?

Thanks!

daler:
@davidechicco I needed to read your code to see what is happening. Looks like `behst.py` is actually just calling `project.sh`. Over in `project.sh` if `--test` is provided you test for the presence of filenames in the directory from which the script was called rather than the existence of executables on the path. Then later in `project.sh` you're calling the executables with a `./` prefix, expecting them to be in the same directory from where the script was called.

The easiest solution:

- remove all of the `./` prefixes in `project.sh`. Since executables were added to the path in `build.sh`, you can expect they are on the path. You would have to change your documentation to have users put the installation dir on their path.
- either remove the `--test` section, or if you want users to be able to test an installation, have it use `hash`
- in `meta.yaml`, individually test for each executable in subsequent entries to the `test: commands:` section, e.g., 

```yaml
test:
  commands:
    - behst.py -h
    - hiC_parser.py -h
    - download_besht_data.sh -h
    - ...etc
```

Alternatively, I think in this case with many executables being called, a small functional test might actually be appropriate since it would test accessibility of the various executables while doing so -- but only if it's relatively small. We want to avoid consuming resources on travis-ci.  For example, if you could get something like this to work in `run_test.sh`:

```bash
# download a minimal data set for testing
download_behst_data.sh /tmp/behst --small

# runs in a minute or two
besht.py /tmp/besht/tiny.bed /tmp/behst
```

then that should test everything all in one shot.

davidechicco:
Thanks @daler, here are the news.

1.  I removed all the ./ prefixes as you requested. Now I call the scripts with the `python2.7` command.

2. I've some troubles in using the `hash`  command (it works if I check the `./fileName` existence, but it does not work if I check the `fileName' existence). So, for the moment, I removed it.

3. I created a new "small" minimal test as you suggested.  Here's the new `run_test.sh` script:
```
#!/bin/bash
#
#$ -cwd
#$ -S /bin/bash
#
set -o nounset -o pipefail -o errexit
set -o xtrace

# download a minimal data set for testing
./download_behst_data.sh ~/thisBEHSTdataFolder --small

# runs in 4 minutes
./besht.py ~/thisBEHSTdataFolder/pressto_LIVER_enhancers.bed ~/thisBEHSTdataFolder

Is it alright?

Where should I put this new test script? Can you do that?

Thanks

**Comment by daler** [comment]
@davidechicco I made the changes here. The `run_test.sh` script runs the test. You can make your own changes and push them to this branch. For now I pointed to the `tip` branch in `meta.yaml` and disabled the sha26sum and the linting that would complain about that. These will need to be fixed before merging, but pieces are in place now for you to make changes to the upstream, and then either make modifications here or not and re-run tests to get them to pass.

Once the tests pass, push a new tag to upstream (e.g. 0.7) and then edit this recipe to point to that tag and add its sha256sum.

**Comment by daler** [comment]
I missed the python2.7 part. You should add a shebang line `#!/usr/bin/env python` to your scripts so they will run with whatever the user's python is rather than calling with python2.7

**Comment by davidechicco** [comment]
Hi @daler 
I made the changes to my recipe and tried to `git push` it, but it does not work. I get this message:

> $ git push https://github.com/davidechicco/bioconda-recipes/
>
> warning: push.default is unset; its implicit value is changing in
> Git 2.0 from 'matching' to 'simple'. To squelch this message
> and maintain the current behavior after the default changes, use:
> 
>   git config --global push.default matching
> 
> To squelch this message and adopt the new behavior now, use:
> 
>   git config --global push.default simple
> 
> See 'git help config' and search for 'push.default' for further information.
> (the 'simple' mode was introduced in Git 1.7.11. Use the similar mode
> 'current' instead of 'simple' if you sometimes use older versions of Git)
> 
> Username for 'https://github.com': davidechicco
> Password for 'https://davidechicco@github.com': 
> To https://github.com/davidechicco/bioconda-recipes/
>  ! [rejected]        master -> master (fetch first)
> error: failed to push some refs to 'https://github.com/davidechicco/bioconda-recipes/'
> hint: Updates were rejected because the remote contains work that you do
> hint: not have locally. This is usually caused by another repository pushing
> hint: to the same ref. You may want to first merge the remote changes (e.g.,
> hint: 'git pull') before pushing again.
> hint: See the 'Note about fast-forwards' in 'git push --help' for details.

I tried `git pull` and it says "Already up-to-date".
Any idea on what to fix?

Regarding the Python version, my package only works with Python 2.7. I have this setting in the shebang, but you wrote me to call the scripts without the `./`, so I've to specify it.

Thanks

**Comment by daler** [comment]
you are on the master branch; this PR is on the `behst-3` branch

**Comment by davidechicco** [comment]
I tried to move to the current branch (`git checkout behst-3`) and the push, but I get errors again:

davide@opennet-33-58 behst]$ git push
warning: push.default is unset; its implicit value is changing in
Git 2.0 from 'matching' to 'simple'. To squelch this message
and maintain the current behavior after the default changes, use:

  git config --global push.default matching

To squelch this message and adopt the new behavior now, use:

> 
>   git config --global push.default simple
> 
> See 'git help config' and search for 'push.default' for further information.
> (the 'simple' mode was introduced in Git 1.7.11. Use the similar mode
> 'current' instead of 'simple' if you sometimes use older versions of Git)
> 
> Counting objects: 7, done.
> Delta compression using up to 4 threads.
> Compressing objects: 100% (3/3), done.
> Writing objects: 100% (3/3), 408 bytes | 0 bytes/s, done.
> Total 3 (delta 2), reused 0 (delta 0)
> remote: Resolving deltas: 100% (2/2), completed with 2 local objects.
> To git@github.com:bioconda/bioconda-recipes
>    396ff05..a0a21f6  behst-3 -> behst-3
>  ! [rejected]        master -> master (non-fast-forward)
> error: failed to push some refs to 'git@github.com:bioconda/bioconda-recipes'
> hint: Updates were rejected because a pushed branch tip is behind its remote
> hint: counterpart. If you did not intend to push that branch, you may want to
> hint: specify branches to push or set the 'push.default' configuration variable
> hint: to 'simple', 'current' or 'upstream' to push only the current branch.
> 

What should I do? Thanks

**Comment by davidechicco** [comment]
@daler I think I was in the wrong directory, now I tried again and it seemed to work.
The current version should be [here](https://github.com/davidechicco/bioconda-recipes/tree/BEHST/recipes/BEHST).
Also the test on Travis CI is running [here](https://travis-ci.org/davidechicco/bioconda-recipes/builds/256162191).

What should I do now? I would like to deliver the recipe as soon as possible. Thanks

**Comment by daler** [comment]
please see our discussion from a month ago about git, forks, and branches: https://github.com/bioconda/bioconda-recipes/pull/5062#issuecomment-311476662. The branch for this PR is called `behst-3`.

**Comment by davidechicco** [comment]
Thank @daler
I ran some new tests on Travis CI but they [failed](https://travis-ci.org/davidechicco/bioconda-recipes/jobs/256179916). I think it's because Travis wants to use Python 3.5, while it should use Python 2.7.

> updating index in: /opt/conda/conda-bld
> Traceback (most recent call last):
>   File "/opt/conda/lib/python3.5/site-packages/conda_build/utils.py", line 133, in _copy_with_shell_fallback
>     stderr=subprocess.PIPE, stdout=subprocess.PIPE)
>   File "/opt/conda/lib/python3.5/subprocess.py", line 581, in check_call
>     raise CalledProcessError(retcode, cmd)
> subprocess.CalledProcessError: Command 'cp -a /tmp/tmp89ek9jl_/info/recipe/run_test.sh /opt/conda/conda-bld/behst_1500664481051/test_tmp/run_test.sh' returned non-zero exit status 1

Why is conda trying to use Python 3.5 and not Python 2.7?
I inserted the line `#!/usr/bin/python2.7` in each Python script.

**Comment by daler** [comment]
The build system uses python3. The error is trying to copy a file `run_test.sh` which doesn't exist. See https://conda.io/docs/building/meta-yaml.html#test-section and many recipes in bioconda which use a `run_test.sh` file.

Your fork still downloads data in the `build.sh` script. Again _please do not do this_. This PR, on which you are reading this text, is the template to use. It has a `run_test.sh` and `build.sh`. I have set it up so that all you need to do is make changes upstream and re-run test. There should be minimal changes you need to make to this PR.

File: recipes/behst/meta.yaml [hunk 1]
```diff
@@ -0,0 +1,39 @@
...
+  fn: behst-{{ version }}.tar.gz
+  url: https://bitbucket.org/hoffmanlab/behst/get/tip.tar.bz2
+
+build:
+  number: 0
+  skip: True # [not py27]
+
+requirements:
+  build:
+    - python
```
daler: note @davidechicco minimal build reqs needed

recipes/behst/meta.yaml
```diff
@@ -0,0 +1,39 @@
+{% set version = "0.6" %}
+
+package:
+  name: behst
+  version: {{ version }}
+
+source:
+  fn: behst-{{ version }}.tar.gz
+  url: https://bitbucket.org/hoffmanlab/behst/get/tip.tar.bz2
```
daler: @davidechicco note the url is set to `tip` so upstream changes will get picked up here.

recipes/behst/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+mkdir -p $PREFIX/bin
+chmod +x bin/*
+cp bin/* $PREFIX/bin
```
daler: @davidechicco note no downloading of data in build.sh. Just copying executables to `$PREFIX/bin`

recipes/behst/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - bedtools
+    - pybedtools
+    - r-base
+    - r-rcurl
+    - r-gprofiler
+    - gnu-wget
+
+test:
+    commands:
+      - behst.py -h
```
daler: other commands can be added here, but `run_test.sh` below should check everything

daler:
@davidechicco currently failing on OSX because `cd ~` in the download script apparently is not supported in the travis-ci build system. Might want to consider `$HOME` rather than `~`, or just `cd`, or omit entirely (since the path needs to be specified anyway).

Also both linux and osx are failing because there's still a working-directory-specific call to `./project.sh` call at https://bitbucket.org/hoffmanlab/behst/src/3b1d3840500720b7eaf91a56cc304bd16eb1694c/bin/behst.py?at=default&fileviewer=file-view-default#behst.py-66

davidechicco:
Thanks @daler for your help, here I am again.

> The build system uses python3. The error is trying to copy a file run_test.sh which doesn't exist. See https://conda.io/docs/building/meta-yaml.html#test-section and many recipes in bioconda which use a run_test.sh file.

I see. In the last version of my recipe, I replaced `run_test.sh` with `test.sh`. Apparently, this change has not been propagated on Travis CI, and I cannot understand why... do you have any idea?

> Your fork still downloads data in the build.sh script. Again please do not do this. This PR, on which you are reading this text, is the template to use. It has a `run_test.sh` and `build.sh`. I have set it up so that all you need to do is make changes upstream and re-run test. There should be minimal changes you need to make to this PR, but any changes you make should be made to this PR.

Thanks but where can I find your versions of `run_test.sh` and `build.sh` on this PR? At the moment, here is [my last version of the BEHST recipe from the behst-3 branch](https://github.com/davidechicco/bioconda-recipes/tree/behst-3/recipes/BEHST). 

> currently failing on OSX because cd ~ in the download script apparently is not supported in the travis-ci build system. Might want to consider $HOME rather than ~, or just cd, or omit entirely (since the path needs to be specified anyway).

Thanks, I replaced it with `cd $HOME`

> Also both linux and osx are failing because there's still a working-directory-specific call to `./project.sh` call at https://bitbucket.org/hoffmanlab/behst/src/3b1d3840500720b7eaf91a56cc304bd16eb1694c/bin/behst.py?at=default&fileviewer=file-view-default#behst.py-66

I see. I tried to use `project.sh` instead of `./project.sh` but it does not work. Apparently, it needs the complete path. Any suggestion on this issue? Thanks

daler:
@davidechicco I don't see any additional commits pushed to this PR. Try running the following:

```
git clone git@github.com:bioconda/bioconda-recipes.git
cd bioconda-recipes
git checkout behst-3

# edit files as needed, then:

git commit <path to changed file[s]> -m "<commit message>"
git push origin behst-3
```

Then refresh this page, https://github.com/bioconda/bioconda-recipes/pull/5302, and your changes should appear at the bottom and a new test will be triggered.

davidechicco:
@daler I think I was able to solve the `subprocess.call()` function call in the `./behst.py` file.
I updated the recipe in this pull request, by telling it to use the 0.7 version of the BEHST package.
I built my recipe locally on my laptop and it worked well; also my latest runs on Travis CI worked out.

I believe my recipe might be considered ready right now.

Please let me know there are other edits needed from me. Thanks for all! :-)

daler:
@davidechicco looking good with the latest commit to master. I had simplified this recipe so you could push commits upstream and they would be picked up without editing the recipe. Now that the tests pass, you need to:

1. push a new tag and then change the URL from 

```
https://bitbucket.org/hoffmanlab/behst/get/tip.tar.bz2
```

to 

```
https://bitbucket.org/hoffmanlab/behst/get/{{ version }}.tar.bz2
```

2. compute the sha256 of that tarball and add it to the `source:` section
3. remove the `extra:` block at the bottom of meta.yaml

Alternatively if you let me know when the tag is pushed upstream I can make these changes.

davidechicco:
@daler Okay, I tagged the last version of the BEHST package as 0.8, computed the sha256 code of the last package archive, and inserted it in the `sha256` field of the `source` section of the `meta.yaml` file.
[Here are the changes.](https://github.com/bioconda/bioconda-recipes/pull/5302/commits/d351cd8d2098bf8092ef3b91aa77ed67b9e47c70)

Let me know if it's alright, thanks

daler:
@davidechicco please check your sha256, build is failing because of a mismatch.

davidechicco:
@daler  I wrongly got the code from the zip file; now I got it from the ..tar.bz2 file and it should work.

Let me know if it's alright, thanks

daler:
@davidechicco at last!

davidechicco:
Thanks @daler for all your help!

davidechicco:
Hi @daler, sorry to disturb you again. A colleague of mine tried to install BEHST from bioconda (`conda install -c bioconda behst`) and got the following error:

> Fetching package metadata ...........
> Solving package specifications: .
> 
> UnsatisfiableError: The following specifications were found to be in conflict:
>  - enable
>  - keras
> Use "conda info <package>" to see the dependencies for each package.

Do you know what it is about? Any idea on how to solve it? Thanks

bgruening:
Try to add the conda-forge channel as well.

daler:
Specifically, see the "Set up channels" section: https://bioconda.github.io/

davidechicco:
Thanks @daler @bgruening, now it apparently works, but the system says:

> The following packages will be DOWNGRADED:
> 
>     astropy:            1.3-np111py27_0                   --> 1.1.2-np110py27_0             
>     numpy:              1.11.3-py27_0                     --> 1.10.2-py27_0                 
>     scikit-learn:       0.18.1-np111py27_1                --> 0.17-np110py27_1              
>     scipy:              0.18.1-np111py27_1                --> 0.16.0-np110py27_1   

It's weird because I do not specify any package version in the [recipe](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/behst/meta.yaml) (I just say I need Python 2.7).
My colleague does not want to downgrade these packages.
Any idea on why this is happening and how to fix it?

Thanks

bgruening:
@davidechicco my guess is that you are installing in the root environment. This is not encouraged. Please use `conda create -n my_cool_environment -c conda-forge -c bioconda behst`

daler:
also the recipe has these requirements:

```yaml
  run:
    - python
    - pandas
    - bedtools
    - pybedtools
    - r-base
    - r-rcurl
    - r-gprofiler
    - gnu-wget
```

davidechicco:
@bgruening Thanks, I'll tell my colleague to try that.

@daler Yes I know, but how is it related to the downgrade? Thanks

daler:
pandas requires numpy, and those other packages also require numpy . . . but other than that it's tough to say without knowing the entire contents of the existing environment and the full dependency tree of everything.

Using the command recommended by Bjoern I get numpy 1.13.1, so my guess is something being pinned in the existing root env is causing downgrade.

davidechicco:
Okay thanks

PR #5330, opened 2017-07-24
Update goleft recipe to include samtools as a run requirement

armish:
This is specifically for `goleft depth` that parallelizes `samtools` calls. `samtools` is not a monstrous dependency to have so better include it rather than forcing user to install it upon `depth` subcommand failures (cc: @brentp)


bgruening:
@armish can you please bump the build number? Otherwise you will not get a new package.

daler:
Thanks @armish!

PR #5341, opened 2017-07-24
Update wkhtmltopdf to 0.12.4

dhimmel:
This is the first time I've worked with a conda recipe, so please review carefully and let me know if I'm missing anything.

I was not able to build locally due to https://github.com/conda/conda-build/issues/1482, so I'll have to use the Travis builds for feedback.

The goal is to upgrade `wkhtmltopdf` to 0.12.4. In addition, enabling installation tests as well as support for other OSes besides Linux would be ideal.

### DIFF
--- a/recipes/wkhtmltopdf/meta.yaml
+++ b/recipes/wkhtmltopdf/meta.yaml
@@ -4,15 +4,16 @@ build:
 
 package:
   name: wkhtmltopdf
-  version: "0.12.3"
+  version: "0.12.4"
 source:
-  fn: wkhtmltox-0.12.3_linux-generic-amd64.tar.xz  # [linux]
-  url: http://download.gna.org/wkhtmltopdf/0.12/0.12.3/wkhtmltox-0.12.3_linux-generic-amd64.tar.xz  # [linux]
+  fn: wkhtmltox-0.12.4_linux-generic-amd64.tar.xz  # [linux]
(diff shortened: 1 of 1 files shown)


kyleabeauchamp:
LGTM once tests pass, thanks for the version update!

dhimmel:
I'm getting the following [error in the build](https://travis-ci.org/bioconda/bioconda-recipes/jobs/257013662#L1662):

```
wkhtmltopdf: error while loading shared libraries: libfontconfig.so.1: cannot open shared object file: No such file or directory
```

@kyleabeauchamp or others: any idea on how to fix this?

One [possible solution](https://stackoverflow.com/a/13046682/4651668) is `apt-get install libfontconfig1`, but then would users be required to also have this package installed? See also https://github.com/conda/conda/issues/1051#issuecomment-67635602.

bgruening:
You can have a runtime dependency on fontconfig, but we should try to compile from source. Can you please try this and show us the error message which you get?

dhimmel:
FYI, I don't intend to work on this PR anymore. We [switched](https://github.com/greenelab/manubot-rootstock/pull/78) from using `wkhtmltopdf` to `weasyprint`.

Furthermore, installing bioconda development environment modified my root environment, which annoyed me.

Others should feel free to pick up where I left off / close this PR for house-keeping.

daler:
@dhimmel thanks for the update. I hope you aren't discouraged from contributing to bioconda in the future -- we now create a completely separate miniconda installation (see http://bioconda.github.io/contribute-a-recipe.html) to avoid touching the root env.

PR #5356, opened 2017-07-26
Add recipe for ushuffle

s-will:

### DIFF
--- a/recipes/ushuffle/meta.yaml
+++ b/recipes/ushuffle/meta.yaml
@@ -27,4 +27,4 @@ requirements:
 test:
   commands:
     - ushuffle --help | head -n1
-    - python -c "import ushuffle; s='xyxyxxy'; print(s, '-->', ushuffle.shuffle(s, len(s), 2))"
+    - python -c "import ushuffle; s=(chr(65)*2+chr(66))*3; print(s,ushuffle.shuffle(s,len(s),2))"
--- a/recipes/ushuffle/meta.yaml
+++ b/recipes/ushuffle/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: ushuffle
-  version: 1.0.0
+  version: 1.2.0
 
 about:
   home: http://digital.cs.usu.edu/~mjiang/ushuffle/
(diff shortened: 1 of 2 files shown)


s-will:
Hi @bgruening, the recipe linting is a great feature! However, I am afraid it is wrong in the case of multiple source files. I included md5 hashs for each of the three source files---it seems conda build recognizes them.

daler:
@s-will I was under the impression that conda uses multiple source urls as alternatives in case one fails. Now that I look more closely, it looks like you're using a new feature of conda-build 3: https://conda.io/docs/building/meta-yaml.html#source-section. Are you using conda-build 3 locally? What happens if you try [skipping the linting](https://bioconda.github.io/linting.html#skipping-a-linting-test) on travis-ci?

We're not yet supporting conda-build 3. Hopefully soon. And when that happens I'll fix the linting to support multiple URLs like this.

In the meantime, assuming this multiple-source feature is only available in conda-build 3,  I'm actually not sure how best to write this recipe. The only way is to download the separate source files individually? No zip or tarball or not available on github or bitbucket or anything?

recipes/ushuffle/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+about:
+  home: http://digital.cs.usu.edu/~mjiang/ushuffle/
+  license: custom
+  summary: uShuffle---a useful tool for shuffling biological sequences while preserving the k-let counts
+
+build:
+  number: 0
+
+source:
+  - url: http://digital.cs.usu.edu/~mjiang/ushuffle/main.c
```
bgruening: I don't think this will work. You can only have one URL.
Moreover, these URLs do not seem to be versioned. As a last resort you can pack those and put them into https://github.com/galaxyproject/cargo-port

s-will:
Hi @bgruening, hi @daler, many thanks for the fast response. Yes, I used conda-build 3.x locally; unaware that I used an unsupported feature ;). 
Strangely enough, the author's decided to distribute everything in single files. Someone else has set up a github repository  guma44/ushuffle. Would it be preferable to use this github repo or rather use cargo-port?

bgruening:
I guess this is a matter of trust in this 3rd party repo :)

s-will:
That's why I rather wanted to link from the author's. Since we have hash sums, it should be reasonably safe to use the repo (after checking diffs to the original files). On the other hand there is little control; e.g. the recipe would be broken if the repo is deleted (or even only updated, since there are no releases).

bgruening:
For github repos you can get tarballs for every commit and as soon as we merge the PR we mirror the tarball so this should address your concerns.

s-will:
It seems like there is a quoting problem in the test commands now; I am going to simplify the tests, but  it could be interesting to fix the quoting issues  (or catch it during linting).

PR #5369, opened 2017-07-28
MultiQC: Extended container base for py3 with click.

ewels:

Suggestion by @bgruening to get python 3 builds to work with click. See discussion in #5249.

### DIFF
--- a/recipes/multiqc/meta.yaml
+++ b/recipes/multiqc/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: multiqc
-  version: "1.1"
+  version: "1.2"
 
 source:
-  fn: v1.1.tar.gz
-  url: https://github.com/ewels/MultiQC/archive/v1.1.tar.gz
-  md5: e7d1b452ba5f89d568c4e68665ccd9f5
+  fn: v1.2.tar.gz
+  url: https://github.com/ewels/MultiQC/archive/v1.2.tar.gz
+  md5: 55dac8b0dc055810ae09ec9b72e0603e
 
 build:
   number: 0
@@ -21,8 +21,10 @@ requirements:
     - lzstring
     - jinja2 >=2.9
     - matplotlib
+    - markdown
     - numpy
     - pyyaml
+    - requests
     - simplejson
     - spectra
 
@@ -33,8 +35,10 @@ requirements:
     - lzstring
     - jinja2 >=2.9
     - matplotlib
+    - markdown
     - numpy
     - pyyaml
+    - requests
     - simplejson
     - spectra
     - libgcc  # [linux]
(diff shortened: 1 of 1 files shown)


ewels:
Hmm - getting close to working! From the last couple of builds it looks like the linux environment wants `C.UTF-8` or `en_US.utf8` as the locale string but OSX wants `en_US.UTF-8` 😠 .

@bgruening - any ideas how to set this in a system-specific manner?

Phil

bgruening:
@ewels you could try something like.
```
    - export LC_ALL=$(locale -a | grep en_US | grep UTF) && export LANG=$(locale -a | grep en_US | grep UTF) && multiqc --version  # [linux]
   -  multiqc --version  # [osx]
```

Sorry, for all the pain!

daler:
@ewels not sure if this will cause more or less pain in the long run, but [`argh`](http://argh.readthedocs.io/en/latest/) is quite similar to click and I've used it a lot. And has no complaints with py3. Looking at the main multiqc script, I think it's almost a drop-in replacement.

ewels:
Thanks both! @bgruening - I take it that it's the `# [linux]` and `# [osx]` that do the system specific stuff? The `$(locale` stuff was my attempt at pulling the correct string manually. Using the comments should be better, will try that now instead.

@daler - yes I have wondered about switching before, even back to argparse which I used in the early days of MultiQC. I use quite a bit of the fancy click stuff though - eg. using setuptools to allow external plugin tools to add their own command line options for example, and even simple stuff like adding a progress bar whilst looping through file searches. So I can imagine that it would still end up taking quite a bit of time...

daler:
@ewels ah, I didn't know it was click that handled the plugin command-line additions and progress bar stuff. That would be an awful lot of work to replicate. 

Also, the `# [linux]` and `# [osx]`  are [preprocessing selectors](https://conda.io/docs/building/meta-yaml.html#preprocessing-selectors) for conda recipes.

ewels:
Aha, nice - thanks! I've seen them around before but was never sure if they were literally just comments or whether they were doing something actively.. Good to see the docs!

ewels:
Linux is being difficult - it's really struggling to set a locale properly (see [here](https://travis-ci.org/bioconda/bioconda-recipes/jobs/258581057#L2039-L2040) and different approach [here](https://travis-ci.org/bioconda/bioconda-recipes/jobs/258584133#L928-L929)).

Final trial-and-error attempt in the latest commit, but going on holiday for a week now. Will come back to it after that.

Phil

ewels:
Ah, oops - sorry, didn't realise that this PR was still open and just pushed the v1.2 release..

ewels:
ok awesome - the Travis build seems to be passing too 🎉

bgruening:
Cool! Great @ewels - hope you had some great holidays.

ewels:
Great stuff - thanks @bgruening! Yup, very chilled holidays thanks ☺️

ewels:
So annoyingly I accidentally overwrote whatever changes we were trying in this PR when I pushed the v1.2 release. Meaning that the v1.1 and v1.2 releases are both still only Python 2.7.

I'll open a new PR and try again with various strategies for getting py 3 to work (I was kind of surprised when the tests passed here so easily!)

Phil

PR #5371, opened 2017-07-28
Add IGV recipe

mvdbeek:

### DIFF
--- a/recipes/igv/build.sh
+++ b/recipes/igv/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+cp *.jar $PREFIX/bin
+cp igv.sh $PREFIX/bin/igv
+chmod +x $PREFIX/bin/igv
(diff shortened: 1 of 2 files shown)


mvdbeek:
I guess we'd need the full X11 stack to make the test pass ... or I can disable the test.

bgruening:
Could you try to add the X11 stack? Otherwise, relaying the test would be fine with me.

mvdbeek:
I don't think we have a complete X11 stack, see https://github.com/conda-forge/staged-recipes/issues/2587.

Another issue with including X11 would be that in most instances people installing IGV already have an Xserver running, so that would lead to conflicts, I believe. Do you know if there is mechanism in conda that would selectively decide whether or not dependencies such as X11 are needed ?

I guess I can disable the test for now ...

bgruening:
Is this package meant to run IGV on a Desktop not in a headless mode?

mvdbeek:
> Is this package meant to run IGV on a Desktop not in a headless mode?

I suppose that would be the most common use ... while ultimately I am going to use it in headless mode. so ...

bgruening:
Thanks!

PR #5398, opened 2017-07-31
survminer R/CRAN package

fabio-cumbo:

Contains the function 'ggsurvplot()' for drawing easily beautiful and 'ready-to-publish' survival curves with the 'number at risk' table and 'censoring count plot'. Other functions are also available to plot adjusted curves for ‘Cox' model and to visually examine ’Cox' model assumptions.

### DIFF
--- a/recipes/r-survminer/build.sh
+++ b/recipes/r-survminer/build.sh
@@ -1,0 +1,13 @@
+#!/bin/bash
+
+# R refuses to build packages that mark themselves as Priority: Recommended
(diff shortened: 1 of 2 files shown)


daler:
Hi @fabio-cumbo, this recipe is not bioinformatics-specific and so should be added to the conda-forge channel instead of bioconda. I'll add similar comments on other PRs for non-bioinformatics packages that point to this one to keep as a single place for discussion.

PR #5423, opened 2017-08-02
Addition of mQC recipe

StevenVerbruggen:
This PR adds a recipe for mQC (MappingQC), a quality control tool for assessing mapping quality in ribosome profiling experiments.


recipes/mqc/run_test.py
```diff
@@ -0,0 +1,22 @@
+#Test Python modules installation
+import traceback
```
bgruening: You can put this into the mata.yaml as well: https://conda.io/docs/building/meta-yaml.html#python-imports

recipes/mqc/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - matplotlib
+    - seaborn
+    - r-base
+    - perl
+    - perl-app-cpanminus
+    - perl-dbi
+    - perl-parallel-forkmanager
+
+test:
+  requires:
```
bgruening: This is not needed as you already have this in your run-time environment.

recipes/mqc/run_test.pl
```diff
@@ -0,0 +1,11 @@
+# Test file for testing Perl modules installation
+$|=1;
```
bgruening: Similar to python tests you can add these for perl as well. E.g. here: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/perl-gd/meta.yaml#L29

recipes/mqc/build.sh
```diff
@@ -0,0 +1 @@
+#! /bin/bash
```
bgruening: This file is empty, it should install your package.
So depending on your installation mechanism you need to add here: `python setup.py install ...` or simply copy your script to $PREFIX.

StevenVerbruggen: As the package exists of one main perl script and a folder with some extra dependency scripts in python and R, I suppose I need to copy all those scripts here to $PREFIX?

bgruening: Yes, you entrypoint should be in $PREFIX/bin

recipes/mqc/run_test.pl
StevenVerbruggen: And how can you make sure that he uses the perl 'use' tests for the perl modules instead of the python 'import' test, which he needs to use for the python modules?

bgruening: Interesting, what happens if you try this? I never had a package with perl and python at once I guess. However, if this does not work you can also use `python -c "import numpy"`

StevenVerbruggen: Now I put the perl packages in the test section of the meta.yaml together with the python packages but he tries to test the perl packages with a python 'import' test, which makes the tests and thus also the whole build fail

StevenVerbruggen: Easiest solutions to me seems to use the "tests: imports:" for python packages and the "test: commands: perl -e "use Parallel::ForkManager;" for perl packages

Issue #5383, opened 2017-07-31
Adding gff3_to_embl and genometools python bindings


Cashalow:
Hello everyone,

I'm trying to add gff3toembl to bioconda. It needs the python bindings to genomictools-genomictools to work.

I tried modifying the recipe of genomictools to put the python bindings during the build. However there is now a problem during the mulled-test of genomictools. This is related to the cairo packages because it fails because its lacking libraries related to X (libXrender, libXext etc...) and completes with the flag cairo=no. Do you think I can update the genometools recipe to discard the cairo support (and annotationsketch and related )? I have been trying for a while to pass the mulled-test with cairo but I think I am now completely stuck.

PR #5439, opened 2017-08-04
Add python binding genometools

Cashalow:

This is an update to add the genometools python bindings in order to add gff3toembl afterwards. (ref https://github.com/bioconda/bioconda-recipes/issues/5383 )
However, independently of my update the build now wouldn't pass the mulled-test with cairo activated so I had to remove it. I tried without removing the cairo dependencies but the test never passed.

### DIFF
--- a/recipes/genometools-genometools/build.sh
+++ b/recipes/genometools-genometools/build.sh
@@ -1,15 +1,16 @@
 #!/bin/bash
 
 
-cp -R gtpython/gt ${PREFIX}/lib/python${CONDA_PY:0:1}.${CONDA_PY:1:2}/site-packages/
+#cp -R gtpython/gt ${PREFIX}/lib/python${CONDA_PY:0:1}.${CONDA_PY:1:2}/site-packages/
 
 #sed -i.bak "s/DEPLIBS:=/DEPLIBS:=-lsupc++ /g" Makefile
 
 #make -n > log.txt 
 #grep "lsupc++" log.txt
-make useshared=yes
+make
 export prefix=$PREFIX
 make install 
 
-
+cd gtpython
+$PYTHON setup.py install
 
--- a/recipes/genometools-genometools/build.sh
+++ b/recipes/genometools-genometools/build.sh
@@ -1,26 +1,6 @@
 #!/bin/bash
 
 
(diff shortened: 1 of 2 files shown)


recipes/genometools-genometools/build.sh
```diff
@@ -1,4 +1,10 @@
 #!/bin/bash
-make
+
+
+make cairo=no
```
bgruening: Any reason to disable it? I think we have all cairo dependencies in conda-forge.

Cashalow: When including cairo, the mulled test fails with 
gt: error while loading shared libraries: libXrender.so.1: cannot open shared object file: No such file or directory
(See for instance : https://travis-ci.org/Cashalow/bioconda-recipes/jobs/260260440)

So after adding xorg-libxrender, xorg-libxext and xorg-libsm (that are also needed, see https://travis-ci.org/Cashalow/bioconda-recipes/jobs/260260440 and https://travis-ci.org/Cashalow/bioconda-recipes/builds/260223795) the mulled-test still fails with 

gt: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory ( https://travis-ci.org/Cashalow/bioconda-recipes/jobs/260223797) 

which does not make a lot of sense to me. 

I tried also building genometools without using all the dependencies already bundled, which is the standard behavior, but I think I am way out of my league here. I think the problem is related to this discussion https://github.com/genometools/genometools/issues/742

Cairo is the only not dependency not bundled with genometools, I think that the problem comes from this but I'm not sure how to solve it. I think the previous build passed because at the time the mulled-test was not used ?

bgruening:
> gt: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory ( https://travis-ci.org/Cashalow/bioconda-recipes/jobs/260223797)
which does not make a lot of sense to me.

This usually means that you need to include `libgcc` as well.

Thanks for trying!

Cashalow:
Indeed it's now working... Thank you very much for the tip and sorry for this oversight.

bgruening:
@Cashalow cool, glad I could help!

PR #5445, opened 2017-08-04
Add gff3toembl

Cashalow:

This PR adds a recipe for gff3toembl. It contains two fixes:

For python version other than 2.7, one needs to change the module urllib to urllib.parse.
The python bindings of genometools are found in lib/python2.7/site-packages, I have to link them if other versions of python are used. I don't know if it is the proper way to do.

Cheers,

SL


recipes/gff3toembl/build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+
+if [ "${CONDA_PY}" != "27" ]
+then
+    sed -i.bak "s/urllib/urllib.parse/" gff3toembl/EMBLContig.py
```
bgruening: Can we patch this properly using `six` or can this be upstreamed?
Thanks @Cashalow!

Cashalow:
Thanks for your help again ! Following your advice, I also sent a PR to sangertools/gff3toembl/ to fix this. They did not use to have support for python 3 because the genometools bindings used to work only for 2.7, but now that they do, one can fix this in gff3toembl.

bgruening:
This really rocks! If you like you can use condas patch functionality to get your patch in here in the meantime until we get a new release.

Cashalow:
Ok sorry about that, I'm still learning the proper way of doing things, thanks for all the indications! I hope this does the trick.

bgruening:
Great! Thanks @Cashalow!

PR #5447, opened 2017-08-05
Remove urllib3

kain88-de:

The urllib3 package is removed since it is available on conda-forge. All recipes requiring urllib3 are changed to require minimum a version larger then available on the bioconda channel.


bgruening:
Nice! Thanks a bunch!

kain88-de:
Any idea why the linter failed?

bgruening:
@kain88-de if you look at the travis output travis tells you the reson(s).
```
recipes/arvados-python-client  [uses_setuptools, setup_py_install_args]
recipes/dropbox                                 [setup_py_install_args]
recipes/httpretty                               [setup_py_install_args]
```
You can get more informations here: https://bioconda.github.io/linting.html
For example we would like to have `$PYTHON setup.py install --single-version-externally-managed --record=record.txt` in the build.sh instead of `setup.py install`.
The use_setuptools looks like there is setuptools as a runtime environment, which is most likely also false. This is not your fault, but we improve our build infrastructure to prevent errors and these recipes are not following our latest recommendations.

kain88-de:
Ah I understand that is easy to fix. Updating your recipes to follow the same standard is a good idea anyway.

bgruening:
Thanks Max!

kain88-de:
@bgruening I don't see how I can fix the last error. The package depends on setuptools in the setup.py

bgruening:
@kain88-de thats fine! Thanks a lot!

PR #5452, opened 2017-08-06
remove dropbox package

kain88-de:

It's provided in a newer version by conda-forge


recipes/snakemake/meta.yaml
```diff
@@ -30,7 +30,7 @@ requirements:
     - docutils
     - pyyaml
     - requests >=2.8.1
-    - dropbox >=5.2
```
bgruening: Keep in mind this will not create a new snakemake package as you did not increased the build-number.
I think this is ok, as it will not break anything and the next update of snakemake will get the new package dependency, just wanted to make you aware.

kain88-de: Good point. I also forgot to do that in the other PR. I can add those changes here  or in a separate PR if you like

bgruening: I'm not even sure this is needed, just wanted to point this out, that you are not surprised to not see a new package. Maybe you can add a comment to those changes so that the maintainer knows about this later on.

bgruening:
Thanks @kain88-de!

Issue #5374, opened 2017-07-28
Streamlined bioconda recipe development


daler:
@bioconda/all we have some new changes that should hopefully make it easier to build and test recipes locally.

## Summary

`simulate-travis.py` will now install a completely isolated conda environment exclusively for its own use. 

To start using it, first perform the setup (putting in whatever path you'd like to install the isolated version in):

```bash
./simulate-travis.py --bootstrap /tmp/miniconda 
```

That's it. You can then use `simulate-travis.py` as before. You can even remove `bioconda-utils` and `conda-build` from your root environment. Some folks use `bioconda-utils` directly, in which case you would specify the full path `/tmp/miniconda/bin/bioconda-utils`.

There are some additional features as well, see below.

## Rationale

Previously, contributors were forced to change their root conda environment to run tests locally. This is because `bioconda-utils` requires `conda-build`, and `conda-build` must be installed in the root environment. This was frustrating if the root environment was used for something other than bioconda development, and was brought up in #5134, bioconda/bioconda-utils#145 (comment), and #5355.

## Changes

- Simplified the steps to get started on recipes. All you need to start developing bioconda recipes is Python (2 or 3) and the PyYAML package. Optionally, docker. Everything else is handled automatically. This also simplifies setup on travis-ci.
- `~/.config/bioconda/conf.yml` stores the location of the bootstrapped installation so you don't have to specify each time
- `mulled-build` will now run by default, but there is an argument to disable it
- `--git-range` argument to `simulate-travis.py` specifies a git range to be used for both linting and building
- `--skip-linting` argument for . . . skipping linting
- Documentation in bioconda-utils has been updated to reflect these changes (specifically https://bioconda.github.io/contribute-a-recipe.html)

sebastian-luna-valero:
Many thanks @daler !

HenningTimm:
Thanks for the work @daler this makes my life a whole lot easier. This solves #5134 (and I will close it).

The installation and creation of the environment seem to work nicely. However I can not test if everything works due to another problem with docker (most likely related to #5027). I will create an Issue for this.

jerowe:
@daler you just do the coolest things. ;-)

peterjc:
Is this source warning expected under macOS?

```
$ python3 simulate-travis.py --bootstrap /tmp/miniconda 
Please run using "bash" or "sh", but not "." or "source"
[...]

PR #5458, opened 2017-08-07
save miniconda temp file with .sh ext to avoid "bash" or "sh" warnings

daler:

Avoids causing the warning `Please run using "bash" or "sh", but not "." or "source"` when using `simulate-travis.py --bootstrap` (xref https://github.com/bioconda/bioconda-recipes/issues/5374#issuecomment-320710105)

### DIFF
--- a/simulate-travis.py
+++ b/simulate-travis.py
@@ -11,14 +11,15 @@ except ImportError:
 import subprocess as sp
 import shlex
 import argparse
+import tempfile
 
 if sys.version_info.major == 3:
     PY3 = True
-    from urllib.request import urlretrieve, urlcleanup
+    from urllib.request import urlretrieve
 
 else:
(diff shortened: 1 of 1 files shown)


simulate-travis.py
```diff
@@ -11,14 +11,15 @@
 import subprocess as sp
 import shlex
 import argparse
+import tempfile
 
 if sys.version_info.major == 3:
     PY3 = True
-    from urllib.request import urlretrieve, urlcleanup
+    from urllib.request import urlretrieve
```
bgruening: if this gets to messy, we should use `six` here ...

daler: Definitely. For now though I think it's simple enough to warrant not having six as an additional dep.

bgruening:
Thanks!

PR #5459, opened 2017-08-07
Add phyloFlash

epruesse:

### DIFF
--- a/recipes/phyloflash/meta.yaml
+++ b/recipes/phyloflash/meta.yaml
@@ -21,6 +21,7 @@ requirements:
   run:
     - perl >=5.13.2
     - bbmap
+    - emirge
     - pigz
     - samtools
     - vsearch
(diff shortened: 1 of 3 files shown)


bgruening:
The tests seems to fail here.

recipes/phyloflash/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+  patches:
+    - real_bin.patch
+
+build:
+  number: 0
+
+requirements:
+  build: []
+  run:
+    - perl >=5.13.2
```
bgruening: You want to have here `r-base` as well.

recipes/phyloflash/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - r-gtable
+    - r-reshape2
+    - r-ggplot2
+    - r-optparse
+
+test:
+  commands:
+    - (phyloFlash.pl -check_env 2>&1; true) | grep "Please provide a path using -dbhome"
+    - phyloFlash_makedb.pl --help
+    - (phyloFlash_heatmap.R --help; true) | grep "Generates a heatmap plot"
```
bgruening: The `true` is usually not needed. The return code that is evaluated is taken from the grep command and not from the first command.

epruesse: ... unless `pipefail` is set. It's not, in this case, so you are right. I was grasping at straws, trying to figure out why the tests are failing on travis.

bgruening:
Thanks!

Issue #5292, opened 2017-07-20
Adding Biopython to conda-forge


souravsingh:
We are looking to add an option for users who like to try biopython to install it using ``conda``. While the binaries for Linux and Mac are updated through the excellent BioConda, the windows binaries for biopython aren't updated to the latest version.

For this reason, I am thinking of adding the biopython recipe to ``conda-forge`` so that we can maintain parity over the various systems. I would like to know if the folks at bioconda are alright with idea of having the biopython recipe at ``conda-forge``.

xref https://github.com/biopython/biopython.github.io/issues/118

cc @bioconda/core

tomkinsc:
If it makes it easier for people to find/use biopython, I'm all for it. We'll be depending on conda-forge for the foreseeable future, so it will remain available to bioconda users.

With the channel precedence [suggested](https://bioconda.github.io/#set-up-channels)  in the bioconda docs, bioconda will be checked first before conda-forge, so it sounds like Windows users will get the package from conda-forge, and Mac/Linux users will source from bioconda, unless it is removed from bioconda and available only on conda-forge.

bgruening:
What ever is the easiest for the package maintainer :)

ping @peterjc

peterjc:
I'm familiar with how BioConda updates work (and how to update them, e.g. #5175), and conda-forge looks somewhat similar (although one repository per package), but I don't now how the default channel is updated.

I take it there are currently no plans for bioconda to expand to cover Windows (using AppVeyor)? I assume not, but that would be great.

Biopython is already on the default channel (Mac, Linux, Windows) but is already two releases out of date https://anaconda.org/anaconda/biopython - if there's an easy way to update that it might be the best option?

Otherwise, given we'd like to have Windows conda packages, @souravsingh seems right - conda-forge looks like a good idea. So +1 on trying this.

souravsingh:
To update the binaries for biopython in the default channel, we would have to ask the uploader of the recipe to do so, which can get quite slow.

If there is no issue, I can quickly create a patch for the biopython recipe.

peterjc:
Biopython is now in conda-forge, which has the advantage of providing Windows binaries:

https://github.com/conda-forge/biopython-feedstock
https://anaconda.org/conda-forge/biopython

Should we now deprecate the BioConda Biopython recipe, or continues to update both?

johanneskoester:
[...]

PR #5465, opened 2017-08-08
Test importing compiled bits of Biopython

peterjc:

Spotted while looking at https://github.com/conda-forge/biopython-feedstock/pull/1 for the conda-forge recipe.

Note as per #5292 we may be deprecating (removing) Biopython from BioConda in favour of conda-forge only.

### DIFF
--- a/recipes/biopython/meta.yaml
+++ b/recipes/biopython/meta.yaml
@@ -18,7 +18,7 @@ about:
   summary: 'Freely available tools for computational molecular biology.'
 
 build:
-  number: 0
+  number: 1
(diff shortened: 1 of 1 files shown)


bgruening:
@peterjc biopython should stay in bioconda?

peterjc:
I'm waiting for a decision on #5292 about if we should remove Biopython from bioconda now that it's in conda-forge - but until then let's keep it up to date?

bgruening:
Fine with me. Just for the record I'm +1 removing it from Bioconda.

PR #5482, opened 2017-08-09
haystack_bio

rfarouni:
@bioconda/core

### DIFF
--- a/recipes/haystack_bio/build.sh
+++ b/recipes/haystack_bio/build.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-$PYTHON setup.py install
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
 
--- a/recipes/haystack_bio/build.sh
+++ b/recipes/haystack_bio/build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+$PYTHON setup.py install
+
--- a/recipes/haystack_bio/build.sh
+++ b/recipes/haystack_bio/build.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-$PYTHON setup_bioconda.py install --single-version-externally-managed --record=record.txt
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
 
--- a/recipes/haystack_bio/build.sh
+++ b/recipes/haystack_bio/build.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-$PYTHON setup.py install --single-version-externally-managed --record=record.txt
+$PYTHON setup_bioconda.py install --single-version-externally-managed --record=record.txt
 
--- a/recipes/haystack_bio/meta.yaml
+++ b/recipes/haystack_bio/meta.yaml
@@ -26,6 +26,20 @@ about:
 requirements:
   build:
     - python
+    - openjdk
+    - meme
+    - bedtools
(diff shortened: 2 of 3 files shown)


recipes/haystack_bio/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+$PYTHON setup.py install
```
bgruening: Can you please use:
`$PYTHON setup.py install --single-version-externally-managed --record=record.txt`

recipes/haystack_bio/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - ucsc-bigwigaverageoverbed
+    - ucsc-bedgraphtobigwig
+    - ghostscript ==9.10
+    - bx-python 
+    - numpy 
+    - scipy 
+    - matplotlib 
+    - argparse 
+    - jinja2 
+    - nose
```
bgruening: is nose really needed during runtime?

recipes/haystack_bio/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - bedtools
+    - sambamba
+    - ucsc-bigwigaverageoverbed
+    - ucsc-bedgraphtobigwig
+    - ghostscript ==9.10
+    - bx-python 
+    - numpy 
+    - scipy 
+    - matplotlib 
+    - argparse
```
bgruening: argparse should be in the stdlib

recipes/haystack_bio/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    summary: Epigenetic Variability and Transcription Factor Motifs Analysis Pipeline
+
+requirements:
+
+  build:
+    - python
+    - bedtools
+    - openjdk
+    - meme
+    - sambamba
```
bgruening: Are sambamba and openjdk ... needed during installation?

rfarouni: @bgruening I have updated the recipe. Thanks for catching the redundancies!

rfarouni:
@bgruening I changed the test command to run a bit faster.

bgruening:
What is setup_bioconda doing? You old recipe was fine, I was just trying to get rid of a few more build-dependencies ... sorry for that.

rfarouni:
@bgruening no problem! Thanks for your help. I  needed to have the option for users to install the package manually using pip, so I created another setup.py file. The problem is that I cannot include the required packages in the setup.py file. When I add the following in setup I get an error.
```
        install_requires=[
            'numpy>=1.8.2',
            'pandas>=0.13.1',
            'matplotlib>=1.3.1',
            'scipy>=0.13.3',
            'jinja2>=2.7.3',
            'bx-python>=0.7.3',
             'pybedtools>=0.7.10']

```

bgruening:
Sorry, I think I unnecessary confused you. If you have build dependencies, like numpy and pandas you need to include them. If you only have run-time dependencies our recipe as it is it totally fine!

rfarouni:
@bgruening I see.  But how to find out which packages are build dependencies? If they have Cython code?

rfarouni:
@bgruening I have added back the packages to the build dependencies section so I can use the setup.py for manual install as well.

bgruening:
Thanks a bunch and sorry for the confusion!

Issue #5483, opened 2017-08-09
data-heavy bioconductor packages


daler:
Now that we have more space provided by continuum, I was wondering if it would be appropriate to add bioconductor data packages. For example I'd like to add `bioconductor-chipqc` but it depends on these data packages:

  - `bioconductor-txdb.celegans.ucsc.ce6.ensgene`
  - `bioconductor-txdb.hsapiens.ucsc.hg18.knowngene`
  - `bioconductor-txdb.hsapiens.ucsc.hg38.knowngene`
  - `bioconductor-txdb.mmusculus.ucsc.mm10.knowngene`
  - `bioconductor-txdb.mmusculus.ucsc.mm9.knowngene`
  - `bioconductor-txdb.rnorvegicus.ucsc.rn4.ensgene`

As an example, [`TxDb.Hsapiens.UCSC.hg38.knownGene`](https://bioconductor.org/packages/release/data/annotation/html/TxDb.Hsapiens.UCSC.hg38.knownGene.html) is 31MB, so these would be roughly 200MB of space on the anaconda servers.

If it doesn't make sense to store those packages there, is there an alternative solution (other than just not having a package for `bioconductor-chipqc`)?

bgruening:
We could create a dummy package that runs a normal R install in the post-install hook maybe?

daler:
Hmm, that's a good idea. I'll see if I can get that to work.

daler:
@bgruening @johanneskoester I have something working in #5485, but it's not optimal.

Quoting from https://conda.io/docs/building/build-scripts.html:

> We strongly recommend that post-link (and pre-unlink) scripts should:
> 
>    - be avoided whenever possible,
>    - not touch anything other than the files being installed,
>    - not write anything to stdout (or stderr), unless an error occurs,
>    - not depend on any installed (or to be installed) conda packages, and
>    - only depend on simple system tools such as rm, cp, mv, ln, and so on.
> 

Installing a bioconductor data package via a post-link script would require `wget` and `R`, so we wouldn't be following the last two recommendations.

Other issues I've discovered with this approach:

- The moment the URL changes, the package is broken. Maybe if we could pre-cache on Cargo Port we could use that URL in the post-link. A possible way to catch this kind of issue would be to set up a cron job to install these packages every night.
- Errors in post-link.sh are hard to debug, so that might open up some support headaches if folks can't figure out why the install is mysteriously failing.
[...]

PR #5485, opened 2017-08-10
[WIP] post-link test for bioconductor data package

daler:

xref #5483

### DIFF
--- a/recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/meta.yaml
+++ b/recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/meta.yaml
@@ -13,6 +13,7 @@ source:
     - https://depot.galaxyproject.org/software/TxDb.Hsapiens.UCSC.hg38.knownGene/TxDb.Hsapiens.UCSC.hg38.knownGene_3.4.0_src_all.tar.gz
   md5: 1d5e07631ea58e96b11905d39e76ca6e
 build:
+  noarch: generic
   number: 0
   rpaths:
     - lib/R/lib/
--- a/recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/meta.yaml
+++ b/recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/meta.yaml
@@ -13,7 +13,7 @@ source:
     - https://depot.galaxyproject.org/software/TxDb.Hsapiens.UCSC.hg38.knownGene/TxDb.Hsapiens.UCSC.hg38.knownGene_3.4.0_src_all.tar.gz
   md5: 1d5e07631ea58e96b11905d39e76ca6e
 build:
-  noarch: generic
+  # noarch: generic
   number: 0
   rpaths:
     - lib/R/lib/
--- a/recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/meta.yaml
+++ b/recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/meta.yaml
@@ -18,10 +18,6 @@ build:
     - lib/R/lib/
     - lib/
 requirements:
-  build:
-    - bioconductor-annotationdbi
-    - 'bioconductor-genomicfeatures >=1.25.18'
-    - r-base
   run:
     # wget is needed in post-link.sh
     - wget
--- a/recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/meta.yaml
+++ b/recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/meta.yaml
@@ -1,0 +1,33 @@
+package:
+  name: bioconductor-txdb.hsapiens.ucsc.hg38.knowngene
+  version: 3.4.0
+source:
+  fn: TxDb.Hsapiens.UCSC.hg38.knownGene_3.4.0.tar.gz
+  url:
+    - http://bioconductor.org/packages/3.5/data/annotation/src/contrib/TxDb.Hsapiens.UCSC.hg38.knownGene_3.4.0.tar.gz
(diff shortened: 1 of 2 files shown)


daler:
I haven't added a hash here, but presumably an expected hash could be added to the post-link.sh and checked against the hash of the downloaded tarball.

recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+build:
+  number: 0
+  rpaths:
+    - lib/R/lib/
+    - lib/
+requirements:
+  build:
+    - bioconductor-annotationdbi
+    - 'bioconductor-genomicfeatures >=1.25.18'
+    - r-base
```
johanneskoester: If this package only installs a post-link script, why does it have build dependencies?

recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+
+# Keeping the sources in here should trigger any warnings about missing URLs or
+# hash mismatches at build time, even though nothing is actually done with the
+# downloaded package.
+source:
+  fn: TxDb.Hsapiens.UCSC.hg38.knownGene_3.4.0.tar.gz
+  url:
+    - http://bioconductor.org/packages/3.5/data/annotation/src/contrib/TxDb.Hsapiens.UCSC.hg38.knownGene_3.4.0.tar.gz
+    - https://depot.galaxyproject.org/software/TxDb.Hsapiens.UCSC.hg38.knownGene/TxDb.Hsapiens.UCSC.hg38.knownGene_3.4.0_src_all.tar.gz
```
johanneskoester: I guess these URLs are not needed now that there is a post-link script?

daler: Technically no, but keeping them in here checks to make sure they are available and checks the hash at build-time. This is at the expense of extra download time/space though. I could go either way on this. I wish there were a good way of pulling out the meta.yaml URLs in the post-link script, then the URLs and hash could be maintained, as normal, in the meta.yaml.

recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/meta.yaml
daler: Good point, I missed that.

recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/meta.yaml
johanneskoester: I see. Your decision.

bgruening: We need it here, otherwise the entire idea does not work as we do not mirror these packages on the Galaxy depot.

recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/post-link.sh
```diff
@@ -0,0 +1,35 @@
...
+URLS=(
+    "https://depot.galaxyproject.org/software/TxDb.Hsapiens.UCSC.hg38.knownGene/TxDb.Hsapiens.UCSC.hg38.knownGene_3.4.0_src_all.tar.gz"
+    "http://bioconductor.org/packages/3.5/data/annotation/src/contrib/TxDb.Hsapiens.UCSC.hg38.knownGene_3.4.0.tar.gz"
+    )
+MD5="1d5e07631ea58e96b11905d39e76ca6e"
+
+# Use a staging area in the conda dir rather than temp dirs, both to avoid
+# permission issues as well as to have things downloaded in a predictable
+# manner.
+STAGING=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG-BUILDNUM
```
bgruening: We could offer an ENV variable that people can use to overwrite this location. If available store there ... otherwise use this one here. E.g. $BIOCONDA_DATA_STAGING

recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/meta.yaml
daler: Ah, right. That solves it then.

recipes/bioconductor-txdb.hsapiens.ucsc.hg38.knowngene/post-link.sh
daler: The staging location is just a temp place to download the tarball that we're guaranteed to have access to, and it's deleted at the end. Pretty sure the data are installed into the R libs dir which has to be in the conda dir anyway. Do you think it's worth it to have a env var for the tempfile location?

daler:
I forgot about this but `noarch: generic` gets skipped on osx. The build before was failing on osx, likely because of using `md5sum` instead of `md5`. So the noarch version, despite passing on travis-ci, will actually fail on osx.

Solutions: 
- remove the noarch designation so it is tested on osx. Since the uploaded conda package is minimal in size, this shouldn't have an impact on quota.
- don't do the md5 check in the post-link script
- change bioconda-utils so that noarch packages are built and tested -- but not uploaded -- when running on osx

bgruening:
@daler you are awesome! https://github.com/bioconda/bioconda-recipes/commit/b857df334838544e237508e3ab0b36ed2253e2eb look great. Is this still WIP. I like it!

daler:
It's still a WIP in the sense that I want to formalize it into bioconda-utils so it can be used for other packages, but as a single standalone package I think its good to go.

PR #5487, opened 2017-08-10
add new version for rgi

raphenya:

### DIFF
--- a/recipes/rgi/3.2.1/build.sh
+++ b/recipes/rgi/3.2.1/build.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+if [ "$(uname)" == "Darwin" ]; then
+    echo "Platform: Mac"
+
+elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+    echo "Platform: Linux"
+fi
+
+# install setup
+$PYTHON setup.py install
--- a/recipes/rgi/3.2.1/build.sh
+++ b/recipes/rgi/3.2.1/build.sh
@@ -1,11 +1,3 @@
 #!/bin/bash
 
-if [ "$(uname)" == "Darwin" ]; then
-    echo "Platform: Mac"
-
-elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
-    echo "Platform: Linux"
-fi
-
-# install setup
 $PYTHON setup.py install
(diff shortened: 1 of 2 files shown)


daler:
@raphenya what's the difference between this git repo and https://card.mcmaster.ca/download? Is it possible to use the latter? We archive tarballs in Galaxy's Cargo Port for long-term reproducibility, and using git urls prevents this.

recipes/rgi/3.2.1/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if [ "$(uname)" == "Darwin" ]; then
```
daler: These echos shouldn't be needed for conda recipes. You should be fine with just the `$PYTHON setup.py install`.

recipes/rgi/3.2.1/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+source:
+  git_url: {{ source_url }}
+  git_rev: {{ commit }}
+
+build:
+  skip: True # [py3k]
+  number: 1
+
+# We pin on gcc version to avoid gcc-5 dual ABI issues depending on boost.
```
daler: this comment describes pinning gcc, but I don't see any pinning below, in which case you can remove this comment (if it doesn't really need to be pinned)

recipes/rgi/3.2.1/build.sh
raphenya: @daler thanks! would remove the echos.

recipes/rgi/3.2.1/meta.yaml
raphenya: @daler Thanks. Would remove comment.

raphenya:
@daler there a way to use git urls,  since it's easier when importing other module to your project?

daler:
@raphenya can you help explain the git repo you're using in this recipe? To me, it looks like the repo is a conda-specific wrapper (at least just based on the name) that wraps rgi proper. It does this by creating a git submodule that pulls in the rgi repo, and essentially installs the rgi scripts as part of a new rgi_conda_dev Python package.

Is that correct? If so, wouldn't it be easier to use the rgi repo directly, copying Python scripts to `$PREFIX/bin`?

Looks like the Python scripts don't have shebang lines, so you would have to add them in the conda recipe. Using the rgi release tarball from https://github.com/arpcard/rgi/archive/3.2.1.tar.gz, in build.sh you could do something like this:

```bash
#!/bin/bash
for fn in *py; do
  sed -i "1s|^|#!/usr/bin/env python\n|" $fn
  chmod +x $fn
  cp $fn $PREFIX/bin
done
```

Would this work for this recipe? It avoids coordinating a wrapper git repo plus submodule and uses a release tarball directly, resulting in easier maintenance in the long term.

raphenya:
@daler yes the repo is a wrapper for rgi.

PR #5492, opened 2017-08-11
Add snmf

avengerpb:
* [v] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [v] This PR adds a new recipe.


recipes/snmf/build.sh
```diff
@@ -0,0 +1,15 @@
...
+sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $snmf/Snmf.pl
+rm -f $snmf/Snmf.bak
+chmod +x $snmf/snmf.sh
+chmod +x $snmf/plink
+chmod +x $snmf/install.command
+make $snmf/install.command
+chmod 777 $snmf/Snmf.pl
+mkdir -p $PREFIX/bin
+mkdir -p $PREFIX/opt/var
+export PATH=$snmf:$PATH >> ./bash_profile
```
daler: @avengerpb `bash_profile` will never get read on the host system; in conda recipes I don't think PATH manipulations performed in build.sh will ever have an effect in the package.

Simply copying executables over to `$PREFIX/bin` here should do the trick, at least for simple packages.

recipes/snmf/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+  version: 1.0.0
+
+source:
+   fn: 1.5.tar.gz
+   url: https://github.com/avengerpb/snmf_conda/archive/1.5.tar.gz
+   md5: f56a662812203d0046ae435ae3eb0494
+
+test:
+ command:
+  - /snmf.sh test.vcf out logs k_out k_log 2 5 ab
```
daler: Once `snmf.sh` is in `$PREFIX/bin`, you won't need the leading `/` here.

avengerpb:
I have changed the recipe follow your comment

recipes/snmf/build.sh
```diff
@@ -0,0 +1,14 @@
...
+snmf=$PREFIX/opt/$PKG_NAME-$PKG_VERSION
+mkdir -p $snmf
+mkdir -p $PREFIX/bin
+cp -r ./* $PREFIX/bin
+sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $PREFIX/bin/Snmf.pl
+rm -f $PREFIX/bin/Snmf.bak
+chmod +x $PREFIX/bin/snmf.sh
+chmod +x $PREFIX/bin/plink
+chmod +x $PREFIX/bin/install.command
+make $PREFIX/bin/install.command
```
daler: As you have it now, the installation process happens inside `$PREFIX/bin`. Instead, it should happen in the build dir (the cwd of this script) and then executables should be copied over from there.

recipes/snmf/build.sh
```diff
@@ -0,0 +1,12 @@
...
+
+snmf=$PREFIX/opt/snmf/
+mkdir -p $snmf
+mkdir -p $PREFIX/bin/
+cp -rf ./* $snmf/
+sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $snmf/Snmf.pl
+rm -f $snmf/Snmf.bak
+chmod +x $snmf/snmf.sh
+chmod +x $snmf/plink
+chmod 777 $snmf/Snmf.pl
```
bgruening: why 777?

recipes/snmf/build.sh
```diff
@@ -0,0 +1,12 @@
...
+cp $snmf/snmf.sh $PREFIX/bin/
```
bgruening: Could this be a symlink?

recipes/snmf/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+build:
+  detect_binary_files_with_prefix: true
+  number: 0
+  skip: False
+
+requirements:
+  build:
+    - perl-bioperl
+    - libgcc
```
bgruening: is libgcc really needed here?

recipes/snmf/meta.yaml
```diff
@@ -0,0 +1,28 @@
+about:
+  home: https://github.com/avengerpb/snmf_conda
+  license: Artistic-2.0
+  summary: Estimating individual admixture coefficients based on sparse non-negative matrix factorization and population genetics
+
+build:
+  detect_binary_files_with_prefix: true
```
bgruening: I think this is activated by default so not need here.

avengerpb:
I have a problem with path of files. The path in the code is relative path and it will be wrong if I only the start file are places or symlink in the /bin then it can't call to other file. Do you have any suggest to resolve it.

avengerpb:
I have a problem with the path since in the code the path is relative. When calling start file ( snmf.sh ) at another dir the calling path is wrong. Do you have any suggestion for this problem.

btw, how to fix permission denied when mkdir in envs/opt.

recipes/snmf/build.sh
```diff
@@ -0,0 +1,12 @@
+ #!/bin/bash
+
+snmf=$PREFIX/opt/snmf
+mkdir -p $snmf
+chmod 777 $snmf
```
daler: do you need the 777 here?

recipes/snmf/meta.yaml
```diff
@@ -0,0 +1,26 @@
...
+  version: 1.0.0
+
+source:
+   fn: 1.5.tar.gz
+   url: https://github.com/avengerpb/snmf_conda/archive/1.5.tar.gz
+   md5: 951beb9a826c0c02fd6c08571b4365d2
+
+test:
+ command:
+  - run.sh test.vcf out logs k_out k_log 2 5 ab
```
daler: You might want to check for output here. I don't know exactly how snmf works, but it looks to me like the build script creates `run.sh` to have hard-coded arguments passed to `snmf.sh`.

That is, no matter what arguments you give to `run.sh`, it will always call `"snmf.sh vcf output logs best_k_output best_k_logfile kmin kmax groups threshold_group"`

recipes/snmf/meta.yaml
```diff
@@ -0,0 +1,26 @@
...
+    - perl-bioperl
+    - perl-lib
+
+package:
+  name: snmf
+  version: 1.0.0
+
+source:
+   fn: 1.5.tar.gz
+   url: https://github.com/avengerpb/snmf_conda/archive/1.5.tar.gz
```
daler: Sorry I'm just seeing this now, but how does the repo referred to here differ from http://membres-timc.imag.fr/Olivier.Francois/snmf/software.htm? Did you create a fork of the original sNMF? If so the recipe name should be changed to differentiate it from the original.

avengerpb: Yes this is the same with original but I also added library and changed path file for suitable conda style.

recipes/snmf/build.sh
avengerpb: The script will create new folder to put all the output in at that dir

recipes/snmf/meta.yaml
avengerpb: This is just something I experiment. I having troubles with the relative path inside the script since everything should be in one directory but I suggest me to put the snmf.sh in the bin so the relative path to call other files won't work right.  Do you have any idea to fix that

recipes/snmf/meta.yaml
daler: Can you clarify what you mean by 'added library'? Any file path changes should happen within the conda recipe, no need to maintain a separate repo. We want to work with the original source as much as possible.

avengerpb: I put 2 Perl lib ( Switch.pm and Bio/Seq.pm ) in the my repo. The file use many relative path and when building conda recipe I had to change some. Furthermore the link above only have the bin code while the version which I used for Galaxy has snmf.sh and Snmf.pl to control the bin code after install.

daler: A cleaner way to do this would be to not use your separate repo and instead have `snmf.sh` and `Snmf.pl` as extra files included in this recipe that are copied over to `$PREFIX/bin` (e.g., in `build.sh` you would have `cp $RECIPE_DIR/snmf.sh $PREFIX/bin/snmf.sh`).

I also just noticed you are copying a pre-compiled `plink` over, which is unlikely to pass tests on  travis-ci. So your recipe would need to include run dependencies of `plink` (or `plink2`), `perl-bioperl`, and `perl-switch`.

It does not look like we have a recipe for `perl-switch`, so you can create one and include it in this PR.

avengerpb: Ok, I will try in that way. However, when excecuting the symlink, the relative path inside will base on current dir where you call the script not the dir that original file is placed. Do you have any suggestion for that ?

daler: Yep, if you check out some of the java packages, the wrapper script has to handle that kind of use case, e.g. https://github.com/bioconda/bioconda-recipes/blob/master/recipes/gatk/3.5/gatk.sh.
But I'm not sure we need to symlink if you do it that way I suggested -- everything can go in the bin dir. @bgruening had suggested it in https://github.com/bioconda/bioconda-recipes/pull/5492#discussion_r133381946, but at that time I don't think it was clear what the recipe was doing.

avengerpb: In the new version, the make file false when compiling so can I pre run the compliment then put those files in the dir later ?

avengerpb:
I think the first version of Snmf should be like this and I will improve for cleaner and more efficiently in the next version. By the way, can you check for me the vcf2hapmap tool that I push before ?

recipes/snmf/build.sh
avengerpb: Okay, I will fix that. Can you check for me the vcf2hapmap recipe ?

avengerpb:
Thank you very much. I would very appreciate that since I am new to Conda

daler:
Looks like that one has some of the same issues as this one: the recipe builds a conda-specific wrapper for a tool rather than for the tool itself. In that one you should write a recipe for vcf2hapmap directly. There will likely need to be changes upstream (for example, the path is hard-coded at https://github.com/avengerpb/vcf2hapmap/blob/master/vcf2FastaAndHapmap.sh#L33) but luckily you're the author so it should be straightforward :)

johanneskoester:

PR #5507, opened 2017-08-12
update sequana to version 0.5.0

cokelaer:


daler:
@cokelaer looks like you're running out of space on travis-ci, https://travis-ci.org/bioconda/bioconda-recipes/jobs/264317277#L1038. @bgruening I think we saw a similar issue cgat-scripts, was there a solution for that?

cokelaer:
@bgruening hey thanks for spotting that. I always have trouble building this recipe and I have changed lots of things but did not realised this was due to running out of space... but now I can see that. I am always surprised to see the mac building fine but the linux version is always causing trouble... while locally it works fine. I have just trigger another build will look at the cgat-scripts before more triggers.

bgruening:
@cokelaer @daler the problem is really that the space is limited on travis and large meta-package might exceed this. I was digging around and we can save a few MB even maybe 1GB but we will always have a space limit.
It might be a good solution to promote the anaconda environments as a solution for this particular problem.

@cokelaer are you ok with this package? Ready to merge?

cokelaer:
@bgruening @daler 
Hi, coming back to you wrt this package. I've updated the version to 0.5.0 and added a couple of
packages (apdes, prokka). Again, the MAC build succeeds but the linux one does not. locally this work. If I understand correctly, the build of the package takes too much space. Is this becuase we rely on too many packages? You said "It might be a good solution to promote the anaconda environments as a solution for this particular problem." what do you mean ? 
thanks in advance for your feedbacks

bgruening:
@cokelaer please see the following links:
https://conda.io/docs/user-guide/tasks/manage-environments.html
https://docs.continuum.io/ae-notebooks/user/anaconda
https://docs.continuum.io/anaconda-cloud/user-guide/tasks/work-with-environments

I also reduced our storage footprint a little bit, maybe rebaseing against master helps a little bit?

cokelaer:
Finally, managed to pass the checks. @bgruening FYI this was prokka package that was most probably the culprit with 2Gb of dependencies. I will add back other small deps in the future but for now that would do it.

PR #5513, opened 2017-08-14
Update meta.yaml for v1.1.10

jdidion:


daler:
@jdidion not sure what's going on here; something's not being passed correctly to the mulled-build test container. Can you maybe try double quotes in the test?

daler:
Thanks. I had merged #5515 before checking back here, sorry about that.

I merged this anyway since they're both the same change; the net effect will be the package being built from #5515 and then the build for this PR should be a no-op since the package will already exist by the time it runs.

PR #5516, opened 2017-08-14
added recipe for SHIVER - Sequences from HIV Easily Reconstructed

notestaff:


bgruening:
@notestaff `pyfastaq` is not build for py27. If this would be possible you could also use this package for py27.

PR #5521, opened 2017-08-15
Bump plant_tribes_assembly_post_prcessor to 1.0.2

gregvonkuster:

### DIFF
--- a/recipes/plant_tribes_assembly_post_processor/1.0.2/meta.yaml
+++ b/recipes/plant_tribes_assembly_post_processor/1.0.2/meta.yaml
@@ -1,0 +1,33 @@
+package:
+  name: plant_tribes_assembly_post_processor
+  version: "1.0.2"
+
+source:
+  fn: v1.0.2.tar.gz
+  url: https://github.com/dePamphilis/PlantTribes/archive/v1.0.2.tar.gz
+  md5: c0a4217048fe8a18e0c13d67667da638
+
+build:
+  number: 0
+  # Requires genometools which is not supported on osx
+  skip: True # [osx]
+
+requirements:
+  run:
+    - cap3
+    - genometools-genometools >=1,<2
+    - hmmer >=3
+    - mafft >=7,<8
+    - perl >=5.22
+    - perl-estscan2 2.1 pl5.22.0_1
+    - transdecoder >=3,<4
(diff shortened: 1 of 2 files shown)


recipes/plant_tribes_assembly_post_processor/1.0.2/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  skip: True # [osx]
+
+requirements:
+  run:
+    - cap3
+    - genometools-genometools >=1,<2
+    - hmmer >=3
+    - mafft >=7,<8
+    - perl >=5.22
+    - perl-estscan2 2.1 pl5.22.0_1
```
bgruening: I don't think this strict pinning is good here. We might need to rebuild all perl packages against `5.22.2.1`

gregvonkuster: I agree, its not ideal, but the number of dependencies here results in a pretty complex chain since many involve Perl and some are much more dated than others.  I spent a bunch of time trying to get the combination to work and pinning in this way is what I found to work in the current environment.  I will take look at it when the Perl packages are rebuilt and make sure everything is kosher at that time as well.

gregvonkuster: Can this be merged?

gregvonkuster: @bgruening version 1.0.1 of this recipe uses the exact same strict pinning as this recipe: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/plant_tribes_assembly_post_processor/1.0.1/meta.yaml, so I'm wondering why this version of the recipe cannot be merged as well.  The perl-estscan2 code has not been touched since 2013 (https://sourceforge.net/projects/estscan/), and its age makes it difficult to get it to work with some versions of perl.  I'm not sure if I could make the pinning less strict here without a lot of effort.  Is there any way that this recipe can be merged as is?  If not, I'll likely continue to use version 1.0.1 of it until I have more time to spend on this.

bgruening:
Sorry, forgot about this one.

gregvonkuster:
Thanks @bgruening ;)

PR #5529, opened 2017-08-15
Add ARB 6.1.rev16233

epruesse:

The stable release series for ARB (6.0.x) was forked off in 2014 and has since only received bugfix patches. No other releases were made upstream since then, although development is ongoing. However, ARB is in production through a CI system right off of the SVN and SVN trunk is reasonably stable at most times.

This PR creates a package with the "official" version (as set by build system itself) `6.1.rev16233`. It is meant to close the gap until upstream eventually publishes the next release.

### DIFF
--- a/recipes/arb-bio/6.1.rev16223/abs_path_fix.patch
+++ b/recipes/arb-bio/6.1.rev16223/abs_path_fix.patch
@@ -741,9 +741,10 @@ Index: util/usecores.pl
 ===================================================================
 --- Makefile	(revision 16228)
 +++ Makefile	(working copy)
-@@ -711,7 +711,7 @@
+@@ -711,5 +711,5 @@
  ifeq ($(DARWIN),1)
-  TIME:=gtime
+- TIME:=gtime
++ TIME:=command time
  else
 - TIME:=/usr/bin/time
 + TIME:=command time
--- a/recipes/arb-bio/6.1.rev16223/abs_path_fix.patch
+++ b/recipes/arb-bio/6.1.rev16223/abs_path_fix.patch
@@ -738,3 +738,15 @@ Index: util/usecores.pl
  
  # determines cores used by jenkins builds
  my $cores = `cat /proc/cpuinfo | grep processor | wc -l`; # no /proc/cpuinfo (e.g. on OSX) -> falls back to minimum
+===================================================================
+--- Makefile	(revision 16228)
++++ Makefile	(working copy)
+@@ -711,7 +711,7 @@
+ ifeq ($(DARWIN),1)
+  TIME:=gtime
+ else
+- TIME:=/usr/bin/time
++ TIME:=command time
+ endif
+
+ #---------------------- SSE vectorizer
--- a/recipes/arb-bio/6.1.rev16223/abs_path_fix.patch
+++ b/recipes/arb-bio/6.1.rev16223/abs_path_fix.patch
@@ -1,0 +1,740 @@
+Index: BINDINGS/PERL/Makefile
+===================================================================
+--- BINDINGS/PERL/Makefile	(revision 16223)
++++ BINDINGS/PERL/Makefile	(working copy)
+@@ -123,8 +123,8 @@
+ MAKE_APERL_FILE = Makefile.aperl
+ PERLMAINCC = $(CC)
+ PERL_INC = /usr/lib/perl/5.10/CORE
+-PERL = /usr/bin/perl
+-FULLPERL = /usr/bin/perl
++PERL = /usr/bin/env perl
++FULLPERL = /usr/bin/env perl
+ ABSPERL = $(PERL)
+ PERLRUN = $(PERL)
+ FULLPERLRUN = $(FULLPERL)
+@@ -854,7 +854,7 @@
+ 
+ # --- MakeMaker makeaperl section ---
+ MAP_TARGET    = perl
+-FULLPERL      = /usr/bin/perl
++FULLPERL      = /usr/bin/env perl
+ 
+ $(MAP_TARGET) :: static $(MAKE_APERL_FILE)
+ 	$(MAKE) $(USEMAKEFILE) $(MAKE_APERL_FILE) $@
+Index: CONVERTALN/show_error_list.pl
+===================================================================
+--- CONVERTALN/show_error_list.pl	(revision 16223)
++++ CONVERTALN/show_error_list.pl	(working copy)
+@@ -1,4 +1,4 @@
+-#!/usr/bin/perl
++#!/usr/bin/env perl
+ 
+ use strict;
+ use warnings;
+Index: CORE/arb_diff.cxx
+===================================================================
+--- CORE/arb_diff.cxx	(revision 16223)
++++ CORE/arb_diff.cxx	(working copy)
+@@ -262,7 +262,7 @@
+     if      (!GB_is_regularfile(file1)) error = GBS_global_string("No such file '%s'", file1);
+     else if (!GB_is_regularfile(file2)) error = GBS_global_string("No such file '%s'", file2);
+     else {
+-        char *cmd     = GBS_global_string_copy("/usr/bin/diff --unified %s %s", file1, file2);
++        char *cmd     = GBS_global_string_copy("/usr/bin/env diff --unified %s %s", file1, file2);
+         FILE *diffout = popen(cmd, "r");
+ 
+         if (diffout) {
+Index: GDE/SATIVA/arb_sativa.pl
+===================================================================
+--- GDE/SATIVA/arb_sativa.pl	(revision 16223)
++++ GDE/SATIVA/arb_sativa.pl	(working copy)
+@@ -1,4 +1,4 @@
+-#!/usr/bin/perl
++#!/usr/bin/env perl
+ 
+ #############################################################################
+ #                                                                           #
+Index: GDEHELP/cleanmenu.pl
+===================================================================
+--- GDEHELP/cleanmenu.pl	(revision 16223)
++++ GDEHELP/cleanmenu.pl	(working copy)
+@@ -1,4 +1,4 @@
+-#!/usr/bin/perl
++#!/usr/bin/env perl
+ 
+ use strict;
+ use warnings;
+Index: GDEHELP/pp.pl
+===================================================================
+--- GDEHELP/pp.pl	(revision 16223)
++++ GDEHELP/pp.pl	(working copy)
+@@ -1,4 +1,4 @@
+-#!/usr/bin/perl
++#!/usr/bin/env perl
+ # ============================================================ #
+ #                                                              #
+ #   File      : pp.pl                                          #
+Index: HELP_SOURCE/generate_index.pl
+===================================================================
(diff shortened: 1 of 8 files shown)


epruesse:
@bioconda/core Anyone able to do a review?

recipes/arb-bio/6.1.rev16223/abs_path_fix.patch
```diff
@@ -0,0 +1,753 @@
+Index: BINDINGS/PERL/Makefile
+===================================================================
+--- BINDINGS/PERL/Makefile	(revision 16223)
++++ BINDINGS/PERL/Makefile	(working copy)
```
bgruening: @epruesse can you get this upstream as well?

recipes/arb-bio/6.1.rev16223/build.sh
```diff
@@ -0,0 +1,124 @@
...
+ cd $PREFIX/bin;
+ for exe in $ARB_INST/bin/arb*; do
+     ln -s "$exe"
+ done
+)
+
+# fix the library paths
+case `uname` in
+    Darwin)
+	# fix library IDs
```
bgruening: Meh this is messy. Glad it seems to work.

recipes/arb-bio/6.1.rev16223/build.sh
```diff
@@ -0,0 +1,124 @@
...
+
+mkdir -p "${PREFIX}/etc/conda/activate.d"
+cat >"${PREFIX}/etc/conda/activate.d/arbhome-activate.sh" <<EOF
+export ARBHOME_BACKUP="\$ARBHOME"
+export ARBHOME="\$CONDA_PREFIX/lib/arb"
+EOF
+mkdir -p "${PREFIX}/etc/conda/deactivate.d"
+cat >"${PREFIX}/etc/conda/deactivate.d/arbhome-deactivate.sh" <<EOF
+export ARBHOME="\$ARB_HOMEBACKUP"
+EOF
```
bgruening: Do we have something like `make test` which we can run here as well?

recipes/arb-bio/6.1.rev16223/make.patch
```diff
@@ -0,0 +1,25 @@
...
+ #  LINK_STATIC=1# link static (testing only)
+ endif
+
+@@ -770,6 +770,10 @@
+
+ #---------------------- differences between linking executables and shared libs:
+
++cflags += $(CFLAGS)
++clflags += $(LDFLAGS)
++cxxflags += $(CXXFLAGS)
```
bgruening: This also looks like it should be upstreamed.

recipes/arb-bio/6.1.rev16223/meta.yaml
```diff
@@ -0,0 +1,93 @@
+{% set name = "ARB" %}
+{% set version = "6.1.rev16223" %}
+
+package:
+  name: {{ name|lower }}-bio
+  version: {{ version }}
+
+source:
+  svn_url: http://vc.arb-home.de/readonly/trunk
```
bgruening: Do you think we should back this up in a tarball and cargo-port?

recipes/arb-bio/6.1.rev16223/meta.yaml
```diff
@@ -0,0 +1,93 @@
...
+    - pthread-stubs
+    - xorg-libxp
+    - xorg-libxmu
+    - xorg-libxi
+    - xorg-libxaw
+    - xorg-libxpm
+    - xorg-libxft
+    - xorg-makedepend
+    - xerces-c
+    - libtiff
```
bgruening: We need more pinning here. Can you check this list and adopt your pinning?
https://github.com/conda-forge/conda-forge.github.io/blob/master/scripts/pin_the_slow_way.py#L39

recipes/arb-bio/6.1.rev16223/meta.yaml
```diff
@@ -0,0 +1,93 @@
+{% set name = "ARB" %}
+{% set version = "6.1.rev16223" %}
```
bgruening: Can you double check that this version is correct in the conda way and that the next version will be called 6.2 or 7.0?

```
>>> LooseVersion("6.1.rev16223") < LooseVersion("6.2")
True
>>> LooseVersion("6.1.rev16223") < LooseVersion("6.1")
False

```

recipes/arb-bio/6.1.rev16223/abs_path_fix.patch
epruesse: I probably could, but the hard coded /usr/bin/perl paths are correct for other scenarios. ARB comes with Perl language bindings and uses them from its scripts, so the Perl used by those has to match the one used during build.

In Bioconda, the package requires and pins Perl, so the match is guaranteed to be in the PATH as long as ARB is.

In e.g. Debian however, the package would be built against the system Perl, and the scripts need to use that one even if the user has a different version in their PATH, lest things break.

Technically, the paths should be set by the build machinery. However, ARB is built using plain Makefiles with a bunch of tests and variables interspersed through those catering to Debian and Homebrew, not with cmake or autotols. I don't dare to mess with those too much, in particular not upstream, as it's too easy to break things under other conditions.

recipes/arb-bio/6.1.rev16223/build.sh
epruesse: "Normally", this would be done by libtool. Under Linux, ARB uses LD_LIBRARY_PATH liberally to force it to work. Under OSX, that's not possible since Apple introduced that annoying SIP thing that removes linker settings from the environment. Macports side steps the issue by linking statically. I've tried a lot of approaches to get this to be right at build time, without post-build-tinkering, but this has been the simplest and most robust way to get the paths to be understood by conda's own path-fixing.

recipes/arb-bio/6.1.rev16223/build.sh
epruesse: Not really. ARB uses a home grown unit test framework that expects flags to be set during library build, so that private members etc. can be inspected. To run the tests, I'd have to build ARB twice, once with unit tests enabled, and once in release mode. And that would kick me past the time limits on Travis.

(Yes, it's a "code smell" to inspect private members and I don't think testing code other than the release version is a good idea, but it's the way it was implemented upstream).

In any case, the tests are run through upstream's CI, so re-running them here does not gain too much. What I'd like to have is a way to launch ARB, click a few buttons and see whether that's going all right. But while I could script ARB to run a test, running a headless X server inside of the test docker would be far from trivial (plus, there is the time limit again, ARB already takes rather long to build).

recipes/arb-bio/6.1.rev16223/make.patch
epruesse: Upstream likes to have the build as isolated as possible. I may be able to push a variant of this through, but it won't be fast or easy.

recipes/arb-bio/6.1.rev16223/meta.yaml
epruesse: The next release version is likely to follow the 6.1.0-rc1 and 6.1.0 scheme. That should be OK, as in conda strings are "less than" integers unless the string is "post". 

```
>>> from conda.version import VersionOrder
>>> VersionOrder("6.1.rev16223") < VersionOrder("6.1")
True
>>> VersionOrder("6.1.rev16223") < VersionOrder("6.1.0")
True
```

What I considered was naming it "6.0.5+6.1.rev16223" (after `+` is the local version, which is always less significant than the main version) so that the official release version would still be preferred unless someone requested the build specifically. OTOH, naive users are probably better served with the current version that has 3 years more development in it.

recipes/arb-bio/6.1.rev16223/meta.yaml
epruesse: I had pondered creating a tar ball. The ones at ftp.arb-silva.de/ARB/builds have seized to appear, but I wasn't sure about self-hosting this as it wouldn't create a safer spot.

How do I use cargo port? We could use this Trac URL to pull the revision as zip archive:
http://bugs.arb-home.de/browser/trunk?rev=16228&format=zip

recipes/arb-bio/6.1.rev16223/meta.yaml
epruesse: The only one out of that list I hadn't pinned was boost-cpp. It doesn't matter much, as ARB uses only one header-only lib and the oldest conda-forge version 1.63 is plenty new enough.

Still, pinned it to be conforming.

recipes/arb-bio/6.1.rev16223/abs_path_fix.patch
epruesse: Hmm. In theory I could script a replacement of `#!/usr/bin/perl` to the current absolute perl path. Conda should then pick that up and change it as it does with all the python script paths when it installs the package.

recipes/arb-bio/6.1.rev16223/meta.yaml
bgruening: ? libtiff is not pinned and should be pinned to `libtiff >=4.0.3,<4.0.8`

recipes/arb-bio/6.1.rev16223/meta.yaml
bgruening: Just submit a PR to this repo: https://github.com/galaxyproject/cargo-port

recipes/arb-bio/6.1.rev16223/meta.yaml
bgruening: Ok, this seems to be fine! Thanks for checking.

recipes/arb-bio/6.1.rev16223/build.sh
bgruening: :)

recipes/arb-bio/6.1.rev16223/abs_path_fix.patch
bgruening: Not needed. I tend to agree with your first conclusion but this is not relevant for this PR. Thanks! Other than the pinning everything looks good.

recipes/arb-bio/6.1.rev16223/meta.yaml
epruesse: Darn. Missed that one. It's probably wrong in the 6.0.6 version as well.

recipes/arb-bio/6.1.rev16223/meta.yaml
epruesse: Done. See galaxyproject/cargo-port#129

epruesse: And you were faster fixing the tabs...

epruesse:
@bgruening I messed up a bit on the PR to cargo port. The version I chose had the last commit to it made in revision 16223. The most recent revision, 16228, has some commits to other branches in it. So while there is no difference to the file stored at cargo_port, I'm not stuck with a file name that is named "wrong".

Do you want me to do a PR to cargo-port to "fix" the entry, to add a correct one, or do we just leave it as it is?

bgruening:
I think we can leave it as it is. Thanks!

johanneskoester:

PR #5531, opened 2017-08-16
Barrnap 0.8

epruesse:

Adds recipe for Barrnap 0.8 and fixes the LFS issues in the old recipes. HMM training data and scripts and test cases are excluded to minimize size of package.

### DIFF
--- a/recipes/barrnap/0.8/meta.yaml
+++ b/recipes/barrnap/0.8/meta.yaml
@@ -19,7 +19,6 @@ requirements:
   run:
     - perl >=5.22.0
     - hmmer >=3.1b
-    - libgcc
 
 test:
   source_files:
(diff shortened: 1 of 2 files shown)


recipes/barrnap/0.8/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+build:
+  number: 0
+
+requirements:
+  build: []
+  run:
+    - perl >=5.22.0
+    - hmmer >=3.1b
+    - libgcc
```
bgruening: What is the error if you remove libgcc?

epruesse: None. Copy-paste from the 0.7 version.

Interesting, didn't see it actually build. It's still failing with `simulate-travis.py`...

epruesse: I assume it was originally in there because Barrnap upstream includes the `nhmmer` binaries for osx and linux and they were just copied into the conda package.

bgruening:
Thanks @epruesse 
@Slugger70 to we need a new Galaxy release here?

PR #5540, opened 2017-08-16
Mdasim

dlaehnemann:

This is a tool that is not developed any more, so this is supposed to be a reproducible snapshot installation. As it was developed for linux only and getting it to compile on OSX seems to require extensive debugging (I tried a few steps, but gave up), it now skips OSX builds -- and the Travis CI OSX build fails, accordingly.


dlaehnemann:
Thanks for catching this, I guess I was too quick on putting this in there!

bgruening:
No problem! Thanks a lot for your contributions!

PR #5553, opened 2017-08-17
samtools - pinned bzip2 to 1.0.* and zlib to 1.2.8

notestaff:

### DIFF
--- a/recipes/samtools/meta.yaml
+++ b/recipes/samtools/meta.yaml
@@ -5,7 +5,7 @@ package:
   version: {{ version }}
 
 build:
-  number: 1
+  number: 2
   skip: False
 
 source:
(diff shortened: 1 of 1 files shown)


bgruening:
Thanks!

notestaff:
@bgruening Maybe, when a recipe gets updated, recipes that depend on it and were not pinned to earlier versions could be automatically re-tested?

johanneskoester:
Yes, this should definitely be done in some form. We can even periodically scan updates in conda-forge and defaults. If the latest version is newer than one of our recipes, we should trigger a test. I'll think about it.

notestaff:
@johanneskoester thanks!  the samtools issue here resulted from a bzip2 update on the defaults channel.

if any test fails, could automatically pin it to the "last good" version.

bgruening:
I don't think this is practically possible with our open-, priceless infrastructure we are currently using. And even if we could do this with our university clouds this would waste a lot of CPU cycles and power - save the environment! :)

I think we are pretty bad at pinning and we need to pin way more libraries and align more with conda-forge here. This would have prevented your issue in the first place.
Luckily this will get much easier in the end of the year with conda-build-3. Until then we should try to pin more things directly, even zlib.

johanneskoester:
You might be right. I'll nevertheless think about it. I still have the feeling that this can be done efficiently when we are careful

PR #5557, opened 2017-08-17
first batch of biotools packages starting with c

ArneKr:


recipes/bioconductor-cancermutationanalysis/build.sh
```diff
@@ -0,0 +1,6 @@
+
```
bgruening: @ArneKr can you please remove the leading newlines here?
This should be fixed in the newest skeleton generator.

PR #5589, opened 2017-08-18
simulate-travis.py: Run (mulled) docker build/test on OSX

epruesse:
On OSX with `docker` installed, we can run both the OSX and the Linux targets. This patch makes `simulate-travis.py` do just that.

The default is to run all possible tests. The Linux/Docker tests are disabled reusing `--disable-docker`, the local OSX test is disabled via a new flag `--disable-local`.

I chose to make "all tests" the default because 
  - that is what the name `simulate-travis.py` implies
  - the `--disable-local` flag matches the existing flags (`--disable-docker`, `--disable-mulled`)
  - the stage in package development where one switches from`conda-build` to `simulate-travis.py` is after the package "builds in principle", so rigorous testing mirroring Travis as closely as possible is usually desired.
  - OSX users having docker installed and running, probably want it to be used for tests

The code will try to get the Docker environment variables via `docker-machine env`. If the binary is missing or the VM is not currently running, a notice is printed but the test is not aborted.

### DIFF
--- a/simulate-travis.py
+++ b/simulate-travis.py
@@ -97,6 +97,12 @@ ap.add_argument('--disable-mulled', action='store_true', help='''By default, if
                 then we run independent mulled-build tests on
                 a successfully-built recipe. Use this argument to disable this
                 behavior''')
+ap.add_argument('--disable-local', action='store_true', help='''If you are on
+                OSX, have docker-machine installed and docker is running, we
+                run building and testing for both OSX and Linux using docker
+                for the latter. You can disable the Linux build by specifying
+                 --disable-docker or disable the OSX build using this argument.
+                ''')
 ap.add_argument('--alternative-conda', help='''Path to alternative conda
                 installation to override that installed and configured with
                 --install-alternative-conda. If the conda executable you want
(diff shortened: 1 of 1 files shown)


simulate-travis.py
```diff
@@ -97,6 +97,12 @@
                 then we run independent mulled-build tests on
                 a successfully-built recipe. Use this argument to disable this
                 behavior''')
+ap.add_argument('--disable-local', action='store_true', help='''If you are on
```
bgruening: Should we rename these to `--disable-linux-tests` and `disable-osx-tests`? `local` and `docker` does not seem to be very meaningful.

epruesse: I though about using a `make` style syntax: `./simulate-travis.py [target [target...]]` where `target` is one of `native`, `docker`, `linux`, `all`, `osx` and `mulled`:

host/target|native|docker|mulled|linux|osx|all (default)
-|-|-|-|-|-|-
linux|native|FAIL|FAIL|native|FAIL|native
osx|native|FAIL|FAIL|FAIL|native|native
linux+docker|native|docker|mulled|mulled|FAIL|mulled
osx+docker|native|docker|mulled|mulled|native|mulled+native

Or, as if-else:

`native` :: build directly
`docker` :: build in `docker`, FAIL if `no docker`
`mulled` :: build in `docker` and run `mulled` tests, FAIL if `no docker`

`linux` :: if `docker` is available, `mulled`; otherwise if on `linux` `native` and if on `osx` FAIL
`osx` :: if on `osx` `native`; otherwise FAIL

`all` (default) :: if on `linux`, `linux`; if on osx `native` _and_ if `docker` is available `mulled`

daler: I really like this target syntax and gives a lot of control esp to osx folks with docker.

simulate-travis.py
```diff
@@ -257,6 +263,58 @@ def _set_channel_order():
...
+    try:
+        sp.check_call(['scripts/travis-run.sh'], env=env, universal_newlines=True)
+    except sp.CalledProcessError:
+        sys.exit(1)
+
+
+def _docker_machine_run_travis(env):
+    try:
+        p = sp.check_output(['docker-machine', 'env'])
+        os.environb.update({
```
daler: Just curious: what's the rationale for bytestrings here, rather than `universal_newlines=True` in the `check_output` call?

daler:
@epruesse looks good as-is; if you can implement the `make`-style syntax, all the better.

simulate-travis.py
epruesse: I suppose I've learned to distrust shell interaction using unicode, too many things that can unexpectedly go wrong because some language flag is set. In this case, it shouldn't matter, either way would work fine.

 The parsing is sort of a hack anyway. A steel wrought solution would evaluate the output in `bash` as it is meant to be used and dump the resulting environment in a controlled fashion. But I believe it is rather unlikely that the docker folks will ever put anything funny in that output.

epruesse:
Merging as is for now. It may be a week or so until I get around to the discussed CLI change.

simulate-travis.py
daler: Makes sense, thanks.

daler:
Great, thanks for this!

PR #5629, opened 2017-08-19
Antismash

zachcp:

This is the Antismash package for finding and classifying biosynthetic gene clusters.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,8 @@ services:
 
 before_install:
     - set -e
-    - if [ "$TRAVIS_OS_NAME" == "linux" ]
+    - |
+      if [ "$TRAVIS_OS_NAME" == "linux" ]
       then
         rm -rf /home/travis/perl5/ /home/travis/.phpenv /home/travis/gopath /home/travis/otp
         rm -rf /home/travis/.gimme/ /home/travis/.rvm /home/travis/.nvm
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,15 @@ sudo: required
 
 services:
   - docker
+
+before_install:
+    - set -e
+    - if [ "$TRAVIS_OS_NAME" == "linux" ]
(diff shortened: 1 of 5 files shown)


zachcp:
looks like linux now builds!!

but macOS is failing. DO you have any idea whats happening here? (thanks, BTW)

bgruening:
Cool I will look into it as soon as I'm back. Can you create a activate.sh script that download all other databases after installation?

bgruening:
Good news is that it should pass the error was just a download problem. So not our fault. For a proper activate/deactivate script you need to copy the scripts into the correct location and you need to add a deactivate script.

zachcp:
Perhaps it is simpler for a user to download the databases prior to their first use?  I am not sure of the benefit of activate/deactivate

bgruening:
@zachcp I think people expect from conda package that they work out of the box. An other usecase is that workflow management systems can simply install the package and everything works, without human invention. If you don't have time for this, I will try to get to it.

Thanks again!

zachcp:
@bgruening ok, that seems like a reasonable answer.

I've changed the download script to be in the post-link.sh file as [per the instructions](https://bioconda.github.io/guidelines.html#link-and-unlink-scripts-pre-and-post-install-hooks).

Thanks again for your help

zachcp:
looks like we're still running out of space  😢 

```
error: [Errno 28] No space left on device
```

bgruening:
@zachcp please remove the post-link script again, I need to work more on space reduction until this will be possible :(
Sorry for taking your time :(

zachcp:
@bgruening no prob. thanks for working on this.

bgruening:
Thanks @zachcp!

zachcp:
@bgruening  BOOM!

bgruening:
!!! Now I need to update my Galaxy integration :)
Thanks @zachcp!

PR #5829, opened 2017-08-22
IgDiscover: Depend on seaborn 0.7

marcelm:
Seaborn 0.8, which has been released recently, contains a regression that
makes IgDiscover fail when creating some plots.

### DIFF
--- a/recipes/igdiscover/meta.yaml
+++ b/recipes/igdiscover/meta.yaml
@@ -24,7 +24,7 @@ requirements:
     - snakemake >=3.9.0
     - cutadapt
     - xopen >=0.1.1
-    - seaborn >=0.6.0
+    - seaborn >=0.6,<0.8
     - scipy >=0.16.1
     - ruamel.yaml
     - muscle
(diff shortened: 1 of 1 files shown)


recipes/igdiscover/meta.yaml
```diff
@@ -24,7 +24,7 @@ requirements:
     - snakemake >=3.9.0
     - cutadapt
     - xopen >=0.1.1
-    - seaborn >=0.6.0
+    - seaborn =0.7
```
bgruening: Can we pin this to <=0.7?

marcelm: Any particular reason for that? I don’t really care as long as I get 0.7.1 most of the time, which is the most recent working version.

marcelm: Would `<0.8` be ok as well?

bgruening: sure!

bgruening: If you pin to a specific version you restrict the possible environments that can be created. Just imagine a package that needs <=0.6. This package would not be compatible with your package here.

marcelm: Ok, I guessed that that was your reason. Unfortunately, we actually require Seaborn >= 0.6, so now that I’ve looked up the syntax for it, I’ve actually made the requirement even more restrictive.

marcelm:
Thanks!

Issue #976, opened 2016-03-07
weblogo - naming convention for python packages


sebastian-luna-valero:
Hi all,

One of my packages depends on a python module called `weblogo`:
https://pypi.python.org/pypi/weblogo

However, there is a recipe in bioconda called `weblogo` that contains a completely different program:
https://github.com/bioconda/bioconda-recipes/tree/master/recipes/weblogo
http://weblogo.berkeley.edu/

I am about to add the `weblogo` package hosted in `PyPI` to bioconda but what would be a proper name for it? I guess we should define a common policy for this kind of problems.

Many thanks,
Sebastian.

johanneskoester:
Maybe the policy could be to prefix with "python-" in case of ambiguities, i.e. python-weblogo? @bioconda/all, any opinions?

daler:
`python-` prefix sounds reasonable. There's precedence at least: I did this with `python-wget` for the PyPI package `wget`.

acaprez:
Sounds fine to me as well.

brentp:
Sounds good

sebastian-luna-valero:
Many thanks all, I will use `python-weblogo`

Best regards,
Sebastian.

vivekiitkgp:
Probably it'll be a good idea to mention this somewhere on README or
CONTRIBUTING guidelines.

--
https://vivekiitkgp.github.io

PR #5837, opened 2017-08-22
added weblogo3

rfarouni:

This recipe is for the latest version of WebLogo (WebLogo 3.5.0). It updates the recipe `python-weblogo` for WebLogo 3.4.0. Since there is another recipe for `weblogo` for WebLogo 2, and `python-weblogo` is not a descriptive name for WebLogo3 (it can be easily missed), I am adding a third recipe with the name `weblogo3`


daler:
weblogo 2 and weblogo 3 seem to be substantially different (perl and python respectively). The separation of `weblogo` and `python-weblogo` bioconda recipes appears to have come from https://github.com/bioconda/bioconda-recipes/issues/976. However, nowadays we would probably resolve this by pinning to `weblogo >3` in a recipe requiring the python version. I believe #976 was also before we had a system for retaining older version, further contributing to the rename.

I would propose the following:

- remove existing recipe for `python-weblogo` from git (the conda package will remain on anaconda.org)
- move existing `weblogo` to `weblogo/2.82`
- track the latest `weblogo 3.x` in the main `weblogo` recipe

@bioconda/core, thoughts on this?

rfarouni:
@daler 

Agree. Also The recipe in  `python-weblogo` doesn't not include the ghostscript in the requirements. Also when I run weblogo I get a numpy error that I don't get when I install WebLogo 3.5 from pip

bgruening:
:+1: for @daler's suggestion!

rfarouni:
@daler The weblogo/2.8.2 recipe keeps on failing. 

```

Here is my latest attempt

daler:
@rfarouni I fixed seqlogo 2.8.2 to use `#!/usr/bin/env perl` instead of the hard-coded `#!/usr/bin/perl`, and added numpy as a build-time dependency to 3.5.0, since its setup.py seemed to need it. Might need an extra test to check for graphviz though.

bgruening:
LGTM! Thanks a lot @rfarouni and @daler!

yhoogstrate:
I think the license is MIT (https://github.com/WebLogo/weblogo/blob/master/LICENSE.txt)

PR #5838, opened 2017-08-23
Du Novo: Include VERSION file in build, to allow --version and --phone-home to work.

NickSto:

This update simply adds a line to `build.sh` to move the `VERSION` file to `$PREFIX/lib`. This allows the various scripts to find the current Du Novo version number, allowing the `--version` option to work. This also allows `--phone-home` to work, since that requires the version number.


NickSto:
Thanks, Bjorn!

Issue #5429, opened 2017-08-03
Missing busco config file


mschatz:
The current busco recipe correctly installs the dependencies, but leaves out the required config.ini file. It is not terribly difficult to make, but would totally streamline the process to have one included. The default config file is available here:
https://gitlab.com/ezlab/busco/blob/master/config/config.ini.default

All of the required paths can be set to the miniconda bin directory:

```
# BUSCO specific configuration
# It overrides default values in code and dataset cfg, and is overridden by arguments in command line
# Uncomment lines when appropriate
[busco]
# Input file
;in = ./sample_data/target.fa
;# Run name, used in output files and folder
;;out = SAMPLE
;# Where to store the output directory
;;out_path = ./sample_data
;# Path to the BUSCO dataset
;;lineage_path = ./sample_data/example
;# Which mode to run (genome / protein / transcriptome)
;;mode = genome
;# How many threads to use for multithreaded steps
;;cpu = 1
;# Domain for augustus retraining, eukaryota or prokaryota
;# Do not change this unless you know exactly why !!!
;;domain = eukaryota
;# Force rewrite if files already exist (True/False)
;;force = False
;# Restart mode (True/False)
;;restart = False
;# Blast e-value
;;evalue = 1e-3
;# Species to use with augustus, for old datasets only
;;species = fly
;# Augustus extra parameters
;# Use single quotes, like this: '--param1=1 --param2=2'
;;augustus_parameters = ''
;# Tmp folder
;;tmp_path = ./tmp/
;# How many candidate regions (contigs, scaffolds) to consider for each BUSCO
;;limit = 3
;# Augustus long mode for retraining (True/False)
;;long = False
;# Quiet mode (True/False)
;;quiet = False
;# Debug logs (True/False), it needs Quiet to be False
;;debug = True
;# tar gzip output files (True/False)
;;gzip = False
;# Force single core for the tblastn step
;;blast_single_core = True

[tblastn]
# path to tblastn
path = /work-zfs/mschatz1/mschatz/build/miniconda/bin/

[makeblastdb]
# path to makeblastdb
path = /work-zfs/mschatz1/mschatz/build/miniconda/bin/

[augustus]
# path to augustus
path = /work-zfs/mschatz1/mschatz/build/miniconda/bin/

[etraining]
# path to augustus etraining
path = /work-zfs/mschatz1/mschatz/build/miniconda/bin/
;
;# path to augustus perl scripts, redeclare it for each new script
[...]

PR #5846, opened 2017-08-23
Busco config

daler:

Another attempt at #5429.

Since there's a lot going on with dependencies and paths, this recipe now includes a minimal functional test. Data (10MB) are temporarily downloaded within `run_test.sh`. The test includes a command similar to https://github.com/bioconda/bioconda-recipes/issues/5429#issuecomment-322312966.

@mschatz your comment shows that the config needs to be adjusted depending on the tool, so I had added back in the more flexible python config generator that is run in the post-link step. However the test seems to be running without special-casing the augustus paths or the env var, so we might be OK as-is. Is there something in the eukaryota_odb9 example you show that triggers a call to gff2gbSmallDNA.pl that is not being triggered in this test with proteobacteria_odb9?

### DIFF
--- a/recipes/busco/build.sh
+++ b/recipes/busco/build.sh
@@ -8,6 +8,10 @@ SHARE=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
 mkdir -p $SHARE
 cp config/config.ini.default $SHARE/config.ini.default
 
+cp $RECIPE_DIR/generate-busco-config.py $SHARE/generate-busco-config.py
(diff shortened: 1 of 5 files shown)


mschatz:
I dont think there should be any differences between a proteobacteria_odb9
and eukaryota_odb9 example (at least not for the tools used). Is there an
easy way I can test this, or should I wait for all automatic tests to
complete?

Mike

daler:
@mschatz I just merged it, we need to wait for these to complete before it's available on the channel: https://travis-ci.org/bioconda/bioconda-recipes/builds/267968314

PR #5849, opened 2017-08-24
Recipe for SeMeta 1.0

keuv-grvl:

Actually, this recipe does not really make the software ready-to-use. I added a note at the bottom of the recipe because running this software requires some additional configuration (get databases, put your data in a `./input/` folder, ...).

Any thoughts ?

### DIFF
--- a/recipes/semeta/build.sh
+++ b/recipes/semeta/build.sh
@@ -1,13 +1,23 @@
+#!/usr/bin/env bash
 make clean
 make
 
-INSTALL_DIR=$CONDA_PREFIX/share/SeMeta-1.0/
-mkdir -p $INSTALL_DIR
+INSTALL_DIR="$CONDA_PREFIX/share/SeMeta-1.0"
 
-cp SeMeta_Assign $INSTALL_DIR
-cp SeMeta_Cluster $INSTALL_DIR
-cp config.conf $INSTALL_DIR
-cp README.txt $INSTALL_DIR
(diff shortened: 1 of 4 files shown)


daler:
Does the data have to go to an `./input` directory next to the executable? It might be useful to either include a script that downloads data (but does not run when building the package) or that at least prints out the location of the current environment's `share` directory so it's easy to figure out where you need to download data.

keuv-grvl:
I wrote a script to download and configure the databases (NCBI nt and NCBI taxonomy). There are then located in `$CONDA_PREFIX/share/SeMeta1.0/database/` and `$CONDA_PREFIX/share/SeMeta1.0/taxonomy/`. The database folder can be anywhere we want as long it configured in `config.conf` but the taxonomy files must be in a `taxonomy` directory next to the executable. A `post-link.sh` script print a message showing which command the user should run the get the databases.

I chose to download NCBI nt because of simplicity (one single file) but final users can set any database (RefSeq is mentioned in README.txt for example).

The `./input` folder have to be in the same directory than the executable (it is hardcoded). All results are saved in a `./output` folder which also must be along with the executable. Maybe we should wait for a new release before add it to Bioconda...

daler:
@keuv-grvl this is looking great. We can only do so much on the packaging side in making things user friendly! If there is a need for this package to be in bioconda, no reason to wait for a new release, you can merge this. You might want to consider opening an issue or PR to upstream though so the authors know about it.

PR #5866, opened 2017-08-25
Removed julia (available in conda-forge)

dfornika:

Julia is now [available on conda-forge](https://github.com/conda-forge/julia-feedstock), which is a more appropriate channel for it.


bgruening:
Awesome, thanks again for your migration effort!

dfornika:
Is there any way to remove the existing julia-0.5.2 build from the bioconda channel on anaconda.org? I'd like to package a julia application and it's pulling the bioconda version of julia, but I need it to depend on the conda-forge version.

bgruening:
@dfornika done!

bgruening:
To provide a little bit background. I just changes the label so they are still in the channel but not available until to specify a label during installation. I tend to really remove this package as it was kind of broken in old releases. Opinions?

dfornika:
I don't know anything about the state of the older builds. It seems reasonable to leave them marked as 'deprecated' for a while (a few months?) and then drop them once the conda-forge build(s) take over.

PR #5875, opened 2017-08-28
Update umi tools to 0.5.0

davebx:
* [x] This PR is based on the hard work of @bgruening


recipes/umi_tools/meta.yaml
```diff
@@ -20,6 +27,11 @@ requirements:
     - future
   run:
     - python
+    - matplotlib
+    - scipy
+    - regex
+    - freetype
+    - certifi ==2016.9.26
```
croth1: Is it necessary to pin certifi to a specific version here?

PR #5877, opened 2017-08-29
Update megahit to 1.1.2

epruesse:
- adds OSX support
 - fixes missing pin for zlib
 - test checks binaries (not just python runner script)


bgruening:
Looks good need to follow up with the cleanups. There is a new mulled release which we should use ... narf too much too do.

epruesse:
@bgruening I put the cause for the workaround in a separate ticket on bioconda-utils. If you choose the preferred fix I can prepare a PR.

PR #5880, opened 2017-08-29
Update ncbi-genome-download to 0.2.5

kblin:
Signed-off-by: Kai Blin <kblin@biosustain.dtu.dk>

### DIFF
--- a/recipes/ncbi-genome-download/build.sh
+++ b/recipes/ncbi-genome-download/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-$PYTHON setup.py install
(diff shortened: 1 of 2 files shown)


peterjc:
Why the addition of ``--single-version-externally-managed --record=record.txt``?

kblin:
I'm cargo-culting https://bioconda.github.io/linting.html#setup-py-install-args here, because the old `build.sh` fell over while linting with that error.

peterjc:
Thanks - a bit of BioConda magic I'd not seen before, but that seems clear enough. Merged.

PR #5886, opened 2017-08-29
added rpackages from biotools (4)

ArneKr:


daler:
@ArneKr @bgruening I just realized that the recent flood of bioconductor packages (this PR among them) are all for Bioconductor 3.5 which is tested with R 3.4. Since we're still on R 3.3, we should probably be targeting Bioconductor 3.4. Tests have been passing, so I guess it's fine, but any thoughts on this?

bgruening:
@daler I'm not aware of problems so far. I will push more the migration and then we can rebuild our packages as well.

PR #5924, opened 2017-09-01
Adding recipe for reaper

jenzopr:

This PR adds a recipe for the reaper, tally, minion and swan binaries from the kraken software tool set (http://www.ebi.ac.uk/research/enright/software/kraken). The tools are designed to streamline the analysis of next-generation sequencing data. Although designed with small RNA sequence analysis in mind the tools can be used to address issues facing next-generation sequencing in general.

Feedback is highly appreciated since this is my first PR to bioconda!
Best and thanks,
Jens

### DIFF
--- a/recipes/reaper/meta.yaml
+++ b/recipes/reaper/meta.yaml
@@ -10,10 +10,10 @@ requirements:
   build:
     - gcc # [not osx]
     - llvm # [osx]
-    - zlib
+    - zlib 1.2.8
   run:
     - libgcc # [not osx]
-    - zlib
+    - zlib 1.2.8
 
 test:
   commands:


recipes/reaper/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+source:
+  url: http://wwwdev.ebi.ac.uk/enright-dev/kraken/reaper/src/reaper-15-065.tgz
+  md5: 33e1b4c8b72d6460d2fc3a45affaa8a2
+
+requirements:
+  build:
+    - gcc # [not osx]
+    - llvm # [osx]
+    - zlib
```
bgruening: can you use `zlib 1.2.8`

recipes/reaper/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+  run:
+    - libgcc # [not osx]
+    - zlib
```
bgruening: here as well

recipes/reaper/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+cd src
+make
+mkdir -p $PREFIX/bin
+mv reaper $PREFIX/bin
+mv tally $PREFIX/bin
```
bgruening: Should tally, swan and minion be in a separate repository?

bgruening:
Thanks @jenzopr, a few comments inline.

recipes/reaper/build.sh
jenzopr: Thats a good point. They do not depend on each other, but can work together (especially reaper and tally, when it comes to trim paired-end sequencing files). The website states that `Tally and Minion are companion programs and shipped alongside reaper`.
How is this solved in other software tool sets? What would be the benefit to have them separately?

jenzopr:
I wanted to continue the discussion that @bgruening started in the review above. The question is, if it makes more sense to pack reaper, tally and minion in seperate recipes or not. If not, we could enhance the visibility of the three programs by changing the name of to kraken (or ebi-kraken, since kraken is taken) to indicate that a tool set is being installed?

bgruening:
@jenzopr you are the maintainer ... but I do not see a disadvantage in splitting this into 3 repos. The benefit, though only slightly, is that I only install what I need. I can upgrade versions independently, save some disc-space. Disadvantage for you is a larger maintenance overhead.
Not sure about visibility - will there be ever someone that searches for `reaper` and do not know about ebi-kraken?

jenzopr:
Thanks a lot @bgruening! I agree that the benefit is a gain in flexibility. However, they are build together anyway and I guess that whoever looks for the toolset wants the full toolset. Independent upstream version upgrades will not very likely happen, since the tools are upgraded together (under the same version number).
I'll leave the name as is because reaper is the most prominent, and the source code fn has reaper inside and will add a note that tally, minion and swan come with the installation. How does that sound?

PR #5926, opened 2017-09-01
Bumped cnvkit version

micknudsen:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ env:
   global:
     - SUBDAGS=1
     - BIOCONDA_UTILS_TAG=master
-    - BIOCONDA_UTILS_BUILD_ARGS="--loglevel=info"
+    - BIOCONDA_UTILS_BUILD_ARGS="--loglevel=debug"
     - BIOCONDA_UTILS_LINT_ARGS=
     - MINICONDA_VER="4.2.12"
   matrix:
--- a/recipes/cnvkit/meta.yaml
+++ b/recipes/cnvkit/meta.yaml
@@ -1,17 +1,17 @@
-{% set version="0.8.6a0" %}
+{% set version="0.9.0" %}
 package:
   name: cnvkit
   version: {{ version }}
 
 source:
-  fn: cnvkit-4c94834.tar.gz
-  url: https://github.com/etal/cnvkit/archive/4c94834.tar.gz
+  fn: v0.9.0.tar.gz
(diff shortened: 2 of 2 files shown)


micknudsen:
How does one proceed when Travis runs out of time?

bgruening:
I guess this was just a hiccup with travis as the OSX build was successful. I restarted the build.

micknudsen:
Thanks. Unfortunately it timed out again. Same thing happened in my own fork. Restarted a couple of times, but same thing...

etal:
Looks like this build completes in 15 minutes on OS X, and bails after 1.5 hours on Linux. Is there anything I can change upstream (or via dependency versions) to speed this up?

bgruening:
Ok, lets see what happens with increased logging :)

etal:
If I'm reading the logs right, mulled-build takes a bit over an hour for each Python version, and the timeout lands in the middle of the second execution. (1) Is that expected? (2) Would it be possible to split the dockerization for Python 2.7 and Python 3.5 into separate jobs to stay under the 90-minute limit?

bgruening:
No this is not expected, a build is more or less a copy of the PREFIX into a container and then running the tests in this new container. I have no clue why this takes so long. I will try to reproduce it locally.

micknudsen:
I have recently had troubles installing certain bioconda packages. Solving dependencies ran for more than a day, so I killed the jobs (-vvv, the extremely verbose option, indicated progress, so it didn't just stall). This also happened for _cnvkit_, but today it all of sudden just worked.

Maybe this was to blame? Could you please try to restart the build?

daler:
I wonder if we're hitting a known bug in conda caused by too many duplicated R pacages across `conda-forge`, `bioconda`, and `r` channels. The issue in https://github.com/conda/conda/issues/5536 suggests modifying your `~/.condarc` to remove the `r` channel from defaults (as in this particular comment: https://github.com/conda/conda/issues/5536#issuecomment-310099762).

I think it will be worth it to convert bioconda-utils to using such a modified condarc. I'll add an issue in bioconda-utils.

chapmanb:
Thanks so much for all this work. I've been trying to debug this and couldn't replicate it with the latest packages so tried and submission and all passed smoothly. I wonder if this was related to the move to 3.4.1 on going at the time? Either way, I ended up bumping CNVkit to 0.9.0 so we should have the latest version now. Thanks again for all the work on this and hopefully all goes smooth on future updates.

PR #5928, opened 2017-09-02
riboSeed

nickp60:
Hi All, bear with me, this is my first time submitting a new recipe!

### DIFF
--- a/recipes/riboseed/meta.yaml
+++ b/recipes/riboseed/meta.yaml
@@ -1,0 +1,71 @@
+{% set name = "riboSeed" %}
+{% set version = "0.4.16" %}
+{% set hash_type = "sha256" %}
+{% set hash_value = "abb680a514058c2233f34a3c574727f61ed65e06b2123964b359683ccf9eb9f5" %}
+package:
+  name: '{{ name|lower }}'
+  version: '{{ version }}'
+
+source:
+  fn: '{{ name }}-{{ version }}.tar.gz'
+  # url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  url: https://test-files.pythonhosted.org/packages/21/59/ef9dbb631f132c312bdd1cc145395050fdd0bfbe46765280c9a157367dcd/riboSeed-0.4.16.tar.gz
+  '{{ hash_type }}': '{{ hash_value }}'
+
+build:
+  number: 0
+  skip: True  # [win]
+  skip: True  # [py2k]
+  skip: True  # [py34]
+  script: $PYTHON setup.py install --single-version-externally-managed --record=record.txt
+requirements:
+  build:
+    - python
+    - setuptools
+    - biopython ==1.68
+    - pysam ==0.11.2.2
+    - jenkspy ==0.1.4
+    - pyutilsnrw ==0.1.0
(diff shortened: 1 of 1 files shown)


recipes/riboseed/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+source:
+  fn: '{{ name }}-{{ version }}.tar.gz'
+  # url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  url: https://test-files.pythonhosted.org/packages/21/59/ef9dbb631f132c312bdd1cc145395050fdd0bfbe46765280c9a157367dcd/riboSeed-0.4.16.tar.gz
+  '{{ hash_type }}': '{{ hash_value }}'
+
+build:
+  number: 0
+  skip: True  # [win]
```
bgruening: This needs to be all in one line, so something like `skip: [win or py2k or py34]`
But it seems to work, so I'm wondering if you really need to skip so many things.

recipes/riboseed/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+  skip: True  # [py2k]
+  skip: True  # [py34]
+  script: $PYTHON setup.py install --single-version-externally-managed --record=record.txt
+requirements:
+  build:
+    - python
+    - setuptools
+    - biopython ==1.68
```
bgruening: Why are you pinning all packages this strictly, is there a reason? Otherwise, just write `biopython` and `pysam` etc...

recipes/riboseed/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+{% set hash_type = "sha256" %}
+{% set hash_value = "abb680a514058c2233f34a3c574727f61ed65e06b2123964b359683ccf9eb9f5" %}
+package:
+  name: '{{ name|lower }}'
+  version: '{{ version }}'
+
+source:
+  fn: '{{ name }}-{{ version }}.tar.gz'
+  # url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  url: https://test-files.pythonhosted.org/packages/21/59/ef9dbb631f132c312bdd1cc145395050fdd0bfbe46765280c9a157367dcd/riboSeed-0.4.16.tar.gz
```
bgruening: Why not using this URL? https://github.com/nickp60/riboSeed/archive/0.4.13.tar.gz

nickp60: Hi @bgruening , thanks for the comments!  This is a alpha test of getting all the dependencies working.  This is why I haven't linked to the stable Pypi link yet.

recipes/riboseed/meta.yaml
nickp60: Great, I will change that.

recipes/riboseed/meta.yaml
nickp60: There is a rather difficult web of dependency availability and congruency.  Having a more updated version of pysam doesn't work with the versions of numpy that work well, and so on.  Its a product of the available versions in conda-forge and bioconda.  I hope to get things sorted in the next minor release that will make the packages less strictly pinned.

PR #5930, opened 2017-09-03
Add Package chewbbaca-1.0

dfornika:


recipes/chewbbaca/build.sh
```diff
@@ -0,0 +1,15 @@
+DEST="$PREFIX/bin/chewBBACA"
```
bgruening: Yould it make sense to use PREFIX/lib/ here?

recipes/chewbbaca/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+  skip: True  # [osx]
+  number: 0
+
+requirements:
+  build:
+    - python
+
+  run:
+    - python
+    - numpy 1.13.1
```
bgruening: Is this strict pinning here needed?
Also for biopython and prodigal ...

recipes/chewbbaca/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - scipy >=0.13.3
+    - plotly 1.12.*
+    - biopython 1.70
+    - prodigal 2.6.3
+    - blast >=2.5.0
+
+test:
```
bgruening: Please add a test.

recipes/chewbbaca/meta.yaml
dfornika: I've done my best to match the versions listed on the [chewBBACA repo](https://github.com/B-UMMI/chewBBACA) and in the `requirements.txt` file

recipes/chewbbaca/build.sh
dfornika: Sure, I've made the change.

bgruening:
Thanks @dfornika!

PR #5931, opened 2017-09-03
Shovill

Slugger70:
Shovill is T Seemann's assembly pipeline.
Shovill is a pipeline which uses SPAdes at its core, but alters the steps before and after the primary assembly step to get near-identical results in far less time than SPAdes alone.


recipes/shovill/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+mkdir -p $PREFIX/bin
```
bgruening: You are not compiling here, but you use gcc as run time and build time dependency.
Do you really need to compile anything? If you just copy, you probably don't need anything as build time dependency.
Also zlib is probably not needed.

PR #5940, opened 2017-09-04
Changing kraken version naming

druvus:

@bioconda/core  I was planning to update kraken to a version that handle the new ncbi stucture better. However the the current bioconda version naming contains the commit id which would not work in the long run. I would therefor like to start by changing the version naming. Either somethin like I suggested here or to include date in ISO format instead of git commit id. Do you have any preferences?

### DIFF
--- a/recipes/kraken/meta.yaml
+++ b/recipes/kraken/meta.yaml
@@ -1,13 +1,15 @@
+{% set version = "0.10.6_beta.a0" %}
+
 package:
   name: kraken
-  version: "0.10.6_eaf8fb68"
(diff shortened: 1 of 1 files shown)


johanneskoester:
Can't we simply ask the devs for releasing a new version? I agree that the commit id is suboptimal, but starting our own versioning scheme seems to be suboptimal as well.

druvus:
I agree it would be best to get them to release a new version. It is two issues open at the kraken repo requesting that so I will keep this on hold for now.

PR #5958, opened 2017-09-06
Added a recipe for UROPA

jenzopr:

Hi all,
UROPA (‘Universal RObust Peak Annotator’) is a python-based tool, intended for genomic region annotation. Based on a configuration file, different target features can be prioritized with multiple integrated queries. These can be sensitive for feature type, distance, strand specificity, feature attributes (eg. protein_coding) or the anchor position relative to the feature. UROPA can incorporate reference annotation files (GTF) from different sources, like Gencode, Ensembl, or RefSeq, as well as custom reference files produced by the user. UROPA is published under the doi: [10.1038/s41598-017-02464-y](https://www.nature.com/articles/s41598-017-02464-y).

Because of the requirement installation burden, we would like to include UROPA in bioconda to enable the usage for a wider target audience.
I would be very happy to address any reviews or remarks.

Thanks a lot,
Jens

### DIFF
--- a/recipes/uropa/build.sh
+++ b/recipes/uropa/build.sh
@@ -9,3 +9,4 @@ mv uropa/* $PREFIX/bin/uropa
 # Auxillary scripts
 mv summary.R $PREFIX/bin
 mv reformat_output.R $PREFIX/bin
+mv utils/uropa2gtf.R $PREFIX/bin
--- a/recipes/uropa/build.sh
+++ b/recipes/uropa/build.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+mkdir -p $PREFIX/bin/uropa
+mv uropa.py $PREFIX/bin
+mv uropa/* $PREFIX/bin/uropa
--- a/recipes/uropa/build.sh
+++ b/recipes/uropa/build.sh
@@ -1,12 +1,3 @@
 #!/usr/bin/env bash
 
-# Main program
-mv uropa.py $PREFIX/bin
-
-mkdir -p $PREFIX/bin/uropa
-mv uropa/* $PREFIX/bin/uropa
-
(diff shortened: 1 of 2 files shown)


recipes/uropa/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - r-venndiagram
+    - r-getopt
+    - r-tidyr
+    - r-upsetr
+    - bioconductor-graph
+    - bioconductor-rbgl
+
+test:
+  commands:
+    - "uropa.py -h > /dev/null"
```
bgruening: The `/dev/null` is not really needed.

recipes/uropa/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+# Main program
+mv uropa.py $PREFIX/bin
+
+mkdir -p $PREFIX/bin/uropa
```
bgruening: Can this be located under lib?

jenzopr: I honestly don't know. Fact is that `uropa.py` relies on this (relative) structure to include further source files located in `uropa/` via pythons `include uropa.config`. Any idea how this could be solved? Creating a proper Pypi source package that can be installed via `python setup.py install`?

bgruening: From a community standpoint a pypi package would be ideal, but I will not push you in this direction. The PR is ok for BioConda as it is I think. But consider a real pypi package.

jenzopr:
@bgruening I packaged UROPA into a proper python package, submitted to PyPI and reflected the new source in the meta.yaml. The whole copying to bin/uropa etc. is not neccessary anymore.

Best,
Jens

bgruening:
Awesome Jens!

PR #5968, opened 2017-09-07
Add bwa 0.7.3a recipe

jtratner:
Based off of the bwa 0.7.8 recipe.

Based on md5sum pulled from sourceforge on 2017-09-07.

To get it to pass the `simulate-travis` tests, I added this to meta.yaml (taken from an earlier PR to add bwa 0.6):

```
test:
  commands:
    - bwa 2>&1 | grep 'index sequences'
```

but there do not appear to be tests for other bwa commands, so I'm not sure whether this is required.

At a minimum, it seems to build just fine.

### DIFF
--- a/recipes/bwa/0.7.3a/meta.yaml
+++ b/recipes/bwa/0.7.3a/meta.yaml
@@ -18,6 +18,10 @@ requirements:
     - libgcc
     - zlib
 
+test:
+  commands:
+    - bwa 2>&1 | grep 'index sequences'
+
 about:
   home: http://bio-bwa.sourceforge.net
   license: MIT
--- a/recipes/bwa/0.7.3a/meta.yaml
+++ b/recipes/bwa/0.7.3a/meta.yaml
@@ -1,0 +1,24 @@
+package:
+  name: bwa
+  version: "0.7.3a"
+
+build:
+  number: 0
+
+source:
+  fn: bwa-0.7.3a.tar.bz2
+  url: http://downloads.sourceforge.net/project/bio-bwa/bwa-0.7.3a.tar.bz2
+  md5: 7ed9f5d750d26f5bf91cee1fa6f65df2
+
+requirements:
+  build:
+    - gcc
+    - zlib
+  run:
+    - libgcc
+    - zlib
+
+about:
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
LGTM

recipes/bwa/0.7.3a/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+source:
+  fn: bwa-0.7.3a.tar.bz2
+  url: http://downloads.sourceforge.net/project/bio-bwa/bwa-0.7.3a.tar.bz2
+  md5: 7ed9f5d750d26f5bf91cee1fa6f65df2
+
+requirements:
+  build:
+    - gcc
+    - zlib
```
bgruening: zlib should be pinned if possible
And it would be nice to use GCC only for Linux

kyleabeauchamp: Sorry I didn't catch this earlier, my linter isn't as sharp.  Do I need to do an update PR or is this fine to leave as is?

bgruening: I think it is fine for the moment. But please keep this in mind. zlib already broke ABI once.

jtratner: what version would you pin to? How would you use gcc only for linux? we can update all of the bwa builds at once after that

bgruening: https://github.com/conda-forge/conda-forge.github.io/blob/0661e2ecf4b213ddf36350ca100a5a820eb17eb7/scripts/pin_the_slow_way.py#L85 :)

kyleabeauchamp: Here is the GCC-linux only logic: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/samtools/meta.yaml#L18

That version of zlib might also be acceptable, not 100% sure what is best.

jtratner: nvm, I see how to only use gcc for linux.

PR #5982, opened 2017-09-10
Add recipe for virema 0.6

npavlovikj:


recipes/virema/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+source:
+  fn: ViReMa_0.6.zip
+  url: https://sourceforge.net/projects/virema/files/ViReMa_0.6/ViReMa_0.6.zip
+  md5: 2e15fcea133a64b4b043dba39d87f289
+  patches:
+    - path.patch
+
+build:
+  number: 0
+  skip: True # [osx or py3k]
```
bgruening: Why is this skiped for OSX?

recipes/virema/meta.yaml
```diff
@@ -0,0 +1,30 @@
+package:
+  name: virema
+  version: "0.6"
+
+source:
+  fn: ViReMa_0.6.zip
+  url: https://sourceforge.net/projects/virema/files/ViReMa_0.6/ViReMa_0.6.zip
+  md5: 2e15fcea133a64b4b043dba39d87f289
+  patches:
+    - path.patch
```
bgruening: What is this patch doing?

npavlovikj: @bgruening , the scripts are python scripts, so this patch adds "/opt/anaconda1anaconda2anaconda3/bin/python", and there is also "^M" in the python script after this (:?), so I remove that as well. Is that ok ?

bgruening: You should add `#!/usr/bin/env python` and not the long /opt/anaconda prefix. I would not touch the line endings if not needed.

npavlovikj: Thanks for the suggestion @bgruening , I fixed that in the newest commit.

bgruening: Can you try to make the patch smaller and not change line endings. Its hard to read the patch this way.

npavlovikj: Sorry about that @bgruening , it is fixed now.

epruesse:
LGTM. Thanks @npavlovikj!

PR #5995, opened 2017-09-11
Add matplotlib-nogui

epruesse:

This adds `matplotlib-nogui` as a version of `matplotlib` with reduced dependency tree and modifies `quast` to rely on that package. The plain `matplotlib` package builds and depends on all GUI backends supported by `matplotlib` which lead to a large amount of packages installed needlessly for non-interactive tools. (Which is an issue when running CI tests of depending tools/pipelines as the conda installation frequently times out).

I'm drawn between wanting this package in bioconda or conda-forge. Conda-forge has the `matplotlib` package, however, the issue has been discussed at conda and conda-forge before and re-starting it would likely go into the "doing it right" direction, splitting the package into multiple sub- and meta-packages, and ultimately get deferred past conda 3.0.

In any case, I'd like to see the more rigorous mulled tests to come through here first.

Opionions re bioconda vs conda-forge? @bgruening what do you think?

### DIFF
--- a/recipes/matplotlib-nogui/build.sh
+++ b/recipes/matplotlib-nogui/build.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+cat <<EOF > setup.cfg
+[directories]
+basedirlist = $PREFIX
+
+[packages]
+tests = False
+sample_data = False
+toolkits = False
+
+[gui_support]
+agg = True
+cairo = False
+gtk = False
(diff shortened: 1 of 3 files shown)


bgruening:
conda-forge :)

epruesse:
@bgruening See conda-forge/matplotlib-feedstock#2 stalled for 1 1/2 years.

PR #5998, opened 2017-09-11
update: rtg-tools 3.8.4

Lenbok:

Recipe no longer installs the "third-party" directory, as this is not required at runtime (it is included in the original distribution in order to provide source code to a modified library class).

Recipe now installs the "scripts" directory, which contains a bash-completion helper and some other small resources that may be useful to the end user. (Although this is optional and could be dropped if required.)

Recipe is explicit in settings for usage logging (false) and crash-only reporting (true).


bgruening:
@Lenbok this looks like a nice improvement. Thanks.

PR #6001, opened 2017-09-12
Quasitools recipe

camytran:


camytran:
Hello! The travis CI build is failing during testing in linux, but not osx. I am not sure what the cause of the error is ( SERR FileNotFoundError: [Errno 2] No such file or directory: 'locale' ), and am wondering if someone can take a look at the recipe I've added. Thanks :)

camytran:
I believe the failure is caused by a known issue with click. Several other recipes state: "# click needs to be patched in order to properly detect unicode in py3k" and currently have build/skip set to true.

bgruening:
Thanks @camytran!

PR #6005, opened 2017-09-12
Fix BLAST 2.6.0 when LD_LIBRARY_PATH is defined

nsoranzo:
If `--with-hard-runpath` configure option is not used, BLAST programs search libraries first in the directories defined by the `LD_LIBRARY_PATH` environment variable, instead of using the rpath specified by conda.

Also add patch to support Boost 1.64.0 and later.

### DIFF
--- a/recipes/blast/build.sh
+++ b/recipes/blast/build.sh
@@ -2,7 +2,7 @@
 
 set -e -x -o pipefail
 
-mkdir -p $PREFIX/bin/
+mkdir -p $PREFIX/bin
 
 if [ "$(uname)" == "Darwin" ]; then
     echo "Platform: Mac"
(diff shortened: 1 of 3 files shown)


recipes/blast/build.sh
```diff
@@ -13,7 +13,7 @@ elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
     echo "Platform: Linux"
 
     cd $SRC_DIR/c++/
-    ./configure --prefix=$PREFIX
+    ./configure --prefix=$PREFIX --with-hard-runpath
```
bgruening: Would be nice to add a comment here as well.

nsoranzo: Will do.

nsoranzo:
`The job exceeded the maximum time limit for jobs, and has been terminated.` should I split in to 2 PRs (one for 2.5.0 and one for 2.6.0)?

bgruening:
Yes please split into two.

nsoranzo:
The fix for BLAST 2.5.0 has been moved to https://github.com/bioconda/bioconda-recipes/pull/6009 .

PR #6012, opened 2017-09-13
update spades to 3.11.0; changed to compile from source.

notestaff:

### DIFF
--- a/recipes/spades/meta.yaml
+++ b/recipes/spades/meta.yaml
@@ -16,7 +16,6 @@ requirements:
   build:
     - gcc >=4.8.2  # [not osx]
     - llvm  # [osx]
-    - python
     - cmake >=2.8.12
     - zlib
     - bzip2
@@ -28,10 +27,13 @@ requirements:
     - bzip2
 
 test:
-  requires:
-    - setuptools
   commands:
     - spades.py --version
+    - rnaspades.py --version
+    - truspades.py --version
+    - dipspades.py --version
+    - metaspades.py --version
+    - plasmidspades.py --version
     - spades.py --test && rm -rf spades_test
 
 about:
--- a/recipes/spades/meta.yaml
+++ b/recipes/spades/meta.yaml
@@ -14,9 +14,10 @@ build:
 
 requirements:
   build:
-    - gcc  # [not osx]
+    - gcc >=4.8.2  # [not osx]
     - llvm  # [osx]
-    - cmake
+    - python
+    - cmake >=2.8.12
     - zlib
     - bzip2
 
@@ -31,7 +32,7 @@ test:
     - setuptools
   commands:
     - spades.py --version
-    - spades.py --test
+    - spades.py --test && rm -rf spades_test
 
 about:
   home: http://cab.spbu.ru/software/spades/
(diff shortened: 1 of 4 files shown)


notestaff:
this PR is a small cleanup of #6010 ; made it a separate PR in case the small changes break the recipe.  But #6010 should be fine to merge.

recipes/spades/meta.yaml
```diff
@@ -1,29 +1,49 @@
...
+  url: http://cab.spbu.ru/files/release{{ version }}/SPAdes-{{ version }}.tar.gz
+  sha256: 308aa3e6c5fb00221a311a8d32c5e8030990356ae03002351eac10abb66bad1f
 
 build:
   number: 0
-  skip: True  # [not py2k and not py35]
 
 requirements:
   build:
+    - gcc >=4.8.2  # [not osx]
```
bgruening: is this pinning really needed?

recipes/spades/meta.yaml
```diff
@@ -1,29 +1,49 @@
...
+    - llvm  # [osx]
+    - cmake >=2.8.12
+    - zlib
```
bgruening: can you pin zlib as well?

recipes/spades/meta.yaml
```diff
@@ -1,29 +1,49 @@
...
+    - bzip2
```
bgruening: This one as well please: https://github.com/conda-forge/conda-forge.github.io/blob/0661e2ecf4b213ddf36350ca100a5a820eb17eb7/scripts/pin_the_slow_way.py#L40-L85

recipes/spades/meta.yaml
notestaff: spades docs say it is: http://cab.spbu.ru/files/release3.11.0/manual.html#sec2.3

recipes/spades/meta.yaml
notestaff: in most bioconda-recipes/recipes/*/meta.yaml zlib is not pinned.  same for bzip2.  is that a bug?

recipes/spades/meta.yaml
bgruening: This is very unusual for Bioconda, we only have gcc 4.8.5.

notestaff: pinning to >=4.8.2 is consistent with that, no?

bgruening: Yes, just saying that this is very unusual ... If you want to keep it, let's merge this PR.

notestaff: Reason to keep is, if someone later looks at the recipe as a guide for other packaging (like I do sometimes with Homebrew formulas), it's best if the requirement from spades docs is preserved.

druvus:
thanks @notestaff for fixing it!

notestaff:
Hm, building the final package takes 49 min 26 sec
https://travis-ci.org/bioconda/bioconda-recipes/jobs/275144287
Time limit for travis jobs is 50 min: https://docs.travis-ci.com/user/customizing-the-build/

But that's because the builds for the three python versions are done sequentially.   Maybe worth configuring travis to make them separate jobs?  @bgruening

bgruening:
@notestaff we have an extended build time in Bioconda, so this should not a problem for this package.

recipes/spades/meta.yaml
tseemann: I agree with @notestaff - if there is a real minimum, put it in!

notestaff: I think @bgruening's concern was, this makes it look like all other bioconda recipes have _no_ minimum, which probably isn't true.

PR #6013, opened 2017-09-13
Bwa 0.7.16

johanneskoester:

I have also updated the homepage and added the two perl scripts.


johanneskoester:
@bgruening there seems to be a problem with mulled: https://travis-ci.org/bioconda/bioconda-recipes/jobs/274969587#L826
It complains about an invalid bind argument...

jtratner:
@johanneskoester - can you copy/template this over to the other bwa recipes too?

recipes/bwa/meta.yaml
```diff
@@ -1,29 +1,34 @@
...
+    - zlib {{CONDA_ZLIB}}*
   run:
-    - zlib
+    - zlib {{CONDA_ZLIB}}*
     - libgcc  # [not osx]
+    - perl
 
 test:
   commands:
     - bwa 2>&1 | grep 'index sequences in the'
```
jtratner: @johanneskoester - this is causing mulled to fail (the `'` causes this command to break out of the shell escaping) - you need to use `"index sequences in the"` instead)

johanneskoester: Thanks!

johanneskoester:
@jtratner question is whether we have the need to update the old versions as well. Usually our policy is that we only modify the latest version unless somebody explicitly needs the older version (of course, the older package version won't be deleted). So, the fact that we have dedicated folders for the old versions here already indicates this. However, it could also be that they are not needed anymore.

* @pinguinkiste do you specifically need bwa 0.5.9 and 0.7.8?
* @JinfengChen do you specifically need bwa 0.6.2?

jtratner:
@johanneskoester - gotcha - well if you could update 0.7.3a that'd be great :) . My understanding is that point versions of bwa are quite different (i.e., it doesn't use semantic versioning)

PR #6025, opened 2017-09-14
Illumina interop: Prefix command names with interop

jtratner:

@matthdsm put up an interop recipe for Illumina's interop library in  #6015, but commented that the command names were very generic and looked for a solution. I agree - the names should be prefixed, otherwise I think they're likely to unintentionally shadow other tool names (plus harder to tell where they are from in scripts).

This recipe updates the illumina-interop recipe to prefix all commands with `interop_`.

Not sure what the policy is regarding these sorts of changes / respecting tool names.

cc @kyleabeauchamp

### DIFF
--- a/recipes/illumina-interop/build.sh
+++ b/recipes/illumina-interop/build.sh
@@ -1,9 +1,13 @@
 #!/bin/bash
 
+mkdir -p $PREFIX/interop/bin
 mkdir -p $PREFIX/bin
 
 mkdir -p build
 cd build
-cmake -DCMAKE_INSTALL_PREFIX:PATH=$PREFIX ..
+cmake -DCMAKE_INSTALL_PREFIX:PATH=$PREFIX/interop ..
 make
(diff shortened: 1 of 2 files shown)


jtratner:
cc @kyleabeauchamp

kyleabeauchamp:
AFAIK we have been somewhat flexible w.r.t. tool renames of this nature.  I believe our wrappers for picard are one example (https://github.com/bioconda/bioconda-recipes/blob/master/recipes/picard/picard.sh) where we've taken some liberty to modify a tool to play nicely with our installation requirements.

@bgruening  thoughts?  To me this approach sounds reasonable...

bgruening:
Yes this looks good to me, we have done this before. But maybe we can add a post installation script to let people know about this? A simple one like: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/dfam/post-link.sh

PR #6035, opened 2017-09-14
clustergrammer

cristhomas:

### DIFF
--- a/recipes/clustergrammer/meta.yaml
+++ b/recipes/clustergrammer/meta.yaml
@@ -1,0 +1,32 @@
+package:
+  name: clustergrammer
+  version: "1.13.5"
+
+source:
+  fn: clustergrammer-1.13.5.tar.gz
+  url: https://pypi.python.org/packages/f4/ad/c0d0fd6e1c917d51df418d084201ef87c9b17288284870a31ac3ecebd571/clustergrammer-1.13.5.tar.gz
+  md5: d7b600b34c5674710f83efbeddbc50f7
+
+requirements:
+  build:
+    - python
+    - numpy
+    - scikit-learn
+    - pandas
+
+  run:
+    - libgcc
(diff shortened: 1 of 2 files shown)


recipes/clustergrammer/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - scikit-learn
+    - pandas
+
+test:
+  imports:
+    - clustergrammer
+
+about:
+  home: https://github.com/MaayanLab/clustergrammer-py
+  license: UNKNOWN
```
bgruening: https://github.com/MaayanLab/clustergrammer-py/blob/master/LICENSE

Its MIT licensed afaik.

recipes/clustergrammer/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+requirements:
+  build:
+    - python
+    - numpy
+    - scikit-learn
+    - pandas
+
+  run:
+    - libgcc
```
bgruening: Is libgcc really needed here?

cristhomas: it was when I tested locally.

recipes/clustergrammer/meta.yaml
cristhomas: thank you, missed that.

recipes/clustergrammer/meta.yaml
cristhomas: as in, that's the only way it passed the mulled tests (and I tried a bunch of things)

bgruening: Yeah, this can be. If so then please add `- libgcc  # [linux]`

cristhomas: ok! thank you

PR #6036, opened 2017-09-15
spades genome assembler - on osx fetch binaries rather than compile

notestaff:
spades genome assembler - on osx fetch binaries rather than compile, because compiling disables multi-threading (see https://github.com/ablab/spades/issues/30#issuecomment-329631647 )


druvus:
Wouldn't it be an idea to also use gcc for osx to get openmp support? I know some other recipes that solved it this way.

notestaff:
@druvus I'll try that next; I just want to first fix the recipe quickly so we aren't distributing a subpar spades.   Travis seems slow to check recipes right now...

PR #6043, opened 2017-09-15
Update newick utils

Cashalow:

The latest stable version is too old (2011) and lacks many features

### DIFF
--- a/recipes/newick_utils/build.sh
+++ b/recipes/newick_utils/build.sh
@@ -3,5 +3,4 @@ autoreconf -fi
 ./configure --prefix=$PREFIX --without-guile --without-lua --with-libxml
 make
 make install
-make check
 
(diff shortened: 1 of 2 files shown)


Cashalow:
Do you think this is okay to be merged @bgruening (one check failed on osx) ? Thanks a lot.

bgruening:
Where do you see a failing test?

Cashalow:
In the previous commit, during the `make check` https://travis-ci.org/bioconda/bioconda-recipes/jobs/275882367

bgruening:
Oh I see, mh not sure it means that something with the OSX stuff if wrong. "FAIL: test_nw_gen.sh" ... not sure why. But I would skip the OSX builds for this or try to fix the error :(

Cashalow:
I tried a little bit but I did not manage. However I don't even know whether the previous version passed these particular tests because I did not run the `make check` before. So you prefer to make this update Linux specific only ?

bgruening:
Can you fill an issue upstream and ask if this is known?

PR #6050, opened 2017-09-15
flowsom & flowutils

cristhomas:

### DIFF
--- a/recipes/bioconductor-flowsom/build.sh
+++ b/recipes/bioconductor-flowsom/build.sh
@@ -1,0 +1,17 @@
+
+#!/bin/bash
+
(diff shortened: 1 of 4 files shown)


recipes/bioconductor-flowsom/build.sh
```diff
@@ -0,0 +1,17 @@
+
```
bgruening: Can you remove the empty lines here?
I think you are not using the latest version of the skeleton generator, which is now part of bioconda-utils. We fixed this bug imho. Sorry for the inconvenience.

cristhomas: not a pb!

PR #6053, opened 2017-09-15
Update Metasv to 0.5.4

acaprez:

Update `metasv` to 0.5.4, and add needed dependencies.

### DIFF
--- a/recipes/metasv/meta.yaml
+++ b/recipes/metasv/meta.yaml
@@ -1,16 +1,13 @@
 package:
   name: metasv
-  version: '0.4.0'
-
+  version: '0.5.4'
 source:
-  #fn: metasv-0.3.2a0.tar.gz
-  #url: https://github.com/bioinform/metasv/archive/0.3.2a0.tar.gz
-
-  fn: metasv-e499f09.tar.gz
-  url: https://github.com/bioinform/metasv/archive/e499f09.tar.gz
+  fn: 0.5.4.tar.gz
+  url: https://github.com/bioinform/metasv/archive/0.5.4.tar.gz
+  md5: de2e21ac4f86bc4d1830bdfff95d8391
 
 build:
-  number: 4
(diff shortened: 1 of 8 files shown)


recipes/metasv/meta.yaml
```diff
@@ -1,32 +1,31 @@
...
 
 requirements:
   build:
     - python
     - setuptools
     - cython
-    - pybedtools
-    - pyvcf
-    - pysam
+    - pybedtools ==0.6.9
```
bgruening: Is this strict pinning really needed?

recipes/pybedtools/0.6.9/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  build:
+    - python
+    - six
+    - samtools
+    - setuptools
+    - pandas
+    - cython
+    - gcc  # [linux]
+    - llvm # [osx]
+    - zlib
```
bgruening: zlib need to be pinned, but I'm wondering if it is needed, because the runtime dependency seems to be not needed...

recipes/metasv/meta.yaml
acaprez: For the versions, I'm following what the authors say to use under "System Requirements" here:  http://bioinform.github.io/metasv.

recipes/pybedtools/0.6.9/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - python
+    - six
+    - samtools
+    - setuptools
+    - pandas
+    - cython
+    - gcc  # [linux]
+    - llvm # [osx]
+    - zlib {{CONDA_ZLIB}}*
+
```
acaprez: The package does include a couple of shared libraries that are linked against zlib, so I went ahead and added the pinning for the build/run reqs.

PR #6062, opened 2017-09-18
spades - on osx compile from source

notestaff:

### DIFF
--- a/recipes/spades/meta.yaml
+++ b/recipes/spades/meta.yaml
@@ -5,29 +5,26 @@ package:
   version: {{ version }}
 
 source:
-  fn: SPAdes-{{ version }}-Darwin.tar.gz  # [osx]
-  url: http://cab.spbu.ru/files/release{{ version }}/SPAdes-{{ version }}-Darwin.tar.gz  # [osx]
-  sha256: 9e6accf2258c379436e92f37f920e4dc8fa4a2da9a9813215709a9723bf15d87  # [osx]
-  fn: SPAdes-{{ version }}.tar.gz  # [linux]
-  url: http://cab.spbu.ru/files/release{{ version }}/SPAdes-{{ version }}.tar.gz  # [linux]
-  sha256: 308aa3e6c5fb00221a311a8d32c5e8030990356ae03002351eac10abb66bad1f  # [linux]
+  fn: SPAdes-{{ version }}.tar.gz
+  url: http://cab.spbu.ru/files/release{{ version }}/SPAdes-{{ version }}.tar.gz
+  sha256: 308aa3e6c5fb00221a311a8d32c5e8030990356ae03002351eac10abb66bad1f
 
 build:
-  number: 1
-  string: py{{CONDA_PY}}_zlib{{CONDA_ZLIB}}_{{PKG_BUILDNUM}}  # [linux]
+  number: 2
+  string: py{{CONDA_PY}}_zlib{{CONDA_ZLIB}}_{{PKG_BUILDNUM}}
 
 requirements:
   build:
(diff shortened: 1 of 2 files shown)


recipes/spades/meta.yaml
```diff
@@ -5,29 +5,26 @@ package:
...
-  sha256: 308aa3e6c5fb00221a311a8d32c5e8030990356ae03002351eac10abb66bad1f  # [linux]
+  fn: SPAdes-{{ version }}.tar.gz
+  url: http://cab.spbu.ru/files/release{{ version }}/SPAdes-{{ version }}.tar.gz
+  sha256: 308aa3e6c5fb00221a311a8d32c5e8030990356ae03002351eac10abb66bad1f
 
 build:
-  number: 1
-  string: py{{CONDA_PY}}_zlib{{CONDA_ZLIB}}_{{PKG_BUILDNUM}}  # [linux]
+  number: 2
+  string: py{{CONDA_PY}}_zlib{{CONDA_ZLIB}}_{{PKG_BUILDNUM}}
```
epruesse: @bgruening Is this this necessary? zlib could just be pinned, and python should be automatic I believe?

recipes/spades/meta.yaml
```diff
@@ -5,29 +5,26 @@ package:
...
 
 requirements:
   build:
-    - gcc >=4.8.2  # [linux]
-    - cmake >=2.8.12  # [linux]
-    - zlib {{CONDA_ZLIB}}*  # [linux]
-    - bzip2 1.0.*  # [linux]
+    - gcc >=4.8.2
+    - cmake >=2.8.12
+    - zlib {{CONDA_ZLIB}}*
```
epruesse: pin to 1.2.8

recipes/spades/meta.yaml
```diff
@@ -5,29 +5,26 @@ package:
...
+    - bzip2 1.0.*
 
   run:
     - python
-    - libgcc  # [linux]
```
epruesse: libgcc only needed for linux

recipes/spades/meta.yaml
```diff
@@ -5,29 +5,26 @@ package:
...
-    - zlib {{CONDA_ZLIB}}*  # [linux]
-    - bzip2 1.0.*  # [linux]
+    - libgcc
+    - zlib {{CONDA_ZLIB}}*
```
epruesse: pin to 1.2.8

epruesse:
Something funky is going on here: https://travis-ci.org/bioconda/bioconda-recipes/jobs/276664014#L1341

Looks like some config header generated by cmake had issues. This would need to be investigated locally on a Linux box.

recipes/spades/meta.yaml
tomkinsc: @epruesse: zlib is pinned correctly here per the [bioconda guidelines](https://bioconda.github.io/guidelines.html#c-c), via the value of the current environment variable defined in [env_matrix.yml](https://github.com/bioconda/bioconda-recipes/blob/master/scripts/env_matrix.yml). Explicitly pinning the version of zlib in the recipe would make mass rebuilds more difficult under a new version of zlib.

recipes/spades/meta.yaml
tomkinsc: Same as above, zlib is pinned correctly here via the environment variable.

epruesse: Sorry, must have missed that you did pin it. 

Shouldn't you use `{{CONDA_BZIP2}}` instead of `1.0.*` as well? IIRC it's in `env_matrix`.

johanneskoester:

PR #6073, opened 2017-09-18
Update BLAST recipe

epruesse:
- build on OSX as well
 - reduce package size
   - build and use blast dynamic libraries
   - use dynamic external libs (zlib, bzip2, openssl, boost)

NOTE: bioconda-utils guidelines.rst needs to be changed if/when this is merged


Cashalow:
I am working on adding gnuTLS on conda-forge so that it can be used by blast for remote blast. I think my PR should be merged relatively soon (https://github.com/conda-forge/staged-recipes/pull/3884), do you think it would be possible to add the gnutls dependency in your PR as soon as its done ?

recipes/blast/build.sh
```diff
@@ -2,25 +2,39 @@
...
+    --without-autodep \
+    --without-makefile-auto-update \
+    --with-flat-makefile \
+    --without-caution \
+    --without-dbapi \
+    --without-lzo \
+    --with-hard-runpath \
+    --without-debug \
+    --with-strip \
+    --without-downloaded-vdb \
```
Cashalow: `--without-vdb` instead of `--without-downloaded-vdb`

Cashalow:
I am moving forward now that gnutls has been merged into conda-forge. How do you want to proceed ? Do you get why we get this error ?

`blastn: error while loading shared libraries: libcom_err.so.3: cannot open shared object file: No such file or directory`

recipes/blast/build.sh
epruesse: thx

epruesse:
I think `--without-krb5` should fix the `libcom_err.so.3` issue. `configure` decided that kerberos was around, but I don't think we need/want that.

Re gnutls: Why is gnutls preferable over openssl? They do the same thing and openssl is more commonly used (well, gnutls is new) in conda-forge. I understand the project was created because the openssl license used to be GPL incompatible. It supposedly is a little faster, but that shouldn't matter for blast.

Cashalow:
Ok good to know for krb5!

For the remote option, blast needs gnutls, it can't do it only with openssl (https://www.ncbi.nlm.nih.gov/books/NBK409171/)

epruesse:
Huh. Ok. Well, the Debian `ncbi-blast` package also uses gnutls. So we'll use that. Maybe upstream didn't implement everything for the openssl API.

Cashalow:
Maybe openssl is needed for some other thing and should be left? I have no idea though.

epruesse:
Technically they do the same thing. Plus https://packages.debian.org/sid/ncbi-blast+ doesn't mention it.

epruesse:
Gnutls for OSX hasn't built yet. If that repo loses the race for travis resources, the osx build here will fail.

Cashalow:
Ok perfect thanks! Let's hope the conda forge repo builds on time.

epruesse:
@daler Why was blast blacklisted? Is it ok to unblacklist? What was broken with it?

Cashalow:
gnutls on osX has been built on conda forge, the job can be restarted now.

bgruening:
restarted!

Cashalow:
Can you add 
```--with-nettle=$PREFIX```
please ?

there is something wrong in the osX build with regards to gnutls 

(```ld: file not found: @rpath/libgnutls.30.dylib for architecture x86_64```)

I am trying to fix this.

**Comment by Cashalow** [comment]
I have to admit I am a bit stuck... Did you also try to resolve the osX error ?

**Comment by epruesse** [comment]
@Cashalow Which error? I'm guessing that the gnutls error on travis was spurious.

**Comment by epruesse** [comment]
Now it's failing OSX because of 
```
dyld: Library not loaded: /anaconda/conda-bld/blast_1506727460060/work/ncbi-blast-2.6.0+-src/c++/ReleaseMT/lib/libwritedb.dylib
  Referenced from: /anaconda/conda-bld/blast_1506727460060/_t_env/bin/makeblastdb
  Reason: image not found
```

Might have to try setting the rpath explicitely. Curiously, that part is working fine on my laptop...

**Comment by Cashalow** [comment]
A new version of Blast go out and it was added but without the dynamic libraries https://github.com/bioconda/bioconda-recipes/pull/6282 :( 

I'm trying myself to build properly the new version but still failing... https://github.com/Cashalow/bioconda-recipes/tree/blast_2.7. 

Are you still dealing with this matter ? I can take it over but I am mostly interested by the Linux version, if I don't manage with the osX I might leave it the old way for the osX build...

**Comment by peterjc** [comment]
I've managed to avoid building BLAST+ myself for years now, in favour of using the NCBI provided binaries - and was perfectly happy using them directly in the BioConda recipe. Clearly you have a use case not covered by this.

**Comment by Cashalow** [comment]
My use case is using the -remote option which has been needing gnutls for over a year now https://www.ncbi.nlm.nih.gov/books/NBK131777/#_Blast_ReleaseNotes_BLAST_2_5_0_September_ 

Using blast 2.6 from bioconda the remote function fails, prompting for GNUTLS, and I assume it is going to be the same once the 2.7 hits the could.

**Comment by peterjc** [comment]
Ah, and the NCBI say on https://www.ncbi.nlm.nih.gov/books/NBK409171/

> Starting with BLAST+ 2.5.0, the usage of the –remote option requires the GNUTLS development libraries (http://www.gnutls.org/) to make a secure connection to NCBI. The pre-compiled Linux and MacOS binaries link these statically and the Windows binaries include the required DLLs.
> 
> If you are compiling the sources, please be sure your system has the appropriate dependencies installed before building the BLAST+ applications.

Is it theoretically possible to use the NCBI provided binaries with BioConda provided gnutls libraries, perhaps with some binary editing to adjust the paths?

**Comment by epruesse** [comment]
My use case is space. The upstream provided binaries are statically linked and huge. 700MB for BLAST is a little over the top.

**Comment by epruesse** [comment]
These actually built fine on OSX. The catch was also not a timeout. In the last build, there had been something off with the library paths. Curiously, it had worked in my docker...

**Comment by nsoranzo** [comment]
Ops, I didn't noticed `blast` was in `build-fail-blacklist` and no 2.7.0 was actually ever built, sorry about that.

**Comment by epruesse** [comment]
This PR modifies the blacklist as well. So the 2.7 binary based recipe didn't actually make it yet, did it?

**Comment by nsoranzo** [comment]
> So the 2.7 binary based recipe didn't actually make it yet, did it?

No 2.7.0 on anaconda.org yet.

**Comment by peterjc** [comment]
RE:

> So the 2.7 binary based recipe didn't actually make it yet, did it?

What recipe are you referring to? The current BioConda BLAST+ recipe compiles from source on 64-bit Linux but takes the NCBI provided macOS binaries as is https://github.com/peterjc/bioconda-recipes/blob/master/recipes/blast/build.sh (true for at least 2.6.0 and with my changes for 2.7.0 as well).

Following #6282, I would have expected this package for BLAST+ 2.7.0 to appear in Anaconda - but it seems due to #6119 ``blast`` is currently black-listed along with a bunch of other stuff while dealing with other package issues? @nsoranzo @bgruening  Do we need to remove BLAST from the black-list?

**Comment by epruesse** [comment]
@peterjc Yes, BLAST can be unblacklisted (as done in this PR).

**Comment by peterjc** [comment]
Given @Cashalow wrote:

> I am working on adding gnuTLS on conda-forge so that it can be used by blast for remote blast.

Cross reference *blast remote fails because GNUTLS is missing*I #5539

**Comment by epruesse** [comment]
@peterjc Do you mind if I add your four humans blast test case to this recipe?

**Comment by peterjc** [comment]
Sure - most of my BLAST test cases are small extracts of public datasets anyway.

I assume you mean the one giving a segmentation fault in BLAST+ 2.7.0 as mentioned on #6290? The NCBI are looking into this.

**Comment by Cashalow** [comment]
There's a linting problem on the boost pining since @bgruening commit preventing the recipe from being built (the boost pining is not recognized anymore) -- is there a change on linting specifications coming soon ?

**Comment by nsoranzo** [comment]
@Cashalow The linting problem is not connected to @bgruening changes, the issue is https://github.com/bioconda/bioconda-utils/issues/190.

**Comment by Cashalow** [comment]
@nsoranzo okay thanks !

**Comment by nsoranzo** [comment]
Now that https://github.com/bioconda/bioconda-utils/pull/191 has been merged, linting passes again. But in the mean time NCBI completely removed the 2.7.0 release from FTP archives and even release notes.
Should we re-target this to 2.6.0 or wait for the next release?

**Comment by peterjc** [comment]
The BLAST team confirmed the https://blastedbio.blogspot.co.uk/2017/10/blast-270-segmentation-fault-with-html.html segmentation fault by email on 12 October 2017, and hope to have a fix out this week - so any day now...

I'd wait for the 2.7.0 fix (whatever they call it).

**Comment by peterjc** [comment]
I've asked by email as the original target date for the 2.7.0 fix has passed, and 2.7.0 has been removed from the NCBI FTP site.

**Comment by peterjc** [comment]
The change log is not up yet, but BLAST+ 2.7.1 is now on the FTP site at https://ftp.ncbi.nlm.nih.gov/blast/executables/blast+/2.7.1/

File: recipes/blast/build.sh [hunk 2]
```diff
@@ -1,26 +1,59 @@
...
+if test x"`uname`" = x"Linux"; then
+    # only add things needed; not supported by OSX ld
+    LDFLAGS="$LDFLAGS -Wl,-as-needed"
+fi
 
-    cd $SRC_DIR/c++/
-    # --with-hard-runpath is needed otherwise BLAST programs would search
-    # libraries first in the directories defined by the LD_LIBRARY_PATH
-    # environment variable, instead of using the rpath specified by conda
-    ./configure --prefix=$PREFIX --with-hard-runpath
```
nsoranzo: If you don't mind waiting a bit, I'd like to test if this is really not needed any more and maybe add a test to prevent future regressions.

For reference, this was needed because on my cluster the `$LD_LIBRARY_PATH` environment variable is set and blast programs were picking up libstdc++.so.6 from there, resulting in errors like:
```
makeblastdb: /PATH/IN/MY/LD_LIBRARY_PATH/libstdc++.so.6: version `GLIBCXX_3.4.15' not found (required by makeblastdb)
```

nsoranzo: As I was worried, the test case I've added fails without `--with-hard-runpath`, I'm going to re-add it, test it and push here, if all goes well.

epruesse: I had already changed it to `--with-runpath` ...

epruesse:
@nsoranzo wrote:
> If --with-hard-runpath configure option is not used, BLAST programs search libraries first in the directories defined by the LD_LIBRARY_PATH environment variable, instead of using the rpath specified by conda.

That is desired behavior. It's the point of LD_LIBRARY_PATH. I've removed the test again.

recipes/blast/build.sh
epruesse: Ugh. I don't know about this. That sounds like a broken cluster setup. A better way to get blast to ignore LD_LIBRARY_PATH is to remove the broken setting: `unset LD_LIBRARY_PATH; export LD_LIBRARY_PATH`.

epruesse:
@bgruening @nsoranzo 

What is the "Bioconda-Way" for RPATH vs RUNPATH? Should LD_LIBRARY_PATH override the libraries used by Bioconda packages (i.e. using RUNPATH) or should Bioconda packages have hard-coded paths set with RPATH that cannot be overridden with LD_LIBRARY_PATH?

I assumed that conda's path fixing was using DT_RUNPATH, but apparently it's using DT_RPATH and ignoring the latter?

nsoranzo:
> That is desired behavior. It's the point of LD_LIBRARY_PATH. I've removed the test again.

LD_LIBRARY_PATH is not supposed to have influence over conda-compiled binaries. In fact, blast is the only C++ package having problems on my cluster.

It took me quite a bit of work to come out with the fix and test you've reverted again. Can you please restore both?

epruesse:
It took me quite a while to compile a working BLAST on both OSX and Linux. The `--with-hard-runpath` option was the culprit for the linking issues, actually. So let's see if the package finally works now. If it does, we can still add more tests and fixes.

epruesse:
(The test would also need to use DYLD_LIBRARY_PATH on OSX and make sure it doesn't use /bin/sh or /bin/bash so SIP doesn't remove the variable.)

nsoranzo:
> The --with-hard-runpath option was the culprit for the linking issues, actually.

I don't think that's the case, since after re-adding it still works and in commit https://github.com/bioconda/bioconda-recipes/pull/6073/commits/bc828f4f498c27d7ebcb282ff0ec2f78b7e95571 the fix was actually adding `--with-runpath=$LIB_INSTALL_DIR`, which can be used together with `--with-hard-runpath` (they map to different `ld` options).

> The test would also need to use DYLD_LIBRARY_PATH on OSX

`--disable-new-dtags` has effect only on ELF binaries, so OSX is not affected, as you probably noticed already.

Anyway, thanks a lot for working on this :+1:

epruesse:
Yes, after finding and inspecting `configure.ac`, I learned that the `--with-hard-runpath` option is not alternative to the `--with-runpath` option but a separate switch disabling `--enable-new-dtags` on Linux. Hence the fix in bc828f4.

I don't think there is a way to disable the effect of `DYLD_LIBRARY_PATH` - if it's not removed by SIP it's always honored. But then on OSX the use of `*LD_LIBRARY_PATH` isn't as widespread as on Linux.

Sorry about the revert. I was annoyed that the build which should have finally passed (the pin linting broke a few builds) had a new test breaking it and no fix to go with that. The package was much more tedious than I'd anticipated. 

I do see the trouble when using `module`.

Will you review? I'd like to merge this ASAP. :)

epruesse:
Rebasing this...

epruesse:
Ok. Ready to merge I think.

nsoranzo:
Thanks a bunch @epruesse!

peterjc:
This has been an epic effort - thanks everyone 👍

PR #6078, opened 2017-09-19
Update: pysam, bcbio

### LINKED ISSUE #6069: pysam issue
@chapmanb @daler @bgruening 

Recipe appears to be busted on python3 versions: https://github.com/pysam-developers/pysam/issues/539

I think the recipe needs to add a python version to the recipe name?  (https://github.com/bioconda/bioconda-recipes/blob/master/recipes/pysam/meta.yaml#L15)  Does that seem to be the problem?

chapmanb:
- pysam -- fix py3 issue with missing CONDA_PY in build string.
  Fixes #6069 Fixes pysam-developers/pysam#539
- bcbio -- add h5py dependency for pizzly usage

### DIFF
--- a/recipes/bcbio-nextgen/meta.yaml
+++ b/recipes/bcbio-nextgen/meta.yaml
@@ -3,16 +3,16 @@ package:
   version: '1.0.6a0'
 
 build:
-  number: 1
+  number: 2
   skip: True # [not py27]
 
 source:
   #fn: bcbio-nextgen-1.0.5.tar.gz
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
Thanks @chapmanb!

Does anyone know if we have any example that are this complicated in the docs for the env version metaprogramming?  I couldn't find one during my brief look...

chapmanb:
Kyle -- I'm not sure. I don't have a good handle on the magic that happens in the build string unfortunately. My approach is to grep for related strings in the recipe directory and use an existing example.

PR #6081, opened 2017-09-19
Squeakr

rob-p:

### DIFF
--- a/recipes/squeakr/meta.yaml
+++ b/recipes/squeakr/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "squeakr" %}
 {% set version = "0.5" %}
-{% set md5 = "848f9cfb6cee5f192eed14f2fae8d5ae" %}
+{% set md5 = "cb71e38003cb96bc61b55817d33c15e7" %}
 
 package:
   name : {{ name|lower }}
--- a/recipes/squeakr/meta.yaml
+++ b/recipes/squeakr/meta.yaml
@@ -1,0 +1,34 @@
+{% set name = "squeakr" %}
+{% set version = "0.5" %}
+{% set md5 = "848f9cfb6cee5f192eed14f2fae8d5ae" %}
+
+package:
+  name : {{ name|lower }}
+  version : {{ version }}
+
+source:
+  url : https://github.com/splatlab/squeakr/archive/v{{ version }}.tar.gz
+  fn : {{ name|lower }}_{{ version }}.tar.gz
+  md5 : {{ md5 }}
+
+build:
+  number : 1
+
+requirements:
+  build:
+    - bzip2
+    - zlib
+    - boost
+  run:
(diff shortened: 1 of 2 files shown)


rob-p:
I'm getting the following failure during the linux build (but the OSX build works fine):

```
squeakr-count: /usr/lib64/libstdc++.so.6: version `GLIBCXX_3.4.15' not found (required by squeakr-count)
```

Any idea what might be causing this?  The executables are built successfully, but the test fails.

recipes/squeakr/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - llvm  # [osx]
+    - bzip2
+    - zlib
+    - boost 
+    - openssl
+  run:
+    - bzip2
+    - zlib
+    - boost 
+    - openssl
```
bgruening: You need to add `- libgcc  # [not osx]` as runtime dependency.

recipes/squeakr/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+requirements:
+  build:
+    - gcc  # [not osx]
+    - llvm  # [osx]
+    - bzip2
+    - zlib
+    - boost 
+    - openssl
+  run:
+    - bzip2
```
bgruening: Any cou probably want to pin zlib, boost and bzip2 as well.

recipes/squeakr/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  name : {{ name|lower }}
+  version : {{ version }}
+
+source:
+  url : https://github.com/splatlab/squeakr/archive/v{{ version }}.tar.gz
+  fn : {{ name|lower }}_{{ version }}.tar.gz
+  md5 : {{ md5 }}
+
+build:
+  number : 1
```
bgruening: build numbers can start with 0

recipes/squeakr/meta.yaml
rob-p: What's the easiest way to do that --- is there a "current version" variable I can use, or should I hard code the versions?

bgruening:
```
- zlib {{CONDA_ZLIB}}*
```
https://github.com/bioconda/bioconda-recipes/blob/b1a58891842d9f7f7945b20f5d4701a435879a51/recipes/pybedtools/0.6.9/meta.yaml

and boost:
`- boost {{CONDA_BOOST}}*`

bgruening: I will include a variable for bzip2 as well.

recipes/squeakr/meta.yaml
rob-p: looks like this does the trick.  I will push the fix as soon as the `{{CONDA_BZIP2}}` variable is merged.

Issue #5842, opened 2017-08-23
Set LC_ALL and LANG env vars in extended base image


ewels:
See previous discussion at #5541, #5369, #5249, #4811 and others.

#### TL;DR;
The `click` package and others require a locale to be set to work properly with Python 3. The normal busybox image doesn't support locales, but the extended-base image does. However, they appear to either not be set, or something strange is happening with them.

After a bunch of trial and error, the MultiQC build seems to pass with the following setup:

```yaml
commands:
  - LC_ALL=C.UTF-8 LANG=C.UTF-8 multiqc --version # [linux]
  - LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 multiqc --version # [osx]
extra:
  container:
    extended-base: true
```

It would great if the locale settings are set so that `LC_ALL` and `LANG` don't have to be set by the recipes themselves.

Note that linux and OSX require different locale strings to be set currently.

Phil

daler:
fixed in bioconda/bioconda-extended-base-image#3

PR #6094, opened 2017-09-21
Downgrade networkx dependency

vladsaveliev:

Networkx has recently been updated to 2.0, which is incompatible with `colormath` `spectra` dependency (see https://github.com/ewels/MultiQC/pull/561 and 
https://github.com/ewels/MultiQC/issues/592). Until a new `colormath` version is released, we have to downgrade `networkx` dependency for MultiQC.

cc @ewels

### DIFF
--- a/recipes/multiqc/meta.yaml
+++ b/recipes/multiqc/meta.yaml
@@ -25,6 +25,7 @@ requirements:
     - requests
     - simplejson
     - spectra
+    - networkx <2
 
   run:
     - python
@@ -39,6 +40,7 @@ requirements:
     - requests
     - simplejson
     - spectra
+    - networkx <2
     - libgcc  # [linux]
 
 test:


vladsaveliev:
@ewels, travis still fails because of that Click error here https://github.com/bioconda/bioconda-recipes/pull/5541 and here https://github.com/bioconda/bioconda-recipes/pull/5888, but as far as I understand it's just Travis related issue and the recipe looks good. Do you have any suggestions how we proceed and get this and #5888 merged?

ewels:
Umm, hope that someone from @bioconda/core can help and come up with a long term solution where the locale is set in the busybox..? ;)

But otherwise I guess our only option is to go back to skipping the python 3 build:

```yaml
build:
  skip: true  # [py3k]
```

ewels:
ps.. Other option could possibly be to just remove all tests? Then at least it should build for people using conda versions on their systems, even if the docker containers are broken. Thoughts @bioconda/core ?

bgruening:
@ewels I will have finished the R3.4 migration soon and will have a look at it.
Sorry that it takes so long.

daler:
@bgruening @vladsaveliev @ewels I think the issue is that we have 3 docker containers in play which have mutually incompatible locale settings.

`conda build` is run on `condaforge/linux-anvil`. This has `en_US.UTF-8` but does not have `C.UTF-8`:
```
docker run -it condaforge/linux-anvil
(root) [root@25975080fb60 /]# export LC_ALL=C.UTF-8
bash: warning: setlocale: LC_ALL: cannot change locale (C.UTF-8): No such file or directory
bash: warning: setlocale: LC_ALL: cannot change locale (C.UTF-8)

(root) [root@123012ab5f3d /]# export LC_ALL="en_US.UTF-8"  # works fine

```

`mulled-build` by default uses busybox, which has no locale installed.

The extended-base container has `C.UTF-8` but not `en_US.UTF-8`:

```
docker run -it bitnami/minideb:jessie
root@cf34d7a492d4:/# LC_ALL=C.UTF-8 # works fine
root@cf34d7a492d4:/# export LC_ALL=en_US.UTF-8 
bash: warning: setlocale: LC_ALL: cannot change locale (en_US.UTF-8): No such file or directory
```

I think this should fix it: https://github.com/bioconda/bioconda-extended-base-image/pull/3

xrefs #5842, #5541, #5369, #5249, #4811

daler:
Whoo!

vladsaveliev:
That's awesome. Thanks for sorting it out!

ewels:
Fantastic! Python3 conda builds again! 🎉  Awesome stuff @daler, thanks!

PR #6126, opened 2017-09-24
r-aroma.* packages

daler:

update r-aroma.core and r-aroma.affymetrix; remove them plus `bioconductor-repitools` (which depends on them) from blacklist


bgruening:
:) updating is always a good solution! Thanks.

Issue #6122, opened 2017-09-24
post R 3.4.1 migration cleanup


daler:
As described in #6119, and mostly thanks to @bgruening, we are now building for R 3.4.1 and the majority of packages can be successfully built. However there still some remaining packages that are having trouble building. There does not seem to be a single solution to fix these, so it will take some time and effort to resolve problems on each package. 

The blacklist is here, and includes comments about the problems with each recipe:
https://github.com/bioconda/bioconda-recipes/blob/master/build-fail-blacklist. The goal is to have an empty blacklist.

This issue can serve as a central place to discuss and coordinate the work on fixing these packages.

daler:
http://www.aroma-project.org/ says there was an update on 2017-09-13 addressing an issue with the `future` package. I'm starting to play around with updates to the various aroma packages and possibly the `future` package as well to see if that will resolve the issues.

daler:
(xref #6126)

daler:
For abyss, there's known incompatibility between older versions of abyss and later versions of boost: https://github.com/bcgsc/abyss/issues/130

sebastian-luna-valero:
Hi @daler 

Congratulations for your great work!

Could you please confirm whether [peakranger](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/peakranger) has been migrated to R 3.4.1?

It only has `r-base` as runtime dependency.

Best regards,
Sebastian

daler:
It must be a boost or zlib version that's restricting it to R 3.3.2. I'll rebuild in a new PR.

daler:
@sebastian-luna-valero see #6174

sebastian-luna-valero:
Brilliant, thank you very much again!

rvalieris:
hello @daler and @bgruening , I just noticed that a few packages are updated on the bioconda repo but not on the recipe, examples:

bioconductor-iranges is 2.8.2 on the recipe and 2.10.5 on the [repo](https://anaconda.org/bioconda/bioconductor-iranges/files)
bioconductor-biocparallel is 1.6.6 on the recipe and 1.10.1 on the [repo](https://anaconda.org/bioconda/bioconductor-biocparallel/files)
there is probably more like this, I haven't check too many yet.

anybody know whats going on here ? should I update these recipes anyway or leave it ?

daler:
[...]

PR #5108, opened 2017-07-04
update spp; highlight boost and icu incompatibility

daler:
* [ ] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [ ] This PR adds a new recipe.
* [x] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

PR #6127, opened 2017-09-24
rm r-spp from blacklist

daler:

Attempting rebuild of r-spp on osx. xref #5108 and #6122

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -33,19 +33,12 @@ recipes/bioconductor-pcaexplorer
 #blocked by https://github.com/conda-forge/staged-recipes/pull/4038
 recipes/bioconductor-monocle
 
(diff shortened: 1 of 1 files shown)


daler:
@yesimon you had helped on the last `r-spp` in #5108, any suggestions for this one?

daler:
(upstream issue is https://github.com/hms-dbmi/spp/issues/12)

yesimon:
My guess is that the failure is linking to boost libraries using `libc++` which is incompatible with compiling spp with `libstdc++`. However, `r-spp` is only using `boost/tokenizer` which is a header only library, so I think the trick is to get autotools to compile with boost header only instead of linking the shared library.

johanneskoester:

PR #6131, opened 2017-09-24
build abyss 2.0.2

daler:

xref #6122 

@bioconda/core I have some questions here:

**Earlier versions of ABYSS:** ABYSS versions <2.0.2 do not work with later version of boost (https://github.com/bcgsc/abyss/issues/130). It looks like the preprocessing selectors are processed before even jinja templating, so we can't do things like `skip: True # [ float({{ CONDA_BOOST }}) < 1.62]`. One solution would be to pin boost (ignoring CONDA_BOOST) in these older recipes; another would be to keep them in the blacklist. Another would be to remove the recipes. I don't know what the demand is for earlier versions of ABYSS to know whether they're worth rebuilding on earlier boost. Here I'm taking the path of least disruption and leaving them in the blacklist.

**Wrapping a makefile:** I updated the main recipe to ABYSS 2.0.2 but this is failing in the mulled-build test. The reason for failure is that `abyss-pe` is a Makefile which has a shebang line of `#!/usr/bin/make -rRf`, but in the container `make` lives at `/usr/local/bin/make`. We can't change the shebang line to `#!/usr/bin/env make -rRf` because of the extra arguments. Would a wrapper script work to solve this? The `abyss-pe` Makefile is over my head; I don't know about the implications of a wrapper script.

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -6,9 +6,14 @@ recipes/mothur/1.36.1
 # recipes/metaprob
 
 # more temporary blacklisting to get other packages to run
-recipes/abyss
+
+# Older versions of abyss need older versions of boost
+# https://github.com/bcgsc/abyss/issues/130
+# recipes/abyss
+recipes/abyss/1.5.2
 recipes/abyss/1.9.0
 recipes/abyss/2.0.1-k128
+
 recipes/augustus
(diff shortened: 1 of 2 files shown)


druvus:
fixed in #6589

PR #6139, opened 2017-09-25
[WIP] Trans-Proteomic Pipeline packaging

jvolkening:

@bgruening Do you want to review this and let me know what you think? It was a pain but things seem to finally be working at this point.


recipes/tpp/build.sh
```diff
@@ -0,0 +1,62 @@
...
+#export HDF5_SUPPORT = 1
+#export MZ5_SUPPORT  = 1
+
+echo "INSTALL_DIR  = ${PREFIX}"      >  site.mk
+echo "TPP_BASEURL  = /tpp"           >> site.mk
+echo "TPP_DATADIR  = ${PREFIX}/data" >> site.mk
+
+# install the dependencies
+
+cpanm FindBin::libs
```
bgruening: This should be a conda dependency as well.

recipes/tpp/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  url: https://downloads.sourceforge.net/project/sashimi/Trans-Proteomic%20Pipeline%20%28TPP%29/TPP%20v5.0%20%28Typhoon%29%20rev%200/TPP_5.0.0-src.tgz
+  fn: {{ name|lower }}_{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+  patches:
+    - tpp5.0.0.patch
+
+build:
+    number: 0
+    skip: true  # [osx]
+    skip: true  # [win]
```
bgruening: Bioconda is not supoprting windows builds, so this can be skiped. If you want to skip two things, the correct way would be `# [osx or win]`.

recipes/tpp/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+        - zlib
+        - bzip2
+        - perl-extutils-makemaker
+        - perl-app-cpanminus
+    run:
+        - libgcc <=4.9 # [linux]
+        - perl
+        - perl-xml-parser
+        - hardklor
+        - zlib
```
bgruening: can you please use `- zlib {{CONDA_ZLIB}}*`

recipes/tpp/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+        - bzip2
```
bgruening: and here, as well as in the build section, `bzip2 1.0.*`

recipes/tpp/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+        - perl-xml-parser
+        - zlib
+        - bzip2
+        - perl-extutils-makemaker
+        - perl-app-cpanminus
+    run:
+        - libgcc <=4.9 # [linux]
+        - perl
+        - perl-xml-parser
+        - hardklor
```
bgruening: hardklor is now a dependency? Interesting!

recipes/tpp/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+        - zlib
+        - bzip2
+test:
+    files:
+        - test_data
+        - assert.sh
+
```
bgruening: can you please add some `commands`-test. Simply call some binary and get the version or the help back.

recipes/tpp/build.sh
jvolkening: Okay, I will have to package it separately

recipes/tpp/meta.yaml
jvolkening: It is wanted/needed by one of the TPP programs. Rather than let TPP build it itself, possibly overruling a version installed by conda or elsewhere, it seemed better to just add it as a dependency.

recipes/tpp/meta.yaml
jvolkening: The only problem here is that, as far as my testing goes, only one of the TPP programs returns 0 if called without data (or with `--help`, `-h`, `-v`, etc). The rest print help text but return 1, which causes the tests to fail. I will add the one working command I know of and then put the test suite into the build script as you suggested above.

jvolkening:
> The most crucial one is that we do not ship test data in bioconda. What about the following we create a tarball of the test data and store them somewhere. In your build.sh you download it and run it inside the build.sh.
> A conda recipe should only container the meta.yml and the build.sh. In addition please add commands-test to the meta.yaml. This will hopefully cover everything.

I will try to address this. Is the `run_test.sh` okay to include (seems to be, based on [this](https://bioconda.github.io/guidelines.html#tests)) if I host the test data elsewhere? I rather think the expanded test suite is important given all of the quirks and little hacks that were necessary to get various bits of the suite working.

recipes/tpp/build.sh
jvolkening: Should this go in bioconda or another conda channel? It's not specific to biology but I see other general-purpose perl packages in bioconda.

bgruening:
run_test.sh is OK, but the test imports will not work in our Docker testing..

recipes/tpp/build.sh
bgruening: BioConda is fine for the moment I think. We did not mass-migrate perl stuff to conda-forge.

recipes/tpp/meta.yaml
bgruening: Ok!

recipes/tpp/meta.yaml
bgruening: You can use the following trick: `foo --help 2>&1 | grep "Something in the help"`

johanneskoester:

bgruening:
@jvolkening do you still plan to work on this?

jvolkening:
I have a few more commits not yet pushed, mostly having to do with which binaries/scripts should and shouldn't be included in the conda package. I'm on vacation but will review your recently merged work and see if I have anything more to add.

There is also a newer version of TPP out but it requires more tweaks to the build process that I haven't had time to work on.

PR #6141, opened 2017-09-26
Updating Last (lastal) to v874

jmeppley:

Adds the 8 byte versions of of the binaries (lastdb8, lastal8, ...) to the build for indexing and searching HUGE dbs on high memory machines.


recipes/last/build.sh
```diff
@@ -20,12 +23,14 @@ last-map-probs \
 last-dotplot \
 "
 
-export CMAKE_CXX_FLAGS="-std=c++11 -stdlib=libc++"
+if [[ $SUBDIR == osx* ]]; then
```
daler: @jmeppley how does `$SUBDIR` get set?

jmeppley:
Yeah, that's a bit of a hack. It's part of the conda build process.

I was just trying to get it to compile at the time, and meant to go back to
the conda docs to look for a better way to get the os.

I'll look one more time before merging...

daler:
I didn't see `$SUBDIR` in https://conda.io/docs/user-guide/tasks/build-packages/environment-variables.html which is why I was asking, but this should do the trick without having to worry about env vars:

```bash
if [[ $(uname) == "Darwin" ]]; then
  export CMAKE_CXX_FLAGS="-stdlib=libc++"
fi
```

jmeppley:
I'll update the PR to use that instead. Thanks!

tomkinsc:
Thanks for this, lastal is normally an annoying package to update because of the 2to3 patching. If you've been using the updated version, have you encountered any issues running it under py3? Especially since it looks like the patch mostly shifts line numbers down relative to a prior patch.

jmeppley:
I honestly have never used the maf-convert tool (the one that needs the patch), so I haven't tested it beyond checking if it compiles.

I don't even use last's MAF format, so I don't have anything lying around to test it on. Have you seen problems with the latest version?

jmeppley:
I was able to test the converted python script on osx, and it seems fine to me.

tomkinsc:
No problems so far! Thanks for the update!

PR #6169, opened 2017-09-28
Recipe for "Feht"

hellothisisMatt:

Instead of using the source code, I used the pre-compiled binary executable. There seems to be a missing file in the source code that is causing errors when building it. The drawback of using the executable is that `feht` won't be usable on OSX.


hellothisisMatt:
As mentioned above, this is using the precompiled executable as there is an issue with building from source. This will be fixed in the next release of `feht` where the recipe will be updated.

bgruening:
@hellothisisMatt can you please report this issue upstream, so that they can fix it?

hellothisisMatt:
@bgruening I've let the author know should be fixed in the next iteration.

bgruening:
Thanks

PR #6195, opened 2017-10-01
adding mappy

wdecoster:

### DIFF
--- a/recipes/mappy/build.sh
+++ b/recipes/mappy/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-$PYTHON setup.py install 
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
 
 # Add more build steps here, if they are necessary.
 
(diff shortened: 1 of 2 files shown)


recipes/mappy/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+export CFLAGS="-I$PREFIX/include"
+export LDFLAGS="-L$PREFIX/lib"
+
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
+
+# Add more build steps here, if they are necessary.
```
bgruening: Can you remove the comments here.

recipes/mappy/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+build:
+  number: 0
+  skip: True  # [py27]
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - zlib
```
bgruening: can you pin zlib like this please, `- zlib {{CONDA_ZLIB}}*`

recipes/mappy/build.sh
wdecoster: Oh right, will do.

Thanks!

PR #6200, opened 2017-10-02
Samtools 0.1.x

scholtalbers:

Adds `samtools 0.1.12, 0.1.13, 0.1.14, 0.1.15` (all following a similar recipe)


scholtalbers:
I believe the osx build job should just be restarted, as it failed to download samtools 0.1.15 - which should just work: https://travis-ci.org/bioconda/bioconda-recipes/jobs/282252663

daler:
@scholtalbers I just restarted it, let's see what happens.

recipes/samtools/0.1.12/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+sed -i.bak 's/-lcurses/-lncurses/' Makefile
+sed -i.bak 's/-lz/-lz -ltinfo/' Makefile
+sed -i.bak 's/\(inline void __ks_insertsort\)/static \1/'  ksort.h
+export CFLAGS="-I$PREFIX/include"
```
epruesse: I think this should be 
```
export CFLAGS="$CFLAGS -I$PREFIX/include"
```
in case the CFLAGS should be default set to something in the future.

recipes/samtools/0.1.12/build.sh
```diff
@@ -0,0 +1,10 @@
...
+make
+mkdir -p $PREFIX/bin
+mv samtools $PREFIX/bin
+mkdir -p $PREFIX/lib
+mv libbam.a $PREFIX/lib
```
epruesse: Do we need the static lib? Could this be made a dll?

recipes/samtools/0.1.12/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - gcc
+    - zlib
+    - ncurses
+  run:
+    - libgcc
+    - zlib
+    - ncurses
+source:
+  fn: samtools-0.1.12.tar.bz2
+  md5: 500ae32fb431de3940e0e955744b9b36
```
epruesse: sha256 preferred

recipes/samtools/0.1.12/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+  summary: Tools for dealing with SAM and BAM files
+build:
+  number: 0
+package:
+  name: samtools
+  version: 0.1.12
+requirements:
+  build:
+    - gcc
+    - zlib
```
epruesse: please pin this with `{{CONDA_ZLIB}}`

recipes/samtools/0.1.12/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - ncurses
+  run:
+    - libgcc
+    - zlib
```
epruesse: same here

recipes/samtools/0.1.12/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+  license: MIT
+  summary: Tools for dealing with SAM and BAM files
+build:
+  number: 0
+package:
+  name: samtools
+  version: 0.1.12
+requirements:
+  build:
+    - gcc
```
epruesse: OSX should be compiled with native CLANG. Add a ` # [linux]` here.

recipes/samtools/0.1.12/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - zlib
+    - ncurses
```
epruesse: pin to `5.8`

recipes/samtools/0.1.12/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+  run:
+    - libgcc
+    - zlib
+    - ncurses
```
epruesse: and here

recipes/samtools/0.1.12/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+package:
+  name: samtools
+  version: 0.1.12
+requirements:
+  build:
+    - gcc
+    - zlib
+    - ncurses
+  run:
+    - libgcc
```
epruesse: only needed when compiling with GCC, so ` # [linux]`

recipes/samtools/0.1.12/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+sed -i.bak 's/-lcurses/-lncurses/' Makefile
```
epruesse: OSX `sed` does not have `-i` ... not sure why this worked at all?

epruesse:
Are these just for completeness sake, or is there an actual need for the older versions of samtools? (Sorry, don't know what changed in samtools along those versions)

recipes/samtools/0.1.12/build.sh
epruesse: Never mind. I overlooked that it just **requires** the backup extension

nsoranzo: It does have `-i`, but it forces the creation of the backup file, see https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/sed.1.html

I think this is correct.

recipes/samtools/0.1.12/meta.yaml
nsoranzo: @epruesse Shouldn't it be pinned to `{{CONDA_NCURSES}}` ?

scholtalbers:
I 'require' the samtools 0.1.12 as a galaxy wrapper depends on it. The others I included for completeness.

recipes/samtools/0.1.12/meta.yaml
scholtalbers: The md5 I just copied from sourceforge which might be good enough right?

recipes/samtools/0.1.12/meta.yaml
epruesse: Err, yes, of course.

recipes/samtools/0.1.13/meta.yaml
```diff
@@ -18,7 +18,7 @@ requirements:
     - ncurses {{CONDA_NCURSES}}*
 source:
   fn: samtools-0.1.13.tar.bz2
-  sha256: 8ab22994c9e899648a8731cdfd6b1e4b1ed8773e
```
epruesse: Looks too short to be an sha256 ;-)

recipes/samtools/0.1.12/meta.yaml
bgruening: Imho it is better to use a md5 if this is given from upstream then calculating your own checksum.

bgruening:
@epruesse does this look ok for you?

scholtalbers:
@epruesse are you waiting for me to correct anything else or is it good to go, would be great to get it merged in..

epruesse:
@scholtalbers Sorry for delay. Thanks for ping and many thanks for recipes! :)

PR #6203, opened 2017-10-02
Bioconductor 3.5 build bumps

daler:

I'm starting to migrate all bioconductor packages to Bioconductor 3.5. In doing so, I'm taking the opportunity to unify and standardize the formatting to make this easier in the future. This PR includes just those recipes that have a build number bump. That is, their version did not change and no dependencies have changed, but they now use jinja templating in URLs, filenames, and tests as well has have the specified `r-base` version as was found in their respective DESCRIPTION files.

Since these *should* build cleanly, I'm using it as a real-world test case to see how a PR into the bulk branch works.

cc @johanneskoester @bgruening


recipes/bioconductor-acme/meta.yaml
```diff
@@ -1,30 +1,34 @@
...
-    - https://depot.galaxyproject.org/software/ACME/ACME_2.32.0_src_all.tar.gz
-  md5: f99ea6b94399fd7a10f55ac7e7ec04fa
+    - 'http://bioconductor.org/packages/3.5/bioc/src/contrib/{{ name }}_{{ version
+      }}.tar.gz'
+    - 'https://bioarchive.galaxyproject.org/{{ name }}_{{ version }}.tar.gz'
+    - 'https://depot.galaxyproject.org/software/{{ name }}/{{ name }}_{{ version }}_src_all.tar.gz'
+  sha256: b3d24b5cd9dd51d7e28866a1e3e92d953ee2953593d95af5840284f70391ed59
 build:
-  number: 0
+  number: 2
```
bgruening: the bump is here one two many.

recipes/bioconductor-adsplit/meta.yaml
```diff
@@ -1,14 +1,18 @@
...
-    - http://bioconductor.org/packages/3.5/bioc/src/contrib/adSplit_1.46.0.tar.gz
-    - https://depot.galaxyproject.org/software/adSplit/adSplit_1.46.0_src_all.tar.gz
-  md5: 7638432c734c1fe458acbb9b29384f57
+    - 'http://bioconductor.org/packages/3.5/bioc/src/contrib/{{ name }}_{{ version
+      }}.tar.gz'
+    - 'https://depot.galaxyproject.org/software/{{ name }}/{{ name }}_{{ version }}_src_all.tar.gz'
+  sha256: d59c68a312b386107a2f6fc776e964b9eb0872218caebae7864601d978c8f225
 build:
-  number: 0
+  number: 2
```
bgruening: here as well

recipes/bioconductor-aldex2/meta.yaml
```diff
@@ -1,15 +1,19 @@
...
-    - https://depot.galaxyproject.org/software/ALDEx2/ALDEx2_1.8.0_src_all.tar.gz
-  md5: 979f4e257eb75248e8059a88674df0c8
+    - 'http://bioconductor.org/packages/3.5/bioc/src/contrib/{{ name }}_{{ version
+      }}.tar.gz'
+    - 'https://bioarchive.galaxyproject.org/{{ name }}_{{ version }}.tar.gz'
+    - 'https://depot.galaxyproject.org/software/{{ name }}/{{ name }}_{{ version }}_src_all.tar.gz'
+  sha256: 04c2bb73c634c3b274937dc3e7268efac9f08a93b8f0e091c7b5babeacd53610
 build:
-  number: 0
+  number: 2
```
bgruening: ok, everywhere :)

recipes/bioconductor-annaffy/meta.yaml
```diff
@@ -1,15 +1,19 @@
...
 source:
-  fn: annaffy_1.48.0.tar.gz
+  fn: '{{ name }}_{{ version }}.tar.gz'
   url:
-    - http://bioconductor.org/packages/3.5/bioc/src/contrib/annaffy_1.48.0.tar.gz
-    - https://bioarchive.galaxyproject.org/annaffy_1.48.0.tar.gz
-    - https://depot.galaxyproject.org/software/annaffy/annaffy_1.48.0_src_all.tar.gz
-  md5: c51219a222b377403fe4bc08a6c57e7f
+    - 'http://bioconductor.org/packages/3.5/bioc/src/contrib/{{ name }}_{{ version
+      }}.tar.gz'
```
bgruening: can we fix this line break here.

bgruening:
Ryan I added a few comments. Most of the packages already exists for bioc 3.5. Is it worth to bump the build number?

recipes/bioconductor-aldex2/meta.yaml
daler: whoops, forgot to `git checkout -- .` after the test. I'll fix all of these . . .

daler:
Thanks for your comments, making the changes now.

I was on the fence about adding these build bumps. I want to eventually automate the process of updating all bioconductor packages, and making them all uniform would simplify that. The question is whether that standardization of recipes is worth spending the build time now or postpone the standardization until 3.6. I could be convinced either way.

bgruening:
My vote is for 3.6. Space is rare. If we template the `{{3.5}}` we could bump this as well via a env var. This would leave us with only the checksum to be updated ...

johanneskoester:
I vote for 3.6 as well. Great work! I always thought it would be great to "hit a button" and get updated bioconductor packages. BTW we should be able to do the same for python and perl. With some additional annotation, eg something like an URL to obtain the latest version string under extra->versionurl it should be possible for other packages as well.

bgruening:
Jupp, for python and probably others as well we should coordinate with the conda-forge people. They are developing webservices for this usecase.

johanneskoester:
Can't we simply unify the recipes now, but just not bump the build number for those were no version change is entailed? Then everything is prepared for 3.6 and your plans without wasting space in our channel.

daler:
Templating `3.5` and unifying without build bumps are good ideas, I'll work on that.

As for PyPI, https://github.com/bioconda/bioconda-utils/pull/178 gets us most of the way for the bioconda-specific stuff; we can then see how that fits into conda-forge's webservices.

johanneskoester:
Great!

daler:
@johanneskoester @bgruening I've been making some judgement calls on a per-recipe basis whether or not to retain `skip: True # [osx]` if it was defined in a previous recipe version, and how to handle `gcc` and other non-R, non-bioconductor deps that were in previous recipe versions (e.g., keeping `autoconf` [here](https://github.com/bioconda/bioconda-recipes/pull/6203/commits/1f88398da39ebcab75ded65d0645091145c26f18) and keeping `r-rcpp`, `gcc` despite not being specified in newer bioconductor package, [here](https://github.com/bioconda/bioconda-recipes/pull/6203/commits/10730d8878ba6189e2c00321909b551e008244d6)).

I was aiming to strike the balance between recipes that build and recipes that have too many restrictions. But maybe instead I should comment out any extra deps, try building in `bulk`, and then uncomment as needed to get them to build again.

Thoughts?

bgruening:
I tend to be for the latter idea, with bulk.

daler:
@bgruening @johanneskoester, I've now gone through all the bioconductor recipes, commenting out any custom changes (patches, osx skips, exported vars in build scripts, extra dependencies not indicated in a package's DESCRIPTION file). As they fail, it should be straightforward to apply any fixes that were previously required by uncommenting them.

To clean up this PR, do you think I should open another one that merges in latest master branch to `bulk`? Or leave this as-is?

daler:
The recent https://github.com/bioconda/bioconda-recipes/pull/6247 is an example of deps not being listed in DESCRIPTION. I'm considering supporting an `extra: custom-bioc-deps:` field in recipes to support cases like this. For example, in #6247, the recipe could also add the following:

```yaml
extra:
  custom-bioc-deps:
    run:
      - r-statmod
```
The next time bioconductor-skeleton runs, it would automatically add `r-statmod` to the run deps of the generated recipe. That way, the next time we do a bulk update, bioconductor skeleton can read and insert custom stuff into updated recipes to avoid going through by hand each time.

While that would work for deps in meta.yaml, I'm not sure how to handle custom build script entries, other than perhaps doing the diff between existing and bioconductor-skeleton-generated build.sh scripts and then commenting out the changes in the file. 

Maybe the solution really is going through by hand each time . . . no substitute for a human brain after all.

PR #6208, opened 2017-10-03
set explicit license instead of reference to LICENSE source file

osallou:
* [ X ] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [ X ] This PR updates an existing recipe.

### DIFF
--- a/recipes/yahmm/meta.yaml
+++ b/recipes/yahmm/meta.yaml
@@ -34,6 +34,7 @@ test:
 about:
   home: http://pypi.python.org/pypi/yahmm/
   license: MIT
+  license_file: LICENSE.txt
   summary: 'YAHMM is a HMM package for Python, implemented in Cython for speed.'
 
 extra:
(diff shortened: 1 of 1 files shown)


recipes/yahmm/meta.yaml
```diff
@@ -33,7 +33,7 @@ test:
 
 about:
   home: http://pypi.python.org/pypi/yahmm/
-  license: LICENSE.txt
```
bgruening: Should this be license_file?

osallou: could be , but needs to access source. Could be added as a link to source location (http:///), don't know how you usually manage references in source files.

bgruening: LICENSE.txt is indeed in the tarball, so you can change this to license_file:

osallou: added license_info

bgruening:
Thanks @osallou!

PR #6209, opened 2017-10-03
Fix invalid license link

osallou:
Explicit set of license and add correct link to license file on github

* [ X ] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [ X ] This PR updates an existing recipe.

### DIFF
--- a/recipes/vphaser2/meta.yaml
+++ b/recipes/vphaser2/meta.yaml
@@ -7,7 +7,9 @@ package:
   name: "vphaser2"
   version: '2.0'
 source:
-  git_url: https://github.com/broadinstitute/v-phaser2.git #[osx or linux64]
+  url: https://github.com/broadinstitute/v-phaser2/archive/6b5483a8e1a9bf24061a0aa211fee18e6b845668.tar.gz  #[osx or linux64]
+  md5: fd22b1c4403903576e62defb51b6ae32
+  #git_url: https://github.com/broadinstitute/v-phaser2.git #[osx or linux64]
 build:
   number: 3
   rpaths:
@@ -23,8 +25,3 @@ requirements:
 test:
     commands:
         - 'vphaser2 | grep "BAM file"'
-extra:
-  skip-lints:
-    - uses_git_url
-    - missing_hash
-
--- a/recipes/vphaser2/meta.yaml
+++ b/recipes/vphaser2/meta.yaml
@@ -11,7 +11,7 @@ source:
   md5: fd22b1c4403903576e62defb51b6ae32
   #git_url: https://github.com/broadinstitute/v-phaser2.git #[osx or linux64]
 build:
-  number: 3
+  number: 4
   rpaths:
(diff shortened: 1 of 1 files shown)


recipes/vphaser2/meta.yaml
```diff
@@ -22,3 +23,8 @@ requirements:
 test:
     commands:
         - 'vphaser2 | grep "BAM file"'
+extra:
```
bgruening: I think it is better to use this URL: https://github.com/broadinstitute/v-phaser2/archive/6b5483a8e1a9bf24061a0aa211fee18e6b845668.tar.gz and keep the linter.

recipes/vphaser2/meta.yaml
```diff
@@ -1,6 +1,7 @@
 about:
     home: 'https://www.broadinstitute.org/scientific-community/science/projects/viral-genomics/v-phaser-2'
-    license: 'https://www.broadinstitute.org/node/4962'
+    license: 'single user license for academic non-commercial research purposes only'
+    license_file: 'https://github.com/broadinstitute/v-phaser2/blob/master/V-Phaser-2.0/license/V_Phaser_2_License_v20130317.pdf'
```
bgruening: This should point to a file in the tarball.

osallou: ok, I saw several other packages using references to github, so did the same, gonna fix this

recipes/vphaser2/meta.yaml
osallou: well, I only updated license part, I did not modify build part, but indeed refering to an archive would be better for reproducibility. I gonna update recipe to do so

osallou: added archive instead of git_url, removing linter excludes

bgruening:
This probably need a build-version bump to get a new release with the latest source.

recipes/vphaser2/meta.yaml
```diff
@@ -1,14 +1,17 @@
...
-    license: 'https://www.broadinstitute.org/node/4962'
+    license: 'single user license for academic non-commercial research purposes only'
+    license_file: 'V-Phaser-2.0/license/V_Phaser_2_License_v20130317.pdf'
     summary: 'V-Phaser 2 is a tool to call variants in genetically heterogeneous populations from ultra-deep sequence data'
 package:
   name: "vphaser2"
   version: '2.0'
 source:
-  git_url: https://github.com/broadinstitute/v-phaser2.git #[osx or linux64]
+  url: https://github.com/broadinstitute/v-phaser2/archive/6b5483a8e1a9bf24061a0aa211fee18e6b845668.tar.gz  #[osx or linux64]
```
epruesse: @bgruening Should this perhaps use cargo-port instead?

bgruening: Lets put is this way, we do backup nowadays everything on cargo-port as soon as we merge it. To prevent data loss.

PR #6231, opened 2017-10-05
Bump gclib.

cbrueffer:


recipes/gclib/meta.yaml
```diff
@@ -1,15 +1,19 @@
...
+
 package:
-  name: gclib
-  version: "0.0.1"
+  name: {{ name }}
+  version: {{ version }}
 
 build:
-    number: 4
+    number: 5
```
bgruening: I'm not sure increase the build number is correct here. We should increase the actual version.

cbrueffer:
gclib is not actually versioned in any way, it's more of a continuous development (at least as I can see). Do you see any alternative in this case?

bgruening:
we can simply increase the 0.1 version by 0.1.1 and increase the last number

johanneskoester:

PR #6236, opened 2017-10-05
Update peddy to 0.3.1.

cbrueffer:

### DIFF
--- a/recipes/peddy/build.sh
+++ b/recipes/peddy/build.sh
@@ -1,3 +1,3 @@
-#!/bin/bash
+#!/bin/sh
 
-$PYTHON setup.py install
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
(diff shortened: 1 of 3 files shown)


cbrueffer:
This update chokes on the ```nose``` package missing when running tests even though it is specified as a test requirement. The recipe builds if I add it as a build and run dependency.

@bgruening , @daler any idea what may be the issue?

daler:
Hi @cbrueffer, I'd remove the `nostests` test command, and remove `nose` from the build and run deps. We generally don't run unit tests, so it shouldn't be needed here.

The failure is because the minimal docker container test uses _only_ the package itself. Anything else (like extra test deps or files or conda itself) are purposefully excluded.

cbrueffer:
Thanks for the explanation! I'll remove the nosetest command as suggested.

PR #6247, opened 2017-10-08
bioconductor-limma - add statmod as dependency

mblue9:

In this PR I've added statmod as a dependency to limma, as discussed in the Galaxy IUC PR here https://github.com/galaxyproject/tools-iuc/pull/1471 (cc @bgruening ) and also because in the limma user guide pg 9 (here: https://www.bioconductor.org/packages/devel/bioc/vignettes/limma/inst/doc/usersguide.pdf) it states "you’ll probably want biocLite("statmod)" as well".

### DIFF
--- a/recipes/bioconductor-limma/meta.yaml
+++ b/recipes/bioconductor-limma/meta.yaml
@@ -15,7 +15,6 @@ build:
 requirements:
   build:
     - r-base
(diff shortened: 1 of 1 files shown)


mblue9:
Thanks @bgruening!

PR #6256, opened 2017-10-09
Omero 5.4.0

sbesson:

### DIFF
--- a/recipes/python-omero/meta.yaml
+++ b/recipes/python-omero/meta.yaml
@@ -26,10 +26,6 @@ test:
     - omero
     - omero.gateway
 
-extra:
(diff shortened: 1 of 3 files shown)


recipes/python-omero/meta.yaml
```diff
@@ -19,14 +19,18 @@ requirements:
...
     - python
-    - zeroc-ice
+    - zeroc-ice >3.5,<3.7
 
 test:
   imports:
     - omero
     - omero.gateway
 
+extra:
```
druvus: Instead of skipping linting it is better to update build.sh to include `$PYTHON setup.py install --single-version-externally-managed --record=record.txt`

PR #6268, opened 2017-10-10
Simwalk 2.91 Recipe

mtekman:

### DIFF
--- a/recipes/simwalk/meta.yaml
+++ b/recipes/simwalk/meta.yaml
@@ -7,14 +7,14 @@ source:
   url: https://github.com/BioTools-Tek/packages/blob/master/simwalk2/SimWalk2-2.91-lin.tar.gz?raw=true # [linux]
   md5: 2a529d50ca48cd7ea2b67f85c290f203 # [linux]
 
-  fn: SimWalk2-2.91-osx.tar.gz # [osx]
-  url: https://github.com/BioTools-Tek/packages/blob/master/simwalk2/SimWalk2-2.91-osx.tar.gz?raw=true # [osx]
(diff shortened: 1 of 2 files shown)


recipes/simwalk/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  md5: 157cf4a869afdacf8e415005886f27c7 # [osx]
+
+  
+build:
+  number: 0
+  #skip: True # [osx]
+
+requirements:
+  build:
+    - libgfortran # [linux]
```
bgruening: This means you only have dependencies for Linux and no dependencies at all for OSX.
You also want to have GCC as build dependency and not libgcc, libgcc is runtime dependency.

recipes/simwalk/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - libgcc # [linux]                                                                                                                                    
+  run:                                                                                                                                                
+    - libgfortran # [linux]
+    - libgcc # [linux]
+    - zlib # [linux]
+
+test:
+  source_files:
```
bgruening: This will not work you don't have files included here.
Can you just call `simwalk2 --help` or similar and catch maybe the expected error.

mtekman: simwalk unfortunately takes no arguments, it just looks for a `BATCH2.DAT` file which specifies where all the other input files are.

I will copy over the relevant files from the work directory, thanks

recipes/simwalk/meta.yaml
mtekman: will fix, thanks

PR #6288, opened 2017-10-11
Riboseed version 0.4.35

nickp60:


recipes/riboseed/meta.yaml
```diff
@@ -1,47 +1,47 @@
...
-  # url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
-  url: https://test-files.pythonhosted.org/packages/21/59/ef9dbb631f132c312bdd1cc145395050fdd0bfbe46765280c9a157367dcd/riboSeed-0.4.16.tar.gz
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
   '{{ hash_type }}': '{{ hash_value }}'
 
 build:
-  number: 0
+  entry_points:
+    - ribo = riboSeed.ribo:main
+  number: 1
```
bgruening: the number should start with 0 as this is a new package

recipes/riboseed/meta.yaml
```diff
@@ -51,10 +51,17 @@ requirements:
...
     - bwa ==0.7.8
     - samtools ==1.4.1
+    - bcftools ==1.5
 
 test:
   imports:
     - riboSeed
+  commands:
+    - ribo
+    - ribo run -h
```
bgruening: This seems to depend on an Xserver? Is this subcommand needed on HPC? If so you probably need to include xlibs as well.

recipes/riboseed/meta.yaml
nickp60: I have version 0 on my personal channel; is it ok to have identical versions on multiple channels?

recipes/riboseed/meta.yaml
nickp60: Thanks, I am working figuring out how to solve this.  It has to do with matplotlib importing, and a recent change should sort this out...

nickp60: @bgruening I reworked the code to use matplotlib's API instead of the pyplot commands which should get rid of this.  So far, not dice; it seems like the tests still fail.  Which xlibs would need to be included? the program doesnt open any windows, so its been difficult to troubleshoot

recipes/riboseed/meta.yaml
peterjc: I believe build numbers on different channels are not relevant

recipes/riboseed/meta.yaml
daler: @nickp60 this is from https://github.com/nickp60/riboSeed? In the master branch, I only see one case using `matplotlib.use("Agg")`. Is there a different branch to check?

nickp60: @daler The script (plotMauveBetter.py) that uses that syntax isn't actually included in the package;  all other instances that used to use that were changed  from using pyplot to the API:

    from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas

as discusses in [this post](http://www.dalkescientific.com/writings/diary/archive/2005/04/23/matplotlib_without_gui.html);  previously (before commit 46173bb), I used the other syntax:

    import matplotlib as mpl
    mpl.use('Agg')
    import matplotlib.pyplot as plt

daler: Not sure if it will help, but usually I do 
```python
import matplotlib
matplotlib.use('agg')
```
as the very first import, just in case any other modules are themselves pulling in matplotlib. The backend gets set the very first time -- and only the very first time -- matplotlib is imported so you have to catch it early.  That said, there's nothing obvious in 46173bb that would be importing it first, but it does look like riboSnag.py had that part commented out which could be causing issues.

You might even want to consider importing matplotlib once in its own module, say `_matplotlib.py`, do the backend-setting there, and then in your other modules, use `from _matplotlib import matplotlib`.



Last, that post you refer to is over a decade old, many things have changed in matplotlib since then . . . but I guess not the problem of running on a headless server!

nickp60: Thanks @daler , I like the idea of importing matplotlib in its own module, I'll give that a try!

PR #6294, opened 2017-10-12
BioTools annotations in BioConda Recipes

ypriverol:


ypriverol:
Dear BioConda core team: 

During the BioContainers hackathon, a couple of people including @hmenager @bgruening @mr-c worked in a definition to link biocontainers/bioconda packages to other initiatives such as: biotools, etc. 

The first proposal was to include the following session on each recipe: 

```
  extra:
      biotools: [amos]
```
This will be a key-value pair system that will allow us to retrieve metadata in the future around each container (see the issue in biocontainer https://github.com/BioContainers/specs/issues/84). @hmenager and @ypriverol have developed a set of scripts to automatically map more than 600 containers with the corresponding bio.tools identifier (list here: https://github.com/BioContainers/biotools-bioconda-ids/blob/master/mapping_matchonly.csv). 

Some of the recipes have been updated with more well-structured spaces/indentation because everything has been programmatically updated. Please have a look and let us know if we can move this forward.

recipes/amos/meta.yaml
```diff
@@ -44,3 +44,5 @@ requirements:
     - perl-dbi
     - jellyfish
     - python
+extra:
+  biotools: [amos]
```
bgruening: can we have here this syntax?
```
biotools:
   - amos
```

bgruening: was there a decision about the key?
```
tool_identifier:
  - ff
  - ggg
```

```
extras:
   biotools_id:
     - ff
   other_id:
     - gg
```

bgruening:
ping @bioconda/core 
Sorry for the large diff.

recipes/amos/meta.yaml
ypriverol: @bgruening @mc-r the I Would like to have somethin that can be auto-resolve and dont need to much parsing: 

```
extras: 
   biotools:
       - amos
```
or 

```
extras: 
  identifiers:
      - biotools: amos 
```
The only problem with the second one is that we need to assume that every provider provides a unique identifier which in some cases probably is not 100% true. And this is one of the main reason we (@hmenager and myself) used this `[amos]`. This can help us in the future for example for one bioconda package to have all the pubmed ids like: `pubmed: [PMID:ZZZ, PMID:ZZZZ]`

Thanks a lot for this. Hope you guys love the idea @bioconda/core

hmenager: What about:

```
extras: 
  identifiers:
      - biotools: [amos]
```

Would that suit you @ypriverol and @bgruening ?

ypriverol: +1 this was what we originally thought. Support multiple providers and multiple ids for same provider.

johanneskoester: Yes, why not. Maybe:
```
extra:
    ids:
        biotools: [amos]
```
What would be the advantage of adding a `-` before biotools?

ypriverol: Hi @johanneskoester as a developer I will said 100% for the shorter 'ids' but as we should expect users doing annotation, probably 'identifiers' is more intuitive. Also because is not really long. But I can accept both. BTW, not real advantage, only that we can have also this:

'''
extra:
    identifiers:
         - biotools: [amos]
         - punned: [pmid:23456]
'''
What do you think?

stain: I hope we don't invent new prefix scheme like `punned` or `pmid:` -- where do we resolve those prefixes? Another prefix registry?

Use http://identifiers.org/  (which admitedly don't really do many tool registries today) or simply full URIs  - with clear recommendations/examples. That would then be automatically extensible without any further work.

hmenager: indeed @johanneskoester , your syntax is what i was thinking about, the dash is useless there i guess.

johanneskoester: @stain, this is a very interesting point, indeed. You mean, instead of introducing a biotools identifier, one would register a tool at identifiers.org, and use that identifier both in bioconda and biotools?

ypriverol: Hi @stain agree with you the punned id was only an example to clarify that we can implement multiple providers. Agree with you about the identifiers.org idea. We can probably get in contact with Sarala or others in identifiers.org for supporting software. 

Something we need to agree in the scheme/implementation is if we want to support full URL or key value pairs.

ypriverol: If you are happy with that I can trigger the discussion with identifier.org

tomkinsc: Are there cases where the biotools ID differs from the package name? Right now it seems redundant to list the name twice.

ypriverol: Hi @tomkinsc the current annotations are the one we can "guess" from biotools + bioconda. The complete list with the mismatch is here https://github.com/BioContainers/biotools-bioconda-ids/blob/master/biotools-biocontainers-venn.ipynb we would like to be able to have a mechanism in bioconda to reference to other ids tools. This will enable to registries such as biocontainers.pro/registry/ to pull the metadata from metadata resources. The machanism we are proposing will also work for other identifiers: publications, identifier.org, etc.

bgruening: @ypriverol @hmenager wasn't the idea to link the entire URI(s)?
I thought the only discussion was if the URI is enough or if we want to use an additional annotation key?

ypriverol: @bgruening We can use URL or the key-value pairs. The problems with the urls is that some of the resources can change their URL and break the system, key-value pair may work in the future better.

ypriverol: Hi all, @stain: I have been talking to Sarala, the main developer/project lead from identifiers.org. The way they can encode in the future bio.tools would bee: 

```
identifiers.org\bio.tools\amos 
``` 
From my point of view, this should be a discussion trigger by bio.tools group with identifiers.org. If we have the current representation `- biotools: [amos]` we can resolve this. I need to say that I don't know any place that standardizes yet the ids of software, tools or packages. Then, it will take some time until something is out there stable. 

This PR is trying to move forward the standardization in bioconda of references to identifiers in other tools.

mr-c: FYI, there is already another source of software identifiers at identifiers.org: Research Resource IDentifier (RRID) https://identifiers.org/rrid/

Here's an example for Bowtie https://identifiers.org/rrid/RRID:SCR_005476

ypriverol: Hi all: 
    I have seen Bio.tools definitions and discussing with @bgruening understanding better the 
    bio.tools id system. In this case, the identifier and URI is `http://bio.tools/amos` instead of 
    amos. Then the current proposal should be the following:   

```
extra:
   identifiers: 
       biotools: http://bio.tools/amos
```  
then the value of this key-value pair (the identifier) should be a URI.

bgruening:
@ypriverol I think you latest push removed the annotations.

ypriverol:
sorry, i will be updated .. sorry for that.

ypriverol:
I have changed to a new structure: 

```
extra:
  identifiers:
     biotools:
         - http://bio.tools/abyss
```
Thanks for the feedback.

mbargull:
> @ypriverol I think you latest push removed the annotations.

It apparently also ditched the comments from the other reviewers :-/

I'm in no way involved in this discussion / proposal, but I like it!

Though I have a big remark regarding the PR itself and how it was created:
Firstly, in the current form, this should not be merged at all as it will break quite a few recipes!
Beware that it is actually considerably risky to automatically alter conda recipes. That is because the `meta.yaml` files are often no valid/unambiguous YAML files at all due to their two independent preprocessors (`[...]` selectors and `jinja2`). Hence I'd advise to avoid parsing and then recreating them (unless someone would create a tool which could handle all possible selector combinations etc., which can be time-consuming..).
One thing that apparently happened for this PR was, that the selectors were not handled correctly, leading to duplicate keys which got removed, see, e.g.:
```diff
--- a/recipes/novoalign/meta.yaml
+++ b/recipes/novoalign/meta.yaml
@@ -5,8 +5,6 @@ package:
   version: '{{ version }}'
 source:
   fn: novocraftV{{ version }}.tar.gz
-  url: http://www.novocraft.com/homebrew/novocraftV{{ version }}.Linux2.6.tar.gz # [linux]
-  md5: d0cd7d20cf6f10f1da8db1e875356aa3 # [linux]
   url: http://www.novocraft.com/homebrew/novocraftV{{ version }}.MacOSX.tar.gz # [osx]
   md5: 1a0f98015fe0bf7c8e135132063b312f # [osx]
```
My advice: Don't try to load, change and dump the `meta.yaml` files but just append the `extra:` section, treating them as plain text files. There not too many files affected by this PR which already have an `extra:` section. Those could be identified via a simple `grep '^extra:'` and then be changed manually.

ypriverol:
@mbargull I can try to improve even more the tool to do the automatic annotation. The problem is that if we use the latest section of the recipe, in the future when we run again this we can have the problem of detecting if the information. 

I have improved the current tool to consider all the variants now. In the future, we can use this tool to accept/correct the recipes. In addition, I suggest and agree with you that we can update the recipes in blocks, for example, make a PR every 10 recipes. But I would prefer to use my actual code and do this automatically.

mbargull:
All right, if this is not a one-time effort and you have updates in mind, then of course it is worthwhile to invest time into such a tool and my less automatic approach would not be practically feasible.
Smaller PRs with around 10 recipes would be nice, indeed. Also if the sources to your tool are available somewhere, I'm sure some people are happy to help with some of those tricky quirks!

ypriverol:
@mbargull code here contributions more than welcome (https://github.com/BioContainers/biotools-bioconda-ids/blob/master/annotation_bioconda.ipynb). I' working now with the local copy of the recipes. thanks for reporting the bug of the duplications. I'm really new to yaml+jinja2 and it will take me some time to finish python notebook. This PR was more to have the discussion and agree in the annotation. If everyone is happy with the structure... I will go for the future changes. @hmenager ? do you like the current representation. ?

bgruening:
@mbargull are 100 recipes per PR also ok? I will spend a few hours and review them. The question for me is more if we also want to fix the linting in the same PRs or do we merge it even with failing PRs as long as it only adds metadata.

ypriverol:
Hi @bgruening and @hmenager @mbargull Now the most important thing is to agree in the structure of the metadata. I can work after that in the system to update the recipes every 10 or 50 or 100 recipes.

mbargull:
@bgruening I'm fine with up to 100 recipes per PR. TravisCI & co might timeout though ;). I'd prefer to fix linting issues separately, but am fine either way. I could prioritize the recipes from this PR in my lint maintenance list to reduce conflicts in advance, though.

@ypriverol You are right, it is currently far more important to decide on the structure than to talk about those technicalities!

bgruening:
@mbargull there will be no builds on the CI as we do not increase the build number. So it should be fine for the CI system. We just changing metadata here.

ypriverol:
Fantastic guys, I will schedule 10 recipes per PR and review them before the PR, then you can review it. We don't need either to do this for tomorrow, the most important thing is to move the annotations forward. @bgruening do you have the email of @hmenager, we can probably contact him and see what he said.

ypriverol:
@bgruening @mbargull another idea that comes to my mind is that we annotate this information independent of the recipe, and we solve the dependency through the biocontainers API. In this way, we can keep separating the container recipe/bioconda recipe from the metadata annotation. if we need to change something in the metadata we update that file rather than the recipe.

daler:
@ypriverol I like the idea of separating API-specific metadata from recipes. The same parsing code could be used, just adding metadata to a separate file rather than incorporating into the recipe (and thereby potentially breaking the recipe). Seems like it would be a lot more flexible that way as well.

bgruening:
@ypriverol so back to this idea? https://github.com/bioconda/bioconda-recipes/blob/master/recipes/locarna/biotools.yaml
We can start this yaml file the tool ids.

ypriverol:
Would be probably better to store this out of the recipe completely. I would think about a btter way because if we put this close to the recipe, we will need to update the recipe every time is updated the metadata. I would take a look to this file and be back to you.

bgruening:
It should be part of my slides from the hackathon. Thanks!

johanneskoester:

PR #6319, opened 2017-10-16
lint setuptools args: recipes "a"

mbargull:

Hi there @bioconda/core, I'm Marcel and currently working for @johanneskoester to do some maintenance around here.
This is part (of a first small set) of maintenance PRs I'm going to issue in the following weeks. I'll mainly try to resolve issues pointed out from `bioconda-utils lint`. For these first few recipes there are also some version updates etc. included (since I took a closer look at the recipes to somehow get a feeling on what I'll be dealing with ;) ). To be able to pick up some pace so that I can handle a somewhat more noticeable amount of recipes, the PRs in the following days / weeks will likely be more lint-error-focused. I just created a `lint` label which I'll add to those.

Notes for this PR: `argparse` is available at conda-forge and thus can likely be deleted here -- I just wasn't sure whether this is something I should include in one of those batch updates..


bgruening:
@mbargull welcome to Bioconda!
I think you can remove the argparse package from github. It is now maintained in conda-forge.

Let us know if you need any help.

recipes/alignlib-lite/meta.yaml
```diff
@@ -16,19 +18,18 @@ requirements:
     - llvm # [osx]
     - python
     - setuptools
-    - boost
+    - boost {{ CONDA_BOOST }}*
```
johanneskoester: Seems like alignlib is linking statically against boost. In such a case, we don't need the pinning. I will update the linter such that it ignores such cases.

johanneskoester:
@bgruening, for your information: Marcel is the student assistant I was mentioning before.

recipes/alignlib-lite/meta.yaml
johanneskoester: xref: bioconda/bioconda-utils#191

daler:
Thanks @mbargull! This will be very helpful.

recipes/alignlib-lite/meta.yaml
mbargull: How would you detect such cases via the linter?
Are you positive about the static linking (haven't checked myself)? So, should I just add the appropriate `skip-lints` for this recipe (with a comment that it links statically)?

mbargull: @johanneskoester / https://github.com/bioconda/bioconda-utils/pull/191:
> If the dependency does not occur in run, we don't need to pin it at all because it is statically linked.

Ok, got it! I'll wait for https://github.com/bioconda/bioconda-utils/pull/191 to be merged an then update this PR by removing the pinning.

johanneskoester: It is merged.

PR #6320, opened 2017-10-16
lint setuptools args: recipes "b" 1

mbargull:

Lint maintenance & updates. @johanneskoester, context at #6319.

Hi @guillermo-carrasco, I just notices that you updated `bcbio_monitor` around the time you created the recipe for it. Is it fine to update the bioconda recipe from `1.0.2` to `1.0.6`, like I did here?


recipes/backports.csv/meta.yaml
```diff
@@ -24,27 +20,11 @@ requirements:
...
-    # Put any additional test requirements here.  For example
-    # - nose
+test:
+  imports:
+    - backports.csv
 
 about:
   home: https://github.com/ryanhiebert/backports.csv
-  license: UNKNOWN
+  license: Python Software Foundation License
```
bgruening: I guess we should stick to:
```
license: PSF
license_family: PSF
```
as done by conda-forge.

mbargull: I used the long version since `grep 'PSF\|Python Software Foundation' recipes/**/meta.yaml` did only output those -- though I'd also prefer `PSF` (or, even better, `PSF 2`)! But nevermind, I checked conda-forge's `python`, `argparse`, and, well..., `backports.csv` feedstocks and their identifiers are also not entirely consistent.
But the main thing is that `backports.csv` is in conda-forge and thus can be deleted here.

PR #6321, opened 2017-10-16
lint setuptools args: recipes "b" 2

mbargull:

Lint maintenance & updates. @johanneskoester, context at #6319.

`biom_format`: I guess it's ok to remove `setuptools` as a run dependency, though I just checked with a quick `grep -r setup` which did not come up with anything `setuptools` related.

`biomaj`: Currently the recipe is at `3.0.19` and could be updated to `3.0.20`. Note that there are even `3.1.x` versions available for which the package is heavily refactored and split into subpackages.


johanneskoester:
biokit needs the extended base image for the container based testing. See, e.g., the "lace" recipe.

johanneskoester:
The reason is that the default minimal container lacks opengl, which has to be delivered by some kind of system library (because it comes with a GPU driver).

mbargull:
I was just about to ask how you deal with those system/OpenGL requirement -- thank you!

johanneskoester:
I think you need the extended container image for the  failing recipe. I hope that the no space left error of the last one goes away once everything else passes (sometimes the container based testing leaves stuff on disk if it fails).

recipes/biom-format/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - pandas >=0.19.2
+    - six >=1.10.0
+
+test:
+  imports:
+    - biom
+    - biom.cli
+
+  commands:
+    - biom --help
```
johanneskoester: I think you need to export C.UTF-8 locale here: https://travis-ci.org/bioconda/bioconda-recipes/jobs/290521631#L974

mbargull: I deliberately didn't set those here since that should be done in the extended base container. It didn't work since there was an error in the Dockerfile, which is now fixed.
Now the job bailed out with an `IOError(28, 'No space left on device')`, so I've taken the recipe out into a separate PR #6400.

PR #6325, opened 2017-10-16
Update version of clustalo

souravsingh:

### DIFF
--- a/recipes/clustalo/build.sh
+++ b/recipes/clustalo/build.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+export CPPFLAGS="-I$PREFIX/include"
+export LDFLAGS="-L$PREFIX/lib"
+
 ./configure --prefix=$SRC_DIR
 make
 
--- a/recipes/clustalo/build.sh
+++ b/recipes/clustalo/build.sh
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+./configure --prefix=$SRC_DIR
+make
+make install
+
 mkdir -p $PREFIX/bin
-cp $SRC_DIR/clustal* $PREFIX/bin/$PKG_NAME
+cp $SRC_DIR/bin/clustal* $PREFIX/bin/$PKG_NAME
 chmod +x $PREFIX/bin/$PKG_NAME
--- a/recipes/clustalo/build.sh
+++ b/recipes/clustalo/build.sh
@@ -2,8 +2,7 @@
 
 ./configure --prefix=$SRC_DIR
 make
-make install
 
 mkdir -p $PREFIX/bin
-cp $SRC_DIR/bin/clustal* $PREFIX/bin/$PKG_NAME
+cp $SRC_DIR/src/clustalo $PREFIX/bin/$PKG_NAME
 chmod +x $PREFIX/bin/$PKG_NAME
--- a/recipes/clustalo/meta.yaml
+++ b/recipes/clustalo/meta.yaml
@@ -10,11 +10,17 @@ source:
 
 requirements:
   build:
-    - gcc
+    - gcc # [linux]
+    - llvm # [osx]
+    - libgcc # [linux]
     - argtable2
   run:
-    - libgcc
+    - llvm # [osx]
+    - gcc # [linux]
+    - libgcc # [linux]
     - argtable2
+    
+    
 
 build:
   number: 0
--- a/recipes/clustalo/meta.yaml
+++ b/recipes/clustalo/meta.yaml
@@ -5,18 +5,16 @@ package:
   version: "{{ version }}"
 
 source:
-  url: http://www.clustal.org/omega/clustal-omega-{{ version }}.tar.gz # [not win]
-  sha256: 8683d2286d663a46412c12a0c789e755e7fd77088fb3bc0342bb71667f05a3ee # [not win]
+  url: http://www.clustal.org/omega/clustal-omega-{{ version }}.tar.gz
+  sha256: 8683d2286d663a46412c12a0c789e755e7fd77088fb3bc0342bb71667f05a3ee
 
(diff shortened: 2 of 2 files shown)


cbrueffer:
I looked into this a while ago, there's no updated OSX binary available. Only updating the Linux one as currently done won't work. Maybe ask the Clustal folks to update the OSX binary as well?

souravsingh:
I just noticed this. I believe this could be fixed by creating separate recipes for Linux and OSX.

cbrueffer:
That may work, but I don't think it's a good solution. Way easier to send an email :-)

souravsingh:
Ok, I have sent a mail to the maintainers of Clustalo

cbrueffer:
Great, thanks!

eseiler:
Hey,

I have two suggestions:

Either
(1) Use different versions for Linux and OSX:
```
{% set version = "1.2.3" %} # [osx] 
{% set version = "1.2.4" %} # [linux]
```

Or
(2) Build from source:
meta.yaml

```
{% set version = "1.2.4" %}

package:
  name: "clustalo"
  version: "{{ version }}"

source:
  url: http://www.clustal.org/omega/clustal-omega-{{ version }}.tar.gz # [not win]
  sha256: 8683d2286d663a46412c12a0c789e755e7fd77088fb3bc0342bb71667f05a3ee # [not win]

requirements:
  build:
    - gcc
    - argtable2
  run:
    - libgcc
    - argtable2

build:
  number: 0

test:
  commands:
    - "clustalo --help > /dev/null"

about:
  home: "http://www.clustal.org/omega/"
  license: "GPLv2"
  summary: "Latest version of Clustal: a multiple sequence alignment program for DNA or proteins"
```

build.sh
```
#!/bin/bash

./configure --prefix=$SRC_DIR
make
make install

mkdir -p $PREFIX/bin
cp $SRC_DIR/bin/clustal* $PREFIX/bin/$PKG_NAME
chmod +x $PREFIX/bin/$PKG_NAME
```

It's usually better to build it from source if possible and it would circumvent the version problem in the future.
I tested both (1) and (2) on travis and they worked just fine.

There may be some additional options you want to pass to ./configure (see their install manual), but this is all that is needed to be compiled from source.

recipes/clustalo/build.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash
 
+./configure --prefix=$SRC_DIR
+make
+make install
```
cbrueffer: I think this shouldn't be necessary given that the binary is copied by hand later on?

eseiler:
Yes, this may be a good idea.
If it just copies the binaries, it shouldn't matter much, though.

recipes/clustalo/meta.yaml
```diff
@@ -1,25 +1,29 @@
...
-  md5: d6f953f498ea4fa1d08f57ea8cd731d0 # [linux]
-  fn: clustal-omega-{{ version }}-macosx # [osx]
-  url: http://www.clustal.org/omega/clustal-omega-{{ version }}-macosx # [osx]
-  md5: e8ec7292c258b8b1ef4f3780b3952fe8 # [osx]
+  url: http://www.clustal.org/omega/clustal-omega-{{ version }}.tar.gz # [not win]
+  sha256: 8683d2286d663a46412c12a0c789e755e7fd77088fb3bc0342bb71667f05a3ee # [not win]
+
+requirements:
+  build:
+    - gcc
```
bgruening: Can you please try to use llvm for osx as written in our documentation: https://bioconda.github.io/guidelines.html#c-c

recipes/clustalo/meta.yaml
```diff
@@ -1,25 +1,35 @@
...
   version: "{{ version }}"
 
 source:
-  fn: clustalo-{{ version }}-Ubuntu-x86_64 # [linux]
-  url: http://www.clustal.org/omega/clustalo-{{ version }}-Ubuntu-x86_64 # [linux]
-  md5: d6f953f498ea4fa1d08f57ea8cd731d0 # [linux]
-  fn: clustal-omega-{{ version }}-macosx # [osx]
-  url: http://www.clustal.org/omega/clustal-omega-{{ version }}-macosx # [osx]
-  md5: e8ec7292c258b8b1ef4f3780b3952fe8 # [osx]
+  url: http://www.clustal.org/omega/clustal-omega-{{ version }}.tar.gz # [not win]
```
bgruening: you don't need the [not win] selector, we don't compile for Windows

recipes/clustalo/meta.yaml
```diff
@@ -1,25 +1,35 @@
...
+
+requirements:
+  build:
+    - gcc # [linux]
+    - llvm # [osx]
+    - libgcc # [linux]
+    - argtable2
+  run:
+    - llvm # [osx]
+    - gcc # [linux]
```
bgruening: gcc should not be a runtime dependency

recipes/clustalo/meta.yaml
```diff
@@ -1,25 +1,35 @@
...
-  url: http://www.clustal.org/omega/clustal-omega-{{ version }}-macosx # [osx]
-  md5: e8ec7292c258b8b1ef4f3780b3952fe8 # [osx]
+  url: http://www.clustal.org/omega/clustal-omega-{{ version }}.tar.gz # [not win]
+  sha256: 8683d2286d663a46412c12a0c789e755e7fd77088fb3bc0342bb71667f05a3ee # [not win]
+
+requirements:
+  build:
+    - gcc # [linux]
+    - llvm # [osx]
+    - libgcc # [linux]
```
bgruening: libgcc should not be a compile time dependency

souravsingh:
Looks good now. @bgruening Would you like to take a look before merging?

PR #6328, opened 2017-10-16
Add gi-eagle

christopher-schroeder:

### DIFF
--- a/recipes/gi-eagle/meta.yaml
+++ b/recipes/gi-eagle/meta.yaml
@@ -1,0 +1,61 @@
+{% set name = "gi-eagle" %}
+{% set version = "0.9.3.3" %}
+{% set hash_type = "sha256" %}
+{% set hash_value = "f7027a67b4ae857cec8ed231bf2f5b001d6a2070909cc7f4590244b749c836d8" %}
+
+package:
+  version: '{{ version }}'
+  name: '{{ name|lower }}'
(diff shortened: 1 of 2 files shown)


recipes/gi-eagle/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+  license_file: ''
+  doc_url: ''
+  summary: Summary of the package
+  license: MIT
+  home: https://bitbucket.org/christopherschroeder/eagle
+  license_family: MIT
+  description: ''
+
+extra:
+  recipe-maintainers: ''
```
johanneskoester: You can remove all the empty sections.

recipes/gi-eagle/meta.yaml
```diff
@@ -0,0 +1,61 @@
+{% set name = "gi-eagle" %}
+{% set version = "0.9.3.3" %}
+{% set hash_type = "sha256" %}
+{% set hash_value = "f7027a67b4ae857cec8ed231bf2f5b001d6a2070909cc7f4590244b749c836d8" %}
```
johanneskoester: We usually only set the parts that change and occur more than once with a jinja variable. Here, this means only `version` should be a jinja variable. This helps to keep the yaml as readable as possible, while an update still only involves to edit the version and the hash once.

PR #6333, opened 2017-10-17
Yara 0.9.9 64x

eseiler:

Hey @bioconda/core ,

there are also binaries that are compiled with SSE4 support and are therefore more efficient.
For the sake of maximizing compatibility, there are only the 64x packages in this recipe.

I have found three different ways so far to include the SSE4 packages which each have different requirements:

- Check for SSE4 support in meta.yaml:
    This could be done by `{% set sse4_support = os.system("grep flags /proc/cpuinfo | grep sse4  > /dev/null") %} # [linux64]` for linux and a similar command for osx, but it would require os to be available in the jinja environment
- Download all sources and check for SSE4 support in build.sh:
    This would require at least cond-build 3.0, so you can download different sources and put them in different folders by using the 'folder: <folder>' directive in the source part.
- Build from source and check for SSE4 support in build.sh:
    This would require gcc >=4.9.* in any channel used by conda-build. There is one in the `https://repo.continuum.io/pkgs/main/linux-64/` channel, but it isn't included.

I guess the best solution would be to wait for gcc >= 4.9 to be available and until then to just distribute the packages without SSE4 support?

Best,
Enrico

### DIFF
--- a/recipes/yara/build.sh
+++ b/recipes/yara/build.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -eu -o pipefail
 
 BINARY_HOME=$PREFIX/bin
 PKG_HOME=$PREFIX/opt/$PKG_NAME-$PKG_VERSION
--- a/recipes/yara/meta.yaml
+++ b/recipes/yara/meta.yaml
@@ -1,22 +1,24 @@
+{% set version="0.9.9" %}
+
 build:
   number: 0
+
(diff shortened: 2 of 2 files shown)


bgruening:
Sounds good to me. I think a new GCC makes many thinks easier. We need to migrate to cb3 as soon as possible. conda-forge first ;)

PR #6336, opened 2017-10-17
Switch to using source package from cargo port

jamesabbott:


eseiler:
Hey, 
just in case you're still trying to figure out why travis doesn't actually build anything:
It has to be meta.y**a**ml, not meta.yml. At least `conda build .` run in the cgview directory says so :)

bgruening:
@eseiler good catch!

jamesabbott:
Ah! Thanks....I seem to making more newbie mistake with this than my first package...

PR #6349, opened 2017-10-18
Cgview

jamesabbott:

Fix broken tests, and make build.sh OSX compliant


recipes/cgview/meta.yaml
```diff
@@ -28,8 +28,3 @@ requirements:
        - perl
        - perl-bioperl
 
-test:
-    commands:
-        - cgview >/dev/null 2>&1
```
bgruening: Can we stick to this but use the following trick: `cgview 2>&1 | grep "some message in the output"`

jamesabbott: That would be seem more sensible...

bgruening:
Thanks!

PR #6357, opened 2017-10-19
Add recipe funcannot v2.8

mtekman:

### DIFF
--- a/recipes/funcannot/meta.yaml
+++ b/recipes/funcannot/meta.yaml
@@ -1,0 +1,32 @@
+package:
+  name: funcannot
+  version: "v2.8"
+
+source:
+  url: https://github.com/BioTools-Tek/funcannot/archive/master.zip
+  md5: 2e70b503a1632998deba759b50c0c07f
+   
+
+build:
+  number: 0
+  #skip: True # [osx]
+
+requirements:
+  build:
+    - gcc
+    - zlib {{CONDA_ZLIB}}*
+    - qt
+
+  run:
+    - libgcc
+    - qt
+
+
+test:
+  commands:
+    - funcannot 2>&1 | grep "INPUTS:"
+
+about:
+  home: https://github.com/BioTools-Tek/funcannot
+  license: GPLv3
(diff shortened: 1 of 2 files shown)


recipes/funcannot/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+  md5: 2e70b503a1632998deba759b50c0c07f
+   
+
+build:
+  number: 0
+  #skip: True # [osx]
+
+requirements:
+  build:
+    - gcc
```
bgruening: llvm on macos is not working?

recipes/funcannot/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - zlib {{CONDA_ZLIB}}*
```
bgruening: is this really needed or linked in statically? I'm just wondering why it is not needed as runtime dependency.

recipes/funcannot/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - qt
+
+  run:
+    - libgcc
+    - qt
```
bgruening: do you need to pin the qt version? version 4 or 5?

recipes/funcannot/meta.yaml
mtekman: untested, but if it can build under gcc, why not use gcc?

recipes/funcannot/meta.yaml
mtekman: It can build for both, but I'll lock it qt4 for now

recipes/funcannot/meta.yaml
mtekman: Ah. I thought zlib was needed by conda at build time to unpack the tool. 
So these are includes that are fed to the actual build script, okay I understand -- will remove.

recipes/funcannot/meta.yaml
mtekman: (I will specify llvm for osx, in future commits - I just want to know why llvm is preferred over a more cross-platform soln)

PR #6362, opened 2017-10-19
Update package dependencies for secimtools

moskalenko:

### DIFF
--- a/recipes/secimtools/meta.yaml
+++ b/recipes/secimtools/meta.yaml
@@ -1,8 +1,8 @@
 {% set name = "secimtools" %}
 {% set version = "1.0.0" %}
-{% set rc_version = "rc20170724" %}
+{% set rc_version = "rc20171016" %}
 {% set hash_type = "md5" %}
-{% set hash_value = "047b10ea87f70a0d46e7dea3f9e1147a" %}
+{% set hash_value = "0490fa17658188c6bc421178cf84ba7e" %}
 package:
   name: '{{ name|lower }}'
   version: '{{ version }}'
@@ -13,7 +13,7 @@ source:
   '{{ hash_type }}': '{{ hash_value }}'
 
 build:
-  number: 0
+  number: 1
   skip: True  # [py3k]
   script: python setup.py install  --single-version-externally-managed --record=record.txt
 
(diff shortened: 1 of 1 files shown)


recipes/secimtools/meta.yaml
```diff
@@ -64,9 +60,12 @@ test:
 about:
   home: https://github.com/secimTools/SECIMTools
   license: MIT License
+  license_file: ''
```
bgruening: LICENSE.txt

moskalenko:
Updated build number and pypi package version

recipes/secimtools/meta.yaml
```diff
@@ -13,46 +13,42 @@ source:
   '{{ hash_type }}': '{{ hash_value }}'
 
 build:
-  number: 0
+  number: 1
```
bgruening: @moskalenko I would rather but an rc version string into the real package version. The way you are doing it here will give you a new version even if you specify 1.0.0.

moskalenko: Can you clarify the above statement? Thank you.

bgruening: If you type in `conda install secimtools=1.0.0` you are expecting one single version with a specific set of features and bugs. You are keeping this version number but increasing the build number and updating the source. This means the version 1.0.0 has now potential new bugs and features. So it is not reproducible for people that want to install the old version 1.0.0.

moskalenko: I really don't want to settle on a 1.0.0 or set any expectations of reproducibility until the paper is accepted at which time I will release 1.0.0 packages and use semantic versioning from there on. This is pre1.0.0 with minor fixes going in to get the galaxy wrappers done, which is the main blocker at this time. Wrappers need this update.

bgruening: In this case should I remove the older version?

moskalenko: Sure. Danke.

moskalenko:
Björn,

Is anything else needed here?

Regards,

Alex

moskalenko:
It looks like it's blocked by CI, so it gets reopened every time you close it.

bgruening:
Thats me that is trying to convince travis to test it. I will take care of this.

PR #6373, opened 2017-10-20
Fix name in cyvcf2

druvus:

### DIFF
--- a/recipes/cyvcf2/meta.yaml
+++ b/recipes/cyvcf2/meta.yaml
@@ -1,4 +1,4 @@
-{% set name = "cyvcf" %}
+{% set name = "cyvcf2" %}
 {% set version = "0.8.0" %}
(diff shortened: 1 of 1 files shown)


recipes/cyvcf2/meta.yaml
```diff
@@ -1,4 +1,4 @@
-{% set name = "cyvcf" %}
+{% set name = "cyvcf2" %}
```
bgruening: mh, should we remove the old package then?

druvus: I was going to ask you that... I think it would be a good idea. It is just the last one for ten days ago that had this name typo.

bgruening: Ok, let me remove the last one.
Thanke Andreas!

PR #6377, opened 2017-10-21
HTSlib update to v1.6

mrphyja:

@bioconda/core <- this mention doesn't seem to be working?

First time contributor, but it's just a simple version update so not too much to break hopefully. I also also had to pin bzip2 to satisfy linting checks. Thanks!


recipes/pysam/meta.yaml
```diff
@@ -9,7 +9,7 @@ source:
 
 
 build:
-    number: 1
+    number: 2
```
bgruening: I don't think you need this version bump. You can even reset it to 0 as you will get a new build string by increasing the htslib package in the env_matrix.

bgruening:
ping @kyleabeauchamp for a review.

mrphyja:
Maybe I should just revert that last commit for pysam and stick with just updating HTSlib? I could probably do another PR for pysam later if I end up needing it, it was kind of an afterthought anyway.

kyleabeauchamp:
Yeah I would leave the pysam as is.

kyleabeauchamp:
Otherwise lgtm

mrphyja:
Ok, reverted the commit that changed pysam, just an HTSlib update now.

bgruening:
@mrphyja better then reverting is to set it back to 0.

mrphyja:
@bgruening I'm a bit confused, Kyle recommended to leave Pysam as is, so I basically just did a git revert on the commit that touched that file so now the only files being changed are `htslib/meta.yaml` and `env-matrix.yml`. Are you saying you think I should modify `pysam/meta.yaml`?

bgruening:
If I remember correctly the pysam recipe has a build number of 1 right? This could be reset to 0 afaik as we get a new build string either way.

mrphyja:
@bgruening Ok, thanks, I reset the build number to 0.

bgruening:
Thanks @mrphyja!

PR #6378, opened 2017-10-21
Asn2gb

jamesabbott:

This is a small NCBI utility for format conversion - one of a number they provide, including tbl2asn. The PR is therefore based on the tbl2asn recipe which uses binary builds of the utils. 

Contrary to the comments in the tbl2asn recipe, I believe the source code for these tools is available, but is part of the ncbi toolkit, which is a massive and complex build for a small utility, and I've not found the necessary configure options to build these utlitiies, so the binaries look a sensible option in this case.

Hopefully the use of binaries is acceptable in this case.

### DIFF
--- a/recipes/asn2gb/meta.yaml
+++ b/recipes/asn2gb/meta.yaml
@@ -2,6 +2,11 @@ package:
     name: asn2gb
     version: "18.2"
 
+about:
+    home: https://www.ncbi.nlm.nih.gov/IEB/ToolBox/C_DOC/lxr/source/doc/asn2gb.txt
+    license: Public Domain
+    summary: asn2gb converts ASN1 format sequence records to Genbank format
+
 source:
     fn: mac.asn2gb.gz # [osx]
     url: ftp://ftp.ncbi.nih.gov/toolbox/ncbi_tools/converters/by_program/asn2gb/mac.asn2gb.gz # [osx]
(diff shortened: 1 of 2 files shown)


recipes/asn2gb/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    home: https://www.ncbi.nlm.nih.gov/IEB/ToolBox/C_DOC/lxr/source/doc/asn2gb.txt
+    license: Public Domain
+    summary: asn2gb converts ASN1 format sequence records to Genbank format
+
+source:
+    fn: mac.asn2gb.gz # [osx]
+    url: ftp://ftp.ncbi.nih.gov/toolbox/ncbi_tools/converters/by_program/asn2gb/mac.asn2gb.gz # [osx]
+    sha256: e251014227d7cb9de7e33f2101b34c154d4ee40bc64b81d41dc4ed0fd5972d63 # [osx]
+    fn: linux.asn2gb.gz # [linux]
+    url: ftp://ftp.ncbi.nih.gov/toolbox/ncbi_tools/converters/by_program/asn2gb/linux.asn2gb.gz # [linux]
```
bgruening: we only compile for linux64 ... so the selector can be just [linux]
Compiling is no option?

jamesabbott:
The source is buried in the ncbi toolbox which is a huge build to get working (and I've not even managed to find the right incantation to make it build these conversion utilities). Since tbl2asn (another of these NCBI converters) has been added just as binaries I borrowed that recipe as a starting point. 

I'll take a look at what it would take to  get the ncbi toobox built...

jamesabbott:
Actually, it doesn't look as as scary as I thought...

I'll have a go at building a recipe for the toolbox - it may duplicate standalone tools which have already been added to bioconda indivudually (i.e. tbl2asn)

bgruening:
You call, I don't want to stop you :)

jamesabbott:
I've hit  a brick wall building the ncbi-toolbox due to it's dependency on a c-shell configuration script. I've tried building a tcsh package for conda-forge but the build fails a bunch of tests. The tcsh project seems dead with the website gone and mailing list address undeliverable so without assistance from someone who can get to the bottom of the test failures I'm stuck.
Think I have to give up and revert to this binary version, if your happy to merge it.

jamesabbott:
pinging @bgruening

bgruening:
@jamesabbott thanks for trying that hard!

bgruening:
Ok, this is a never ending story. The checksum does not match anymore and it was not uploaded to the archive. Can you mirror the binaries to cargo-port and we take these?

PR #6403, opened 2017-10-23
Add some new recipes

erasche:

### DIFF
--- a/recipes/galaxy-parsec/meta.yaml
+++ b/recipes/galaxy-parsec/meta.yaml
@@ -36,6 +36,10 @@ requirements:
     - xunit-wrapper >=0.12
     - future
 
+extra:
+  container:
+    extended-base: true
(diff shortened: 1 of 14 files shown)


recipes/galaxy-parsec/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - bioblend
+    - wrapt
+    - pyyaml
+    - justbackoff
+    - xunit-wrapper
+    - future
+
+  run:
+    - python
+    - click >=6.7
```
bgruening: @erasche since you are using click. Can you please add this to the recipe:
```
extra:
  container:
    extended-base: true
```

We will not use busybox here for testing but a slightly larger container with full UTF support.

bgruening:
Thanks @erasche

PR #6408, opened 2017-10-23
Mimeo

Adamtaranto:


recipes/mimeo/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py install
```
bgruening: I think this needs a change: https://bioconda.github.io/linting.html#setup-py-install-args

Adamtaranto: Fixed in latest commit, thanks.

bgruening:
Nice team work. Thanks both!

PR #6411, opened 2017-10-23
Add tirmite

Adamtaranto:

### DIFF
--- a/recipes/tirmite/build.sh
+++ b/recipes/tirmite/build.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-$PYTHON setup.py install
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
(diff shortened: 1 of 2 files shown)


recipes/tirmite/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py install
```
druvus: Please update it to `$PYTHON setup.py install --single-version-externally-managed --record=record.txt`.

Adamtaranto: Fixed in latest commit

Adamtaranto: Not sure why linux build test is failing

druvus: I fixed the linux build. The issue was that libgcc was in the build reqs section instead of run reqs. Just need to wait for the osx travis build to finish.

PR #6414, opened 2017-10-23
lint setuptools args: recipes "c"

mbargull:

Lint fixes and cleanup.

### DIFF
--- a/recipes/chainmap/meta.yaml
+++ b/recipes/chainmap/meta.yaml
@@ -8,6 +8,8 @@ source:
   md5: 2ddf536db921c82c9559791047b88caf
 
 build:
+  # Skip builds since it is most likely not needed for Python 3.
+  # Before enabling those builds, rather consider to move the package to conda-forge.
   skip: True # [py3k]
   number: 0
 
--- a/recipes/chainmap/meta.yaml
+++ b/recipes/chainmap/meta.yaml
@@ -8,15 +8,16 @@ source:
   md5: 2ddf536db921c82c9559791047b88caf
 
 build:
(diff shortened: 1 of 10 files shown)


recipes/chainmap/meta.yaml
```diff
@@ -8,16 +8,15 @@ source:
   md5: 2ddf536db921c82c9559791047b88caf
 
 build:
-  skip: True # [py3k]
-  number: 0
+  number: 1
```
bgruening: It looks like this recipe has skipped python3 intentionally.

mbargull: Yes, but I could see any reason why from the Python package's perspective, which should support any Python version. Maybe it was skipped because it should actually not be needed for `python >=3.3`, having `collections.ChainMap`?
If it's necessary/reasonable (for BioConda) to avoid those extra builds, I can restore the `skip`.

johanneskoester: We should add a lint that requires an explanatory comment for such skips...

The reason you guess seems to be right. I think we should keep the skip for now, and move the package to conda-forge at some point.

mbargull: Regarding the explanations: yes, please! I was actually planning to open one or more issues/discussing about that in the bioconda-utils repository.

Otherwise, yes, this package should at some point be moved to conda-forge. I'll restore the skip for now by reverting 26f4b52b367a4fa982f1002b11d23e3bc86da943.

mbargull: Done.

PR #6415, opened 2017-10-23
lint setuptools args: recipes "c" 2

mbargull:

Lint fixes, cleanup, update.

### DIFF
--- a/recipes/cobra/meta.yaml
+++ b/recipes/cobra/meta.yaml
@@ -1,14 +1,17 @@
-build:
-  number: 0
-  skip: True # [osx]
+{% set version = "0.4.0" %}
 
 package:
   name: cobra
-  version: "0.4.0"
+  version: {{ version }}
 
 source:
-  git_url: https://github.com/opencobra/cobrapy.git
-  git_tag: de484b5a044da58496b4733f9d7c0302a55f4285
+  fn: cobra-{{ version }}.tar.gz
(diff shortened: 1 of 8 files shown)


recipes/cobra/meta.yaml
```diff
@@ -1,14 +1,17 @@
...
 
 source:
-  git_url: https://github.com/opencobra/cobrapy.git
-  git_tag: de484b5a044da58496b4733f9d7c0302a55f4285
+  fn: cobra-{{ version }}.tar.gz
+  url: https://pypi.io/packages/source/c/cobra/cobra-{{ version }}.tar.gz
+  sha256: 104e1f73f78d463849161143e8f2b98fcd7cb5ce8ee2c5b0c704ce3153036e30
+
+build:
+  number: 1
```
bgruening: any reason to increase there the build number?

mbargull: Hm.. The source is definitely different since one is a whole repository and the other one is a Python sdist. Would that justify a rebuild? If we want to avoid a rebuild, I'd then rather use a tarball from GitHub for the previously given tag; and just comment out the PyPI source for future references/uses. Would that be fine?
That way we would avoid a version number increase/rebuild *and* be sure the source did not change.

bgruening: Fair enough. In principle you would assume it is the same content, but better save than sorry.

PR #6417, opened 2017-10-23
Update pyProphet

blankclemens:


eseiler:
Since the package is only compatible with python2, we need to either skip the python3 builds, or see if we can convert the files to be compatible with python3 (see [HISAT2 recipe](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/hisat2/build.sh) ).

blankclemens:
Thanks!

bgruening:
Thanks @blankclemens and @eseiler!

PR #6418, opened 2017-10-23
Added a new recipe for Tablet.

imilne:

### DIFF
--- a/recipes/tablet/build.sh
+++ b/recipes/tablet/build.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+outdir=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+mkdir -p $outdir
+mkdir -p $PREFIX/bin
+cp -R * $outdir/
+chmod +x $outdir/tablet
+ln -s $outdir/tablet $PREFIX/bin
--- a/recipes/tablet/meta.yaml
+++ b/recipes/tablet/meta.yaml
@@ -1,0 +1,28 @@
+{% set name = "tablet" %}
+{% set version = "1.17.08.17" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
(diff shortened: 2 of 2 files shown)


peterjc:
Ideally I would include a copy of the licence in the zip file (as a text file, e.g. ``LICENSE.txt``), and then reference this in the ``meta.yaml`` file via the ``license_file`` key.

(I am not merging this new recipe due to the potential conflict of interest as a colleague of Iain).

peterjc:
Or you could use this -  from the filename I didn't initially spot this is the license file:

```
license_file: docs/tablet.html
```

imilne:
Thanks, Peter, that makes sense. I wouldn't merge anything yet, as I still don't know if this builds properly - I couldn't get Travis to play nice with my fork at all. I have a feeling it's going to die on the version-check test as Tablet may spool up a gui-event thread even before that happens, so some kind of "java.awt.AWTError: Can't connect to X11 window" might happen. It didn't show up with local testing as both my linux and mac laptops aren't headless. Flapjack (#6419) doesn't use that so may be ok.

peterjc:
Personally I would use something like ``tablet --version | grep '^Tablet`` for the version check (so that a human can inspect the log), given making an automatic exact regular expression will be tricky (escaping dots etc).

Anyway, that's minor polish - fingers crossed the build checks pass, and we can merge this.

imilne:
We had that, Peter, but it failed every time on Linux - passed fine on macOS though.

imilne:
And build number is 0 again - that was a just a random change to try and get things working in frustration late last night.

peterjc:
Merging given tests now passing and @bgruening's earlier approval.

peterjc:
@imilne the package ought to make its way to the public repository automatically now, check back tomorrow.

peterjc:
I was being too pessimistic on the deployment times, it is live now: https://anaconda.org/bioconda/tablet

bgruening:
Thanks all, nice work!

PR #6426, opened 2017-10-23
fishtaco: downgrade to 1.0.5 due to licensing restrictions

mbargull:

The FishTaco package should be removed since its very restrictive license actually forbids distribution, which BioConda does via its Conda channel, right? (Furthermore, the license actually even makes the software *self-destructing* (date: 2020-12-31) ?!?).

See the license file: https://github.com/borenstein-lab/fishtaco/blob/712607f0926fdf8be81495e2e932ddb06be8b6da/LICENSE
Some excerpts:
```
1. [...] The FishTaco software remains at your University and is not published, distributed, or otherwise transferred or made available to other than Academic Users.
```
```
2. You may not distribute FishTaco or any modification to FishTaco to any third party.
```
```
8. This Software License Agreement and all rights granted under it terminate on December 31st, 2020. Upon termination, you agree to remove so as to make unrecoverable the original FishTaco software, all copies and all modifications thereof.
```

To my understand the software should **not** have even be made available by the author themselves on GitHub as well as PyPI (due to `1.` above).
Even though it is available on PyPI (= source for this recipe), the license still prohibits the use on `anaconda.org/bioconda` (to my understanding).

IANAL(!), but IMO the license is pretty clear about that, right?

EDIT: Updated PR to not remove the recipe but revert to version 1.0.5 which is BSD-3-Clause licensed.


mbargull:
The recipe itself really isn't a problem license-wise, yes. But its whole purpose is to create a conda package to redistribute the software via `anaconda.org/bioconda`. And if the software's license forbids that redistribution, the recipe is actually useless and can be removed..

mbargull:
@ar0ch, do you see any reason to keep this recipe? BioConda doesn't redistribute FishTaco via this repository but with its conda channel it of course does so.

ar0ch:
From my reading, yes it should be fine.  As long as you're not directly serving the download, keeping a backup tar isn't in violation of the license. 

I do agree @mbargull that self destructing licenses are wierd and not very nice in spirit. 

All else fails, bioconda can pin 1.0.5 and blacklist updates. 1..0.5 and previous are under BSD-3 clause and are unencumbered as far as redistribution goes

ar0ch:
I don't personal use fishtaco, so I don't really mind if it is removed. I do support users who use it through conda however

mbargull:
Well yes, the backup tarball will not violate the license -- unless the date is 2021-01-01 or later ;D.
Now, BioConda **is** serving the download (using `anaconda.org` as a service provider), so we do violate the license there.

mbargull:
But I didn't know about the BSD license for versions up to 1.0.5! Then certainly it is the right thing to revert to that version instead of removing it altogether.

bgruening:
@mbargull fyi bioconda/core can remove in special cases like this single packages from the channel. Just let me know.

mbargull:
Yes, I know, and, yes, I guess we'll eventually have to do that. Maybe wait until anyone who might want to give input has done so.

druvus:
It is my mistake it got added to Bioconda and I am ok with that we remove it. 

The authors are using the standard WU license so it probably correct that they shouldn't have uploaded it in the first case to pypi/github.

johanneskoester:
Yes, all of the versions with incompatible license should be removed from the channel, and the recipe should be reverted to the last compatible version, including a comment that it may not be updated until the license changes upstream. Also, we should add a note to `extra::notes` saying that this recipe cannot be updated to a more recent version due to license restrictions.

@mbargull very good (and important) catch. You are doing a great job!

mbargull:
> , including a comment that it may not be updated until the license changes upstream. Also, we should add a note to extra::notes saying that this recipe cannot be updated to a more recent version due to license restrictions.

Perfect! That would have been my only question, to put that information into comments or extra/notes. Thanks for reading my mind! ;D

mbargull:
FYI, I opened https://github.com/borenstein-lab/fishtaco/issues/2 to make the authors aware of the licensing issues.

mbargull:
Updated PR to not remove the recipe but revert to version 1.0.5 which is BSD-3-Clause licensed.

mbargull:
Well, here we go... There were some problems:
FishTaco depends on MUSiCC and MUSiCC is incompatible with `scikit-learn >=0.18`. Now, the `musicc` recipe did not have this `scikit-learn <0.18`-constraint and its test did not catch this since it imported the empty module `musicc` and not `musicc.core`. This will now be fixed for the updated package in this PR.
But the builds of the previous version do not have this constraint and can be picked up by conda's solver. IIRC, even if we would issue a new build for `musicc 1.0.1`, the solver would still be able to consider the previous builds, depending on which packages' versions it tries to maximize. Hence we'd maybe need to either remove those builds or do something what the conda-guys call a metada-hotfix (IIRC). (That's a major unsolved issue with conda, currently..)
That aside, builds for Python 3.6 of `scikit-learn <0.18` are not available on conda-forge, thus I skipped those builds for `musicc` and `fishtaco`.
Overall, it's a messy situation. But I guess, without touching the previous builds and without requesting builds of `scikit-learn <0.18` for Python 3.6 from conda-forge, the workarounds in the `fishtaco` recipe are necessary..

mbargull:
Well, that's annoying. My local build & test for `musicc-1.0.2-py35_0` succeeded (which picked up `scikit-learn-0.17.1-np111py35_2`), but now with the CI (which picked up `scikit-learn-0.16.1-np19py35_0` and no `openblas`) for the test we get:
```python
[Oct 24 15:05:03] SERR File "/usr/local/lib/python3.5/site-packages/musicc/core.py", line 21, in <module>

[Oct 24 15:05:03] SERR from sklearn import cross_validation, linear_model # for the LASSO

[Oct 24 15:05:03] SERR File "/usr/local/lib/python3.5/site-packages/sklearn/cross_validation.py", line 32, in <module>

[Oct 24 15:05:03] SERR from .metrics.scorer import check_scoring

[Oct 24 15:05:03] SERR File "/usr/local/lib/python3.5/site-packages/sklearn/metrics/__init__.py", line 31, in <module>

[Oct 24 15:05:03] SERR from . import cluster

[Oct 24 15:05:03] SERR File "/usr/local/lib/python3.5/site-packages/sklearn/metrics/cluster/__init__.py", line 19, in <module>

[Oct 24 15:05:03] SERR from .unsupervised import silhouette_samples

[Oct 24 15:05:03] SERR File "/usr/local/lib/python3.5/site-packages/sklearn/metrics/cluster/unsupervised.py", line 10, in <module>

[Oct 24 15:05:03] SERR from ..pairwise import pairwise_distances

[Oct 24 15:05:03] SERR File "/usr/local/lib/python3.5/site-packages/sklearn/metrics/pairwise.py", line 29, in <module>

[Oct 24 15:05:03] SERR from .pairwise_fast import _chi2_kernel_fast, _sparse_manhattan

[Oct 24 15:05:03] SERR ImportError: libopenblas.so.0: cannot open shared object file: No such file or directory
```
Possibly builds/dependencies of one of those old packages are broken.. Any suggestions?

ar0ch:
Looks like you might need openblas (https://anaconda.org/conda-forge/openblas) as a build dep

mbargull:
The build was fine, so it would not be a build but run dependency. Either way, some other package musicc depends on (`scikit-learn`) would need that dependency.

However, I threw in the towel and just removed all those workarounds from before and just patched the import in `musicc` so that (at least of the test...) it is compatible with `scikit-learn >=0.18`. All in all this might be the least troublesome solution.

mbargull:
FishTaco's test failed since it picked up the older non-patched version of MUSiCC. Does the bioconda channel have a higher priority than the local channel from the build system?

johanneskoester:
@mbargull it shouldn't of course. But the tests pass now. Is the problem resolved and we can merge?

mbargull:
I just updated the patch for `musicc`. Other than that, yes, from my part it's good to go.
`fishtaco=1.1.1` should then be removed from the Conda channel.
@bgruening: That license/distribution restriction also concerns https://depot.galaxyproject.org/software/fishtaco/fishtaco_1.1.1_src_all.tar.gz

PR #6428, opened 2017-10-24
update schema-salad

eseiler:


eseiler:
Apparently, it cannot be compiled for python3, because the avro.schema.make_avsc_object function exists in avro-python2, but not in avro-python3.
They use a fork of avro-python3 called [avro-cwl](https://pypi.python.org/pypi/avro-cwl/1.8.4) instead.
I guess, we would either have to package the avro-cwl to use it with python3, or package it for py27 only and see what the future brings.
Any thoughts @bgruening ?

PR #6431, opened 2017-10-24
Barrnap python

nickp60:
This recipe adds a python implementation of barrnap for those Linux users working in environments where they do not have rights to update Perl packages.


peterjc:
Will ``barrnap-python`` fully replace ``riboseed`` from now on? Is that why you are removing the riboseed package?

nickp60:
@peterjc I may have misstepped at some point here, but I should have two open PR's: one for barrnap-python, and the other for riboseed.  I forgot to checkout the python-barrnap branch initially, and ended up with both the riboseed recipe and barrnap-python recipes in this branch.  I only intended to remove riboseed from this branch;  should I have reverted the riboseed changes instead of removed?

peterjc:
Cross reference #6288 for the Riboseed recipe

peterjc:
Right now this file exists on the master branch:

https://github.com/bioconda/bioconda-recipes/blob/master/recipes/riboseed/meta.yaml

This pull request (wrongly?) removes it, as well as adding the new barrnap-python recipe.

nickp60:
@peterjc thanks, I checkout the deleted recipe from the master branch and committed it to this branch.  Thanks so much for your help!

peterjc:
Cross reference https://github.com/nickp60/barrnap-python/issues/1

recipes/barrnap-python/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - python
+
+test:
+  commands:
+    - barrnap.py -h
+
+about:
+  home: https://github.com/nickp60/barrnap-python
+  license: GNU General Public License v3.0
+  summary: 'python package for TSs barrnap'
```
peterjc: I would expand TSs here, meaning is unclear.

recipes/barrnap-python/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  skip: True  # [py2k]
+  skip: True  # [py34]
+
+requirements:
+  build:
+    - python
+    - setuptools
+
+  run:
+    - barrnap >=0.8
```
peterjc: So this depends on the original Perl ``barrnap`` tool as packaged here:

https://github.com/bioconda/bioconda-recipes/tree/master/recipes/barrnap

nickp60: Yes;  I have added a clarifying note.  This is to make sure that (1) the actual authors can get accurate usage information of their work, and (2) to ensure that the HMM database is installed in a predictable place.

recipes/barrnap-python/meta.yaml
nickp60: I have clarified that in the upcoming push, thanks for the comment!

recipes/barrnap-python/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  version: '{{ version }}'
+
+source:
+  fn: '{{ name }}-{{ version }}.tar.gz'
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  '{{ hash_type }}': '{{ hash_value }}'
+
+build:
+  number: 0
+  skip: True  # [win]
```
bgruening: This should be one line. # [win or py2k or py3k]

bgruening: The win skip is not needed and I'm wondering if the py34 skip is really needed.

recipes/barrnap-python/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
+
+# Add more build steps here, if they are necessary.
```
bgruening: Can you also cleaup this file please?

nickp60: yes, thanks, sorry about that!

recipes/barrnap-python/meta.yaml
nickp60: Thanks, I changed that.  The pertinent change between py3.4 and later versions is the syntax for running system commands (subprocess.call() vs the more updated subprocess.run())

PR #6433, opened 2017-10-24
Racon: Remove python and mummer run deps

ar0ch:

This is an update to the run dependencies for racon.  The current recipe prevents installs on py3 environments because of a conflicting numpy/matplotlib requirement in the example script.  I don't think these should be enforced as strict run dependencies since the example script is discarded during install.  Also, the Unicycler recipe needs racon and is only py3 compatible

### DIFF
--- a/recipes/racon/meta.yaml
+++ b/recipes/racon/meta.yaml
@@ -7,7 +7,6 @@ package:
 
 build:
   number: 0
-  skip: False
 
 source:
   fn: {{ name|lower }}_{{ version }}.tar.gz
--- a/recipes/racon/meta.yaml
+++ b/recipes/racon/meta.yaml
@@ -7,7 +7,7 @@ package:
 
 build:
   number: 0
-  skip: True # [not py27]
+  skip: False
 
 source:
   fn: {{ name|lower }}_{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


recipes/racon/meta.yaml
```diff
@@ -7,7 +7,7 @@ package:
 
 build:
   number: 0
-  skip: True # [not py27]
+  skip: False
```
peterjc: Can you just delete the skip line now that is not used?

ar0ch: Oh, yeah that would be the right thing to do. Thanks

ar0ch:
@druvus Are we good to merge this change? We're going to do an offline build of the new Unicycler which will pull this as a runtime dependency

Thanks!

druvus:
@ar0ch @peterjc sorry I missed this PR. I have added some modification and agree with to remove python as a requirement.

bgruening:
Thanks all!

nekrut:
somehow this is still in conflict with unicycler because racon 0.5.0 build 1 has this:
```
nsatisfiableError: The following specifications were found to be in conflict:
  - libcurl -> zlib[version='>=1.2.11,<1.3.0a0']
```
is there a reason that this particular zlib is required?

PR #6436, opened 2017-10-24
Add run dependencies for Quast

bebatut:

### DIFF
--- a/recipes/quast/4.5/build.sh
+++ b/recipes/quast/4.5/build.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+export INCLUDE_PATH="${PREFIX}/include"
+export LIBRARY_PATH="${PREFIX}/lib"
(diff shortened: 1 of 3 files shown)


druvus:
@bebatut @bgruening @alexeigurevich I just realized that quast is included in `build-fail-blacklist` so those updates were never built. Currently on my mobile so I can't fix it right now.

alexeigurevich:
@druvus Do you know whether quast was removed from the black list? It looks like currently available last version is still 4.5 (see https://bioconda.github.io/recipes.html#recipes)

druvus:
It seems it was not removed so I suggest to make a new PR where you remove it from the black list. I can try to do it during the weekend if you don't have time.

alexeigurevich:
Thanks, I've just do this here: https://github.com/bioconda/bioconda-recipes/pull/6591

PR #6437, opened 2017-10-24
Discasm, adding a new recipe.

tstrnity:


recipes/discasm/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+source:
+  fn: DISCASM-v{{ version }}.tar.gz # [unix]
+  url: https://github.com/DISCASM/DISCASM/releases/download/DISCASM-v{{ version }}/DISCASM-v{{ version }}.tar.gz # [unix]
+  md5: 571b2265018608298752f40f3f375b0f # [unix]
+
+build:
+    number: 0
+    skip: true  # [py3k]
+    skip: True  # [osx]
```
bgruening: If you want to skip both you need to add something like this [osx or py3k]

bgruening: I'm wondering if the py4k skip is really needed.

recipes/discasm/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+cp -R $SRC_DIR/* $PREFIX/bin
+cd $PREFIX/bin
+chmod a+x DISCASM
+pwd
+ls -l
```
bgruening: Can you remove the last two lines?

recipes/discasm/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+        - perl
+        - star >=2.4
+        - trinity >=2.4
+        - oases >=0.2
+
+test:
+    commands:
+        - DISCASM --help > /dev/null 2>&1
+        # DISCASM requires the environment variable TRINITY_HOME to be specified,
+        # but can find it if it is in the PATH variable.
```
bgruening: Can you explain this comment? It not super clear.

recipes/discasm/post-link.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+echo "DISCASM requires TRINITY_HOME to be set to the install directory of trinity."
+echo "This version of DISCASM will look for Trinity in PATH, if TRINITY_HOME is not set."
```
bgruening: If this is the case, it is how it should be and I would remove this file.

recipes/discasm/meta.yaml
tstrnity: The DISCASM code has not been tested on python 3. It works on python 2.7.

recipes/discasm/build.sh
tstrnity: Yes, I will do that.

recipes/discasm/meta.yaml
tstrnity: Sorry. I had written the first part of the comment (the code simply checked for TRINITY_HOME and terminated if it was not defined). But that made it impossible to pass the installation test. I contacted Brian Haas, and then wrote some code to find the TRINITY_HOME directory and set the variable. Then did a pull request and Brian added that code into the distribution. So rather than removing the first comment, I added to it. The code is setting TRINITY_HOME for the other software that is invoked by DISCASM (i.e. trinity).

recipes/discasm/post-link.sh
tstrnity: ok. I will do that.

recipes/discasm/meta.yaml
tstrnity: I will try to make the comment more clear.

recipes/discasm/meta.yaml
tstrnity: Regarding the python 3k skip, if skipping is not the correct way to do that, please let me know. When I recently added a recipe for slncky, I had initially just set a run requirement to be python < 3, and was told that the correct way to do that was to skip the python 3k build. Thanks for any advice you may have about the correct way to do this.

tstrnity: Going ahead and pushing new version with skip [osx or py3k]. Feel free to change it to what ever it should be, if that is not right.

mbargull: Both statements are correct:
- DISCASM needs to skip builds for Python 3 since it uses Python 2-specific code, e.g. `print`-statement.
- In most (not all) cases `python` should not be pinned (see: https://github.com/conda/conda-build/issues/1962) which is also true for DISCASM, so the `skip: True` + selector is correct.

mbargull:
@tstrnity, would you mind being more specific what files should be copied in `build.sh`?
As it currently stands, files with pretty generic names like
- `.gitmodules`
- `Changelog.txt`
- `LICENSE`
- `README.md`
- `testing`

will be copied into `$PREFIX/bin`. `DISCASM.wiki` shouldn't be copied there, too, I suppose.
Overall DISCASM does not adhere to the [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) hence it needs those other folders and vendored dependencies, e.g. `util`, `SciEDPipeR`, to be copied into `$PREFIX/bin`. This isn't optimal but doesn't necessarily prevent to create a conda package.
A workaround could be do copy all those files not into `$PREFIX/bin` directly, but into another, newly created, e.g., `DISCASM-v0.1.2` folder somewhere inside `$PREFIX/` and add a symlink `$PREFIX/bin/DISCASM` to the executable `$PREFIX/other-directory/DISCASM`. Though I'm not sure if `$PREFIX/bin/DISCASM-v0.1.2` or another place should be used. @bgruening, any ideas?

daler:
@tstrnity @mbargull, `$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM` is a good option for location to copy to and symlink from (from http://bioconda.github.io/guidelines.html#java, but isn't specific to Java . . . I should probably add that info somewhere else in the docs)

recipes/discasm/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+test:
+    commands:
+        - DISCASM --help > /dev/null 2>&1
+        # DISCASM requires the environment variable TRINITY_HOME to be set.
+        # If TRINITY_HOME is not set before it starts, it uses "which" to find
+        # Trinity and sets TRINITY_HOME to be the directory where Trinity resides.
+
+about:
+    home: "https://github.com/DISCASM/DISCASM"
+    license: "BSD 3-clause New or Revised License - https://github.com/DISCASM/DISCASM/blob/master/LICENSE"
```
mbargull: Just use
```yaml
license: BSD-3-Clause
```
or
```yaml
license: BSD-3-Clause
license_file: LICENSE
```

recipes/discasm/meta.yaml
```diff
@@ -0,0 +1,37 @@
+{% set version = "0.1.2" %}
+
+package:
+    name: discasm
+    version: {{ version }}
+
+source:
+  fn: DISCASM-v{{ version }}.tar.gz # [unix]
```
mbargull: You can drop the `[unix]` selectors: those are always true since Bioconda only builds packages for Linux and MacOS both of them being "unix".

recipes/discasm/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+  url: https://github.com/DISCASM/DISCASM/releases/download/DISCASM-v{{ version }}/DISCASM-v{{ version }}.tar.gz # [unix]
+  md5: 571b2265018608298752f40f3f375b0f # [unix]
+
+build:
+    number: 0
+    skip: true  # [osx or py3k]
+    # The reason for skipping osx is that trinity >=2.4 is not available in the osx channel
+    # The python code in DISCASM runs on python 2.7.
```
mbargull: Thanks for adding those comments! IMO, those are tremendously helpful!

recipes/discasm/meta.yaml
tstrnity: done

recipes/discasm/meta.yaml
tstrnity: done

recipes/discasm/meta.yaml
tstrnity: @mbargull you're welcome. Also, changed build.sh to incorporate concern about moving everything to bin. I move it to share, as suggested by @daler, but rather than creating a symlink, I create a script that invokes DISCASM using its full path. This is the method used by the star-fusion tool, for reasons similar to the situation with this tool.

recipes/discasm/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+DISCASM_INSTALL_DIR="$PREFIX/share/DISCASM-0.1.2-0"
```
mbargull: If you want, you can literally use `$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM`, as @daler suggested. Then you wouldn't need to change it when you update the recipe to a later version.
Those `$PKG_` variables are available at build time, see https://conda.io/docs/user-guide/tasks/build-packages/environment-variables.html.

recipes/discasm/meta.yaml
mbargull: Right, the script is the way to go since with a symlink would result in `__file__` (inside your Python script) being the path of the link not the destination -- I hadn't thought of that before, good catch!

mbargull: [BTW: (unspoken) convention is to use 2 spaces instead of 4 for the indentation inside the `meta.yaml` -- but that's just nitpicking ;)]

recipes/discasm/build.sh
daler: In case it's useful, here's how to resolve the symlink (so it looks like it's the right directory) in python. https://github.com/bioconda/bioconda-recipes/blob/07aa987a0c9ebe7975bd66afd783026431b86883/recipes/busco/generate-busco-config.py#L25, which avoids hard-coded paths.

tstrnity: Cool. Thanks.

tstrnity: I just scanned through the code to see if that is done. It is done for some, but not all situations. Not sure why Brian did use realpath some places, but not others. If it gets changed in the future, I will be able to change the shell scripts to symlinks.

bgruening:
Top @tstrnity!

PR #6439, opened 2017-10-24
Added ssu-align 0.1.1

mworkentine:


recipes/ssu-align/post-link.sh
```diff
@@ -0,0 +1,22 @@
+echo "
+ssu-align has been installed.  To run it properly please set the
+environment variable SSUALIGNDIR using one the of the following
```
bgruening: @mworkentine why is this needed? Can this program not simply use the pwd from the the binary?

ar0ch: SSUAlign is weird and uses that ENVVAR to find libraries instead of the binary path since the install lets you install them to different places if you want

mworkentine:
Yes, it’s not really ideal at all. I wasn’t sure how to deal with this. Do you know if there are other bioconda packages that have dealt with similar situations?  In the mean time I’ll have a closer look at the source to see if it might be patched.

recipes/ssu-align/post-link.sh
bgruening: Can we include a very tiny bash wrapper for both binaries that do the export and forward all parameters to the original binary. So something like:

* rename original binary to PREFIX/bin/_ssu-align
* create a wrapper called PREFIX/bin/ssu-align that calles _ssu-align but exports the ENV before

ar0ch:
Alternatively you can patch the perl scripts:

```
my $ssualigndir = $ENV{'SSUALIGNDIR'};
```

```
my $ssualigndir = $PREFIX/share/ssu-align-${version};
```

bgruening:
@ar0ch PREFIX might not be available in the runtime env.

ar0ch:
@bgruening Should be available to build.sh for patching though

bgruening:
@ar0ch yes, during build time yes. But then you end up with a path in this file, which is wrong during runtime, right?

ar0ch:
If its patched during build (either a diff just sed the whole line), that path will be correct during runtime

mworkentine:
@bgruening I went with your suggestion for wrapper scripts which are created on the fly in build.sh.

PR #6453, opened 2017-10-25
Updated the receipe for Tablet

imilne:

Minor changes (and an md5 update) to support a remote package with an improved launch script

### DIFF
--- a/recipes/tablet/meta.yaml
+++ b/recipes/tablet/meta.yaml
@@ -1,5 +1,6 @@
 {% set name = "tablet" %}
 {% set version = "1.17.08.17" %}
+{% set build = "1" %}
 
 package:
   name: {{ name }}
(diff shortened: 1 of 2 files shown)


recipes/tablet/meta.yaml
```diff
@@ -1,17 +1,18 @@
...
 
 package:
   name: {{ name }}
   version: {{ version }}
 
 source:
   fn: {{ name }}-{{ version }}.zip
-  url: http://bioinf.hutton.ac.uk/tablet/installers/bioconda/{{name}}-{{version}}.zip
-  md5: 39c421c85df941a6d27c7755ea9ad55b
+  url: http://bioinf.hutton.ac.uk/tablet/installers/bioconda/{{name}}-{{version}}-{{build}}.zip
```
bgruening: build numbers should be used when the tarball stays the same, but the recipe changes. E.g. new/updated dependencies. You should not change the tarball in one tool version-cycle.

peterjc: Build number now also bumped from 0 to 1.

imilne: If the bundle changes, that changes the md5, which as part of the recipe would mean a build increment, no? Or is it ok to update the bundle (for whatever reason), update the md5, but leave everything else alone?

peterjc: Updating the upstream bundle (for the same version number) and rebuilding the BioConda package would give a different package, which should be distinguished by an incremented build number.

imilne: So what I've done is fine then?

peterjc: I thought this pull request was fine (which is why I merged it), but best to have this confirmed by @bgruening for future reference.

imilne: Still learning so please bear with me. In this case the tool (and its version) hasn't changed, but the (bioconda-specific) launch script that we've had to develop for it has. This script is included in the bundle, which obviously changes the md5 when we update it. I'm happy to go with whatever the recommendation is for dealing with this situation, be it incrementing build numbers or otherwise?

PR #6462, opened 2017-10-25
Updating Feht to 1.1.0

hellothisisMatt:

Unfortunately, we still have to use the precompiled executable as building the source code results in a linker error. This is yet to be resolved.


recipes/feht/meta.yaml
```diff
@@ -12,7 +13,8 @@ source:
   md5: '{{ hash_value }}'
 
 build:
-  number: 0
+  number: 1
```
druvus: Please set build number to 0 for new versions.

hellothisisMatt: Ah, thanks for catching that!

bgruening:
LGTM!

PR #6463, opened 2017-10-25
Cleanup flaimapper

mbargull:

This is just a cleanup/fix for the flaimapper recipe.

@yhoogstrate:
1. Were there any reasons you added `numpy` as a run/build and `htseq` as a build requirement? FlaiMapper does not seem to use either of those, right?!
2. Are there reasons to keep the older version explicitly? Please note this should only be needed in special cases, see: http://bioconda.github.io/guidelines.html#versions
3. Can you give an explanation why `pysam` is pinned to `==0.11`? In `setup.py` there is `'pysam >= 0.8.0,<0.9'`. Could we use `>=0.8.0,<0.9|>=0.10` or are `0.10*` and `>=0.12` not compatible?

@yhoogstrate / @druvus:
- I re-added `setuptools` as a run requirement since `flaimapper.Data` imports `pkg_resources` which is part of `setuptools`. Any objections?


yhoogstrate:
Back then I had some problems in getting the wrapper working and only ofter some hacks it did. 

1. Numpy (and HTSeq als built) - I think this is there for historical reasons. I have switched libraries over time. This could go (tested it a few minutes ago). I suppose I only got it working by including them at some point in time but your changes do install on the current infrastructure it's even better.
2. I am fine with removing old versions for FlaiMapper but (for other tools) I have been requested not to overwrite (https://github.com/bioconda/bioconda-recipes/pull/5117). Please use the bioconda policy to make sure we don't get any punishment :wink: - I don't know whether this is keeping old files or patching them.
3. Main reason: because pysam is/was a backwards and forwards incompatible mess. When I made the wrapper, 0.12 didn't exist and I simply didn't want to trigger myself testing it after every update (https://github.com/yhoogstrate/flaimapper/issues/17, http://pysam.readthedocs.io/en/latest/release.html#release-0-8-1). Maybe I am a bit oldskool but I've learned that major version changes have compatibility issues and therefore I prefer hard version management unless there is a thorough testing procedure. I think after pysam 0.10 compatibility issues it got better but I still wouldn't mind staying stick to one version because I don't trust pysam updates. I tried to include tests in bioconda (just the flaimapper built-in functional tests: `python setup.py install, nose2 src/tests/*.py`) but if I remember correctly, the source code was removed after the built process and before running the tests. I didn't find an elegant solution so I decided not to continue with the tests and hard-defined the pysam version.

I have no objectoins with any of your changes. Once they you push/merge, I will test it with pleasure and I want to thank you for this work so far!

all best,

Youri

bgruening:
Given Youris approval I will merge despite the different pysam reqs.
Thanks all!

PR #6471, opened 2017-10-25
ProPhyle 0.3.0.0

karel-brinda:

### DIFF
--- a/recipes/prophyle/0.1.0/meta.yaml
+++ b/recipes/prophyle/0.1.0/meta.yaml
@@ -23,7 +23,7 @@ requirements:
     - bitarray
     - scipy
     - wheel
-    - zlib
+    - zlib {{CONDA_ZLIB }}*
     - psutil
   run:
     - python
@@ -32,7 +32,7 @@ requirements:
     - bitarray
     - scipy
     - wheel
-    - zlib
+    - zlib {{CONDA_ZLIB }}*
     - psutil
 
 test:
--- a/recipes/prophyle/0.2.0/meta.yaml
+++ b/recipes/prophyle/0.2.0/meta.yaml
@@ -23,7 +23,7 @@ requirements:
     - bitarray
     - scipy
     - wheel
-    - zlib
+    - zlib {{CONDA_ZLIB }}*
     - psutil
     - pysam
   run:
@@ -33,7 +33,7 @@ requirements:
     - bitarray
     - scipy
     - wheel
-    - zlib
+    - zlib {{CONDA_ZLIB }}*
     - psutil
     - pysam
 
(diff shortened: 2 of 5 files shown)


karel-brinda:
@bioconda/core Do you have any idea why this doesn't work?

1. The version should be 0.3.0.0, see https://github.com/bioconda/bioconda-recipes/blob/0ad02ce2dffc46cc1eea5331ebaaa24b4d29ba50/recipes/prophyle/meta.yaml#L1
2. Zlib is pinned, see https://github.com/bioconda/bioconda-recipes/blob/0ad02ce2dffc46cc1eea5331ebaaa24b4d29ba50/recipes/prophyle/meta.yaml#L40

daler:
@karel-brinda looks like this commit additionally has a 0.2.1.3 version in the 0.2.1 subdirectory, and that's what the linter is complaining about. If you want to re-build the older version, you can fix the zlib pinning in 0.2.1.3 (and maybe change the subdirectory to match the version). If you just want to keep the older built package on anaconda without rebuilding the recipe, you can omit the older recipe.

karel-brinda:
@daler I used to keep old versions of recipes in the subdirectories and thought that it was a recommended practice. So I should update all the old recipes if I want to keep them? Are they going to be rebuilt?

daler:
Coincidentally we were just talking about this over at https://github.com/bioconda/bioconda-recipes/pull/5117#issuecomment-339692774 (which references http://bioconda.github.io/guidelines.html#versions). The idea is that old recipes will always be available in the git history, so generally no need to hold on to them in subdirs. Exceptions of course are when an older recipe is broken for some reason; a fixed recipe can be added to a subdir while the latest version remains in the top-level dir.

karel-brinda:
@daler I've completely missed this major update of Bioconda policy. I usually kept the last minor version of every "edition" of programs (e.g., 0.2.0.3 for 0.2.0.x). I am a little bit afraid that some useful information might get lost if we don't keep this type of "version" directories.

Is it going to be replaced by some database mapping conda packages builds to specific dirs+commits in this repo (e.g., a link to meta.yaml)? How can I easily find out what recipe was used to build individual versions of packages?

daler:
@karel-brinda policy-wise we haven't been enforcing this in any way, it's not that big of a deal. But the easy answer to your question is that each package tarball in the channel has a `recipe` contains the recipe it was built from. As for tying it to the git history, I mentioned in https://github.com/bioconda/bioconda-recipes/pull/5117#issuecomment-339697632 we could build a tool that figures out the git commit given the package tarball and the git repo.

karel-brinda:
@daler Oh, I see. I didn't know that the recipes were included in the packages. In this case, I can imagine that people could keep just recipes for last versions and, in addition to that, there could an automatically generated Github repository with all the recipes for all the versions of all the binaries. Then it would be easy to find a recipe for a specific version of every package. Such a repo would be a great database, e.g., for people interested in automatic conversion of Conda recipes to other build systems.

PR #6479, opened 2017-10-26
Updated yara version

gurgese:


recipes/seqan_tcoffee/meta.yaml
```diff
@@ -0,0 +1,23 @@
...
+  home: https://github.com/seqan/seqan/blob/master/apps/seqan_tcoffee/README
+  license: https://github.com/seqan/seqan/blob/master/apps/seqan_tcoffee/LICENSE
+  summary: SeqAn::T-Coffee - Multiple Sequence Alignment.
+
+package:
+  name: seqan_tcoffee
+  version: {{ version }}
+
+source:
+  url: http://packages.seqan.de/seqan_tcoffee/seqan_tcoffee-{{ version }}-Mac-x86_64.zip # [osx]
```
bgruening: Can this be compiled, instead of downloaded?

gurgese: Yes, it can be also compiled. 
Do you prefer the compiled version? 
If so I can update the yaml and run a new pull request

bgruening: Yes, the source version is preferred. You can just use this branch and commit the updated yaml and build.sh script.

gurgese:
Dear @bgruening 
I updated my commit with the src version of seqan_tcoffee. 
However, for compiling the latest release (that make use of seqan-v2.3.2) a version of gcc supporting c++14 is required.
I tried to set up manually some flags for the compiler but the ./simulate-travis.py script ends always with some errors.
Do you have some suggestions or some build.sh file to be used a s reference?

PR #6490, opened 2017-10-26
Loompy recipe

gioelelm:

### DIFF
--- a/recipes/loompy/meta.yaml
+++ b/recipes/loompy/meta.yaml
@@ -1,0 +1,50 @@
+{% set name = "loompy" %}
+{% set version = "1.0.2" %}
+{% set hash_type = "sha256" %}
+{% set hash_value = "36159c7d1423d4beb9848efc17ca5b2893398f1c7a6dccb408a413a112134c2b" %}
+
+package:
+  name: '{{ name|lower }}'
+  version: '{{ version }}'
+
+source:
+  fn: '{{ name }}-{{ version }}.tar.gz'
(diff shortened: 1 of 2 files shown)


recipes/loompy/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+source:
+  fn: '{{ name }}-{{ version }}.tar.gz'
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  '{{ hash_type }}': '{{ hash_value }}'
+
+build:
+  skip: True  # [not py36]
+  number: 0
+  script: python setup.py install  --single-version-externally-managed --record=record.txt
```
bgruening: If you include the this line, you don't need the script file.

gioelelm: Ok, should I remove the script? Should I just commit the change in the forked repo or make a new pull request?

bgruening: Yes please remove the build.sh script and commit it into the same branch, the PR will then be updated automatically.

gioelelm: Done.
A question slightly OT, now that python3.7 is being released, probably [not py36] is not a good preprocessing tag to mean python>=3.6, right ? maybe [py < "36"] or [int(py) < 36] are more correct?

bgruening: Normally you just write [py36 or py37]. Not sure if your examples will work, never tried it.

PR #6506, opened 2017-10-27
Add backspinpy from PyPI.

jdblischak:

Some notes for reviewers:

* [backspinpy](https://pypi.python.org/pypi/backspinpy) implements the [BackSPIN](https://github.com/linnarsson-lab/BackSPIN) algorithm for clustering single cells based on their gene expression levels.
* I created the recipe using conda-build 3.0.19.
* The package exports a command-line tool, `backspin`, so I added a test for this.
* The license situation is confusing. In the `setup.py` it says the license is MIT ([link](https://github.com/linnarsson-lab/BackSPIN/blob/406c150e0e29979b06bdd418b9680186333e128d/setup.py#L21)); however, the file [LICENSE](https://github.com/linnarsson-lab/BackSPIN/blob/406c150e0e29979b06bdd418b9680186333e128d/LICENSE) in the repo is BSD-2. I tried to manually add the license file using the [conda-forge trick](https://conda-forge.org/docs/meta.html#packaging-the-license-manually) of `'{{ environ["RECIPE_DIR"] }}/LICENSE'`, but I received an error stating that the variable `environ` was not defined.
* The recipe was [built successfully](https://travis-ci.org/jdblischak/bioconda-recipes/builds/293645416) on my fork.
* I tried adding `noarch: python` to the recipe since it only contains Python files, but this [build failed](https://travis-ci.org/jdblischak/bioconda-recipes/builds/293666125) on linux but passed on macOS. It's unclear to me what a "pure python package" is anyways. The recipe depends on numpy/scipy, so I assume that means a pure python package cannot have dependencies which require compilation.

Unrelated to this specific recipe: I get strange errors on the Travis build when I modify my commit history and force push to the feature branch on my fork. For example, I removed the commit that tried `noarch` and pushed the change back to GitHub, and the second time around my build failed (even though it was the exact same commit that had just passed). Is there anyway to avoid/fix this? I like to overwrite my history to keep the commit history clean when I submit a PR.

https://travis-ci.org/jdblischak/bioconda-recipes/jobs/293666127#L821

### DIFF
--- a/recipes/backspinpy/meta.yaml
+++ b/recipes/backspinpy/meta.yaml
@@ -1,0 +1,49 @@
+{% set name = "backspinpy" %}
+{% set version = "0.2.1" %}
+{% set hash_type = "sha256" %}
+{% set hash_value = "91f8b6b3d5acb318844d8ecdd4b67bf8c37a19fa129466cebf2083dc0fc5476c" %}
+
+package:
+  name: '{{ name|lower }}'
+  version: '{{ version }}'
+
+source:
+  fn: '{{ name }}-{{ version }}.tar.gz'
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  '{{ hash_type }}': '{{ hash_value }}'
+
+build:
+  number: 0
+  script: python setup.py install  --single-version-externally-managed --record=record.txt
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - future
+    - numpy
+    - scipy
+    - scikit-learn
+  run:
+    - python
+    - future
+    - numpy
+    - scipy
+    - scikit-learn
+
+test:
+  imports:
+    - backspinpy
+
+  commands:
+    - backspin -h
+
+about:
+  home: https://github.com/linnarsson-lab/BackSPIN
+  license: MIT
+  license_family: MIT
+  summary: backSPIN clustering algorythm
+
+extra:
+  recipe-maintainers:
+    - jdblischak


daler:
Thanks @jdblischak. 

https://travis-ci.org/jdblischak/bioconda-recipes/jobs/293666127#L904 seems to suggest you need a `chmod +x` on `$PREFIX/bin/backspin`. I would have thought the `setup.py` would have handled it, but maybe use a separate `build.sh` and make that change?

As for `noarch`, there are compatibility issues with conda that we're still working out, not too surprised it didn't work.

I haven't used that `'{{ environ["RECIPE_DIR"] }}/LICENSE'` before, I wonder if it's only supported in conda-build-3 (which we are not supporting just yet)?

jdblischak:
> https://travis-ci.org/jdblischak/bioconda-recipes/jobs/293666127#L904 seems to suggest you need a chmod +x on $PREFIX/bin/backspin. I would have thought the setup.py would have handled it, but maybe use a separate build.sh and make that change?
>
> As for noarch, there are compatibility issues with conda that we're still working out, not too surprised it didn't work.

@daler That error message was only when I tried to specify `noarch`. If `noarch` isn't working in general, I plan to leave the recipe as is for now.

> I haven't used that '{{ environ["RECIPE_DIR"] }}/LICENSE' before, I wonder if it's only supported in conda-build-3 (which we are not supporting just yet)?

I don't think that is the issue. conda-forge is also very conservative with updating conda-build (they are using 2.1.17).

https://anaconda.org/conda-forge/conda-build

Any insight in the error about `git diff --relative` that I receive when amending commits. It makes it very hard for me to iteratively test changes to a recipe.

https://travis-ci.org/jdblischak/bioconda-recipes/jobs/293666127#L821

daler:
I think you mean a different build, https://travis-ci.org/jdblischak/bioconda-recipes/jobs/293733109?

I'm not sure what the best fix would be for that, but I'm pretty sure checking out a branch from bioconda-recipes (rather than a fork) will fix it. You can do a squash merge at the end once it's passing to keep the history clean.

jdblischak:
> I think you mean a different build, https://travis-ci.org/jdblischak/bioconda-recipes/jobs/293733109?

@daler Yes, sorry about that. I had a bunch of builds open and must have copied the wrong one. But yes, I get errors from `git diff --relative` on my fork.

https://travis-ci.org/jdblischak/bioconda-recipes/jobs/293733109#L781

If this happens again, I'll change my strategy to cloning/branching instead of forking. Thanks!

daler:
@jdblischak actually, I think the problem is that travis does a shallow clone of the repo such that the build infrastructure would need to do a `git fetch <commit>` to fetch whatever commits were provided as part of the range argument. We could probably provide something along these lines in https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/utils.py#L879.

Anyway, we've streamlined the building to reduce the load on travis-ci, so clone/branch works pretty well.

bgruening:
@jdblischak thanks for you contribution!

PR #6515, opened 2017-10-27
Update Unicycler - needs offline build

ar0ch:

### DIFF
--- a/recipes/unicycler/meta.yaml
+++ b/recipes/unicycler/meta.yaml
@@ -7,7 +7,9 @@ package:
 
 build:
   number: 0
-  skip: True # [py27]
+  skip: True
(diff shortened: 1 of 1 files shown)


recipes/unicycler/meta.yaml
```diff
@@ -1,18 +1,18 @@
...
+{% set version = "0.4.2" %}
 
 package:
   name: {{ name|lower }}
   version: {{ version }}
 
 build:
   number: 0
-  skip: True
+  skip: True # [py27]
```
bgruening: Now this will be only skipped for py27. You want to skip it for everything - so that I can build it offline. Maybe add a comment, why you skip the build and that this is build offline for the time being.

PR #6517, opened 2017-10-27
Fgbio should be supported in any python version

nh13:
@chapmanb I think you added fgbio originally (thank-you!).  I am not sure why it requires python 2.7 only (I want python 3 support); perhaps the included script to run `fgbio`?  It feels odd to me to depend on a specific version of python when `fgbio` itself does not.  An alternate solution would be to (picard/meta.yaml)[https://github.com/bioconda/bioconda-recipes/blob/master/recipes/picard/meta.yaml]

### DIFF
--- a/recipes/fgbio/meta.yaml
+++ b/recipes/fgbio/meta.yaml
@@ -9,7 +9,7 @@ package:
     version: {{ version }}
 build:
   number: 0
-  skip: False
+  skip: false
 source:
   fn: fgbio-{{ tag }}.tar.gz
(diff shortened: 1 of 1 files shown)


chapmanb:
Nils -- thanks for this. I believe I did py27 only at the start due to space issues, since having py3 creates 3 copies of all the jars in anaconda.org. The long term solution to this is meant to be noarch but I'm not sure where that support is currently at. Happy to merge this in the meantime to not exclude use of py3. Thanks much for the PR.

PR #6548, opened 2017-10-30
Add freetype to fastqc deps

### LINKED ISSUE #5026: Missing ttf-dejavu/fontconfig dependencies for fastqc
If fontconfig or ttf-dejavu are not installed, fastqc can produce the following error:

```
java.lang.Error: Probable fatal error:No fonts found
```

which is identical to the error described here: https://stackoverflow.com/questions/14066525/java-lang-error-probable-fatal-errorno-fonts-found. This issue is at least encountered in the snakemake container used by snakemake-wrappers.

I think this error is not necessarily specific to fastqc, but likely a more general dependency of openjdk which is currently missing. 

We can easily fix it for fontconfig by adding the dependency to the fastqc or openjdk recipes. However, for ttf-dejavu there does not seem to be an existing package as far as I can see. I have no experience with creating non-R/Python packages, so was wondering if someone would be prepared to build a ttf-dejavu package or otherwise provide me with pointers on how to do so.

epruesse:

Should fix #5026

### DIFF
--- a/recipes/fastqc/meta.yaml
+++ b/recipes/fastqc/meta.yaml
@@ -27,6 +27,9 @@ source:
     - java_xms.patch
 
 test:
+  files:
+    - test.fq.gz
   commands:
     - fastqc -h >/dev/null
     - fastqc --version
+    - fastqc -o tmpdir test.fq.gz && rm -rf tmpdir
--- a/recipes/fastqc/meta.yaml
+++ b/recipes/fastqc/meta.yaml
@@ -6,13 +6,14 @@ about:
 
 build:
   detect_binary_files_with_prefix: true
-  number: 2
+  number: 3
   skip: False
 
 requirements:
   run:
     - openjdk
     - perl
+    - freetype 2.7
 
 package:
   name: fastqc


recipes/fastqc/meta.yaml
```diff
@@ -6,13 +6,14 @@ about:
...
   detect_binary_files_with_prefix: true
-  number: 2
+  number: 3
   skip: False
 
 requirements:
   run:
     - openjdk
     - perl
+    - freetype 2.7
```
bgruening: This is strange. fastqc is just a Java program and should not depend on it. Is it missing for openjdk or perl?

epruesse:
The conda-forge openjdk package is "missing" a number of library dependencies for "full" functionality. These are the freetype lib needed for font rendering as done by fastqc and a bunch of X11 libs for screen rendering. Since the JVM works fine without those, it's a matter of taste whether they need to be included by packages using Java or Java itself, potentially though meta-packages (say openjdk-x11, openjdk-fonts). To get fastqc working asap I'm just adding it here. 

I did post an issue at conda forge a while ago ... see https://github.com/conda-forge/openjdk-feedstock/issues/19

epruesse:
@bgruening Can I merge this? It should fix the issue in the container.

bgruening:
Go for it. I'm still not a fan of this and think this should go upstream to open.jdk - but its important to fix it now.

mbargull:
How about adding a short explanatory comment to that dependency?

epruesse:
Hmm. This is weird. Technically, the package shouldn't need freetype. conda-forge::openjdk uses the azul/zulu java packages which bundle the required libs including freetype. The linking seems to be ok too. So in theory it should work without.

That said, I did have the problem in my pipeline in the unit tests and adding freetype fixed it there...

epruesse:
Ok, looks like the quay container is using the defaults channel java, rather than the conda-forge one. 
Closing this for now.

PR #6554, opened 2017-10-31
added R and UpSetR dependency to Intervene

asntech:


recipes/intervene/meta.yaml
```diff
@@ -48,8 +48,6 @@ test:
     - intervene.modules.pairwise
     - intervene.modules.upset
     - intervene.modules.venn
-  commands:
-    intervene --help
```
mbargull: There was just a typo, i.e., missing dash, since `commands` is supposed to be a list. Use:
```yaml
  commands:
    - intervene --help
```
If the program has an entry point `intervene`, then including it in `test/commands`, like you did before, was indeed a good thing to do.

asntech: @mbargull thanks for catching the typo.

PR #6557, opened 2017-10-31
Add Mummer4 recipe

ar0ch:

This PR adds [mummer4](https://mummer4.github.io), which is partially backwards compatible with mummer3 syntax but includes niceties like multithreaded nucmer, out of the box 64bit support (any sized genome), read-based analysis and SAM output.

This does conflict with the `mummer` recipe that exists, in that the binaries have the same name.  I'm adding a new, `mummer4` recipe instead of updating the existing mummer recipe since the build process has changed and requires newer libgcc, which may not be compatible with existing environments.

### DIFF
--- a/recipes/mummer4/build.sh
+++ b/recipes/mummer4/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+./configure --prefix=$HOME
+make
+make install
--- a/recipes/mummer4/build.sh
+++ b/recipes/mummer4/build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-./configure --prefix=$HOME
+./configure --prefix=$PREFIX
 make
 make install
--- a/recipes/mummer4/meta.yaml
+++ b/recipes/mummer4/meta.yaml
@@ -1,0 +1,29 @@
+{% set name = "mummer4" %}
+{% set version = "4.0.0beta2" %}
+
+about:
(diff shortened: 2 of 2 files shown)


bgruening:
Looks good to me! ping @peterjc 

> This does conflict with the mummer recipe that exists, in that the binaries have the same name. I'm adding a new, mummer4 recipe instead of updating the existing mummer recipe since the build process has changed and requires newer libgcc, which may not be compatible with existing environments.

The binaries still have the same name, right? So you can still not mix both versions in one environment.

ar0ch:
Yes, the binaries have the same names as mummer3

peterjc:
What does "partially backwards compatible" mean? Adding new stuff is fine, but do they remove anything?

It is likely that anyone would need mummer3 and mummer4 installed at the same time (as does happen sometimes with hmmer2 and hmmer3), in which case disambiguated binary names might helpful?

ar0ch:
It doesn't appear all the old mummer3 options have been ported over yet.  At least, the help text for the `mummer` binary: 

> Implemented MUMmer v3 options...

makes it seem like its not 100% backwards compatible?

peterjc:
We can come back and add an extra renamed-binary package (as done for hmmer2) later if there is demand. I have no objection to merging this.

bgruening:
Thanks @peterjc for the quick response. Thanks @ar0ch for the recipe!

PR #6561, opened 2017-10-31
Du Novo: Update to 2.0.8.

NickSto:
Update to install Du Novo 2.0.8 (a bugfix release).
Also, change networkx dependency to version 1.10 to allow compatibility with Galaxy Toolshed (which lacks a 1.11).

### DIFF
--- a/recipes/dunovo/build.sh
+++ b/recipes/dunovo/build.sh
@@ -2,10 +2,10 @@
 
 # Download submodules and move them to lib.
 ## kalign
-kalignver=0.1.2
+kalignver=0.1.3
 wget --no-check-certificate "https://github.com/makrutenko/kalign-dunovo/archive/v$kalignver.tar.gz"
 hash=`sha256sum v$kalignver.tar.gz | tr -s ' ' | cut -d ' ' -f 1`
-[ $hash = 398f699491b2aa607c32062007228a1de853e2f665340e82dc1fc967b0ee8d85 ]
+[ $hash = 712b149038a818d296d0bcc53e997487c768a151e46e500f1753ba6c3dd9852d ]
 tar -zxvpf v$kalignver.tar.gz
 rm -rf kalign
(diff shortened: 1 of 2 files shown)


bgruening:
> Galaxy Toolshed (which lacks a 1.11)

Are you referring to `packages` here? Packages are depreacted and you should not rely on them. Use in your conda package what ever fits your application and Galaxy will use it.

NickSto:
Oh, I'm not sure what you mean by `packages`. I think the issue was that my tool `<requirements>` originally included networkx 1.9, since there was no `package_networkx_1_11` in the Toolshed. Then later I changed the conda `meta.yaml` to say networkx 1.11, since it exists in conda. But I didn't realize that means Du Novo would be installed via conda with networkx 1.11, but the tool says it requires 1.9, so there was a conflict. 1.10 exists in both, so it resolves the conflict.

bgruening:
> Oh, I'm not sure what you mean by packages

-> package_networkx_1_11 ;)

Completely, forget about those packages. They are not needed anymore and you don't need to care about them. What I'm saying don't change this recipe just because such a package is missing in the TS.

NickSto:
Oh okay, that's what I thought.

But if I specify 1.11 in conda and 1.9 in the tool xml, I wasn't able to actually run the tool because of the version conflict. I installed via conda and told Galaxy to resolve all dependencies with conda. Was I doing something wrong?

bgruening:
Do you have the error message?

NickSto:
Oh, not on hand. Let me reproduce it.

Btw, for now, could I just remove the version number in `meta.yaml`? That's how it was when it was working and before I started messing with it.

Maybe later I can PR a working version with a version number specified.

NickSto:
..and I'd really like to get this version out :D

bgruening:
Just pinned it to <2.0.

NickSto:
Ohh yeah that works!

PR #6562, opened 2017-11-01
Add Recipe smctc

dfornika:

This is a dependency for [sts](https://github.com/bioconda/bioconda-recipes/pull/5855). I'm submitting it here (not conda-forge) because it's a non-standard fork of the [canonical library](https://www2.warwick.ac.uk/fac/sci/statistics/staff/academic-research/johansen/smctc/)


cbrueffer:
Hi Dan, this looks non-biology specific and may be more appropriate for conda-forge? Even the summary says it's "very general" :-) What do you think?

cbrueffer:
Sorry, missed the text in the PR where you actually mention why you submit it here.

bgruening:
Can you add the info about the non-standard fork as extras/notes or somewhere in the code as comments?

cbrueffer:
I think it would be good to at least rename the package so it doesn't block the smctc name for the canonical package (maybe sts-smctc or similar?), and also give the homepage of the fork and not the original package. Otherwise it's very easy to confuse the two.

dfornika:
Thanks for the feedback. I'm going to have to prepare a similar package for `lcfit` which is also used in `sts`.

https://github.com/OnlinePhylo/sts/tree/master/lib
https://github.com/matsengrp/lcfit

johanneskoester:
Thanks!

PR #6572, opened 2017-11-01
Add parsnp

Cashalow:

### DIFF
--- a/recipes/parsnp/build.sh
+++ b/recipes/parsnp/build.sh
@@ -1,11 +1,4 @@
 #!/bin/bash
-
-
-if [[ `uname` == Darwin ]]; then
-    export CC=gcc
-    export CXX=g++
-fi 
-
 ./autogen.sh
 ./configure --prefix=${PREFIX}
 make
--- a/recipes/parsnp/build.sh
+++ b/recipes/parsnp/build.sh
@@ -3,3 +3,4 @@
 ./configure --prefix=${PREFIX}
 make 
 make install
+mv bin/parsnp ${PREFIX}/bin/ 
--- a/recipes/parsnp/build.sh
+++ b/recipes/parsnp/build.sh
@@ -3,6 +3,5 @@
 ./configure --prefix=${PREFIX}
 make
 sed -i.bak '1i#!/usr/bin/env python\n' Parsnp.py
-sed -i.bak 's/sys\.stderr\.write(\ BOLDME/sys\.stdout\.write(\ BOLDME/' Parsnp.py
 mv Parsnp.py parsnp
 /usr/bin/install parsnp ${PREFIX}/bin/
--- a/recipes/parsnp/build.sh
+++ b/recipes/parsnp/build.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+./autogen.sh
+./configure --prefix=${PREFIX}
+make
+mv src/parsnp ${PREFIX}/bin/
+
--- a/recipes/parsnp/build.sh
+++ b/recipes/parsnp/build.sh
@@ -6,4 +6,21 @@ make install
 ls -Srthl src/*
 ls -Srthl bin/*
 
-mv src/parsnp ${PREFIX}/bin/ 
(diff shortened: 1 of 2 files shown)


Cashalow:
The osX build fails because clang is not compatible with openMP for conda-forge version apparently ? I have removed the osX build because I'm not sure about how to fix this.

recipes/parsnp/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+./autogen.sh
+./configure --prefix=${PREFIX}
+make
+sed -i.bak '1i#!/usr/bin/env python\n' Parsnp.py
+sed -i.bak 's/sys\.stderr\.write(\ BOLDME/sys\.stdout\.write(\ BOLDME/' Parsnp.py
```
mbargull: Can you give any information as to why changing the program's behavior this way is necessary?

Cashalow: For the second sed: whenever the program used to be called, the version and the link to the documentation would be sent to stderr, prompting any test to fail for instance.

For the first one: I thought this was standard procedure for directing the system to use python for interpreting the script files.

mbargull: But your change will alter the program's output (on `stdout`). I don't believe this should be done.
The `stderr` stream is commonly used not only for outputting errors but also logging information. Hence I believe the original usage is totally legitimate. If this caused any tests to fail, then those tests are likely broken

Cashalow: Sorry you are right I confused to things at the same time. It was calling parsnp without any arguments that throws an error code, I'll remove the second sed.

mbargull: Great, thanks!

Cashalow:
Please don't merge this yet. I am still working on it and something is not quite right. Is it okay if I add to the recipe a few test files (one genome as gbk and one as fasta ) so that I can check that parsnp is running as expected ?

mbargull:
> Is it okay if I add to the recipe a few test files

If the build tests are *very lightweight* (both, in terms of runtime as well as storage), that's fine.
No matter if you add the test data or not, I'd like to refer to
http://bioconda.github.io/contribute-a-recipe.html#test-locally
for the early stages of creating a recipe. Testing locally has the benefit that one can inspect the created Conda packages from the local build cache (i.e., `conda-bld` folder of the Miniconda installation) manually or even install it with `conda [create|install] -n test-env --use-local package-name`.

bgruening:
@Cashalow any help needed?

Cashalow:
Thanks a lot for asking @bgruening !

I had more or less given up on adding parsnp on bioconda because the version that is shipped in the github, that I tried compiling, is actually different from the binaries made available on the website ( http://harvest.readthedocs.io/en/latest/content/parsnp/quickstart.html)

Basically I ran into the exact same problem as mentioned on this thread when somebody tried to add it to homebrew: 

https://github.com/marbl/parsnp/issues/3

Given that the software as not been updated for almost 3 years now, that the last time the author answered to an issue on the github was more than one year ago, I basically gave up understanding how to build it correctly.

The alternative would be to add directly the binary available on the website to bioconda but I see little added value for that... What do you think ?

bgruening:
If it is not needed lets close this recipe.
It is ok to add binaries to bioconda, but we prefer to compile from source for several reasons. I guess we can make an exception here if you need this package. Your call.

Just to understand it correctly, your test in this PR is not sufficient and it actually is broken what you currently have?

Cashalow:
The binary that I obtain in this PR is working, the only problem is that it does not correspond at all to the binary distributed : it is not called the same way, needs a config file, and there is no information as to how to use this binary. 

I understood from the homebrew-related thread that I linked that the final distributed binary was compiled using pyinstaller but I did not managed. 

As it stands, the binary I obtain is useless.

Cashalow:
Closing now. I got the software to work more or less as intended but the compile binary now throws Segmentation errors and I don't really know why. I'm not sure I see the point of adding the frozen binary so we will leave it like that for the time being.

PR #6585, opened 2017-11-02
added recipe for telseq

zd1:

### DIFF
--- a/recipes/telseq/build.sh
+++ b/recipes/telseq/build.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+set -x
+set -e
+
+pushd $SRC_DIR/src
+bamtools=$(conda list bamtools -c | awk -F'::' '{print $2}')
+./autogen.sh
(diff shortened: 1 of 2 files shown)


recipes/telseq/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -x
+set -e
+
+pushd $SRC_DIR/src
+bamtools=$(conda list bamtools -c | awk -F'::' '{print $2}')
+./autogen.sh
+./configure --prefix=$PREFIX --with-bamtools=$SYS_PREFIX/pkgs/$bamtools  && make && make install
```
bgruening: This should be $PREFIX/lib or $PREFIX/include, isn't it?

zd1: Thanks @bgruening for looking into this. lib and include subdirectories are taken care of inside the configure, which is built by autoconf with configure.ac. 

I followed the tutorial of contributing a package but unfortunately couldn't manage to build  a workflow of my own. In the end I decided to postpone this and use docker for now. This pull request can be dropped. 

Thanks.

PR #6589, opened 2017-11-02
Pinned boost to 1.61 to prevent rebuilds against incompatible versions

jamesabbott:


druvus:
@jamesabbott abyss is included in the ´build-fail-blacklist´ file so you would need to remove it there to test the updated recipe.

bgruening:
@jamesabbott you might need to pin other versions as well.

jamesabbott:
WIll do...hadn't realised some versions were already blacklisted...

bgruening:
This is interesting, in your Docker testing we get: `/bin/sh: abyss-pe: not found`

jamesabbott:
Indeed...I don't think all those abyss versions were blacklisted due to the boost problem since they are borked even with boost 1.61. The release notes for abyss 2.0.2 specifically mention fixes to build againt boost >1.62. Strangely it seems to build fine using simulate_travis with --disable-docker, but not when using  docker so it looks like something to do with the build environment.
I've been trying to get to the bottom of this but 'real-life/job' have been getting in the way...

daler:
I think the fix is to wrap the makefile, but I was unable to figure that out -- see https://github.com/bioconda/bioconda-recipes/pull/6131

jamesabbott:
Thanks @daler...I think I see what is happening....I've been getting a bit confused since I assumed the abyss and abyss-k128 recipes were the same apart from the k-mer compilation option. In fact, abyss-2.0.1-k128 does not include 'make' in the run requirements, and has also dropped the 'abyss-pe' test. It therefore passes tests fine since it is not testing actually abyss-pe, but abyss-pe runs fine when run from a conda installation  (or via simulate-travis --disable-docker)  since I happen to have /bin/make locally. This behaviour  seems broken since it is reliant on make being present outside the conda environment - the older abyss versions take the same approach and therefore pass tests and seem ok, but can not be guaranteed to work.

The standard abyss recipe includes make as a runtime requirement, but since the shebang line is hard-coded to /bin/make it is not actually using the make installed via conda, so is also probably considered broken....again it will only work if /bin/make is available despite the runtime make requirement.

In my head the cleanest solution is to patch the shebang line abyss-pe to $PREFIX/bin/make, but suspect this would have to be done via sed in a post-install script, which seems evil and messy.
Any better suggestions?

mbargull:
No need for the post link script. Just use
```bash
sed -i "1 s|/usr/bin/make|$PREFIX/bin/make|"  bin/abyss-pe
```
at the top of `build.sh`. That way `/usr` is replaced with the prefix from the build environment which `conda-build` replaces with a placeholder which in turn is replaced by `conda` with the user's environment prefix at install time. So `conda-build`/`conda` already do the job of `post-install.sh` for us!

jamesabbott:
Neat...I tested with a post-link script last night and this does resolve the issues, so have moved it to build.sh...just need to let travis to it's thang...

bgruening:
@jamesabbott this is awesome! Thanks a bunch!

Thanks all for your reviews and help!

PR #6591, opened 2017-11-02
Removing quast from the build-fail blacklist

alexeigurevich:
Needed for updating quast to 4.6.0

Quast was included in the build-fail blacklist, so update to v4.6.0 (https://github.com/bioconda/bioconda-recipes/pull/6420) has no desired effect.

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -14,7 +14,6 @@ recipes/abyss/1.9.0
 recipes/abyss/2.0.1-k128
 
 recipes/clever-toolkit
-recipes/quast
 recipes/transcomb
 recipes/rsem
 recipes/metaprob


bgruening:
@alexeigurevich thanks for working on this.
It seems that is fails to compile gnuplot:
```
NOTICE: Failed to compile gnuplot (/opt/conda/conda-bld/quast_1509634438084/_t_env/lib/python2.7/site-packages/quast_libs/gnuplot)! Try to compile it manually. You can restart Quast with the --debug flag to see the compilation command.
```

Maybe you can depend on gnuplot during build time as well and convince it to not compile it?

alexeigurevich:
Gnuplot compilation is not required, so we can work if it fails (with a little bit limited functionality). Moreover, we did not changed this behaviour since v.4.5 which is successfully incorporated to Bioconda.

It looks like the problem is with GeneMark installation:
```
2017-11-02 16:21:25
Running GeneMarkS...
[Errno 2] No such file or directory: 'UNKNOWN/.gm_key'
....
```
This is caused by their licensing system which requires .gm_key file in user's home dir. 
I know how to fix this, since GeneMark is also an optional module: we just do not need to crash if we fail to install it properly. 
But the fix will require a new 4.6.1 release, so will take some amount of time (hopefully not much).

PR #6593, opened 2017-11-02
Update Nonpareil recipe

bebatut:


recipes/nonpareil/meta.yaml
```diff
@@ -1,24 +0,0 @@
-package:
-  name: nonpareil
-  version: "2.4.01"
-source:
-  git_url: https://github.com/lmrodriguezr/nonpareil.git
```
bgruening: This recipe had a higher version number than you included now. Is this intended? Any idea what we should do with the "older" version that has a higher version number?

bebatut: The old version is failing with `simulate_travis` because of `git_url`, it is why I changed to the  closest (older) release. There is also a release for 2.4.1 if you think it is better

bgruening: 2.4.1 would at least ensure that 2.4.01 is not the default one what you get with a normal install.

bebatut: I added the latest version (3.1.1). The normal install should be this one. No?

bgruening: Not sure what I have seen before. All good.

PR #6600, opened 2017-11-03
Add recipe for valet

bebatut:

### DIFF
--- a/recipes/valet/build.sh
+++ b/recipes/valet/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+mkdir -p $PREFIX/bin
+
+cp src/py/* $PREFIX/bin
+export VALET=$PREFIX/bin
--- a/recipes/valet/build.sh
+++ b/recipes/valet/build.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 mkdir -p $PREFIX/bin
+mkdir -p $PREFIX/bin/src/py/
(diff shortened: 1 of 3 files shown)


bebatut:
I found an issue when testing it locally. I am fixing it

recipes/valet/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+mkdir -p $PREFIX/bin
+
+cp src/py/* $PREFIX/bin
+export VALET=$PREFIX/bin
```
bgruening: Is this export needed at runtime?

bebatut: yes. Is there a different way to do that?

bgruening: This is currently only valid in the build environment.
If you would have a single binary, I would recommend a wrapper. E.g. a file called `_binary` that is a shell script that `export VLATE=`which binary` && binary`.
In your case it is probably better to patch the python scripts to automatically set VALET to `os.path.realpath(__file__)`

bebatut: :) 
I was testing the second option

bebatut:
I fixed most of the issues locally. But I still have 
```
dyld: Library not loaded: @rpath/libintl.9.dylib
  Referenced from: /Users/bebatut/miniconda3/envs/valet/lib/R/lib/libR.dylib
  Reason: image not found
```

bgruening:
Fixed and green! Thanks!

PR #6604, opened 2017-11-03
update vcftools only C++ libraries

Cashalow:

### DIFF
--- a/recipes/vcftools/meta.yaml
+++ b/recipes/vcftools/meta.yaml
@@ -19,8 +19,6 @@ requirements:
     - gcc  # [not osx]
     - llvm  # [osx]
     - zlib {{CONDA_ZLIB}}*
-    - perl
-    - perl-vcftools-vcf
     - autoconf
     - automake
     - libtool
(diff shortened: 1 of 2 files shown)


Cashalow:
Sorry there was a misunderstanding on my part, I will update first vcftools perl libraries

recipes/vcftools/meta.yaml
```diff
@@ -1,27 +1,34 @@
...
+    - autoconf
+    - automake
+    - libtool
+    - pkg-config
   run:
     - libgcc  # [not osx]
-    - zlib
-    - perl-vcftools-vcf ==0.953
+    - zlib {{CONDA_ZLIB}}*
+    - perl-vcftools-vcf
```
bgruening: can you add `perl` as a dependency as well

bgruening: I'm wondering why the perl tools are added at all? Should this not be a separate package, for people that don't need perl.

Cashalow: I don't know, as both C and perl libraries are included in the original package, people loading it would expect to have both available when using the bioconda package and discarding the perl libraries would cause confusion on the user's side no ? I didn't understand why they were split in the first place but I wanted to update the package following the previous work on bioconda ?

Cashalow:
I have removed the perl dependencies and stated in the summary that only the C++ libraries are present in this package.

PR #6605, opened 2017-11-03
Jalview v1.12.2b2

martin-mann:


recipes/jalview/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - openjdk >=7
+    - ant
+    - xorg-libxtst
+  run:
+    - openjdk >=7
+    - xorg-libxtst
+
+test:
+    commands:
+      - jalview -h
```
bgruening: Do you need to run the tests also headless?

martin-mann: ? Not sure i get what you mean?!

martin-mann: But i changed the call, seems the argument was wrong, lets see...

martin-mann:
Ping @eggzilla , close to working...

eggzilla:
@martin-mann , really nice seeing such a useful tool added in one day! Congrats!

recipes/jalview/build.sh
```diff
@@ -16,10 +16,10 @@ echo -e '#!/usr/bin/env bash\n\
...
 java -Djava.awt.headless=true -Djava.ext.dirs=\$JALVIEWDIR -cp \$JALVIEWDIR/jalview.jar jalview.bin.Jalview \${@};\n\
 ' | tr -d '\\' > $JALVIEWDIR/jalview.sh; 
-chmod 0755 $JALVIEWDIR/jalview.sh; 
+chmod +x $JALVIEWDIR/jalview.sh; 
 
 # link wrapper script
 mkdir -p $PREFIX/bin
-ln -s $JALVIEWDIR/jalview.sh $PREFIX/bin/jalview
-chmod 0755 $PREFIX/bin/jalview; 
+ln -s ../share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM/jalview.sh $PREFIX/bin/jalview
```
bgruening: should this be $JALVIEWDIR?

martin-mann: Nope, tried to use a relative rather than an absolute path link to the wrapper.. not good? Done by the system anyway..

bgruening: This links from the build environment to the $PREFIX. If you want to use relative path, you need to change to $PREFIX at first.

martin-mann: Mhh.. sounds reasonable.. *grin* will change it back to $PREFIX to avoid confusion...

Funny thing: seems to be somehow (partially?) Working, since the test finds the linked shell script.. strange..

martin-mann:
Mhhh.. when browsing the travis log it seems the hardcoded installpath is lost or wrong.. will change to something dynamic based on the script location as in the python wrapper from the example... but mostlikely on monday..

martin-mann:
@bgruening grmpf.. seems now there is something wrong with the travis biconda build.. please have a look at the travis log from the last commit (I restarted the travis build already without any resulting change). thanks!

martin-mann:
@bgruening seems readlink ist the problem since it doesnt understand '-f' argument.. :/

maybe I will swap to a perl wrapper in the end.. or do you have an example shell script that identifies its own directory for further use?

good night, Martin

bgruening:
This might be helpful https://github.com/bioconda/bioconda-recipes/blob/master/recipes/peptide-shaker/peptide-shaker.py or the shell version: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/snpeff/snpeff.sh

martin-mann:
@bgruening thank! i have seen the trick from the snpeff recipe before but thought the 'short thingy' i tried might be sufficient.. seems I was wrong. ;)

recipes/jalview/build.sh
```diff
@@ -0,0 +1,33 @@
...
+# create target directory
+JALVIEWDIR=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+mkdir -p $JALVIEWDIR
+
+# copy files to target
+cp -R dist/* $JALVIEWDIR/.
+
+# create wrapper script
+echo -e '#!/usr/bin/env bash\n\
+# Find original directory of bash script, resolving symlinks\
```
bgruening: Just for the sake of readability, wouldn't it be better it have this as separate file in the repo?

recipes/jalview/meta.yaml
```diff
@@ -0,0 +1,32 @@
+
+about:
```
bgruening: can you move this to the bottom?

recipes/jalview/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    home: http://www.jalview.org/
+    license: GNU GENERAL PUBLIC LICENSE, Version 3.0
+    summary: |
+        Jalview is a free program for multiple sequence alignment editing, visualisation and analysis. 
+        Use it to view and edit sequence alignments, analyse them with phylogenetic trees and principal 
+        components analysis (PCA) plots and explore molecular structures and annotation.
+package:
+    name: jalview
+    version: 2.10.2b2
```
bgruening: You could use template-vars for the number, would make updates easier!

recipes/jalview/build.sh
martin-mann: jup, I decided the same last night..

recipes/jalview/meta.yaml
martin-mann: problem is, the link uses underscores instead of points... any idea?

recipes/jalview/meta.yaml
martin-mann: done

bgruening:
Cool thanks!

PR #4798, opened 2017-05-24
Feature/perl file temp

jerowe:
* [ ] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [x] This PR adds a new recipe.
* [x] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

More low level perl recipes.

PR #6606, opened 2017-11-03
updating perl-vcftools-vcf

Cashalow:

I have put the version to 0.1.14 because that is what is stated on the Vcf.pm file (0.1.14-12 exactly, the version of the package used to be extracted from there) although the version of the full package is 0.1.15

### DIFF
--- a/recipes/perl-vcftools-vcf/build.sh
+++ b/recipes/perl-vcftools-vcf/build.sh
@@ -1,10 +1,9 @@
 #!/bin/bash
 set -eu
-
 mkdir -p perl-lib/lib
 cp src/perl/*.pm perl-lib/lib
 # Fix naming scheme of package to make Builder happy
 sed -i.bak 's/Vcf.pm.  Module/Vcf - Module/' perl-lib/lib/Vcf.pm
 cp $RECIPE_DIR/Build.PL perl-lib
 cd perl-lib
-cpanm -v -i .
+HOME=/tmp cpanm -v -i .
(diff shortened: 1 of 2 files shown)


Cashalow:
To fix the cpanm error on osX, I have used the fix presented here : https://github.com/bioconda/bioconda-recipes/pull/4798, please tell me if it's not okay !

recipes/perl-vcftools-vcf/build.sh
```diff
@@ -1,12 +1,9 @@
...
-
 mkdir -p perl-lib/lib
 cp src/perl/*.pm perl-lib/lib
-# Fix version to something cpan installable tools like
-sed -i.bak 's/r953/0.953/' perl-lib/lib/Vcf.pm
 # Fix naming scheme of package to make Builder happy
 sed -i.bak 's/Vcf.pm.  Module/Vcf - Module/' perl-lib/lib/Vcf.pm
 cp $RECIPE_DIR/Build.PL perl-lib
 cd perl-lib
-cpanm -v -i .
```
bgruening: Why can you not call Build.PL directly? https://github.com/bioconda/bioconda-recipes/pull/4798/files#diff-07b0d6ce7a6e79394d4384e7db728326R4

My fear is that cpanm does more magic things, like pulling down other libs that should be defined more explicitly.

Cashalow: Indeed it works perfectly well thanks ! I have removed the cpan dependency accordingly.

peterjc:
Looks good to me, and you've address the issues flagged by @bgruening - merging...

peterjc:
Thanks!

PR #6608, opened 2017-11-03
Update extract-sv-reads to v1.2.1

ernfrid:

This updates extract-sv-reads to v1.2.1. See the release notes for details.


recipes/extract-sv-reads/meta.yaml
```diff
@@ -4,22 +4,26 @@ build:
...
-    - zlib
+    - curl
+    - bzip2 {{ CONDA_BZIP2 }}*
+    - xz
   run:
     - libgcc
-    - zlib
+    - curl
+    - bzip2 {{ CONDA_BZIP2 }}*
+    - xz
```
bgruening: can we pin xz 5.2.* as well

ernfrid: Sure thing. What's required to do that? Add a line to env_matrix.yml like so?
```
CONDA_XZ: "5.2"
```

bgruening: Yes, this would be supercool!

ernfrid:
Updated as requested. Let me know if this addresses your comments @bgruening

bgruening:
Fantastic! :)

PR #6610, opened 2017-11-03
Remove MultiQC from the blacklist. It's already set to use the extend…

dpryan79:
…ed image.

This is mostly just removing MultiQC from the blacklist. It was already requesting the extended docker image for testing, though apparently since it was blacklisted that was never actually used.


recipes/multiqc/meta.yaml
```diff
@@ -116,8 +117,7 @@ test:
 
   commands:
     # click requires a unicode locale
-    - LC_ALL=C.UTF-8 LANG=C.UTF-8 multiqc --version # [linux]
-    - LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 multiqc --version # [osx]
+    - LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 multiqc --version
```
mbargull: You can even just use
```yaml
    - multiqc --version
```
since the `en_US.UTF-8` locale is already set in the extended base image.

recipes/multiqc/meta.yaml
```diff
@@ -14,6 +14,7 @@ build:
 requirements:
   build:
     - python
+    - setuptools
```
mbargull: Could you also add `setuptools` to the run requirements and add
```yaml
  skip-lints:
```
to the `extra` section?

dpryan79: Does it actually need `setuptools` to run? That'd be rather unusual.

recipes/multiqc/meta.yaml
dpryan79: True

recipes/multiqc/meta.yaml
mbargull: Yes, it uses the `pkg_resources` package which is part of `setuptools`, see: https://github.com/ewels/MultiQC/search?q=pkg_resources. But you are right, most packages do not require `setuptools` to run. The usage of `pkg_resources` is one of those special cases which make it a requirement, though.

dpryan79:
Once this passes on TravisCI I'll merge (or someone can do it for me).

bgruening:
Gracias!

PR #6620, opened 2017-11-04
Update pysam

kyleabeauchamp:

PS: does anyone here know why we switched from hard-pinning HTSLib in pysam to pinning it to a global Bioconda variable?  I feel like having multiple versions of HTSLib/Samtools in a single environment is asking for trouble, and that's what one could get with having a global variable rather than pinning this to HTSLib 1.6

### DIFF
--- a/recipes/pysam/meta.yaml
+++ b/recipes/pysam/meta.yaml
@@ -1,15 +1,15 @@
 package:
     name: pysam
-    version: 0.12.0.1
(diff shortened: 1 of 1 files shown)


kyleabeauchamp:
cc @AndreasHeger just merged bioconda recipe, build should be online in an hour or so.

AndreasHeger:
Great, thanks!

PR #6630, opened 2017-11-06
Beagle 4.0

dlaehnemann:

Sorry to be bothering @core again, this recipe adds in an older version of a tool, where a recipe already exists for newer versions. So while this should pass all tests and I could then merge it, I'm not sure if this setup with a subdirectory in the recipe directory is correct / wanted. It is roughly modelled after the version subdirectories I found for peptide-shaker:
https://github.com/bioconda/bioconda-recipes/tree/master/recipes/peptide-shaker

### DIFF
--- a/recipes/beagle/4.0_06Jun17/beagle
+++ b/recipes/beagle/4.0_06Jun17/beagle
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+# Wraps beagle.jar
+#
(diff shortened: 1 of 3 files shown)


daler:
@dlaehnemann this is exactly what the subdirectories are for (adding an older version when a newer one is already maintained) so this should be fine. Thanks!

dlaehnemann:
@daler will merge then, thanks for the info!

PR #6639, opened 2017-11-06
DISCASM: Added missing dependency (pysam) and added tests of util scripts.

tstrnity:

### DIFF
--- a/recipes/discasm/meta.yaml
+++ b/recipes/discasm/meta.yaml
@@ -33,10 +33,10 @@ test:
     # Trinity and sets TRINITY_HOME to be the directory where Trinity resides.
     # The following commands need to be invoked from where they are located.
     # DISCASM runs them from that location. These tests invoke them that way.
-    - "$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM/util/retrieve_SF_chimeric_and_unmapped_reads.py --help 2>&1 | grep \"usage\" "
-    - "$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM/util/retrieve_SF_chimeric_reads.py --help 2>&1 | grep \"usage\" "
-    - "$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM/util/run_OASES_multiK.pl --help 2>&1 | grep \"##################################\" "
-    - "$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM/util/run_OASES.pl --help 2>&1 | grep \"usage\" "
+    - "$PREFIX/share/discasm-0.1.2-1/util/retrieve_SF_chimeric_and_unmapped_reads.py --help 2>&1 | grep \"usage\" "
+    - "$PREFIX/share/discasm-0.1.2-1/util/retrieve_SF_chimeric_reads.py --help 2>&1 | grep \"usage\" "
+    - "$PREFIX/share/discasm-0.1.2-1/util/run_OASES_multiK.pl --help 2>&1 | grep \"##################################\" "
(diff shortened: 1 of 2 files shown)


mbargull:
Would it make sense to also just use wrapper scripts for the additional `util/*.{py,pl}` scripts? Just like the one you already used:
```bash
echo "#!/bin/bash" > $PREFIX/bin/DISCASM
echo "$DISCASM_INSTALL_PATH/DISCASM \$@" >> $PREFIX/bin/DISCASM
chmod +x $PREFIX/bin/DISCASM
```

tstrnity:
@mbargull Thanks for the comment. I thought about doing that. The wrapper scripts would only be used for the recipe tests, which is why I didn't. It would make the meta.yaml look cleaner, but would require each one of the scripts to have a corresponding wrapper built which makes build.sh longer.  I will be submitting another similar recipe for gmap-fusion that has about twice as many scripts to test. I will do whatever would be preferred: wrapper scripts vs. referencing through the $PREFIX in the meta.yaml.

mbargull:
If they are only used for the tests, then it's just fine like you did it as they don't need to have a wrapper in `$PREFIX/bin`. Then you could also consider to turn the build number into a jinja2 variable just like the `{% set version = "0.1.2" %}` so that you can use `$PREFIX/share/discasm-{{ version }}-{{ build_num }}/util/` later on -- that way you won't have to update those paths for subsequent builds.

The decision whether or not to make the scripts accessible via `$PREFIX/bin/...` is totally up to you. It shouldn't be in any way important if either `meta.yaml` or `build.sh` are aesthetically more pleasing. It just depends on if the purpose of the software is to make them directly callable by the user.
I don't know what those scripts do, hence essentially my question could be rephrased as "Should those scripts be directly accessible to the user?" :wink:

tstrnity:
DISCASM is the only program that is intended to be directly callable by the user. I will set build as a jinja2 variable, as you suggest, so the paths don't need to be altered in subsequent builds.

recipes/discasm/meta.yaml
```diff
@@ -1,38 +1,47 @@
...
   fn: DISCASM-v{{ version }}.tar.gz
   url: https://github.com/DISCASM/DISCASM/releases/download/DISCASM-v{{ version }}/DISCASM-v{{ version }}.tar.gz
   md5: 571b2265018608298752f40f3f375b0f
 
 build:
-    number: 0
-    skip: true  # [osx or py3k]
-    # The reason for skipping osx is that trinity >=2.4 is not available in the osx channel
-    # The python code in DISCASM runs on python 2.7.
+  number: 1
```
mbargull: Use `number: {{ build }}` to be consistent.

tstrnity: right. Will do.

recipes/discasm/meta.yaml
```diff
@@ -1,38 +1,47 @@
...
-    commands:
-        - DISCASM --help > /dev/null 2>&1
-        # DISCASM requires the environment variable TRINITY_HOME to be set.
-        # If TRINITY_HOME is not set before it starts, it uses "which" to find
-        # Trinity and sets TRINITY_HOME to be the directory where Trinity resides.
+  commands:
+    - DISCASM --help 2>&1
+    # DISCASM requires the environment variable TRINITY_HOME to be set.
+    # If TRINITY_HOME is not set before it starts, it uses "which" to find
+    # Trinity and sets TRINITY_HOME to be the directory where Trinity resides.
```
mbargull: That comment about `TRINITY_HOME` isn't really relevant for the recipe, but should rather be part of the software's documentation, e.g., readme file or wiki.

tstrnity: ok. will remove.

mbargull:
Great, LGTM! Let's wait for the CI to run the tests and then we're good to go. Thanks @tstrnity

PR #6649, opened 2017-11-07
initial nest conda build files

planetmija:

### DIFF
--- a/recipes/nest/build.sh
+++ b/recipes/nest/build.sh
@@ -1,5 +1,11 @@
 #!/bin/sh
 
+export MPI_FLAGS=--allow-run-as-root
+
+if [ $(uname) == Linux ]; then
+    export MPI_FLAGS="$MPI_FLAGS;-mca;plm;isolated"
+fi
+
 export CFLAGS="-I$PREFIX/include"
 export LDFLAGS="-L$PREFIX/lib"
 
--- a/recipes/nest/build.sh
+++ b/recipes/nest/build.sh
@@ -11,7 +11,7 @@ cmake -DCMAKE_INSTALL_PREFIX:PATH=$PREFIX \
       -Dwith-python=ON \
       ..
 
-make -j
+make -j2
 make install
+# this is needed to make the python bindings work
 cp $PREFIX/lib64/* $PREFIX/lib -r
-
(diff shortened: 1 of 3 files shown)


recipes/nest/build.sh
```diff
@@ -21,3 +21,8 @@ make -j2
 make install
 # this is needed to make the python bindings work
 cp $PREFIX/lib64/* $PREFIX/lib -r
+
+extra:
+  container:
+	    # openmpi needs ssh/rsh
+	    extended-base: true
```
mbargull: `build.sh` -> `meta.yaml` 😉

bgruening:
@bioconda/core this is our first recipe with openmpi. Openmpi needs openssh installed, which is really tricky.
This recipe builds with with following patches for our build-infrastructure:
* https://github.com/bioconda/bioconda-utils/pull/211
* https://github.com/bioconda/bioconda-extended-base-image/pull/6

This work was done and discussed during our workshop yesterday with admins from out local compute center - (HPC and Cloud).

johanneskoester:
Sounds like a reasonable strategy to me. I guess openssh as a conda package is not feasible?

daler:
The conda package for `nest` will still require a system installation of openssh though, right? I guess the linux-anvil image has it, which is why this builds successfully. 

If there's no good way to get a conda package, then +1 on adding to extended base.

bgruening:
@johanneskoester at least conda-forge is installing it via yum and following a similar strategy. I guess part of the reason are security updates etc.

I will take your comments as ok and proceed with merging the PRs?

bgruening:
@daler I would just follow conda-forge here for the time being and not mess with openssh.
And yes all systems will need to have openssh-client installed. The linux-anvil has __not__ included openssh. Thats the reasons I added them in these PRs.

* https://github.com/bioconda/bioconda-utils/pull/211
* https://github.com/bioconda/bioconda-extended-base-image/pull/6

daler:
Ah, I missed the bioconda-utils commit with the RUN line. And it makes sense to use the yum repos for security updates.

https://github.com/bioconda/bioconda-utils/pull/184 was an attempt to move toward a pre-built docker container which should be fine now that the build system is no longer changing frequently. If later we decide the build times get too long, moving to that pre-built model and adding openssh there would speed things up considerably.

bgruening:
@planetmija thanks!

PR #6681, opened 2017-11-09
ConsensusFixer

DrYak:
Add a recipe for ConsensusFixer (https://github.com/cbg-ethz/ConsensusFixer) based on the currently present recipe for InDelFixer.

### DIFF
--- a/recipes/consensusfixer/ConsensusFixer.sh
+++ b/recipes/consensusfixer/ConsensusFixer.sh
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+# ConsensusFixer wrapper script
+# based on picard wrapper
+set -eu -o pipefail
+
+set -o pipefail
+export LC_ALL=en_US.UTF-8
+
+# Find original directory of bash script, resovling symlinks
+# http://stackoverflow.com/questions/59895/can-a-bash-script-tell-what-directory-its-stored-in/246128#246128
+# this is basically a bash re-implementation of the (GNU-only, non Mac OS X) "realpath" command
+SOURCE="${BASH_SOURCE[0]}"
+while [[ -h "$SOURCE" ]]; do # resolve $SOURCE until the file is no longer a symlink
+    DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+    SOURCE="$(readlink "$SOURCE")"
+    [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
+done
+DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+
+JAR_DIR=$DIR
+ENV_PREFIX="$(dirname $(dirname $DIR))"
+# if JAVA_HOME is set (non-empty), use it. Otherwise 
+# Use Java installed with Anaconda to ensure correct version
+java="${JAVA_HOME:-$ENV_PREFIX}/bin/java"
+
+# extract memory and system property Java arguments from the list of provided arguments
+# http://java.dzone.com/articles/better-java-shell-script
+default_jvm_mem_opts=( "-Xms512m" "-Xmx1G" )
+jvm_mem_opts=()
+jvm_prop_opts=()
+pass_args=()
+for arg in "$@"; do
+    case $arg in
+        '-D'*)
+            jvm_prop_opts+=( "${arg}" )
+            ;;
+        '-XX'*)
+            jvm_prop_opts+=( "${arg}" )
+            ;;
+         '-Xm'*)
+            jvm_mem_opts+=( "${arg}" )
+            ;;
+         *)
+            pass_args+=( "${arg}" )
+            ;;
+    esac
+done
+
+if (( ${#pass_args[@]} )) && [[ "${pass_args[0]}" == ch.ethz* ]]; then
+    jar_arg='-cp'
+else
+    jar_arg='-jar'
+fi
+exec -a "${BASH_SOURCE[0]}" "$java" "${jvm_mem_opts[@]:-"${default_jvm_mem_opts[@]}"}" ${jvm_prop_opts[@]:+"${jvm_prop_opts[@]}"} "${jar_arg}" "$JAR_DIR/ConsensusFixer.jar" ${pass_args[@]:+"${pass_args[@]}"}
+
--- a/recipes/consensusfixer/build.sh
+++ b/recipes/consensusfixer/build.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
+
+consensusfixer="$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM"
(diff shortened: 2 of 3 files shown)


recipes/consensusfixer/ConsensusFixer.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
SoapZA: `#!/usr/bin/env bash`
Prefer portable shebangs: https://blogs.gentoo.org/mgorny/2016/02/08/a-quick-note-on-portable-shebangs/

recipes/consensusfixer/ConsensusFixer.sh
```diff
@@ -0,0 +1,65 @@
...
+# based on picard wrapper
+set -eu -o pipefail
+
+set -o pipefail
+export LC_ALL=en_US.UTF-8
+
+# Find original directory of bash script, resovling symlinks
+# http://stackoverflow.com/questions/59895/can-a-bash-script-tell-what-directory-its-stored-in/246128#246128
+SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
```
SoapZA: always use `[[ ]]` over POSIX `[ ]`, as you're already relying on bash

recipes/consensusfixer/ConsensusFixer.sh
```diff
@@ -0,0 +1,65 @@
...
+done
+DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
+
+JAR_DIR=$DIR
+ENV_PREFIX="$(dirname $(dirname $DIR))"
+# Use Java installed with Anaconda to ensure correct version
+java="$ENV_PREFIX/bin/java"
+
+# if JAVA_HOME is set (non-empty), use it. Otherwise keep "java"
+if [ -z "${JAVA_HOME:=}" ]; then
```
SoapZA: same

recipes/consensusfixer/ConsensusFixer.sh
```diff
@@ -0,0 +1,65 @@
...
+done
+
+if [ "$jvm_mem_opts" == "" ]; then
+    jvm_mem_opts="$default_jvm_mem_opts"
+fi
+
+pass_arr=($pass_args)
+if [[ ${pass_arr[0]:=} == org* ]]
+then
+    eval "$java" $jvm_mem_opts $jvm_prop_opts -cp "$JAR_DIR/ConsensusFixer.jar" $pass_args
```
SoapZA: `eval` is usually never necessary in such circumstances, have you tried it without it?

recipes/consensusfixer/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
SoapZA: ditto

recipes/consensusfixer/ConsensusFixer.sh
DrYak: Yup, good idea. I was just reusing @sposadac 's work on InDelFixer, but this makes it a bit more portable.
(We did use the same approach in SIB Vital-IT's RPMs).

recipes/consensusfixer/ConsensusFixer.sh
DrYak: `eval` will e.g. re-interpolate any pending `$` in the strings, etc.

I think, in the mind of the original wrapper's author, `eval` enables to  handle arguments containing white-spaces (that would otherwise be destroyed by the while loop), by wrapping additional quotes around them or giving extra back-slashes. (Similar to one approach of handling remote white-spaces in `rsync`).

(I would have personally went for much simpler handling through bash arrays and quoted array expansion.
But, again, I preferred to reuse whatever is already in bioconda)

DrYak:
@SoapZA : Maybe I shall re-use the [peptide shaker java wrapper](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/peptide-shaker/peptide-shaker.py) instead ?

DrYak:
@SoapZA : ...or if you want, we could even meet in person so we get to talk over a beer ;-)

recipes/consensusfixer/ConsensusFixer.sh
SoapZA: @DrYak the proper way to handle whitespace in bash is to use arrays. Using the peptide shaker is also an option, although I think this bash script can be made to work fine. Your choice really

DrYak: Choose the *redo it properly using arrays* route.

DrYak:
Wrapper script redone using modern Bash's arrays approach at @SoapZA 's request.

(After work beer invitation still applies, though ;-) )

recipes/consensusfixer/ConsensusFixer.sh
```diff
@@ -0,0 +1,55 @@
...
+set -eu -o pipefail
+
+set -o pipefail
+export LC_ALL=en_US.UTF-8
+
+# Find original directory of bash script, resovling symlinks
+# http://stackoverflow.com/questions/59895/can-a-bash-script-tell-what-directory-its-stored-in/246128#246128
+# this is basically a bash re-implementation of the (GNU-only, non Mac OS X) "realpath" command
+SOURCE="${BASH_SOURCE[0]}"
+while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
```
SoapZA: double brackets `[[ ]]`, please.

DrYak: I think my first attempt at answer got lost somewhere...

 - `test` and `[` bash built-ins clone the behaviour of the POSIX `test` command line tool.
 - `[[` built-in is a bash specific extension that adds C-style logical operation, pattern matching, regex matching, etc.

Here the script's original author uses standard behaviour (checking for symlinks) so the POSIX-compliant bash built-in suffices.

DrYak:
Commits squashed.

But somehow Travis doesn't trigger but stays expecting

DrYak:
NOTE: Successfully triggered TravisCI by force-pushing a rebased commit against upstream master.

recipes/consensusfixer/ConsensusFixer.sh
SoapZA: I am well aware of the difference between POSIX `[` and the bashism `[[`. There are good reasons to prefer `[[` over POSIX-isms, one of which is the fact that `[[` provides for less surprising behaviour than the POSIX stuff. https://stackoverflow.com/a/669486
Gentoo also recommends it over POSIX shell features, https://devmanual.gentoo.org/tools-reference/bash/index.html#single-versus-double-brackets-in-bash
Please just do this.

DrYak: And none of the weird surprising edge case are hit by the above code (which usually have to do with file pattern matching,  badly quoted empty string matching, etc.)
But, well, it's your call. Let's get the thing merged already... 
(squashed/rebased/pushed, waiting for Travis)

DrYak:
It's ready to merge !

PR #6690, opened 2017-11-10
spades 3.9.0 with python 3.6

pirovc:

I noticed that spades version 3.9.0 is not running on python 3.6 but conda does install it without dependency errors. I changed the recipe but I am not sure if that is the correct way of doing it.

```
$ spades.py --test; echo $?


== Error ==  python version 3.6 is not supported!
Supported versions are 2.4, 2.5, 2.6, 2.7, 3.2, 3.3, 3.4, 3.5


In case you have troubles running SPAdes, you can write to spades.support@bioinf.spbau.ru
Please provide us with params.txt and spades.log files from the output directory.
1
```


bgruening:
@pirovc we should remove the python3.6 builds then correct?

pirovc:
@bgruening correct

PR #6703, opened 2017-11-10
WIP: fix JAMM

dmaticzka:

Current installations of JAMM segfault. I've added the official test to check out what's going on and try to fix this. Sould be some dependency, the test is working with manual JAMM install and current R from RCRAN.

### DIFF
--- a/recipes/jamm/meta.yaml
+++ b/recipes/jamm/meta.yaml
@@ -18,6 +18,7 @@ build:
 requirements:
 
   run:
+    - util-linux
     - perl
     - r-signal
     - r-mclust >=4
--- a/recipes/jamm/meta.yaml
+++ b/recipes/jamm/meta.yaml
@@ -18,7 +18,6 @@ build:
 requirements:
 
   run:
-    - util-linux
     - perl
     - r-signal
     - r-mclust >=4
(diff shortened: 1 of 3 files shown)


recipes/jamm/meta.yaml
```diff
@@ -10,21 +10,28 @@ source:
...
     - r-signal
     - r-mclust >=4
+    - curl
 
 test:
   commands:
     - JAMM.sh | grep JAMM > /dev/null
     - SignalGenerator.sh | grep JAMM > /dev/null
+    - mkdir testdata
+    - curl bioinf.uni-freiburg.de/~maticzkd/jamm_testdata/ctcfRep1.chr21.bed > testdata/ctcfRep1.chr21.bed
```
bgruening: can you just create a 3 line BED file on the fly?

bgruening:
Seems you are passing the conda tests, but not the Docker tests. So no segault here. If you need tiny test files you could also put them into the package and fetch them during build.sh.
Thanks.

dmaticzka:
Thanks for having a look. Yes, the test case is not optimal yet. But it's temporary for now.

So far, R doesn't even start because of missing or incompatible libreadline. No chance for a segfault there :) Let's see if enforcing conda-forge r-base gets this further.

dmaticzka:
That's it. Should fix the mulled-test, osx and hopefully use the right dependencies after install.

PR #6710, opened 2017-11-10
Clean up perl packages

druvus:

### DIFF
--- a/recipes/perl-test-base/meta.yaml
+++ b/recipes/perl-test-base/meta.yaml
@@ -8,25 +8,23 @@ source:
   md5: ced0cd86b099f9fd2cd73df1f46ab5e9
 
 build:
-  number: 0
+  number: 1
 
 requirements:
   build:
-    - perl-threaded
+    - perl
     - perl-spiffy
     - perl-text-diff
     - perl-test-deep
 
   run:
-    - perl-threaded
+    - perl
     - perl-spiffy
     - perl-text-diff
 
 test:
   imports:
     - Test::Base::Filter
-  # Perl 'use' tests trigger "readline() on unopened filehandle DATA" errors,
-  # so we have to use this "dummy" test package instead.
   commands:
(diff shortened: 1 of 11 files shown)


recipes/perl-test-base/meta.yaml
```diff
@@ -8,25 +8,23 @@ source:
...
-    - perl-threaded
+    - perl
     - perl-spiffy
     - perl-text-diff
 
 test:
   imports:
     - Test::Base::Filter
-  # Perl 'use' tests trigger "readline() on unopened filehandle DATA" errors,
-  # so we have to use this "dummy" test package instead.
```
mbargull: I suppose this is a useful comment as it explains the reason for the "non-standard" test case. I'd vote to keep it as long as that workaround is needed.

recipes/perl-test-yaml/meta.yaml
```diff
@@ -8,20 +8,18 @@ source:
...
     - perl-test-base
 
   run:
-    - perl-threaded
+    - perl
     - perl-test-base
 
 test:
-  # Perl 'use' tests trigger "readline() on unopened filehandle DATA" errors,
-  # so we have to use this "dummy" test package instead.
```
mbargull: I suppose this is a useful comment as it explains the reason for the "non-standard" test case. I'd vote to keep it as long as that workaround is needed.

mbargull:
Build failed for `perl-test-base` and `perl-test-yaml` (skipped) due to the single quotes in the test commands. This is a known issue with the usage of `mulled`. Until we fixed that, we'll have to use double quotes:
`- perl -e "use Test::Base tests => 1; pass;"`
`- perl -e "use Test::YAML tests => 1; pass;"`

Not sure why `perl-tie-ixhash` failed as the error message isn't very informative..

recipes/perl-test-base/meta.yaml
druvus: good catch! I am adding it back.

PR #6726, opened 2017-11-11
Add bioconductor-fabia.

cbrueffer:

### DIFF
--- a/recipes/bioconductor-fabia/build.sh
+++ b/recipes/bioconductor-fabia/build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+mv DESCRIPTION DESCRIPTION.old
+grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
+$R CMD INSTALL --build .
--- a/recipes/bioconductor-fabia/build.sh
+++ b/recipes/bioconductor-fabia/build.sh
@@ -1,4 +1,7 @@
 #!/bin/bash
+
 mv DESCRIPTION DESCRIPTION.old
 grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
+
+autoconf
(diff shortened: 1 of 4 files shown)


johanneskoester:

mbargull:
Did some closer investigation so we can better understand what happens with R packages that are troublesome when compiled with Clang and output ambiguous messages like
```
configure: error: cannot run C compiled programs.
```
. Turns out, this package's `configure.ac` adds
```bash
LDFLAGS=`"${R_HOME}/bin/R" CMD config LDFLAGS`
```
in which `R CMD config LDFLAGS`'s output includes `-lgfortran`. And since it assigns to `LDFLAGS`, it overwrites any previous flags. That includes `-Wl,-rpath,$PREFIX/lib` and thus the compiled binaries are linked to `libgfortran` but can't find it:
```
dyld: Library not loaded: @rpath/./libgfortran.3.dylib
```
Since those compiler/linker flag assignments are [given in the R documentation](https://cran.r-project.org/doc/manuals/r-devel/R-exts.html#Configure-and-cleanup-1), this might be a common idiom.
There might be another/better way than the fix I added here. Maybe someone who's more familiar with R and https://svn.r-project.org/R/trunk/src/scripts/config would know.
In the last commit I added
```diff
-$R CMD INSTALL --build .
+$R CMD INSTALL --build . || (echo "error code: $?" ; cat config.log ; exit 1)
```
. Maybe, going forward, we should add something like this to R recipes to make debugging faster?

cbrueffer:
Great investigation @mbargull ! I agree, adding some more debug output would make this a lot easier. Please commit if you think this is ready now.

cbrueffer:
I've merged this since the recipe looks good now. Thanks a lot @mbargull !

PR #6758, opened 2017-11-13
Staged and conditional testing.

johanneskoester:

This PR modifies builds in two ways.
1. Linux and macOS builds happen in two stages. If the first stage (linux) fails, the second is not tried. While this could in theory make builds slower, the osx queue is in practice so slow that the difference should not be noticeable. Moreover the benefit is that we only fill the osx queue for builds that pass the linux tests. This should significantly improve the backlog situation within Bioconda.
2. The skipping of pushes in pull requests is now handled via conditions in the travis configuration. This means that, instead starting a job and exiting immediately, we do not even need to start the job. Again, this should help a lot with our queue. Further, it also makes the PR experience cleaner, because we have no pseudo-succeeded jobs.

ping @bioconda/core.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,3 @@
-os:
-  - osx
-  - linux
 language: generic
 sudo: required
 
@@ -18,7 +15,14 @@ before_install:
 
 install: scripts/travis-setup.sh
 
-script: scripts/travis-run.sh
+jobs:
+  include:
+    - stage: Linux
+      os: linux
+      script: scripts/travis-run.sh
+    - stage: macOS
+      os: osx
+      script: scripts/travis-run.sh
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,12 +17,19 @@ install: scripts/travis-setup.sh
 
 jobs:
   include:
+    # in forks, master and bulk, we run osx and linux in parallel on each push
+    - stage: Linux+macOS
+      if: (type = push) AND ((branch IN (master, bulk)) OR (fork = true))
+      os:
+        - linux
+        - osx
+    # in pull requests, we first run linux and only if that succeeds, we run osx
     - stage: Linux
-      if: (branch IN (master, bulk)) OR (type = pull_request) OR (fork = true)
+      if: type = pull_request
       os: linux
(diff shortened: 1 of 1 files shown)


daler:
@johanneskoester great ideas!

daler:
@johanneskoester now that I think about it, this will not work well for our current workflow on `bulk` where we expect both linux and osx build and fail due to timeouts. With this PR we would have to get all recipes successfully built on linux before starting any work on osx.

I guess we could get around this with blacklists, by temporarily blacklisting failing linux packages just to get the osx builds to run. That extra complexity might be worth the overall time and resource savings from this PR.

bgruening:
Thanks for working on this @johanneskoester!
A few (meta? ;)) Questions from my side.

> Linux and macOS builds happen in two stages. If the first stage (linux) fails, the second is not tried. While this could in theory make builds slower, the osx queue is in practice so slow that the difference should not be noticeable. Moreover the benefit is that we only fill the osx queue for builds that pass the linux tests. This should significantly improve the backlog situation within Bioconda.

Do we really have such a large backlog? I think it is actually ok'ish. We have a high turnaround, but waiting 24h for a package is ok for me. Maybe we should communicate this better. 
For many people osx is the only thing we can not test locally. So we would end up waiting for linux to finish until we know if it works for OSX, and debugging OSX is then really painful. Today we can kill travis linux jobs manually and just wait for the OSX one, this would not be possible anymore.

> The skipping of pushes in pull requests is now handled via conditions in the travis configuration. This means that, instead starting a job and exiting immediately, we do not even need to start the job. Again, this should help a lot with our queue. Further, it also makes the PR experience cleaner, because we have no pseudo-succeeded jobs.

Can we simply encourage people to use a fork and not the main repo?

P.S. Also recent slow down is also due to the bulk branch. Which is not always the case and we could do this in a different organisation if really needed.

johanneskoester:
@bioconda/core any objections against the final version of this PR? Am I missing a use case?

daler:
Looks good to me, I like keeping the option of OSX-and-Linux in parallel on forks. And it looks like there's enough flexibility in the config to change things if/when we find issues or more use-cases.

PR #6781, opened 2017-11-14
package pfam_scan.pl (searching Pfam with HMMER 3)

peterjc:

### DIFF
--- a/recipes/pfam_scan/VERSIONS.txt
+++ b/recipes/pfam_scan/VERSIONS.txt
@@ -1,0 +1,91 @@
+All the pfamscan files listed below were downloaded from the EBI FTP site,
+which has a confusing folder structure with no clear and consistent version
+information in the filenames. Which is why I made these notes.
+
+Peter Cock, 2017-11-14
+
+===========
+Version 0.7
+===========
+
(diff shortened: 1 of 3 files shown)


peterjc:
This is also similar to the ``build.sh`` used for the ``mugsy`` package by @acaprez

simonvh:
Hi @peterjc, I'm not a Perl expert, but to me the package looks fine. Any specific issues with the build?

peterjc:
(Turns out it was breaking on the multi-line bash syntax to create the wrapper script, copy-paste had gone funny)

peterjc:
Green lights :)

The only thing I am unsure of is how I have used ``$PREFIX/bin`` for the wrapper (which sets ``$PERL5LBI``), and  ``$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM`` for the actual tool and its Perl bits.

Would it be enough to just stick the unzipped files directly under ``$PREFIX`` instead?

simonvh:
Don't know how other packages approach this, but I like it this way. More tidy IMO.

peterjc:
Yep, I didn't say here in the PR comments, but in the commit notes stated this was based on (your) ``trawler`` recipe - which is why I put you down as a reviewer.

Merging now, thank you!

PR #6783, opened 2017-11-14
Update discosnp++ to 2.3.0

cmonjeau:

### DIFF
--- a/recipes/discosnp/2.2.10/build.sh
+++ b/recipes/discosnp/2.2.10/build.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+mkdir -p ${PREFIX}/bin
+
+# decrease RAM needed
+sed -i.bak 's/make -j8/make -j1/' INSTALL
+
+# remove tests because of too much RAM needed
+sed -i.bak 's/. .\/simple_test\.sh/#. .\/simple_test\.sh/' INSTALL
+
+# installation
+sh INSTALL
+
+# change run_discoSnp++.sh deps path
+sed -i.bak 's/\$EDIR\/bin/\$EDIR/' run_discoSnp++.sh
+sed -i.bak 's/scripts\//..\/scripts\//' run_discoSnp++.sh
+
+# copy binaries
+cp run_discoSnp++.sh ${PREFIX}/bin
+chmod +x ${PREFIX}/bin/run_discoSnp++.sh
+
+cp build/ext/gatb-core/bin/dbgh5 ${PREFIX}/bin
+cp build/bin/read_file_names ${PREFIX}/bin
+cp build/bin/kissnp2 ${PREFIX}/bin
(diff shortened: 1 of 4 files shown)


cbrueffer:
Thanks!

cmonjeau:
Hi,

the 2.3.0 release is not in https://bioconda.github.io after the merge. Is it normal?

cbrueffer:
I don't think so, I see the same issue with more recipes. Any idea @daler or @bgruening ?

daler:
While the website only gets updated nightly, packages should be available immediately after building on travis-ci. The fact that they're not is definitely not normal, but it looks like it's due to a change in our build config. There's a fix in #6802. Once that fix is merged, the build system should detect that discosnp is missing and will build it (along with any other packages that haven't built on master in the past 19 hrs).

cbrueffer:
:+1:

mbargull:
`scripts/travis-run.sh` will only build packages if their recipes changed acc. to Git.
See: https://github.com/bioconda/bioconda-recipes/blob/c28499039a8dc6d1480863f741a3bc1f86c4b28b/scripts/travis-run.sh#L27-L70
So we'll explicitly need to trigger their builds again.

daler:
I merged master into bulk branch and it's picking them up: https://travis-ci.org/bioconda/bioconda-recipes/builds/302580513. However, it doesn't look like an OSX build is being triggered. I suspect the `os:` list is only seeing linux, but from the travis-ci docs at https://docs.travis-ci.com/user/conditional-builds-stages-jobs/ it's not clear how to fix this. @johanneskoester any ideas?

mbargull:
Ah, I just opened #6808 to trigger those builds again. Should I just close this in favor of your `bulk` changes?

daler:
I think you can close it, that way we don't have that trailing whitespace in the recipes.

daler:
From https://docs.travis-ci.com/user/build-stages/#Build-Stages-and-Build-Matrix-Expansion:

> Each job included in jobs.include inherits the first value of the array that defines a matrix dimension. In the example above, without explicitly setting rvm: 2.4, the included job inherits rvm: 2.3.

and from https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#Conditional-Jobs:

> Jobs need to be listed explicitely, i.e., using jobs.include (or its alias matrix.include), in order to specify conditions for them. Jobs created via matrix expansion currently cannot have conditions.

I can't quite work out what the syntax should be in order to get this to work. @johanneskoester would it be OK to revert #6758 until we work this out?

mbargull:
This affects `master` too, of course: https://travis-ci.org/bioconda/bioconda-recipes/builds/302575014

daler:
@mbargull I want to get #6810 in so we don't accumulate more unbuilt packages. I'll use the bulk branch to do the rebuilds.

PR #6786, opened 2017-11-14
Updated breseq to version 0.31.1.

ysu20:
The new version of breseq contains a fix to support bowtie2's new version number format.

### DIFF
--- a/recipes/breseq/meta.yaml
+++ b/recipes/breseq/meta.yaml
@@ -6,7 +6,7 @@ package:
   version: {{ version }}
 
 build:
-  number: 0
+  number: 1
 
 source:
   fn: {{ name|lower }}_{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


mbargull:
Hi @ysu20, thanks for doing this!
And sorry I hadn't already done this! I checked if requirements needed to be updated -- they don't -- but then forgot to update the recipe itself 😶.

We could refine the `bowtie2` run requirement to
```yaml
# constraints from https://github.com/barricklab/breseq/blob/v0.31.1/src/c/breseq/settings.cpp#L1117-L1162
- bowtie2 >=2.0.0,!=2.0.3,!=2.0.4,!=2.3.1
```
Though that's currently not necessarily needed since none of those versions are in the Bioconda channel.

With or without the additional version constraints on `bowtie2` this LGTM, thanks!

PR #6791, opened 2017-11-15
Pin Perl verison in Meraculous to the current defualt in conda-forge

ysu20:

Meraculous use Perl functions that are depreciated in the newer version of Perl, which is the case in the default channel.  This may is probably not necessary to pin the Perl version right now if we can assume that users will always set conda's the channel preference based on bioconda's recommendation.  However it has not been the case in my experience.  Additionally, this change will mitigate the risk of conda-forge upgrading its default Perl to a version that is not compatible with Meraculous in the future.

### DIFF
--- a/recipes/meraculous/meta.yaml
+++ b/recipes/meraculous/meta.yaml
@@ -9,7 +9,7 @@ source:
 
 build:
     skip: True  # [osx]
-    string: pl_5_22_{{PKG_BUILDNUM}}
+    string: pl5.22_{{PKG_BUILDNUM}}
     number: 1
 
 requirements:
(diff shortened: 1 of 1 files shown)


recipes/meraculous/meta.yaml
```diff
@@ -9,22 +9,23 @@ source:
 
 build:
     skip: True  # [osx]
-    number: 0
+    string: pl_5_22_{{PKG_BUILDNUM}}
```
bgruening: the build string looks like: `pl.5.22` see here for example: https://anaconda.org/bioconda/perl-bioperl/files

bgruening:
Thanks!

PR #6795, opened 2017-11-15
Short read connector tool

cmonjeau:

### DIFF
--- a/recipes/short-read-connector/build.sh
+++ b/recipes/short-read-connector/build.sh
@@ -4,6 +4,7 @@ mkdir -p ${PREFIX}/bin
 
 # decrease RAM needed
 sed -i.bak 's/make -j/make -j1/' INSTALL
+sed -i.bak 's/2> log_linker_err//' test/simple_test.sh
 
 # installation
 sh INSTALL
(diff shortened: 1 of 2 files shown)


recipes/short-read-connector/build.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+mkdir -p ${PREFIX}/bin
+
+# decrease RAM needed
+sed -i.bak 's/make -j/make -j1/' INSTALL
+sed -i.bak 's/2> log_linker_err//' test/simple_test.sh
+
+# change run_discoSnp++.sh deps path
+sed -i.bak 's/\$EDIR\/bin/\$EDIR/' short_read_connector.sh
```
bgruening: @cmonjeau you can also write this in the following way: `'s|\$EDIR/bin|\$EDIR|'`

recipes/short-read-connector/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+build:
+  number: 0
+
+requirements:
+  build:
+    - gcc   # [linux]
+    - llvm  # [osx]
+    - cmake
+    - zlib {{CONDA_ZLIB}}*
+    - git
```
bgruening: why is git needed?

cmonjeau: Hi @bgruening. Thanks for the review.

Inside the INSTALL file, there are git commands to get external repositories but it seems the dependency is not needed :)

recipes/short-read-connector/build.sh
cmonjeau: Done

PR #6804, opened 2017-11-15
abyss: add abyss-pe wrappers, fixes installs on long prefixes

mbargull:

@jamesabbott: I didn't consider overly long shebang lines when I recommended to replace `/usr` with `$PREFIX` in `abyss-pe`, my bad.
This creates a wrapper for `abyss-pe` and moves the original `abyss-pe` to `abyss-pe.Makefile`. Although it would be nicer to move it to `$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM/abyss-pe.Makefile` or the like, I didn't do that since the makefile has a line
```
path?=$(shell dirname `which $(MAKEFILE_LIST)`)
```
which depends on the directory the makefile is in. If we want to move it out of `$PREFIX/bin`, we'd have to patch that too.

@pvanheus: sorry for the troubles!

### DIFF
--- a/recipes/abyss/1.5.2/build.sh
+++ b/recipes/abyss/1.5.2/build.sh
@@ -13,12 +13,5 @@ fi
 make AM_CXXFLAGS=-Wall
 make install
 
-WRAPPER_PATH="$PREFIX/bin/abyss-pe"
-WRAPPED_PATH="$PREFIX/bin/abyss-pe.Makefile"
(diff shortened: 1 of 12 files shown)


pvanheus:
`abyss/bin/abyss-bloom-dist.mk` also uses make - does this also need a wrapper?

mbargull:
I guess so, as it uses the same `make -rRf`. Good catch! I'll include this as well.

PR #6805, opened 2017-11-15
Update subread to 1.6.0

lparsons:


daler:
@lparsons I don't think we need to keep the subdirectory here, just updating the existing recipe is fine. The older version of the package remains in the channel.

daler:
And thanks for updating this so soon after the release!

lparsons:
@daler Removed the subdir. I ran into a segfault on the old version of `featureCounts`, which appears to be resolved in 1.6.0, thus the quick upgrade. :wink:

Are we good to merge this?

daler:
Looks good, merge when green!

PR #6814, opened 2017-11-15
shorah version bump to 1.1.1

DrYak:
New version of Shorah added to the existing ones.
(Also includes a patch to make the C/C++ command line utilities return nicely with `-h/`--help` options)

### DIFF
--- a/recipes/shorah/1.1.1/shorah_nice_exit.patch
+++ b/recipes/shorah/1.1.1/shorah_nice_exit.patch
@@ -1,0 +1,99 @@
+diff --git a/src/shorah/b2w.cpp b/src/shorah/b2w.cpp
+index be21bab..21d1b92 100644
+--- src/shorah/b2w.cpp
++++ src/shorah/b2w.cpp
+@@ -178,7 +178,7 @@ int main(int argc, char* argv[])
+                 break;
+             case 'h':
+                 fprintf(stderr, "%s", help_string);
+-                return 1;
++                return 0;
+         }
+     }
+     if (argc < 11 || argc > 12) {
+diff --git a/src/shorah/contain.cpp b/src/shorah/contain.cpp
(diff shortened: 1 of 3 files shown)


armintoepfer:
Rather ping @ozagordi, who is the owner

recipes/shorah/1.1.1/shorah_nice_exit.patch
```diff
@@ -0,0 +1,99 @@
...
++    exit(status);
+ }
+ 
+ int main(int argc, char** argv)
+@@ -439,7 +439,7 @@ int main(int argc, char** argv)
+                 help();
+                 break;
+             case 'h':
+-                help();
++                help(0);
```
SoapZA: @DrYak please change all call sites to use `EXIT_SUCCESS` and `EXIT_FAILURE` macros. Then, don't include the patch, and instead send in a PR to our shorah upstream. We should avoid carrying patches upstream if at all possible (which it is, seeing that we are upstream).

Once the patch is integrated, I will cut a 1.1.2 release instead.

DrYak: of course the end plan is to integrate it upstream.  (pull request will be underway later this, once I find some free time during the Bern NGS Meeting).


But @sposadac needs it by tomorrow, so this 1.1.1 pull request is the temporary emergency solution, until 1.1.2.

armintoepfer:
Duct tape programming 101

DrYak:
Duck Tape

Yup...

I'll be starting on 1.1.2 shortly...

DrYak:
Trying to [remove my layer of duck tape](https://github.com/cbg-ethz/shorah/pull/37) without loosing too many feathers...

DrYak:
...and [as promised](https://github.com/bioconda/bioconda-recipes/pull/6824).

PR #6820, opened 2017-11-16
add structureharvester

lecorguille:

### DIFF
--- a/recipes/structureharvester/build.sh
+++ b/recipes/structureharvester/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+mkdir -p $PREFIX/bin
+cp $SRC_DIR/*.py $PREFIX/bin && chmod +x $PREFIX/bin/structureHarvester.py
+
--- a/recipes/structureharvester/meta.yaml
+++ b/recipes/structureharvester/meta.yaml
@@ -1,14 +1,13 @@
-{% set name = "structureHarvester" %}
 {% set version = "0.6.94" %}
 {% set commit = "e15952601b946267ce566df6ff0dd56df5c8b8f6" %}
 
 package:
-  name: {{ name|lower }}
+  name: structureHarvester 
   version: {{ version }}
 
 source:
   fn: {{ commit }}.zip
-  md5: 41ce1a230ba9a4ee9b7dca2dd63cc700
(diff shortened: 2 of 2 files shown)


recipes/structureharvester/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+mkdir -p $PREFIX/bin
+cp $SRC_DIR/*.py $PREFIX/bin && chmod +x $PREFIX/bin/structureHarvester.py
```
johanneskoester: Does this mean that other (non-executable) python files will pollute $PREFIX/bin? If so, please put them into $PREFIX/lib/structureharvester instead, and have a bash wrapper or symlink that points to the main python script.

recipes/structureharvester/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+{% set version = "0.6.94" %}
+{% set commit = "e15952601b946267ce566df6ff0dd56df5c8b8f6" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  fn: {{ commit }}.zip
+  md5: 41ce1a230ba9a4ee9b7dca2dd63cc700
```
johanneskoester: Please use sha256.

recipes/structureharvester/meta.yaml
```diff
@@ -0,0 +1,30 @@
+{% set name = "structureHarvester" %}
+{% set version = "0.6.94" %}
+{% set commit = "e15952601b946267ce566df6ff0dd56df5c8b8f6" %}
+
+package:
+  name: {{ name|lower }}
```
johanneskoester: Please put the name directly in here. Our goal is to only use jinja for stuff that occurs more than once and changes upon new releases.

lecorguille: Done

recipes/structureharvester/meta.yaml
lecorguille: Done

recipes/structureharvester/build.sh
lecorguille: There is only on other file.py which indeed isn't executable.
But sorry, I'm not sure to understand this alternative:
`symlink that points to the main python script.`
Do you have an example in mind?

PR #6823, opened 2017-11-16
add pureclip 1.2.1

dmaticzka:

Requires gcc >=4.9, so this will probably just sit here a while.

### DIFF
--- a/recipes/pureclip/build.sh
+++ b/recipes/pureclip/build.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+
+mkdir build
+cd build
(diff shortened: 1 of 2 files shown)


bgruening:
For openmpi support, please follow this recipe: https://github.com/bioconda/bioconda-recipes/blob/eac862cc446a5e961d09f13cf649cab8a91385ec/recipes/nest/meta.yaml

especially the extra section.

dmaticzka:
Thanx, this was in error and needs openmp (I think :). Switched to using the official static versions now.

PR #6827, opened 2017-11-17
Update fgbio 0.2.1->0.4.0

gshiba:

Ping @nh13 @tfenne @chapmanb

### DIFF
--- a/recipes/fgbio/build.sh
+++ b/recipes/fgbio/build.sh
@@ -7,13 +7,8 @@ mkdir -p $PREFIX/bin
 
 cd $SRC_DIR
 
-# Development version build
-sbt update
-sbt assembly
-cp target/scala-*/fgbio-*.jar $outdir/fgbio.jar
-
 # Release jar
-#cp fgbio*.jar $outdir/fgbio.jar
+cp fgbio*.jar $outdir/fgbio.jar
 
 cp $RECIPE_DIR/fgbio.py $outdir/fgbio
 chmod +x $outdir/fgbio
--- a/recipes/fgbio/meta.yaml
+++ b/recipes/fgbio/meta.yaml
@@ -1,5 +1,5 @@
-{% set version="0.2.1b" %}
-{% set tag="2af6747" %}
+{% set version="0.4.0" %}
+{% set tag="a9445b4" %}
 about:
     home: https://github.com/fulcrumgenomics/fgbio
     license: MIT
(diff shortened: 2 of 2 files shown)


recipes/fgbio/meta.yaml
```diff
@@ -1,5 +1,5 @@
-{% set version="0.2.1b" %}
-{% set tag="2af6747" %}
+{% set version="0.4.0" %}
+{% set tag="a9445b4" %}
```
nh13: Why not use the tag rather than the commit hash?

recipes/fgbio/meta.yaml
```diff
@@ -15,7 +15,7 @@ source:
   url: https://github.com/fulcrumgenomics/fgbio/archive/{{ tag }}.tar.gz
   #fn: fgbio-{{ version }}.jar
   #url: https://github.com/fulcrumgenomics/fgbio/releases/download/{{ version }}/fgbio-{{ version }}.jar
-  md5: c770b50025284f4baf9dceb45a7fd826
+  sha256: e3f1b73fc623741430dfc0058f7835a8cdc6d7fd6272d85a84fd7215d378d0d6
```
nh13: Since this is a release, can you use fn and url on lines 16-17?

recipes/fgbio/meta.yaml
nh13: May be a good idea to check how the meta.yaml was setup when it referenced a prior release.

gshiba: To be honest, this is the first time I've poked at [conda build recipes](https://conda.io/docs/user-guide/tasks/build-packages/recipe.html); I wanted the 0.4.0 release via conda as it really simplifies reproducibility for me, so I made the minimal changes necessary. I've made another commit, so take a look there please. Thanks for the prompt release, by the way.

recipes/fgbio/meta.yaml
gshiba: Done; please see next commit.

recipes/fgbio/build.sh
```diff
@@ -7,13 +7,8 @@ mkdir -p $PREFIX/bin
 
 cd $SRC_DIR
 
-# Development version build
-sbt update
```
bgruening: Is sbt as build dependency still needed?

bgruening: and scala?

gshiba: The tests passed, which implies they're not needed (hence they were removed)...?

tfenne: @gshiba I think @bgruening is referring to the requirements that are left in here:

```
   build:
      - openjdk >=8
     - sbt
     - scala
```

Since it's just pulling the release, you can remove the sbt and scala requirements from the build section.

gshiba: Ah, of course. Another push coming up shortly... thanks for your patience.

PR #6838, opened 2017-11-17
Bumped up the default memory for fgbio's JVM to 4g.

tfenne:

I'm bumping up the memory allocated in the default invocation of fgbio via the wrapper script to 4GB from 1GB.  Most of the programs are fairly memory intensive and will struggle in 1GB.  4GB should be enough for most to succeed.


bgruening:
Thanks!

PR #6843, opened 2017-11-18
Weka

visze:

Hi bioconda team. I added the famous machine learning library weka to bioconda. I spit weka up into weka-stable and weka-dev because the weka-dev add a lot of ne features (this is similar to the uploaded maven repository).

### DIFF
--- a/recipes/weka-dev/build.sh
+++ b/recipes/weka-dev/build.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+# Setup path variables
+BINARY_HOME=$PREFIX/bin
+PACKAGE_HOME=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+
+# Create destination directories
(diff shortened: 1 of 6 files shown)


bgruening:
Hi @visze! Welcome to Bioconda.
We don't have currently a rule for development packages, but I would not name them `stable` and `dev`. Weka should be `weka` if possible. This makes it easier to user to find it. You could think about a version prefix that indicates the dev version for example.

kyleabeauchamp:
Presumably this should go upstream on conda-forge as it's a general purpose ML library?

visze:
Yes conda-forge might be better. I just wanted to add a software on top of
it later for variant prioritization. So that's why I went to bioconda.

But maybe (or for sure) weka is not only used in the bioinformatics field
(even if I use it exclusive there).



Am 18.11.2017 21:33 schrieb "Kyle Beauchamp" <notifications@github.com>:

johanneskoester:

PR #6851, opened 2017-11-20
package PARalyzer on macOS

croth1:

@dyusuf do you see a problem with packaging PARalyzer on macOS as well? since it's java only, it should be fine - what do you think?


dyusuf:
to build a MAC pkg is always more tricky. 

I can not recall the exact issue of the OS failure while sometimes it relates to the configuration and OS infrastructure of bioconda itself.

however, the current failure of your build is not due to OS but because the recipe failed at linting which is a new controlling method introduced some time ago. for reference, https://bioconda.github.io/linting.html

croth1:
Let's see if I can make the linter happy :)

croth1:
Linter seems to be happy now! What do you think - can we merge this?

bgruening:
Thanks @croth1!

dyusuf:
@croth1 beautiful :-)

PR #6856, opened 2017-11-20
add smina

dkoes:

### DIFF
--- a/recipes/smina/build.sh
+++ b/recipes/smina/build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+mkdir -p $PREFIX/bin
+cp -R smina $PREFIX/bin
+chmod 0755 ${PREFIX}/bin/*
--- a/recipes/smina/meta.yaml
+++ b/recipes/smina/meta.yaml
@@ -4,15 +4,17 @@ about:
     summary: A fork of AutoDock Vina that is customized to better support scoring function development and high-performance energy minimization. 
 package:
     name: smina
-    version: "latest"
+    version: "Nov  9 2017"
 build:
   number: 0
   skip: False
 source:
     fn: smina                                                                   # [linux]
     url: https://downloads.sourceforge.net/project/smina/smina.static           # [linux]
+    md5: 2c36acc7c512a800640012355ea791bc                                       # [linux]
     fn: smina                                                                   # [osx]
     url: https://downloads.sourceforge.net/project/smina/smina.osx              # [osx]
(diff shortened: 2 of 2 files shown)


bgruening:
@dkoes is there any way to compile this from source?

dkoes:
Well, yes, but the static binary will be much more robust considering the number of dependencies.  The autodock-vina recipe is also a binary distribution.

mwojcikowski:
Does it play nice with new OpenBabel (2.4+)? Last time I tried compiling there was a problem with adding hydrogens.

bgruening:
@mwojcikowski are you ok with merging it. 
I'm, but it would be nice to have this build from source at some point.

dkoes:
One of the reasons I prefer to distribute a static executable is I can control what version of OpenBabel is used.  I'm not aware of any problems with hydrogens in this current distribution.

mwojcikowski:
I'm fine with merging it as binary. I'm using smina as a static binary myself.

But ultimately I'd prefer smina with recent OpenBabel, 2.3.2 is over 5 years and has many known and resolved bugs.

mwojcikowski:
@dkoes I see it was in fact fixed a while ago, sorry for misinforming. But the fact that OB 2.3.2 is ancient is valid.

The best way to figure this recipe out would be to make a static build using specific OB version, this way we could require OB during build and not on runtime. I'm afraid that OB package does not contain static libraries you can link against.

bgruening:
@dkoes @mwojcikowski please keep the discussion going. Its great to see the cheminformatic community here more and more!

I will merge this as a first step!
Thanks.

dkoes:
It isn't using 2.3.2, it's using a development version (specifically, github.com:dkoes/openbabel.git).  It has been a while since I've pulled from upstream (about a year) but it isn't ancient.

mwojcikowski:
Ok, fine. But unless it's specified somewhere we do not know that. You can pin to a git commit during a build from source and not use a openbabel package.

I just noticed that te versioning scheme is really bad. `01.30.2018` will be older than `11.09.2017`. Let's reverse the date to `Y-M-D`, as in RDKit if you prefer dates to release versions.

dkoes:
I submitted a pull request (#6925) to fix the versioning.

I don't think there is any particular dependency on the openbabel version in smina - it should build fine with any version.  But the static binary is built with whatever version is installed on my workstation, which happens to be my fork (which I will updated from upstream in the next build).

PR #6867, opened 2017-11-21
Abyss osx build

mbargull:
* x] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).

Trying out osx builds. WIP

### DIFF
--- a/recipes/abyss/build.sh
+++ b/recipes/abyss/build.sh
@@ -1,9 +1,5 @@
 #!/bin/bash
 
-#strictly use anaconda build environment
-CC=${PREFIX}/bin/gcc
-CXX=${PREFIX}/bin/g++
-
 ./configure \
     --prefix="$PREFIX" \
     --with-boost="$PREFIX" \
--- a/recipes/abyss/build.sh
+++ b/recipes/abyss/build.sh
@@ -13,8 +13,7 @@ fi
     --prefix="$PREFIX" \
     --with-boost="$PREFIX" \
     --with-mpi="$PREFIX" \
-    --with-sparsehash="$PREFIX" \
-    --enable-maxk=96
+    --with-sparsehash="$PREFIX"
 make AM_CXXFLAGS=-Wall
 make install
 
--- a/recipes/abyss/build.sh
+++ b/recipes/abyss/build.sh
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+if [[ "$(uname)" == Darwin ]]; then
+    # Fix for install_name_tool error:
+    #   error: install_name_tool: changing install names or rpaths can't be
+    #   redone for: $PREFIX/bin/abyss-overlap (for architecture x86_64) because
+    #   larger updated load commands do not fit (the program must be relinked,
+    #   and you may need to use -headerpad or -headerpad_max_install_names)
+    export LDFLAGS="$LDFLAGS -headerpad_max_install_names"
+fi
+
 ./configure \
     --prefix="$PREFIX" \
     --with-boost="$PREFIX" \
--- a/recipes/abyss/build.sh
+++ b/recipes/abyss/build.sh
@@ -1,15 +1,6 @@
 #!/bin/bash
 
-if [[ $(uname) == "Darwin" ]]; then
-	echo "Configuring for OSX..."
-	export CPPFLAGS="${CPPFLAGS} -I$PREFIX/include -I$PREFIX/include/boost --stdlib=libstdc++"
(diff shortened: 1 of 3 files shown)


recipes/abyss/build.sh
```diff
@@ -1,15 +1,15 @@
...
-	echo "Configuring for OSX..."
-	export CPPFLAGS="${CPPFLAGS} -I$PREFIX/include -I$PREFIX/include/boost --stdlib=libstdc++"
-	export CXXFLAGS="${CXXFLAGS} --stdlib=libstdc++"
-else
-	echo "Configuring for Linux..."
-	export CPPFLAGS="-I$PREFIX/include"
-fi
+#strictly use anaconda build environment
+CC=${PREFIX}/bin/gcc
+CXX=${PREFIX}/bin/g++
```
mbargull: Copied from:
https://github.com/bioconda/bioconda-recipes/blob/382f5765ac3d78903f2f9030edc406d5d964178e/recipes/google-sparsehash/build.sh#L3-L5

mbargull:
@bioconda/osx 
`google-sparsehash` is built/configured with `gcc` under `osx`. Because of that `abyss` would need to use `gcc` too to build for `osx`.
I don't believe that would be the correct fix, though. Rather than that, `google-sparsehash` should be built with `llvm` instead. Now `google-sparseshash/build.sh` has:
https://github.com/bioconda/bioconda-recipes/blob/382f5765ac3d78903f2f9030edc406d5d964178e/recipes/google-sparsehash/build.sh#L3-L5
Does that cause any trouble if we exchange `gcc` by `llvm` on `osx` for `google-sparsehash`?

yesimon:
Give it a shot! Normally there's a good reason for it and fails with `llvm`. However, if `abyss` works fine with gcc why change it? Lots of bio software are built using `gcc` without much thought for `clang` compatibility

druvus:
conda-forge also got `sparsehash` which is based on the same source. Maybe worth trying using it instead ? I checked and several bioconda recipes are now using the conda-forge version instead of the bioconda version. 

Should we retire `google-sparsehash` recipe in favour for conda-forge `sparsehash` and update the 2-3 packages that are depending on `google-sparsehash`?

mbargull:
@yesimon: I was eying at `conda-forge`'s `toolchain`/`toolchain3`, thus was wondering what'd be best in terms of compatibility/unified build environments for each platform, which would favor `llvm`.

@druvus: I wasn't aware of `conda-forge::sparsehash`! Thanks for pointing that out. Yes that's exactly what we should do. It's only `abyss`, `stride` and `tedna` which depend on `google-sparsehash`. I'll tend to this.

mbargull:
At last, successful builds on both platforms with `conda-forge::sparsehash`!
I'll update `1.5.2`, `1.9.0`, `2.0.2-k128` in a separate PR.

mbargull:
The patch was submitted and merged upstream, hence it can be removed for later versions: https://github.com/bcgsc/abyss/pull/174

yesimon:
Looks great!

yesimon:
Quick question: should `--enable-maxk=96` be set? It seems like it will change the defaults.

mbargull:
Good question! I just checked the default values:
```
$ grep enable_maxk= abyss-*/configure
abyss-1.5.2/configure:  enable_maxk=64
abyss-1.9.0/configure:  enable_maxk=96
abyss-2.0.1/configure:  enable_maxk=128
abyss-2.0.2/configure:  enable_maxk=128
```
Looks like @druvus added `--enable-maxk=96` in the initial commit (version `1.9.0` at that point that already has a default of 96). This was then unchanged for the `2.0.1` update (actual default: 128). @jerowe then added a `2.0.2-k128` build which uses the actual default explicitly via `--enable-maxk=128`. Additionally the `1.5.2` recipe from @jamesabbott copied the 96 (instead of the default 64) over.
@druvus, @jerowe, @jamesabbott: Any thoughts?

druvus:
I vote for using the actual default.

mbargull:
Hence `2.0.2-k128` could be removed since would then be equivalent to `abyss`.

jamesabbott:
I'd agree with going to the defaults...1.5.2 is only included for transabyss compatibility,  and read lengths and kmer sizes would typically be shorter in transcriptome sequencing so there would be no impact on utility by going for the default. Makes more sense from a  user perspective to get a default config as standard.

mbargull:
Great, thanks @druvus and @jamesabbott!
@jerowe: Are you ok with me removing the recipe for `abyss-k128` in a subsequent PR once this one is merged?

PR #6874, opened 2017-11-22
TASSEL Build Recipes

ysu20:

This PR include build recipes for TASSEL5, 4 and 3.


recipes/tassel/build.sh
```diff
@@ -0,0 +1,27 @@
...
+# Fix shabang lines
+sed -i.bak '1 s|^.*$|#!/usr/bin/env perl|g' $TASSEL_HOME/*.pl
+
+# Create wrapper script in bin to call TASSEL's launch scripts
+cat >"${BINARY_HOME}/run_pipeline.sh" <<EOF
+exec $CONDA_PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM/run_pipeline.pl "\$@"
+EOF
+
+cat >"${BINARY_HOME}/run_anything.sh" <<EOF
+exec $CONDA_PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM/run_anything.pl "\$@"
```
bgruening: Can we use a symlink here instead?

ysu20: Unfortunately no.  If I created symbolic links for TASSEL's launch wrappers, they would fail to construct Java's CLASSPATH and to find the jar file correctly when they are invoked via the symbolic links.  This is because they would use the file location of the symbolic links rather than the file location of the actual wrappers as the base path to find other libraries.

recipes/tassel/3.0.137/build.sh
```diff
@@ -0,0 +1,31 @@
+BINARY_HOME=$PREFIX/bin
+TASSEL_HOME=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+
+# Download TASSEL4 from its public repo
+git clone git://git.code.sf.net/p/tassel/tassel3-standalone tassel3-standalone
+cd tassel3-standalone
+git checkout V$PKG_VERSION
```
mbargull: Why isn't the source given in `meta.yaml` like so:
```yaml
source:
  git_url: https://git.code.sf.net/p/tassel/tassel3-standalone.git
  git_rev: V{{ version }}
```

mbargull: I'm not too familiar with Git on SourceForge, but is it possible (and reliable) to get source tarballs from there?

recipes/tassel/3.0.137/build.sh
```diff
@@ -0,0 +1,31 @@
...
+cat >"${BINARY_HOME}/run_pipeline.sh" <<EOF
+exec $CONDA_PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM/run_pipeline.pl "\$@"
+EOF
+
+cat >"${BINARY_HOME}/run_anything.sh" <<EOF
+exec $CONDA_PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM/run_anything.pl "\$@"
+EOF
+
+cat >"${BINARY_HOME}/start_tassel.sh" <<EOF
+exec $CONDA_PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM/start_tassel.pl "\$@"
```
mbargull: We shouldn't rely on `$CONDA_PREFIX` while building, but use `$PREFIX` instead. Since those variables are expanded at build time, you can just use `$TASSEL_HOME` instead of `$CONDA_PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM` in all three wrapper scripts.

recipes/tassel/build.sh
ysu20: Actually, I don't really like my workaround either.  I will crate patches for TASSEL's launch wrappers and make them work when they are called via symbolic links.

recipes/tassel/3.0.137/meta.yaml
```diff
@@ -0,0 +1,24 @@
+{% set name = "TASSEL" %}
+{% set version = "3.0.137" %}
```
mbargull: Are there any reasons to use this specific version?
Looks like `V3.0.174` is the latest: https://sourceforge.net/p/tassel/tassel3-standalone/ref/master/tags/

recipes/tassel/4.3.15/build.sh
```diff
@@ -0,0 +1,31 @@
+BINARY_HOME=$PREFIX/bin
+TASSEL_HOME=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+
+# Download TASSEL4 from its public repo
+git clone git://git.code.sf.net/p/tassel/tassel4-standalone tassel4-standalone
+cd tassel4-standalone
+git checkout V$PKG_VERSION
```
mbargull: Same as above: In `meta.yaml`:
```yaml
source:
  git_url: https://git.code.sf.net/p/tassel/tassel4-standalone.git
  git_rev: V{{ version }}
```

recipes/tassel/3.0.137/build.sh
ysu20: Changed to download snapshot from SourceForge in the latest commit.

recipes/tassel/3.0.137/build.sh
ysu20: Removed any reference to ```CONDA_PREFIX``` in the latest commit.

recipes/tassel/3.0.137/meta.yaml
ysu20: It was a mistake.  I have revised the recipe to build for V3.0.174 in the latest commit.

recipes/tassel/4.3.15/build.sh
ysu20: Changed to download snapshot from SourceForge in the latest commit.

ysu20:
Hi guys, thanks for all your feedback and suggestions.  I have updated the PR to address all your concerns.

ysu20:
Hi @bgruening and @mbargull, do you have any comments for the revised PR? If you are happy with my changes, would you mind accept this PR and merge it to biconda?

recipes/tassel/build.sh
```diff
@@ -0,0 +1,14 @@
+BINARY_HOME=$PREFIX/bin 
+TASSEL_HOME=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+
+# Copy source to the conda environment
+mkdir -p $TASSEL_HOME 
+cp -R $SRC_DIR/* $TASSEL_HOME/
+
+# Convert CRLF in wrappers to LF
+sed -i $'s/.$//' $TASSEL_HOME/*.pl
```
mbargull: Does the leading `$` for the `sed` have any purpose?
However, maybe use
```bash
sed -i 's/\r$//' $TASSEL_HOME/*.pl
```
instead. That way the `sed` won't cause any trouble in case the TASSEL developers at some point in the future decide to use Unix line endings.

ysu20:
Thank you @mbargull for your suggestions. Revised the PR accordingly.

mbargull:
Thank you, @ysu20!

ysu20:
Hi @mbargull, thanks for merging this PR.  One silly question though, when I try to search for tassel via conda, the only version returned is the latest, although my PR includes build recipes for older version of TASSEL. Do I need to do anything specific to get them published?

mbargull:
No silly question at all! Apparently SourceForge creates the Git snapshots only on demand and removes them after a certain while. Because of that the older versions were not built on the `master` branch. I've submitted a change to use `git_url`/`git_rev` for them in #7015.

PR #6886, opened 2017-11-22
Ensure 2 perl wrapper scripts do not mask failures

peterjc:

e.g. Using -pfamB with pfam_scan.pl should return error code 1, but as written the wrapper script returned 0.

I've opted to use the simple one-line approach to setting the variable, which ensures the call is the last line of the script - and thus its return value gets passed back.

### DIFF
--- a/recipes/pfam_scan/build.sh
+++ b/recipes/pfam_scan/build.sh
@@ -9,8 +9,7 @@ cp -r * $outdir
 # Short wrapper script which sets PERL5LIB while running
 cat > $PREFIX/bin/pfam_scan.pl <<EOF
 #!/bin/bash
-export PERL5LIB=$outdir:$outdir/
-$outdir/pfam_scan.pl \$@
-unset PERL5LIB
+# Last line of script to ensure any failure return code is passed on:
+PERL5LIB=$outdir:$PERL5LIB $outdir/pfam_scan.pl \$@
 EOF
 chmod +x $PREFIX/bin/pfam_scan.pl
--- a/recipes/pfam_scan/build.sh
+++ b/recipes/pfam_scan/build.sh
@@ -10,6 +10,6 @@ cp -r * $outdir
 cat > $PREFIX/bin/pfam_scan.pl <<EOF
 #!/bin/bash
 # Last line of script to ensure any failure return code is passed on:
-PERL5LIB=$outdir:$PERL5LIB $outdir/pfam_scan.pl \$@
+PERL5LIB=$outdir:$outdir/treg_comparator exec $outdir/pfam_scan.pl "\$@"
 EOF
 chmod +x $PREFIX/bin/pfam_scan.pl
--- a/recipes/pfam_scan/build.sh
+++ b/recipes/pfam_scan/build.sh
@@ -9,7 +9,6 @@ cp -r * $outdir
 # Short wrapper script which sets PERL5LIB while running
 cat > $PREFIX/bin/pfam_scan.pl <<EOF
 #!/bin/bash
-# Last line of script to ensure any failure return code is passed on:
-PERL5LIB=$outdir:$outdir/treg_comparator exec $outdir/pfam_scan.pl "\$@"
+PERL5LIB=$outdir exec $outdir/pfam_scan.pl "\$@"
 EOF
 chmod +x $PREFIX/bin/pfam_scan.pl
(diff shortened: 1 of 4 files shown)


peterjc:
Fixing my silly mistake while fighting network glitches...

peterjc:
Fixing my silly mistake while fighting network glitches...

peterjc:
I think the recent TravisCI failure was due to trawler's use of the system perl:

```
+ source /opt/conda/bin/activate /opt/conda/conda-bld/trawler_1511354797317/_t_env
+ /bin/bash -x -e /opt/conda/conda-bld/trawler_1511354797317/test_tmp/run_test.sh
+ trawler
/opt/conda/conda-bld/trawler_1511354797317/_t_env/bin/trawler: /opt/conda/conda-bld/trawler_1511354797317/_t_env/share/trawler-2.0-1/bin/trawler.pl: /usr/bin/perl: bad interpreter: No such file or directory
```

I've tweaked the recipe to update the trawler hashbangs to use ``/usr/bin/env perl`` instead.

Cross reference https://github.com/Ramialison-Lab-ARMI/Trawler-2.0/issues/1

recipes/trawler/build.sh
```diff
@@ -2,15 +2,21 @@
...
 cp -r * $outdir
 
 # Short wrapper script
 cat > $PREFIX/bin/trawler <<EOF
 #!/bin/bash
-export PERL5LIB=$outdir:$outdir/treg_comparator
-$outdir/bin/trawler.pl \$@
-unset PERL5LIB
+# Last line of script to ensure any failure return code is passed on:
+PERL5LIB=$outdir:$outdir/treg_comparator $outdir/bin/trawler.pl \$@
```
mbargull: Good call on that change! You could even add two more enhancements with
```bash
PERL5LIB=$outdir:$outdir/treg_comparator exec $outdir/bin/trawler.pl "\$@"
```
* `exec`: replace caller process with child, see, e.g., https://unix.stackexchange.com/questions/279024/reason-for-exec-in-wrapper-scripts
* `"$@"`: avoid splitting of input parameters, e.g., `trawler "1 2" 3` will call `trawler.pl` with `["1 2", "3"]` instead of `["1", "2", "3"]`

peterjc:
The trawler test is failing due to https://github.com/Ramialison-Lab-ARMI/Trawler-2.0/issues/2 (previously the non-zero return code from the trawler usage was masked by the bug in our wrapper script).

peterjc:
Hmm, the ``pfam_scan.pl`` tests are now failing (but they used to work), looks like a missing Perl package?  Line 36 of is ``Bio/Pfam/Scan/PfamScan.pm``

```
use IPC::Run qw( start finish );
```

Now get:

```
SERR Can't locate IPC/Run.pm in @INC (you may need to install the IPC::Run module)
```

There are two third party packages on conda for this, https://anaconda.org/search?q=perl-ipc-cmd

recipes/pfam_scan/build.sh
```diff
@@ -9,8 +9,7 @@ cp -r * $outdir
 # Short wrapper script which sets PERL5LIB while running
 cat > $PREFIX/bin/pfam_scan.pl <<EOF
 #!/bin/bash
-export PERL5LIB=$outdir:$outdir/
-$outdir/pfam_scan.pl \$@
-unset PERL5LIB
+# Last line of script to ensure any failure return code is passed on:
+PERL5LIB=$outdir:$outdir/treg_comparator exec $outdir/pfam_scan.pl "\$@"
```
mbargull: Copy-pasted `:$outdir/treg_comparator` over :wink:. Doesn't really harm, but a single `PERL5LIB=$outdir` should suffice:
```bash
PERL5LIB=$outdir exec $outdir/pfam_scan.pl "\$@"
```
Also, the comments in the wrapper scripts can now be deleted, too. With `exec` the following code shouldn't be executed anyway.

mbargull:
Minor additional remarks, otherwise looks good to me!

recipes/pfam_scan/build.sh
peterjc: Thanks

mbargull:
Thanks a bunch, @peterjc!

peterjc:
Thanks for the review and feedback. I've learnt another Unix trick (``exec``) from this.

PR #6888, opened 2017-11-22
HG-CoLoR tool

cmonjeau:

### DIFF
--- a/recipes/hg-color/build.sh
+++ b/recipes/hg-color/build.sh
@@ -1,16 +1,30 @@
 #!/bin/bash
 
 mkdir -p ${PREFIX}/bin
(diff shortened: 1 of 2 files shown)


abretaud:
Hey,
There is something exotic in this tool: it recompiles a modified version of the pgsa lib (which itself is in conda thanks to #6868).
Currently, this HG-CoLoR PR wget the pgsa source code from github and compile its own version (with modified code).

Do you think of a better way to write this recipe? We could modify pgsa recipe to copy its source code in $PREFIX and then use it in HG-CoLoR recipe...

any thoughts?

abretaud:
Merging as is as we don't see a better solution and we prefer not to pollute the pgsa recipe

PR #6893, opened 2017-11-22
Remove google-sparsehash

mbargull:

This removes `bioconda::google-sparsehash` in favor of `conda-forge::sparsehash` which is essentially the same package.
`stride` and `tedna` (together with `abyss` the only recipes depending on `google-sparsehash`) are updated to depend on `sparsehash` instead.


bgruening:
Thanks @mbargull!

PR #6896, opened 2017-11-23
vgraph recipe: https://github.com/bioinformed/vgraph

pvanheus:


recipes/vgraph/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+{% set version = "1.2.0" %}
+# held back until github archives work
+#{% set sha256 = "f727208daeb0a23e5248ddbe2fe2a3fdd8ca5e510cc4ce7aa6bbccb9b2bc5805" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  held back until github archives work
```
mbargull: typo, missing `#`

recipes/vgraph/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+    - setuptools
+    - setuptools_scm 1.15.0
+    - setuptools_scm_git_archive 1.0
+    - cython
+    - pysam
+    - pytest
+    - pytest-runner
+    - coverage
+  run:
+    - python >3
```
mbargull: If this isn't `noarch: python`, `conda-build 2.1.x` will create defunct packages with this.
Just remove the ` >3`.

recipes/vgraph/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+    - cython
```
mbargull: Is `cython` necessary as a `run` requirement?

recipes/vgraph/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+  build:
+    - python >3
+    - setuptools
+    - setuptools_scm 1.15.0
+    - setuptools_scm_git_archive 1.0
+    - cython
+    - pysam
+    - pytest
+    - pytest-runner
+    - coverage
```
mbargull: `pytest`, `pytest-runner` and `coverage` are only needed to run the tests so they could be removed from the build requirement. But sadly `pytest-runner` is added to `setup_requires` in `setup.py`. Hence, I think `setuptools` will complain if we remove it.
Two options for `pytest-runner`:
 * leave it as a build requirement
 * path `setup.py` to remove it from `setup_requires`

(+ maybe open an issue on their issue tracker for the `setup_requires` change)

recipes/vgraph/meta.yaml
pvanheus: When I tried to test without it, I got an error - it was looking for some cython modules.

pvanheus: I think this might be because of the 'install_requires' in setup.py - I'm testing that. It is not used as an import by any of the actual code.

recipes/vgraph/meta.yaml
pvanheus: So `pytest-runner` etc should be in `tests_require` but not in `setup_requires`?

recipes/vgraph/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+  git_url: https://github.com/{{ github_author }}/{{ name }}.git
+  git_rev: v{{ version }} 
+  patches:
+    - setup_py.patch
+
+build:
+  number: 0
+  skip: True  # [py<35 and py==36]
+  script: python setup.py install --single-version-externally-managed --record record.txt
+
```
mbargull:
```diff
+  entry_points:
+    - vgraph = vgraph.vgraph:main
```
That way we let Conda create the the entry point and don't get the `setuptools`' one.

recipes/vgraph/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+  #url: http://github.com/{{ github_author }}/{{ name }}/archive/v{{ version }}.tar.gz
+  #sha256: {{ sha256 }}
+  git_url: https://github.com/{{ github_author }}/{{ name }}.git
+  git_rev: v{{ version }} 
+  patches:
+    - setup_py.patch
+
+build:
+  number: 0
+  skip: True  # [py<35 and py==36]
```
mbargull: This would be `[py<35 and py==36]`, though there is currently not need to skip Python 3.6 (build failure was due to `setuptools_scm_git_archive` availability, but that isn't need right now anyways). Even Python 3.4 should be ok (although Bioconda  doesn't build against that anymore).

So I'd just use one of `[py<34]`, `[not py3k]`, `[py2k]`, `[py27]`.

recipes/vgraph/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+  script: python setup.py install --single-version-externally-managed --record record.txt
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - setuptools_scm 1.15.0
+    - setuptools_scm_git_archive 1.0
```
mbargull: `setuptools_scm_git_archive` not needed for now

recipes/vgraph/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+    - cython
+    - pysam
```
mbargull: `pysam` isn't needed while building if `setup.py` is patched accordingly.

recipes/vgraph/setup_py.patch
```diff
@@ -0,0 +1,15 @@
...
++++ setup.py	2017-11-23 13:03:28.000000000 +0200
+@@ -25,9 +25,9 @@
+ from distutils.extension import Extension
+ 
+ 
+-install_requires = ['Cython', 'pysam']
+-setup_requires   = ['pytest-runner', 'pysam', 'setuptools_scm==1.15.0']
+-tests_require    = ['pytest', 'coverage']
++install_requires = ['pysam']
++setup_requires   = ['pysam', 'setuptools_scm==1.15.0', 'setuptools_scm_git_archive==1.0']
```
mbargull:
```diff
+setup_requires   = ['setuptools_scm==1.15.0'']
```
`'pysam'` isn't required for building, but we could add `'Cython'` here.
`setuptools_scm_git_archive` isn't needed for now.

recipes/vgraph/meta.yaml
mbargull: Yeah, that's an entry point from `setuptools`/`pkg_resources` that checks `install_requires`. If we let Conda create the entry point, we get one that doesn't use `pkg_resources` and don't get that error.

recipes/vgraph/meta.yaml
mbargull: You got it right :D

mbargull:
@pvanheus: If the developers accept your `setuptools_scm_git_archive` PR, then you'll need some Python 3.6 builds of that in the future. I opened an issue for that: https://github.com/conda-forge/setuptools_scm_git_archive-feedstock/issues/1

johanneskoester:
We have made changes to our build system (migrated from Travis CI to Circle CI). Please update your fork with the latest upstream/master.

daler:
Closing due to inactivity., please re-open if needed.

PR #6899, opened 2017-11-23
abyss-k128: remove recipe in favor of abyss-2.0.2-boost1.64_3

mbargull:

Followup on https://github.com/bioconda/bioconda-recipes/pull/6867#issuecomment-346438576
@jerowe: I suppose this one is for you to decide. From build `abyss-2.0.2-boost1.64_3` onward the default value of `maxk=128` from `abyss 2.0` is used. That makes `abyss-k128` redundant so that we could remove its recipe.


jerowe:
@mbargull , if 128 is the new default this is redundant. Would you happen to know about the other compilations? I mainly build things for researchers at my job, and I know my boss very particularly wanted the other compilation options. Removing the recipe doesn't remove the package from conda.

jerowe:
Thanks!

jerowe:
Yes, the 96 kmer size.

mbargull:
Three options come to mind:
* explicitly depend on an older build `abyss-2.0.2-boost1.64_1` -- But I wouldn't like to depend on that implicit assumption, "builds `<2` => 96, `>=2` => 128" if I were you.
* use `1.9.0` which has a default of 96 -- of course only applicable if `1.9` and `2.0` are otherwise identical for your use case.
* create a `abyss-k96` package

PR #6907, opened 2017-11-24
update salmon

rob-p:

### DIFF
--- a/recipes/salmon/build.sh
+++ b/recipes/salmon/build.sh
@@ -2,8 +2,8 @@
 set -eu -o pipefail
 
 if [[ "$(uname)" == Darwin ]]; then
-    CC=clang
-    CXX=clang++
+    export CC=clang
+    export CXX=clang++
 fi
 
 mkdir -p $PREFIX/bin
--- a/recipes/salmon/build.sh
+++ b/recipes/salmon/build.sh
@@ -4,6 +4,8 @@ set -eu -o pipefail
 if [[ "$(uname)" == Darwin ]]; then
     export CC=clang
     export CXX=clang++
+    export CXXFLAGS="${CXXFLAGS} -I${PREFIX}/include/c++/v1"
+    export LDFLAGS="${LDFLAGS} -L${PREFIX}/lib -Wl,-rpath,${PREFIX}/lib"
 fi
 
 mkdir -p $PREFIX/bin
--- a/recipes/salmon/build.sh
+++ b/recipes/salmon/build.sh
@@ -1,6 +1,11 @@
 #!/bin/bash
 set -eu -o pipefail
 
+if [[ "$(uname)" == Darwin ]]; then
+    CC=clang
+    CXX=clang++
+fi
+
 mkdir -p $PREFIX/bin
 mkdir -p $PREFIX/lib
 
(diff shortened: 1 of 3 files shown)


rob-p:
It looks like the issue is due to the usage of `thread_local` in the Fasta Parser on OSX.  Travis says this is "not supported on the current target".  However, `thread_local` is a simple C++11 feature.  What exactly is the config being used to do the CI testing on OSX?

jerowe:
@bioconda/osx, any ideas?

mbargull:
Looks like a newer version of `clang` is needed: https://stackoverflow.com/a/29929949.
Will a build dependency on `clangdev` let us use a newer one?

rob-p:
A newer version of clang would be ideal.  I could remove the thread_local specifier for those variables all together, but that could have an adverse effect on performance.

mbargull:
Without knowing the code (/performance critical paths), I'd also rather avoid the patch, yes. Maybe try adding `clangdev  # [osx]` as build requirement. Not sure if we'd need an additional
```bas
if [[ "$(uname)" == Darwin ]]; then
    CC=clang
    CXX=clang++
fi
```
or something in `build.sh` or if `CXX` is `clang++` anyway. Not sure though if there are any other implications by not using the default `clang`. (And I know next to nothing of macOS -- that's that fancy pear-logoed system, right?! :P)

rob-p:
@mbargull ,

  Simply adding `clangdev # [osx]` doesn't seem to work.  Is there some other way to force a different clang version during the CI build?

mbargull:
Maybe try adding 
```bash
if [[ "$(uname)" == Darwin ]]; then
    CC=clang
    CXX=clang++
fi
```
to the top of `build.sh`. Alternatively there are also some `cmake` parameters with which those can be set, but I don't them from the top of my head.

rob-p:
But its building using clang, even without the llvmdev dependency.  The issue is that i need a newer version of clang (apple clang 8 or later).

mbargull:
Yes, but it's picking up the default one, see: https://travis-ci.org/bioconda/bioconda-recipes/jobs/306930733#L325.
If you use `clangdev` as a build dependency AND set `CC`/`CCX` in `build.sh`, it should pick the newer `clang`/`clang++` from the build environment.

rob-p:
Ok, ill give that a try. Thanks!

rob-p:
Still doesnt seem to force the use of a newer clang.  Is there a specific path or env variable i can use to set clang as the new one instead of the ild one?

mbargull:
Argh, I'm a dummy and forgot the `export`, sorry! Use the following instead:
```bash
if [[ "$(uname)" == Darwin ]]; then
    export CC=clang
    export CXX=clang++
fi
```

rob-p:
Sighh, this seems to lead to quite a different error (now it dies due to failure to include a header in Jellyfish --- `<new>`).

mbargull:
> Sighh

Indeed...
The error meant that it couldn't find the standard library header. I pushed a commit that added `libcxx` to the dependencies and uses the same version for `clangdev` to have them synchronized.

rob-p:
W00t, the patch seems to work.  Awesome! Thanks for all the help!

mbargull:
Nice, glad to hear that!

PR #6908, opened 2017-11-24
Quasitools py3

mbargull:

Enables Python 3 builds, updates dependencies, uses Conda entry point.

### DIFF
--- a/recipes/quasitools/build.sh
+++ b/recipes/quasitools/build.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-python setup.py install
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
--- a/recipes/quasitools/meta.yaml
+++ b/recipes/quasitools/meta.yaml
@@ -15,15 +15,11 @@ build:
 requirements:
   build:
     - python
-    - scipy
-    - samtools ==0.1.19
-    - pysam ==0.9
-    - click
-    - biopython
-    - bowtie2
+    - setuptools
 
   run:
     - python
+    - numpy
     - scipy
     - samtools ==0.1.19
     - pysam ==0.9
--- a/recipes/quasitools/meta.yaml
+++ b/recipes/quasitools/meta.yaml
@@ -1,15 +1,16 @@
+{% set version = "0.2.1" %}
+
 package:
   name: quasitools
-  version: "0.2.1"
+  version: {{ version }}
 
 source:
-  fn: quasitools-0.2.1.tar.gz
-  url: https://github.com/phac-nml/quasitools/archive/v0.2.1.tar.gz
+  fn: quasitools-{{ version }}.tar.gz
+  url: https://github.com/phac-nml/quasitools/archive/v{{ version }}.tar.gz
   md5: bf3c01af9c30ee5e0add80a141f8678c
 
 build:
- # click needs to be patched in order to properly detect unicode in py3k
-  skip: True # [py3k]
+  number: 1
 
 requirements:
   build:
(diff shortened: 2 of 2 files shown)


mbargull:
@camytran: I relaxed the dependency on `pysam` so that `pysam 0.9.1` releases can be used since currently there are not Python 3.6 builds of `pysam 0.9.0` in the Bioconda channel.
Is that change fine, i.e., are `quasitools` also compatible with `pysam 0.9.1`, or should we rather add Python 3.6 builds for `pysam 0.9.0` instead.

mbargull:
I didn't really pay attention to the `samtools ==0.1.19` before, which is also specified in https://github.com/phac-nml/quasitools/blob/v0.2.1/README.md.
If this is accurate, then we'd actually have to use a `pysam` version between `0.7.5` and `0.8.1` (for which builds/recipes are currently missing). See the release notes:

http://pysam.readthedocs.io/en/latest/release.html#release-0-7-5
> switch to samtools 0.1.19

http://pysam.readthedocs.io/en/latest/release.html#release-0-8-1
> Pysam now wraps htslib and samtools versions 1.1.

http://pysam.readthedocs.io/en/latest/release.html#release-0-9-0
> The 0.9.0 release upgrades htslib to htslib 1.3 and numerous other enchancements and bugfixes. 

The `quasitools` recipe's dependency on `pysam ==0.9.0` only worked since Conda picked up some old builds for that version which did not correctly pin their dependency on `samtools 1.3`, cf.:
https://github.com/bioconda/bioconda-recipes/blob/c2027455a09aa4df9956e4eccd1a24ea81a8f4aa/recipes/pysam/meta.yaml#L25
https://github.com/bioconda/bioconda-recipes/blob/4b1f6dd8a9a38616072ddb08de64fc7713642cc7/recipes/pysam/meta.yaml#L27

camytran:
Hi @mbargull. The samtools dependency for 0.1.19 is accurate. I just re-tested quasitools with different pysam versions and found that it is compatible with pysam>=0.8.1, so the quasitools recipe can be updated accordingly.

mbargull:
My question is, does it work with `pysam <=0.8.0`? Those are the versions that are actually compatible with `samtools 0.1.19`.

mbargull:
The problem here is that `pysam` is bound to certain `samtools` versions. So if we want to use `pysam` with `samtools 0.1.19`, we'd have to use an older `pysam` version.
The fact that you could/can install `quasitools` with `pysam 0.9.0` and `samtools 0.1.19` at the same time is actually a flaw (from my point of view even a bug) in how Conda works. It always includes outdated builds of packages in its dependency resolution and hence is free to pick older build with non-accurate/defect run requirement over newer fixed builds. That is exactly the case for `pysam 0.9.0` for which the correct `samtools ==1.3` requirement was only added in later builds.

It might be that `quasitools` in fact uses parts of `pysam` that are independent on the `samtools` version that is used. But sadly we have to "take all or nothing" when it comes to proper dependency resolutions in the [Conda] package environments.

camytran:
> we'd actually have to use a pysam version between 0.7.5 and 0.8.1

My mistake, I thought samtools 0.1.19 was compatible with pysam 0.7.5-0.8.1, _inclusive_. Unfortunately, quasitools requires pysam 0.8.1 and up. I am working on updating quasitools to work with samtools 1.3 so that we can use pysam >=0.8.1 as is required. Thanks for the explanation about dependencies.

mbargull:
> >we'd actually have to use a pysam version between 0.7.5 and 0.8.1
> 
> My mistake, I thought samtools 0.1.19 was compatible with pysam 0.7.5-0.8.1, inclusive. 

Oh wow, now as I see it in your quote -- no, no, my mistake -- my wording wasn't clear at all! XD

> I am working on updating quasitools to work with samtools 1.3 so that we can use pysam >=0.8.1 as is required. Thanks for the explanation about dependencies.

That's more than I would ever had expected. Thank you!

camytran:
Hi @mbargull, there is a new release of quasitools 0.2.2](https://github.com/phac-nml/quasitools/releases), so the meta.yml file can be updated now to reflect this new release and the samtools and pysam dependencies can be bumped to samtools==1.3 and pysam>=0.8.1.

mbargull:
Neat! Should this then be exactly `samtools ==1.3` or is `samtools >=1.3` likely fine, too?

camytran:
Yes you're right, samtools >=1.3 will work :smiley:

camytran:
Thanks @mbargull! :smiley:

mbargull:
Thanks to you, too!

PR #6910, opened 2017-11-24
adding the OTHER python newick package

nickp60:
Again, I'm trying to get Ragout scaffolder on bioconda, and as such I am attempting to add some missing pre-requisites

### DIFF
--- a/recipes/python-mailund-newick/build.sh
+++ b/recipes/python-mailund-newick/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py install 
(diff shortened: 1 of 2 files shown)


nickp60:
Could anyone give me some insight? This package uses py27. I first was getting the linting error ```setup_py_install_args```.  So I tried adding ```$PYTHON setup.py install --single-version-externally-managed --record=record.txt``` but now I am getting the ```error: option --single-version-externally-managed not recognized
``` error, as this is an old version of setup tools.

This seems like a catch 22, how can I solve this?

druvus:
It is using distutils and not setuptools so you need add `$PYTHON setup.py install --record=record.txt`

nickp60:
@druvus Thanks so much!

PR #6912, opened 2017-11-24
remove joblib recipe

kain88-de:
The version here is 2 years old. The conda-forge version is better maintained. On conda-forge I can install 0.11. on this channel only version 0.9.3 (from 2015) is available.


mbargull:
Please only remove the recipe itself. Don't change the dependencies of the other packages (those are minimum requirements anyways, hence the newer releases from `conda-forge` always satisfy them).

kain88-de:
> (those are minimum requirements anyways, hence the newer releases from conda-forge always satisfy them).

That is the point. They should be forced to pick up the newer version from conda-forge. This is the only sane way I know of to remove a package with the channel pinning from conda forge. For example, if you have a condarc like this.

```yaml
channels:
  - bioconda
  - conda-forge
  - defaults
```

Then conda will *always* use the old joblib version available in bioconda when it's installed as a dependency. Requiring a larger version number will ensure that the newer package version is installed.

mbargull:
Ah, I must suddenly became colorblind as I thought you removed the constraints rather than add them. Really sorry about that!
Yes, without those minimum version requirements this is currently something you can only make `conda` do with `--no-channel-priority`. So I understand the need for this workaround but I am not happy with it -- and cannot offer an alternative at the same time. I'd actually like to here some other opinions on this matter. @bioconda/core: thoughts?
If we'll add those constraints, we should add at least a comment as to why this was done.

bgruening:
@mbargull thats the only solution that works at the moment. An other way, also suboptimal, is to mark the bioconda package as `broken` and remove the `main` label.

kain88-de:
> If we'll add those constraints, we should add at least a comment as to why this was done.

I can add a comment once the discussion has reached a conclusion.

mbargull:
@bgruening: Yeah, it's a bit like choosing between pest and cholera. Seeing that there is only `joblib 0.9.3` and not multiple versions on the Bioconda channel, I'd lean more to making them unavailable via a label. The existing label `deprecated` would be more appropriate than `broken`, though.

> I can add a comment once the discussion has reached a conclusion.

@kain88-de: Thank you for that and thanks for doing this overall!

kain88-de:
> @kain88-de: Thank you for that and thanks for doing this overall!

This version problem has cost me 20min of work time today. Fixing it took less time (yay emacs) and hopefully helps others avoid this.

bgruening:
Ok, I will deprecate it. @kain88-de can you revert the strict pinning?

kain88-de:
done

recipes/triform2/meta.yaml
```diff
@@ -33,7 +33,7 @@ requirements:
     - pandas
     - numpy
     - natsort
-    - joblib
+    - jobli
```
bgruening: this looks wrong

recipes/pomegranate/meta.yaml
```diff
@@ -15,15 +15,15 @@ requirements:
     - python
     - cython >=0.22.1
     - numpy >=1.8.0
-    - joblib >=0.9.0b4
+    - joblib
```
bgruening: Not sure why this restriction was in place, but better leave it.

kain88-de:
Oh yeah sorry. You can change it yourself. Github allows you to push to my repo on that branch. 

 I won’t get back to this before Monday.

PR #6916, opened 2017-11-25
Make bioconductor-dexseq py3 compatible

mblue9:

This adds 2to3 into the recipe for dexseq_count.py and dexseq_prepare_annotation.py to avoid errors like:

raise ValueError, "Same name found on two chromosomes: %s, %s" % ( str(l[i]), str(l[i+1]) )
^
SyntaxError: invalid syntax


recipes/bioconductor-dexseq/build.sh
```diff
@@ -5,6 +5,11 @@ $R CMD INSTALL --build .
 
 python_scripts=$PREFIX/lib/R/library/DEXSeq/python_scripts
 
+if [ "$PY3K" == 1 ]; then
+    2to3 -w $python_scripts/dexseq_count.py
+    2to3 -w $python_scripts/dexseq_prepare_annotation.py
+fi
```
bgruening: It seem that this collide with the sed command a few lines below on OSX.
Maybe patching both scripts would be more clean?

mblue9: Ok, do you mean I should submit py3 changes for these scripts upstream to the dexseq authors? (I don't know that much python but I guess I could try)

bgruening: Ideally yes, but the patch can also be included here for the time being.

bgruening:
@mblue9 thanks!

PR #6921, opened 2017-11-26
adding fgap recipe

nickp60:
Hi all, this represents the first (to my knowledge) use of an octave script in bioconda.  What I have done is use the build stage to create a runner script that tells octave where to find the source and passes the appropriate args. Comments and critiques welcome;  I am aware this is far from optimal, but lets consider it a starting point!

https://github.com/pirovc/fgap

### DIFF
--- a/recipes/fgap/build.sh
+++ b/recipes/fgap/build.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
-echo "#!/bin/bash" >> FGAP
-echo "octave --path \"$PREFIX/bin/\" --no-gui --eval \"fgap \$@\"" >> FGAP
+cat > FGAP <<EOF
+#!/bin/bash
+octave --path "$PREFIX/bin/" --no-gui --eval "fgap \$@"
+EOF
+
 chmod +x FGAP
 cp fgap.m $PREFIX/bin/
 cp FGAP $PREFIX/bin/
--- a/recipes/fgap/build.sh
+++ b/recipes/fgap/build.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+echo "#!/bin/bash" >> FGAP
+ls
+echo "echo 'starting FGAP'" >> FGAP
+echo "octave --path \"$PREFIX/bin/\" --no-gui --eval \"fgap \$@\"" >> FGAP
+chmod +x FGAP
+cp fgap.m $PREFIX/bin/
+cp FGAP $PREFIX/bin/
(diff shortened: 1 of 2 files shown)


recipes/fgap/meta.yaml
```diff
@@ -0,0 +1,24 @@
...
+build:
+  number: 0
+  skip: True  # [osx]
+  
+package:
+  name: fgap
+  version: 1.8.1
+source:
+  fn: fgap-1.8.1.tar.gz
+  # this is nickp60's fork because bioconda needs tagged releases
```
mbargull: It's totally fine to use Git archives for certain commits, e.g., for this project that would be:
```yaml
  url: https://github.com/pirovc/fgap/archive/e3004998f9058e8505b7812544d4c1dbd595a169.tar.gz
```
That way you won't have to maintain a fork only to have tags.
Additionally, there is also nothing wrong by asking upstream if they could tag their releases in the future :wink:

nickp60: @mbargull ooo, I did not know about Git archives!  I move to that until upstream makes a release.  thanks for the tip!

recipes/fgap/meta.yaml
```diff
@@ -0,0 +1,24 @@
...
+  summary: "FGAP: an automated gap closing tool"
+build:
+  number: 0
+  skip: True  # [osx]
+  
+package:
+  name: fgap
+  version: 1.8.1
+source:
+  fn: e3004998f9058e8505b7812544d4c1dbd595a169.tar.gz
```
mbargull: You can still use 
```yaml
  fn: fgap-1.8.1.tar.gz
```
`conda-build` would default to `e3004998f9058e8505b7812544d4c1dbd595a169.tar.gz` if `fn` isn't given, but we can also use a more meaningful name like `fgap-1.8.1.tar.gz` if we want to.

recipes/fgap/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+echo "#!/bin/bash" >> FGAP
+ls
+echo "echo 'starting FGAP'" >> FGAP
```
mbargull: Do we really need the `echo 'starting FGAP'` here?
Couldn't we just use something like:
```bash
cat > FGAP <<EOF
#!/bin/bash
octave --path "$PREFIX/bin/" --no-gui --eval "fgap \$@"
EOF
```
?

mbargull: [Sorry that I added those comments after each other and not as part of a full review. Clicked the wrong button..]

mbargull:
Nice! Thank you, @nickp60!

PR #6931, opened 2017-11-27
Mykrobe

Takadonet:


recipes/ga4ghmongo/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  md5: 6557d21feb7540ad7cc2548036563a2b
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - mongoengine
```
mbargull: `mongoengine` can be removed from the build requirements.

recipes/ga4ghmongo/meta.yaml
```diff
@@ -14,7 +14,6 @@ requirements:
   build:
     - python
     - setuptools
-    - mongoengine
```
mbargull: Wow, you're fast :smile:

Takadonet: 👍 Have two more packages to go after this one.

recipes/mykatlas/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  # Python imports
+  imports:
+    - mykatlas
+  commands:
+    - atlas --help
+
+about:
+  home: http://github.com/phelimb/atlas
+  license: MIT
+  summary: '.'
```
mbargull: Even though there doesn't seem to be an "official" description in the projects's `setup.py` or `README.md`, can you add a short one?

recipes/mykrobe/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+requirements:
+  build:
+    - python
+    - setuptools
+    - mykatlas
+    - ga4ghmongo
+    - pyvcf
+    - mongoengine
+    - biopython
+    - future
```
mbargull: Apart from `python` and `setuptools`, I don't think we need any other build requirements.

Takadonet:
So looks like there is one more hidden requirement called mccortex31.

recipes/mykrobe/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+  run:
+    - python
+    - mykatlas
+    - ga4ghmongo
```
mbargull: `grep`-ing for `import` on the project's source shows that it's also directly dependent on `monoengine` => add it to the run requirements.

recipes/mykrobe/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  # Python imports
+  imports:
+    - mykrobe
+  commands:
+    - mykrobe --help
+
+
+about:
+  home: https://github.com/iqbal-lab/Mykrobe-predictor
+  license: https://github.com/iqbal-lab/Mykrobe-predictor/blob/master/LICENSE
```
mbargull: There is a special `license_file` field we can use that takes a filename from the source archive:
```yaml
  license: Custom non-commercial license
  license_file: LICENSE
```
(The project's license also requires us to include it in the package this way.)

recipes/mykrobe/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  summary: '.'
```
mbargull: Even though there doesn't seem to be an "official" description in the projects's `setup.py` or `README.md`, can you add a short one?

recipes/mykatlas/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  name: mykatlas
+  version: "0.6.1"
+
+source:
+  fn: mykatlas-0.6.1.tar.gz
+  url: https://pypi.python.org/packages/bf/33/946e43324bc24ba331a1bee06e609832bfbe7a4f07dde1846246fb5a68cf/mykatlas-0.6.1.tar.gz
+  md5: 87d59770ceeb5406a0146c3c51d93ea9
+
+build:
+  number: 0
```
mbargull: Please add the following to the `build:` section to get entry points created by Conda:
```yaml
  entry_points:
    - atlas = mykatlas.atlas_main:main
```

recipes/mykrobe/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  name: mykrobe
+  version: "0.5.6"
+
+source:
+  fn: mykrobe-0.5.6.tar.gz
+  url: https://pypi.python.org/packages/7e/35/a662ec560cfc91f35519e1459de917001ca19a9dbefd53c5b3a2b14122d0/mykrobe-0.5.6.tar.gz
+  md5: 5e912497f98cc192e3f0fbf426210560
+
+build:
+  number: 0
```
mbargull: Please add the following to the `build:` section to get entry points created by Conda:
```yaml
  entry_points:
    - mykrobe = mykrobe.mykrobe_predictor:main
```

mbargull:
> hidden requirement

Who doesn't like those... 😑

recipes/mykrobe/meta.yaml
Takadonet: Added

recipes/mykatlas/meta.yaml
Takadonet: Fixed

recipes/mykrobe/meta.yaml
Takadonet: Fixed

recipes/mykrobe/meta.yaml
Takadonet: Fixed

recipes/mykrobe/meta.yaml
Takadonet: Fixed

recipes/mykrobe/meta.yaml
Takadonet: Since using pipy, the license_file is not part of the source  url:

recipes/mykrobe/meta.yaml
mbargull: I think `pyvcf` and `biopython` aren't direct dependencies.

recipes/mykrobe/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  imports:
+    - mykrobe
+  commands:
+    - mykrobe --help
+
+
+about:
+  home: https://github.com/iqbal-lab/Mykrobe-predictor
+  license: Custom non-commercial license
+  license_file: https://github.com/iqbal-lab/Mykrobe-predictor/blob/master/LICENSE
```
mbargull: > Since using pipy, the license_file is not part of the source url

Bummer! I opened up an issue for this at the project's repository: https://github.com/iqbal-lab/Mykrobe-predictor/issues/142.
Could you then use GitHub as the source for now, i.e., use:
```yaml
source:
  fn: mykrobe-0.5.6.tar.gz
  url: https://github.com/iqbal-lab/Mykrobe-predictor/archive/933301d9f1e673a2a282255426a885ef72e67c07.tar.gz
  sha256: adcf75a8e873192047921947ef6c367fadee9dff09637f0acad7a4c9c4bcf53e
```
and
```yaml
  license_file: LICENSE
```

recipes/mykatlas/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  # Python imports
+  imports:
+    - mykatlas
+  commands:
+    - atlas --help
+
+about:
+  home: http://github.com/phelimb/atlas
+  license: MIT
+  summary: 'Assists in discoverly of antibiotic-resistance with mykrobe'
```
mbargull: `discoverly` typo :wink:

recipes/mykatlas/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - ga4ghmongo
+
+  run:
+    - python
+    - future
+    - biopython
+    - mongoengine
+    - pyvcf
+    - ga4ghmongo
+
```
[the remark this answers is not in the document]
Takadonet: Fixed

recipes/mykrobe/meta.yaml
Takadonet: Fixed

recipes/mykrobe/meta.yaml
Takadonet: Changed to github source

recipes/mykatlas/meta.yaml
Takadonet: fixed

mbargull:
@Takadonet: Everything looks fine and dandy -- apart from that `mccortex31` dependency you mentioned.. I pushed a commit which just uses the included precompiled binary as it looks like quite a bit of work to sift through all `make` targets and dependencies to get things working, see:
https://github.com/Phelimb/atlas/blob/015c2b1fb92bdbedcd1146b4343d5b4fdd8b3b5b/setup.py#L9:L21
https://github.com/Phelimb/mccortex/blob/22410eacc3c416c73dec86a133f063cf1405663e/Makefile
https://github.com/Phelimb/mccortex/blob/22410eacc3c416c73dec86a133f063cf1405663e/libs/Makefile
That also means I had to skip OSX builds.

But if you or any one else is willing to look into building `mccortex` from source, be my guest 😉.

mbargull:
@Takadonet: Nice work on this one! Should we merge or do you want to give compiling `mccortex` a try?

Takadonet:
@mbargull Thanks for all your help! Going to give mccortex a try since without it, mykrobe will not run correctly. One added factor to all of this is that he is using his forked version of mccortex on a specific commit and not a release... Will see how it goes.

mbargull:
That's correct; the links I provided above point exactly to the commit that is being used.
If you are successful in building `mccortex` we could either bundle it with `mykatlas` as it's currently being done or create a new package whose name would reflect the fact that it is a fork, e.g., `mccortex-mykatlas`, `mccortex-phelimb` or the like.
If it's too much trouble to get it to compile, it's fine by me to use the precompiled `mccortex31` for now and revisit the compilation attempt at a later date.

Takadonet:
I am ready just to precompiled mccortex31 and revisit for another day. Or could get one of my co-op students to do it later!

For the precompiled, you thinking of putting it straight into mykatlas then or still its own repo?

Cashalow:
I really like Mykrobe predictor and at some point I wanted to start myself the work of adding it to conda. However, I had some doubts about the licensing of the software and thought this was not possible. Indeed commercial use of it is prohibited without the agreement of the copyright holder. Are you sure this is okay with bioconda distribution ?

I would love to see this coming though.

mbargull:
@Cashalow: To my understanding of the license it should be ok to distribute in on Bioconda. Bioconda itself doesn't intend commercial use and users of the software, i.e., someone who `conda install`s it, should agree to the licensing terms, as it is case for any other software.
The only part of the license that might be debatable is
> the Licensee will ensure that the copyright notice set out above appears prominently wherever the Software is reproduced, modified, transmitted or transferred in any form.

since currently there is no mechanism in `conda` which will display licensing information during installations. Even though I find it highly important to adhere to licensing conditions, I'd just live with not being too conforming for now..
If we want to fix this, we can either
* add a `post-link.sh` that somehow displays that information, or
* add that functionality to `conda` itself (which should be the desirable long-term goal).

mbargull:
@Takadonet:
> For the precompiled, you thinking of putting it straight into mykatlas then or still its own repo?

Good question! I'd like to get another opinion on this. Does anyone from @bioconda/core have an opinion on this? To summarize the issue at hand: `mykatlas` releases on PyPi bundle a precompiled binary `mccortex31` which stems from (a specific commit of) a fork of `mccortex`. Should we extract that binary into a separate package (which indicates it comes from a fork of `mccortex`) or should we bundle it just like in PyPi's release?
(Note: The inclusion of `mccortex31` in the PyPi release *could* be just a technicality since `mccortex` is not a Python package and hence not on PyPi.)

johanneskoester:
As long as there is no other use case for that binary, let's keep it simple and include it here.

mbargull:
Alright, if there won't be any other objections, I'm going to merge this in a bit.

mbargull:
Merged. Thanks for this, @Takadonet!

Takadonet:
No thank you!

Cashalow:
Hello @Takadonet @mbargull 
You appear to have added version 0.5.6 whereas the latest release on the github page is 0.3.6. When I downloaded the archive v0.5.6 the VERSION file states : 

```v0.5.6-0-gbd7923a-dirty```

I tried using the Mykrobe with some date of mine and the version from bioconda, I am running into a python error because some value are not defined : 

 ```
  File "/home/sacha/miniconda3/lib/python3.5/site-packages/mykatlas/typing/typer/variant.py", line 119, in _type_variant_probe_coverages
    if "LOW_GT_CONF" in self.filters and (confidence < self.confidence_threshold):
TypeError: unorderable types: int() < NoneType()
```

Have you tried this mykrobe version with your data ? Does it work ?

Takadonet:
hey @Cashalow 

The latest versions of mykrobe are being push to pypi and not github release page. The latest version on pypi is https://pypi.python.org/pypi/mykrobe/0.5.6 .

My end user finally got around testing with this datasets and is not working as well but with different error.

`  File "/share/apps/conda_env/mykrobe-v0.5.6.0/lib/python3.5/site-packages/mykatlas/cortex/mccortex.py", line 226, in coverages_cmd
    raise ValueError("Need either seq or ctx binary to run coverages")
ValueError: Need either seq or ctx binary to run coverages
`

Looks like ctx was not installed correctly on the path. 

What was your command that you got for your error?

Takadonet:
So looks like ctx was on the path but I was running the tool incorrectly. I have the same error as you now. Will take a look.

Takadonet:
So look like their is a bug with the latest version of mykatlas where the user MUST supply flag of **--min_variant_conf** when running **mykrobe predict**  The default of 'nothing' causing it to fail.  I have no idea what should be the min variant conf should be.

Takadonet:
I reported an issue with the author [here](https://github.com/iqbal-lab/Mykrobe-predictor/issues/143)

mbargull:
Hi @Cashalow, @Takadonet,
looks like right *after* the release of `mykatlas 0.6.1` on PyPi there was a commit that added that default value: https://github.com/Phelimb/atlas/commit/98c6545e4d2021069e431815a78c302dc0a342fd. (However skimming through the code I don't understand why the `None` value would be passed trough since it is replaced in https://github.com/Phelimb/atlas/blob/98c6545e4d2021069e431815a78c302dc0a342fd/mykatlas/cmds/genotype.py#L24-L25...)
@Takadonet: Does the program work fine if you use `--min_variant_conf 100`? If yes, we could either wait for the developer to issue another release on PyPi, or just use a patch from the aforementioned commit for a new build, I suppose.

Takadonet:
@mbargull The application works fine when any value is passed to _--min_variant_conf_

Cashalow:
Thanks for all your input ! I have just tried to run my analysis with --min_variant_conf defined and indeed it works well.

mbargull:
@Cashalow: I added the patch for `--min_variant_conf` to a new `mykatlas 0.6.1` build in #7017.
For the other bug, I guess it's best to ask the author, @Phelimb, if they would add a new release on PyPi. (Another possibility would be to build from the Git repo, but then someone might have to look into building `mykatlas`'s version of `mccortex` :wink:).

PR #6936, opened 2017-11-28
[FIX] add python 3 dep to discosnp recipe

cmonjeau:


recipes/discosnp/meta.yaml
```diff
@@ -8,19 +8,22 @@ source:
...
+  skip: True  # [not py3k]
+  number: 4
 
 requirements:
   build:
     - gcc   # [linux]
     - llvm  # [osx]
     - cmake
     - zlib {{CONDA_ZLIB}}*
+    - python
```
mbargull: I don't think `python` is needed as a build requirements, or am I wrong?

cmonjeau: You've right. There is a test during the install but the python requirement is not necessary.

mbargull:
We can also use the version constraint on `python` here, just like with https://github.com/bioconda/bioconda-recipes/pull/6940#issuecomment-347542429.

mbargull:
If you add that constraint, the `skip: True  # [not py3k]` can be removed, too.

cmonjeau:
On a previous PR, @bgruening told me not to pin python in the requirements and use the build section.

https://github.com/bioconda/bioconda-recipes/pull/6154#pullrequestreview-65695173

mbargull:
In the *vast* majority of cases (i.e., mostly Python pacakges) @bgruening's advice is the way to go. Here and for `hg-color`, however, we've got a special case that allows us to circumvent the creation of Python-specific builds.
But it's a good thing being critical about those advices -- for Python packages and `conda-build 2.x` those will even do [more harm than help](https://github.com/conda/conda-build/issues/1962).
Since it really is an uncommon/special case, think it would be best to also include a comment alongside that constraint, i.e.,
```yaml
requirements:
  run:
    # add version constraint to prevent Python-specific builds
    - python >2.6
```
(If you know that some scripts from `discosnp` don't work with Python 3.x, then the constraint should be something like `python <3.0a0` of course.)

@bgruening: Do you have any objections against that strategy?

bgruening:
No objections from my side. We should just be very careful with these kind of modification as long as we don't use conda-build 3.

mbargull:
> We should just be very careful with these kind of modification as long as we don't use conda-build 3.

Indeed! Thanks.

cmonjeau:
Thanks to @mbargull @bgruening and all people for these advices :)

cmonjeau:
and thanks @abretaud too 🥇

abretaud:
@cmonjeau ahah I did nothing here!

PR #6940, opened 2017-11-28
Fix hgcolor

cmonjeau:


mbargull:
You'll need to adjust the shebang line for the Python scripts in `build.sh`, e.g., by
```bash
sed -i.bak '1 s|^#!.*$|#!/usr/bin/env python|' bin/*.py
```

mbargull:
@bioconda/core: It looks like this wouldn't need to be pinned to a specific Python version since those scripts are not installed as Python modules/packages but only copied to `$PREFIX/bin` and look they are compatible with any Python version:
https://github.com/pierre-morisse/HG-CoLoR/blob/073add7bb0ba74f64c954f81338bcb208a2d72c1/bin/filterOutShortAlignments.py
https://github.com/pierre-morisse/HG-CoLoR/blob/073add7bb0ba74f64c954f81338bcb208a2d72c1/bin/filterShortReads.py
https://github.com/pierre-morisse/HG-CoLoR/blob/073add7bb0ba74f64c954f81338bcb208a2d72c1/bin/formatLongReads.py

But we don't have anything in place that could prevent the creation of Python-specific builds for such packages, right? (Note that `noarch: python` wouldn't make sense here, since this is no Python package but just requires some Python interpreter to be present.)

@msarahan: Quick question: I believe `conda-build 3.x` (for backwards compatibility) will still handle Python in a special way as to apply implicit pinning, right? Will there be any mechanism to prevent that behavior?

daler:
@mbargull right: we'll still get versions of the package for 2.7, 3.5, and 3.6.

msarahan:
I think you could do this a few different ways.  The one that comes to mind first is to have a run-only dependency on Python, preferably with a bounded constraint:

```
requirements:
  run:
    - python  >2.6
```

mbargull:
Got it. Thank you very much, @msarahan!

mbargull:
@cmonjeau: Can you please add a constraint like ` >2.6` to the `python` run requirements, just like @msarahan pointed out?

abretaud:
Green! Thanks!

PR #6951, opened 2017-11-29
travis: easy_install pip (trying xcode8.3 image)

mbargull:

@bioconda/osx: should we try to use [Travis' new default image](https://blog.travis-ci.com/2017-11-21-xcode8-3-default-image-announce) or should we just add `osx_image: xcode7.3` to `.travis.yml` for now?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ sudo: required
 services:
   - docker
 
+osx_image: xcode8.3
+
 before_install:
     - set -e
     - |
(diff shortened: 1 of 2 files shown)


johanneskoester:
I think we should stick with the old version as long as possible, in order to keep packages compatible with old osx. @daler, does that make sense (I don't know enough about macOS)?

mbargull:
xref: #6952

mbargull:
FYI: Apparently it was only a temporary issue that caused `pip` to not be available in Travis' new default  OSX image.
However, I suppose that was a "happy accident" that made us aware of the change and let us pin down the build environment a bit more via #6952 :smile:

PR #6957, opened 2017-11-29
Patch srst2 versions 0.1.6 and 0.2.0

Takadonet:


mbargull:
Do the changes from the patch alter the program's behavior for use cases which are not affected by https://github.com/katholt/srst2/issues/99? If they do alter the program's default behavior, I don't think we should add those changes here but rather wait for an upstream update.

As for the failed tests: Those are due to a current limitation in the testing framework concerning the use of single quotes in the test commands. If you replace the single quotes with double quotes, it should work fine.

Takadonet:
Fixed failed tests for version 0.1.4.6 . 

They don't alter the end results but just address the issue where file path cannot have '.' character. I am for sure willing to remove them from this branch for the latest version of the tool.  It will limit our ability to update our wrapper since we don't have control on when the author will accept and publish a new release.

Takadonet:
@mbargull Removed patch and hope for srst2 to be updated with bug fix.

mbargull:
Do you want to reinstate the patch for the current version or should we merge without it? After having a closer look at the patch, I think it is a justifiable addition/fix.

Takadonet:
I will reinstate the patch tomorrow. Working from home with a sick kid at the moment.

mbargull:
No worries, I pushed the necessary changes. Every get well soon!

mbargull:
@ericenns: In #3767 you added the recipe for `srst2 0.1.4.6`. But the software's actual version seems to be `0.1.6`. Was that just a typo?

ericenns:
@mbargull yes that is a typo. I believe I introduced it when we were updating our srst2 galaxy tool to use conda and I used the version we had in there which was a typo.

ericenns:
@mbargull I guess I should clarify the typo is the 0.1.4.6 which should have been 0.1.6

mbargull:
Thanks for clarification, @ericenns. I just added a commit to this PR to fix this.

PR #6973, opened 2017-11-30
Haystack

rfarouni:


recipes/haystack_bio/meta.yaml
```diff
@@ -1,6 +1,7 @@
 {% set version = "0.5.2" %}
 
-{% set sha256 = "1275a8e5f098ac859d8c71b7fb237e99d87158f9e0238f53650cd3959463681b" %}
+{% set sha256 = "17197f3185fe22917d4aff0dda139a538cefb091f3cf1086550e83c4f2c7c57e" %}
```
bgruening: Why does the checksum change if the version stays the same? Please do not force-push tags and make old versions non-reproducible.

rfarouni: An upgrade to a dependent package caused an error. The code  change to fix it was minimal and I was not sure if a version upgrade was warranted. Also, I wasn't sure  if it is possible to have the previous version removed so user won't accidentally install it. If I upgrade to 0.5.3, can users still install 0.5.2?

bgruening: Yes users can install 0.5.2 even if 0.5.3 is available.

rfarouni: And that is the situation that I am trying to avoid.  Although 0.5.2 worked fine a while ago, a change in a dependency package broke it. What course of action would adhere to best practices in this situation? Having users download broken software is not a good idea I think?

PR #6976, opened 2017-11-30
Add tombo

druvus:

### DIFF
--- a/recipes/tombo/build.sh
+++ b/recipes/tombo/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
You were quite fast in building this recipe :), only 1 day since public release...

bgruening:
@druvus it seems you have forgotten to `git add` the build.sh file?

druvus:
@bgruening You're right... did it too late.

marcus1487:
Wow incredibly fast indeed! I did not see this before adding my version of tombo recipe (#7007). One difference with the recipe I have added it to name the recipe ont-tombo in order to indicate that this should be an Oxford Nanopore Technologies supported recipe (as with ont-fast5-api). I would like to request that the ont-tombo be used before much larger distribution is adopted. Let me know if there are any issues with this request at this point. Thanks!

druvus:
I added this when ONT made the tombo repo public at github. Later ONT also released it at pypi were they added the prefix. I am fine to remove this and instead use ont-tombo. Any opinions @bioconda/core

PR #6978, opened 2017-12-01
added watchdog WMS recipe

klugem:

This is my first recipe for bioconda. If I can improve something, please let me know.
Watchdog: https://github.com/klugem/watchdog


klugem:
I have no idea why Macos build fails...Can anyone give me a hint ?

[https://travis-ci.org/bioconda/bioconda-recipes/jobs/309948558](url)

The only error I can find is this one:

> Users/travis/.travis/job_stages: line 57: shell_session_update: command not found

recipes/watchdog-wms/build.sh
```diff
@@ -0,0 +1,32 @@
...
+mkdir -p "${OUT}"
+mkdir -p "${PREFIX}/bin"
+cp -R * "${OUT}/"
+
+# rename the sh scripts
+mv "${OUT}/watchdog.sh" "${OUT}/watchdog-cmd.sh"
+mv "${OUT}/workflowDesigner.sh" "${OUT}/watchdog-gui.sh"
+
+# modify path of included files
+sed -i "s|\$SCRIPT_FOLDER/|\$SCRIPT_FOLDER/../share/${VERSION}/|g" "${OUT}/watchdog-cmd.sh"
```
mbargull: You'd have to triple-escape `\\\$SCRIPT_FOLDER` since then `bash` will pass on that part of the double quoted string as `\$SCRIPT_FOLDER` to `sed` which also needs the `$` to be escaped.

klugem: Thanks for the hint! However, there was another problem with the macos version of sed as the -i parameter espects another argument.

mbargull: Ah, right, `sed` behaves differently/incompatible for inline modifications on Linux and OSX (i.e., the "OSX-style" `sed -i ''` won't work on Linux since there an additional argument is not valid :/). The only way to directly use it in a manner that works under both systems is to use something like `sed -i.bak` (i.e., create a backup file).

mbargull: I really should have noticed that before! Sorry 'bout that.

klugem: No problem. Now it is passing the tests. :)

klugem:
Any more comments on this one ? If not, can it be merged ? :-)

recipes/watchdog-wms/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  build:
+      - openjdk >=8
+      - coreutils >=8
+  run:
+      - openjdk >=8
+      - coreutils >=8
+
+test:
+  files:
+      - workflow.test.xml
```
johanneskoester: This is not a valid test case. Sorry for being late on this. Can you please add a test case that runs the software. It is sufficient to call it with --help or so. We just need to see if it executes. Also, the `files` is not supported by our testing framework.

klugem: There is a test script (run_test.sh) that calls the binary with
watchdog-cmd --help 
and 
watchdog-cmd -x "workflow.test.xml".

klugem: Is there an additional testing framework apart from the travis CI tests ?

bgruening: can you add `watchdog-cmd --help` and Co just as test/commands and add all other more elaborated test directly into the build.sh script for the moment.

I think this would help a lot.
Thanks.

klugem: Sure, I can do that. Can you explain me what's wrong with the run_test.sh file I wrote ?

bgruening: There is nothing "wrong" with it per se. Its just that we are also testing your recipes in a busybox based Docker container to make sure you have defined all requirements. In this Docker container we don't have Conda and not test-environment. Which makes it hard to test run_test.sh files.

klugem: Thanks for the explanation. I will make the proposed changes.

klugem: The changes are made are in #7001. Any comments on this one ? :)

PR #6980, opened 2017-12-01
busco: remove wget requirement, update post-link script

mbargull:

@daler: In #5846 you
* added `wget` as a run requirement. I believe it is only used for the test though, correct?
* reinstated the Python script for the post-link step. I think we should rather stick with `sed` to do this since both of them do the same and we should only use basic tools here as [advised](https://conda.io/docs/user-guide/tasks/build-packages/link-scripts.html).


mbargull:
> the final answer is "whatever makes the package not broken"

😨 "broken" "package" 😰...
too much discussion on that topic lately 😫...
make that pain go away, please 😆

PR #6982, opened 2017-12-01
DISCASM: Adding bzip2 to recipe to fix pysam linking problem (https:/…

tstrnity:
…/github.com/ContinuumIO/anaconda-issues/issues/2441).

### DIFF
--- a/recipes/discasm/meta.yaml
+++ b/recipes/discasm/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "discasm" %}
 {% set version = "0.1.2" %}
-{% set build = "2" %}
+{% set build = "3" %}
 
 package:
   name: {{ name }}
@@ -19,6 +19,9 @@ build:
 
 requirements:
   build:
+    - python
+    - pysam >=0.10.0
+    - bzip2
   run:
     - python
     - pysam >=0.10.0


bgruening:
@tstrnity does this also happen with the conda-forge bzip2 package. I can not reproduce the problem afaik.
Do you have a special case when this happens?

tstrnity:
I am not sure. I am trying to use DISCASM as a galaxy tool. When the galaxy tool is loaded with the previous recipe, it happened. So I tried creating an empty anaconda environment and installed discasm, and it happened in that situation as well. When I removed the automatically loaded bzip2 and installed the bzip2 from bioconda, then all works fine. The error occurs within pysam. When a component is used that needs something from bzip2, it tries to load the .so library, but the default bzip2 package only has the static .a library, so the program fails.

tstrnity:
I investigated, and it does seem like the one in the conda-forge channel is the one that works. The one in the default channel has a later build number than the one in the conda-forge channel. I do not know why placing the requirement within the DISCASM recipe causes the one in the conda-forge channel to get loaded, whereas leaving it to pysam to load the dependency causes the one from the default channel to be loaded.

tstrnity:
Well, now I am doing another test with a new conda environment and it seems like the conda-forge bzip2 is being loaded, so not sure why it wasn't doing that for me the other day... It is possible that there was something strange with that conda environment or its configuration that I am not repeating today. So let's put a hold on this pull request until I can investigate the Galaxy tool a bit more to see if there is something in its environment that is not right.

tstrnity:
OK. So I created a new galaxy from scratch to test the tool and it works fine, so I am closing this pull request as unnecessary.

tstrnity:
Just fyi, I think the situation can be recreated if one has a prior environment that had loaded the default channel's bzip2 into the conda's pkgs directory. Conda seems to just grab that local one, as it appears to have a larger build number, rather than downloading the one with a lower build number from the conda-forge channel.

PR #6983, opened 2017-12-01
Recipe for scan_indel

hellothisisMatt:

### DIFF
--- a/recipes/scanindel/meta.yaml
+++ b/recipes/scanindel/meta.yaml
@@ -1,10 +1,10 @@
-{% set name = "ScanIndel" %}
+{% set name = "scanindel" %}
 {% set version = "1.3" %}
 {% set hash_type = "md5" %}
 {% set hash_value = "d51dc67c19763e1f5630020a261f4afc" %}
 
 package:
-  name: ScanIndel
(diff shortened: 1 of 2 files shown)


recipes/scanindel/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  build:
+  run:
+    - scipy           
+    - numpy
+    - bedtools ==2.17.0
+    - samtools <=1.0
+    - bwa ==0.7.12
+    - blat
+    - freebayes ==0.9.21.7
+    - python ==2.7.13
```
bgruening: The python pinning is not need as you skip the build in the build section.
Are the other strict pinnings really needed? We usually don't do this and have them more relaxed.

hellothisisMatt: Thanks for catching that. I'll remove the python pinning. I believe the other strict pinning is needed as the tool mentions that it requires certain versions of the listen dependencies.

hellothisisMatt: It's a bit of an old tool so I'm unsure if it can function on the newer versions of the dependencies

Issue #6809, opened 2017-11-15
Travis-CI complains "Build config did not create any jobs" since commit 5a76a5


fgp:
Since I pushed commit 5a76a5 to my fork's master branch, pushing to my recipe branch no longer triggers a travis-ci build. Travis-CI lists the commit under "Requests", but reports that ""Build config did not create any jobs". I suspect the problem is the "fork = true" condition -- could it be that Travis-CI somehow doesn't know that my repository is a fork of the main bioconda-recipe repo? Do I have to tell Travis-CI that somehow? The "Contributing" section of the manual does not seem to specify any special steps one must take to configure that, and the Travis-CI documentation (https://docs.travis-ci.com/user/conditional-builds-stages-jobs/) seems to indicate that "fork" is a built-in variable (I guess Travis-CI queries it from github somehow).

daler:
@fgp we've been making some improvements to the build system in the past day or so. Things are still settling down, and it may be that we're missing something here. The latest fix we're working on is https://github.com/bioconda/bioconda-recipes/pull/6811. Once that's merged into master and you pull those changes to your fork, would you mind trying again? If it still doesn't work then we'll work on a fix.

In the meantime, can you post a link to the travis-ci build you're referring to? I don't see any bioconda-recipes forks under your account.

fgp:
I just pulled the latest master and pushed that master plus my branch rebased on top of it (named r-akima) to https://github.com/Cibiv/bioconda-recipes. It seems that the problem persists - still nothing scheduled on Travis-CS (well, the master push trigger a build, but I cancelled that. The build I want is the one of my r-akima branch)

daler:
I tested it on my fork as well and I'm also not seeing the branch tests. I'll play around with it some more. The logic looks correct in the job selection. 

You're right, travis-ci has the existing fork variable. We're testing out these beta features: 

- https://docs.travis-ci.com/user/build-stages/
- https://docs.travis-ci.com/user/conditional-builds-stages-jobs/

Also, if this is in preparation for an `r-akima` PR (as implied by your branch name) would you please submit it to conda-forge instead since it's not a bio-specific package?

fgp:
[...]

PR #6991, opened 2017-12-04
Samtools 0.1.12 adding misc. tools

scholtalbers:

Add the misc. tools `maq2sam-short`, `maq2sam-long`, `wgsim` so they can be used with this samtools. In particular I need these for an old galaxy tool (BAMedit).

Creating a WIP PR instead of developing/testing on my own branch due to #6809


bgruening:
Can you please include tests for these tools @scholtalbers?

scholtalbers:
@bgruening added the tests and this works with `./simulate-travis.py --docker`. But fails here - same for #6992 . Do you have any tip to fix this?

recipes/samtools/0.1.12/meta.yaml
```diff
@@ -24,4 +24,8 @@ test:
   files:
     - ex1.fa
   commands:
-    - "samtools faidx ex1.fa"
+    - samtools faidx ex1.fa
+    - "maq2sam-short 2>&1 | grep '^Version: r439'"
+    - "maq2sam-long 2>&1 | grep '^Version: r439'"
+    - "wgsim 2>&1 | grep '^Version: 0.3.2'"
```
mbargull: typo in `wgsim` version: `0.2.3`
Also it might be best to avoid single quotes in the strings' contents for the moments, use instead:
```yaml
    - 'maq2sam-short 2>&1 | grep "^Version: r439"'
    - 'maq2sam-long 2>&1 | grep "^Version: r439"'
    - 'wgsim 2>&1 | grep "^Version: 0\.2\.3"'
```

mbargull:
The `mulled` tests fail due to the `ex1.fa` test file.
@bgruening: Should we maybe just move the two commands that use the test file to a `run_test.sh` to "hide" it from `mulled` (for now, until we can support `mulled`'s `--test-files`)?
If so, we should also add `samtools --help` or `samtools --version` or whatever works to `test/commands`.

Additionally, `samtools faidx ex1.fa` didn't seem to abort the tests: https://travis-ci.org/bioconda/bioconda-recipes/jobs/312603100#L848. Hence that test case must be modified to actually catch a failure.

recipes/samtools/0.1.12/meta.yaml
```diff
@@ -22,8 +22,6 @@ source:
   url: https://downloads.sourceforge.net/project/samtools/samtools/0.1.12/samtools-0.1.12a.tar.bz2
 test:
   commands:
-    - '(printf ">seq1\nCACTAGTGGCTCATTGTAAATGTGTGGTTTAACTCGTCCATGGCCCAGCATTAGGGAGCTGTGGACCCTGCAGCCTGGCTGTGGGGGCCGCAGTGGCTGAGGGGTGCAGAGCCGAGTCAC\n>seq2\nTTCAAATGAACTTCTGTAATTGAAAAATTCATTTAAGAAATTACAAAATATAGTTGAAAGCTCTAACAATAGACTAAACCAAGCAGAAGAAAGAGGTTCAGAACTTGAAGACAAGTCTCT\n" > ex1.fa && samtools faidx ex1.fa || exit 1)'
```
mbargull: Please also add 
```yaml
    - 'samtools 2>&1 | grep "^Version: 0\.1\.12a"'
```

recipes/samtools/0.1.12/run_test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+# create ex1.fa
+printf ">seq1\nCACTAGTGGCTCATTGTAAATGTGTGGTTTAACTCGTCCATGGCCCAGCATTAGGGAGCTGTGGACCCTGCAGCCTGGCTGTGGGGGCCGCAGTGGCTGAGGGGTGCAGAGCCGAGTCAC\n>seq2\nTTCAAATGAACTTCTGTAATTGAAAAATTCATTTAAGAAATTACAAAATATAGTTGAAAGCTCTAACAATAGACTAAACCAAGCAGAAGAAAGAGGTTCAGAACTTGAAGACAAGTCTCT\n" > ex1.fa
+
+# samtools can fail with exit 255 which is not picked up as an error
+samtools faidx ex1.fa || exit 1
```
mbargull: Looking at the code, it seems like `samtools faidx` will actually exit with status code 0 in this case 😕...
But it behaves correctly if we use additional arguments, hence use this instead:
```bash
samtools faidx ex1.fa seq1
```

bgruening:
@mbargull @scholtalbers I would keep the tests as self-contained as possible. So if printf works super, lets keep it. The reason is that we want to test as many build artefacts as we can, e.g. Singularity containers. Mounting always test-files in etc is cumbersome and error-prone. Having everything in one line is easy.

That that, if it is not possible otherwise we should use test-files.

bgruening:
> @bgruening: Are you ok with the run_test.sh workaround? Both affected commands are also tested in test/commands so we don't loose anything there.

Yes, I'm ok with merging it. But in general would keep test/files to a minimum. Then better ship them with the package.

scholtalbers:
@mbargull & @bgruening thx for the feedback! For a next package or updating this one I will stick with the printf method in a `./run_test.sh` 🤪

PR #7004, opened 2017-12-05
Feature/perl moose

jerowe:

I am working on moving the perl-moose* modules to conda-forge, but I am trying to get through some lower level package first. It is also a dependency of BioPerl, which is on the todo list.


jerowe:
@bioconda/core @daler, perl builds are not with bioconda-utils. They work with conda build. Any ideas why?

mbargull:
Hi @jerowe, there's currently an issue with parts of the test framework when the test commands contain single quotes. If you replace them with double quotes, the tests might work.
(BTW: your commit message never cease to crack me up 😆)

jerowe:
Thanks @mbargull . ;-) Gotta keep entertained while building the software . I will try out the double quotes.

PR #7007, opened 2017-12-05
Ont tombo

marcus1487:


marcus1487:
Tombo is a package for the analysis of raw nanopore sequencing data and detection of modified bases. This package is a derivative work of nanoraw (currently on bioconda), so seems to fit quite well into the bioconda channel.

marcus1487:
As noted in #6976 I think the best course moving forward would be to merge this request and remove the previous tombo recipe in order to have consistent naming between pypi and conda. Shall I include the removal of the tombo recipe in this pull request or is should that be handled within the other tombo pull request? I am not sure how the removal of a recipe would effect the bioconda system, so please advise on how to proceed. Thanks!

bgruening:
@marcus1487 I'm fine making this a separate PR. Removing the recipe is not enough we also need to remove/migrate the package from the channel. I can do this if the recipe is remove from the channel.

PR #7008, opened 2017-12-06
quast 4.6.1

alexeigurevich:
Fixed the GeneMark installation issue preventing creation of bioconda recipe for v.4.6.0. No functional changes since 4.6.0.


alexeigurevich:
Hi, Björn @bgruening! Could you help with this? I cannot figure out where the problem is. I see in the end of the log:
`2017-12-06 16:36:20`
`Running Contig analyzer... `
`Compiling gnuplot... `
`NOTICE: Failed to compile gnuplot (/opt/conda/conda-bld/quast_1512577423627/_t_env/lib/python2.7/site-packages/quast_libs/gnuplot)! Try to compile it manually. You can restart Quast with the --debug flag to see the compilation command. `
`  1  contigs_1`
`  1  Logging to files /opt/conda/conda-bld/quast_1512577423627/test_tmp/quast_test_output/contigs_reports/contigs_report_contigs_1.stdout and contigs_report_contigs_1.stderr...`
`  1  Aligning contigs to the reference`

followed immediately by

`16:37:40 BIOCONDA ERROR BUILD FAILED recipes/quast, CONDA_BOOST=1.64;CONDA_BZIP2=1.0;CONDA_GMP=5.1;CONDA_GSL=1.16;CONDA_HDF5=1.8.17;CONDA_HTSLIB=1.6;CONDA_NCURSES=5.9;CONDA_NPY=112;CONDA_PERL=5.22.0;CONDA_PY=27;CONDA_R=3.4.1;CONDA_XZ=5.2;CONDA_ZLIB=1.2.8;MACOSX_DEPLOYMENT_TARGET=10.9
16:37:41 BIOCONDA ERROR BUILD SUMMARY: of 1 recipes, 1 failed and 0 were skipped. Details of recipes and environments follow.
16:37:41 BIOCONDA ERROR BUILD SUMMARY: FAILED recipe quast-4.6.1-boost1.64_2.tar.bz2, environment CONDA_BOOST=1.64;CONDA_NPY=112;CONDA_BZIP2=1.0;CONDA_PERL=5.22.0;CONDA_HDF5=1.8.17;CONDA_HTSLIB=1.6;CONDA_GMP=5.1;CONDA_R=3.4.1;CONDA_GSL=1.16;MACOSX_DEPLOYMENT_TARGET=10.9;CONDA_XZ=5.2;CONDA_NCURSES=5.9;CONDA_ZLIB=1.2.8;CONDA_PY=27`

But I do not see what the error is! Note that the message about failed gnuplot compilation is written to STDOUT and there is no exit code afterwards (it is just a notice -- not crucial module for Quast). Is it possible to look at `/opt/conda/conda-bld/quast_1512577423627/test_tmp/quast_test_output/contigs_reports/contigs_report_contigs_1.stderr` somehow? Or ask bioconda to give a traceback or at least exit code. 
Thanks!

recipes/quast/meta.yaml
```diff
@@ -1,13 +1,13 @@
...
+{% set version = "4.6.1" %}
+{% set hash = "29118bff957af6d1b36136735450ddc8" %}
 
 package:
   name: {{ name|lower }}
   version: {{ version }}
 
 build:
-  number: 1
+  number: 2
```
bgruening: This should start from 0 if you increase the version number.

recipes/quast/meta.yaml
```diff
@@ -49,11 +49,13 @@ requirements:
     - sambamba
 
 test:
+  source_files:
```
bgruening: Is test_data really needed?

bgruening:
@alexeigurevich you could try to `cat ./quast_test_output/contigs_reports/contigs_report_contigs_1.stderr` directly after you test command. Concat both with `;`.

Also if you run `simulate_travis.sh` you can dive into the container and look at it more closely.
Hope that helps!

recipes/quast/meta.yaml
alexeigurevich: Maybe I understand build-number policy incorrectly. Initially I set it to 0 when I made the first commit changing Quast version from 4.5 to 4.6.1. After that, I got failed tests and made several changes in the recipe only (not in Quast v.4.6.1). I thought that I need to change build-number after such changes (I did this twice in this patch, so build number became 2).

recipes/quast/meta.yaml
alexeigurevich: It is needed for running quast --test and metaquast--test executed below. However, it is not needed for installation itself, so it is not installed by default. So, here I ask conda to copy the data to temporary dir before testing (if I understand the syntax correctly).

alexeigurevich:
Hi, @bgruening ! Sorry for bothering again. I tried both suggestions but the problem is still unclear. 
1. I tried to debug locally, but `./simulate-travis.py --git-range HEAD --packages quast` ends with `BUILD SUCCESS` and no errors.
2. I tried to add `cat ./quast_test_output/contigs_reports/contigs_report_contigs_1.stderr` to the test command (see my commit above) but it looks like it is not working for some reason. In the travis log I see `+ quast --test --force-nucmer` without mentioning `cat` and the run of this command ended the same as before (actually it is expected since the command is the same): 

`1 Aligning contigs to the reference`

followed immediately by

`16:37:40 BIOCONDA ERROR BUILD FAILED recipes/quast, `

recipes/quast/meta.yaml
mbargull: You nearly got it right, with the exception that the build numbers are only "manifested" for builds that are released to the Conda channel, i.e., (usually) builds that happened on the `master` branch. That means that for additional changes you make before your PR is merged, you won't have to increase the number as no builds were actually uploaded. Hence `quast 4.6.1` is always at build number `0` for this PR.

recipes/quast/build.sh
```diff
@@ -16,7 +16,7 @@ QUAST_HOME=$PREFIX/opt/quast-$PKG_VERSION
 mkdir -p $BINARY_HOME
 mkdir -p $QUAST_HOME
 
-python "setup.py" install_full
+python "setup.py" install
```
mbargull: Beware that `quast` behaves very "non-standard-ish": As is, you need its custom `install_full` argument here to instruct it to build its dependencies. The build failure you got earlier is likely because with the change to `install` it now tries to compile `gnuplot` etc. during its run time but then misses the required compilers.
If you'd want `quast` to behave more conventional (which is likely the preferable long-term solution), you would have to patch it to avoid the custom compilation steps, i.e., patch/look at `setup.py` and *all* files that use `LIBS_LOCATION` from `quast_libs/qconfig.py`..

recipes/quast/meta.yaml
alexeigurevich: Ok, got it, thanks!

recipes/quast/build.sh
alexeigurevich: Hi @mbargull, I've changed `install_full` to `install` here just for the sake of speed. Note that both options process `gnuplot` compilation equal likely, the difference between them is in few other modules (stuff for structural variations and without reference mode of MetaQuast -- i.e. relatively rarely needed). I plan to return `install_full` back when we solve the main issue with the build failure. Note that the error is absolutely the same and I initially faced it after my first commits in this patch where I used `install_full` as before.

As for the general solution -- I will try to patch Quast accordingly.
However, I still do not understand what is going wrong with the build right now. Quast v.4.6.1 is pretty similar to v.4.5 which was successfully added to bioconda. All differences are mostly cosmetics and should not affect compilation/installation at all..

mbargull: Oh, wow, my bad about the `install`/`install_full` thing! I must've read something wrong while skimming `setup.py`, sorry.
But it looks like the `gnuplot` directory isn't added to `package_data`. Maybe adding a `find_package_files('gnuplot')` in there will fix the current `gnuplot` problem.

alexeigurevich: Well, it was our fault, I've added a patch to include `find_package_files('gnuplot')` in `setup.py`. Now, Quast does not try to recompile gnuplot in the test run as it did previously, which is good.
However, it has not solved the main issue. Now I don't have any more clues why the test is failing :)  Or why `; cat ....stderr` in `meta.yaml` is not working.

mbargull: > Well, it was our fault

Ah, I didn't even notice before, I was talking to one of the tool's devs!

To look into what failed during the `gnuplot` compilation, you'd have to inspect `quast_libs/gnuplot/make.err`, I'd guess from https://github.com/ablab/quast/blob/quast_4.6.1/quast_libs/ca_utils/misc.py#L132.
Also, I would suggest to let the build fail if some compilations failed, e.g., at https://github.com/ablab/quast/blob/quast_4.6.1/setup.py#L271 insert via a patch something like
```python
if not_installed_message:
    raise RuntimeError(not_installed_message)
```
since we really want to avoid client-side compilation and get a readily build package from `conda-build`.
[Either way, if you want to integrate Quast properly into the Conda/Bioconda ecosystem (and avoid those `gnuplot` build issues :wink: overall), the best thing would be to avoid/patch out all custom compilation steps and let `conda` install all dependencies]

alexeigurevich: Well, I committed one more patch, so we now reuse system `gnuplot` if it is available. So, no there is no warnings/notices/etc about any problems with it. However, the test is still failing!

mbargull: Maybe the test went fine and just didn't create the `.stderr` file you are trying to read and thus `cat` fails? Just a hunch.
But it's great that an external `gnuplot` can now be used, thanks! Do you think it would be possible to do similar patches for some of the other tools? Looking inside `quast_libs`, I see a couple of programs that are available on `bioconda`, e.g., `bwa`, `bedtools`, `glimmer`, `gnuplot`, `mummer`, `sambamba`.

alexeigurevich: Well, initially I did not use `cat` but the test was crashing and I was curios what is happening. The glue is probably in this .stderr file, so Björn (@bgruening) suggested me to add `cat`. However, it is not printed anyway :( 
Note: this files is always created until something extraordinary has happened. But again, it is not the source of problem because it is the same as before adding `cat`.

alexeigurevich:
Hi, @bgruening and @mbargull! We definitely need your assistance to finalize this pull request since the Testing system behaves very strange and we do not understand what is going on. 

After almost a week of trial-and-error commits we get close to the solution. In particular we finally were able to read .stderr and .stdout files using few patches. There were no errors but at least we found a particular point where conda stops the execution and reports failed test message. It looks like something is wrong during creating of gnuplot plots in the pipeline (even when we use system `gnuplot` and there is no warnings about its incorrect installation). Finally, it is not a big deal, since we have a lot of other plots and we anyway plan to remove gnuplot drawings in the next major release. So, it is pretty easy to disable gnuplot using ad hoc patch and we wanted to do it to finally add v.4.6.1 to the recipes. BUT something strange is happening with the testing system even without applying the patch! We demonstrate it by using simple `--no-plots` option in the test commands. Note that this option disables ALL plots generation including gnuplot ones, so should lead to passed tests.

Let's compare two build logs before and after my last commit (`53d25a8`) in this thread. 
(1) Before: https://api.travis-ci.org/v3/job/319249822/log.txt
(2) After: https://api.travis-ci.org/v3/job/319265071/log.txt

In (1) we added `--no-plots`  to our first test command (`quast --test --force-nucmer --no-plots`). It finally passed correctly but the entire test is still failed due to the second test command (`metaquast --test --force-nucmer`) because it has enabled plots and crashed similarly as before during gnuplot processing.
In (2) we added `--no-plots` to the second command (`metaquast --test --force-nucmer --no-plots`) and expected to see the entire test passed. Note that this is the only change in the last commit. However, as you can see in the log, a completely different problem occurs. This time conda does not provide `test_data` folder to Quast for some reason and it fails because of this! Note that we explicitly ask about this folder in `meta.yaml` (see `test` section) and that this folder was provided in (1).

Could you explain what is going on and how should we proceed? Note that this `test_data` presence/absence issue appears spontaneously for us and we also get it when tried to apply gnuplot-disabling patch instead of `--no-plots` solution.
Thanks!

mbargull:
Hi @alexeigurevich,
the error you see in the latest build attempt is because of a certain feature of Bioconda's build system. The tests are run a second time but inside a minimal busybox container, but that additional test run does not support test files, test requirements, and test scripts. Thus `quast` is not supplied with the appropriate test files. I think (but am not certain of it) that `wget` is available in busybox, so maybe you can try to add
```yaml
    - wget quast.sf.net/test_data.tar.gz && tar xzf test_data.tar.gz  # [linux]
```
to the test commands. IF that does not work, as a workaround you could move those tests, i.e.,
```bash
quast --test --force-nucmer --no-plots
metaquast --test --force-nucmer --no-plots
```
to a `run_test.sh`. But the other tests, i.e.,
```yaml
     - quast --version
     - metaquast --version
```
must remain inside `meta.yaml` so that they can be tested inside the minimal system.

Regarding the failures for the `*quast --test --force-nucmer` commands without `--no-plots`, sadly I have no idea what went wrong, sorry.

alexeigurevich:
Hi @mbargull, thanks for the reply!
Your explanation makes sense and I will try to use your kind suggestions (I think the one with `run_test.sh` is more appropriate). However, I don't understand why the last but one build worked correctly for `quast --test --force-nucmer --no-plots` if there should not be any test data for it.

mbargull:
The tests run in succession, such that first the tests are run on the build system and if those succeed, then the container-based test is run. So for the last but one build *on the build system*, which has access to the test files, `quast --test --force-nucmer --no-plots` ran fine but `metaquast --test --force-nucmer` failed and as such the container-based test wasn't run at all.
In comparison to that, for the last build, all first tests (`quast` and `metaquast` alike) were successful such that `quast --test --force-nucmer --no-plots` was run in the container for the first time and failed there.

alexeigurevich:
Ahh, now it is clear, thanks for the explanation! 
One final question, is it fine to have a bunch of trial-and-error commits in pull request like we have here? Or I can close this request and start a new one with a single working commit updating meta.yaml to v.4.6.1.

mbargull:
If you want to clean up your PR, you can always force push to your branch -- it's your branch after all. So no need to open up another PR.
But you don't really have to do this since pull requests are always "squashed and merged" for this repository, so that only one commit on `master` will be created.
As for the `gnuplot`/`--no-plots` issue: You could build the recipe locally (with `--no-plots` tests to build it successfully) and run/debug the tests locally, e.g., with
```bash
conda create -n quast-plot-debug --use-local quast && \
activate quast-plot-debug && \
wget quast.sf.net/test_data.tar.gz && tar xzf test_data.tar.gz && \
quast --test --force-nucmer
```
if you want.

alexeigurevich:
Finally we get the checks passed! I will clean up everything and let you know when this pull request is ready for merging

mbargull:
Just an FYI: In my `bioconda-recipes` fork I let `quast` build on OSX as well a couple of days ago and interestingly there the tests without `--no-plots` succeeded (https://travis-ci.org/mbargull/bioconda-recipes/builds/319958427). (But I still couldn't see what went wrong on Linux.)

alexeigurevich:
Hi, guys! Finally we are done -- please revise and merge if there is no any comments. Everything is working except `gnuplot` drawings but it is not critical and anyway will be removed in the next major release. @mbargull note that you requested few changes which are now satisfied.

recipes/quast/gnuplot_issue.patch
```diff
@@ -0,0 +1,61 @@
+From ffc589098223e021f604ae033cbd2574a1de58f3 Mon Sep 17 00:00:00 2001
+From: Alexey Gurevich <alexeigurevich@gmail.com>
+Date: Mon, 25 Dec 2017 15:22:42 +0300
+Subject: [PATCH] Temp: disable gnuplot compilation and run
+
+---
+ quast_libs/ca_utils/align_contigs.py | 6 +++---
```
mbargull: In https://github.com/ablab/quast/blob/quast_4.6.1/quast_libs/contigs_analyzer.py#L260-L261
```diff
-    if qconfig.draw_plots:
-    compile_gnuplot(logger, only_clean=False)
```
should be patched too, I guess.

recipes/quast/meta.yaml
```diff
@@ -1,19 +1,21 @@
...
-  number: 1
+  number: 0 
   string: boost{{CONDA_BOOST}}_{{PKG_BUILDNUM}}
 
 source:
   fn: {{ name|lower }}-{{ version }}.tar.gz
   url: http://downloads.sourceforge.net/project/{{ name|lower }}/{{ name|lower }}-{{ version }}.tar.gz
   md5: {{ hash }}
+  patches:
+    - gnuplot_issue.patch
```
mbargull: Maybe use a filename that reflects the behavioral change, e.g., `disable-mummer-plots.patch`, or be more verbose, e.g., `gnuplot-issue_disable-mummer-plots.patch`.
To make the change in behavior more prominent, you can also add something along the lines of
```yaml
extra:
  notes: Drawing MUMmer plots is disabled due to issues with gnuplot.
```
at the end of `meta.yaml`.

recipes/quast/gnuplot-issue_disable-mummer-plots.patch
```diff
@@ -0,0 +1,26 @@
+From 02231ef13c093017f329abd6527da1474b5e670a Mon Sep 17 00:00:00 2001
```
mbargull: I think you renamed the wrong file :wink:

alexeigurevich: yep, thanks! One more try to go :)

alexeigurevich:
Hmm, some strange things have happened again. MacOS build failed with GeneMark fail, however, the last changes did not affect this at all. Moreover, I saw exactly the same issue 3 days ago -- we fixed all tests (passed tests at `e94327a`), than I added simple clean up (`3821d09` .. `462ff1a`) and got this strange MacOS-GeneMark error. I reverted everything back (`290bd06`) and got passed tests again. After that, I added pretty the same clean up commits one by one (`dea7b4f` .. `a3ca05d`) and everything became OK again. Later, I added tiny changes after @mbargull's code review (`cd6bc29` & `8ad9574`) and got failed build again :(
Trying to find any meaningful difference between working `a3ca05d` and failing `8ad9574`..

mbargull:
How strange, indeed... As you said, your last changes shouldn't have affected the GeneMark usage -- maybe there is just some race-condition (maybe during the parallel execution at https://github.com/ablab/quast/blob/quast_4.6.1/quast_libs/genemark.py#L307?) or another non-deterministic bug in QUAST, GeneMark, or something?!? As I've been curious, I restarted that failed OSX build from the previous commit -- and it miraculously passed xD. So I'd say it is best to revert the last commit and reset to https://github.com/bioconda/bioconda-recipes/pull/7008/commits/8ad95749dc973e2df17858be849bd46087b1a1f0. Additionally (under the assumption/hunch that parallel GeneMark runs might be the problem), if possible, add `--memory-efficient` to the test commands in `run_tests.sh` to avoid the parallel execution in https://github.com/ablab/quast/blob/quast_4.6.1/quast_libs/genemark.py#L307.

alexeigurevich:
Well, we've never seen a race condition in GeneMark execution (or any other parallel unit of Quast, I feel `joblib` is a pretty good lib for parallel execution). But who knows, maybe it occurred here, anyway we do not have any better ideas. 
I added `--threads 1` to the tests to avoid parallel execution. It is more straightforward than `--memory-efficient` ;) Also, I reverted the previous commit, as you suggested. Hopefully everything will be fine with tests now!

alexeigurevich:
Hm, we have macOS-GeneMark issue again despite the switch to single-thread execution!
Maybe you can rerun the failed test again and if it turns to passed you can finally accept this request :) ?

bgruening:
@alexeigurevich I have restarted the test.

alexeigurevich:
@bgruening are there any news on the rerun result?

mbargull:
Looks like it failed again -.-
I would've suggested to then replace `run_tests.sh` with something like
```bash


#!/bin/bash

# skip tests on OSX as they fail unreliably
if [ "$(uname)" != Darwin ] ; then
    quast --test
    metaquast --test
fi
```
to just skip the tests on OSX for now to get a reliable build process.
But I only just now noticed we've got another problem:
The recipe states `license: GPL2` but this sadly isn't accurate as QUAST includes a lot of other software that is not GPL-2.0-licensed. Hence the license for the Conda package cannot be just `GPL2` but must include all license parts of the whole software package, so it should at least be something like `license: Custom` instead of `GPL2`.
But then there is another problem: I got aware of the licensing issue while re-reading `genemark.py` and thus read GeneMark's license, which states (highlighting added by me):
>1. Permitted Uses.  Licensee may:
>
>a) use the Product solely for Licensee's **own internal research** purposes.
>
>2. **Prohibited** Uses.  Licensee may not:
>
>a) transfer, **distribute**, lease or sub-license the Product.
>b) use the Product for any commercial purpose.
>c) charge for access or viewing of the Product.
>d) alter, modify, or adapt the Product or documentation, or portions thereof.
>
>3. Timeframe
>
>The License is valid for two year period from date of download. At the end
>of this period you will be reminded to renew the license and to download
>a new version of the software.

So to my understanding, we cannot distribute GeneMark as part of the `quast` package and hence GeneMark would have to be removed from the package. Regarding GeneMark's usage, you could
* remove that functionality from QUAST
* or download it via an `post_install.sh` script
* or download it in QUAST (like it's done for, e.g, Manta) (I would prefer the `post_install.sh` instead of this, though)

. (I didn't check all the licenses of the other included software and cannot comment whether anything else needs special treatment, though.)
@bgruening: What do you think about that licensing issue?

mbargull:
An even better alternative to options 2. and 3. would be: 

4. create a `genemark` recipe that uses `post-link.sh` to download GeneMark (something akin to the `gatk` recipe) and use that as a run requirement.

alexeigurevich:
As far as I understand, you can't download genemark using `wget` or something, you need to complete license form at http://exon.gatech.edu/GeneMark/license_download.cgi and after that you get the download link. So, probably we probably cannot add its downloading in `post-link.sh`.
Note: we have a separate license from Gene Probe Inc (current version is valid until August 2019 and can be renewed) for distributing GeneMark programs inside Quast package for academic use only: 
>1. Permitted Uses. Licensee may:
>a) use the Product for Licensee's own internal research purposes as well as;
>b) include the Product for distribution within the “QUAST” package for academic use to external, third party academic research institutions.

Quast commercial users should contact Gene Probe directly for obtaining appropriate license (see Quast License or Manual starting section, note that our own License with Gene Probe is not included in the package and is not publicly available online).
I think you are right and we should exclude GeneMark functionality from Quast package distributed via bioconda since bioconda commercial users could be not aware about GeneMark licensing specifics. Also I am not 100% sure that 1b) above allows such distribution from third-party installers like bioconda.

Note that in Quast package on Illumina BaseSpace (part of SPAdes App) we disabled GeneMark because of similar License issues. There is a functionality for such "disabling" in Quast already, so we could probably shoot two birds with one stone (not stable macOS tests and GeneMark licensing). Do you agree with it?

mbargull:
> Note: we have a separate license from Gene Probe Inc (current version is valid until August 2019 and can be renewed) for distributing GeneMark programs inside Quast package for academic use only

Good to hear that for QUAST itself everything is in place license-wise 😄 
> I am not 100% sure that 1b) above allows such distribution from third-party installers like bioconda.

I agree with your doubts since it's being distributed by Bioconda and I guess Bioconde itself would need a license for distribution then.

> There is a functionality for such "disabling" in Quast already, so we could probably shoot two birds with one stone (not stable macOS tests and GeneMark licensing). Do you agree with it?

Sounds great to me! I'd say just use that same "disabling"-functionality and extend the `exta: notes:` note saying it is disabled due to licensing issues. (And to make sure we don't distribute GeneMark unintentionally, be sure to modify `setup.py` to exclude the appropriate folders -- `MANIFEST.in` might be modified too with a `recursive-exclude` since `setup.py` uses `include_package_data=True`)

alexeigurevich:
Well, it looks like everything is fine now :) Both tests and licensing issues.

alexeigurevich:
Hi guys! Happy New Year!
Let me know if you want something more from our side or just merge this request if everything is fine already :)
Thank you for all your help! Now we become much more familiar with bioconda recipe creation which is really cool.

mbargull:
> Happy New Year!

To you too! I just commit some last changes to prevent build for different Python version to use the same filenames, so that `quast` will be available for Python 2.7, 3.5, 3.6.

One last question:  `quast` is storage-wise one of the largest packages on Bioconda due to the included dependencies. The biggest one of those is the `SILVA 16S rRNA gene database` which is downloaded with `setup.py install_full`. Would it make sense to not include that in the package?
If it is problematic to exclude it on default for now, we could work this out in some future update, I suppose.

Thanks again for all your effort!

mbargull:
Python 3 builds failed since Python 2-only libraries are unconditionally included, see https://travis-ci.org/bioconda/bioconda-recipes/jobs/324244229#L7248 and onward.
Most of them could simply be removed from the package so that  Python 3-compatible versions installed by Conda will be used instead. But, IIUC, Manta is Python 2-only, so if that one is included, `quast` needs to be Python 2-only anyways...

alexeigurevich:
Thanks for the update! See my answers to your questions/suggestions below.
 
> One last question: quast is storage-wise one of the largest packages on Bioconda due to the included dependencies. The biggest one of those is the SILVA 16S rRNA gene database which is downloaded with setup.py install_full. Would it make sense to not include that in the package?

My rough estimate is that >95% QUAST users do not need `install_full` and use only simple `install` functionality. The difference is in SILVA + blast utils (for MetaQuast without ref mode) and Manta (for structural variants detection) which both are rather specific use cases. I think it is fine to switch to `install` and skip SILVA installation which will reduce the size of the package by an order of magnitude. Note that if a user run QUAST in a mode requiring SILVA or Manta or something else not installed, QUAST will automatically download required stuff.
Do you agree with such change?

> Python 3 builds failed since Python 2-only libraries are unconditionally included, see https://travis-ci.org/bioconda/bioconda-recipes/jobs/324244229#L7248 and onward.
Most of them could simply be removed from the package so that Python 3-compatible versions installed by Conda will be used instead. But, IIUC, Manta is Python 2-only, so if that one is included, quast needs to be Python 2-only anyways...

Yes, Manta is Python 2-only, so for users with Python 3 we just disable this functionality. And we switched to another structural variants caller in the new release, so this problem is Quast v.4 only. Is there anyway to conditionally remove some parts of the code depending on the Python version? Something like conditional patching or whatever else?

alexeigurevich:
Hi, guys!
How should we finally proceed with this long-awaiting pull request? :)

I see two possible solutions for now:
1) replicate currently available quast-4.5 recipe, i.e. do not resolve the last issues with SILVA, Manta and Python versions and leave them "as is". At least this recipe will be better than 4.5 in some aspects and we can solve everything later in the new major update (5.0). 
2) switch to `install` where no SILVA and Manta are installed (reduce bioconda package, user can install SILVA and related tools later) and completely remove python source causing Python2/3 issues (substitute unconditionally included Python libs with bioconda imports; completely remove Manta since it is Python2-only).

mbargull:
Hi Alexey!
Apologies for yet another delay..
>  Is there anyway to conditionally remove some parts of the code depending on the Python version? Something like conditional patching or whatever else?

There are a couple of ways to achieve that, but I would say let's postpone that for another release :wink:.

> I see two possible solutions for now:

I opted for the middle ground solution: make the change from `install_full` to `install` now, but otherwise only build for Python 2 for now to get a `4.6.1` package out already. Let's tackle Python 3 builds in a subsequent PR (possibly for the future release then). Feel free to ping then!

The `pre-unlink.sh` I added in the latest commit is needed as a `conda remove` will only remove files that are part of the Conda package/installed by `conda` itself.

If you are fine with everything for now, we can finally merge this once the CI shows green (which sadly might take some time as Travis CI has a massive backlog for OSX builds currently 😒).

alexeigurevich:
Hi, Marcel!
Thanks for the update. I agree with your plan and the changes.

However, I see that tests failed even on Linux but with very strange error. Actually the log contains just a single word "null". Could you debug this?

mbargull:
No worries, Travis CI has had some problems, see https://www.traviscistatus.com/. Their latest message includes
> We are now back to processing jobs on all infrastructures and closely monitoring the situation. As a clean-up effort, we have to cancel all jobs that were previously stuck. 

. We'll just have to try to restart the build a couple of times over the next hours, I guess. (First retry just failed - will do the next one a bit later..)

mbargull:
Travis finally completed the PR! Linux builds should be available in a bit after the merge; OSX builds may still take some more time, though.
Alexey, thanks for your patience

alexeigurevich:
Great, thanks!

PR #7015, opened 2017-12-06
tassel: use git_url for old versions

mbargull:

Git snapshots on SourceForge are not persistent, hence we have to use `git_url`/`git_rev` directly, see failed build for commit https://github.com/bioconda/bioconda-recipes/commit/449cdda333fd98e1fc63a5aa7acbf9676f7343e5.

### DIFF
--- a/recipes/tassel/3.0.174/meta.yaml
+++ b/recipes/tassel/3.0.174/meta.yaml
@@ -6,9 +6,8 @@ package:
   version: {{ version }}
 
 source:
-  fn: {{ name|lower }}-{{ version }}.zip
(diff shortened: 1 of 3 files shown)


bgruening:
FYI:

> Git snapshots on SourceForge are not persistent, hence we have to use git_url/git_rev directly, see failed build for commit 449cdda.

We have this problem all over, but instead of using git, which is manipulable, we try to mirror everything to https://depot.galaxyproject.org. 

One versione we already have: https://depot.galaxyproject.org/software/tassel/tassel_5.2.40_src_all.tar.gz

mbargull:
So how would the process in this case work? I'm all in for using checksummed archives, but can you clarify what would exactly need to be done in such "time constrained" cases? We have to make SourceForge generate their snapshot but then how to instruct `cargo-port`(?) to mirror them quickly before they're already gone from the SourceForge server?

PR #7016, opened 2017-12-06
Bump snp-pipeline to version 1.0.1

stevendavis:

This is my first PR bumping snp-pipeline to version 1.0.1 with some dependency changes and a command test added. TravisCI tests passed.

### DIFF
--- a/recipes/snp-pipeline/build.sh
+++ b/recipes/snp-pipeline/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-$PYTHON setup.py install
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
 
 # Add more build steps here, if they are necessary.
 
--- a/recipes/snp-pipeline/build.sh
+++ b/recipes/snp-pipeline/build.sh
@@ -2,8 +2,6 @@
 
 $PYTHON setup.py install --single-version-externally-managed --record=record.txt
 
-# Add more build steps here, if they are necessary.
-
 # See
 # http://docs.continuum.io/conda/build.html
 # for a list of environment variables that are set during the build process.
--- a/recipes/snp-pipeline/meta.yaml
+++ b/recipes/snp-pipeline/meta.yaml
@@ -31,6 +31,23 @@ test:
   requires:
     - testfixtures
 
+  commands:
+    - cfsan_snp_pipeline                  --version
+    - cfsan_snp_pipeline run              --version
+    - cfsan_snp_pipeline data             --version
+    - cfsan_snp_pipeline index_ref        --version
+    - cfsan_snp_pipeline map_reads        --version
+    - cfsan_snp_pipeline call_sites       --version
+    - cfsan_snp_pipeline filter_regions   --version
+    - cfsan_snp_pipeline merge_sites      --version
+    - cfsan_snp_pipeline call_consensus   --version
+    - cfsan_snp_pipeline merge_vcfs       --version
+    - cfsan_snp_pipeline snp_matrix       --version
+    - cfsan_snp_pipeline distance         --version
+    - cfsan_snp_pipeline snp_reference    --version
+    - cfsan_snp_pipeline collect_metrics  --version
+    - cfsan_snp_pipeline combine_metrics  --version
+
 extra:
   skip-lints:
     - uses_setuptools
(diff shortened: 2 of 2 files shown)


recipes/snp-pipeline/meta.yaml
```diff
@@ -14,7 +14,6 @@ requirements:
   build:
     - python
     - pyvcf >=0.6.7
-    - setuptools
```
mbargull: This shouldn't be removed.

recipes/snp-pipeline/meta.yaml
```diff
@@ -33,6 +31,13 @@ test:
   requires:
     - testfixtures
 
+  commands:
+    - cfsan_snp_pipeline --version
+
+extra:
```
druvus: Try to remove `setuptools` from run reqs. It is normally only needed during build step.

mbargull: Please don't as it it really required: https://github.com/CFSAN-Biostatistics/snp-pipeline/blob/v1.0.1/snppipeline/data.py#L12.

druvus: thanks @mbargull my mistake! I missed it used `pkg_resources`

recipes/snp-pipeline/meta.yaml
stevendavis: @mbargull Thank you for the review.  May I ask, what are the guidelines for deciding which packages are build requirements?

mbargull: You're welcome.
The build requirements can be very individual for each package. But for simple Python packages, like the one at hand, it is quite easy to determine:
- need `python`
- need anything that's not part of Python's stdlib which is imported in `setup.py`
- need anything specified in `setup(...)`'s `setup_requires` parameter

Exceptions may occur, like for example:
 - if `setup.py` imports the package itself, most likely it's best to add all run requirements as build requirements, too
 - if some extensions are built, then additional compiler packages are required

In `snp-pipline`'s case we thus only need `python` and `setuptools`, i.e. `pyvcf`, `psutil` and `biopython` can be removed from build requirements.

recipes/snp-pipeline/meta.yaml
```diff
@@ -1,11 +1,11 @@
...
 source:
-  fn: snp-pipeline-0.7.0.tar.gz
-  url: https://pypi.python.org/packages/62/6b/5c7c7d1ad8239f2dd051a91e8bc55cd98418df5719d0fd4605f863f71d9a/snp-pipeline-0.7.0.tar.gz
-  md5: 8f5971ca4729364fd22e26e712be862d
+  fn: snp-pipeline-1.0.1.tar.gz
+  url: https://pypi.python.org/packages/9c/86/e2a212735e229b5dd65c7f58f53822e820507bb4f95f870687f50046f99c/snp-pipeline-1.0.1.tar.gz
+  md5: d42bfefd2ccccdb5f880f3a74f4ff165
 
 build:
   preserve_egg_dir: True
```
mbargull: `preserve_egg_dir: True` can be removed, I think

mbargull: Please also add to the `build:` section:
```yaml
  entry_points:
    - cfsan_snp_pipeline = snppipeline.cfsan_snp_pipeline:main
    - qarrayrun = snppipeline.qarrayrun:main
```
That way we get scripts created by `conda`/`conda-build` instead of `setuptools`, which should be preferred.

recipes/snp-pipeline/meta.yaml
```diff
@@ -14,7 +14,6 @@ requirements:
   build:
     - python
     - pyvcf >=0.6.7
-    - setuptools
     - psutil
     - biopython
```
mbargull: So we can remove `pyvcf`, `psutil` and `biopython` from the build requirement and only use
```yaml
  build:
    - python
    - setupools
```

recipes/snp-pipeline/meta.yaml
```diff
@@ -33,6 +31,13 @@ test:
   requires:
     - testfixtures
 
+  commands:
+    - cfsan_snp_pipeline --version
+
+extra:
+  skip-lints:
+    - uses_setuptools
```
mbargull: Can you add a comment for this? E.g.,
```yaml
```

recipes/snp-pipeline/build.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-$PYTHON setup.py install
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
 
 # Add more build steps here, if they are necessary.
```
mbargull: You can remove that comment.

Issue #6974, opened 2017-11-30
sambamba optional runtime dependency on samtools/bcftools and package-split samtools


vinjana:
I need to install sambamba 0.5.9 with samtools/bcftools 0.1.19. Unfortunately, sambamba has a runtime dependency declared on bcftools that was split out of the samtools package some time after 0.1.19. Therefore, if I want to install sambamba 0.5.9 it will pull in the newer bcftools (which will BTW **silently** override the original 0.1.19 version). I would like to note that samtools/bcftools is an **optional** runtime dependency of sambamba, i.e. most of sambamba subtools work without it.

What should I do here? I think the correct solution would be to completely remove bcftools as dependency from sambamba. Any objections? Also, bcftools should get a dependency `samtools >=1.2` (or so, need to figure out the exact version) to avoid the overwriting of bcftools (which will result in a mix of samtools 0.1.19 and a newer bcftools). 

Another solution that would keep the runtime dependency would be to add a complex dependency of the sort "samtools <1.2 | samtools >=1.2, bcftools >=1.2" -- unfortunately this is the wrong syntax and I am not yet even sure whether conda[-build] can deal with such complex dependencies.

It would be nice to get some feedback on this before all the fuzz with a PR. So, comments are appreciated!

bgruening:
@chapmanb @tomkinsc do you have any input here? Thanks!

chapmanb:
+1 on removing the samtools and bcftools requirements from the recipe. I think it's fine to mention these as optional functionality but they're not required for most things you'd do with sambamba.

vinjana:
Sorry for asking again, but parts of my question remain unanswered. What about older package versions? I need the sambamba 0.4.6 and 0.5.9 package versions w/o the samtools dependency. I guess, there are people who do not include the build version in their environment dependency explicitly. Would backporting then still be o.k.?

johanneskoester:
Yes, backporting would be ok. You can add those versions as subfolders (if not already there), and increase the build number.

bgruening:
I think this is resolved - feel free to reopen if there is still something unclear.
Thanks @vinjana!

tomkinsc:
Removing the build dependenciew seems like the right call; thanks for handling this. We can always a `sambamba-full` package if it becomes an issue for users.

PR #7024, opened 2017-12-07
Sambamba/samtools disconnect and Sambamba 0.5.9

vinjana:

### DIFF
--- a/recipes/sambamba/meta.yaml
+++ b/recipes/sambamba/meta.yaml
@@ -11,13 +11,7 @@ source:
   md5: 07d82a2102126565e6796658face7c71 # [osx]
 
 build:
-  number: 0
-
-requirements:
-  build:
-  run:
-    - samtools # required for mpileup
-    - bcftools # required for mpileup
+  number: 1
 
 test:
   commands:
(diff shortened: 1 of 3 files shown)


bgruening:
@vinjana what about leaving 0.6.6 as it is and creating 0.6.7 without samtools as dependency?

vinjana:
You mean 0.5.9 and 0.6.7 without the samtools dependency and leave 0.6.6 as is? I can do that. Just validate that this is what you meant.

recipes/sambamba/meta.yaml
```diff
@@ -11,13 +11,7 @@ source:
   md5: 07d82a2102126565e6796658face7c71 # [osx]
 
 build:
-  number: 0
-
-requirements:
-  build:
-  run:
-    - samtools # required for mpileup
-    - bcftools # required for mpileup
```
johanneskoester: Does that mean those are no longer needed by sambamba?

vinjana: No. It means they were never really needed by sambamba (depending on how you look at it :-) ). They are only needed for sambamba mpileup, whose help-page also says so. Please see the discussion at  #6974.

vinjana:
O.k. how to proceed here? Acceptible?

johanneskoester:
Thanks!

PR #7025, opened 2017-12-07
FlashLFQ initial recipe

jj-umn:
Installs FlashLFQ release 0.1.99
https://github.com/smith-chem-wisc/FlashLFQ


recipes/flashlfq/elements.dat
```diff
@@ -0,0 +1,2878 @@
+<html>
+<head>
```
bgruening: Why is this file needed here @jj-umn? Can we store it on cargo-port?

jj-umn: @bgruening The FlashLFQ requires that file and expects it in the same dir as the executable.  I asked the code developer to include it in the distribution for the next release.

PR #7038, opened 2017-12-08
Watchdog RC Update

klugem:


recipes/watchdog-wms/meta.yaml
```diff
@@ -1,5 +1,5 @@
 {% set name = "watchdog-wms" %}
-{% set version = "1.2.4b" %}
+{% set version = "1.2.0" %}
```
bgruening: You are decreasing the version number here. This will not be the default version if people install watchdog.

recipes/watchdog-wms/meta.yaml
```diff
@@ -22,12 +23,12 @@ package:
     version: {{ version }}
 
 build:
-  number: 0
+  number: 1
```
bgruening: If you change the version number of the tool you don't need to increase the build number.

recipes/watchdog-wms/meta.yaml
klugem: That's what I thought...I will change the version number. ;)

PR #7049, opened 2017-12-11
Update R MutationalPatterns

vladsaveliev:

Updating [MutationalPatterns](http://bioconductor.org/packages/release/bioc/html/MutationalPatterns.html) R package.

`r-mutationalpatterns` recipe points to a [GitHub release](https://github.com/UMCUGenetics/MutationalPatterns/releases), however it looks like the authors are releasing new versions only in [Bioconductor](http://bioconductor.org/packages/release/bioc/html/MutationalPatterns.html): the latest GitHub version is 1.1.0, the latest bioconductor one is 1.4.1. So it makes sense to re-point recipe to bioconductor.

However it seems that bioconductor packages in bioconda are named using a `bioconductor-` prefix instead of `r-`, so I'm submitting here a new package `bioconductor-mutationalpatterns`. But I'm also updating and keeping `r-mutationalpatterns` for back-compatibility. Let me know if that's a bad decision and we should stick with only one recipe, either `r-` or `bioconductor-`.


bgruening:
@vladsaveliev I think it would be better to remove the r-* package altogether.

vladsaveliev:
@bgruening, thanks for the suggestion! Removed the r- one.

druvus:
Pinging @chapmanb who added the initial version to inform about swapping from r- to bioconductor- prefix

chapmanb:
+1 for swapping over. Thanks for taking this on Vlad. Merging now.

PR #7050, opened 2017-12-11
Watchdog: Testing fails caused by too small timeout ?

klugem:


klugem:
[https://travis-ci.org/bioconda/bioconda-recipes/jobs/314701984](url)

Seems for me that testing of the build failed for Macos because the host is so slow ?

klugem:
Was indeed the problem...

recipes/watchdog-wms/run_test.sh
```diff
@@ -7,7 +7,7 @@ VERSION="${PKG_NAME}-${PKG_VERSION}-${PKG_BUILDNUM}"
 OUT="${PREFIX}/share/${VERSION}"
 PATH="$PATH:$OUT/core_lib/wrapper/"
 
-function timeoutbash() { bash -c '(sleep 10; kill -9 $$ 2>/dev/null) & exec $@' $0 $@; }
+function timeoutbash() { bash -c '(sleep 180; kill -9 $$ 2>/dev/null) & exec $@' $0 $@; }
 
 # basic test
 timeoutbash watchdog-cmd --help 2>&1>/dev/null
```
mbargull: Does the program not terminate itself after it output the help information?

recipes/watchdog-wms/run_test.sh
```diff
@@ -7,7 +7,7 @@ VERSION="${PKG_NAME}-${PKG_VERSION}-${PKG_BUILDNUM}"
 OUT="${PREFIX}/share/${VERSION}"
 PATH="$PATH:$OUT/core_lib/wrapper/"
 
-function timeoutbash() { bash -c '(sleep 10; kill -9 $$ 2>/dev/null) & exec $@' $0 $@; }
+function timeoutbash() { bash -c '(sleep 180; kill -9 $$ 2>/dev/null) & exec $@' $0 $@; }
```
mbargull: Since you added the `coreutils` run requirement, you can use the provided `timeout` command. This should have the advantage that it doesn't wait for the timeout even if the callee is already terminated.
Also, IIRC, `coreutils` provides GNU sed, hence you can just use `sed -i` instead of your custom `sedinline`.

recipes/watchdog-wms/run_test.sh
klugem: Sure, it does. I included the timeout command for the more advanced test.
("not terminating" could be caused by a bug in the github version in that case.)

recipes/watchdog-wms/run_test.sh
klugem: Good point, I will change that the next time I update the package.

PR #7071, opened 2017-12-13
dascrubber update

mictadlo:


mictadlo:
Hi @druvus  and @bgruening
By any chance, do you know what did I miss?

Thank you in advance.

Michal

druvus:
It seems you missed to include `REPtrim`. 

Those are very old recipes from the days before bioconda-utils. Nowadays it is better to not use ` > /dev/null` so you can see what is going wrong.

PR #7077, opened 2017-12-13
Add recipe seq-seq-pan

cjandrasits:

@bioconda/core please review, this is my first slightly complicated recipe... thanks!

### DIFF
--- a/recipes/seq-seq-pan/meta.yaml
+++ b/recipes/seq-seq-pan/meta.yaml
@@ -10,11 +10,11 @@ source:
     
 build:
     number: 0
-    skip: True # [osx]
+    skip: True # [osx or py27]
     
 requirements:
     run:
-        - python ==3.5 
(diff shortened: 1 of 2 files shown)


recipes/seq-seq-pan/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+    url: https://gitlab.com/chrjan/seq-seq-pan/repository/v{{ version }}/archive.tar.gz
+    md5: 2135730fee2eec0b018605bef4029f38
+    
+build:
+    number: 0
+    skip: True # [osx]
+    
+requirements:
+    run:
+        - python ==3.5
```
druvus: It requires py35 or higher so you should remove restriction here and instead include a skip for py27 in the build section.

cjandrasits:
@druvus thanks!

PR #7078, opened 2017-12-13
New recipe for SeroBA

hellothisisMatt:

### DIFF
--- a/recipes/seroba/build.sh
+++ b/recipes/seroba/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# Have to download the precompiled binary as we don't have a compiler on conda which supports C++14.
+# We don't have a compiler on Conda for C++14, so downloading pre-compiled binary and moving to bin.
 wget https://github.com/refresh-bio/KMC/releases/download/v3.0.0/KMC3.linux.tar.gz
 tar xzf KMC3.linux.tar.gz
 
--- a/recipes/seroba/build.sh
+++ b/recipes/seroba/build.sh
@@ -1,15 +1,3 @@
 #!/bin/bash
 
-# We don't have a compiler on Conda for C++14, so downloading pre-compiled binary and moving to bin.
-wget https://github.com/refresh-bio/KMC/releases/download/v3.0.0/KMC3.linux.tar.gz
-tar xzf KMC3.linux.tar.gz
-
-chmod 755 kmc
-chmod 755 kmc_tools
-chmod 755 kmc_dump
-
-cp kmc $PREFIX/bin
-cp kmc_tools $PREFIX/bin
-cp kmc_dump $PREFIX/bin
-
 $PYTHON setup.py install --single-version-externally-managed --record=record.txt
--- a/recipes/seroba/build.sh
+++ b/recipes/seroba/build.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# Have to download the precompiled binary as we don't have a compiler on conda which supports C++14.
+wget https://github.com/refresh-bio/KMC/releases/download/v3.0.0/KMC3.linux.tar.gz
+tar xzf KMC3.linux.tar.gz
+
+chmod 755 kmc
+chmod 755 kmc_tools
+chmod 755 kmc_dump
+
+cp kmc $PREFIX/bin
+cp kmc_tools $PREFIX/bin
+cp kmc_dump $PREFIX/bin
+
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
--- a/recipes/seroba/meta.yaml
+++ b/recipes/seroba/meta.yaml
@@ -49,3 +49,6 @@ about:
 extra:
   skip-lints:
     - uses_setuptools  # seroba uses pkg_resources
+  container:
+    # matplotlib (via ariba) depends on libGL.so.1 from the system
+    extended-base: true
--- a/recipes/seroba/meta.yaml
+++ b/recipes/seroba/meta.yaml
@@ -1,0 +1,49 @@
+{% set name = "seroba" %}
+{% set version = "1.0.0" %}
+{% set md5sum = "fcd6802070aa90d3f4810bb72a23cdce" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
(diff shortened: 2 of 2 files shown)


hellothisisMatt:
Trying to figure out why travis-ci testing is failing, but it seems like a dependency cannot find SeroBA. This recipe works locally using ./simulate-travis.py, and conda build.

recipes/seroba/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
+  url: https://github.com/sanger-pathogens/{{ name }}/archive/v{{ version }}.tar.gz
+  md5: {{ md5sum }}
+
+build:
+  number: 0
+
+requirements:
```
mbargull: Please add `setuptools` to build as well as run requirements.

recipes/seroba/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  commands:
+    - seroba -h
+
+about:
+  home: https://github.com/sanger-pathogens/seroba
+  license: GPL3.0
+  license_family: GPL
+  summary: 'SeroBA is a k-mer based Pipeline to identify the Serotype from Illumina NGS reads for given references.'
+
+description: |
```
mbargull: `description` should be part of the `home` section.

recipes/seroba/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+      SeroBA is a k-mer based Pipeline to identify the Serotype from Illumina NGS reads for given references. 
+      You can use SeroBA to download references from (https://github.com/phe-bioinformatics/PneumoCaT) 
+      to do identify the capsular type of Streptococcus pneumoniae.
```
mbargull: Please append
```yaml
extras:
  skip-lints:
```

recipes/seroba/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# We don't have a compiler on Conda for C++14, so downloading pre-compiled binary and moving to bin.
```
mbargull: Is it possible for you to update the `kmc` recipe to version `3.0.0` instead of including it here? The existing `kmc` package also provides `kmc`, `kmc_dump`, as well as `kmc_tools`.

recipes/seroba/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
+  url: https://github.com/sanger-pathogens/{{ name }}/archive/v{{ version }}.tar.gz
+  md5: {{ md5sum }}
+
+build:
+  number: 0
```
mbargull: Please add
```yaml
  skip: True  # [not py3k]
```
as is it is Python 3 only. (This will also solve the build failure you got.)

recipes/seroba/meta.yaml
```diff
@@ -0,0 +1,55 @@
+{% set name = "seroba" %}
+{% set version = "1.0.0" %}
+{% set md5sum = "fcd6802070aa90d3f4810bb72a23cdce" %}
```
mbargull: `sha256` [should be preferred](http://bioconda.github.io/guidelines.html#hashes) unless only `md5` is provides alongside the source.

recipes/seroba/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
+  url: https://github.com/sanger-pathogens/{{ name }}/archive/v{{ version }}.tar.gz
+  md5: {{ md5sum }}
+
+build:
+  number: 0
+
+requirements:
+  build:
```
mbargull: Oh, and the build requirements are only
```yaml
  build:
    - python
    - setuptools
```
, nothing else is needed of the build itself. (Ok, apart from `wget` -- but only because it is used in `build.sh` which could be removed when the `kmc` recipe is updated.)

recipes/seroba/meta.yaml
hellothisisMatt: After adding this to the recipe, it still fails linting for some reason. Passes on local `./simulate-travis.py`

mbargull: Very sorry about my typo!

recipes/seroba/build.sh
mbargull: I'll give it a shot myself, no worries.

recipes/seroba/meta.yaml
hellothisisMatt: Sorry, which is the typo? :laughing:

mbargull: The surplus `s` in `extras` that you had to track down :X :wink:

mbargull:
@hellothisisMatt: I let `kmc-3.0.1` be built and updated your PR to depend on it -- let's see if it works.

mbargull:
This LGTM so far -- one last question, though:
Where do the requirements
```yaml
  run:
    - pyqt ==4.11.4
    - glib
    - xorg-libxext 
    - xorg-libsm
    - xorg-libxrender
```
come into play?

hellothisisMatt:
@mbargull Thank you so much for your help!
```
    - pyqt ==4.11.4
    - glib
    - xorg-libxext 
    - xorg-libsm
    - xorg-libxrender
```
These requirements were necessary because Travis didn't have these libraries so it was failing testing.

mbargull:
You are very welcome!
The test/build on Travis might've likely failed due to different reasons and might succeed now while leaving them out. Do you mind if I add an additional commit to remove those dependencies just to see whether it works without them?

hellothisisMatt:
Yeah for sure! It didn't work on my local ./simulate-travis.py.

mbargull:
I let it run locally and saw what the issue was: it was a `seroba` -> `ariba` -> `matplotlib` -> `pyqt 5` -> `libGL.so.1` dependency which you circumvented by using `pyqt 4`. I just added the appropriate fix so that the package can be installed with either of `pyqt 4` and `pyqt 5`.
When the CI passes, this should be good to go, I think 😄.

hellothisisMatt:
Perfect! Thank you so much for your help with the recipe.

mbargull:
I ended up using another more direct and even better fix 😄.
Thanks for the recipe @hellothisisMatt, merging!

PR #7079, opened 2017-12-13
Update minvar

ozagordi:

A rebuilt was necessary to correct some bugs in `data_package`. Apologies again.


recipes/minvar/meta.yaml
```diff
@@ -3,14 +3,14 @@ package:
...
-  number: 0
+  number: 1
   # minvar written for python 3.x only, so skip py27
   # setuptool_scm_git_archive not available in py36
   skip: True # [py27 or py36]
 
 source:
   url: https://github.com/ozagordi/MinVar/archive/v2.0.tar.gz
-  md5: 5ba1c9814079c770e0c1502abba068b6 
+  md5: 06f8fde34050ef9228722594939b22b5
```
bgruening: Why has the checksum changed? Is this a new release?

ozagordi:
I updated the tag (deleted old v2.0 and pushed a new one) rather than pushing a v2.0.1. I didn’t feel like having a version 2.0 that was completely unusable.

recipes/minvar/meta.yaml
ozagordi: I updated the tag (deleted old v2.0 and pushed a new one) rather than pushing a v2.0.1. I didn’t feel like having a version 2.0 that was completely unusable.

bgruening: This is not very reproducible. We encourage everyone to not change released version.

kyleabeauchamp: Agree with @bgruening

ozagordi: Ok. I will keep it in mind for next update, thanks.

PR #7080, opened 2017-12-13
Ectyper

UBCSamSung:


recipes/ectyper/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+    summary: Ectyper is a python script for serotyping e.coli genomes
+    author: Camille La Rose, Chad Laing, Sam Sung
+    home: https://github.com/phac-nml/ecoli_serotyping
+
+package:
+    name: ectyper
+    version: 0.0.8
+
+source:
+    url: https://github.com/phac-nml/ecoli_serotyping/archive/0dbfb220d76f7c0fe840ed7c304d8cc1f6e994db.zip
```
bgruening: There is a 0.1 alpha release: https://github.com/phac-nml/ecoli_serotyping/releases
Could you maybe ask for a new release tarball?

UBCSamSung:
@bgruening Thank you for reviewing. The conda package now use latest release tarball.

recipes/ectyper/meta.yaml
UBCSamSung: It is done.

recipes/ectyper/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+    name: ectyper
+    version: 0.0.8
+
+source:
+    url: https://github.com/phac-nml/ecoli_serotyping/archive/v0.1.tar.gz
+    md5: 06b5793ed0f7000a05e20600b1ac5d0c
+
+requirements:
+    build:
+        - python 3.6.3.*
```
bgruening: Python should not be pinned this way.
Is there any reason to skip python 3.5? Please skip all python version in the build/skip section.

recipes/ectyper/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+    run:
+        - python 3.6.3.*
+        - pandas 0.21.0.*
```
bgruening: All other pins are here very strict, is this really needed? Biopython for example is stable afaik and does not need to be pinned to 1.69.

UBCSamSung:
@bgruening Thank you for reviewing. I have removed the strict dependency requirement and skipped python 2.7. All checks have passed.

recipes/ectyper/meta.yaml
```diff
@@ -0,0 +1,33 @@
+about:
+    license: MIT
+    summary: Ectyper is a python script for serotyping e.coli genomes
+    author: Camille La Rose, Chad Laing, Sam Sung
+    home: https://github.com/phac-nml/ecoli_serotyping
+
+package:
+    name: ectyper
+    version: 0.1.0
```
bgruening: Now the number is still not the number from the github release. It should be 0.1 isn't it?
You could also use template vars for the version number, this way an upgrade is easier later on.

bgruening:
Thanks @UBCSamSung. Just one last comment.

UBCSamSung:
@bgruening It should be good now.

bgruening:
Thanks @UBCSamSung!

PR #7083, opened 2017-12-13
Moff bump to v1.2

caleb-easterly:


recipes/moff/meta.yaml
```diff
@@ -1,12 +1,14 @@
+{% set version = "v1.2" %}
```
druvus: I think it is better praxis to call it 1.2 and add "v" in front of it in the url.

caleb-easterly: Changed!

PR #7086, opened 2017-12-14
Bump IgBLAST to 1.8.0

marcelm:

### DIFF
--- a/recipes/igblast/build.sh
+++ b/recipes/igblast/build.sh
@@ -16,7 +16,7 @@ if [ $(uname) == Linux ]; then
     # provide this. See https://github.com/bioconda/bioconda-recipes/pull/3020
 
     cd c++
-    ./configure --prefix=$PREFIX --with-sqlite3=$PREFIX
+    ./configure --prefix=$PREFIX --with-sqlite3="$PREFIX" --with-lmdb="${PREFIX}"
     make -j2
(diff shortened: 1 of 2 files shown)


marcelm:
Ok, was afraid this is going to happen. The error is here:
```
configure: trying to build the NCBI SRA/VDB Toolkit from GitHub.
configure: error: SRA/VDB build failed with status 1
```
The build system for IgBLAST tries to clone and install https://github.com/ncbi/ncbi-vdb.git and https://github.com/ncbi/ngs.git . The script in the IgBLAST source distribution that does this and produces the above error is `ncbi-igblast-1.8.0-src/c++/scripts/common/add_vdb.sh`. Any messages are suppressed, so I cannot even tell what is going wrong on Travis.

The sra-tools recipe seems to have the same problem. It solves it (sort of) by downloading ncbi-vdb and ngs in a more controlled way as part of the `build.sh` script. In my opinion, that is not the way to go either. Separate ncbi-vdb and ngs packages should be created on which both igblast and sra-tools then could depend.

I don’t think I have time for this at the moment, so I’ll close this PR in a few days unless someone steps up.

marcelm:
Considering <https://github.com/ncbi/ngs/issues/14>, I’m not looking forward at all to packaging the dependencies separately, so closing for now.

PR #7098, opened 2017-12-15
GMAP-fusion: Adding new recipe.

tstrnity:

### DIFF
--- a/recipes/gmap-fusion/meta.yaml
+++ b/recipes/gmap-fusion/meta.yaml
@@ -21,8 +21,9 @@ requirements:
     - perl-set-intervaltree
     - perl-db-file
     - samtools >=1.3
-    - bowtie2
-    - gmap
+    - bowtie2 >=2.1
+    # The versions of gmap later than 2017-05-08 do not work with GMAP-fusion version 0.3.0.
+    - gmap <=2017.05.08
 
 test:
   commands:
--- a/recipes/gmap-fusion/meta.yaml
+++ b/recipes/gmap-fusion/meta.yaml
@@ -16,6 +16,7 @@ build:
 
 requirements:
   build:
+    - python
   run:
     - perl
     - perl-set-intervaltree
--- a/recipes/gmap-fusion/meta.yaml
+++ b/recipes/gmap-fusion/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "gmap-fusion" %}
 {% set version = "0.3.0" %}
-{% set build = "0" %}
+{% set build = "1" %}
 
 package:
   name: {{ name }}
@@ -16,7 +16,6 @@ build:
 
 requirements:
   build:
-    - python
   run:
     - perl
     - perl-set-intervaltree
--- a/recipes/gmap-fusion/meta.yaml
+++ b/recipes/gmap-fusion/meta.yaml
@@ -1,0 +1,47 @@
+{% set name = "gmap-fusion" %}
+{% set version = "0.3.0" %}
+{% set build = "0" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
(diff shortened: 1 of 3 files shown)


recipes/gmap-fusion/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - samtools >=1.3
+    - bowtie2 >=2.1
+    # The versions of gmap later than 2017-05-08 do not work with GMAP-fusion version 0.3.0.
+    - gmap <=2017.05.08
+
+test:
+  commands:
+    - "GMAP-fusion --help 2>&1 | grep Required"
+    # The following scripts all need to be tested from their install location.
+    - "$PREFIX/share/{{ name }}-{{ version }}-{{ build }}/util/compare_fusion_align_to_best_nonfusion_align.pl 2>&1 | grep usage"
```
johanneskoester: In Bioconda such scripts have to be made available within $PREFIX/bin. The most simple way is to create symlinks. If that does not work, you have to either write shell script wrappers or patch the scripts. Ideally, this would be fixed upstream, because it is a major design flaw in the software. But for now, wrappers or patches are fine.

tstrnity: Hello Johannes,
Thank you for considering my pull request. I understand what you are saying. However, I would like you to consider the following information.
These scripts would not normally be called by a GMAP-Fusion user. They are called by the GMAP-fusion program as a part of its pipeline. These "programs" basically operate as subroutines for the GMAP-fusion program. The designers of the software have a consistent model with which they write programs. They design using discrete program units that can be unit tested and pipelined. One feature of this design is that it allows programs whose jobs get interrupted to be restarted at discrete places in the pipeline. That is a design consideration that this model fulfills.
I can go ahead and put in symlinks, but they would be superfluous as far as a user of the software is concerned and would clutter up the user's bin directory (and the contents of the build.sh file as well). Here is the list of files:
    compare_fusion_align_to_best_nonfusion_align.pl
    estimate_fusion_read_support.pl
    filter_gmap_chims_via_dist_from_exon_bounds.pl
    gmap_genome_gff3_to_chim_summary.pl
    prep_evidence_for_IGV.pl
    retrieve_fusion_transcript_candidates.pl
    revise_chim_candidate_list_via_split_align_reassessment.pl
    screen_fusion_genome_breakpoint_for_confounding_kmers.pl
    screen_fusion_trans_genes_for_kmer_confounders.pl
    split_fusion_gmap_parts.pl
    write_GMAP_Fusion_annot_report.pl

I would also reference my pull request for DISCASM: https://github.com/bioconda/bioconda-recipes/pull/6437
and its recipe meta.yaml file: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/discasm/meta.yaml
Also, please look at my pull request for Fusion-Inspector: https://github.com/bioconda/bioconda-recipes/pull/7161
and its recipe meta.yaml file: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/fusion-inspector/meta.yaml
(FusionInspector's has 29 sub-programs in its util directory.)

Let me know if you still want me to add symlinks for these 11 sub-programs and I will do it.
Thanks,
Cicada Dennis

johanneskoester: Ok, in that case you are indeed right to not expose them in bin. Sorry for mypremature judgement. I misinterpreted their purpose because of the separate test cases.

johanneskoester:
And sorry for the long delay until review! Sometimes, we miss a PR...

recipes/gmap-fusion/meta.yaml
tstrnity: Thanks so much! I should have changed the comment just before the test, the way I did in the Fusion-Inspector. With regard to the long wait, I saw there were a lot of requests in the queue and appreciate that they can get buried.
Sincerely,
Cicada Dennis

PR #7099, opened 2017-12-15
Pinned prestor's pandoc dependency to 1.19.2 since 2.x removes an opt…

tfenne:
…ion that is used.

Just a small update to a recipe to pin the version of pandoc that is used.  Without the pin pandoc resolves to pandoc 2.x which removed an option that prestor uses.

### DIFF
--- a/recipes/r-prestor/meta.yaml
+++ b/recipes/r-prestor/meta.yaml
@@ -47,7 +47,7 @@ requirements:
     - r-bibtex
     - r-captioner
     - r-alakazam
-    - pandoc 1.19.2
+    - pandoc <2.0
     - texlive-core
 
 test:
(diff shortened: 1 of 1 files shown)


recipes/r-prestor/meta.yaml
```diff
@@ -47,7 +47,7 @@ requirements:
     - r-bibtex
     - r-captioner
     - r-alakazam
-    - pandoc
+    - pandoc 1.19.2
```
bgruening: can we just pin to <2.0

tfenne: Done!

bgruening:
@tfenne do you have time to grep through the source and fix other recipes as well?

bgruening:
Thanks!

PR #7113, opened 2017-12-18
kmc: update to 3.0.1

mbargull:

@notestaff: If possible, can you take a look at this. Instead of the `autotools` patch, this uses the project's makefiles, though. (The big `autotools` patch was a bit unwieldy to review for me, sorry :X.)

The patches incorporate:
* alterations to makefiles to make them a bit more customizable, i.e., https://github.com/refresh-bio/KMC/pull/45.
* on Linux, don't statically link the standard library as I don't think we have a package for a static `stdlibc++` available. The provided libraries remain statically linked, though.
* backport `std::make_unique` which is the only thing that makes `kmc_tools` require C++14 instead of C++11. That way `kmc_tools` is compiled, too

As for `zlib`/`bzip2`: Those are currently -- just like in the original source -- statically linked to on Linux (to the ones provided by KMC) and hence actually needn't be build/run requirements -- but I feel dynamically linking should be preferred, right?

As with the previous recipe, usage of `asmlib` is disabled. But actually KMC now uses another library, `vectorclass`, which -- to my understanding -- poses licensing issues for KMC... See: https://github.com/refresh-bio/KMC/issues/44.


johanneskoester:
@mbargull I think switching to dynamic linking should be only preferred it would be no effort. Especially for relatively small libraries like zlib and bzip2. Hence, you can keep the static linking.

mbargull:
It didn't really pose any effort -- I just hadn't pushed those changed yet.. Thanks for the comment, though!

PR #7124, opened 2017-12-19
Updating Bio_Hansel to v1.1.0

hellothisisMatt:

### DIFF
--- a/recipes/bio_hansel/meta.yaml
+++ b/recipes/bio_hansel/meta.yaml
@@ -13,7 +13,7 @@ source:
   '{{ hash_type }}': '{{ hash_value }}'
 
 build:
-  number: 2
+  number: 0
   skip: True # [not py3k]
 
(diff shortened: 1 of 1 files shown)


recipes/bio_hansel/meta.yaml
```diff
@@ -31,6 +32,7 @@ requirements:
     - numpy >=1.12.1
     - pandas >=0.20.1
     - attrs ==17.2.0
+    - pyahocorasick >=1.1.6
```
bgruening: This seems to be missing in our supported channels.

hellothisisMatt: Should be available on `conda-forge` now.

hellothisisMatt: Ah, this package is not available for OSX. I added `[not py3k]` to the skip parameter.

PR #7134, opened 2017-12-20
DeepVariant: initial recipe (0.4.1)

chapmanb:
Initial work on DeepVariant package. It works now on recent machines
Ubuntu 16.04+, Centos 7+ with pre-built binaries but left notes for
additional work to do:

- Uses pre-built binaries, building problematic due to clif dependency
- Need recent kernel with GLIBC > 2.23 due to pre-built htslib
- Build patches __main__.py in zip files to generalize python dependency
- Requires numpy 1.13, want to sync with CONDA_NPY
- Uses python wrappers that don't expose all underlying options
- Ships with pre-built model. Longer term will need way to fetch models on demand.

Fixes google/deepvariant#9


bgruening:
I looked into it as well and it seems that the model is several 100 MB. I guess we need to use the same trick as we do with bioconductor data packages before we can merge this. 

Thanks @chapmanb!

chapmanb:
Björn;
Thanks for the review on this. What trick would you like me to do? I was going to bundle together into one big package from now since there is only a single model. Do you prefer a different approach before getting it in?

bgruening:
@chapmanb we could simply use a post-link and a post_unlink script as we do for R. See here for the latest iteration which is currently building on `bulk`.

https://github.com/bioconda/bioconda-recipes/tree/bulk/recipes/bioconductor-fdb.infiniummethylation.hg19

All you need to do is a simple wget call. I can work on this if you prefer. I guess we can store the model on https://github.com/galaxyproject/cargo-port and give it a version as well.

chapmanb:
Thanks, that makes sense although is complicated a bit here because currently the model and pre-built binaries are in one tarball from GitHub so you'd be double downloading them. I can discuss with Google team about making the model a separate download. At some point they'll have multiple models so the way they have it now won't scale.

Since I want to sync with them on some of the other points as well (building on older CentOS, trying to fix the clif dependency to make it easier) should we merge this as a first pass to make it available to folks and then iterate to improve?

chapmanb:
We could do, I was suggesting coordinating with the Google team to push hosting the models as part of their deployment process. I feel like that will be more sustainable long term to keep models/code in sync, so we don't have to do it as part of each DeepVariant release, and can just update to point at their releases in the future post-link script.

So I was trying to be lazy and avoid doing work that we hopefully won't have to do later for the sake of getting something out that people can use. If you think the size of this package is going to be too much, though, we can start trying to figure that all out.

bgruening:
I think it is too much. Its around 300MB per version and it violates a few of our rules afaik. I'm not even sure `crcmod` belongs into bioconda. I will try to allocate some time and help you with this if you like. I will start with migrating the `crcmod` and use the post-link script if you agree.

recipes/deepvariant/dv_call_variants.py
```diff
@@ -0,0 +1,44 @@
+#!/opt/anaconda1anaconda2anaconda3/bin/python
```
bgruening: This should be `#!/usr/bin/env python`

recipes/deepvariant/dv_call_variants.py
```diff
@@ -0,0 +1,44 @@
...
+#
+# Wrapper script for DeepVariant call_variants
+
+BINARY_DIR="/opt/anaconda1anaconda2anaconda3/BINARYSUB"
```
bgruening: This does not look very portable.

recipes/deepvariant/dv_make_examples.py
```diff
@@ -0,0 +1,57 @@
+#!/opt/anaconda1anaconda2anaconda3/bin/python
```
bgruening: This should be `#!/usr/bin/env python`

bgruening:
The model is also available here as it seems: https://console.cloud.google.com/storage/browser/deepvariant/models/DeepVariant/0.4.0/DeepVariant-inception_v3-0.4.0+cl-174375304.data-wgs_standard?pli=1

chapmanb:
Bjorn;
Thanks for the review and thoughts. I'm open to whatever you think works best but could use this for some work I'm doing so let me know how I can help make it happen.

On `/opt/anaconda1anaconda2anaconda3/`: this gets replaced with the install path during install, and is the trick I use to be able to correctly find the binaries and models. It should be fully portable, and I've used it a few times in the past for jars and other things where we need to resolve to an install directory in share. I don't know any other tricks to be able to correctly reference the final install: is there a better way to do it?

rpoplin:
Hey there, looks like you already found the model in our GCS bucket but just wanted to say that both options are available-- We have two options for getting the pre-built binaries and models.

1.) Download from the GCS bucket with gsutil
gs://deepvariant/binaries
gs://deepvariant/models
- or with curl using a URL like
http://storage.googleapis.com/deepvariant/binaries/DeepVariant/0.4.1/DeepVariant-0.4.1+cl-178668854/make_examples.zip

or 2.) Download from the GitHub release
https://github.com/google/deepvariant/releases/download/v0.4.1/deepvariant.zip

Cheers!

bgruening:
@chapmanb I migrated crcmod to conda-forge. Next step would be a download script for the model.

chapmanb:
Thanks so much for all the help and suggestions. I prepared a clean branch that incorporates all these suggestions and does on-demand download of the model using a post-link script: #7167. I'll close this in favor of that. Please let me know if you have any other suggestions at all.

PR #7150, opened 2017-12-22
Update meta.yaml

ozagordi:
Wrong version. The download file was updated to 2.1.2, but the version field was still 2.1.1

### DIFF
--- a/recipes/minvar/meta.yaml
+++ b/recipes/minvar/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: minvar
-  version: "2.1.1"
+  version: "2.1.2"
 
 build:
   number: 0


recipes/minvar/meta.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: minvar
-  version: "2.1.1"
+  version: "2.1.2"
```
bgruening: Mh, this means the old version was overwritten by the 2.1.2? Can you use a variable name for the version, so that this does not happen again. Thanks.

ozagordi: Yes, the old version was overwritten. My mistake, sorry. Could you please point me to an example, for next time? Is there a consensus now on whether the old versions should be there as directories?
Thanks.

bgruening: Not a consensus, you as package maintainer can decide this. What I was referring is that package for 2.1.1 contains actually the 2.2.2 package at the moment, right? Should this be corrected?

ozagordi: If it’s easier, I would just keep the 2.1.2 containing the correct version package. If not, 2.1.1 should be changed to contain 2.1.1 version and 2.1.2 should be added.

bgruening: That would be nice I guess.

ozagordi: Anything I can do?

bgruening: Can you just create both recipes, so we can also fix the old one?

ozagordi: Ok, I'll do so. After my holidays, though. About to leave. Thanks!

ozagordi: Hi @bgruening.
Just to be sure how to proceed.
I will update the recipe such that package 2.1.1 contains version 2.1.1 (points to the correct release from GitHub). Once this will be merged into bioconda, I will update the recipe such that package 2.1.2 will contain 2.1.2. In this way, bioconda will have both 2.1.1 and 2.1.2 packages, both containing the correct version of the tool. Is this correct?

bgruening: Yes :)

johanneskoester:
We have made changes to our build system (migrated from Travis CI to Circle CI). Please update your fork with the latest upstream/master.

ozagordi:
Closing here and working in #7347 with dir/subdirs structure.

PR #7166, opened 2017-12-29
goatools support py35 and py36

yyminmax:


recipes/goatools/meta.yaml
```diff
@@ -5,40 +5,37 @@ about:
...
+  md5: f2ab989ec9c4acdd80504b263c3b3188
 
 build:
-  skip: True # [not py27]
   number: 0
 
 requirements:
   build:
     - python
-    - setuptools
```
mbargull: `setuptools` is a build requirement. Actually only
```yaml
requirements:
  build:
    - python
    - setuptools
```
is needed for the build.

recipes/goatools/build.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-$PYTHON setup.py install
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
 
 binaries="\
```
mbargull: Those scripts should be installed by `setuptools`, hence `build.sh` can be reduced to only
```
 #!/bin/bash
 
$PYTHON setup.py install --single-version-externally-managed --record=record.txt
 ```

mbargull:
I added some commits, mainly to use `setuptools` for the build and also to exclude the test data from the package as they don't seem to be required for normal operation (but make the package require >100 times more storage space). I hope you don't mind, that I added those changes.

Otherwise your changes look good to me, thanks!

yyminmax:
Thanks for your changes. It looks great.

bgruening:
Cool, thanks a lot!

PR #7181, opened 2018-01-01
add new recipes

cokelaer:

### DIFF
--- a/recipes/bioconvert/build.sh
+++ b/recipes/bioconvert/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+$PYTHON setup.py install 
+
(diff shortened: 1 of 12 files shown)


druvus:
I saw some possible issues with indentation when you tried the extended image. 

Instead of
```
extra:
    container:
    # needs libGL.so.1
    extended-base: true
```

it should be: 
```
extra:
  container:
    # needs libGL.so.1
    extended-base: true
```

Not sure it would help but it could be worth a try.

cokelaer:
ah thanks, I will have a go. The previous build succeeded but there were no imports (all commented) although the standalones worked. I have fixed the indent, and add the tests back. Let us see the outcome

PR #7195, opened 2018-01-04
libmaus2 a new recipe

mictadlo:


mictadlo:
I think I also need https://github.com/bioconda/bioconda-recipes/pull/7196 as a dependency.

recipes/libmaus2/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+    run:
+        - snappy
+        - pkg-config    
+        - libgcc    # [not osx]
+        - zlib {{ CONDA_ZLIB }}*
+
+test:
+  commands:
+    - echo
```
cbrueffer: Something missing here? If there are no tests I think the whole section can be removed.

mictadlo: I tried without a test but lint complained.

daler: @mictadlo a good solution here would be to generate and run a minimal compilation example inside a `run_test.sh` file. This will verify the libs are successfully installed and available.

mictadlo: @daler Could you please provide an example?

daler: Sure, have a look at https://github.com/bioconda/bioconda-recipes/blob/master/recipes/talloc/run_test.sh

mictadlo: I noticed that inside `https://github.com/gt1/libmaus2/tree/master/test` are test scripts. Shoud I add those scripts in the following way to meta.yaml?
```
test:
  commands:
    - test/testAIO.sh
    - testBitIO.sh
    
```

mictadlo:
How is it possible to fix `https://travis-ci.org/bioconda/bioconda-recipes/jobs/328876977#L2706`?

dpryan79:
I see the issue. The shell scripts you reference in the test section are present in the source code, but that no longer exists when it comes to running tests. What you would need to do is to copy those shell scripts from the source code into the recipe directory. However, even then the tests will fail, since they require the source code. My personal suggestion would be to not bother with the testing in the recipe on a library like this (though if there's a `make test` then you might do that during the build). That same strategy is used in the htslib recipe, since the resulting library will end up being used by other recipes, which have their own testing.

PR #7206, opened 2018-01-05
Add recipe for spaln 2.3.1

nathanweeks:
This is an optional dependency for a forthcoming (hopefully) update of braker to version 2.0.1

* [ X] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).


druvus:
@nathanweeks the bioconda guidelines promotes using urls instead of git_urls.  I suggest you are using https://github.com/ogotoh/spaln/archive/a552a0cbe85f7c14a49b37fc211c9f5e4581787e.tar.gz and also add a sha256.

nathanweeks:
@druvus thanks for the edification. I've made the suggested changes to the meta.yaml.

druvus:
It might be a good idea to also include the perl part in the test section to make sure they are working.

nathanweeks:
One of the seqdb/*.pl scripts, makeidx.pl, must be invoked from the seqdb.pl directory (it looks for seqdb/Makefile). I was considering submitting a patch to make it look for the Makefile in the pathname specified by the spaln ALN_DBS environment variable if set and not null (this would also require modifying the seqdb/Makefile to use that environment as an alternative to looking in the current directory for the makblk.pl perl script), but I was unsure if that would diverge too much from the upstream behavior.

Would something like the following suffice?

```
test:
  commands:
    - $PREFIX/seqdb/catchr.pl -? 2>&1 | grep -q 'catchr -hPrefix'
    - $PREFIX/seqdb/makblk.pl | grep -q Usage:
    - test -f $PREFIX/seqdb/makeidx.pl
...
```

druvus:
New version of spaln added so closing.

PR #7207, opened 2018-01-05
Add flye

druvus:

### DIFF
--- a/recipes/flye/build.sh
+++ b/recipes/flye/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+sed -i.bak 's/-rdynamic//' src/Makefile
+python setup.py install  --record record.txt.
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
Cool!

kyleabeauchamp:
Would be good to double check that installing this doesn't clobber an existing installation of minimap2 and/or mappy.

druvus:
Good catch. I missed that it was incorporated since the last abruijn version in bioconda. I will double check it.

druvus:
Flye will install the incorporated minimap2 as flye-minimap2 and mappy is not installed so it should be ok.

kyleabeauchamp:
Looks like there were some very recent bugfixes (https://github.com/fenderglass/Flye/issues/26) that some folks have experienced (e.g., missing config files in the installer).  This definitely impacted your recipe :(

druvus:
@kyleabeauchamp The flye issues should be fixed with #7270

PR #7213, opened 2018-01-07
Bump kaiju to latest version

brooksph:
Modified url to master from latest release to take advantage of recent changes including gzip compatibility. @bgruening Not sure if a tar.gz file is required. 


druvus:
@brooksph It will not be reproducible if you are are using the master tag. It is better practise to access a specific commit, i.e. https://github.com/bioinformatics-centre/kaiju/archive/5557a95f78d04d641e3131c6d6602bbf0c0faa0c.tar.gz

Did you ask the kaiju authors for an updated release?

brooksph:
@druvus Thank you as I had not considered the repoducibility aspect of referencing master here. I've updated the recipe to reference the latest commit. I will bump the kaiju authors to see if they are planning a release anytime soon.

brooksph:
@luizirber any ideas. I can't get this one to pass on macOS. cc @charlesreid1

charlesreid1:
Mac CI build is failing due to failed compilation of zstr in `strict_fstream.hpp`, issue mateidavid/zstr#5 is possibly related. It requires patching zstr.

charlesreid1:
A possible culprit of this error is the `strerror_r` return value:
* the [Mac OS X man page for strerror_r](http://www.manpages.info/macosx/perror.3.html) states: " The strerror_r() function renders the same result into strerrbuf for a maximum of buflen characters and **returns 0 upon success**."
* the [Linux man page for strerror_r](https://linux.die.net/man/3/strerror_r) states: "The GNU-specific strerror_r() **returns a pointer to a string containing the error message**."

The error happens on the subsequent line, when p is passed to another function. The problem seems to be that GNU behavior is expected, but not happening. The compilation command that's failing on the Mac build is a `g++` command, but I would guess this `strerror_r` behavior is not up to the compiler.

pmenzel:
There is a [pull request](https://github.com/mateidavid/zstr/pull/6/commits/1dc84fa4e38970237b03703859e2437adc3803f4) in zstr that aims to fix the strerror_r() problem. 

Could you try, if the change fixes the problem in your CI tests? Then I could integrate it into the zstr header in Kaiju. Unfortunately, I don't have a Mac to try it myself.

charlesreid1:
I have verified the (slightly simpler) pull request mateidavid/zstr#7 will fix the issue, it just requires changing 

```
 #elif (_POSIX_C_SOURCE >= 200112L || _XOPEN_SOURCE >= 600) && ! _GNU_SOURCE 
```

to

```
 #elif (_POSIX_C_SOURCE >= 200112L || _XOPEN_SOURCE >= 600) && ! _GNU_SOURCE || defined(__APPLE__) 
```

and everything is fine.

While the zstr library doesn't seem to be active, its contents are included directly into Kaiju, so a PR to Kaiju should do the trick. I've submitted a PR at bioinformatics-centre/kaiju#60 to fix that in Kaiju. In the meantime, we could shim the `|| defined(__APPLE__)` bit in with a sed script, but I'll defer to folks who know how bioconda usually handles this sort of thing.

Also note, I tested the `-fpermissive` flag but the build still fails with that flag set, so I was incorrect above.

pmenzel:
I merged the PR in kaiju in the latest commit. Could you test if it works, then I could tag it as version 1.6.1

charlesreid1:
@pmenzel Tested and working great! Thanks.

recipes/kaiju/meta.yaml
```diff
@@ -1,25 +1,29 @@
...
+    - zlib {{ CONDA_ZLIB }}*
+
   run:
     - libgcc
     - perl
     - gnu-wget >=1.16
+    - zlib {{ CONDA_ZLIB }}*
 
 test:
   commands:
```
mbargull: Can you change/simplify the test command to 
```yaml
    - kaiju 2>&1 | grep "Kaiju {{ version }}"
```
? This is a common workaround for such cases, hence you can then also remove the comment below.

pmenzel:
I made a new release, [version 1.6.1](https://github.com/bioinformatics-centre/kaiju/releases/tag/v1.6.1)  that contains this commit. thanks for the help!

charlesreid1:
Huh... ci/circleci: lint is failing on this:

```
Setup bioconda-utils00:00

#!/bin/bash -eo pipefail
.circleci/setup.sh

.circleci/common.sh: line 2: syntax error near unexpected token `newline'
Exited with code 2
```

mbargull:
Yeah, I just restarted it. Don't know what happened there. Some kind of fluke...

charlesreid1:
🤦‍♂️ Forgot to upgrade the Kaiju version. Let's give this another go...

recipes/kaiju/meta.yaml
```diff
@@ -12,7 +12,7 @@ source:
 build:
   number: 0  
   no_link:
-    - bin/makeDB.sh -p
+    - bin/makeDB.sh -e
```
mbargull: Please note that `build: no_link:` is just a list of files (or globs), but does not correspond to any command execution.

charlesreid1:
pbj time

PR #7216, opened 2018-01-07
Fix a couple snap issues

nathanweeks:

maker uses hmm-align.pl from snap, and was failing due to hard-coded "#!/usr/bin/perl". This patch:

- replaces the hard-coded "#!/usr/bin/perl" paths with "#!/usr/bin/env perl" in all included perl scripts
- fixes an issue in the generated snap driver script that could cause problems if any arguments were quoted strings containing spaces
- updates the meta.yaml to add some additional regression tests for the perl scripts, add a sha256, and correct the license

Note that cds-trainer.pl and noncoding-trainer.pl fail due to missing Perl libraries that are not currently available in bioconda, but can be obtained from:
https://github.com/KorfLab/Perl_utils


druvus:
I would recommend to package the missing perl_utils in a separate package. I will see if I could assist in doing it later this week.

bgruening:
@druvus what do you mean here? Anything that you would like to before merging?

nathanweeks:
@dpryan79 : It looks like the only change that was made since this PR was submitted was the addition of a sha256 to the meta.yaml (and to clarify, this PR was for *snap*, not *snap-aligner*)

dpryan79:
OK, I'll try to make this mergeable then.

dpryan79:
@bioconda-bot update

dpryan79:
If someone with a mac could look into the test failures on OSX (just run the conda install commands mentioned by the bot for it) then I'd appreciate it. I won't have a chance for a couple days.

nathanweeks:
@dpryan79 : it looks like the osx test passed?

dpryan79:
@nathanweeks I had it skip testing on OSX. Someone with a mac needs to install the binaries the bot posted and see why they fail testing.

dpryan79:
It's completely unclear why the testing fails on OSX. I can install the package on a mac and the tests pass when I manually run them.

nathanweeks:
OK, I'll try to debug next week.

dpryan79:
The OSX failure appears to be due to a segfault in snap itself rather than an issue with something else. Perhaps this is caused by a clang issue, since it's not being thrown on Linux. Given that, it may be best to simply skip building on OSX.

dpryan79:
Let's just skip Linux for now then.

nathanweeks:
OK, strange that snap would segfault on OS X just displaying the usage message---could indicate a latent (on Linux/gcc) bug in the code? Anyway, thanks for tracking that down!

PR #7220, opened 2018-01-08
VEP: bumped subversion

matthdsm:

### DIFF
--- a/recipes/ensembl-vep/meta.yaml
+++ b/recipes/ensembl-vep/meta.yaml
@@ -1,4 +1,4 @@
-{% set version="91.0" %}
+{% set version="91.1" %}
 
 package:
   name: ensembl-vep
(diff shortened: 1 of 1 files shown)


mbargull:
There was a temporary communication error to `anaconda.org` for the corresponding build on `master`. Because of that, the update is not yet available on the channel.
I wanted to simply restart that build, but now we got
```
$ git clone --depth=50 --branch=master https://github.com/bioconda/bioconda-recipes.git bioconda/bioconda-recipes
Cloning into 'bioconda/bioconda-recipes'...
$ cd bioconda/bioconda-recipes
$ git checkout -qf 1ce27bc4f4155cf44570afee6e577eb0733dfa07
fatal: reference is not a tree: 1ce27bc4f4155cf44570afee6e577eb0733dfa07
The command "git checkout -qf 1ce27bc4f4155cf44570afee6e577eb0733dfa07" failed and exited with 128 during .
```
which means the commit is older than the `clone --depth=50`.
@bioconda/core: Should I just submit a new PR to do some non-harming (but reasonable) modification like changing the checksum to `sha256` to trigger a  new build?

daler:
@mbargull sounds good!

PR #7241, opened 2018-01-10
Add bioconda-utils recipe.

johanneskoester:

This is necessary for the reproducibility of our Bioconda paper.

### DIFF
--- a/recipes/bioconda-utils/meta.yaml
+++ b/recipes/bioconda-utils/meta.yaml
@@ -1,0 +1,81 @@
+{% set version = "0.11.0" %}
+
+package:
+  name: bioconda-utils
+  version: {{ version }}
+
+source:
+  fn: bioconda-utils-{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


bgruening:
Restarted the tests.

johanneskoester:
@bgruening bioconda-utils won't install into the mulled container because it needs to be installed in the root environment. As we don't want a mechanism to disable mulled tests, I guess the solution is to provide a hardcoded exception in bioconda-utils itself (disable mulled testing if you build yourself), right?

bgruening:
@johanneskoester You could in theory use run_test files that are currently ignored by mulled. Otherwise, I I can not think of a better way, currently.

bgruening:
Restarted with latest changes from Ryan.

daler:
I just added a new v0.11.1 release with the updates and updated this recipe, so that the built package can theoretically build itself.

bgruening:
Super cool!

johanneskoester:
Thanks @daler for taking care of this!

PR #7288, opened 2018-01-17
FastQC - Add fonts to fix biocontainer on quay.io 

brooksph:
* [X] This PR does something else (explain below).\

I tried to run the biocontainer created with this recipe and got the following error: 
```
$ docker run -it -v $PWD:/data quay.io/biocontainers/fastqc:0.11.6--pl5.22.0_0
# cd /data
# fastqc example.fq.gz -o /tmp
Started analysis of example.fq.gz
Analysis complete for example.fq.gz
Exception in thread "Thread-1" java.lang.Error: Probable fatal error:No fonts found.
	at sun.font.SunFontManager.getDefaultPhysicalFont(SunFontManager.java:1236)
```

I tried to figure out how the container was built but it seems that the biocontainer on Dockerhub works but is different than the one on quay.io (this is the one missing the font). 

cc @bgruening @luizirber

### DIFF
--- a/recipes/fastqc/meta.yaml
+++ b/recipes/fastqc/meta.yaml
@@ -15,7 +15,7 @@ source:
 
 build:
   detect_binary_files_with_prefix: true
-  number: 0
+  number: 1
 
 requirements:
   run:
--- a/recipes/fastqc/meta.yaml
+++ b/recipes/fastqc/meta.yaml
@@ -21,6 +21,7 @@ requirements:
   run:
     - openjdk
     - perl
+    - font-ttf-dejavu-sans-mono
 
 test:
   commands:


bgruening:
I think this was due to the change from defaults to conda-forge. I think the defaults Java package does not have the fonts installed. Hopefully this solved it. Thanks.

bgruening:
Thanks @brooksph!

caseywdunn:
Hi all - I just created a recipe for the most recent fastqc version (https://github.com/bioconda/bioconda-recipes/tree/master/recipes/fastqc), and ran into some issues with the `font-ttf-dejavu-sans-mono` dependency that was added. Since that is from the default channel, it then forces openjdk to install from the default channel rather than the conda-forge channel. That leads to a variety of other problems, and is also inconsistent with the guidelines which indicate openjdk should be installed from conda-forge (http://bioconda.github.io/guidelines.html). In the recipe for the new version I therefore reverted to not including the font dependency. 

Is there another way to add the needed fonts on the systems that don't have them that doesn't change the channel that opensdk is installed from?

PR #7304, opened 2018-01-18
Add Clinker package

lonsbio:

### DIFF
--- a/recipes/clinker/meta.yaml
+++ b/recipes/clinker/meta.yaml
@@ -1,0 +1,35 @@
+{% set name = "clinker" %}
+{% set version = "1.31" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  fn: "v{{ version }}.tar.gz"
+  url: https://github.com/Oshlack/Clinker/archive/v{{ version }}.tar.gz
+  md5: "5487f99af6b9499c87d4d945e9cd9790"
+
+build:
+  number: 0
+
+requirements:
+  run:
+    - samtools
+    - bpipe
+    - star 2.5.3a.*
+    - python 2.7.*
+
+
+
+
+test:
+  commands:
+    - bpipe --help
+
+about:
+  home: https://github.com/Oshlack/Clinker
+  summary: "test"
+  license: GPL3
+ 
+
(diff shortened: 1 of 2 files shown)


recipes/clinker/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+build:
+  number: 0
+
+requirements:
+  run:
+    - samtools
+    - bpipe
+    - star 2.5.3a.*
+    - python 2.7.*
```
bgruening: python should not be pinned like this. Please use the build/skip section for this like `skip: True  # [py3k]`

recipes/clinker/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  md5: "5487f99af6b9499c87d4d945e9cd9790"
+
+build:
+  number: 0
+
+requirements:
+  run:
+    - samtools
+    - bpipe
+    - star 2.5.3a.*
```
bgruening: Is this strict star pinning necessary?

recipes/clinker/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+
+
+test:
+  commands:
+    - bpipe --help
+
+about:
+  home: https://github.com/Oshlack/Clinker
+  summary: "test"
```
bgruening: Can you please add here a slightly more verbode summary :)

recipes/clinker/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  license: GPL3
```
bgruening: license_family: GPL

bgruening: license_file: LICENSE

recipes/clinker/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+    - bpipe
+    - star 2.5.3a.*
+    - python 2.7.*
+
+
+
+
+test:
+  commands:
+    - bpipe --help
```
bgruening: This does not look like a clinker test :)

lonsbio: Clinker is invoked via bpipe - but we may need to write a wrapper.

lonsbio:
Thanks @bgruening and apologies  - we're being introduced to creating BioConda packages with @Slugger70  and I got a little ahead of things! Will fix things up.

PR #7308, opened 2018-01-18
Snakemake 4.5.0

johanneskoester:


cbrueffer:
Hi Johannes,

```python-drmaa``` is needed to make DRMAA submissions work; do you think it makes sense to add it as a run dependency here, or do you rather consider it a dependency of the cluster environment?

johanneskoester:
So, the dependencies are chosen in a way that optional stuff is only included if it might be needed for certain snakefiles (e.g. the remote file stuff). Everything that is best decided by installation (e.g., whether you want to have DRMAA) should remain optional. Although I agree that the overhead would be minimal.

johanneskoester:
Usually, I would rather say this should not at all be denoted together with a workflow, because a workflow should be independent of the underlying platform. Instead, somebody installing Snakemake for usage in a cluster should rather install it together with python-drmaa, independent of the workflow.

Issue #3496, opened 2017-01-16
bioconda-utils hangs on test section


sebastian-luna-valero:
Hello,

My recipes are hanging on the test part of the recipe:

https://github.com/bioconda/bioconda-recipes/pull/3491
https://travis-ci.org/bioconda/bioconda-recipes/builds/192344798

https://github.com/bioconda/bioconda-recipes/pull/3493
https://travis-ci.org/bioconda/bioconda-recipes/builds/192346796

Running `simulate-travis.py` locally I can see that the execution hangs on the test section:
```
INFO bioconda_utils.build:build(124): BIOCONDA TEST START via mulled-build recipes/cgat-pipelines-nosetests, CONDA_BOOST=1.60;CONDA_GMP=5.1;CONDA_GSL=1.16;CONDA_HDF5=1.8.17;CONDA_NCURSES=5.9;CONDA_NPY=110;CONDA_PERL=5.22.0;CONDA_PY=27;CONDA_R=3.3.1
updating index in: /sebastian/i35/conda-bld/linux-64

DEBUG bioconda_utils.pkg_test:test_package(97): Tests to run: python --version && R --version && python -c "import numpy" && python -c "import pandas" && python -c "import pysam" && python -c "import pybedtools" && python -c "import rpy2"

DEBUG bioconda_utils.pkg_test:test_package(121): No container push commands are generated

DEBUG bioconda_utils.pkg_test:test_package(122): mulled-build command: ['mulled-build', 'build-and-test', 'cgat-pipelines-nosetests=0.0.1--py27r3.3.1_0', '-n', 'biocontainers', '--test', 'python --version && R --version && python -c "import numpy" && python -c "import pandas" && python -c "import pysam" && python -c "import pybedtools" && python -c "import rpy2"', '--extra-channels', 'file://sebastian/i35/conda-bld,r,defaults,conda-forge']
```

Has something changed recently in that regard in the build system?

Many thanks,
Sebastian

bgruening:
Not recently, but since a few month we are testing packages in a busybox Docker container. It seems you package is not running in this environment, which indicates that you are missing something ... or at least something is wrong.

sebastian-luna-valero:
That's strange. Even when I remove the test section completely, the `simulate-travis.py` execution hangs on:
```
[...]

PR #7311, opened 2018-01-18
Remove unnecessary gcc build/run dependencies

nathanweeks:

When attempting to update a perl recipe that depends on this one (https://github.com/bioconda/bioconda-recipes/pull/7261), the mulled-build failed when installing gcc as previously required to build/run this recipe, apparently related to issue https://github.com/bioconda/bioconda-recipes/issues/3496. However, as no compiler is actually needed to build/run this perl module, I propose removing such dependencies from the recipe.

### DIFF
--- a/recipes/perl-text-balanced/build.sh
+++ b/recipes/perl-text-balanced/build.sh
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-cpanm -n -f -i .
+HOME=/tmp cpanm -i .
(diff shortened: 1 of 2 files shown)


bgruening:
Good catch, gcc should never be a build dependency.

foreveremain:
(Jumps in...) Any reason why the macos Travis build is stuck for this PR ?  Seems exceptionally long.

nathanweeks:
@foreveremain : there seems to be a big backlog for all macOS Travis builds...

foreveremain:
On 25/01/2018 13:28, Nathan T. Weeks wrote:
> @foreveremain <https://github.com/foreveremain> : there seems to be a
> big backlog for all macOS Travis builds...
argh - yes - thought I found the one that was stuck - here's hoping it clears soon !

bgruening:
Thanks @nathanweeks!

PR #7314, opened 2018-01-18
Bump MultiQC to 1.4

rasmusagren:


daler:
@rasmusagren see #7252. We're unfortunately still waiting on a MacOS build for that PR due to the recent travis-ci hardware problems.

rasmusagren:
Ah, didn't see that one. Then I'll close this one.

The problem is that 1.3 doesn't work as it is now, probably due to mismatching dependencies after conda-forge colormath was updated. Should I make a PR for 1.3, or will that also have Travis issues?

@ewels Phil, could you maybe look at this?

rasmusagren:
@ewels Looked at your colormath PR at conda-forge, and the problem (I think) comes from that networkx>=2.0 should be used as run requirement.

ewels:
@rasmusagren - in the colormath PR you mean? I think to patch 1.3 of MultiQC we just need to specify an older version of `spectra` (before they started requiring networkx > 2). Basically pin those three packages to the versions where they worked ok.

@daler - is it worth prepping and merging a patched version of the v1.3 release that works, before releasing v1.4? Then we'll have working archived versions without needing to use subdirectories. It should be a minor change (just pinning versions) so hopefully quick & easy once Travis is working again..

rasmusagren:
Yes, I was thinking that it would be easier to change at conda-forge since building here seems problematic now. Colormath 3.0.0 requires networkx>=2.0, but in your PR it just states networkx. If that was changed, conda would install the old version of colormath so as not to break the networkx<2 requirement for MultiQC. Maybe there is a similar issue for the spectra PR?

ewels:
I guess you're looking at my first colormath PR, but there's been another since then and the current colormath recipe does specify networkx >=2.0 now: https://github.com/conda-forge/colormath-feedstock/blob/master/recipe/meta.yaml#L27

The tracebacks I've seen have been due to MultiQC v1.3 asking for networkx <2 but not specifying which spectra package. The latest spectra package is therefore installed, which then asks for networkx >2 (or ==2 maybe). These conflicting requirements then cause a crash.

rasmusagren:
Ah, I didn't see that one. Still, if MultiQC specifies networkx <2 then it shouldn't also have to specify spectra version (assuming that the spectra recipe has requirements set correctly). Maybe it works now after your last PRs have been accepted. I'll try again.

daler:
@ewels v1.3 first and then v1.4 sounds fine for keeping the recipe dir clean. @rasmusagren conda-forge also uses travis-ci for their MacOS builds, so unfortunately it won't be any faster there. We're currently looking at CircleCI as an alternative, but that will require rewriting some of our infrastructure so that also won't be a fast solution.

mbargull:
If `conda` picks up the newer versions regardless, then that is due to a known bug in `conda` which still isn't fixed (and for which I can't give any ETA). Then the `colormath=3.0.0` with build number `0` will have to be labelled as `broken` on `conda-forge`.

mbargull:
To clarify: That "bug" in `conda` is something *I* consider a *bug* but is actually a (IMHO, very bad) conceptual decision that lets `conda` consider any previous builds of packages instead of only choosing the one with the highest number (on default).
I just tried it out with the updated `colormath=3.0.0` uploaded: A `conda create -nx multiqc=1.3` will (IMO, wrongly) pick `colormath=3.0.0=py_1` 😒. The right fix would be to mark the `colormath=3.0.0=*_1` builds as broken on `conda-forge`. Changing `multiqc=1.3` would only be a bad workaround, IMHO.

ewels:
Ah crap, so this is due to me messing up the first colormath 3 conda-forge package? How do we mark it as broken? Is that something I can do?

mbargull:
There are a couple of "non-optimal" things at play -- I wouldn't just blame yourself 😉.
You would need access to the Conda channel. I don't know who exactly you should ask, but some one from @conda-forge/core (<-- https://github.com/orgs/conda-forge/teams/core/members; GitHub doesn't render team mentions for other organisations..), should be able to help.

ewels:
Ok, asked: https://github.com/conda-forge/colormath-feedstock/issues/5

ewels:
Argh, now I'm getting another new installation error message! 😱 

```
error: networkx 2.1rc1 is installed but networkx==2.0 is required by {'spectra'}
```
https://travis-ci.org/ewels/MultiQC/builds/330352701

Now that the packages themselves are hopefully all doing sensible things, am I right in thinking that I can list _just_ `spectra` in the MultiQC installation requirements and let the packaging managers sort everything else out themselves? Will that solve all of our problems? I'm pretty sure that I only added `networkx` and `colormath` to the MultiQC requirements because they weren't specifying the necessary versions at the time.

What a nightmare!

mbargull:
If MultiQC directly requires those packages, it should depend on them.
If they are only required for other dependencies to work properly, the packages for those dependencies must depend on them, but MultiQC should have to.
Only for some very special cases it would make sense to directly include actual transitive dependencies.

If you think the `networkx==2.0` requirement for `spectra=0.0.10` is too restrictive and should rather be `networkx>=2` or something like `networkx>2,<3.0a0`, you should file an issue upstream at https://github.com/jsvine/spectra.

ewels:
Yes exactly - it was a very special case before because they didn't have the required dependencies themselves after `networkx` `2.0` came out. So that's what I was suggesting, that we can remove them from MultiQC now and everything should be easier.

Request for less stringent requirements in spectra here: https://github.com/jsvine/spectra/issues/12

rasmusagren:
MultiQC 1.3 works now after they labelled the erroneous colormath builds as
broken.

Den 20 jan. 2018 10:54 fm skrev "Phil Ewels" <notifications@github.com>:

PR #7316, opened 2018-01-18
[BUG] Fix hdf5 version for blasr

jamestwebber:
blasr seems to require 1.8.17 to run, otherwise it complains about a missing library. It's possible there is a way to resolve this via their code, but in the meantime this might fix it.


dpryan79:
Honestly the version for blasr should just be updated and pinned to an hdf5.

dpryan79:
This should no longer be an issue as we now pin against hdf5 for everything.

PR #7329, opened 2018-01-21
Circleci builds

johanneskoester:
This PR enables testing via CircleCI, which promises to be much faster. We have received free MacOS and Linux builds from them. Thank you so much, guys!

# TODO

The following things will have to be done before we can merge this PR.

* [x] Add ANACONDA_TOKEN, INVOLUCRO_AUTH, QUAY_OAUTH_TOKEN to CircleCI settings before merging.
* [x] Test a real build. I will artificially increase bwa build number for this.
* [x] Disable travis. This is unavoidable. Otherwise, we could have problems with both services trying to upload.
* [x] Add a scheduled workflow that grabs missed builds on master branch every night.


bgruening:
I added the ENV vars.

daler:
I've been reading up on CircleCI, got through all the docs, got confused because the circleci.yml here was so different, then realized there's a v2 docs! Made it through and finally understand the setup. This is really cool. 

I have a question about the caching. As I understand the current config, a cache hit will bring over the latest conda, then run the setup, then re-save it. However that setup script does conda-related stuff only if `/tmp/workspace/miniconda` is missing, which means that we effectively use exactly the last cached installation. Great for speed, but if any package versions are updated in the channels, we need to wait for a cache miss to update them. As currently configured, this means bumping the bioconda-utils tag or miniconda version, both of which will change less frequently than channel package updates.

Is that interpretation correct? If so, would it make sense to have:

```bash
/tmp/workspace/miniconda/bin/conda update --all
```
in `.circleci/setup.sh` that would always run? That way we would get the benefit of the cache for the majority of packages, but conda would take care of updating any packages that changed since the last cache.

johanneskoester:
@daler, yes, we could do that. I opted for not doing it, because then we essentially loose the benefit of pinning a miniconda release. This way, we would always be up-to-date. However, the effect is little, because conda-build should fetch the latest software for all relevant dependencies anyway, shouln'd it? So the effect of an update would only be that the environment of bioconda-utils is newer, but for this we anyway pin certain versions.

mbargull:
Technically, we don't pin anything but only install specific versions. But I think we should use pinning, i.e., in `simulate-travis.py` not only do
```bash
conda install --file bioconda_utils-requirements.txt
```
but also
```bash
cp bioconda_utils-requirements.txt "${CONDA_ROOT}"/conda-meta/pinned
```
, otherwise a `conda update --all` would update `conda`, `conda-build`, ...

johanneskoester:
@mbargull exactly, and if we do that, we essentially can also skip the update, because it would just affect packages that we do not need for bioconda-utils at all.

.circleci/config.yml
```diff
@@ -0,0 +1,132 @@
+version: 2
+
+variables:
+  restore_cache: &restore_cache
+    restore_cache:
+      keys:
+        - miniconda-{{ .Environment.MINICONDA_VER }}-{{ .Environment.BIOCONDA_UTILS_TAG }}-{{ arch }}
```
mbargull: Those `.Environment.*` variables are specified outside the repo at CircleCI?
Currently we have `BIOCONDA_UTILS_TAG == 'master'`. I think it would be better do use the commit rather than the tag. Otherwise we'd reuse and outdated cache when `bioconda-utils` has been changed since the last `save_cache`.
If we aren't able to directly include the commit's checksum, maybe it's possible to workaround that with something like
```bash
run:
  command: 'git ls-remote https://github.com/bioconda/bioconda-utils.git {{ .Environment.BIOCONDA_UTILS_TAG }} >bioconda-utils-tag-commit'
restore_cache:
  keys:
    - miniconda-{{ .Environment.MINICONDA_VER }}-{{ checksum "bioconda-utils-tag-commit" }}-{{ arch }}
```
?

.circleci/config.yml
```diff
@@ -0,0 +1,132 @@
...
+          command: bioconda-utils build recipes config.yml --git-range master HEAD --docker --mulled-test
+
+  # build and test on macos
+  test-macos:
+    <<: *macos
+    steps:
+      - checkout
+      # We cannot cache the miniconda installation on macos.
+      # The reason is that tar seems to have a bug when handling sym and hardlinks in the miniconda installation.
+      # The error is, e.g.: Cannot extract through symlink tmp/workspace/miniconda/LICENSE.rst
```
mbargull: If this is due to hard-/symlinks created by the `conda install` command, we could try `conda install --copy` during the setup. If this happen because of files/links Miniconda itself already installed, then nevermind.

.circleci/config.yml
johanneskoester: Good point. Maybe we should even properly tag bioconda-utils releases again. This should also increase the reliability. I will move to the latest release for now.

.circleci/config.yml
johanneskoester: I think it is both, see [here](https://circleci.com/gh/bioconda/bioconda-recipes/73?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

We could still try to manually copy those symlinks away before the cache is created.

mbargull: Ah, thanks for pointing me at the log. I think it's something non-Conda related as it even shows those errors for the cached `repodata.json` files which are very unlikely to be linked somewhere else:
```
tmp/workspace/miniconda/pkgs/cache/: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache
tmp/workspace/miniconda/pkgs/cache/08159f17.json: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/08159f17.json
tmp/workspace/miniconda/pkgs/cache/09cdf8bf.json: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/09cdf8bf.json
tmp/workspace/miniconda/pkgs/cache/09cdf8bf.q: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/09cdf8bf.q
tmp/workspace/miniconda/pkgs/cache/0cdb2265.json: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/0cdb2265.json
tmp/workspace/miniconda/pkgs/cache/0cdb2265.q: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/0cdb2265.q
tmp/workspace/miniconda/pkgs/cache/2a957770.json: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/2a957770.json
tmp/workspace/miniconda/pkgs/cache/2a957770.q: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/2a957770.q
tmp/workspace/miniconda/pkgs/cache/31ce02e0.json: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/31ce02e0.json
tmp/workspace/miniconda/pkgs/cache/31ce02e0.q: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/31ce02e0.q
tmp/workspace/miniconda/pkgs/cache/35adf087.json: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/35adf087.json
tmp/workspace/miniconda/pkgs/cache/35adf087.q: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/35adf087.q
tmp/workspace/miniconda/pkgs/cache/716b88c0.json: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/716b88c0.json
tmp/workspace/miniconda/pkgs/cache/87d00fe333551440c29e16192807b6e6.png: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/87d00fe333551440c29e16192807b6e6.png
tmp/workspace/miniconda/pkgs/cache/9779607c273dc0786bd972b4cb308b58.png: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/9779607c273dc0786bd972b4cb308b58.png
tmp/workspace/miniconda/pkgs/cache/9ca791dd.json: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/9ca791dd.json
tmp/workspace/miniconda/pkgs/cache/9ca791dd.q: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/9ca791dd.q
tmp/workspace/miniconda/pkgs/cache/abae4451.json: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/abae4451.json
tmp/workspace/miniconda/pkgs/cache/abae4451.q: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/abae4451.q
tmp/workspace/miniconda/pkgs/cache/d85a531e.json: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/d85a531e.json
tmp/workspace/miniconda/pkgs/cache/d85a531e.q: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/d85a531e.q
tmp/workspace/miniconda/pkgs/cache/df7feebede9861a203b480c119e38b49.png: Cannot extract through symlink tmp/workspace/miniconda/pkgs/cache/df7feebede9861a203b480c119e38b49.png
```
So maybe a `/tmp` is symlinked in CircleCI's macOS setup?

.circleci/config.yml
```diff
@@ -0,0 +1,132 @@
...
+      key: miniconda-{{ .Environment.MINICONDA_VER }}-{{ .Environment.BIOCONDA_UTILS_TAG }}-{{ arch }}
+      paths:
+        - /tmp/workspace/miniconda
+  setup: &setup
+    run: 
+      name: Setup bioconda-utils
+      command: .circleci/setup.sh
+  macos: &macos
+    macos:
+      xcode: "9.0"
```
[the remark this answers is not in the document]
johanneskoester: This was just for testing, but I forgot to determine the minimum. Good catch.

.circleci/config.yml
johanneskoester: Done.

bgruening:
@johanneskoester what about using bioconda-legacy or a test repo to submit from circle-ci and use travis and circle-ci for a week in parallel?

johanneskoester:
This could work. On the other hand, Travis seems almost unusable for us right now. Hence, even if it continues to be there over the next week, it cannot really serve as a backup for circleci. Furthermore, once circleci is activated in all PRs they cannot be merged anyway if something on circleci fails. Once bioconda/bioconda-utils#253 is merged, we can also be sure that building, testing and uploading works fine. Then, the only risk is that something on the master or bulk branch does not work as expected. And that can be rescued pretty easily by the nightly job that builds and uploads all packages that are not yet there.

johanneskoester:
@bioconda/core this is now merged. I will now deactivate travis and notify all PRs to update to the latest master.

chapmanb:
Wow, thanks Johannes. This is an amazing turnaround time for swapping to a new CI systems. Fingers crossed on Circle CI stability.

johanneskoester:
Yes, I am a bit nervous :-). Feel free to ping me in our core chat on gitter if you see any problems.

nsoranzo:
It seems there are some other files we can remove now, e.g. `scripts/travis-run.sh`, `scripts/travis-setup.sh` and `scripts/travis-wait.sh` .

bgruening:
Lets keep this files for a little bit more time. We still need to see how this scales - also in regards to our free plan. @johanneskoester is on contact with CCI.
I would like to keep these files for a month or so and see how everything goes, so that we can easily switch back if we need it.

nsoranzo:
@bgruening Everything will be forever in the git repository, what's the point?

PR #7347, opened 2018-01-23
Correct MinVar version 2.1.1

ozagordi:


cbrueffer:
Could you be more specific what this fixes and how?

Versions generally shouldn't go backwards; if something needs to be fixed in an earlier package I think the better way is to add a 2.1.1 subdirectory with the corrected 2.1.1 recipe.

ozagordi:
2.1.1 was already present, but the meta.yaml pointed to the wrong tarball. As a consequence, the bioconda package 2.1.1 actually contains the wrong version. The reason for this new build is to correct that mistake. This is what I understood from @bgruening, but I might have misunderstood.

cbrueffer:
Makes sense! I still think this should go into a separate subdir though, otherwise this gets very confusing for the next person doing an update.

ozagordi:
I’m not sure I understand how. One subdir for each old version? But how would bioconda build then?

cbrueffer:
Well, one for each old version you want to touch. An example is Picard: https://github.com/bioconda/bioconda-recipes/tree/master/recipes/picard

ozagordi:
I can organise the recipe like this, but I was under the impression that the subdirs are not built, i.e. they are just a way to save the old versions. Am I wrong? Further, I already opened #7355 to update 2.1.1. How should I proceed?

cbrueffer:
Subdirs are built if something in them changes that requires a rebuild (e.g. build number increase). In principle you can fix both versions in this PR and just close the other one (or the other way around).

PR #7364, opened 2018-01-26
updated secapr

tobiashofmann88:


recipes/secapr/meta.yaml
```diff
@@ -1,28 +1,28 @@
-{% set version = "1.1.0" %}
+# build with: conda build --python 2.7 .
 
 package:
   name: secapr
-  version: {{ version }}
+  version: '1.1.4'
```
cbrueffer: Any reason for reverting the variable usage and other cleanups that @druvus committed?

tobiashofmann88: actually no, I'm just seeing this now, I had an older version of the meta.yaml on my computer which i updated and then uploaded. That version didn't have the variable in it...
But I might just leave it like this for now, unless there is a reason to change it back?

cbrueffer: Well, it would be good to keep the improvements made. Looking at this the update diff should be as small as exchanging the version and hash.

tobiashofmann88: it is important for me to get this submitted asap, since I submitting a manuscript about this pipeline today and the reviewers need access to the newest verison. I will re-implement these changes in the next update, promise ;-)

bgruening: Please, fix this. We will merge today promised if you fix it ;)

johanneskoester:
@tobiashofmann88 ready to merge, just waiting on your fixes.

cbrueffer:
I opened a cleaned up PR in #7378 so we can get this in quickly while keeping the recipe clean, I hope that works for you @tobiashofmann88

cbrueffer:
I've merged PR #7378.

tobiashofmann88:
thanks for your help and sorry for not responding earlier!

PR #7381, opened 2018-01-26
add umitools

kepbod:

### DIFF
--- a/recipes/umitools/build.sh
+++ b/recipes/umitools/build.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+$PYTHON setup.py install
+
+# Add more build steps here, if they are necessary.
+
+# See
+# http://docs.continuum.io/conda/build.html
+# for a list of environment variables that are set during the build process.
+
--- a/recipes/umitools/meta.yaml
+++ b/recipes/umitools/meta.yaml
@@ -1,0 +1,42 @@
+package:
+  name: umitools
+  version: "0.3.0"
+
+source:
+  fn: umitools-0.3.0.tar.gz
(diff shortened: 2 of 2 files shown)


bgruening:
@kepbod do you know this one? https://github.com/bioconda/bioconda-recipes/blob/master/recipes/umi_tools/meta.yaml

I foresee confused users for both packages :(

kepbod:
I don't know umi_tools. umitools is a different one. Do I need to rename it?

yfu:
Hi Björn,

I am the owner of umitools and I can confirm that my umitools is different from umi_tools.

Thanks!

bgruening:
@kepbod @yfu I know. I'm just wondering if we should choose a prefix or something to distinguish both tools better. I fear that people easily install the wrong tool.

kepbod:
@bgruening Do you have some suggestions? In PyPI, [umitools](https://pypi.python.org/pypi/umitools/0.3.0) and [umi_tools](https://pypi.python.org/pypi/umi_tools/0.5.3) both exist. It would make people more confusing if we rename them with a new name inconsistent with the name in PyPI.

bgruening:
No good idea here, sorry. Both projects are just too similar with the (nearly) same name. @yfu any change you rename the entire project? I have seen it is quite young? This is also just a recommendation, we don't have a strict guideline for this. My feeling is just, that also your users could be confused if they install the wrong version.

cbrueffer:
I agree the naming similarity is very unfortunate, but both projects seem to be well established by now (and both were started quite a while ago)...

cbrueffer:
I'd like to get this in, If no one sees a striking reason not to. The naming issue is unfortunate, but I think that's something for the upstream projects to sort out.

bgruening:
Ok :)

PR #7382, opened 2018-01-27
Porechop

jvolkening:

### DIFF
--- a/recipes/porechop/meta.yaml
+++ b/recipes/porechop/meta.yaml
@@ -1,3 +1,9 @@
+# NOTE: Currently we require a modified version of Porechop with downgraded
+# SeqAn to compile with the latest GCC available in conda (4.8).
+#
+# When GCC 4.9 or later is available on conda, the source metadata should be
+# updated to point to the latest mainstream Porechop release.
+
 package:
   name: porechop
   version: '0.2.3_seqan2.1.1'
--- a/recipes/porechop/meta.yaml
+++ b/recipes/porechop/meta.yaml
@@ -1,0 +1,31 @@
+package:
+  name: porechop
+  version: '0.2.3_seqan2.1.1'
+
+source:
+  fn:  Porechop-0.2.3_seqan2.1.1.tar.gz
+  url: https://github.com/jvolkening/Porechop/archive/v0.2.3_seqan2.1.1.tar.gz
+  md5: 648fbf44b8fe70aa771786b5180b4e22
+
+build:
+  number: 0
+  skip: True # [py27]
+
+requirements:
+  build:
+    - python >=3.4
+    - gcc  # [not osx]
+    - llvm # [osx]
+
+  run:
+    - python >=3.4
+    - libgcc # [not osx]
+
+test:
+  commands:
+    - porechop -h
+
+about:
+  home: https://github.com/rrwick/Porechop
+  license: GPL3
+  summary: Adapter removal and demultiplexing of Oxford Nanopore reads
--- a/recipes/porechop/build.sh
+++ b/recipes/porechop/build.sh
@@ -1,0 +1,1 @@
+$PYTHON setup.py install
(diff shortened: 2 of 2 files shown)


recipes/porechop/meta.yaml
```diff
@@ -0,0 +1,31 @@
+package:
+  name: porechop
+  version: '0.2.3_seqan2.1.1'
```
bgruening: I'm wondering if you want to encode this in the build-string?

recipes/porechop/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+source:
+  fn:  Porechop-0.2.3_seqan2.1.1.tar.gz
+  url: https://github.com/jvolkening/Porechop/archive/v0.2.3_seqan2.1.1.tar.gz
```
bgruening: @jvolkening any reason you are using your fork and the the original repo?

recipes/porechop/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  md5: 648fbf44b8fe70aa771786b5180b4e22
+
+build:
+  number: 0
+  skip: True # [py27]
+
+requirements:
+  build:
+    - python >=3.4
+    - gcc  # [not osx]
```
bgruening: I'm wondering why seqan is not a dependency, and if its only shipping header, if you really need to encode it in the version string?

recipes/porechop/meta.yaml
jvolkening: I'm not sure -- see below.

recipes/porechop/meta.yaml
jvolkening: See below for a general discussion. I'm hoping that Ryan will accept this as a stable branch parallel to master in his repo. When that happens, I will change the source to point to upstream. I'm probably being a little impatient, but I really want to get this into bioconda so that I can polish up the Galaxy wrapper and push it from testing to the main Galaxy toolshed (we have people using it in production and in soon-to-be-published workflows).

recipes/porechop/meta.yaml
jvolkening: The underlying problem here is the general one of conda not having a recent GCC supporting `c++14`. To compile Porechop for bioconda, it is therefore necessary to downgrade SeqAn to the last release buildable with GCC 4.8, which is SeqAn 2.1.1.

The only difference between my forked branch and mainstream Porechop is the downgraded SeqAn. That is why I encode this information in the version string -- it doesn't seem quite right to just use the version from upstream. If you have another suggestion I'm happy to listen.

As for making SeqAn a dependency, that's not how it's done in upstream Porechop so I didn't see a reason to change it. The alternative would be to modify upstream to make SeqAn a dependency. Either way it requires a modified version.

Feel free to push whatever modifications you wish to this. I'm just trying to find a way to make it work. Hopefully a recent GCC will make it into conda and then all of this can go away.

jvolkening:
Also, @bgruening, it seems that `$PYTHON -m unittest` would be better in the test section of the YAML, but I was unable to get it to work from there. It doesn't fail but reports running zero tests, which is why I moved it to the build script. I guess these test commands don't run from the build directory? If you have any suggestions on this, please advise.

bgruening:
> It doesn't fail but reports running zero tests, which is why I moved it to the build script. I guess these test commands don't run from the build directory? If you have any suggestions on this, please advise.

This is true, it installed the packages and runs it from a test environment, which does not have the build-directory available. So running it from within the build.sh is completely fine.

recipes/porechop/meta.yaml
bgruening: Ok, got it. Can you please write this into the recipe as a comment. Also that the URL needs to be changed as soon as we get a new compiler?

bgruening:
Thanks!

Issue #1332, opened 2016-04-15
Using llvm in OSX builds


acaprez:
I'm trying to understand the motivation for specifying `llvm` as opposed to `gcc` for OS X builds.  Best I can tell, the conda `llvm` package doesn't include the clang compiler.  It's just the Core part of llvm with the library and utilities.  So when you specify `llvm`, what will really get used is the Xcode-provided clang (and llvm) that is symlinked to `/usr/bin/gcc`.  I'm just curious if that was the intention or not.

yesimon:
It does look like a mistake, and should be removed.

I think it's yet to be decided whether we are using clang or gcc for OS X builds in the future. One thing to note is that conda-recipes uses clang which would be my vote.

johanneskoester:
Thanks for pointing this out. Indeed, the intention was to follow anaconda.org and conda-recipes and build using clang on osx. However, I am personally also fine with using the real gcc on osx, if that does not introduce incompatibilities with anaconda.org or conda-recipes.

johanneskoester:
@bioconda/core we should collect some opinions and decide how to proceed.

croth1:
I think it may make sense to check the policies of the conda-forge community project and see if their recipe format fits for us as well.

yesimon:
I think we should use built-in clang and default travis OS X setup for max compatibility with `conda-recipes` and `conda-forge`. 

I was thinking it could be a good idea to provide an alternative channel such as `bioconda-gcc-osx` for gcc builds on OS X.

molpopgen:
The OS X/GCC thing came up for me while prepping PR #2682, which is a C++11 package.  In short, the issues are:

1. The built-in clang on OS X was not feature-complete for C++11 until 10.9  or so.
2. The built-in clang on the Travis install is simply too old, resulting in "unrealistic" failures (there just can't be that many 10.5/10.6 machines still in use).
3. Ignoring the Travis issue, clang on OS X Yosemite removed built-in support for OpenMP, which is a problem for a lot of scientific computing.  Some uses of OpenMP further don't easily support conditional compilation to get around this--I'm thinking Cython here. For HomeBrew users, one can install clang-omp to band-aid the problem.

Using GCC for OS X builds addresses all of these issues: old and new OS X targets are supported, as is -fopenmp.

I'm not sure there's a completely perfect answer here, but GCC does check of a lot of boxes.

bgruening:
[...]

PR #7388, opened 2018-01-28
Add bioconductor-ffpe.

cbrueffer:

New PR since #6792 can't be re-opened.

### DIFF
--- a/recipes/bioconductor-ffpe/build.sh
+++ b/recipes/bioconductor-ffpe/build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+mv DESCRIPTION DESCRIPTION.old
+grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
+$R CMD INSTALL --build .
--- a/recipes/bioconductor-ffpe/meta.yaml
+++ b/recipes/bioconductor-ffpe/meta.yaml
@@ -1,0 +1,52 @@
+{% set version = "1.22.0" %}
+{% set name = "ffpe" %}
+{% set bioc = "3.6" %}
+
+package:
+  name: 'bioconductor-{{ name|lower }}'
+  version: '{{ version }}'
+
+source:
+  fn: '{{ name }}_{{ version }}.tar.gz'
+  url:
+    - 'http://bioconductor.org/packages/{{ bioc }}/bioc/src/contrib/{{ name }}_{{ version }}.tar.gz'
+    - 'https://bioarchive.galaxyproject.org/{{ name }}_{{ version }}.tar.gz'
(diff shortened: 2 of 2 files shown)


bgruening:
@bioconda/os @bioconda/core it seems that the new build system, or more concrete the new OSX image has some problems. It seems that llvm does not work anymore. I think if we remove llvm and use GCC this PR will pass.

Any idea what is going on here? Was our old llvm never used and instead some compile on the travis image? In conda-forge we don't use llvm, so maybe we should migrate the GCC for all recipes?

daler:
@bgruening I don't know about whether llvm was actually used in the past, but I'm in favor of following conda-forge moving forward. Sounds like another `bulk` run though, for which we'll have to work out details on CCI.

cbrueffer:
Seems strange though, other packages that need compiler support worked fine. LLVM can't be completely unused then.

rvalieris:
i am not really familiar with osx, but this has been discussed before, here: https://github.com/bioconda/bioconda-recipes/issues/1332

I think the important part is that llvm package does **not** contain clang.

mbargull:
> In conda-forge we don't use llvm

Are you sure? `toolchain`/`toolchain3` use `clang`:
- https://github.com/conda-forge/toolchain-feedstock/blob/f6661f3766f27bb36157ca2efaafc4f77ba9573a/recipe/activate.sh
- https://github.com/conda-forge/toolchain3-feedstock/blob/fc8244974e5df85fd11cb9c0fb031f840f86e547/recipe/activate.sh

mbargull:
(Also, I think this discussion was mean to address the other PR, i.e., https://github.com/bioconda/bioconda-recipes/pull/6726. This one might just work with `conda 4.3.33` and only fails due to an R package coming from `defaults`, I think.)

bgruening:
@mbargull see here: https://github.com/conda-forge/r-nloptr-feedstock/blob/master/recipe/meta.yaml#L33 & https://github.com/conda-forge/r-sp-feedstock/blob/master/recipe/meta.yaml as an example. toolchain is not always used.

mbargull:
It's not always used, yes. I was just a bit confused about the "we don't use llvm" -- did you mean that more in reference to R packages? (Sorry for going into it that much, I just want to make sure to get the right mindset for this issue.)

bgruening:
Yeah, just talking here about R and non-windows stuff :)

johanneskoester:
@rvalieris is right about the clang thing. I think on travis, the clang command was a symlink to gcc. So, I guess all those packages used the system gcc on travis macos. At least it was like that when we discussed this some months ago.

mbargull:
The package's configure script added `-lgfortran` but not the required RPATH linker flags, see: https://github.com/bioconda/bioconda-recipes/pull/6726#issuecomment-366690571.

PR #7398, opened 2018-01-29
secapr update

tobiashofmann88:

### DIFF
--- a/recipes/secapr/meta.yaml
+++ b/recipes/secapr/meta.yaml
@@ -18,13 +18,13 @@ build:
 
 requirements:
   build:
-    - python
+    - python2.7*
     - setuptools
     - biopython
     - pandas
 
   run:
-    - python
+    - python2.7*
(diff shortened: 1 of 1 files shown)


tobiashofmann88:
an update of secapr was merged to the bioconda channel last friday, but unfortunately it doens't install because of conflicting specifications. The error is different depending on if I'm trying to install it on my personla macOS or on our departments linux-cluster:

on macOS:
UnsatisfiableError: The following specifications were found to be in conflict:
  - click
  - secapr

on linux cluster:
UnsatisfiableError: The following specifications were found to be in conflict:
  - bottleneck
  - phyluce

I think it may have to do with the python version. All dependencies and the secapr pipeline itself are written and tested in python2.7 only, so I tried to add that to the meta.yaml file (not sure if I did it in the correct way, please give me feedback in case it is incorrect).

Best,
Tobi

PR #7399, opened 2018-01-29
Add ClonalFrameML: Efficient Inference of Recombination in Whole Bacterial Genomes

thanhleviet:

### DIFF
--- a/recipes/clonalframeml/1.11/build.sh
+++ b/recipes/clonalframeml/1.11/build.sh
@@ -1,8 +1,4 @@
 #!/bin/bash
-export CFLAGS="-I$PREFIX/include"
-export LDFLAGS="-L$PREFIX/lib"
-export CPATH=${PREFIX}/include
-
 mkdir -p $PREFIX/bin
 cd src
 bash make.sh
--- a/recipes/clonalframeml/1.11/build.sh
+++ b/recipes/clonalframeml/1.11/build.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+export CFLAGS="-I$PREFIX/include"
+export LDFLAGS="-L$PREFIX/lib"
+export CPATH=${PREFIX}/include
+
+mkdir -p $PREFIX/bin
+cd src
+bash make.sh
+
+cp ClonalFrameML $PREFIX/bin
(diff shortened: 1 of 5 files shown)


recipes/clonalframeml/meta.yaml
```diff
@@ -0,0 +1,28 @@
+package:
+  name: clonalframeml
+  version: "1.11"
+
+build:
+  number: 3
```
bgruening: This should start with 0.

recipes/clonalframeml/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+  run:
+  - libgcc # [linux]
+
+test:
+  commands:
+    - ClonalFrameML 2>&1 | grep "ClonalFrameML"
+
+about:
+  home: https://github.com/xavierdidelot/ClonalFrameML
+  license: GPLv3
```
bgruening: Can you provide a `license_file`

recipes/clonalframeml/1.11/meta.yaml
```diff
@@ -0,0 +1,28 @@
+package:
+  name: clonalframeml
+  version: "1.11"
+
+build:
+  number: 3
```
bgruening: start with 0

recipes/clonalframeml/1.11/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+export CFLAGS="-I$PREFIX/include"
```
bgruening: Are these exports needed? You do not depend on any third party library as it seems.

recipes/clonalframeml/meta.yaml
thanhleviet: Thanks for your comments. The authors did not provide the license file but mentioning that the source code follows GPLv3, should I include this one https://www.gnu.org/copyleft/gpl.html ?
Do I have to create new PR?

bgruening: I will discuss a general approach in an issue as including in every recipe the GPL (as hardcopy) also seems wired, we should do better.

PR #7407, opened 2018-01-29
add Seurat

rvalieris:

### DIFF
--- a/recipes/r-seurat/meta.yaml
+++ b/recipes/r-seurat/meta.yaml
@@ -108,7 +108,8 @@ test:
     - $R -e "library('Seurat')"
 
 about:
-  home: http://www.satijalab.org/seurat, https://github.com/satijalab/seurat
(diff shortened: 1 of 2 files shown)


recipes/r-seurat/meta.yaml
```diff
@@ -0,0 +1,120 @@
...
+    - r-tsne
+    - libgcc
+    - openjdk >=6
+
+test:
+  commands:
+    - $R -e "library('Seurat')"
+
+about:
+  home: http://www.satijalab.org/seurat, https://github.com/satijalab/seurat
```
mbargull: `home` should be a valid URL. Please use
```yaml
  home: http://www.satijalab.org/seurat
  dev_url: https://github.com/satijalab/seurat
```

rvalieris: good call !

PR #7426, opened 2018-01-30
.circleci: set LC_ALL+LANG to en_US.UTF-8 on macos

mbargull:

In CircleCI's macOS image the locale isn't set with `LC_ALL` and `LANG`. Hence `click` will complain if run with Python 3, see https://github.com/bioconda/bioconda-recipes/pull/7417.
The variables are set to `en_US.UTF-8` which mirrors the `extended-base` image.
Tested on https://github.com/bioconda/bioconda-recipes/pull/7417 and works as intended.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,6 +21,9 @@ variables:
   macos: &macos
     macos:
       xcode: "8.3.3"
+    environment:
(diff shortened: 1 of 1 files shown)


bgruening:
This is very unfortunate, thanks for catching!

PR #7428, opened 2018-01-30
Lca classifier

HassanAmr:

### DIFF
--- a/recipes/lcaclassifier/build.sh
+++ b/recipes/lcaclassifier/build.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
-$PYTHON setup.py install
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
 
--- a/recipes/lcaclassifier/build.sh
+++ b/recipes/lcaclassifier/build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+cd LCAClassifier/
+$PYTHON setup.py install
+
--- a/recipes/lcaclassifier/build.sh
+++ b/recipes/lcaclassifier/build.sh
@@ -1,4 +1,3 @@
 #!/bin/bash
-cd LCAClassifier/
 $PYTHON setup.py install
 
--- a/recipes/lcaclassifier/lcaclassifier.patch
+++ b/recipes/lcaclassifier/lcaclassifier.patch
@@ -1,0 +1,59 @@
+Binary files a/.DS_Store and b/.DS_Store differ
+Binary files a/LCAClassifier/.DS_Store and b/LCAClassifier/.DS_Store differ
+Only in b/LCAClassifier/etc: .DS_Store
+Binary files a/LCAClassifier/src/.DS_Store and b/LCAClassifier/src/.DS_Store differ
+Only in b/LCAClassifier/src/LCAClassifier: .DS_Store
+diff -u -r a/LCAClassifier/src/LCAClassifier/arbor.py b/LCAClassifier/src/LCAClassifier/arbor.py
+--- a/LCAClassifier/src/LCAClassifier/arbor.py	2017-03-07 13:51:40.000000000 +0200
++++ b/LCAClassifier/src/LCAClassifier/arbor.py	2017-03-07 16:41:39.000000000 +0200
+@@ -1,3 +1,5 @@
++#!/usr/bin/env python
++
+ import sys
+ import re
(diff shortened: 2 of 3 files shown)


recipes/lcaclassifier/meta.yaml
```diff
@@ -0,0 +1,34 @@
+package:
+  name: lcaclassifier
+  version: 3.0.7
+
+source:
+  url: https://github.com/lanzen/CREST/releases/download/3.0.7/LCAClassifierV3.0.7.tar.gz
+  md5: ca072806b9251f280978c6a3ba520395
```
mbargull: Please use `sha256` instead.

recipes/lcaclassifier/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+build:
+  skip: True  # [py3k]
+  number: 0
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - biom-format
```
mbargull: Only `python` and `setuptools` are required for the build; you can remove the other two.

recipes/lcaclassifier/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - biopython
+
+  run:
+    - python
+    - biom-format
+    - biopython
+
+test:
+  commands:
+    - classify -h
```
mbargull: We can also test the other entrypoint, i.e.,
```yaml
    - nds2CREST -h
```

recipes/lcaclassifier/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+source:
+  url: https://github.com/lanzen/CREST/releases/download/3.0.7/LCAClassifierV3.0.7.tar.gz
+  md5: ca072806b9251f280978c6a3ba520395
+  patches:
+    - lcaclassifier.patch
+
+build:
+  skip: True  # [py3k]
+  number: 0
```
mbargull: Can you add
```yaml
  entry_points:
    - classify = LCAClassifier.classify:main
    - nds2CREST = LCAClassifier.nds2CREST:main
```
to the `build` section?

recipes/lcaclassifier/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - biom-format
+    - biopython
+
+  run:
+    - python
```
mbargull: Acc. to `setup.py`, `setuptools` is also a run requirement.
But I can't see why, so you can either
1. add a patch to remove `setuptools` from `install_requires` in `setup.py` (preferred), or
2. add add `setuptools` to the run requirements

recipes/lcaclassifier/meta.yaml
```diff
@@ -0,0 +1,34 @@
+package:
+  name: lcaclassifier
+  version: 3.0.7
+
+source:
+  url: https://github.com/lanzen/CREST/releases/download/3.0.7/LCAClassifierV3.0.7.tar.gz
+  md5: ca072806b9251f280978c6a3ba520395
+  patches:
+    - lcaclassifier.patch
```
mbargull: Please remove that patch.

recipes/lcaclassifier/lcaclassifier.patch
```diff
@@ -0,0 +1,100 @@
+diff -u -r a/LCAClassifier/src/LCAClassifier/arbor.py b/LCAClassifier/src/LCAClassifier/arbor.py
```
mbargull: Please remove this patch file. It doesn't add anything valuable as it only adds shebangs to files that end up in `$PREFIX/lib`.

recipes/lcaclassifier/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
```
mbargull: Looks like the software is meant to be installed with `install.sh` instead of using `setup.py` directly.

HassanAmr: I will need `buildout`/`zc.buildout' in that case, so, I will add package for it to conda-forge maybe. Thanks.

daler:
Closing due to inactivity., please re-open if needed.

PR #7429, opened 2018-01-30
Update Kallisto to 0.44.0.

cbrueffer:

### DIFF
--- a/recipes/kallisto/meta.yaml
+++ b/recipes/kallisto/meta.yaml
@@ -8,6 +8,8 @@ source:
   fn: kallisto-{{ version }}.tar.gz
   url: https://github.com/pachterlab/kallisto/archive/v{{ version }}.tar.gz
   sha256: 35a81201a56f4557697e6fe693dc6b701bbbd0a7b2b6e1c6c845ef816d67ca29
+  patches:
+    - cmakelists.patch
 
 build:
     number: 0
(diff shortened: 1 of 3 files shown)


recipes/kallisto/meta.yaml
```diff
@@ -7,20 +7,23 @@ package:
...
 build:
     number: 0
     string: hdf5{{CONDA_HDF5}}_{{PKG_BUILDNUM}}
 
 requirements:
     build:
-        - hdf5 {{CONDA_HDF5}}
+        - hdf5 {{ CONDA_HDF5 }}*
         - cmake
         - zlib
```
druvus: zlib line should also be updated so it is pinned

cbrueffer: Good point! Curious this didn't came up during linting.

bgruening: Me we fixed this recently ... lets try again.

bgruening: Oh wait, you don't have it as run-time dependency, so this is a static build right? If so you don't need to pin it.

bgruening:
Seems like htslib is missing?

cbrueffer:
Yes, needs a bit more work to find the conda htslib headers. Haven't had time to continue with this yet.

cbrueffer:
An update was committed in https://github.com/bioconda/bioconda-recipes/commit/4ec8ca5144388aac70303e2fb7384d6c97a9c5fe so close this PR.

PR #7436, opened 2018-01-30
Porechop update

jvolkening:

This PR points the source metadata to the (now updated) mainstream repo. As the upstream author chose a different version string for the branch/tag, and as this is a temporary workaround for the GCC issue, I have replaced the templated URL string with an explicit string. This seems preferable to modifying the recipe version when there are no actual changes to the package code. This can be replaced for the next upstream release.

### DIFF
--- a/recipes/porechop/meta.yaml
+++ b/recipes/porechop/meta.yaml
@@ -12,6 +12,7 @@ package:
 
 source:
   fn:  porechop-{{ version }}.tar.gz
+# url: https://github.com/rrwick/Porechop/archive/v{{ version }}.tar.gz
   url: https://github.com/rrwick/Porechop/archive/v0.2.3-C++11.tar.gz
   sha256: 817371dceed77e583e387e9ca7c10214c0faff7c0a89320d60a66efdc6c0d35a
 
--- a/recipes/porechop/meta.yaml
+++ b/recipes/porechop/meta.yaml
@@ -12,12 +12,11 @@ package:
 
 source:
   fn:  porechop-{{ version }}.tar.gz
-  # url: https://github.com/rrwick/Porechop/archive/v{{ version }}.tar.gz
-  url: https://github.com/jvolkening/Porechop/archive/v{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


bgruening:
Do we want to keep the package version as `0.2.3_seqan2.1.1`?

jvolkening:
Was this a question for me or for others' opinions?

If the package version changes, then both versions show up in the package listing, correct? It seems a likely source of confusion to have two different versions with identical code and different suffixes. Or is it possible to remove the previous version from the listing?

mbargull:
I think there was a misunderstanding here. The `0.2.3_seqan2.1.1` version should be kept.
That release is the same as @jvolkening' previous contribution but now only comes from the main repo, see:
https://github.com/rrwick/Porechop/releases/tag/v0.2.3-C%2B%2B11
and
https://github.com/rrwick/Porechop/pull/41.
The `0.2.3` is still a distinct other release.

bgruening:
I'm just wondering if we should drop the seqan sufix now.

mbargull:
The reason being
1. it's unlikely (and currently not possible) we'll build a package for the "vanilla" `0.2.3`,
2. @jvolkening's changes are "sanctioned" and maintained by the main devs,
3. and, most importantly, the changes would likely not or not at all be noticeable by the user (given we are talking about a dependency for the backend)?

If this is the reasoning, I can get behind that. @jvolkening, what do you think?

jvolkening:
Right, I second all of those points. In an ideal situation, GCC ≥ 4.9 will find its way to conda and then this package can just track the master branch of Porechop. Until that happens, Ryan has agreed to keep the "C++11" branch in sync with master (except for SeqAn) and create separate tags from that branch for each stable release, and this bioconda package needs to track those branch tags.

I would leave the suffix in place as long as we are tracking the non-master releases, but have no problems doing whatever those of you with a better understanding of the conda ecosystem think is best.

bgruening:
Thanks @jvolkening. I don't care that much about it, so lets get it in.

PR #7438, opened 2018-01-30
Stacks 2 0

pbordron:

### DIFF
--- a/recipes/stacks/2.0/build.sh
+++ b/recipes/stacks/2.0/build.sh
@@ -5,7 +5,7 @@ export LDFLAGS="-L${CONDA_PREFIX}/lib" # for both stacks and included htslib dep
 export CPATH=${PREFIX}/include         # for included htslib dependency
 export CFLAGS="-I${PREFIX}/include"    # for included htslib dependency
 
-./configure --prefix=$PREFIX --enable-bam --includedir=${CONDA_PREFIX}/include/bam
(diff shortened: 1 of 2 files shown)


pbordron:
ping @lecorguille @bioconda/core
The conda-build version of bioconda doesn't allow to use the jinja2 function {{ compiler('c') }} which solves the issue with gcc compiler. Is there some project to update to a new one?

https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#host

lecorguille:
ping @bioconda/core

pbordron:
@bgruening @lecorguille the htslib library included with stacks didn't take in account the env. variables  like CC, CXX, etc.
Now it works.

lecorguille:
Congrats @pbordron!

chapmanb:
Philippe -- nice work and thanks for the examples about how to use the newer gcc compilers. This will really help me get unstuck on some recipes I'd been working on that require new versions.

bgruening:
@pbordron nice work. Please note that these compiler where designed to work with conda-build-3. I'm not sure about the consequences. conda-forge has choose to stick to `gcc` until the transition to cb3.
I guess its ok to make an exception for those packages, just wanted to point this out.

ping @bioconda/core

johanneskoester:
We should be careful using the new compiler packages for libraries, but otherwise, if it cannot be avoided, it should be fine.

PR #7440, opened 2018-01-30
Added k8 javascript shell

lh3:
This recipe adds the interpreter needed to run multiple scripts written by me. These scripts include converting GFF to BED12, converting long-read spliced alignment to BED12, calling variants from contig-to-genome alignment, evaluating mapping accuracy, beautifying long-read alignment, etc. The applications are directly relevant to biological sequence analyses.

It is non-trivial to compile k8 because it is built on top of an old version of V8 and compiling V8 is complicated by itself. As a result, this recipe does not compile k8. Instead, it downloads portable precompiled binaries from the k8 github repo. These binaries are known to work on CentOS5 and Mac OSX 10.6.

I am sending this pull request to see the test. It is likely that the PR does not work on my first try. In that case, I may need the help from more experienced users. Please don't merge even if the test passes. I will do that myself. Thank you.

### DIFF
--- a/recipes/k8/build.sh
+++ b/recipes/k8/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+mkdir -p $PREFIX/bin
+
+cp k8-`uname -s|tr [A-Z] [a-z]` $PREFIX/bin/k8
--- a/recipes/k8/meta.yaml
+++ b/recipes/k8/meta.yaml
@@ -1,0 +1,28 @@
+{% set name = "k8" %}
+{% set version = "0.2.3" %}
+{% set hash = "e178cd0073bdeb1907b9b73beea1ad5ff7a09b318d67cc44205fb4f248107214" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/attractivechaos/k8/releases/download/v{{ version }}/{{ name }}-{{ version }}.tar.bz2
+  fn: {{ name|lower }}_{{ version }}.tar.bz2
+  sha256: {{ hash }}
+
+build:
+    number: 0
+
+requirements:
+    run:
+        - libgcc
+        - zlib {{ CONDA_ZLIB }}*
+test:
+  commands:
+    - k8 -v | grep 'K8'
+
+about:
+  home: https://github.com/attractivechaos/k8
+  license: MIT
+  summary: 'Lightweight javascript shell based on V8.'


bgruening:
@lh3 this looks alright to me. My only concern is that this package actually belongs into conda-forge.

lh3:
K8 is only used by myself so far, and probably will be used for my scripts only. I am not sure general developers would be interested in that.

lh3:
I will merge this PR by the end of the day if no one objects. Thank you, @bgruening, for the confirmation that it seems to build ok.

johanneskoester:
@lh3 would it be a problem to submit to conda-forge instead? The recipe is straightforward, I would not expect a lot of review overhead, and if you ping me, I can even merge it if it seems like they are too busy. You will have write access to the resulting feedstock.

lh3:
I will try conda-forge sometime later.

johanneskoester:
Ok, thanks! Sorry if this causes you trouble, but if we start making exceptions it will be only harder to clean up in the future.

lh3:
Looking at the existing recipes in conda-forge, I think those are general-purpose and much more essential tools/libraries. This PR is not. It is just for niche applications in bioinformatics only. No one at conda-forge would be interested in k8.

I understand you concern, but I don't want to contaminate conda-forge with niche recipes. Probably I will just let users download from github. It is a single portable binary and won't cause much trouble anyway.

johanneskoester:
Ok, if you think this is only relevant for bioinformatics, we can reopen this PR. I was thinking that it is just a javascript shell. What are the intended applications then and why is it not relevant for people who would otherwise use nodejs?

lh3:
There were quite a few javascript shells several years ago. Most died because they couldn't compete with node.js on features. K8 has much fewer features than all javascript shells, let alone node.js. I don't think any node.js developers would consider k8.

Then why not just use node.js? Because k8 has more convenient and efficient synchronous streaming APIs and allows to create objects taking large RAM, bypassing the V8 garbage collector. It works better with text processing and daily bioinformatics scripting while benefiting from the performance of V8. I don't think bioinformatics developers would use k8, either, but having k8 in bioconda would allow them to use a bunch of my scripts I mentioned earlier.

johanneskoester:
Ok, that all makes sense, but it still could be that somebody outside of bioinformatics wants to use it, right? At least, it has quite some github stars. Also, being a niche product does not disqualify it for conda-forge. There is a lot of niche stuff on conda-forge, and there is no policy that they only want to have one software of each type. So, I don't think putting it on conda-forge would be a problem. If you like, I can even do it quickly...

PR #7441, opened 2018-01-30
bumped fastqc version 10 0.11.7

caseywdunn:
* [X ] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).
* [X ] This PR updates an existing recipe.

### DIFF
--- a/recipes/fastqc/meta.yaml
+++ b/recipes/fastqc/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "fastqc" %}
-{% set version = "0.11.6" %}
-{% set sha256hash = "86790d1de58d715ce95ec30eed7b16a1bcc0c08874ec37db877bbebbd8172a5e" %}
+{% set version = "0.11.7" %}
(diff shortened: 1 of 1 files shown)


caseywdunn:
For some reason the `openjdk` dependency for this packages is installed from the default channel rather than the conda-forge channel. It works under conda-forge, but there are font problems under the default version.

I am testing in a docker container run with `docker run -it conda/miniconda2 /bin/bash`. This is how I prepare the container:

    conda config --add channels defaults

If I run `conda install openjdk`, it installs from conda-forge as expected:

        openjdk:   8.0.144-zulu8.23.0.3_1 conda-forge

But if I run `conda install fastqc`, which depends on openjdk, it installs from the default channel instead:

        fastqc:                    0.11.7-pl5.22.0_0 bioconda   
        font-ttf-dejavu-sans-mono: 2.37-0                       
        openjdk:                   8.0.121-1                    
        perl:                      5.22.0.1-0        conda-forge

Perhaps the dependency on `font-ttf-dejavu-sans-mono`, which is from the default channel, is leading it to install open-jdk from default rather than conda-forge.

PR #7448, opened 2018-01-31
bioconda-utils 0.11.4: fix conda dependency

mbargull:

Updates the dependency on `conda`.
Also removes all build requirement apart from `python` and `setuptools` as they are not needed during the actual build.


bgruening:
Thanks Marcel!

PR #7494, opened 2018-02-02
Update nudup to 2.3.3

lparsons:


recipes/nudup/meta.yaml
```diff
@@ -1,11 +1,11 @@
 package:
   name: nudup
-  version: 2.3.2
+  version: 2.3.3
```
bgruening: @lparsons could you use next time the fancy template variables? Thanks!

PR #7506, opened 2018-02-05
Bump wub.

bsipos:
This commit updates wub to the latest version.

### DIFF
--- a/recipes/wub/meta.yaml
+++ b/recipes/wub/meta.yaml
@@ -31,7 +31,7 @@ requirements:
   - tqdm
   - h5py
   - statsmodels
-  - htslib <1.4
+  - htslib {{ CONDA_HTSLIB }}*
 
   run:
   - python
@@ -47,7 +47,7 @@ requirements:
   - tqdm
   - h5py
   - statsmodels
-  - htslib <1.4
+  - htslib {{ CONDA_HTSLIB }}*
 
 test:
   imports:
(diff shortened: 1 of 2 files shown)


recipes/wub/meta.yaml
```diff
@@ -31,7 +31,7 @@ requirements:
   - tqdm
   - h5py
   - statsmodels
-  - htslib <1.4
+  - htslib {{ CONDA_HTSLIB }}*
```
mbargull: Looks like `wub` is a well-behaving Python-only packge. So we should only use
```yaml
requirements:
  build:
    - python
    - setuptools
```
for the build requirements.

recipes/wub/meta.yaml
```diff
@@ -47,7 +47,7 @@ requirements:
...
```
mbargull: When we remove `htslib` from the `build` requirements, we don't have to pin it anymore. If the previous restriction on `<1.4` still applies, that should be restored. If there are no restrictions regarding the `htslib` version, then
```yaml
    - htslib
```
should suffice.
Also there seems to be a version constraint in https://github.com/nanoporetech/wub/blob/v0.2.0/setup.py#L19 for `pandas` -- we should add that one here, too, i.e.,
```yaml
    - pandas >=0.20.2
```

mbargull:
Thanks for the update @bsipos and @druvus, also!

PR #7507, opened 2018-02-05
Update RGI to 4.0.0

sachalau:


Cashalow:
The pypi package filetype ( https://pypi.python.org/pypi/filetype/1.0.0 ) is now need as a dependency. I will add it to conda-forge and then start back working on this recipe.

recipes/rgi/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+{% set sha256 = "17abb2e89608b3d82a491d3b8e894456da6d3c955df6e1563ae9240b02b3d0d1" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+
+  url: https://card.mcmaster.ca/download/1/software-v{{ version }}.tar.gz
+  fn: card-software.tar.bz2
```
druvus: I think `rgi_{{ version }}.tar.gz` is a better option here to avoid collisions between updated versions.

druvus:
@Cashalow I was also planning to get this recipe into Bioconda so I tried to help you by opening a PR for filetype at [conda-forge](conda-forge/staged-recipes#5117)

recipes/rgi/meta.yaml
Cashalow: I'm not sure I understand how their download folder is organized. The full address for the rgi software is https://card.mcmaster.ca/download/1/software-v4.0.1.tar.gz that downloads a file called card-software.tar.bz2. Then when you extract this file (on the build file) you get the more standard rgi-4.0.0.tar.gz. I'm not sure I can get access directly to this tar file but I will try.

Cashalow:
According to the failed circle CI build, I also need to download the card.json file from their website to load into rgi. Would you know which field I could add in the yaml for this file to be downloaded too ? 

Thanks a lot !

mbargull:
Blocked by https://github.com/conda-forge/staged-recipes/pull/5117

druvus:
The blocking recipe is now merged

recipes/rgi/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+tar xf rgi-4.0.0.tar.gz 
+cd rgi-4.0.0
+$PYTHON setup.py install
```
mbargull: Please use
```bash
$PYTHON setup.py install --single-version-externally-managed --record record.txt
```

Cashalow:
Hey everyone @druvus @mbargull ! 

I have started back working on the recipe but I'm still missing something. I need to download the CARD data and load it into rgi. I have tried to add different entries in the source part of the file (as described here: https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html). However linting now fails because source being a dictionary instead of a list is not expected I think.

What should I do ? Should I skip linting ? Or is there another way to download extra files from the meta yaml ?

Thanks for your help !

recipes/rgi/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  sha256: {{ sha256 }}
+  
+
+build:
+  skip: True # [not py27]
+  number: 0
+
+requirements:
+  build:
+    - biopython
```
mbargull: Please add
```yaml
    - python
    - setuptools
```

recipes/rgi/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - six >=1.7.0
+    - gcc   # [not osx]
+    - llvm  # [osx]
+    - zlib {{ CONDA_ZLIB }}*
+  run:
+    - filetype
```
mbargull: Please add
```yaml
    - python
```

recipes/rgi/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - rgi main -h
+    - rgi tab -h
+    - rgi parser -h
+    - rgi clean -h
+    - rgi galaxy -h
+
+
+about:
+  home: https://card.mcmaster.ca
+  license: https://card.mcmaster.ca/about
```
mbargull: Please add
```yaml
  license_file: LICENSE
```
As for the license itself, it should definitely state that it's non-commercial. I didn't read https://card.mcmaster.ca/about, so can't give exact advices yet

mbargull:
> I need to download the CARD data and load it into rgi.

I think it's only a warning that is shown. Let's see if the tests run with the test command you fixed

> I have tried to add different entries in the source part of the file (as described here: https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html). However linting now fails because source being a dictionary instead of a list is not expected I think.

Multiple source files is a `conda-build 3` feature -- we still use version 2, so can't use that one yet.

> What should I do ? Should I skip linting ? Or is there another way to download extra files from the meta yaml ?

If it's really needed to download additional files, you'd have to do that "manually" in `build.sh`, e.g., with `curl`. But I think we'll be fine without that JSON file, given that the Linux test ran through.

Cashalow:
Indeed just realized that it was only a warning... however I am not sure rgi can be run without this step, I'll try now. I'll add it in the build if needed then. Thanks !

mbargull:
If you need to add it, then you should add
```yaml
    # curl/openssl are used to download/verify WHATEVER
    - curl
    - openssl
```
to the `requirements/build` section and use something akin to
```bash
FN='WHATEVER'
URL='https://some.thing/WHATEVER'
SHA256=0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
curl -Lo "$FN" "$URL"
openssl sha256 "$FN" | grep $SHA256 || exit 1
```
in `build.sh`.

mbargull:
>  by any non-academic organization whether or not for non-commercial (including research) or commercial purposes is prohibited

Oh, yeah, that sounds like a show-stopper ☹️.

Cashalow:
I'll contact them to know what they think. I didn't find license files for the previous versions of the software already in bioconda however (I looked for just one version).

mbargull:
I mean if someone asked them for
>  except with written permission of McMaster University.

then we could add their response (if positive) as a license file in the recipe. But I'm not sure whether we have such individual licenses in any other recipe. @bioconda/core: Any opinions?

> I didn't find license files for the previous versions of the software already in bioconda however

Well, sometimes license conditions get overlooked, unfortunately.
Best case would be to get a positive feedback, worst case would be that we'd have to remove RGI from the channel all together...

Cashalow:
Sorry I didn't want to throw the stone at anybody, actually I looked on the folder of rgi 3.2.1 downloaded directly from their website and I can't find a license file, so apparently this was not mentioned before.

mbargull:
> Sorry I didn't want to throw the stone at anybody

No need to apologize! For one, it was me that brought up those licensing restrictions and even then I don't feel any stone was thrown, also. It's objectively important we get that legal stuff right, no need to (`git`) blame anyone 😉.
Thanks for contacting the authors, though!

Cashalow:
So actually the CARD developers were planing to add the new version themselves anyway, so they don't see any problem (I think the previous version were added by them too). I've let them know that the new license file caused doubt on our side.

mbargull:
> I've let them know that the new license file caused doubt on our side.

Thanks! Yeah, IMO, they now have to provide us with some permission text.
Also, I think it would be best to copy their text from https://card.mcmaster.ca/about into a LICENSE file in the recipe folder, like described in https://conda-forge.org/docs/meta.html#packaging-the-license-manually.
For reference, relevant changes from http://web.archive.org/web/20170707033834/https://card.mcmaster.ca/about to the current version are:
```diff
-5. Terms of Use for Commercial Reproduction
+5. Terms of Use for Commercial or Non-Academic Organizations
 
 Use or reproduction of Materials on the site, in whole or in part,
-for the purposes of commercial distribution or use is prohibited,
+by any non-academic organization whether or not for non-commercial
+(including research) or commercial purposes is prohibited,
 except with written permission of McMaster University. Commercial
 uses of the Materials on the site are offered only pursuant to a
 written license and user fee. To obtain permission and begin the
 licensing process, please contact McMaster University using the
 address in section 3.
```

arpcard:
Hi all. Yes, the license was tweaked with release of RGI 4.0.0, but it was not our intent to block Conda usage. We know license conditions sometimes get overlooked, but we still want to RGI to be available to Conda users. If we add a LICENSE file with the existing license disclaimer text, can we add text to put RGI in the clear for Conda? Suggestions? Please also note we have made previous versions of RGI available on Conda via GitHub, see https://card.mcmaster.ca/download and https://github.com/arpcard and were just preparing to update it to 4.0.0. Shall we make maintaining RGI in Conda the dedicate task of the CARD team? Just trying to avoid duplicate effort.

Cashalow:
As you wish ! This PR is ready to be merged, the only thing left to decide is whether the card.json data should be loaded during the build of the recipe or whether we should let the user do that once rgi is installed with bioconda. I'll let you decide what you think is best !

arpcard:
@Cashalow The `card.json` should not be loaded during the build. We will let the users download it directly from https://card.mcmaster.ca/download.

@Cashalow Is it possible to keep the recipes in directories with the corresponding version number as seen on this tree: https://github.com/Cashalow/bioconda-recipes/tree/637d3cc8cf5c24b89160bff5c1cb842f53a3a323/recipes/rgi

mbargull:
> If we add a LICENSE file with the existing license disclaimer text, can we add text to put RGI in the clear for Conda?

You are the licensor, so sure you can. You could either make an exception specifically for Bioconda/Conda or more generally for public software distribution systems/services, if you will, I guess.
> Suggestions?

I don't really feel qualified, sorry. If I would have to compose a license or license exception myself, I'd definitely look at other existing licenses to "crib" from 😉.

> Shall we make maintaining RGI in Conda the dedicate task of the CARD team? Just trying to avoid duplicate effort.

Bioconda is an open community project, so everyone could update the RGI recipe. To avoid duplicate effort, it's best to just look for whether the recipe is up to date or if there already is a PR which updates it. If you come forward to update the recipe along new software releases, then this is always appreciated, of course.

> @Cashalow Is it possible to keep the recipes in directories with the corresponding version number as seen on this tree: https://github.com/Cashalow/bioconda-recipes/tree/637d3cc8cf5c24b89160bff5c1cb842f53a3a323/recipes/rgi

Do you plan to maintain the older versions? We have yet to agree/formulate a recommended approach regarding recipe subdirectories. But, to minimize maintenance burden, it's common to have only one (main) recipe for the latest version and add additional subdirs only for other (maintained) versions for which new builds/packages should be created. See https://github.com/bioconda/bioconda-recipes/pull/7697#issuecomment-366082188 and the linked discussion for details.

arpcard:
> Do you plan to maintain the older versions?

We wanted to keep the directories, if for example you build multiple versions.

mbargull:
Note that builds of older versions won't be removed from the Conda channel. But if you think it's not unlikely that new builds for older versions will be created, then it's totally fine to keep the versioned subdirectories.
Any news on the updated license/permission?

arpcard:
>Any news on the updated license/permission?

We are still working on this, we will keep you posted. Cheers.

arpcard:
Apologies for the delay. I'm proposing to add this text to the license file we are drafting:

The only exception is the distribution of the Resistance Gene Identifier software on the Conda
open source, cross-platform, language-agnostic package manager and environment management
system (https://conda.io/). Conda users are still required to abide by the Terms of Use 
listed, specifically sections 4 and 5 below.

arpcard:
Section 4 is text letting academics and government agencies unrestricted use. Here is Section 5:

5. Terms of Use for Commercial or Non-Academic Organizations

Use or reproduction of Materials on the site, in whole or in part, by any non-academic 
organization whether or not for non-commercial (including research) or commercial purposes 
is prohibited, except with written permission of McMaster University. Commercial uses of 
the Materials on the site are offered only pursuant to a written license and user fee. To 
obtain permission and begin the licensing process, please contact McMaster University 
using the address in section 3.

arpcard:
So, new text allows Conda to distribute RGI without restrictions. The hope is that users read the license and contact us if industrial. We know this may get overlooked sometimes and we are ok with that.

Cashalow:
I think this is fine but I think you would need to update the LICENSE file in your software distribution so that the tar we are downloading for bioconda incorporates those changes. What do you think @mbargull ?

arpcard:
updated LICENSE file already drafted and will be added shortly to the repo. thanks!

druvus:
@sachalau and @arpcard is this ready to be fixed and updated to v4.0.3?

sachalau:
I think we are good to go now.

sachalau:
@mbargull Could you validate the changes so that this PR could be merged please ? Thanks a lot.

sachalau:
Done ! If think this could be merged if there are no other comments.

sachalau:
@bgruening Hello ! Could I get a merge on this please ?

druvus:
thanks @sachalau

mbargull:
> @mbargull Could you validate the changes so that this PR could be merged please ? Thanks a lot.

Sorry, didn't see the notifications before.
From my (non-lawyer-) understanding, the "Conda" exception definitely covers distribution by Bioconda -- surely some caviler could object, but I think the intent is very  clear.
Thanks @sachalau and @arpcard for working on this!

PR #7524, opened 2018-02-05
R canopy

vinjana:


jenzopr:
Looks good, but is there a specific reason to use version 1.2.0 instead of 1.3.0 that you currently get by default via CRAN? 
You could also use `conda skeleton cran canopy` to automatically create the recipe (see https://bioconda.github.io/guidelines.html#r-cran for guidelines).

Best,
Jens

vinjana:
Yes. That's the version we need for our workflow publication. Getting the researchers to re-test their workflows with newer versions is a hard business.

Thanks for the r-cran tip!

jenzopr:
Alright! I think we can merge it for now so that the version 1.2.0 is available for you.
However, once its build, we should update the recipe to 1.3.0 (with `conda skeleton cran canopy`).
Would that be fine for you?

vinjana:
Great. Thanks. I guess I should leave the 1.2 in a versioned subdirectory, right?

jenzopr:
No need for that. Once 1.2.0 is built successfully, the version will be persistent in the bioconda channel even if we update to 1.3.0 afterwards.

PR #7525, opened 2018-02-05
Perl json create

vinjana:


bgruening:
I think this fits better into conda-forge.

vinjana:
Correct me if I am wrong, but all perl-json[-.*] packages are in bioconda! None in conda-forge! In particular, perl-json-parse, a dependency (at least during testing) of perl-json-create is only in bioconda.

bgruening:
@vinjana you are correct, but it would be good if we can migrate them to conda-forge.

PR #7536, opened 2018-02-06
replace shorah version 1.1.2 with 1.1.3

DrYak:
The new version in this PR [fixes](https://github.com/cbg-ethz/shorah/pull/42) data corruptions [found by](https://github.com/cbg-ethz/shorah/issues/41) @sposadac  [and](https://github.com/cbg-ethz/shorah/issues/9) @ozagordi , caused by a '*use of uninitialised data*' bug


DrYak:
Hum seems that the tests are running on my CircleCI  account ?
(And I only have the free account, without MacOS machines)

How should I proceed ?

bgruening:
If you don't want to maintain the 1.1.2 version you can put the recipe directly under recipes/shorah/

DrYak:
@bgruening : There are other older versions (which predate my arrival at cbg-ethz) that I've left untouched.

1.1.3 is exactly 1.1.2 but with the bug removed.

I am not currently aware of any publication being base on versions 1.1.1 and 1.1.2 (as their publication was rather recent), and thus probably nobody would need to be able to replicate results "*down to the bugs*"-level. So I propose to retire version 1.1.2 and only keep 1.1.3 for now (but I could re-make version 1.1.2 available on request).

DrYak:
CircleCI tests passed (currently using the macOS trial plan).

bgruening:
@DrYak that our CI plan is used is not normal. I'm not aware of any other PR that has this problem. Please let us know if this happens again.

PR #7543, opened 2018-02-07
Add biodiff

0xaf1f:

@bioconda/core, this package uses the program `diff` on the system. The implementation that comes with MacOS may not be compatible.  Is there a way to exclude MacOS? I'd rather not maintain a conda package for GNU diff.

### DIFF
--- a/recipes/biodiff/meta.yaml
+++ b/recipes/biodiff/meta.yaml
@@ -1,0 +1,41 @@
+{% set version = "0.2.2" %}
+
+package:
+  name: biodiff
+  version: {{ version }}
+
+source:
+  fn: archive.tar.gz
+  url: https://gitlab.com/LPCDRP/biodiff/repository/{{ version }}/archive.tar.gz
+  sha256: 30568afcacdc40581f8b1689f35aee45103aab0ec33dac07b38ab155b1833a5b
+  patches:
+    - osx.patch
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - autoconf
+    - automake
+    - pandoc
+  run:
+    - perl
+
+test:
+  source_files:
+    - test/lambda-phage/*
+  commands:
+    - biodiff test/lambda-phage/reference.fasta test/lambda-phage/snps-indels.fasta
+
+about:
+  home: https://gitlab.com/LPCDRP/biodiff
+  license: GPLv3+
+  license_file: LICENSE
+  summary: 'exact comparison of biological sequences'
+  description: >
+    biodiff is a variant caller that determines the exact differences between two biological sequences.
+    It can operate on DNA and protein sequences, as long as they are in fasta format.
+    The sequences to be compared must have the same fasta header (up to the first whitespace).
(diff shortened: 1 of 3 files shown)


0xaf1f:
It's passing for me locally with `circleci build`. Although, looking in `.circleci/config.yml`, it looks like `--mulled-test` doesn't get passed for local builds.

0xaf1f:
Ok, @bioconda/core, I believe this one's all set now. Many thanks!

tomkinsc:
Looks good. Is it possible to build `biodiff` for OSX?

recipes/biodiff/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+  home: https://gitlab.com/LPCDRP/biodiff
+  license: GPLv3+
+  license_file: LICENSE
+  summary: 'exact comparison of biological sequences'
+  description: >
+    biodiff is a variant caller that determines the exact differences between two biological sequences.
+    It can operate on DNA and protein sequences, as long as they are in fasta format.
+    The sequences to be compared must have the same fasta header (up to the first whitespace).
+    If the reference and query each have only one sequence, however, the header need not match and the comparison will be done, but a warning will be emitted.
+    Output is in the Variant Call Format.'
```
tomkinsc: There's an unneeded `'` at the end of the multi-line description.

0xaf1f:
> Is it possible to build biodiff for OSX?

Not at the moment. OSX's `mktemp` function behaves differently than the busybox/GNU implementations.

recipes/biodiff/meta.yaml
0xaf1f: oh, I missed that when changing it to a block quote. I could get to it tonight.

tomkinsc:
Is the usage in the shell script [here](https://gitlab.com/LPCDRP/biodiff/blob/master/biodiff#L11)? Can you get equivalent behavior by changing the args by `uname`? Something like:
```
if [ "$(uname)" = "Darwin" ]; then
      # mktemp call with bsd/osx-specific args
  else
     # mktemp call with gnu/linux-specific args
fi
```
(and same question for `diff`)

0xaf1f:
Yes, of course it's possible to make it work one way or another, but I don't have a Mac and testing it by pushing to CircleCI is much more cumbersome.  I would probably do it at some point, but if OSX support is going to be a hard condition for acceptance, I'll have to revisit this at some later point.

tomkinsc:
We can definitely accept it without OS X, I just wanted to ask in case it would be trivial to make it work.

0xaf1f:
ok, so it seems you were right. Most of the tedious iterations were already done yesterday while making it compatible with busybox.

so this now includes a patch to enable compatibility with OS X and removes the stray quote in the description.

tomkinsc:
Great! Let's merge it in. Any plans to include the patched changes in a future release?

0xaf1f:
Yes, I likely will incorporate it in some form. I appreciate your patience!

Issue #7349, opened 2018-01-23
Quast 4.6.1 incomplete package


epruesse:
Since 4.6.1, the `quast` package no longer includes the BLAST database built from SILVA and does not include an install script or post_inst hook to complete the package after installation.

This leads to problems e.g. with Snakemake workflows using conda to implicitly provide environments. Firstly, the files downloaded by quast on the fly end up in the local environment, not in the `$CONDA_ENV/pkgs/quast...` folder, so they are downloaded for each environment again and again (i.e. for each Snakemake working directory). If a Snakemake workflow is run with parallelism, quast will also terminate with a bunch of exceptions resulting from a race condition downloading and building that reference.

Could we have a `quast_download_databases` style script in the package? Or even a post_inst hook? Or maybe just have that blast database distributed with the package, or in a companion package `quast-dbs` (as it's life cycle is probably independent from quast itself).

epruesse:
ping @alexeigurevich 

See also #7008

alexeigurevich:
Hi!
Could you add execution of `metaquast --test-no-ref` in the appropriate place (post_inst/script/etc?). This would download and configure BLAST/SILVA related stuff. 
Note that the maintainers asked to remove this part from the Quast default package since SILVA is really huge (and relatively rarely needed for the majority of users).

mbargull:
@epruesse, if this is a pressing issue for you and you don't want to wait until we decided how to approach this problem for the `quast` package, I believe, you could (as a temporary measure) run
```bash
python -c 'from quast_libs.search_references_meta import download_blastdb; download_blastdb()'
```
inside the environment you installed `quast 4.6.1` into. @alexeigurevich, correct me if I'm wrong and that command doesn't suffice to get the database ready.

If you have a workflow that frequently creates new environments and you are concerned about the repetitive downloads, a `quast_download_databases` or `post-link.sh` script won't get you far as they will download the database for every `quast` install (i.e., a `post-install.sh` can't make use of Conda's cache). So for this scenario I only see the two possibilities to avoid multiple downloads:
1. put the database (semi-)manually in place,
2. create a `quast-dbs` package.

Regarding 1., this could be done
 * semi-automatic by redirecting the download of http://www.arb-silva.de/fileadmin/silva_databases/release_123/Exports/SILVA_123_SSURef_Nr99_tax_silva.fasta.gz by a local proxy and using the command I gave above.
 * manually by letting `quast` download it once and then copying the file manually every time.

[...]

PR #7546, opened 2018-02-07
quast 4.6.3

alexeigurevich:


mbargull:
Can we include the request from https://github.com/bioconda/bioconda-recipes/issues/7349 in this one?
I think what we need is
* add `blast` dependency (something I missed in the previous PR, sorry)
* add script `quast-download-blastdb`:
```python
#!/usr/bin/env python

from quast_libs.search_references_meta import download_blastdb
download_blastdb()
```
* add script `quast-download-manta` (replace `_SOME_SUBMODULE_` with an appropriate one):
```python
#!/usr/bin/env python

from quast_libs.ra_utils.misc import download_manta
from quast_libs._SOME_SUBMODULE_ import logger
download_manta(logger)
```
* make `build.sh` copy those two scripts, making sure they are executable:
```bash
cp $RECIPE_DIR/quast-download-* $PREFIX/bin
chmod +x $PREFIX/bin/quast-download-*
```
* add a `post-link.sh` (feel free to change the text as you see fit):
```bash
#!/bin/bash

echo 'The quast package does not bundle Manta as well as the BLAST database from SILVA.'
echo 'To be able to use those, please run'
echo '    quast-download-manta'
echo '    quast-download-blastdb'
exit 0
```
* What about Gage? `compile_gage()` is only run for `install_full`. Do we need to do something about this, or does QUAST call that function on the fly if needed?

(Note that I wrote the above quickly without testing and all.)

alexeigurevich:
> Can we include the request from #7349 in this one?

Ok, makes sense, will add.

> add a post-link.sh (feel free to change the text as you see fit):
> ...
> echo '    quast-download-manta'

Should it be `./quast-download-manta` ? Should we also add `.py` extension for these scripts?

> What about Gage? compile_gage() is only run for install_full. Do we need to do something about this, or does QUAST call that function on the fly if needed?

Quast call this function on the fly, but (1) java compiler is needed (we have `openjdk` in `meta.yaml`, so this is fine) and (2) Quast should have write permissions under `quast_libs` (to save `.class` files compiled from `.java`). Does it always have such permissions in runtime?

mbargull:
> Should it be ./quast-download-manta ?

No, that way the user would try to run a `quast-download-manta` from their current working directory. Since we copy `quast-download-manta` to `$PREFIX/bin`, it will be on `$PATH` when the environment is activated.
> Should we also add .py extension for these scripts?

If you are more comfortable with naming them `quast-download-*.py`, you can do that. It isn't needed though, as the shebang `#!/usr/bin/env python` is what instructs the system to run them with `python`.

> Does it always have such permissions in runtime?

Good question! Maybe not always, e.g., in a multi-user environment. So, I think it would be best to just patch `setup.py` to also run `compile_gage()` when doing a "normal" `install`. Then we can also remove its special handling from `pre-unlink.sh` as `conda` will do its job, i.e., in https://github.com/bioconda/bioconda-recipes/blob/3303e73516844f3ff703561109bf41f94e4c85ee/recipes/quast/pre-unlink.sh#L17-L20
```diff
-from quast_libs.gage import gage_dirpath
-for dirpath in (manta_dirpath, gage_dirpath):
-    if isdir(dirpath):
-        rmtree(dirpath, ignore_errors=True)
+if isdir(manta_dirpath):
+    rmtree(manta_dirpath, ignore_errors=True)
```

alexeigurevich:
All done, please check!

mbargull:
Thanks for the "fix fix" ;). Tested the scripts and now everything works as expected! (Well apart from that I didn't see the message from `post-link.sh`, but I believe there was something else amiss, unrelated to the package and script themselves.)

mbargull:
Thanks @alexeigurevich! Once it's build and available on the channel, I'll try it out again locally and see whether `conda` shows the `post-link.sh` message -- if not I'll debug why that's the case.

alexeigurevich:
Great, thanks!

mbargull:
I think #7567 should fix the `post-link.sh`/`pre-unlink.sh` scripts' output.

PR #7556, opened 2018-02-07
updated to v360 and pointed to github instead of hgdownload. 

karl616:
… if osx will build as it is

### DIFF
--- a/recipes/ucsc-bedgraphtobigwig/357/build.sh
+++ b/recipes/ucsc-bedgraphtobigwig/357/build.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+mkdir -p "$PREFIX/bin"
+if [ "$(uname)" == "Darwin" ]; then
+    cp bedGraphToBigWig "$PREFIX/bin"
+else
+    export MACHTYPE=x86_64
+    export BINDIR=$(pwd)/bin
+    mkdir -p "$BINDIR"
+    (cd kent/src/lib && make)
+    (cd kent/src/htslib && make)
+    (cd kent/src/jkOwnLib && make)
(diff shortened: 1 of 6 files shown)


daler:
Thanks @karl616! Great to see it's building now on macOS, and thanks for changing to github versions. This looks good. I'm not sure we need the old version subdirectory in this case.

To update all the `ucsc-*` tools, we regenerate recipes with https://github.com/bioconda/bioconda-recipes/tree/master/scripts/ucsc, and then we use the [bulk branch](https://bioconda.github.io/build-system.html#the-bulk-branch) since the building generally takes a while.

Would you please take a look at those UCSC scripts and templates to update them with your changes? You can put them in this PR or in another one (or wait till this weekend and I'll work on it). Then I'll start a run on the bulk branch to update everything.

daler:
By the way, gcc on macOS in conda is kind of in flux . . . for now, can you please modify to use `llvm` on osx as indicated here: https://github.com/bioconda/bioconda-recipes/pull/7568#issuecomment-364389661

karl616:
Hi @daler, I missed the script folder and went on with sed/awk. It is passing through circleci build locally at the moment.

The templating is much neater. I will have a look. It might be that I don't have time to finish today, but I could continue on Tuesday (Monday is holiday). I'll make a new PR of the template changes.

With regard to macOS, my only proof that it works is the circleci tests.... Is that enough?

daler:
@karl616 thanks. My only proof for macOS is circleci tests as well, I don't have access to a dev mac machine.

karl616:
Replaced by new PR:

https://github.com/bioconda/bioconda-recipes/pull/7592

@daler Thanks for the feedback!

PR #7557, opened 2018-02-07
Adding svdb v1.0.7

holtgrewe:

### DIFF
--- a/recipes/svdb/build.sh
+++ b/recipes/svdb/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+$PYTHON setup.py install --single-version-externally-managed --record record.txt
+
+# Add more build steps here, if they are necessary.
+
+# See
+# http://docs.continuum.io/conda/build.html
+# for a list of environment variables that are set during the build process.
--- a/recipes/svdb/meta.yaml
+++ b/recipes/svdb/meta.yaml
@@ -1,0 +1,49 @@
+{% set version = "1.0.7" %}
+{% set sha256sum = "f7ca1cd7b1decae4f993a6c486d614cbd7a5b29b591630376623d04fdabd2876" %}
+
+package:
+  name: svdb
+  version: "{{ version }}"
+
+source:
+  fn: svdb-{{ version }}.tar.gz
+  url: https://github.com/J35P312/SVDB/archive/SVDB-{{ version }}.tar.gz
+  sha256: "{{ sha256sum }}"
+
+  # patch not required beyond 1.0.7
+  patches:
+  - svdb.patch
+
+build:
+  skip: True  # [not py27]
+  number: 0
+
+entry_points:
+- svdb = svdb.__main__:main
+
+requirements:
+  build:
+  - cython
+  - python
+  - numpy
+  - scipy
+  - scikit-learn
+  # missing dependency of scipy
+  - libgcc
+
+  run:
+  - python
+  - numpy
+  - scipy
+  - scikit-learn
+  # missing dependency of scipy
+  - libgcc
+
+test:
+  commands:
+  - svdb --help
+
+about:
+  home: https://github.com/J35P312/SVDB
+  license: MIT
+  summary: 'structural variant database software'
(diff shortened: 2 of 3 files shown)


holtgrewe:
Hm, @bioconda/python does anyone see what's wrong here? I'm clueless. Building appears to fail for `scikit-learn` but it's in my `meta.yaml`...

mbargull:
What happened here is that `conda` installed `scikit-learn 0.19.1`, but `setuptools` wants `scikit-learn ==0.15.2`. You'll have to use
```bash
python setup.py install --single-version-externally-managed --record record.txt
```
to prevent `setuptools` from trying to download dependencies by itself. And it's best to also use
```yaml
    - scikit-learn ==0.15.2
```
in `meta.yaml` as well, if that version constraint is really needed. If it's not needed, you should patch `setup.py` to remove that constraint.

recipes/svdb/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+{% set md5sum = "2764c03866f3aa827bec3fcc50fa1742" %}
+
+package:
+  name: svdb
+  version: "{{ version }}"
+
+source:
+  fn: svdb-{{ version }}.tar.gz
+  url: https://github.com/J35P312/SVDB/archive/SVDB-{{ version }}.tar.gz
+  md5: "{{ md5sum }}"
```
mbargull: It's [usually preferred to use `sha256`](http://bioconda.github.io/guidelines.html#hashes).

recipes/svdb/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+build:
+  skip: True  # [not py27]
+  number: 0
```
mbargull: Can you add
```yaml
  entry_points:
    - svdb = svdb.__main__:main
```
to the build section.

recipes/svdb/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  # missing dependency of scipy
+  - libgcc
+
+test:
+  commands:
+  - svdb --help
+
+about:
+  home: https://github.com/J35P312/SVDB
+  license: MIT
```
mbargull: I couldn't find the license declaration in `setup.py` or `README.md` -- are you sure it's MIT licensed?

recipes/svdb/meta.yaml
mbargull: It's [usually preferred to use `sha256`](http://bioconda.github.io/guidelines.html#hashes).

recipes/svdb/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+$PYTHON setup.py install
```
mbargull: Use
```bash
python setup.py install --single-version-externally-managed --record record.txt
```

recipes/svdb/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  skip: True  # [not py27]
+  number: 0
+
+requirements:
+  build:
+  - cython
+  - python
+  - numpy
+  - scipy
+  - scikit-learn
```
mbargull: Use
```yaml
    - scikit-learn ==0.15.2
```
or patch `setup.py`.

recipes/svdb/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  # missing dependency of scipy
+  - libgcc
+
+  run:
+  - python
+  - numpy
+  - scipy
+  - scikit-learn
```
mbargull: Use
```yaml
    - scikit-learn ==0.15.2
```
or patch `setup.py`.

recipes/svdb/meta.yaml
mbargull: Ah, found the [issue you created](https://github.com/J35P312/SVDB/issues/12), thanks!

recipes/svdb/meta.yaml
mbargull: Argh, I wasn't notified of your update so I didn't review them/approve them (maybe GitHub doesn't notify if no commit was added and force pushed -- strange thing...). That `entry_points:` should've been part of the build section.

holtgrewe: Gosh, ok, let's continue this on #7565?

PR #7563, opened 2018-02-08
Add analysis tool collection from molpopgen

pbordron:

### DIFF
--- a/recipes/libsequence/1.8.4/meta.yaml
+++ b/recipes/libsequence/1.8.4/meta.yaml
@@ -29,7 +29,8 @@ requirements:
 
 test:
   commands:
-    - ls -ls $PREFIX/lib/libsequence.so
+    - ls -ls $PREFIX/lib/libsequence.so # [linux]
+    - ls -ls lib/libsequence.dylib      # [osx]
 
 about:
   home: http://http://molpopgen.github.io/libsequence/
--- a/recipes/libsequence/1.8.4/meta.yaml
+++ b/recipes/libsequence/1.8.4/meta.yaml
@@ -20,7 +20,6 @@ requirements:
     - zlib {{ CONDA_ZLIB }}*
     - tbb
     - htslib {{CONDA_HTSLIB}}*
-    - boost {{CONDA_BOOST}}*
   run: 
     - libgcc
     - zlib {{ CONDA_ZLIB }}*
--- a/recipes/libsequence/1.8.4/meta.yaml
+++ b/recipes/libsequence/1.8.4/meta.yaml
@@ -29,8 +29,8 @@ requirements:
 
 test:
   commands:
-    - ls -ls $PREFIX/lib/libsequence.so # [linux]
-    - ls -ls lib/libsequence.dylib      # [osx]
+    - ls -ls $PREFIX/lib/libsequence.so    # [linux]
+    - ls -ls $PREFIX/lib/libsequence.dylib # [osx]
 
 about:
   home: http://http://molpopgen.github.io/libsequence/
--- a/recipes/libsequence/1.8.4/meta.yaml
+++ b/recipes/libsequence/1.8.4/meta.yaml
@@ -29,7 +29,7 @@ requirements:
 
 test:
   commands:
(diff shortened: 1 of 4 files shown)


pbordron:
@bioconda/core @bgruening 
I get the following error when circle.ci try deploy the conda test environment. The recipe build without error on my comuter. Someone has an idea if it is related to the recipe or circle.ci ?

```
OSError: Failed to copy /opt/recipe/lib/libsequence.so to /opt/conda/conda-bld/libsequence_1518093214223/test_tmp/lib/libsequence.so.  Error was: Command 'cp -a /opt/recipe/lib/libsequence.so /opt/conda/conda-bld/libsequence_1518093214223/test_tmp/lib/libsequence.so' returned non-zero exit status 1.
```

pbordron:
OK find why.
A problem in my recipe.

recipes/molpopgen-analysis/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - command -v kimura80
+    - command -v polydNdS
+    - command -v sharedPoly
+    - command -v descPoly
+    - command -v HBKpermute
+    - command -v MKtest
+    - command -v rsq
+
+about:
+  home: http://http://molpopgen.github.io/libsequence/
```
bgruening: Is this link correct?

recipes/libsequence/1.8.4/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  version: {{ version }}
+
+source:
+  fn:  {{ version }}-1.tar.gz 
+  url: http://github.com/molpopgen/libsequence/archive/{{ version }}-1.tar.gz
+  sha256: {{ hash }}
+
+build:
+  number: 0
+  string: "htslib{{CONDA_HTSLIB}}_{{PKG_BUILDNUM}}"
```
bgruening: boost should be also added here

pbordron: Boost was a dependency of the `make check` step that keeps falling on macos x.

recipes/molpopgen-analysis/meta.yaml
pbordron: The recipe fail if the link is incorrect, meaning the build was unsuccessfull

pbordron:
ping @bgruening

recipes/molpopgen-analysis/meta.yaml
bgruening: But wouldn't this link be better? https://github.com/molpopgen/analysis

pbordron: Copy/paste error :p

bgruening:
@pbordron thanks!

molpopgen:
I'd like to request that this recipe be removed (if that is possible).  I'm the author of the package in question, and it is deprecated/end of life'd.  It has no place is a current-era workflow, and it will no longer track updates to libsequence.  As per the README:

As of July 13, 2015, this package is "retired" and will no longer be supported or maintained.

The programs in this package have their origin in the days of using Sanger sequencing to study genetic variation. The movement towards short-read-based resequencing means that software like this may not be applicable to modern data sets based on modest coverage.

This software did its job for a decade or more, and still has some use in certain applications, but it is time to stop supporting it.

pbordron:
@molpopgen @bgruening  I created this package to allow some users from our cluster to use thoses scripts. They know thoses scripts are "retired", but they still need them. I used conda to get rid of many constraints existing in our cluster and I think it is still a good idea to propose the recipe in order to have the right version of the libsequence depency.

We can modify the description in order to add a warning that the package is not maintained nor updated anymore.

molpopgen:
> They know thoses scripts are "retired", but they still need them.

It is hard to see how this is the case.  Unless people are still collecting Sanger data, results based on this package are very likely to simply be wrong.  I'm guessing that the "need" is because of some existing workflow that should be updated to tools more suitable for NGS data.

Any warning should also reflect what I said above re: incorrect results.

PR #7568, opened 2018-02-08
Add llvm as runtime req for osx (libc++ fail without it)

kastman:

When running MacOS 10.13 (at least), llvm is required at runtime - without it you get the standard:

    dyld: Library not loaded: @rpath/libc++.1.dylib
      Referenced from: /Users/kastman/miniconda/envs/agape/bin/sga
      Reason: image not found
    Abort trap: 6

### DIFF
--- a/recipes/sga/meta.yaml
+++ b/recipes/sga/meta.yaml
@@ -24,7 +24,7 @@ requirements:
     - zlib {{ CONDA_ZLIB }}*
   run:
     - libgcc  # [not osx]
-    - llvm  # [osx]
+    - libcxx  # [osx]
     - sparsehash
     - bamtools
     - zlib {{ CONDA_ZLIB }}*
--- a/recipes/sga/meta.yaml
+++ b/recipes/sga/meta.yaml
@@ -10,7 +10,7 @@ source:
   sha256: 1b18996e6ec47985bc4889a8cbc3cd4dd3a8c7d385ae9f450bd474e36342558b
 
 build:
-  number: 0
+  number: 1
 
 requirements:
   build:
@@ -24,6 +24,7 @@ requirements:
     - zlib {{ CONDA_ZLIB }}*
   run:
     - libgcc  # [not osx]
+    - llvm  # [osx]
     - sparsehash
     - bamtools
     - zlib {{ CONDA_ZLIB }}*


mbargull:
Interesting. So, @bioconda/osx, should we recommend adding `llvm` to the run requirements for `[osx]` in http://bioconda.github.io/guidelines.html#c-c or is this some kind of special case here?

mbargull:
But note (regarding adding it to the guidelines) that `llvm` is a beefy package:
```
$ CONDA_SUBDIR=osx-64 conda create --dry-run -qnx llvm
Solving environment: ...working... done

## Package Plan ##

  environment location: /home/maba/code/conda/conda-4.4/envs/x

  added / updated specs: 
    - llvm


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    cctools-895                |       h7512d6f_0         1.5 MB
    llvm-4.0.1                 |       hc748206_0       136.7 MB
    llvm-lto-tapi-4.0.1        |       h6701bc3_0        11.7 MB
    ld64-274.2                 |       h7c2db76_0         2.6 MB
    libcxxabi-4.0.1            |       hebd6815_0         149 KB
    libcxx-4.0.1               |       h579ed51_0         957 KB
    ------------------------------------------------------------
                                           Total:       153.6 MB

    cctools:       895-h7512d6f_0  
    ld64:          274.2-h7c2db76_0
    libcxx:        4.0.1-h579ed51_0
    libcxxabi:     4.0.1-hebd6815_0
    llvm:          4.0.1-hc748206_0
    llvm-lto-tapi: 4.0.1-h6701bc3_0
```

kastman:
@mbargull To be honest, I'm not sure; this is the first time I've run into a bioconda recipe failing on OSX because it's missing libc++. I briefly looked into compiling a static library at build time, but didn't come across any easy options and adding it at runtime worked for me, and (as recommended) libgcc is included as a runtime dep for unix, so I figured that was an acceptable option for osx as well.

I agree it's pretty heavy; do you think we should look harder into static compiling it?

mbargull:
@kastman, does `libcxx` instead of `llvm` also work? That seems a little more sane

mbargull:
> I agree it's pretty heavy; do you think we should look harder into static compiling it?

Nah. Better try out `libcxx` first. That doesn't add a lot:
```
$ CONDA_SUBDIR=osx-64 conda create --dry-run -qnx libcxx
Solving environment: ...working... done

## Package Plan ##

  environment location: /home/maba/code/conda/conda-4.4/envs/x

  added / updated specs: 
    - libcxx


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    libcxx-4.0.1               |       h579ed51_0         957 KB
    libcxxabi-4.0.1            |       hebd6815_0         149 KB
    ------------------------------------------------------------
                                           Total:         1.1 MB

    libcxx:    4.0.1-h579ed51_0
    libcxxabi: 4.0.1-hebd6815_0
```

kastman:
Yeah, I think you're right - llvm was working _because_ it was installing libcxx. Updating the recipe now.

mbargull:
Great. Okay, then let's rephrase the question for the OSX team:
@bioconda/osx, should we recommend to add `- libcxx  # [osx]` to the run requirements in http://bioconda.github.io/guidelines.html#c-c?

yesimon:
No because it's not always needed. Recipes using `gcc` on OS X don't use it.

mbargull:
But those wouldn't use `llvm` anyway. So, the question is, if we should recommend the `build: [llvm]`+`run: [libcxx]`-combo for recipes that use `llvm`. Given your objection, we would have to clarify that `llvm` should only be added if `gcc` isn't used..

kastman:
That makes sense to me, with `libcxx` as the [osx] corollary to `libgcc [not osx]`. Any idea why this hasn't come up for other `llvm` osx recipes?

mbargull:
@kastman, for this PR you'll have to merge in `master` because your branch isn't yet updated to use the new CI.

kastman:
@mbargull I rebased from upstream/master and the checks are now passing. 

I didn't realize but I was using an old master from a week ago; the project must be moving quick!

jenzopr:
This is great! I guess, it could solve the issues in #7549 and #7551

bgruening:
@mbargull am I right that we will go in the near future (cb3) to GCC only? If so we can do this here as well and drop llvm entirely?

mbargull:
I assume that the plan for `conda-forge` is to use `{{ compiler("cxx") }}` which will use `llvm`/`clang`:
```
|(base) $ conda-build --version
conda-build 3.4.1
```
```
|(base) $ cat ./meta.yaml
package:
  name: pkg
requirements:
  build:
    - {{ compiler("cxx") }}
```
```
|(base) $ CONDA_SUBDIR=linux-64 conda render .
package:
    name: pkg
requirements:
    build:
        - gcc_impl_linux-64 7.2.0 hc5ce805_2
        - gxx_impl_linux-64 7.2.0 hd3faf3d_2
        - binutils_linux-64 7.2.0 25
        - libstdcxx-ng 7.2.0 h7a57d05_2
        - gcc_linux-64 7.2.0 25
        - gxx_linux-64 7.2.0 25
        - binutils_impl_linux-64 2.28.1 h04c84fa_2
    run:
        - libstdcxx-ng >=7.2.0
        - libgcc-ng >=7.2.0
extra:
    copy_test_source_files: true
    final: true

```
```
|(base) $ CONDA_SUBDIR=osx-64 conda render .
setting build platform. This is only useful when pretending to be on another another platform, such as for rendering necessary dependencies on a non-native platform. I trust that you know what you're doing.
package:
    name: pkg
requirements:
    build:
        - clangxx_osx-64 4.0.1 h22b1bf0_8
        - libcxx 4.0.1 h579ed51_0
        - llvm-lto-tapi 4.0.1 h6701bc3_0
        - libcxxabi 4.0.1 hebd6815_0
        - ld64 274.2 h7c2db76_0
        - clang 4.0.1 h662ec87_0
        - compiler-rt 4.0.1 h5487866_0
        - cctools 895 h7512d6f_0
        - clangxx 4.0.1 hc9b4283_0
        - llvm 4.0.1 hc748206_0
        - clang_osx-64 4.0.1 h1ce6c1d_8
    run:
        - libcxx >=4.0.1
extra:
    copy_test_source_files: true
    final: true

```
. So to be compatible I guess we should do the same, or do you think otherwise? I think, for now, it makes sense to recommend `gcc`+`libgcc` on Linux for C/C++ and `llvm`+`libcxx` for C++ on macOS (unless `gcc` is be used for some reason, then also `gcc`+`libgcc`) -- if we compile only C  but not C++ on macOS, then `libcxx` is of course not needed. After the `conda-build 3`-switch replace all that by `{{ compiler("cxx") }}` (or, e.g., `{{ compiler("c") }}` if only C is used).
In comprehensible form:
* now:
  *  for C++ builds, use
    ```yaml
    requirements:
      build:
        - gcc  # [not osx]
        - llvm  # [osx]
      run:
        - libgcc  # [not osx]
        - libcxx  # [osx]
    ```
  *  for C-only builds, use
    ```yaml
    requirements:
      build:
        - gcc  # [not osx]
        - llvm  # [osx]
      run:
        - libgcc  # [not osx]
    ```
* in future:
  *  for C++ builds, use
    ```yaml
    requirements:
      build:
        - {{ compiler("cxx") }}
    ```
  *  for C-only builds, use
    ```yaml
    requirements:
      build:
        - {{ compiler("c") }}
    ```

kastman:
I’m confused after reading the cb3 doc - is there a compiler type flexible function for the runtime requirement? Or does the platform target mean it’s not needed anymore? Is all of conda-forge / bioconda switched to cb3? Or is that currently in process?

mbargull:
First and foremost: Let's merge 😄 -- no need to postpone that due to discussions on future changes!

Neither conda-forge nor Bioconda use `conda-build 3` right now. conda-forge is working for quite some time on switching to the new version, so I hope it won't take much longer until that's good to go. Bioconda can/will then switch shortly after, I think. So, yeah it's a WIP.
Regarding the other question, `conda-build` uses that `compiler(...)` function to choose which compiler packages to use for the build. The run requirements are actually introduced via another new feature, named [`run_exports`](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#pin-downstream), which is used in the new compiler packages, e.g.:
- https://github.com/AnacondaRecipes/aggregate/blob/master/ctng-compilers-feedstock/recipe/meta.yaml
- https://github.com/AnacondaRecipes/aggregate/tree/master/ctng-compilers-activation-feedstock/recipe/meta.yaml

But I have to admit that I also still have to do some reading on this to do and thus don't know all nuances, yet. I'll have to do that and then summarize some of the important things for Bioconda soon.

PR #7571, opened 2018-02-08
Added recipe for msaboot

matthew-fogel:


recipes/msaboot/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py install
```
mbargull: Please use
```bash
$PYTHON setup.py install --single-version-externally-managed --record record.txt
```

recipes/msaboot/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  name: msaboot
+  version: "0.1.0"
+
+source:
+  url: https://github.com/phac-nml/msaboot/archive/v0.1.0.tar.gz
+  sha256: 41a34899b4c768abec4bf4e15d5a0a08451208fab9403b00454f1b7203971fa6
+
+build:
+  number: 0
+  skip: True # [win]
```
mbargull: You can remove that `skip` line since there are no Windows builds on Bioconda anyway.

recipes/msaboot/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+requirements:
+  build:
+    - python
+    - numpy
+    - biopython
```
mbargull: You can remove `numpy` and `biopython` from the build requirements, and please add `setuptools` instead.

recipes/msaboot/meta.yaml
mbargull: Please add
```yaml
  entry_points:
    - msaboot = msaboot.msaboot:main
```
to the `build:` section.

recipes/msaboot/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+test:
+  commands:
+    - msaboot -h | grep "bootstraps FASTA input data"
+
+about:
+  home: https://github.com/phac-nml/msaboot
+  license: Apache 2.0
```
mbargull: Can you also add
```yaml
  license_file: LICENSE.txt
```

recipes/msaboot/meta.yaml
```diff
@@ -0,0 +1,31 @@
+package:
+  name: msaboot
+  version: "0.1.0"
+
+source:
+  url: https://github.com/phac-nml/msaboot/archive/v0.1.0.tar.gz
```
mbargull: Can you also add
```yaml
  fn: msaboot-0.1.0.tar.gz
```
to the `source:` section?

recipes/msaboot/meta.yaml
matthew-fogel: Thank you. Does setuptools have to be listed first or is the order that I have listed okay/

recipes/msaboot/meta.yaml
matthew-fogel: Thank you, Please let me know if there is anything else I should change.

mbargull:
Perfect! Thanks @matthew-fogel

PR #7579, opened 2018-02-09
Change path quoting in case of special chars like '@'

peterjc:
If the recipe is used with a conda environment with an at sign in the directory name, given the double quotes Perl would try to interpret this - and fail.

Note given the way the recipe works there will be no dollar signs present to worry about.

This is intended to fix a bug seen under Galaxy which used the form ``__prefix@_uv_`` in an environment name, and thus in the path. Perl didn't like this:

```
Global symbol "@_uv_" requires explicit package name (did you forget to declare "my @_uv_"?)
```

Cross reference https://github.com/galaxyproject/galaxy/pull/5498


peterjc:
Going further, we could change the conda packaging testing to deliberately use this kind of directory naming to discover (and thus fix) any other packages which suffer from this limitation?

PR #7593, opened 2018-02-09
Antismashsmiles

zachcp:

Update for the new 4.1 Antismash release. Include a few patches to fix OSX including removing the SMILES generation code and allowinf *.info files in the MANIFEST.in

### DIFF
--- a/recipes/antismash/manifest.patch
+++ b/recipes/antismash/manifest.patch
@@ -1,0 +1,12 @@
+--- MANIFEST.in
++++ MANIFEST.in
+@@ -4,7 +4,7 @@ include run_antismash.py
+ include download_databases.py
+ 
+ recursive-include antismash *.py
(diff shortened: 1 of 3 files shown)


bgruening:
@zachcp any reason to remove the smiles code?

zachcp:
@bgruening errors were being reported that were caused by the Indigo toolkit. Currently the binaries are distributed by Antismash but they don't seem to work for many users. By inactivating the smiles generation we are ensuring functionality of 99% of what Antismash has to offer but sacrificing very little.

A better long term solution would be to package the [Indigo toolkit](https://github.com/epam/indigo) or, as is the plan, move to RDKit/CDK/some other structure generation toolkit.

See issue here:
https://bitbucket.org/antismash/antismash/issues/55/anaconda-on-macosx

bgruening:
👍 for moving to RDKit :)
Thanks for the explanation.

PR #7606, opened 2018-02-10
Bumped GATK4 version

micknudsen:

### DIFF
--- a/recipes/gatk4/meta.yaml
+++ b/recipes/gatk4/meta.yaml
@@ -21,7 +21,7 @@ requirements:
   build:
   run:
     - openjdk >=8,<9
-    - python
+    - python >0
 
 test:
   commands:
--- a/recipes/gatk4/meta.yaml
+++ b/recipes/gatk4/meta.yaml
@@ -1,4 +1,4 @@
-{% set version="4.0.1.1" %}
+{% set version="4.0.1.2" %}
 
 about:
   home: https://www.broadinstitute.org/gatk/
@@ -15,7 +15,7 @@ build:
 source:
   fn: gatk-{{ version }}.zip
   url: https://github.com/broadinstitute/gatk/releases/download/{{ version }}/gatk-{{ version }}.zip
-  sha256: 11abae169b73946d52e7831d8ada8b01ab05c517bb60abbf4046eeed3c696065
+  sha256: d667c08ec44f4dc2029d00ca16cfcfe7850ae9bfdcdd6e35f3048b8e7e83647b
 
 requirements:
   build:
--- a/recipes/gatk4/meta.yaml
+++ b/recipes/gatk4/meta.yaml
@@ -21,6 +21,7 @@ requirements:
   build:
   run:
     - openjdk >=8,<9
+    # constrain python to prevent multiple builds with conda-build 2.1.x
     - python >0
 
 test:


mbargull:
Thanks for the update @micknudsen.

Couple of questions for @bioconda/core:
This is a, comparatively, storage-heavy package that, if I see it correctly, gets build 6 times with the same content (for each supported Python version and for Linux as well as macOS). As it currently stands, each `gatk4` release uses \~ 1.5 GB instead of the \~ 250 MB it would actually need. As of now, all builds of `gatk4` on the channel need \~ 14 GB..
  * Should we try to make this `noarch: generic`? (TBH, I haven't tried `generic` with `conda-build` before.) We have to `skip: True  # [osx]` for this then, of course -- but I'm not sure whether our build infrastructure has any other problem with `noarch` packages..
  * If not `noarch: generic`, we could change it to depend on `- python >0`. That way we would at least only build twice instead of six times (note that that `>0` workaround is only needed for `conda-build 2.1.x`). `gatk4` isn't really a Python package but merely uses it, hence no need to build for multiple versions.
  * @chapmanb: In https://github.com/bioconda/bioconda-recipes/pull/4835 you added
    ```sed -i.bak 's#!/usr/bin/env python#!/opt/anaconda1anaconda2anaconda3/bin/python#' gatk```
    . I think we should remove that line, because
        1. `/opt/anaconda1anaconda2anaconda3` is an implementation detail, hence it should rather be replaced by `$PREFIX`.
        2. But overall, on Bioconda, we often recommend to explicitly use `/usr/bin/env python` instead of the full path to support running environments with long prefixes. I personally don't like the `env` workaround either, but, esp. on Linux, shebang lines have very limited capabilities. I recommend everyone to read https://www.in-ulm.de/~mascheck/various/shebang/ (or at least take a look at the [included table](https://www.in-ulm.de/~mascheck/various/shebang/#results)).

chapmanb:
Marcel;
Thanks much for looking at this. It would be awesome if we could get this building with `noarch: python`. As you said, it doesn't need a specific python version, just python available to run the wrapper script. I've tried in the past to get this working and couldn't find enough information to make it happen. If you know how do it, please do convert over. I'd be excited to learn how.

This uses `/opt/anaconda1anaconda2anaconda3` because it gets converted into the installation path on the machine. So this gives you a direct reference to the anaconda installed python. $PREFIX would give you a hard coded path on the build machine and `/usr/bin/env` gets the first python in your PATH so you can end up picking up another python on the machine unless your anaconda directory is first. This can break things on systems where you symlink from the anaconda directory into another path to avoid injecting everything from anaconda into your PATH. I've used the `anaconda1anaconda2anaconda3` directory a few times for this approach.

Thanks again for looking into and improving this recipe.

mbargull:
> get this building with `noarch: python`

It would actually be better to use `noarch: generic` since `noarch: python` does some special handling for Python packages, which `gatk4` isn't one of. I'll try to find some time to try test `generic` with this one.

> $PREFIX would give you a hard coded path on the build machine

Actually `conda-build` will replace  all (expanded) `$PREFIX` strings with `conda.base.constants.PREFIX_PLACEHOLDER` (i.e., the string you used) which `conda` will then replace with the installation environments prefix. Although you can assume that string won't be changed in foreseeable future, I'd treat it as an implementation detail and use `$PREFIX` instead :wink:. For some other cases (compiled binaries) it's even crucial to use `$PREFIX`.

> `/usr/bin/env` gets the first python in your PATH so you can end up picking up another python on the machine unless your anaconda directory is first. This can break things on systems where you symlink from the anaconda directory into another path to avoid injecting everything from anaconda into your PATH.

The supported use case for Conda is to run programs in activated environments. IMO, you shouldn't rely on being able to run them without prior activation (due to *several* reasons). If you want to run those programs, I'd advice to use (e.g., in a shell) `(. activate env-name && gatk)` (note the subshell `( )`usage) instead of `gatk`. Also note that you'll get different behavior if you use
```bash
conda create gatk -n short-env-name
```
and
```bash
conda create gatk -n long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-long-env-name
```
as `conda` will (for the better or worse) replace your shebangs with `#!/usr/bin/env` ones.
I could keep on babbling about some more implementation details and corner cases, etc. Overall it's just that shebangs are not generally suited for modern use cases and the current workarounds are all very unsatisfying. I'm planning on revisiting the logic around shebangs in `conda` itself and proposing some additional workarounds -- not necessarily soon, though.
In my eyes, the *most consistent* workaround for now is to use `/usr/bin/env` (even though it has its downsides) and so we should stick with that.

mbargull:
I couldn't find the time to test out `noarch: generic`, so let's just use `python >0` for now. That'll at least reduce the number of builds by a factor of 3.

micknudsen:
What is the difference between `python >0` and `python`? Does the `>0` have a special meaning?

mbargull:
It's just a (dirty) workaround for `conda-build 2.1.x` to make it not pin to a specific `python` version which would create multiple builds (for `2.7`, `3.5`, `3.6`). `conda-build 3` will do the right thing on it's own and not pin `python` unless it's also a build requirement.
(And that workaround should be used with caution as it'll cause trouble with real Python packages; so better not apply it, unless you certainly know what you are doing 😉)

chapmanb:
Marcel;
Thanks for the python trick here and for the comments.

Thanks also for the PREFIX trick, I'll try that going forward for referencing the work directory. It does sound more future proof.

I'm afraid I'll have to disagree about the activated environments. I've tried environmental activation for a while previously back in the day of virtualenv and friends, and it was very difficult to support for external users of our tools. They'd often end up in problematic environments that were hard to reproduce and debugging what was going on remotely was quite difficult. The symlinking and PATH changing are pretty easy to apply and see what is happening. I'd definitely vote against requiring activation if we get to vote, and stick with strategies that work in both cases.

Thanks again for all this work and thanks for listening to my suggestions and usage.

mbargull:
@micknudsen, thanks for the update and sorry for the delay!

@chapmanb, for now I didn't change the `build.sh` as I don't want to postpone the merge of this PR any longer -- given the latest patch release frequency for GATK, we could have more than enough possibilities to do so 😆. As for activated environments, it's just that `conda` has some features (activation scripts) that sometimes demand prior environment activations. If you have any suggestions or stumble upon problematic use cases concerning Conda environment activation, please go ahead and ping me -- I'm involved with `conda`'s development and not infrequently around activation stuff (which, btw, was quite overhauled in `conda 4.4`!).

PR #7608, opened 2018-02-10
Update pysam to 0.14

kyleabeauchamp:

cc @AndreasHeger


kyleabeauchamp:
So pysam 0.14 wraps HTSLib 1.7, samtools 1.7, and BCFtools 1.6 (bcftools 1.7 has yet to be released).  Thus, I've rolled back the use of the global `CONDA_HTSLIB` variable in the pysam recipe.

bgruening:
@kyleabeauchamp please look at the bulk branch. We have increased the HTSLIB env var there already. If you target the bulk I can merge it.

kyleabeauchamp:
I think I want to avoid using the HTSLIB env variable because this recipe (v0.14) depends on a heterogenous mixture of versions (1.6 and 1.7).

kyleabeauchamp:
See, e.g., https://github.com/pysam-developers/pysam/pull/609, for more reading.

dpryan79:
Anyone object to this getting merged in? It certainly looks good to go.

bgruening:
My only concern is that this will not make it into many environments because of the strict pinning.
@kyleabeauchamp please put the ENV vars back in as soon as bcftools are released. 

Thanks for the update!

AndreasHeger:
Thanks @kyleabeauchamp . Apologies for the version mix, but I thought it would be important to get a htslib 1.7 version out despite bcftools not at the same version.

kyleabeauchamp:
I totally agree.  Thanks!

PR #7660, opened 2018-02-12
Bloomfiltertrie

Colelyman:

As per the guidelines, I would like to request a review from @bioconda/core because this is my first recipe. Thanks!


druvus:
@Colelyman An alternative here would be enabling openmp support again and use gcc for both linux and osx as clang doesn't support it.

mbargull:
Would depending on
```yaml
- llvm
- llvm-openmp
```
help? I haven't had the change to try that one yet.

recipes/bloomfiltertrie/popcnt.patch
```diff
@@ -0,0 +1,22 @@
+diff --git a/src/intersection.c b/src/intersection.c
+index 56599bb..443418e 100755
+--- a/src/intersection.c
++++ b/src/intersection.c
+@@ -137,7 +137,7 @@ uint32_t* intersection_uint32_SIMD(uint32_t* list_a, uint32_t* list_b) {
+         //[ copy out common elements
+         __m128i p = _mm_shuffle_epi8(v_a, shuffle_mask[mask]);
+         _mm_storeu_si128((__m128i*)&C[count], p);
+-        count += _mm_popcnt_u32(mask);
++        count += __builtin_popcountl(mask);
```
mbargull: Is this a behavioral change? The original code does a 32-bit pop-count, your change would make that 64-bit.

Colelyman:
Unfortunately depending on `llvm-openmp # [osx[` didn't work. @druvus would removing the `llvm` dependency completely have the MacOS use `gcc` or is there something different that I should do?

Colelyman:
@druvus eliminating the `llvm` dependency worked! Thank you

recipes/bloomfiltertrie/popcnt.patch
mbargull: Due to compatibility reasons, I don't think we can use SSE 4.2 yet; though, TBH, I'm not 100 % sure what the expected minimum capabilities/extensions we target are.
Here, I think you can just use `__builtin_popcount` (minus the `l`ong); that's also what GCC would do anyway if SSE >=4.1 and POPCNT were enabled: https://github.com/gcc-mirror/gcc/blob/gcc-4_8_5-release/gcc/config/i386/popcntintrin.h#L35.

mbargull:
> Unfortunately depending on `llvm-openmp # [osx]` didn't work.

Was worth a shot -- thanks for trying it out. Using `gcc` should be fine.

Colelyman:
@mbargull anything else that I should do before I squash and merge?

mbargull:
Looks good, thank you @Colelyman!

PR #7670, opened 2018-02-13
Update ont-tombo package to version 1.2.

marcus1487:

### DIFF
--- a/recipes/ont-tombo/meta.yaml
+++ b/recipes/ont-tombo/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "ont-tombo" %}
-{% set version = "1.1.1" %}
-{% set hash = "e481e7ff88d2c752cb39ddf25e5b37871748e4178a285053b7dde19f2566c93b" %}
+{% set version = "1.2" %}
+{% set hash = "db3258d7f4e3d9b60d66d457aa8821ac8d0c2058a2eaa6a79e99dff54de8c18a" %}
 
 package:
   name: {{ name|lower }}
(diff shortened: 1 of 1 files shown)


recipes/ont-tombo/meta.yaml
```diff
@@ -15,39 +15,37 @@ build:
...
-    - bwa
-    - minimap2
+    - r-gridextra
+    - mappy
+    - pyfaidx
+    - future
 
   run:
     - python
+    - setuptools
```
mbargull: This should be `- setuptools >=18.0`, acc. to https://github.com/nanoporetech/tombo/blob/1.2/setup.py#L61

recipes/ont-tombo/meta.yaml
```diff
@@ -15,39 +15,37 @@ build:
...
     - h5py
-    - rpy2 >=2.4.2
+    - rpy2
```
mbargull: Given https://github.com/nanoporetech/tombo/blob/1.2/setup.py#L37-L40, this should be
```yaml
    - rpy2  # [py3k]
    - rpy2 <=2.8.6  # [not py3k]
```

recipes/ont-tombo/meta.yaml
```diff
@@ -15,39 +15,37 @@ build:
...
   number: 0
-  skip: True  # [not py27]
 
 requirements:
   build:
     - python
     - setuptools
     - h5py
-    - rpy2 >=2.4.2
+    - rpy2
```
mbargull: see below

mbargull:
Thanks @marcus1487!

marcus1487:
@mbargull I am running into some issues with users hitting a clash between python versions and r 3.4.1. There are certainly workarounds, but it does present a hurdle for users. It seems as though this comes down to an issue withe the libedit package https://github.com/conda/conda/issues/6574 .

I am wondering why/how this tombo release was linked to r-base version 3.4.1 when a newer version is available and AFAICT there are no conflicts with ont-tombo and this r-version. The r-base in the bioconda recipe does not pin a version, so I guess I'm just trying to figure out how to specify this updated r-base version in this recipe appropriately. Thanks!

mbargull:
There are a couple of package, e.g., `python` and `r-base`, that get pinned to exact versions. With `python` we currently build (unless explicitly skipped) for 3 different versions. For `r-base` we only use one version, i.e., `3.4.1` as of now; we follow `conda-forge` on this one where there is no newer version available for now. See https://github.com/bioconda/bioconda-recipes/blob/master/scripts/env_matrix.yml for the currently pinned versions.
We do this for compatibility reasons between different R packages. It's not an optimal solution and may change/improve in future, but it isn't something we can change yet.
Do your users use the `python` and `r-base` packages from `conda-forge` or do they mix those with ones from `defaults`? For compatibility, esp. regarding `r-base`, it's best to make sure the packages from `conda-forge` are used, i.e., let them try to explicitly install with `conda create -c bioconda -c conda-forge ont-tombo r-base python`.

marcus1487:
Interesting. I had previously encountered R packages and particularly `rpy2` (which were on the `r` channel) that were not found in `conda-forge`. These R package installs caused error modes for which the only workaround I found was to completely remove the conda-forge channel completely due to c library linking issues. It appears as though this is no longer the case, so I guess instead I should advise users to create the environment in this fashion now. Thanks for the clarification.

mbargull:
Ah, thanks for clarifying. Yes, some time ago there were some more packages only available on `defaults`/`r`. I usually don't interact with R at all, so I don't know the exact current state of this. Hopefully everything will work fine for you/your users with `conda-forge` now!

PR #3500, opened 2017-01-17
Feature/fix augustus

jerowe:
* [ ] I have read the guidelines above.
* [ ] This PR adds a new recipe.
* [x] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

I'm refactoring this PR a bit.

https://github.com/bioconda/bioconda-recipes/pull/3283

Issue #3284, opened 2016-12-22
Augustus - filterBam.cc:31:10: fatal error: 'unordered_map' file not found


mictadlo:
Hi,
I tried to follow linuxbrew [recipe](https://github.com/Linuxbrew/homebrew-science/blob/master/augustus.rb), but it failed ([PR](https://github.com/bioconda/bioconda-recipes/pull/3283)) which also would allow compiling on Mac OS X and Linux. I got the following error:

```
+ make -C auxprogs/filterBam/src BAMTOOLS=/anaconda/conda-bld/augustus_1482381198348/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/include/bamtools COMPGENEPRED=true
c++ -DUSE_BOOST -I/anaconda/conda-bld/augustus_1482381198348/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/include -L/anaconda/conda-bld/augustus_1482381198348/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/lib  -L/anaconda/conda-bld/augustus_1482381198348/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/lib  -std=c++0x  -c filterBam.cc -o filterBam.o -I/anaconda/conda-bld/augustus_1482381198348/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/include/bamtools -Iheaders -I./bamtools
clang: warning: argument unused during compilation: '-L/anaconda/conda-bld/augustus_1482381198348/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/lib'
[...]

PR #7676, opened 2018-02-14
Fix Augustus build errors on OSX

kastman:

Finishes @jerowe 's work in #3500 to get Augustus compiling on OSX; this patches the filterBam Makefile to use  c++11 which fixes the "fatal error: 'unordered_map' file not found" also listed in #3284.

### DIFF
--- a/recipes/augustus/3.2.3_boost1.57/patches/filterBam.src.Makefile.patch
+++ b/recipes/augustus/3.2.3_boost1.57/patches/filterBam.src.Makefile.patch
@@ -12,8 +12,9 @@
 +INCLUDES =  -I${PREFIX}/include/bamtools/ -Iheaders -I./bamtools
 +#LIBS = -lbamtools -lz
 +LIBS = ${PREFIX}/lib/libbamtools.a -lz
- CFLAGS = -std=c++0x
+-CFLAGS = -std=c++0x
++CFLAGS = -std=c++11
 +CXXFLAGS += -Wall -O2 # -g -p -g -ggdb 
  VPATH = functions
- 
+
  all : $(PROGRAM) $(OBJECTS) CHECKBAM BIN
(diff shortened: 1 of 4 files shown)


bgruening:
@kastman nice! Exactly the help which we need on OSX.
Could you try without bumping the build number? This will produce a OSX build but no Linux one.

recipes/augustus/3.2.3_boost1.57/patches/filterBam.src.Makefile.patch
```diff
@@ -12,8 +12,9 @@
 +INCLUDES =  -I${PREFIX}/include/bamtools/ -Iheaders -I./bamtools
 +#LIBS = -lbamtools -lz
 +LIBS = ${PREFIX}/lib/libbamtools.a -lz
- CFLAGS = -std=c++0x
+-CFLAGS = -std=c++0x
++CFLAGS = -std=c++11
 +CXXFLAGS += -Wall -O2 # -g -p -g -ggdb
```
mbargull: We could either remove
```diff
+CXXFLAGS += -Wall -O2 # -g -p -g -ggdb
```
from the patch or use
```diff
-CFLAGS = -std=c++0x
+CFLAGS += -std=c++11 -Wall -O2
```
instead. Currently, setting `CXXFLAGS` is a no-op since the `Makefile` doesn't adhere to convention and uses `CFLAGS` instead of `CXXFLAGS`.

jerowe:
This is great! Thanks!

recipes/augustus/3.2.3_boost1.57/patches/filterBam.src.Makefile.patch
kastman: What’s the `-O2` doing in `CXXFLAGS`? If the `Makefile` doesn’t use them, are they actually doing anything? Thanks for the catch.

mbargull: > If the `Makefile` doesn’t use them, are they actually doing anything?

Nope, should be a no-op that we can remove, I think

mbargull:
The patch now has one less line:
You need to replace
`@@ -8,10 +8,13 @@`
with
`@@ -8,10 +8,12 @@`
-- or create a new patch if uncertain.

(Huh, interesting that on Linux there were no complaints about the malformed patch...)

kastman:
Yep! Just got it.

I don’t think anything got queued on Linux (“Nothing to be done”), so the patch wasn’t applied at all and thus didn’t fail.

PR #7535, opened 2018-02-06
Update to gromacs 2018

stain:
* [x] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [ ] This PR adds a new recipe.
* [x] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).
* [x] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

Note that this updates to the latest gromacs 2018.

It might be that this can be optimized better, enabling openmp, gpu and hwloc.

```
GROMACS version:    2018
Precision:          single
Memory model:       64 bit
MPI library:        thread_mpi
OpenMP support:     disabled
GPU support:        disabled
SIMD instructions:  AVX_256
FFT library:        fftw-3.3.5-fma-sse2-avx-avx2-avx2_128
RDTSCP usage:       enabled
TNG support:        enabled
Hwloc support:      disabled
Tracing support:    disabled
Built on:           2018-02-06 15:38:42
Built by:           stain@biggie [CMAKE]
Build OS/arch:      Linux 4.13.0-26-generic x86_64
Build CPU vendor:   Intel
Build CPU brand:    Intel(R) Core(TM) i7-3520M CPU @ 2.90GHz
Build CPU family:   6   Model: 58   Stepping: 9
Build CPU features: aes apic avx clfsh cmov cx8 cx16 f16c htt intel lahf
mmx msr nonstop_tsc pcid pclmuldq pdcm popcnt pse rdrnd rdtscp sse2 sse3
sse4.1 sse4.2 ssse3 tdt x2apic
C compiler:
/tmp/miniconda/conda-bld/gromacs_1517931402329/_t_env/bin/cc GNU 4.8.5
C compiler flags:    -mavx    -Wno-unknown-pragmas  -O3 -DNDEBUG
-funroll-all-loops -fexcess-precision=fast
C++ compiler:
/tmp/miniconda/conda-bld/gromacs_1517931402329/_t_env/bin/c++ GNU 4.8.5
C++ compiler flags:  -mavx    -std=c++11  -Wno-unknown-pragmas  -O3
-DNDEBUG -funroll-all-loops -fexcess-precision=fast
```

PR #7684, opened 2018-02-14
Circleci fixes

mbargull:

This does a couple of small things in regard to the CircleCI configuration:
1. Explicitly installs `git` and `pip` to our Conda environment. Should be a no-op since `python` (and thus `pip`) and `git` are dependencies of `bioconda-utils`. But since we `pip install git+https://...` it doesn't hurt to be sure we have those dependencies in place.
2. Uses `--system` parameter for `conda config`. Without that parameter, I think, the configuration is written to the home directory and not `$WORKSPACE`. Since we cache `$WORKSPACE`, we should write it there. \[NOTE: I'm a little confused as to why we didn't already run into problems on macOS since the cache was activated there...\]
3. Checks `CIRCLE_PROJECT_USERNAME == bioconda` for the nightly uploads and exits early on forks.
4. ~\[**EDIT**: Added another commit to make the `PATH` changes conditional. I think that should be fine and is definitely nicer to look at.\]
   Instead of `$BASH_ENV`, uses a custom `.setup_bash_env` script. It's a little ugly since we need to run that for every `run/command` explicitly -- does anyone have a better idea? We shouldn't use `$BASH_ENV` since this resets `PATH` for every `bash` execution, which includes `post-link.sh` scripts and the like. \[NOTE: again, I'm surprised we didn't run into problems on macOS already.\]~ [nvm](https://github.com/bioconda/bioconda-recipes/pull/7684#discussion_r168213812)
5. Sources the `condaforge/linux-anvil` `entrypoint_source` which will enable `devtoolset-2` and thus be a little more like when run with `--docker`, I think. But still, the local configuration doesn't mirror the one on CircleCI -- I've yet to try out some other things.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,13 +52,17 @@ jobs:
       - *setup
       - run:
           name: Linting
-          command: bioconda-utils lint recipes config.yml --git-range master HEAD
+          command: |
+            source .setup_bash_env
+            bioconda-utils lint recipes config.yml --git-range master HEAD
       - run:
           name: Testing
           # Currently, using --docker below causes a permission error on Fedora 27, 
           # although it should be supported by the setup_remote_docker directive 
           # above. A workaround or an upstream fix would be welcome.
-          command: bioconda-utils build recipes config.yml --git-range master HEAD
+          command: |
+            source .setup_bash_env
+            bioconda-utils build recipes config.yml --git-range master HEAD
 
   lint:
     <<: *linux
@@ -70,7 +74,9 @@ jobs:
       - *save_cache
       - run:
           name: Linting
-          command: bioconda-utils lint recipes config.yml --git-range master HEAD
+          command: |
+            source .setup_bash_env
+            bioconda-utils lint recipes config.yml --git-range master HEAD
 
   # build and test on linux
   test-linux:
@@ -83,7 +89,9 @@ jobs:
       - *save_cache
       - run:
           name: Building and testing
-          command: bioconda-utils build recipes config.yml --git-range master HEAD --docker --mulled-test
+          command: |
+            source .setup_bash_env
+            bioconda-utils build recipes config.yml --git-range master HEAD --docker --mulled-test
 
   # build and test on macos
   test-macos:
@@ -96,7 +104,9 @@ jobs:
       - *save_cache
       - run:
           name: Building and testing
-          command: bioconda-utils build recipes config.yml --git-range master HEAD
+          command: |
+            source .setup_bash_env
+            bioconda-utils build recipes config.yml --git-range master HEAD
   
   # build, test and upload on linux
   upload-linux:
@@ -110,7 +120,9 @@ jobs:
       # build only current commit (since we use squashed merging, this is safe)
       - run:
           name: Building, testing, and uploading
-          command: bioconda-utils build recipes config.yml --docker --mulled-test --anaconda-upload --mulled-upload-target biocontainers --git-range $CIRCLE_SHA1~1 $CIRCLE_SHA1
+          command: |
+            source .setup_bash_env
+            bioconda-utils build recipes config.yml --docker --mulled-test --anaconda-upload --mulled-upload-target biocontainers --git-range $CIRCLE_SHA1~1 $CIRCLE_SHA1
 
   # build, test and upload on macos
   upload-macos:
@@ -124,7 +136,9 @@ jobs:
       # build only current commit (since we use squashed merging, this is safe)
       - run:
           name: Building, testing and uploading
-          command: bioconda-utils build recipes config.yml --anaconda-upload --git-range $CIRCLE_SHA1~1 $CIRCLE_SHA1
+          command: |
+            source .setup_bash_env
+            bioconda-utils build recipes config.yml --anaconda-upload --git-range $CIRCLE_SHA1~1 $CIRCLE_SHA1
 
   # build, test and upload for bulk branch on linux
   bulk-linux:
@@ -138,7 +152,9 @@ jobs:
       # build everything that differs compared to master
       - run:
           name: Building, testing, and uploading of all unpublished recipes
-          command: bioconda-utils build recipes config.yml --docker --mulled-test --anaconda-upload --mulled-upload-target biocontainers --git-range master HEAD
+          command: |
+            source .setup_bash_env
+            bioconda-utils build recipes config.yml --docker --mulled-test --anaconda-upload --mulled-upload-target biocontainers --git-range master HEAD
 
   # build, test and upload for bulk branch on macos
   bulk-macos:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -26,6 +26,12 @@ variables:
       LANG: en_US.UTF-8
   linux: &linux
     machine: true
+  skip_fork: &skip_fork
```
johanneskoester: Nice! Please also add this to all other bulk and upload jobs.

mbargull:
6. Lets builds fail if their `.circleci` is not up to date with upstream. Ref: https://github.com/bioconda/bioconda-recipes/pull/7535 -- that one was merged since it showed "all green", but actually only because it didn't ran for macOS due to an outdated config.

mbargull:
DISCLAIMER: I didn't test any of those changes. If some look more meaningful than others, we can also split this PR into smaller ones.

.circleci/config.yml
mbargull: done

.circleci/setup.sh
```diff
@@ -7,8 +7,21 @@ WORKSPACE=`pwd`
...
+if ! git diff --quiet HEAD...upstream/master -- .circleci/; then
+    echo 'The CI configuration is out of date.'
+    echo 'Please merge in bioconda:master.'
+    exit 1
+fi
+
 # Set path
-echo "export PATH=$WORKSPACE/miniconda/bin:$PATH" >> $BASH_ENV
+cat >> $BASH_ENV << EOF
+if ! [[ "\$PATH" =~ (^|:)"$WORKSPACE/miniconda/bin"(|/)(:|\$) ]]; then
```
johanneskoester: Can you please add a comment that explains what the condition is supposed to check :-)?

johanneskoester: How is this different from what we had before?

.circleci/config.yml
```diff
@@ -40,8 +46,8 @@ jobs:
             mkdir -p ~/.ssh
             ssh-keyscan -H github.com >> ~/.ssh/known_hosts
       - run: 
-          name: linking conda
-          command: ln -s /opt/conda $HOME/miniconda
+          name: Setup base system
+          command: echo '. /opt/docker/bin/entrypoint_source' >> $BASH_ENV
```
johanneskoester: Isn't this entrypoint the one with the gigantic chown? If so, why do we include it now (remember the performance problems)? Didn't we have a fix in the entrypoint of our derived image?

.circleci/setup.sh
mbargull: Previously we reset `PATH` unconditionally so that any `bash` script would reset it when it automatically sourced `$BASH_ENV`. E.g., with a `post-link.sh` script the activated environment would have been be kicked out of `PATH` since it would run `$BASH_ENV` which would've reset `PATH`.

.circleci/config.yml
mbargull: The `chown` one is `/opt/docker/bin/entrypoint` which sources `/opt/docker/bin/entrypoint_source` and does all the user creation and `chown` stuff.
`/opt/docker/bin/entrypoint_source` is lightweight, but I'm actually not happy with hardcoding its path here. It would be better to turn the `docker-entrypoint` from the `bioconda-utils` repo into `docker-entrypoint` and `docker-entrypoint-source` and then use the latter here.
Well, even better would be to use `--docker`, but I'm not finished with some workaround to enable that.

.circleci/setup.sh
johanneskoester: Is this a particular behavior of post-link? In general, non-interactive shells should not source .bashrc again (which is $BASH_ENV on circle ci).

mbargull: Ah, right, I think that was only a brain fart on my end. Yeah, `$BASH_ENV` shouldn't be run because we won't be using interactive shells. I'll just revert that change.

.circleci/config.yml
johanneskoester: Nice to hear that --docker might eventually work!

Ok, sounds reasonable, but why not loading entrypoint_source directly in the bioconda-utils docker image?

mbargull: It is loaded in the docker image. But it only makes non-permanent changes `PATH`, i.e., doesn't store it in `BASH_ENV` & co. And I'm pretty sure `steps` in `.circleci/config.yml` will run `/bin/bash` instead of that entrypoint, though.

.circleci/setup.sh
```diff
@@ -7,6 +7,15 @@ WORKSPACE=`pwd`
 # This file can be used to set BIOCONDA_UTILS_TAG and MINICONDA_VER.
 source .circleci/common.sh
 
+# make sure the CircleCI config is up to date
+git remote add -t master upstream https://github.com/bioconda/bioconda-recipes.git
+git fetch upstream
+if ! git diff --quiet HEAD...upstream/master -- .circleci/; then
+    echo 'The CI configuration is out of date.'
+    echo 'Please merge in bioconda:master.'
+    exit 1
```
mbargull: @johanneskoester: Do you think it's reasonable to just abort the builds like this? (Reason for the change: https://github.com/bioconda/bioconda-recipes/pull/7684#issuecomment-365637360)

.circleci/config.yml
johanneskoester: I thought the whole point of entrypoints was that they run in addition to bin/bash. But I guess you checked that and discovered that the entrypoint is not loaded?

mbargull: > But I guess you checked that and discovered that the entrypoint is not loaded?

Correct, learned that by observation, not by deduction. And I'm still very confused about that. You can even override the Docker entrypoint with a custom one in the CircleCI config. But I still don't know when/how/if that is used (tested with `circleci build` locally).
It's also the reason why people couldn't run `git` until you added it to `bioconda-utils`'s requirements -- if the entry point would have been used, `git` from `/opt/conda/bin` would've been on `PATH`. 😕

.circleci/setup.sh
johanneskoester: Yes, this is a very nice idea.

.circleci/config.yml
mbargull: I changed it so that `master`/`bulk` upload jobs fail on forks instead. The reason for this is that if a contributor tries to use their `master`/`bulk` for pull requests, we don't want it to silently succeed, but rather just fail and inform them to use another branch. For the nightly builds, I think it's nice to not let them fail and as such not bug the contributor with "failed build" notifications.

(Context for others: What we actually want is to be able to use `repo:branch` as a filter in the CircleCI `jobs`; but that is currently not supported. Hence this `skip_fork`/`fail_fork` workaround is the next best thing we can do if we still want to use `jobs`.)

johanneskoester:
LGTM. Feel free to merge.

.circleci/config.yml
mbargull: Looks like, as intended, the builds are terminated quickly after < 10 seconds:
https://circleci.com/gh/mbargull/bioconda-recipes/3
https://circleci.com/gh/mbargull/bioconda-recipes/4

PR #7685, opened 2018-02-14
IntaRNA : enable multithreading (linux)

martin-mann:

### DIFF
--- a/recipes/intarna/meta.yaml
+++ b/recipes/intarna/meta.yaml
@@ -31,7 +31,8 @@ requirements:
   run:
     - libgcc # [linux]
     - libcxx # [osx]
-    - boost {{CONDA_BOOST}}* # [linux]
+    - boost {{CONDA_BOOST}}*
+    # [linux]
     #- boost 1.63.0 # [osx]
     # for old interface wrapper
     - perl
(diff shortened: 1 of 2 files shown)


recipes/intarna/meta.yaml
```diff
@@ -31,7 +31,7 @@ requirements:
   run:
     - libgcc # [linux]
     - libcxx # [osx]
-    - boost {{CONDA_BOOST}}* # [linux]
+    - boost =={{CONDA_BOOST}}* # [linux]
```
bgruening: the `==` is not needed.

martin-mann:
Moin @bgruening , any idea why linting says "boost not pinned"? Worked so far, what has changed?

martin-mann:
might the problem occure due to the non-CONDA_BOOST version I am pinning for OSX?

martin-mann:
nope, even then the same error...

martin-mann:
also not working

martin-mann:
just found that CONDA_BOOST is now 1.60 too.. so no distinction should be needed.. right

but anyway.. seems to be a problem with the linting if system specific pinning is not working... or?

PR #7695, opened 2018-02-15
OSX build for clustalo with OpenMP

colinbrislawn:

### DIFF
--- a/recipes/clustalo/meta.yaml
+++ b/recipes/clustalo/meta.yaml
@@ -10,12 +10,10 @@ source:
 
 requirements:
   build:
-    - gcc # [linux]
-    - llvm # [osx]
+    - gcc
     - argtable2
   run:
-    - llvm # [osx]
-    - libgcc # [linux]
+    - libgcc
     - argtable2
 
 build:
--- a/recipes/clustalo/meta.yaml
+++ b/recipes/clustalo/meta.yaml
@@ -19,7 +19,7 @@ requirements:
     - argtable2
 
 build:
-  number: 0
+  number: 1
 
 test:
   commands:
--- a/recipes/clustalo/meta.yaml
+++ b/recipes/clustalo/meta.yaml
@@ -24,7 +24,6 @@ build:
 test:
   commands:
     - "clustalo --help > /dev/null"
-    - "clustaloMP --help > /dev/null"
 
 about:
   home: "http://www.clustal.org/omega/"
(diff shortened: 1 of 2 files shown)


colinbrislawn:
Do you know if I have to run `make` twice, once for each `configure`, or if I can configure everything then run make one time?

souravsingh:
Configuring once should be fine

colinbrislawn:
I want to end up with two executables. One for single threaded or systems without OpenMP, and a second on compiled with OpenMP. FastTree2 uses this setup and it works well.

Do you think we should do that or only have an OpenMP version of ClustalO on bioconda?

colinbrislawn:
Also, I'm getting this error:
```
#!/bin/bash -eo pipefail
.circleci/setup.sh
From ssh://github.com/bioconda/bioconda-recipes
 * [new branch]          master     -> upstream/master
The CI configuration is out of date.
Please merge in bioconda:master.
Exited with code 1
```

Any idea? This looks like a simple error but I'm not sure how to solve it.

mbargull:
@colinbrislawn, my bad concerning the error. Can you merge or rebase on master again? I made some additional changes to the CI configuration around the time you merged the first time. Sorry for the bother.

colinbrislawn:
Sure thing! 

I know this is a basic git question, but what's the best way to go about doing this, without just opening a new PR? How do make my branch point to a newer master?

colinbrislawn:
@brwnj Can you give me some feedback on this PR? I'm not sure if this is the best way to run the build to enable OpenMP. Let me know if you have any suggestions!

colinbrislawn:
Turns out this is only needed on OSX. It already works on Linux because
>ClustalO will automatically support multi-threading if your compiler supports OpenMP.

 🙄 I'll try to get a simpler build that also works on OS!

recipes/clustalo/meta.yaml
```diff
@@ -17,8 +17,6 @@ requirements:
     - llvm # [osx]
     - libgcc # [linux]
     - argtable2
-    
-    
 
 build:
   number: 0
```
rvalieris: you need to increase the build number, otherwise no build is triggered.

colinbrislawn:
On OSX I'm getting
`clang: error: unsupported option '-fopenmp'` 

Would it help to switch to another version of gcc?
```
You could use a non-Apple gcc installed via MacPorts, adding

CC=/opt/local/bin/gcc-mp-4.5

to the configure call (you will have to change the exact string to match
your gcc version).
```

colinbrislawn:
Could @rvalieris or @souravsingh make sure that the all gcc build is set up correctly? I think it's good to merge, but I wanted to make sure!

rvalieris:
yes, it should be alright, thanks !

Issue #6491, opened 2017-10-26
How to handle older recipes?


daler:
As software is updated, we update recipes. What should happen to the older recipes? In some cases, we retain the older recipes in subdirectories; in others we overwrite the old with the new. Currently our policy at http://bioconda.github.io/guidelines.html#versions is to overwrite the old with the new but we haven't been enforcing this.

This recently came up in https://github.com/bioconda/bioconda-recipes/pull/5117 and https://github.com/bioconda/bioconda-recipes/pull/6471.

I think that overwriting old with new is OK since recipes are always bundled with the package tarball, and recipes will always be in the git repo history. We could build a tool that, given a package and the repo, will identify which commit the recipe came from. This also keeps the repo directory structure clean. However there is also value in retaining a more visible record of the recipe changes in the repo in the form of older recipe subdirectories.

Thoughts?

cc @yhoogstrate @karel-brinda @jerowe @bioconda/core

jerowe:
I have a slight preference towards keeping the old recipe, but like you
said it's redundant information. I think I've only had one time I really
had to go looking for an old version of a recipe, and I was able to get
that from the git history.

So very slight personal preference, but I can also see why it's not
necessary or practical.

yhoogstrate:
From a dev and git perspective it is more convenient to track all versions. Lets assume we have a tool which is superseded by a new version and the recipe is overwritten. Suddenly the recipe of the previous version needs a fix. Then the old versions recipe is taken from the archive, reinserted, and lost all its git history. I think file size is the least to worry about because they're quite small flat files. It has my preference to include recipes for all ( at least major) versions, ideally all versions. If we keep appropriate directory names en separate by versions, maybe even major nested with minor versions, it doesnt need to become chaotic.

rvalieris:
imho, we should always update recipes in-place, keeping the old recipe is an exception for when we need to rebuild/update the older version *after* a newer version already exists, and this is not common occurrence.

thats not to say we shouldn't allow it, but I think its better to only keep older recipes we really need instead of all of them.

karel-brinda:
I think the question is what's the final product of Bioconda – the binaries? the recipes? tarballs? the github repository (with its history)?

If "updating recipes in place" becomes the recommended policy, then I would be happy for an additional, automatically generated, Github repository with all the versions (of all the recipes for all  the packages) that were used to generate the binaries. Knowing the recipe for a given package is important for reproducibility.

[...]

PR #7697, opened 2018-02-15
edlib: update to 1.2.3

mbargull:

Updates `edlib` to version `1.2.3`.
Also removes the subdirectories for older versions (they don't seem to be maintained -- all have build number 0).


bgruening:
@mbargull have you see https://github.com/bioconda/bioconda-recipes/pull/7709 ?
Imho build number 0 does not mean its not maintained. @karel-brinda any opinion here?

mbargull:
Thanks @bgruening.
> @mbargull have you see #7709 ?

Not yet, nice.

> Imho build number 0 does not mean its not maintained. @karel-brinda any opinion here?

Ah, I meant to ping Karel -- forgot to do so; thanks for doing that. I agree, that build number 0 doesn't imply unmaintained. The overall look (i.e., incl. history) let me assume that the old versions aren't likely to be rebuilt in future, i.e., are unmaintained, though. @karel-brinda: Do you intend to update the older version at some point? If not, we can remove their recipes as suggested in this PR.

mbargull:
Thanks for the feedback @karel-brinda.
> However, is there any reason to remove the old recipes?

Maintenance burden. All recipes, including those in subdirectories, are possibly subject to occasional maintenance work. An example for that would be when a dependency is added to the globally pinned packages (like the `zlib` update @bgruening did recently on the `bulk` branch). To be fair, that didn't apply to `edlib`, but future updates might (e.g., possibly the upcoming update to `conda-build 3`).
> It seems that the BC community has not agreed on one recommended approach.

That's correct and I know you are aware of and took part of the discussion in https://github.com/bioconda/bioconda-recipes/issues/6491. You are right that we should work out a *recommended approach*. IMHO, it's sensible to use https://github.com/bioconda/bioconda-recipes/issues/6491#issuecomment-357678042 as a guideline for now. (Although I don't agree that it's easy enough to trace back a recipe/commit from a package; we need to push https://github.com/conda/conda-build/pull/2489 forward, IMO!)
> In this specific case, the old recipes might be useful for documentation (the main program was renamed between two versions).

I didn't even notice that! Do you think this is important information? If so, we could add a comment, or better, add that under `extra: notes:` in `meta.yaml`.

karel-brinda:
@mbargull Thank you very much for the explanation. The main executable from version 1.0.0 was named just `aligner`. From 1.1.2 it is `edlib-aligner`.

mbargull:
> The main executable from version 1.0.0 was named just `aligner`. From 1.1.2 it is `edlib-aligner`.

Do you think this information is very important such that I should put a note in `extra: notes:` for that?

karel-brinda:
I am not sure. This information might be useful in situations when people install an environment for a pipeline that was originally using a slightly older version of the tool and some manual adjustments are thus required. However. I am not sure if people read `extra: notes:` anyway. :)

Personally I often look into Conda recipes to see how certain packages were built, what compilation parameters were used, etc. But this is probably not a typical behavior of Conda users. And I should start to look into tarballs rather than into the Github repo.

Is there any guideline about what information is expected to appear in `extra: notes:`? Is this information propagated somewhere (e.g., to the associated webpages of the packages)?

mbargull:
> However. I am not sure if people read `extra: notes:` anyway. :)

That's what I'm also always wondering about 😆 

> Is there any guideline about what information is expected to appear in extra: notes:?

Notes! Hehe, no I don't think there are any guidelines. `grep`ping the guidelines for `notes`, the only thing I can find is:
- > If the recipe installs custom wrapper scripts, usage notes should be added to
  ``extra -> notes`` in the ``meta.yaml``.

IMHO, we can include any information there if we think it's useful.

> Is this information propagated somewhere (e.g., to the associated webpages of the packages)?

Yes! E.g., http://bioconda.github.io/recipes/quast/README.html#notes generated from
https://github.com/bioconda/bioconda-recipes/blob/b8e85bbc4a1b6a4f5a3f3f8dc124a042b2fe69d9/recipes/quast/meta.yaml#L62-L64

----
I just added the note. Should be ready to merge now!

mbargull:
Ha, error!
```
From ssh://github.com/bioconda/bioconda-recipes
 * [new branch]          master     -> upstream/master
The CI configuration is out of date.
Please merge in bioconda:master.
Exited with code 1
```
. I like when things work as intended 😄. Rebasing...

PR #7698, opened 2018-02-15
Circle ci add upstream check

karl616:

Pointing the remote upstream to the main repository does not work for the local build if it is already set. Here is a simple check, but I'm not sure it solves the problem completely. I still get an error:

```
====>> Setup bioconda-utils
  #!/bin/bash -eo pipefail
.circleci/setup.sh
/usr/bin/scl_source: line 34: _scls[@]: unbound variable
Error: Exited with code 1
Step failed
Task failed

```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -58,7 +58,14 @@ jobs:
             ssh-keyscan -H github.com >> ~/.ssh/known_hosts
       - run: 
           name: Setup base system
-          command: echo '. /tmp/repo/docker-entrypoint-source' >> $BASH_ENV
+          command: |
+            echo >> $BASH_ENV <<EOF
+            if [[ $- == *u* ]] ; then
+                set +u ; . /tmp/repo/docker-entrypoint-source ; set -u
+            else
+                . /tmp/repo/docker-entrypoint-source
+            fi
+            EOF
       - *common
       - *setup
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,8 +59,8 @@ jobs:
       - run: 
           name: Setup base system
           command: |
-            echo >> $BASH_ENV <<EOF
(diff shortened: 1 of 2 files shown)


mbargull:
Thanks @karl616 for looking into this. Sorry I broke it before; should've tested it before merging..
I added a commit to use another approach which simply uses a semi-random name for the remote which should just circumvent the collision. I also added a possible workaround for the other error you got. Didn't test it yet myself -- can you give it a try?

mbargull:
Okay, should be good to go now!

karl616:
Thanks @mbargull  for including me.

The remote problem is resolved. I like your solution better as it doesn't put restrictions on the repository.

I cannot say about the second problem as I get a new error:

```
====>> Setup bioconda-utils
  #!/bin/bash -eo pipefail
.circleci/setup.sh
.circleci/setup.sh: line 13: git: command not found
Error: Exited with code 127
Step failed
Task failed
```
Considering this I'm not sure if I should say anything about the first problem either... although it very much looks resolved...

I had the missing git problem earlier.  I think there was a solution upstream (of me). and I tried:
```docker pull bioconda/bioconda-utils-build-env```

Is it related to the config change?

Best,
Karl

mbargull:
Yes, a `docker pull bioconda/bioconda-utils-build-env` was needed as I now use a new file, `docker-entrypoint-source`, that I only added yesterday. The `git: command not found` error should be gone now with the last two commits. If it runs now fine on your site with those last two changes, we can merge when the CI gives it green light.

mbargull:
`circleci build` now works again on my machine 😄, thanks for your help @karl616!

karl616:
Here as well :)

@mbargull you're welcome, but you did all the work

mbargull:
Work =/= putting the final fix in place 😉.
I appreciate you provided the bug report, the initial change and did the testing -- all of that is work as well!

PR #7701, opened 2018-02-15
Pin bamtools

johanneskoester:

Pin bamtools all recipes that depend on the library. Update ngs-disambiguate to latest version.


bgruening:
@johanneskoester please have a look in bulk. I'm working since a week on this. It should be done for this tool already. I will blacklist all remaining failures to merge it back into master as soon as possible.

johanneskoester:
Oh nice, sorry Björn! Thanks for letting me know!

johanneskoester:
Mhm, I have had a look, and I cannot find the pinning in bulk. Are you sure?

bgruening:
The zlib changes yes. I just fear we are diverging to much. Let me blacklist everything and get it merged into master.

johanneskoester:
Ok, good point! Please let me know once bulk is merged.

bgruening:
@johanneskoester go for it. bulk is free and merged.

recipes/ngs-disambiguate/meta.yaml
```diff
@@ -1,23 +1,27 @@
+{% set version = "1.0.0" %}
+
 package:
   name: ngs-disambiguate
-  version: "2016.11.10"
```
bgruening: @johanneskoester the new version will be `older` than the old one right? So we want to retire the older version?

johanneskoester: I just checked and the release was on the same day as the tag. Nevertheless, people might already pin against the "date"-version. Unsure how to solve this...

bgruening: We could label it as `broken`? And people who have pinned it needs to enable this label?

johanneskoester: Yes, you are right. Thinking about it, that makes also sense because the version is actually broken, due to the missing pinning. Let's do that.

mbargull: +1 It also doesn't look like that version was an official one anyway: https://github.com/AstraZeneca-NGS/disambiguate/blame/v1.0.0/setup.py

Issue #7415, opened 2018-01-29
Perl XS won't build due to unsupported compiler option -fstack-protector-strong


nathanweeks:
I'm not sure if this is the right repo to submit this issue to, but while attempting to update perl-dbi, I ran into the following build issue:
```
$ circleci build
...
cc -c   -D_REENTRANT -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2 -O2   -DVERSION=\"1.640\" -DXS_VERSION=\"1.640\" -fPIC "-I/opt/conda/conda-bld/perl-dbi_1517236834053/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/perl5/5.22.0/x86_64-linux-thread-multi/CORE"  -W -Wall -Wpointer-arith -Wbad-function-cast -Wno-comment -Wno-sign-compare -Wno-cast-qual -Wmissing-noreturn -Wno-unused-parameter Perl.c
cc: error: unrecognized command line option ?-fstack-protector-strong?
```
The default perl (5.22.2.1-0 from conda-forge) was apparently compiled with a RHEL version of gcc 4.8.2, which seems to have backported support for -fstack-protector-strong from gcc 4.9, and adds this option to the perl cppflags when compiling XS files:
```
# perl -V | grep -e ccversion -e cppflags
    cppflags='-D_REENTRANT -D_GNU_SOURCE -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -I/usr/local/include'
    ccversion='', gccversion='4.8.2 20140120 (Red Hat 4.8.2-15)', gccosandvers=''
```
The gcc 4.8.5-7 in the conda defaults channel is apparently not the RHEL version, and lacks support for -fstack-protector-strong.

FWIW, perl 5.20.3.1-2 from the conda-forge channel was compiled with only -fstack-protector, which is supported in gcc 4.8.5:

I'm not sure of the "most correct" way to handle this.

nathanweeks:
[...]

PR #7707, opened 2018-02-15
enable perl for vrna

mmiladi:


mmiladi:
According to https://github.com/bioconda/bioconda-recipes/issues/7415 there is an incompatibility issue between Docker gcc and conda gcc.

mtw:
@mmiladi maybe you need the ``check`` dependency for building?

mtw:
maybe adding  ``-fPIC`` to CFLAGS helps?

mmiladi:
@mtw What's the best way to add `-fPIC`? I assumed --enabled-shared would do this?

mtw:
you should be able to pass CFLAGS right after the configure call, like ``./configure --foo --bar CFLAGS="-fPIC -O3 -g" ``

RaumZeit:
May I ask why you use the `--enable-shared` as configure option? As far as I understand, thsi will break libtool, and therefore you'll see the `-fPIC` suggestion. Please try removing both, `-fPIC` and the `--enable-shared` but add a `V=1` to the `make` command. This will at least show us the actuall parameters that are passed to the compiler and linker in the CI

RaumZeit:
Another note: To circumvent running the entire autotools toolchain, I suggest creating a distribution tarball where all the patches are already applied, e.g. as ViennaRNA-2.4.3.1 or something else...

mmiladi:
@RaumZeit `--enable-shared` was a try and error to fix the complain about `-fPIC`
Preparing a patched tarball is really nice idea! I am working on that.

mmiladi:
Can this be due to having both `shared` and `static` enabled? Hint from the second answer here: https://stackoverflow.com/questions/19364969/compilation-fails-with-relocation-r-x86-64-32-against-rodata-str1-8-can-not/42388145

mtw:
I guess now we should get rid of the entire CFLAGS statement again :-/

mmiladi:
True! Another try :)

RaumZeit:
This persistent linker error is odd... For some reasons that I do not understand yet the linker tries to link against a static libperl library in the final step of making the Perl interface. Therefore the linker complains about adding `-fPIC` during compilation, not for anything from ViennaRNA but for libperl!
Could you somehow try to find out whether there is any libperl.so in the build environment?

RaumZeit:
Hi again!
Sorry that I couldn't be much of help yet, I just figured that bioconda only provides `libperl.a` to link the ViennaRNA perl interface to. I'm still in the process of finding out how and why linking should be done if a static version of libperl is available...Probably, the `Makefile.am` in `interfaces/Perl` needs to be adjusted...
Any insights and help is appreciated!

RaumZeit:
Unfortunately, I'm out of ideas on how to solve the linker problem.
Probably, we should wait for any progress made here: https://github.com/conda-forge/perl-feedstock/issues/19

martin-raden:
@mmiladi think this is outdated!?! recent version supports perl, so I am closing this..

PR #7710, opened 2018-02-15
 added recipes/perl-params-coerce

mtw:

### DIFF
--- a/recipes/perl-params-coerce/0.14/meta.yaml
+++ b/recipes/perl-params-coerce/0.14/meta.yaml
@@ -1,0 +1,44 @@
+{% set name = "perl-params-coerce" %}
+{% set version = "0.14" %}
+{% set sha256 = "5c67b11b9859e8d0d048c1c529e16ff49458fe4a98a16ccbfec6cd66e49b7713" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  fn: Params-Coerce-0.14.tar.gz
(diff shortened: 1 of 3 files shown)


bgruening:
@mtw we do not build for windows, so bat files are not needed to commit.

recipes/perl-params-coerce/0.14/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+build:
+  number: 0
+
+requirements:
+  build:
+    - perl #>=5.22.1
+    - perl-params-util
+
+  run:
+    - perl #>=5.22.1
```
bgruening: is this a comment? You can pin your recipe if you want to restrict it to specific perl versions. Would it make sense to also include a <6.0?

mtw: this was just for testing. Restriction <6.0 definitely makes sense.

bgruening: You can for example do something like this: `>=5.22,<6.0`

mtw: in this case, is it necessary to increase the build number from 0 to 1?

bgruening: No, the build number only needs to be increased if there is already a package with this version and you want o have a new-fixed one.

bgruening:
Thanks @mtw!

PR #7714, opened 2018-02-15
Update version of expansionhunter to v2.5.5

nh13:


bgruening:
@nh13 our standard compiler it too old for this version of the package.
You could use the new compiler chain in defaults, but the consequences are not totally clear. We hope to move to conda-build-3 in a few weeks. If you can wait that long ... otherwise try the new compilers.

PR #7827, opened 2018-02-17
Maker OSX Deps and build

kastman:

Initial perl-forks (maker dep) OSX build. Does it fail on circleci?


recipes/perl-test-exception/meta.yaml
```diff
@@ -12,11 +12,11 @@ build:
 
 requirements:
   build:
-    - perl-threaded
+    - perl
```
bgruening: You need to bump the build number here.

recipes/perl-forks/meta.yaml
```diff
@@ -9,15 +9,28 @@ source:
 
 build:
   number: 1
-  skip: True # [ not linux ]
```
bgruening: please bump the build number if you want a new version of this package

recipes/perl-test-exception/meta.yaml
kastman: I updated `perl-threaded` to `perl` because linting failed with threaded; wasn't sure this needed a new build number. But I've updated it below; good point.

recipes/perl-forks/build.sh
```diff
@@ -1,2 +1,23 @@
...
+    perl Build.PL
+    ./Build
+    ./Build test
+    # Make sure this goes in site
+    ./Build install --installdirs site
+elif [ -f Makefile.PL ]; then
+    # Make sure this goes in site
+    perl Makefile.PL # INSTALLDIRS=site
+    cat Makefile | grep fstack
+    sed -i 's/-fstack-protector-strong/-fstack-protector/g' Makefile
```
kastman: The ExtUtils::MakeMaker adds `-fstack-protector-strong` _somewhere_, but I can't figure out where. That flag however is only valid in >= gcc 4.9. So, one hack is to switch to the older `-fstack-protector` which is acceptable, but I'm not really sure what the implications are for speed or security here. I also don't know why gcc-4.8 is being used, when `llvm` / `clang` should be used on the mac build, but this replacement does at least seem to work. Is it a valid approach?

kastman:
Added a comment about the Makefile `-fstack-protector` in `perl-forks/build.sh`, but it's not showing up in the conversation so I thought I'd mention it here. Also, any idea why the macos circlebi build is still "waiting for status to be reported" instead of "pending"?

bgruening:
@kastman thanks. OSX will only start if Linux succeed.

kastman:
@bgruening Hmm, thanks. Looks like most things are working, but `repeatmasker` is failing on linux which is odd, since the only changes to the recipe were to bump the build number and not skip for mac. Is it possible the build/test system has changed since the last time `repeatmasker` was built?

recipes/perl-forks/build.sh
mbargull: For now you should rather remove the `gcc` dependency as is done with `perl` and other Perl recipes. That way the "system" `gcc` from the Docker container is used. That one is also just `gcc 4.8.2` but a patched one, see: https://github.com/conda-forge/perl-feedstock/pull/12#issuecomment-282003891.
(Note that this will change in future, but currently this is needed due to compatibility/legacy reasons.)

mbargull:
If you replace the single quotes with double quotes in the `repeatmasker` test command, that should fix it.

recipes/perl-forks/meta.yaml
```diff
@@ -8,16 +8,26 @@ source:
   md5: 48a1819dd2ff67d756f268771e1a9827
 
 build:
-  number: 1
-  skip: True # [ not linux ]
+  number: 2
+  skip: True # [ not unix ]
```
mbargull: You can just remove the `skip` line now; we don't build for non-`unix` anyway.

recipes/repeatmasker/meta.yaml
```diff
@@ -8,11 +8,11 @@ source:
     patches:
       - patch1
 build:
-  number: 9
+  number: 10
   detect_binary_files_with_prefix: False
   has_prefix_files:
     - bin/RepeatMasker
-  skip: True # [ not linux ]
+  skip: True # [ win ]
```
mbargull: You can just remove the `skip` line now; we don't build for non-`unix` anyway.

recipes/perl-forks/build.sh
kastman: Thanks @mbargull, those discussions help. I think that will work for the linux build, but is the patched `gcc 4.8.2` also used for the macOS build? I'm checking now, but I don't believe that will work locally with `circleci build` and I don't think there's any reason to believe it would be different on the hosted CI, right?

kastman: Also, does that mean this guideline is wrong? https://bioconda.github.io/guidelines.html#c-c

mbargull: macOS builds should use `clang` and `-fstack-protector-strong` shouldn't be added there, I think. Let's just see if the CI chokes on it. And the local `circleci build` should also use that patched `gcc 4.8.2` -- if not, let me know.
Regarding the guidelines, I think (not 100 % sure, though) that it's just the Perl packages that are an oddball here. We should adjust the guidelines anyway in near future when we switch to a newer `conda-build` version. So, for now, I'd take them as "subject to change".

kastman:
Looks like there might be a version incompatibility between List::MoreUtils and forks, but I'm confused by the versioning. Forks seems to require List::MoreUtils version `0.15` (https://metacpan.org/pod/forks#REQUIRED-MODULES) but MoreUtils version on cpan goes from `.400` to `.428` (https://cpan.metacpan.org/authors/id/R/RE/REHSACK/ ), or is version ".415" the same as "0.15"?. 

Honestly, I'm confused about how perl-forks built [previously](https://github.com/bioconda/bioconda-recipes/pull/7827/files#diff-db0be1001bbc163d5d049e0aa807a692), depending only perl-threaded, and not on `List::MoreUtils`, `Test::Exception`, `Acme::Damn`.... etc. , How did this linux build work previously?

@pvanheus , did you seen this when you were writing the recipe initially?

kastman:
Any idea why this would pass on my local `circleci build`, but fail remotely? 

I'm also confused as to why trnascan is throwing this environment error, [it's requirements](https://github.com/bioconda/bioconda-recipes/blob/8c11b68daa49bb5cabc46aabcd7ba162c868858f/recipes/trnascan-se/meta.yaml#L24) look quite simple. Any suggestions? Thanks!

kastman:
The problem was that tRNAscan-se 1.3.1 was Linux only and not built for mac, but 2.0 was. Although the maker recipe is version-locked to 1.3.1, somehow my local `circleci build` must have been accepting 2.0 and building, while the remote wasn't. 

I've backported the 1.3.1 version (in a separate directory) and fixed it to build on both mac and linux, and maker is now happy.

Could I get a final code review before merging, @bgruening or @mbargull? I did bump the build number on tRNAscan-SE. Thanks,

recipes/perl-list-moreutils/0.15/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+# Make sure this goes in site
+set -x -e
+export PATH=/opt/rh/devtoolset-2/root/usr/bin/:$PATH
```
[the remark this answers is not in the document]
kastman: Good catch - I had also left cpanminus in as a dep which I don't believe is needed from some earlier build attempts. Thanks!

recipes/maker/meta.yaml
```diff
@@ -32,6 +31,7 @@ requirements:
     - perl-inline-c
     - perl-perlio-gzip
     - perl-forks
+    - perl-list-moreutils ==0.15
```
kastman: I added perl-list-moreutils version; should I also bump the build number? Linux was building successfully without it.

kastman:
@bioconda/core is this good to merge? Thanks!

recipes/maker/meta.yaml
bgruening: This is strange. Maybe it is just a buildtime dependency and its installed inside our container? Can you try removing the run-time dependency. Also is there a need to pin the exact version?

kastman: @bgruening perl-list-moreutils is more or less two different packages, from different authors at different times. I'm not sure 0.15 is exactly required, but it's the version listed in the `perl-forks` cpan dependencies and I decided it was known good. A later version fails (I'm re-running the test without pinning to get the error again). 

Not sure what's going on, so I'll also remove the run-time dep after that failure.

kastman: I've reverted the breaking changes (removing the run-time dependency and version-locking to a known-good version because newer versions were conflicting). It's now breaking on the linux circleci builds, so it's not a mac specific issue. The conflicting errors from the [circleci build](https://circleci.com/gh/bioconda/bioconda-recipes/4574?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) were:

```
Undefined value assigned to typeglob at /opt/conda/conda-bld/maker_1520105724755/_t_env/lib/perl5/site_perl/5.22.0/x86_64-linux-thread-multi/List/MoreUtils.pm line 94.
...
Undefined subroutine &List::MoreUtils::firstidx called at /opt/conda/conda-bld/maker_1520105724755/_t_env/lib/perl5/site_perl/5.22.0/x86_64-linux-thread-multi/forks/signals.pm line 33.
Undefined subroutine &List::MoreUtils::minmax called at /opt/conda/conda-bld/maker_1520105724755/_t_env/lib/perl5/site_perl/5.22.0/x86_64-linux-thread-multi/forks.pm line 1593.
Compilation failed in require at /opt/conda/conda-bld/maker_1520105724755/_t_env/bin/maker line 42.
BEGIN failed--compilation aborted at /opt/conda/conda-bld/maker_1520105724755/_t_env/bin/maker line 42.
```  

It's caused by `perl-forks` trying to use a method in perl-list-moreutils that I suspect is present in the older version but not in the newer one, so the only explanation I can think of is that the last time this recipe was created the older version was used and the newer one wasn't present yet. 

So, unfortunately it seems like version locking and including as a run-time dep are the correct ways to handle it. :-/

kastman:
@bgruening Sorry to keep pestering, but it would be great to start using this. Anything else I could do to help?

PR #7935, opened 2018-02-20
update Bamutil

karl616:

### DIFF
--- a/recipes/bamutil/build.sh
+++ b/recipes/bamutil/build.sh
@@ -2,7 +2,9 @@
 
 VERSION=1.0.14
 curl -o libStatGen-v$VERSION.tar.gz -L https://github.com/statgen/libStatGen/archive/v$VERSION.tar.gz
+openssl sha256 libStatGen-v$VERSION.tar.gz | grep 70a504c5cc4838c6ac96cdd010644454615cc907df4e3794c999baf958fa734b
 tar -xzvpf libStatGen-v$VERSION.tar.gz
+# remove profile flag to avoid "ld: cannot find gcrt1.o: No such file or directory"
 sed -i -e 's/-pg//' libStatGen-1.0.14/Makefiles/Makefile.include
 make LIB_PATH_GENERAL=./libStatGen-$VERSION
 make install INSTALLDIR=$PREFIX LIB_PATH_GENERAL=./libStatGen-$VERSION
--- a/recipes/bamutil/build.sh
+++ b/recipes/bamutil/build.sh
@@ -3,6 +3,7 @@
 VERSION=1.0.14
 wget -O libStatGen-v$VERSION.tar.gz https://github.com/statgen/libStatGen/archive/v$VERSION.tar.gz
 tar -xzvpf libStatGen-v$VERSION.tar.gz
+sed -i -e 's/-pg//' libStatGen-1.0.14/Makefiles/Makefile.include
 make LIB_PATH_GENERAL=./libStatGen-$VERSION
 make install INSTALLDIR=$PREFIX LIB_PATH_GENERAL=./libStatGen-$VERSION
 mkdir -p $PREFIX/bin/
--- a/recipes/bamutil/build.sh
+++ b/recipes/bamutil/build.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 VERSION=1.0.14
-wget -O libStatGen-v$VERSION.tar.gz https://github.com/statgen/libStatGen/archive/v$VERSION.tar.gz
+curl -o libStatGen-v$VERSION.tar.gz -L https://github.com/statgen/libStatGen/archive/v$VERSION.tar.gz
 tar -xzvpf libStatGen-v$VERSION.tar.gz
 sed -i -e 's/-pg//' libStatGen-1.0.14/Makefiles/Makefile.include
 make LIB_PATH_GENERAL=./libStatGen-$VERSION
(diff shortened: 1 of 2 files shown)


recipes/bamutil/build.sh
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 
 VERSION=1.0.14
-wget -O libStatGen-v$VERSION.tar.gz https://github.com/statgen/libStatGen/archive/v$VERSION.tar.gz
+curl -o libStatGen-v$VERSION.tar.gz -L https://github.com/statgen/libStatGen/archive/v$VERSION.tar.gz
 tar -xzvpf libStatGen-v$VERSION.tar.gz
+sed -i -e 's/-pg//' libStatGen-1.0.14/Makefiles/Makefile.include
```
mbargull: Can we be make this a bit more specific and descriptive, e.g.:
```bash
# remove profile flag to avoid "ld: cannot find gcrt1.o: No such file or directory"
sed -i.bak '/OPTFLAG_PROFILE?=/ s/-pg//' Makefiles/Makefile.include
```

recipes/bamutil/build.sh
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
 
 VERSION=1.0.14
-wget -O libStatGen-v$VERSION.tar.gz https://github.com/statgen/libStatGen/archive/v$VERSION.tar.gz
+curl -o libStatGen-v$VERSION.tar.gz -L https://github.com/statgen/libStatGen/archive/v$VERSION.tar.gz
```
mbargull: Can you also add
```bash
openssl sha256 libStatGen-v$VERSION.tar.gz | grep 70a504c5cc4838c6ac96cdd010644454615cc907df4e3794c999baf958fa734b
```

recipes/bamutil/meta.yaml
```diff
@@ -8,14 +8,17 @@ source:
   sha256: "f5ec8d5e98a3797742106c3413a4ab1622d8787e38b29b3df4cddb59d77efda5"
 
 build:
-  number: 1
+  number: 2
 
 requirements:
   build:
-    - gcc # [linux]
```
mbargull: Please also add
```yaml
    # curl/openssl are used to download/verify libStatGen
    - curl
    - openssl
```

recipes/bamutil/build.sh
karl616: I added a sentence in the commit, but you are right, it is much better to keep it here.

recipes/bamutil/build.sh
karl616: This is nice! I thought about that the file isn't checked, but didn't think of a way to do it... perhaps a bit too focused on getting it to work

recipes/bamutil/meta.yaml
karl616: Done! Could it be that curl is part of the default set of dependencies? I ran it through circleci locally and it worked

recipes/bamutil/build.sh
mbargull: No worries, there are a couple of recipes don't yet check those downloads. But in future, once we make `conda-build 3` available to Bioconda, we can use multiple source files in `meta.yaml` and can remove those "manual" downloads and checks :wink:

recipes/bamutil/build.sh
mbargull: Ah, didn't even see that one (maybe with the excuse that GitHub collapses extended messages on default :laughing:). Yeah, having a comment beside the code is more obvious (for ppl like me).

karl616:
@mbargull Google helped me with this fix, but it took me quite a while to understand what exactly was going wrong. Thanks for the feedback! It really helps understanding how to do things right

recipes/bamutil/meta.yaml
mbargull: `curl` is already available in the base system, hence it worked fine. If we want to be "good citizens", we are a bit more explicit/robust here.

mbargull:
> but it took me quite a while to understand what exactly was going wrong

I can relate to that :wink:. I haven't tried it out yet, but possibly with newer compilers we can remove that workaround (or "fix" -- why does it even always compile profile builds??) again.
Thanks for figuring this out!

recipes/bamutil/build.sh
```diff
@@ -5,7 +5,7 @@ curl -o libStatGen-v$VERSION.tar.gz -L https://github.com/statgen/libStatGen/arc
 openssl sha256 libStatGen-v$VERSION.tar.gz | grep 70a504c5cc4838c6ac96cdd010644454615cc907df4e3794c999baf958fa734b
 tar -xzvpf libStatGen-v$VERSION.tar.gz
 # remove profile flag to avoid "ld: cannot find gcrt1.o: No such file or directory"
-sed -i.bak '/OPTFLAG_PROFILE?=/ s/-pg//' Makefiles/Makefile.include
+sed -i.bak '/OPTFLAG_PROFILE?=/ s/-pg//' ./libStatGen-$VERSION/Makefiles/Makefile.include
```
mbargull: Mea culpa!

karl616:
@mbargull I'll merge it in... Thanks for the help!

karl616:
Should also say that I have written too little c++ code to have a experienced view on profiling... I think today was my first encounter...

mbargull:
Oh, that was just because of some intricacies of the build system, no worries.
Thanks for the fix and merge!

PR #7939, opened 2018-02-20
updated for package update

oena:

### DIFF
--- a/recipes/cmappy/meta.yaml
+++ b/recipes/cmappy/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "3.0.0" %}
+{% set version = "3.1.0" %}
 
 package:
   name: cmappy
@@ -7,7 +7,7 @@ package:
 source:
   fn: cmapPy-{{ version }}.tar.gz
   url: https://pypi.io/packages/source/c/cmappy/cmapPy-{{ version }}.tar.gz
-  sha256: 7def0596d20a653b0f135926e8a80d5d15d0f722cfc95bbe14ba00aaa519de6b
+  sha256: 450467fa6f2392907b4a4d71003c5492ac948db98fc11b13d387d44a2bc57093
 
 build:
   skip: True  # [not py27] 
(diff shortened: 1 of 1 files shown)


oena:
Thanks for the helpful review @mbargull. Just wondering, I thought gcc & llvm were needed because of bioconda not working with Windows... did this change or am I misunderstanding? Happy to update the requirements otherwise.

mbargull:
You just need `gcc`/`llvm` for packages that use them to compile things. But since you don't compile any C-Extensions, you just need `python` itself (and `setuptools` because that's used by `setup.py`).
As for Windows, yes, Bioconda still only supports Linux and macOS. On Windows you'd just have other compilers/runtimes to (possibly) depend on, but likewise only if you were to compile extensions.
Pure Python packages just need their `python` to be happy :smile:.

oena:
Hmm @mbargull I implemented the changes you suggested and now the CircleCI tests are failing. Thoughts?

mbargull:
Hi @oena, that's a good test failure! I mean, annoying for you, but it surfaced an issue in a very popular package, `pandas`. I opened a PR on the `pandas` feedstock to fix this: https://github.com/conda-forge/pandas-feedstock/pull/40.
Do you need the `cmappy` update urgently or would it be fine for you to wait a bit on the fix to be applied upstream?
Also, acc. to `setup.py`, `six` is also a run dependency, right?

oena:
Aha, thanks for filing that. I'm willing to wait for a little, but obviously would also like this update to occur sooner rather than later since it fixes a bug that potentially could cause people some problems. I guess we'll see. 

And `six` isn't a requirement anymore for the time being; you are correct that the setup.py file is inconsistent about that though (will update). Basically a contributor added some changes to implement some Python 3 compatibility (but not all the way), and it ultimately ended up causing more problems than we had the bandwidth to address so we removed py3 compatibility for the time being.

mbargull:
> I'm willing to wait for a little, but obviously would also like this update to occur sooner rather than later since it fixes a bug that potentially could cause people some problems.

Let's "fix" this here then for now :wink:

oena:
Works for me. Thanks!

PR #7940, opened 2018-02-20
Identifiers added to the abyss container

ypriverol:

### DIFF
--- a/recipes/abyss/1.5.2/meta.yaml
+++ b/recipes/abyss/1.5.2/meta.yaml
@@ -47,3 +47,6 @@ extra:
   container:
     # openmpi needs ssh/rsh
     extended-base: true
+  identifiers:
+    - biotools:abyss
+    - doi:10.1101/gr.089532.108
--- a/recipes/abyss/1.9.0/meta.yaml
+++ b/recipes/abyss/1.9.0/meta.yaml
@@ -49,3 +49,6 @@ extra:
   container:
     # openmpi needs ssh/rsh
     extended-base: true
+  identifiers:
+    - biotools:abyss
+    - doi:10.1101/gr.089532.108
--- a/recipes/abyss/meta.yaml
+++ b/recipes/abyss/meta.yaml
@@ -49,3 +49,6 @@ extra:
   container:
     # openmpi needs ssh/rsh
     extended-base: true
+  identifiers:
+    - biotools:abyss
+    - doi:10.1101/gr.089532.108


bgruening:
@bioconda/core @epruesse @peterjc  what do you think about this one?
xref: https://github.com/bioconda/bioconda-utils/issues/252

Imho, this is a very good step forward. If we want to change the layout later on, this should not be a problem and we can do this with some scripts. Imho we should start annotating our tools and use the momentum here. Some people from ELIXIR will help us with this :)

peterjc:
The sample syntax looks clean - but is there a spec (I know how to resolve a DOI, but what about the other potential prefixes like ``biotools``) and/or recipe changes to lint and ultimately consume this metadata?

ypriverol:
@peterjc good point, from my point of view we should use compact identifiers in the way ```domain: identifier```. We will start with the biotools:identifier because we know for more than 1000 entries the corresponding bioconda package. We also know an easy way to resolve the identifiers because the url prefix is always https://bio.tools/abyss

rvalieris:
wouldn't it be better to write:
```
identifiers:
  biotools: abyss
  doi: i10.1101/gr.089532.108
```

so that the yaml package can properly parse these ?

ypriverol:
@rvalieris @bgruening do you want me to update the PR. ?

bgruening:
I do think this is a nice improvement, but lets wait a little bit longer before you spend time on it :)

mbargull:
Ref. from the previous discussion: https://github.com/bioconda/bioconda-recipes/pull/6294#issuecomment-336617692
> I have changed to a new structure: 
```
extra:
  identifiers:
     biotools:
         - http://bio.tools/abyss
```

----
I think it could be fine to use the pure identifiers only (as currently proposed) instead of a full URL. What we additionally need, are multiple identifiers per kind. And ABySS is a good example for this, as it has multiple DOIs: https://github.com/bcgsc/abyss/tree/de8c892583cdf677faeaf7d73b06b76743f03947#citation.
If we want to make the `identifiers` section concise, we could allow both, a single string or a list of strings per identifier kind:
```
extra:
  identifiers:
    biotools: abyss
    doi:
      - 10.1101/gr.214346.116
      - 10.1101/gr.089532.108
```

epruesse:
I like it! :)

Questions:
 
 - Why `identifiers` and not, say, `xref`?
 - Keeping multiple DOIs is useful, but then there is often a "canonical" one that is supposed to be cited, and that's not necessarily the most recent one. Specifying that it's "the first one in the list" would work, but it's not very readable given the cryptic nature of DOI references.

ypriverol:
@epruesse 


>Why identifiers and not, say, xref?

We can be called xref but I think people would be more identified with `identifiers`. The idea of mention identifiers is because we want to reference other tools registries that mention the same tool. I like the idea of using `xref` also..It needs to be decided by the community.  

>Keeping multiple DOIs is useful, but then there is often a "canonical" one that is supposed to be cited, and that's not necessarily the most recent one. Specifying that it's "the first one in the list" would work, but it's not very readable given the cryptic nature of DOI references.

Here the doi, for example, can be references to multiple publications related to the same tool. For example the main paper could be the first `doi` and the second could be a manual publication. 

@mbargull definitely the way to reference the multiple ids should be the way you are proposing: 

```yml
extra:
  identifiers:
    biotools: abyss
    doi:
      - 10.1101/gr.214346.116
      - 10.1101/gr.089532.108
```

epruesse:
So what we want is *link* a package to other online resources. Technically, there is already a pretty generic way to do that, involving *uniform resource locators* (aka URLs, or maybe URIs)... ;-)

Raw, it looks a little ugly:

```yaml
extra:
  identifiers:
    - http://bio.tools/abyss
    - http://doi.org/10.1101/gr.214346.116
    - http://doi.org/10.1101/gr.089532.108
```

Though making `http://` the default if nothing else is specified, it's not so different from what you suggested:

```yaml
extra:
  identifiers:
    - bio.tools/abyss
    - doi.org/10.1101/gr.214346.116
    - doi.org/10.1101/gr.089532.108
```

We could allow annotations, like this:

```yaml
extra:
  identifiers:
    - bio.tools/abyss
    - primary: doi.org/10.1101/gr.214346.116
    - doi.org/10.1101/gr.089532.108
```

That data is easily parsed and interpreted:
  - Publications are links pointing to **pubmed** , **doi**, ....
  - Software directories are links pointing to **bio.tools**, **omicstools**, ...;
  - Contact addresses start with `"mailto:"`
  - ...

Most links are easily classified by checking for a prefix of the URI. Where multiple instances of the same type are needed, annotating could be used to distinguish them, e.g. using `primary:` to mark one as more important than the others (the publication to be used for citing, or the email to be used for asking for help.

Alternatively, if you guys prefer a more form-style format with things to be filled in, the type annotation could be mandatory, at which point it might as well be a top level entry like before:

```yaml
extra:
  identifiers:
    directories:
      - bio.tools/abyss
      - omictools.com/abyss-tool
    publications:
      - primary: doi.org/10.1101/gr.214346.116
      - doi.org/10.1101/gr.089532.108
    contacts:
      - primary: groups.google.com/d/forum/abyss-user
      - mailto:sjackman at gmail
    releases: github.com/bcgsc/abyss/releases
    
```

epruesse:
TL;DR the above:

 1. Writing `bio.tools/abyss` instead of `biotools: abyss` turns an annotation we have to know things about into a URI we don't have to understand to process.
 2. Generic groups like `directories` or `publications` are more meaningful and less plentiful than `biotools`, `omicstools`, `doi`, `pubmed`, `crossref`, etc. 
 3. Named groups of URIs can be easily turned into a pretty and helpful package page.
 4. Grabbing specific references such as the biotools entry out of the data is a trivial prefix match.

ypriverol:
Hi @epruesse:

We have a long discussion in biocontainers about using or not compact identifiers like `biotools:abyss` or the `full url`. With this mechanism, we can allow both:  

```yml  
extra:
   identifiers:
       - bio.tools:abyss
       - doi.org/10.1101/gr.214346.116
       - doi.org/10.1101/gr.089532.108
```
I understand that the url make a lot of sense for `doi` because this will not change or for example for identifiers.org. However, for some services, we should allow `compact identifier` for two main reasons: 

1- URLs can change in the future, then we don't need to change our url only the logic of the tool that will build the nice web page or reference. 

2- In some cases would be nice to reuse that id to query API, or other urls that are based in the same identifier. For example, it is now well promoted that compact identifiers get to use in publications rather that url, then journals are using identifiers.org to resolve the reference. biotools can be now resolve by this mechanism. 

I vote 100% for grouping the identifiers in multiple groups. Then, I will refine your proposal with the following approaches: 

```yml
extra:
  identifiers:
    directories:
      - biotools:abyss
      - omictools.com/abyss-tool
    publications:
      - primary: doi.org/10.1101/gr.214346.116
      - doi.org/10.1101/gr.089532.108
    contacts:
      - primary: groups.google.com/d/forum/abyss-user
      - mailto:sjackman at gmail
    releases: github.com/bcgsc/abyss/releases
```
The only different is that in some cases we will allow compact identifiers.

bgruening:
I'm not a fan of the grouping at least not as this stage. We making things to complicated at the beginning and inventing another ontology that we need to explain and maintain.
Lets start very simple and see how this is accepted by the community and what people will add. Groups can be added easily in a second step if really needed.

epruesse:
@ypriverol Ahhhh - so the term `identifiers` derives from `identifiers.org`.

Allowing to use URI's with default http:// as well as the identifiers.org name space makes things difficult to parse though. In the example, `mailto:` could refer to either a URI scheme or a identifiers.org namespace.

Options:
 1. Stick to URI as the established means of linking resources:
    - ` - identifiers.org/biotools/abyss`
    - ` - urn:miriam:biotools/abyss` (don't like this as much)
    ```yaml
    extra:
      identifiers:
        directories:
          - identifiers.org/biotools/abyss
          - omictools.com/abyss-tool
        publications:
          - primary: doi.org/10.1101/gr.214346.116
          - doi.org/10.1101/gr.089532.108
        contacts:
          - primary: groups.google.com/d/forum/abyss-user
          - mailto:sjackman at gmail
        releases: github.com/bcgsc/abyss/releases
    ```
 2. Make `https://identifiers.org/` the default prefix
    ```yaml
    extra:
      identifiers:
        directories:
          - biotools/abyss
          - http://omictools.com/abyss-tool
        publications:
          - primary: doi/10.1101/gr.214346.116
          - doi/10.1101/gr.089532.108
        contacts:
          - primary: http://groups.google.com/d/forum/abyss-user
          - mailto:sjackman at gmail
        releases: https://github.com/bcgsc/abyss/releases
    ```
  3. Create two sections, one called `identifiers` and one called `xref` or `links`, for identifiers.org identifiers and URIs, respectively. 
     ```yaml
     extra:
      identifiers:
        directories:
          - biotools: abyss
        publications:
          - primary: doi/10.1101/gr.214346.116
          - doi/10.1101/gr.089532.108
      links:
        directories:
          - http://omictools.com/abyss-tool
        contacts:
          - primary: http://groups.google.com/d/forum/abyss-user
          - mailto:sjackman at gmail
        releases: https://github.com/bcgsc/abyss/releases
    ```

I guess my mind is more on "annotating packages with cross references" than on "using identifiers.org", so personally I'm leaning towards option 1.

epruesse:
@bgruening 
> I'm not a fan of the grouping at least not as this stage. We making things to complicated at the beginning and inventing another ontology that we need to explain and maintain.
Lets start very simple and see how this is accepted by the community and what people will add. Groups can be added easily in a second step if really needed.

Well... changes to this might require, or at least trigger, a package rebuild and upload etc, so while not a huge issue, it'd certainly be nice to get it right the first time round.

I'm not sure that a half-dozen terms qualify as an ontology, but I do hear you. So just visualizing it without groups:

   ```yaml
    extra:
      identifiers:
          - identifiers.org/biotools/abyss
          - omictools.com/abyss-tool
          - primary: doi.org/10.1101/gr.214346.116
          - doi.org/10.1101/gr.089532.108
          - primary: groups.google.com/d/forum/abyss-user
          - mailto:sjackman at gmail
          - github.com/bcgsc/abyss/releases
   ```

Just visually, I don't like it. Even though the groups don't add anything semantic that can't be specified wholesale somewhere else, they do give some structure. 

Just so we're all on the same page - where and how do we want to use these annotations?

I'm seeing these use cases:
 1. Displaying links on the package pages ([bioconda abyss package](https://bioconda.github.io/recipes/abyss/README.html))
 2. Preparing a list of things-to-cite from a conda environment
 3. Step towards a mechanism to check for updated releases

Where 1. really will be the primary use case. 

Or am I missing something?

bgruening:
> Well... changes to this might require, or at least trigger, a package rebuild and upload etc, so while not a huge issue ...

No it will not. Only if we increase the build version.

> it'd certainly be nice to get it right the first time round.

Indeed it would, but I don't see this realistically happen. We had this discussion 6 month back already and this PR was just abandoned. People are motivated now to spend time on this, for the second time ;)

> I'm not sure that a half-dozen terms qualify as an ontology, but I do hear you. 

Yeah, I think this is how every ontology starts :)

> Just visually, I don't like it.

Remove the primary and it looks better ;)

> Just so we're all on the same page - where and how do we want to use these annotations?

Not yet clear. The community will be creative I'm sure and if there are new use-case we will adapt to those. Currently, the bio.tool and doi will most likely be used to inter-link bio.tools with OpenEBench, Galaxy TS and so on.

> Where 1. really will be the primary use case.

That at least not the use-case @ypriverol has afaik. Bioconda can make use of this and we can add groups by mapping identifiers as we define them. But the intend of these ids  is currently to have a mapping between different registries, package managers etc ...

epruesse:
I don't want to hold this up. As @bgruening said, if we don't bump the package build number, changing things later is easy enough. So merge as you wish with my blessing ;-)

W.r.t. bio.tools and identifiers: I don't think there are that many annotations a package would usefully have, and we don't need to replicate those annotations everywhere. If I understand the situation right, bio.tools would like to have back-links, so let's just add a section to 'extra:' where those can go. Personally, I still believe staying within the URI scheme is useful as it requires least knowledge on this side. If the scheme needs to be persistent with high reliability on very long term, contract based solutions like the one used by doi.org to ensure permanent availability of the scheme could be implemented by bio.tools.

bgruening:
Thanks @epruesse!

ypriverol:
@epruesse I will open an issue with all the comments you have done in biocontainers. The idea is to keep track of these suggestions for future improvements. I agree with @bgruening that in order to move forward the present PR we should try to implement this simple version and see how it fly. Here the Issue in BioContainers.  https://github.com/BioContainers/specs/issues/92

bgruening:
Cool thanks!
@ypriverol I guess this means we can proceed and encourage everyone to add DOIs and bio.tools IDs.

ypriverol:
@bgruening 
I will start coordinating the inclusion of more recipes. I will let you know how we will do it. Thanks, @epruesse for your understanding.

PR #7942, opened 2018-02-20
Compile dsrc under macox

cokelaer:

### DIFF
--- a/recipes/dsrc/meta.yaml
+++ b/recipes/dsrc/meta.yaml
@@ -12,7 +12,7 @@ source:
   md5: {{ md5 }}
 
 build:
-  skip: False 
+  skip: True # [osx]
   number: 3
   string: boost{{CONDA_BOOST}}_{{PKG_BUILDNUM}}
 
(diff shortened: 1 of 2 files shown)


recipes/dsrc/meta.yaml
```diff
@@ -1,33 +1,37 @@
+{% set name = "dsrc" %}
+{% set version = "2014.12.17" %}
```
druvus: why is version set to a date instead of real version?

cokelaer: I don't remember what was the rationale for that decision. I guess we can switch to 2.0.2 but I'm wondering what would be the effect if we switch to 2.0.2. I mean how the version 2014.12.17 would compare to 2.0.2 Unfortunately 2.0.2 is smaller... but the build number may help if we increase it.

cokelaer:
the macosx compilation fails due to lib boost that cannot be found. I will need to figure out what is missing or wrong in the build.sh file.

cokelaer:
won't fix the macosx version for now. Closing the pull request and will be done later re-using the different proposal to be found in this pull request.

PR #7963, opened 2018-02-21
Trying to update samtools build with ncurses

jpuritz:

### DIFF
--- a/recipes/samtools/build.sh
+++ b/recipes/samtools/build.sh
@@ -23,5 +23,5 @@ make
 cd ..
 # Problem with ncurses from default channel we now get in bioconda so skip tview
 # https://github.com/samtools/samtools/issues/577
-./configure --prefix=$PREFIX --enable-libcurl --enable-plugins --with-plugin-path=$PREFIX/htslib-$PKG_VERSION LDFLAGS='-Wl,--add-needed'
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
AFAIK it we should have ncurses in the meta.yaml if we're going to build with it.  E.g. it's bad to have requirements that are not explicitly annotated.

jpuritz:
Ok, I think I made all the appropriate changes.  This should allow the reactivate the `tview` module of samtools

rvalieris:
>checking for gcc... gcc
checking whether the C compiler works... no
configure: error: C compiler cannot create executables

it doesn't look like this issue was introduced by this patch, but I don't know either. ping @bioconda/osx

mbargull:
Could be that it's because the linker flags have been overridden. Let's see what the logs show in a bit..

mbargull:
```
configure:2449: checking whether the C compiler works
configure:2471: gcc  -arch x86_64 -I/Users/distiller/project/miniconda/conda-bld/samtools_1519762966388/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/include -Wl,--add-needed conftest.c  >&5
ld: unknown option: --add-needed
```
Removed that flag for macOS -- is that fine?

jpuritz:
Awesome.  Thanks for the help.

PR #7972, opened 2018-02-22
Add libcurl as dependency.

roryk:
This library does not install correctly on Red Hat 4.4.7-4 without installing libcurl first. I confirmed that installing libcurl fixes this issue.

```
Error: package or namespace load failed for ‘GenomeInfoDb’ in dyn.load(file, DLLpath = DLLpath, ...):
unable to load shared object '/gpfs/ngs/usr/safary/miniconda/envs/bcb_rna/lib/R/library/RCurl/libs/RCurl.so':
/gpfs/ngs/usr/safary/miniconda/envs/bcb_rna/lib/R/library/RCurl/libs/../../../../libcurl.so.4: undefined symbol: SSL_CTX_set_alpn_protos
Error: package ‘GenomeInfoDb’ could not be loaded
```


roryk:
Working with the folks on the conda `r-rcurl` feedstock to fix it there, since that seems like the root of the problem.

PR #7978, opened 2018-02-22
Fixing racon

druvus:

This update uses git_url even if it is not recommended to fix the broken 0 first build. A typo in the recipe made it to download previous version. The reason for using git_url is 5 additional submodules in this updated version of racon. Maybe we could make an exception until the submodules are added to bioconda/conda-forge?

### DIFF
--- a/recipes/racon/build.sh
+++ b/recipes/racon/build.sh
@@ -3,8 +3,7 @@
 mkdir -p $PREFIX/bin
 mkdir build
 cd build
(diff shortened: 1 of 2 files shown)


druvus:
Working on adding separate sub modules as individual packages

rvalieris:
FYI: you can get an archive link of any commit like this:
`https://github.com/isovic/racon/archive/5cfd1499bcec6e8b2a6584717cb6a3e7ae11b85e.zip`

that way you don't have to skip the lint checks.

PR #7997, opened 2018-02-25
update perl-yaml to 1.24

mtw:

### DIFF
--- a/recipes/perl-yaml/1.24/meta.yaml
+++ b/recipes/perl-yaml/1.24/meta.yaml
@@ -1,0 +1,34 @@
+{% set name = "perl-yaml" %}
+{% set version = "1.24" %}
+{% set sha256 = "b58d045a72a5e1ce714f1d42b64ccddbb055a4afac10b74972dba3b6a47dffb6" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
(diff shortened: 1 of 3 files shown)


nsoranzo:
@mtw Any specific reason why you added the `1.24/` subdirectory instead of updating the main `recipes/perl-yaml/meta.yaml` ?

recipes/perl-yaml/1.24/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+    - perl >=5.22,<6.0
+    - perl-extutils-makemaker
+    - perl-test-yaml
+    - perl-test-more
+
+  run:
+    - perl >=5.22,<6.0
+
+test:
+  impprts:
```
nsoranzo: This should have been `imports:`

mtw: no special reason for that

mtw: thanks :)

nsoranzo: OK, I think the best practice is to modify the main `meta.yaml`, unless you need to keep an old version around.
Updated to 1.26 in https://github.com/bioconda/bioconda-recipes/pull/9876

PR #8007, opened 2018-02-26
Pin flexmix for bioconductor-scde.

jdblischak:

The latest version of flexmix (2.3-14) broke scde. Needs to be pinned to 2.3-13:

https://hms-dbmi.github.io/scde/help.html?place=msg%2Fsinglecellstats%2FrbFUTOQ9wu4%2Fyg4cxLbeAAAJ

xref: #7859 https://github.com/conda-forge/r-flexmix-feedstock/commit/92f7ee3936cc1dfc5483473fb19524c99f95ccee

Unrelated, but I also learned that the scde has an additional license file that restricts commercial use:

https://github.com/hms-dbmi/scde/blob/master/license.txt

Should I add this file and insert a `license_file` line? Is there anyway to alert users in `meta.yaml` that the software isn't really licensed under GPL-2?

### DIFF
--- a/recipes/bioconductor-scde/meta.yaml
+++ b/recipes/bioconductor-scde/meta.yaml
@@ -12,7 +12,7 @@ source:
     - 'https://depot.galaxyproject.org/software/{{ name }}/{{ name }}_{{ version }}_src_all.tar.gz'
   sha256: 226c6ac1699c507546144313027b844280092e311da1fb4cb10b288d6854aa33
 build:
-  number: 0
+  number: 1
   rpaths:
     - lib/R/lib/
     - lib/
@@ -24,7 +24,7 @@ requirements:
     - r-base
     - r-cairo
     - r-extremes
-    - r-flexmix
+    - r-flexmix <=2.3_13
     - r-mass
     - r-mgcv
     - r-nnet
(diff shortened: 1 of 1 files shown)


rvalieris:
good catch with the license, I think we should set the about.license field to "custom" and include the license file, according to dev comments the software is really not intended to be distributed with a GPL-2 license. https://github.com/hms-dbmi/scde/issues/38

bgruening:
@jdblischak not sure what to do about the license. We should at least change it to `others` or something, as @rvalieris pointed out. Do you know how bioconductor is handling this?

jdblischak:
Bioconductor lists the license as GPL-2: https://www.bioconductor.org/packages/release/bioc/html/scde.html

I assume they do not know about the non-standard license because their guidelines require adding this file to the repository: https://www.bioconductor.org/developers/package-guidelines/#license

bgruening:
So should be back-off here, ping bioconductor and pressure upstream to change this?

jdblischak:
> So should be back-off here, ping bioconductor and pressure upstream to change this?

@bgruening Agreed. Let's merge this minor fix of the dependencies and save the license issue until later.

What's the best way of contacting the Bioconductor team about a license issue like this? The support site doesn't seem appropriate since that is for software usage questions. And please let me know if you'd like me to reach out to Bioconductor or if you'd prefer to do it.

bgruening:
I will ask a friend for a contact and let you know.

PR #8013, opened 2018-02-27
Kaiju: Scripts have been moved to util directory

dfornika:

### DIFF
--- a/recipes/kaiju/build.sh
+++ b/recipes/kaiju/build.sh
@@ -9,14 +9,17 @@ make
 cd $SRC_DIR/bin/
 cp addTaxonNames $PREFIX/bin
 cp convertNR $PREFIX/bin
-cp gbk2faa.pl $PREFIX/bin
 cp kaiju $PREFIX/bin
 cp kaiju2krona $PREFIX/bin
 cp kaijup $PREFIX/bin
 cp kaijuReport $PREFIX/bin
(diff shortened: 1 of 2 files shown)


recipes/kaiju/build.sh
```diff
@@ -9,14 +9,17 @@ make
...
 cp kaiju $PREFIX/bin
 cp kaiju2krona $PREFIX/bin
 cp kaijup $PREFIX/bin
 cp kaijuReport $PREFIX/bin
 cp kaijux $PREFIX/bin
-cp makeDB.sh $PREFIX/bin
 cp mergeOutputs $PREFIX/bin
 cp mkbwt $PREFIX/bin
 cp mkfmi $PREFIX/bin
+
```
mbargull: Instead of all there changes, you should just add
```bash
cp convert_mar_to_kaiju.py $PREFIX/bin
```

recipes/kaiju/meta.yaml
```diff
@@ -10,7 +10,7 @@ source:
     sha256: ee6c319ab3e58db846139810705bfe33306618d78d75d715fac623f302be9e72
 
 build:
-  number: 0  
+  number: 1
```
mbargull: Instead of the build number increase, can you update the package to version `1.6.2` and update the `sha256` accordingly?

PR #8017, opened 2018-02-27
[WIP] bump rmats

daler:

The new version of rMATS ([homepage](http://rnaseq-mats.sourceforge.net/)) comes with C extensions pre-compiled. I'm unable to find the source, so here I'm copying the precompiled extensions in a platform-specific manner.

EDIT: please hold off on merging until I work out the license with the authors.


daler:
@bioconda/osx any advice on this error? This package comes pre-compiled and I'm unable to find the source to recompile.

https://circleci.com/gh/bioconda/bioconda-recipes/4088

```
ImportError: dlopen(/Users/distiller/project/miniconda/conda-bld/rmats_1519758907192/_t_env/rMATS/rmatspipeline.so, 2): Library not loaded: /usr/local/opt/gcc@5/lib/gcc/5/libgomp.1.dylib
  Referenced from: /Users/distiller/project/miniconda/conda-bld/rmats_1519758907192/_t_env/rMATS/rmatspipeline.so
  Reason: image not found
```

mbargull:
It doesn't find the OpenMP lib. In theory you could try to use `install_name_tool -change` to that hardcoded `/usr/local/opt/gcc@5/lib/gcc/5/libgomp.1.dylib` path to `"${PREFIX}/lib/libgomp.1.dylib"`. But then there is the problem that the latest `libgcc` build on `osx-64` doesn't include that library anymore. With `libgcc <4.8.5  # [osx]` you'd get that file but, anyway, I don't know if GCC 4 vs. GCC 5 introduces any problems...
Hm, maybe just `skip: True  # [osx]`?

yesimon:
I would also recommend skipping osx

daler:
Easy enough! Thanks both.

mbargull:
I took a quick look but couldn't find any license information. Are you sure it's MIT licensed?

daler:
Ah good point! I had copied the earlier recipe. I will contact the authors to clarify before merging.

dpryan79:
@daler We just had someone on biostars ask about this. The license is on the website and looks like be a typical "free for non-commercial use" license like Kallisto used to have. It allows redistribution, but an commercial users would need a separate license. I don't know if that's considered acceptable or not, it's certainly a bit problematic.

daler:
@dpryan79 no updates to speak of; I had contacted the authors twice but no answer . . . and then this dropped off my radar. I suspect we have quite a few tools with similar licenses already and for that reason would lean toward including it. That's really just passing the buck to the user though.

dpryan79:
We've since updated this, so I'll close this PR.

PR #8028, opened 2018-02-28
molpopgen-analysis : add a warning following PR7563

pbordron:
Add a warning following https://github.com/bioconda/bioconda-recipes/pull/7563#issuecomment-367736770

### DIFF
--- a/recipes/molpopgen-analysis/build.sh
+++ b/recipes/molpopgen-analysis/build.sh
@@ -9,9 +9,9 @@ echo "#!/bin/bash
 echo '
 /!\ Warning/!\\
 
-DO NOT USE THIS PACKAGE IN NGS STUDIES.
+DO NOT USE molpopgen-analysis IN NGS STUDIES.
 
-This is a RETIRED package for the (pre-NGS-era) analysis of population-genetic data.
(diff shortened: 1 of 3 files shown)


rvalieris:
in this case I think we should add a post-link warning, so that the deprecated message will appear after the package is installed.

see here for information: https://conda.io/docs/user-guide/tasks/build-packages/link-scripts.html
example: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/variant-effect-predictor/post-link.sh

pbordron:
Maybe for cluster users having a warning when the user loads the environment is a good idea.

rvalieris:
could be, I never tried but you can create `$PREFIX/etc/conda/activate.d/warning.sh`
see: https://conda.io/docs/user-guide/tasks/manage-environments.html#saving-environment-variables

rvalieris:
looks good just 2 more things:
1. include the name of the package in the warnings, so that the user know where its coming from
2. I think the activate script should also contain the name of the package because the file will be installed in a common place, it could conflict with other packages.

rvalieris:
perfect ! thanks a lot !

Issue #4945, opened 2017-06-12
rna-seqc bug


rlegendre:
Hi,

I try to use the rna-seqc software, installed with conda (thanks for adding it @holtgrewe !) but I encounter an error when running. Note that my local version works correctly with the same settings

Error log is in attach
[RNAseQC.txt](https://github.com/bioconda/bioconda-recipes/files/1067358/RNAseQC.txt)

Thanks for solving

daler:
Would you please provide the commands that cause this error? Better yet would be a minimal example with input files that reproduces the problem.

rvalieris:
it seems rna-seqc is using a version of javassist that is incompatible with java 8.  runniong with java 7 works.

rlegendre:
I tried with a minimal example (only one bam file). If I give the bam with "-s myfile.bam", it works.. if I give a [sample_index.txt](https://github.com/bioconda/bioconda-recipes/files/1070585/sample_index.txt), I have the error !
Here my command line : 
`rna-seqc -o RNAseQC  -r chr19.fa -s sample_index.txt -t Genome/mm9.gtf  -BWArRNA mouse_rRNA/mouse_all_rRNA.fasta`

ranjit58:
Adding some more info.
The conda install also forces to install Java8 (openjdk) version which is not compatible. It works when provided with java 7 (java-jdk: 7.0.91). 

It might be more clear from this docker file
https://hub.docker.com/r/alexiswl/rnaseqqc_1.1.8/~/dockerfile/

Hope it helps.

rvalieris:
sorry, I forgot about this, since rna-seqc is not being updated anymore the best we can do in this case is pin the recipe to use java 7. I will submit a PR.

rvalieris:
package rna-seqc 1.1.8-1 is up, and pins openjdk 7.

however, conda seems to still be preferring to install the previous version. @bgruening could we mark the -0 builds as broken ?

bgruening:
Done and there is now also a new version. This should now work. Please create a new issue if you encounter any problems.

PR #8030, opened 2018-02-28
pin recipe to use java 7, see #4945

rvalieris:

ping recipe to use java 7.
rna-seqc is using a version of javassist that is incompatible with java 8.


ranjit58:
how about replacing it with java-jdk (instead of openjdk) v 7

PR #8044, opened 2018-02-28
Updating runtime dependency of vcfpy v0.11

holtgrewe:
`python` is missing

### DIFF
--- a/recipes/vcfpy/meta.yaml
+++ b/recipes/vcfpy/meta.yaml
@@ -13,12 +13,13 @@ source:
 build:
   skip: True # [py27]
   script: python setup.py install --single-version-externally-managed --record record.txt
-  number: 0
+  number: 1
 
 requirements:
   build:
   - python
   run:
+  - python
   - pysam >=0.10.0
 
 test:


holtgrewe:
@bioconda/python Hm, could someone help me here? I have a problem with my package (I maintain `vcfpy` itself) and I thought I could fix this by adding the runtime dependency of `python`...

So (1) this build fails now. What am I doing wrong?

Also (2) with a fresh Miniconda install and Python 3.6, it fails as well:

```
# conda install vcfpy
# python --version
Python 3.6.4
# python -m vcfpy
/tmp/miniconda3/bin/python: No module named vcfpy
# conda install python=3.5
# python --version
Python 3.5.5
# python -c 'import vcfpy'
-> works
```

mbargull:
@bgruening: The build failure is due to `mulled` pulling `continuumio/miniconda3:latest` which, unfortunately, has been updated to `4.4.10` a couple of hours ago: https://hub.docker.com/r/continuumio/miniconda/tags/. That caused the `mulled` build to run into https://github.com/conda/conda/issues/6811

holtgrewe:
Ah, at least it's good to see that it's not my fault and software packaging is painful everywhere.

yikes! Thanks, I'll sit this one out then, I guess.

mbargull:
I don't understand 
> Also (2) with a fresh Miniconda install and Python 3.6, it fails as well:

yet, though..

mbargull:
Which version of `conda` did you use when you ran the update `python` update in (2)?

mbargull:
Ah, sorry, misread
```bash
# conda install vcfpy
# python --version
Python 3.6.4
# python -m vcfpy
/tmp/miniconda3/bin/python: No module named vcfpy
```
So, instead of the previous question: What version of `vcfpy` was installed by `conda install vcfpy`?

holtgrewe:
@mbargull The one in channel `bioconda`: `0.11.0-0`

mbargull:
(Urgh, brain is slipping... should go to sleep...)
Right, that `vcfpy` package is broken since it is compiled with `python 3.5` but not pinned to that version. Your changes in this PR will fix this. We'll need to label the previous build as `broken` once this is merged and built.

holtgrewe:
@bgruening @mbargull Can I do anything to speed up the resolution?

bgruening:
@mbargull that would be bad. But this container is only used to install a package, this would mean that we can not install bioconda packages 4.4.10? Also I'm wondering why other PR pass if the container is broken. I will try to look into it at the evening.

mbargull:
That bug is in `conda >4.4.7,<4.4.11` and is only hit with installing certain packages (and with Docker/overlay2 fs). See the comments from https://github.com/conda/conda/issues/6811#issuecomment-364894274 onward.
I don't know if it reliably fails or not. I'll just restart the build here one or two times.

mbargull:
@holtgrewe: There is nothing I can currently do to resolve this faster. For now, we need changes from upstream project for this...

holtgrewe:
Yay software development!

Thanks for the info. Let's wait.

holtgrewe:
@mbargull How do I find out if it is fixed upstream?

mbargull:
Hi Manuel, sorry for not informing you earlier, but @bgruening found a nice little workaround in https://github.com/bioconda/bioconda-recipes/pull/8056. Please just merge `master` and you should be good to go.

holtgrewe:
Yay! @mbargull should we review regardlessly?

holtgrewe:
I guess I'll just merge.

mbargull:
Yes, totally fine, thanks!

PR #8054, opened 2018-03-01
Add libgcc runtime dependency for libgomp.so

davebx:

Without this dependency, some tool runs will fail with the message `error while loading shared libraries: libgomp.so.1: cannot open shared object file: No such file or directory`


mbargull:
It doesn't really make sense to add that dependency here as the package itself consist only of a single Perl script. Can you open up an issue describing which tools fail? In that issue, please provide the platform (Linux/macOS), the `conda` version and version of `blast` and `hmmer` that were installed.

PR #8140, opened 2018-03-02
Anvio 4 no cherrypy

jmeppley:
* [* ] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [* ] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).
* [* ] This PR updates an existing recipe.


druvus:
Any specific reason it is only built for py35? Could it be relaxed to `skip: True # [py27]` so it is built for both py35 and py36. The py36 build seems to pass circleci testing.

jmeppley:
That's all I could get to work back at version 2.4. I don't think it worked on py36 at the time. I haven't tried it since. If it passes the tests, then it may be OK, but I would want to do some real tests first.

I have a bunch of people asking for anivo 4 in conda, so I'd prefer to let this go through as is, and then work on testing it under python 3.6. I would push a new build once that's confirmed.

PR #8149, opened 2018-03-04
add old version of fatotwobit

bgruening:


dpryan79:
Is this still needed? That is, do we need to update the recipe for cb3? It's failing at the moment and I presume the presence of mysql in the run and build environment now removes the need for this.

PR #8155, opened 2018-03-05
Quote arguments in EMBOSS wrapper scripts.

peterjc:
This fixes examples where filenames include spaces, e.g. https://github.com/biopython/biopython/issues/1523

```
$ needle -outfile="Emboss/temp with space.needle" -asequence=asis:ACCCGGGCGCGGT -bsequence=asis:ACCCGAGCGCGGT -gapopen=10 -gapextend=0.5 -snucleotide=True

$ water -outfile="Emboss/temp with space.water" -asequence=asis:ACCCGGGCGCGGT -bsequence=asis:ACCCGAGCGCGGT -gapopen=10 -gapextend=0.5
```

Without the fix, these examples used "Emboss/temp" as the output filename.

### DIFF
--- a/recipes/emboss/fix_acd_path.py
+++ b/recipes/emboss/fix_acd_path.py
@@ -14,7 +14,7 @@ export EMBOSS_ACDROOT=../share/EMBOSS/acd/
 export EMBOSS_DATA=../share/EMBOSS/data
 export PLPLOT_LIB=../share/EMBOSS
 
-_water $@
+_water "$@"
 
 --------------------------------------------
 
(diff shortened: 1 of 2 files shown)


recipes/emboss/fix_acd_path.py
```diff
@@ -14,7 +14,7 @@
 export EMBOSS_DATA=../share/EMBOSS/data
 export PLPLOT_LIB=../share/EMBOSS
 
-_water $@
+_water "$@"
```
bgruening: Maybe single quotes? Do we want the shell code to be evaluated?

peterjc: We do want to allow environment variables to be used, e.g.

```
$ water -outfile="$PWD/temp with space.water" -asequence=asis:ACCCGGGCGCGGT -bsequence=asis:ACCCGAGCGCGGT -gapopen=10 -gapextend=0.5
```

Of course, ``$PWD`` is a silly example - something like ``$PROJECT`` or whatever might make sense.

Anyway, this example breaks using  ``_water '$@'`` so the double quote version seems better to me.

bgruening: Makes sense, just double checking.

peterjc:
Thank you!

PR #8205, opened 2018-03-08
Add bxtools (10X sequencing tool)

brainstorm:

### DIFF
--- a/recipes/bxtools/build.sh
+++ b/recipes/bxtools/build.sh
@@ -8,4 +8,4 @@ export LIBS=-lpthread
 ./configure --prefix=$PREFIX
 make
 mkdir -p $PREFIX/bin
-cp src/variant $PREFIX/bin
+cp src/bxtools $PREFIX/bin
--- a/recipes/bxtools/build.sh
+++ b/recipes/bxtools/build.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+set -eu -o pipefail
+
+export CPPFLAGS=-I$PREFIX/include
+export LDFLAGS=-L$PREFIX/lib
+export LIBS=-lpthread
+
+./configure --prefix=$PREFIX
+make
+mkdir -p $PREFIX/bin
+cp src/variant $PREFIX/bin
--- a/recipes/bxtools/meta.yaml
+++ b/recipes/bxtools/meta.yaml
@@ -6,7 +6,6 @@ package:
 
 build:
   number: 0
-  skip: True
 
 extra:
   # Needs to do recursive Git clone to get sub modules
--- a/recipes/bxtools/meta.yaml
+++ b/recipes/bxtools/meta.yaml
@@ -26,6 +26,8 @@ requirements:
     - zlib {{ CONDA_ZLIB }}*
   run:
     - libgcc
+    - bzip2 {{ CONDA_BZIP2 }}*
+    - zlib {{ CONDA_ZLIB }}*
 
 test:
     commands:
--- a/recipes/bxtools/meta.yaml
+++ b/recipes/bxtools/meta.yaml
@@ -1,0 +1,39 @@
+{% set version="0.0" %}
+{% set revision="bdf2982" %}
+package:
+  name: bxtools 
+  version: {{ version }}
+
+build:
+  number: 0
+  skip: True
+
+source:
+  # Needs to do recursive Git clone to get sub modules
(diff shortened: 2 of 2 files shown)


brainstorm:
Can anybody @bioconda/review help me out with this one? It seems to compile fine, but it fails on the test?... tried locally for the usual suspects: return codes, binary location, pinning, etc...

cbrueffer:
Can you remove the "| grep "Usage:" from the test so the actual error shows up in the log?

cbrueffer:
Looks like the bxtools internal htslib is build with lzma support, but the necessary library isn't present. Adding a dependency on "xz {{ CONDA_XZ }}*" should work.

brainstorm:
Interesting, it works flawlessly on one of our clusters, but on the other the BZIP2 library cannot be found:

```
bash-4.1$ bxtools
bxtools: error while loading shared libraries: libbz2.so.1.0: cannot open shared object file: No such file or directory
bash-4.1$ lsb_release -a
LSB Version:    :base-4.0-amd64:base-4.0-noarch:core-4.0-amd64:core-4.0-noarch:graphics-4.0-amd64:graphics-4.0-noarch:printing-4.0-amd64:printing-4.0-noarch
Distributor ID: CentOS
Description:    CentOS release 6.9 (Final)
Release:        6.9
Codename:       Final
bash-4.1$ ldd /g/data/gx8/tmp/miniconda_roman/bin/bxtools
        linux-vdso.so.1 =>  (0x00007ffd4cf28000)
        librt.so.1 => /lib64/librt.so.1 (0x00007f7c981e5000)
        libbz2.so.1.0 => not found
        liblzma.so.5 => /g/data/gx8/tmp/miniconda_roman/bin/../lib/liblzma.so.5 (0x00007f7c97fbf000)
        libz.so.1 => /g/data/gx8/tmp/miniconda_roman/bin/../lib/libz.so.1 (0x00007f7c97da8000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007f7c97b8a000)
        libstdc++.so.6 => /g/data/gx8/tmp/miniconda_roman/bin/../lib/libstdc++.so.6 (0x00007f7c977ce000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f7c9754a000)
        libgcc_s.so.1 => /g/data/gx8/tmp/miniconda_roman/bin/../lib/libgcc_s.so.1 (0x00007f7c97337000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f7c96fa3000)
        /lib64/ld-linux-x86-64.so.2 (0x00005562c351b000)
bash-4.1$ ls -alh /g/data/gx8/tmp/miniconda_roman/lib/libbz*
-rw-rw-r-- 2 rg3930 gx8 270K May 25  2016 /g/data/gx8/tmp/miniconda_roman/lib/libbz2.a
```

Any ideas why? On the other cluster the bzip2 lib seems to be installed:

```
        libbz2.so.1.0 => /home/brainstorm/.miniconda/bin/../lib/libbz2.so.1.0 (0x00001534b7046000)
```

Same conda version on both: `conda 4.4.11`.

acaprez:
The recipe has`bzip2` as a build and runtime dependency, so it shouldn't need to rely on the underlying system library.   It looks like on the Raijin machine the installed `bzip2` package only has the static version of the lib.  Can you run a `conda list` in each environment to get the full `bzip2` version strings and compare the two?

brainstorm:
Thanks @acaprez for the feedback! Here it is:

RHEL cluster (Spartan):

```
$ conda list | grep bzip
bzip2                     1.0.6                h9a117a8_4
```

Legacy Centos6 cluster (NCI/Raijin):

```
$ conda list | grep bzip
bzip2                     1.0.6                         3
```

Is there a way to somehow force/hint the package to not look/use the system's lib?

dpryan79:
On NCI/Raijin you have an outdated build of bzip2 via conda. If you update that then the same version will be installed everywhere and bxtools will presumably work.

acaprez:
You might also want to double-check your channel config/order.  The version of `bzip2` you have is from the defaults channel, which is missing the shared version of the lib.  The `bzip2` from `conda-forge` is the one you need.  It should prefer that one automatically if the channel order is correct.

brainstorm:
Working now after reinstalling bzip2 via conda(-forge), thanks @acaprez and @dpryan79!

PR #8232, opened 2018-03-12
update checkm-genome

pbordron:

### DIFF
--- a/recipes/checkm-genome/meta.yaml
+++ b/recipes/checkm-genome/meta.yaml
@@ -15,7 +15,7 @@ build:
 
 requirements:
   build:
-    - python <3
+    - python
     - numpy >=1.8.0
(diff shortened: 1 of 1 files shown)


recipes/checkm-genome/meta.yaml
```diff
@@ -11,19 +11,19 @@ source:
 
 build:
   number: 0
-  skip: True # [not py27]
+  skip: True # [not py27 or not osx]
 
 requirements:
   build:
-    - python
+    - python <3
```
druvus: You don't need the pinning here as you already added skipping above.

pbordron: Copy/Paste error solving issue like the 1.0.7 version that uses python3 when used with conda3

PR #8258, opened 2018-03-14
metaSNV recipe for v1.0.2

luispedro:

### DIFF
--- a/recipes/metasnv/build.sh
+++ b/recipes/metasnv/build.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+export CFLAGS="-I$PREFIX/include"
+export CPPFLAGS="-I$PREFIX/include"
+export LDFLAGS="-L$PREFIX/lib"
+tag=share/metasnv-${PKG_VERSION}/
+odir=$PREFIX/$tag
+
+make
+mkdir -p $odir
+cp -pr src $odir
+cp -pr metaSNV.py $odir
+cp -pr metaSNV_post.py $odir
+
+cat >$PREFIX/bin/metaSNV_post.py <<EOF
+#!/usr/bin/env bash
+
+BINDIR="\$(cd "\$(dirname "\${0}")" && pwd -P)"
+
+exec python \$BINDIR/../$tag/metaSNV_post.py "\$@"
+EOF
+
+cat >$PREFIX/bin/metaSNV.py <<EOF
+#!/usr/bin/env bash
+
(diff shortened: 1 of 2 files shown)


luispedro:
MetaSNV is at http://metasnv.embl.de/

luispedro:
@bioconda/core: This is a new package. Several users have asked for an easier way to install metaSNV, so it should be used by many the moment it's merged.

recipes/metasnv/build.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+export CFLAGS="-I$PREFIX/include"
+export CPPFLAGS="-I$PREFIX/include"
+export LDFLAGS="-L$PREFIX/lib"
+tag=share/metasnv-1.0.1/
```
rvalieris: you can use variables like `$PKG_VERSION` here,
https://conda.io/docs/user-guide/tasks/build-packages/environment-variables.html

recipes/metasnv/build.sh
```diff
@@ -0,0 +1,31 @@
...
+odir=$PREFIX/$tag
+
+make
+mkdir -p $odir
+cp -pr src $odir
+cp -pr metaSNV.py $odir
+cp -pr metaSNV_post.py $odir
+
+cat >$PREFIX/bin/metaSNV_post.py <<EOF
```
rvalieris: why not put the original `metaSNV*.py` in `bin/` ?

recipes/metasnv/build.sh
luispedro: Thanks, I'll update that

recipes/metasnv/build.sh
luispedro: Because the code expects that everything will be located relative to it.

So, it's either this or patching the code (which I don't want to go into).

rvalieris: would a symbolic link from odir to bin work ?

luispedro: No. I tried that. The ARGV[0] path is still set to the file on bin and then the package fails to find its other files.

rvalieris: alright !

luispedro:
Thanks @rvalieris for your feedback (using the $PKG_VERSION variable improved the recipe).

I will merge unless there are any remaining concerns.

rvalieris:
looks good, go ahead.

PR #8187, opened 2018-03-07
Bump version to 5.4.4

jburel:
This PR updates an existing recipe.

Bump version to 5.4.4.

PR #8260, opened 2018-03-14
idr-py: add recipe from PyPI

joshmoore:
This library is used for accessing https://idr.openmicroscopy.org, will be used in concert with an environment file: https://github.com/IDR/idr-py/pull/11, and depends on `python-omero` which we [maintain](https://github.com/bioconda/bioconda-recipes/pull/8187).

 * Created via `conda skeleton pypi idr-py`
 * Dependency on python-omero added manually

----

### DIFF
--- a/recipes/idr-py/meta.yaml
+++ b/recipes/idr-py/meta.yaml
@@ -33,9 +33,9 @@ requirements:
     - seaborn
 
 test:
-  imports:
-    - idr
-    - idr.tests
+  commands:
+    - env MPLBACKEND=Agg python -c "import idr"
+    - env MPLBACKEND=Agg python -c "import idr.tests"
 
 about:
   home: https://github.com/IDR/idr-py
--- a/recipes/idr-py/meta.yaml
+++ b/recipes/idr-py/meta.yaml
@@ -1,0 +1,48 @@
+{% set name = "idr-py" %}
+{% set version = "0.2.0" %}
+{% set file_ext = "tar.gz" %}
+{% set hash_type = "sha256" %}
+{% set hash_value = "af50a517b2fa324a1030830d95803cda18ecf2f9381550f401651f8c25d2f322" %}
+
+package:
+  name: '{{ name|lower }}'
+  version: '{{ version }}'
+
+source:
+  fn: '{{ name }}-{{ version }}.{{ file_ext }}'
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.{{ file_ext }}
+  '{{ hash_type }}': '{{ hash_value }}'
+
+build:
+  number: 0
+  script: python setup.py install --single-version-externally-managed --record=record.txt
+
+requirements:
+  host:
+    - python
+    - setuptools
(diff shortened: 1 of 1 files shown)


recipes/idr-py/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+source:
+  fn: '{{ name }}-{{ version }}.{{ file_ext }}'
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.{{ file_ext }}
+  '{{ hash_type }}': '{{ hash_value }}'
+
+build:
+  number: 0
+  script: python setup.py install --single-version-externally-managed --record=record.txt
+  skip: true  # [py3k]
```
nsoranzo: [Guidelines](https://bioconda.github.io/guidelines.html#python) recommend:
```yaml
  skip: True  # [not py27]
```

recipes/idr-py/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+requirements:
+  host:
+    - python
+    - setuptools
+  run:
+    - python >=2.7,<3
```
nsoranzo: Just `- python` here should be fine.

recipes/idr-py/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.{{ file_ext }}
+  '{{ hash_type }}': '{{ hash_value }}'
+
+build:
+  number: 0
+  script: python setup.py install --single-version-externally-managed --record=record.txt
+  skip: True  # [not py27]
+
+requirements:
+  host:
```
nsoranzo: This should be `build:`, no?

recipes/idr-py/meta.yaml
joshmoore: Thanks, @nsoranzo. Unfortunately seeing the same error with this change:

```
/opt/conda/lib/python3.6/site-packages/idr/tests/test_attributes.py
```

(I had tried various variants of this and force pushed them away)

recipes/idr-py/meta.yaml
joshmoore: Interesting. This seems to have come from the initial code generation. I'll certainly give `build:` a try.

nsoranzo: It builds successfully now, at least. Some test fails, but I think you can work from there.

joshmoore:
@nsoranzo : `s/host/build/` did the trick. With a subsequent matplotlib issue handled, CI is now green. Thanks for the help.

@bioconda/core : any comments on a new recipe?

daler:
Looks good; I would say the only issue is a messy description field. We don't currently render this anywhere, so it's really just a style choice. Up to you -- merge when ready, and thanks!

joshmoore:
We should have fairly regular updates; I'll incrementally improve the description (and perhaps open a PR for `conda skeleton...`) Thanks all.

PR #8273, opened 2018-03-16
Added recipe for ngs-bits

imgag:

### DIFF
--- a/recipes/ngs-bits/build.sh
+++ b/recipes/ngs-bits/build.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 #build
+make build_3rdparty
 make build_tools_release
 
 #deploy
--- a/recipes/ngs-bits/build.sh
+++ b/recipes/ngs-bits/build.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 
 #build
-make build_3rdparty
 make build_tools_release
 
 #deploy
--- a/recipes/ngs-bits/build.sh
+++ b/recipes/ngs-bits/build.sh
@@ -1,10 +1,20 @@
 #!/bin/bash
 
+#link include and lib folders to allow using htslib
+ln -s $PREFIX/include htslib/include
+ln -s $PREFIX/lib htslib/lib
+
 #build
-CPLUS_INCLUDE_PATH=$PREFIX/include
 make build_tools_release
 
-#deploy
-mkdir -p $PREFIX/bin
+#remove test files from bin folder
 rm -rf bin/out bin/cpp*-TEST bin/tools-TEST
+
+#deploy (lib)
+mkdir -p $PREFIX/lib
+cp -r bin/libcpp* $PREFIX/lib/
+rm -rf bin/libcpp*
+
+#deploy (bin)
+mkdir -p $PREFIX/bin
 cp -r bin/* $PREFIX/bin/
--- a/recipes/ngs-bits/build.sh
+++ b/recipes/ngs-bits/build.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+#build
+make build_3rdparty
+make build_tools_release
+
+#deploy
+mkdir -p $PREFIX/bin
+rm -rf bin/out bin/cpp*-TEST bin/tools-TEST
+cp -r bin/* $PREFIX/bin/
--- a/recipes/ngs-bits/build.sh
+++ b/recipes/ngs-bits/build.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 #build
-make build_3rdparty
+CPLUS_INCLUDE_PATH=$PREFIX/include
 make build_tools_release
 
 #deploy
--- a/recipes/ngs-bits/meta.yaml
+++ b/recipes/ngs-bits/meta.yaml
@@ -13,13 +13,13 @@ requirements:
   build:
     - python
     - zlib {{ CONDA_ZLIB }}*
-    - gcc
+    - libgcc 
     - git
     - cmake
     - qt >=5.5.0
   run:
     - zlib {{ CONDA_ZLIB }}*
-    - gcc
+    - libgcc
     - python
     - matplotlib
     - qt >=5.5.0
--- a/recipes/ngs-bits/meta.yaml
+++ b/recipes/ngs-bits/meta.yaml
@@ -18,7 +18,7 @@ requirements:
     - llvm  # [osx]
     - git
     - qt >=5.5.0
-    - htslib
+    - htslib {{ CONDA_HTSLIB }}*
   run:
     - bzip2 {{ CONDA_BZIP2 }}*
     - xz {{ CONDA_XZ }}*
@@ -27,7 +27,7 @@ requirements:
     - python
     - matplotlib
     - qt >=5.5.0
-    - htslib
+    - htslib {{ CONDA_HTSLIB }}*
 
 about:
   home: https://github.com/imgag/ngs-bits
--- a/recipes/ngs-bits/meta.yaml
+++ b/recipes/ngs-bits/meta.yaml
@@ -13,13 +13,14 @@ requirements:
   build:
     - python
     - zlib {{ CONDA_ZLIB }}*
-    - libgcc 
+    - gcc   # [linux]
+    - llvm  # [osx]
     - git
     - cmake
     - qt >=5.5.0
   run:
     - zlib {{ CONDA_ZLIB }}*
-    - libgcc
+    - libgcc  # [linux]
     - python
     - matplotlib
     - qt >=5.5.0
--- a/recipes/ngs-bits/meta.yaml
+++ b/recipes/ngs-bits/meta.yaml
@@ -18,7 +18,7 @@ requirements:
     - llvm  # [osx]
     - git
     - qt >=5.5.0
-    - htslib {{CONDA_HTSLIB}*
+    - htslib {{CONDA_HTSLIB}}*
   run:
     - bzip2 {{ CONDA_BZIP2 }}*
     - xz {{ CONDA_XZ }}*
@@ -27,7 +27,7 @@ requirements:
     - python
     - matplotlib
     - qt >=5.5.0
-    - htslib {{CONDA_HTSLIB}*
+    - htslib {{CONDA_HTSLIB}}*
 
 about:
   home: https://github.com/imgag/ngs-bits
--- a/recipes/ngs-bits/meta.yaml
+++ b/recipes/ngs-bits/meta.yaml
@@ -6,7 +6,7 @@ build:
   number: 0
 
 source:
-  git_rev: 6bf497bed52461422a1e21ef2149795ed419f443
+  git_rev: cbc0399a15818d1e5a824b5d261d9a1bc3b243fa
   git_url: https://github.com/imgag/ngs-bits.git
 
 requirements:
--- a/recipes/ngs-bits/meta.yaml
+++ b/recipes/ngs-bits/meta.yaml
@@ -6,7 +6,7 @@ build:
   number: 0
 
 source:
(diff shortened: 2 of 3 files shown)


apeltzer:
I just found this here: https://github.com/ContinuumIO/anaconda-issues/issues/1392 Maybe there are some hints what might help ....

imgag:
@apeltzer found a similar issue, but I'm not really sure the problem is the same.

@bgruening @johanneskoester do you have any suggestion how to fix the issue (it works when I run `conda-build` on the command line):
> ERROR conda.core.link:_execute_actions(337): An error occurred while installing package 'defaults::gcc-4.8.5-7'.

Best,
  Marc

imgag:
@bgruening @johanneskoester do you have any suggestion how to fix the issue?
It works when I run `conda-build` on the command line, so it's difficult to debug for me...

The error is (see ci/circleci: test-linux):
> ERROR conda.core.link:_execute_actions(337): An error occurred while installing package 'defaults::gcc-4.8.5-7'.

Thanks, Marc

daler:
@imgag you could try removing `gcc` from run dependencies and adding `libgcc` instead. @bioconda/lib any other suggestions?

imgag:
@daler @dpryan79 Thanks for your comments. Now the linux tests pass :)

There is still a problem with building bamtools for macos, but I will figure that out...

dpryan79:
I think your OSX issues are related to the build script. On OSX you might be getting .dylib files or something like that rather than .so files.

imgag:
@dpryan79 Thanks for the hint. I'm currently replacing bamtools by htslib. I'll check the OSX build once that is committed.

marc-sturm:
@dpryan79 @apeltzer @daler Does anyone of you have a clue how to fix the macOS linking problem? I have no Mac and no clue about linking etc on MacOS...

Here is what I do:
1) compile htslib library
2) compile ngs-bits libraries and executables 
3) copy all libraries and executablies to the `$PREFIX/bin` folder
4) execute `$PREFIX/bin/FastqFormat --help`

This works for perfectly on Linux.

However, this is what I get on macOS:
> FastqFormat --help
> dyld: Library not loaded: /Users/distiller/project/miniconda/conda-bld/ngs-bits_1522311560328/work/htslib/lib/libhts.2.dylib
>   Referenced from: /Users/distiller/project/miniconda/conda-bld/ngs-bits_1522311560328/_t_env/bin/FastqFormat
>   Reason: image not found

The ngs-bits libraries are correctly linked.
However, the path of the htslib library seems to be hard-coded in the executables somehow.

epruesse:
You should be able to enter the macos build on CircleCI by clicking on "Rerun job with SSH" in the upper right corner. That way you can actually poke around at the macos side without access to a mac yourself.

recipes/ngs-bits/meta.yaml
```diff
@@ -0,0 +1,38 @@
+package:
+  name: ngs-bits
+  version: 2018_03
+
+build:
+  number: 0
+
+source:
+  git_rev: d6ade58f2fd4b663e216bb7809afb856eeebd773
+  git_url: https://github.com/imgag/ngs-bits.git
```
epruesse: Upstream does have a 2018_03 release. Any reason why you need to check out a git revision directly? (Either explain git necessity with comment or switch to release tar ball please)

recipes/ngs-bits/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+#build
+make build_3rdparty
```
epruesse: I don't think this should be necessary. As far as I can tell, it only builds htslib. And that may well be the cause of your macos build issues. Use the htslib package please if at all possible. Or else explain with comment why that didn't work. Thanks!

recipes/ngs-bits/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+requirements:
+  build:
+    - bzip2 {{ CONDA_BZIP2 }}*
+    - xz {{ CONDA_XZ }}*
+    - zlib {{ CONDA_ZLIB }}*
+    - gcc   # [linux]
+    - llvm  # [osx]
+    - git
+    - qt >=5.5.0
```
epruesse:
```
 - htslib {{CONDA_HTSLIB}*
```

recipes/ngs-bits/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  run:
+    - bzip2 {{ CONDA_BZIP2 }}*
+    - xz {{ CONDA_XZ }}*
+    - zlib {{ CONDA_ZLIB }}*
+    - libgcc  # [linux]
+    - python
+    - matplotlib
+    - qt >=5.5.0
```
epruesse:
```
 - htslib {{CONDA_HTSLIB}*
```

marc-sturm:
Hi Elmar, thanls for looking onto this. our project contains git
submodules, which are not part of the release tarball. that is why i use
the git tag to download it.

Am 02.04.2018 19:21 schrieb "Elmar Pruesse" <notifications@github.com>:

*@epruesse* requested changes on this pull request.
------------------------------

In recipes/ngs-bits/build.sh
<https://github.com/bioconda/bioconda-recipes/pull/8273#discussion_r178594994>
:

> @@ -0,0 +1,10 @@
+#!/bin/bash
+
+#build
+make build_3rdparty

I don't think this should be necessary. As far as I can tell, it only
builds htslib. And that may well be the cause of your macos build issues.
Use the htslib package please if at all possible. Or else explain with
comment why that didn't work. Thanks!

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/bioconda/bioconda-recipes/pull/8273#pullrequestreview-108675865>,
or mute the thread

epruesse:
Stupid question - if the contents of the submodules is required, why are they not in the release? (You can upload those manually if github doesn't produce the right result)

epruesse:
FYI: When rerunning on CircleCI, use `. /tmp/.bash*` to get the conda path set right.

epruesse:
You need 
```
CPLUS_INCLUDE_PATH=$PREFIX/include
```
in build.sh so that the htslib headers are found.

epruesse:
I've tried building this, and it basically builds fine. There is just one issue: the libraries `libcppCORE` etc. are installed into the bin folder, the linker therefore expects them to be right next to the binary. Since this is built with `qmake` I can offer little help on how to fix this (or rather, it'd take me more time than I can offer). I'm sure that Qt has the necessary features to make this work platform independently.

I'd also recommend moving the libraries, both on linux and on macos, into `$PREFIX/lib/ngs-bits/` rather than `$PREFIX/lib`. The names are just too generic for a ngs-bits package. Since you are obviously familiar with upstream ... maybe those names could be altered to make namespace collisions a little less likely?

marc-sturm:
@epruesse hi Elmar, libraries are now correctly installed in the lib folder and executables in the bin folder.
I can try and make the library names more specific in the next release, but they are actually used in several of my projects, which is why I need to think about how to do it right first. Can you merge the recipe as soon as the test of version 2018_04 passes on macOS? Thanks, Marc

recipes/ngs-bits/build.sh
```diff
@@ -1,10 +1,20 @@
...
 
-#deploy
-mkdir -p $PREFIX/bin
+#remove test files from bin folder
 rm -rf bin/out bin/cpp*-TEST bin/tools-TEST
+
+#deploy (lib)
+mkdir -p $PREFIX/lib
+cp -r bin/libcpp* $PREFIX/lib/
+rm -rf bin/libcpp*
```
epruesse: ‘mv’?

epruesse:
Since you are using those libraries from multiple packages, the by far best way would definitely be to put them into a separate package and have ngs-bits depend on that package. That would solve the submodules cludge as well. You can submit multiple recipes in one PR, and they will be built in order  of dependency and available to each other. 

If you are in a hurry and need to think about turning your shared code into full-on re-usable libraries, we can leave this as it is. 

I'd appreciate if you could upload a tgz for 2018_03 to your GH release and use the tgz instead of checking out a git tag. Git tags just aren't very stable - they are just name-hash pairs, pointers to commits that are easily, even accidentally, changed: `git tag -d 2018_03; git tag 2018_03; git push --tags` and we will never know which commit was originally tagged.

marc-sturm:
@epruesse 

I would like to leave the libraries as they are for now. The interfaces of the libraries are not completely stable and they are used to share code between my projects and different executables only. I don't think they qualify for an own recipe.

Using sub-modules in Git makes the auto-generated tarballs completely useless. I don't know why they don't check out recursively... I can create a tgz archive of the release tomorrow, but the release process requires much more manual work then: (1) create the tarball, (2) upload it, (3) remove the auto-created tarball through an API (not possibe through the GUI afaik). However if it is a requirement for the merge, I will create it. Does the tarball contain the project files only, or does it include the project folder as well? I guess without the folder, right?

epruesse:
> Using sub-modules in Git makes the auto-generated tarballs completely useless. I don't know why they don't check out recursively... 

Probably because they use `git archive` to make the zip file, and that doesn't support submodules.

> I can create a tgz archive of the release tomorrow, but the release process requires much more manual work then: (1) create the tarball, 

Manual work in distribution? Nah, that's what we have CI for. Not using Travis myself, but something like this should work:

Add to your .travis.yml
```
before_deploy:
   git stash --all
   git submodule foreach 'git stash --all'
   REL=ngs-bits_$TRAVIS_TAG
   mkdir $REL
   cp -r * $REL
   tar -cjvf $REL.tar.bz2 $REL
```

> (2) upload it, 

Run `travis setup releases` and edit travis.yml:

```
deploy:
      provider: releases
      api_key: "GITHUB OAUTH TOKEN"
      file: ngs-bits*.tar.bz2
      skip_cleanup: true
```

> (3) remove the auto-created tarball through an API (not possibe through the GUI afaik). 

If you think it's necessary, sure. I'd probably just skip that.

You might want to consider using something like `bumpversion` to streamline this even further. For my projects, `bumpversion minor; git push --tags` is all I need to do to trigger a release with bumped minor version. Bumpversion increases the version number in all places, creates a commit from that and tags it with the new version name. The git push then triggers the CI to build, test and deploy the release.

> However if it is a requirement for the merge, I will create it. 

I'm sorry if I'm being a nit picker here - I've been on the receiving side of this as well, so I do understand the frustration. Reproducibility is part of the Bioconda mission, hence the desire to have archived tar balls. There are e.g. >1k recipes using cargo-port / depot.galaxyproject.org to stash a tarball for that reason. It's what I would have suggested if you weren't adding your own project.

> Does the tarball contain the project files only, or does it include the project folder as well? I guess without the folder, right?

Most commonly, the source distribution archive contains one folder named according to project name and version. Not every project adheres to that, though, and `conda-build` can deal with either.

marc-sturm:
Hi Elmar,

no worries, I'm not annoyed. I know how fast a project with many contributors can become unmaintainable when there are too few rules :)

I wrote a script that creates the tarball and adapted the recipe to use the tarball.
Thanks for the TravisCI hints. I will try and create and upload the tarball using TravisCI later.

How about merging now :)

Best,
  Marc

epruesse:
Thank you! :)

liqg:
Here is a solution.
Install  `libgcc`  and `gcc` into a new env, and copy the new env dir to the base dir of conda.
Then it is ok when I reinstall `gcc` of any version .
In my computer, the problem came after a conda rollback when an installation failed, so I guessed it should be that the rollback remove some files, because I found my original conda gcc also dispeared.

For example:
``
#create a new env
conda create -n gcctest
source activate gcctest
conda install libgcc gcc # it is ok for the new env
source deactivate gcctest
cp -r ~/anaconda3/envs/gcctest/*  ~/anaconda3/
conda install libgcc gcc
``

PR #8290, opened 2018-03-17
Bump clever-toolkit version to 2.2 and add matplotlib dependency

tobiasmarschall:


johanneskoester:
This had no effect because the recipe was blacklisted. See #8311 for a fix.

PR #8302, opened 2018-03-19
new tool rcorrector

bimbam23:

### DIFF
--- a/recipes/rcorrector/build.sh
+++ b/recipes/rcorrector/build.sh
@@ -2,7 +2,9 @@
 
 export C_INCLUDE_PATH=$PREFIX/include
 export OBJC_INCLUDE_PATH=$PREFIX/include
-export CPLUS_INCLUDE_PATH=$PREFIX/include
+export CPLUS_INCLUDE_PATH=$PREFIX/include 
+LD_LIBRARY_PATH=$PATH/lib
+
 
 make
 mkdir -p $PREFIX/bin
--- a/recipes/rcorrector/build.sh
+++ b/recipes/rcorrector/build.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
 
 export C_INCLUDE_PATH=$PREFIX/include
+export OBJC_INCLUDE_PATH=$PREFIX/include
+export CPLUS_INCLUDE_PATH=$PREFIX/include
+
 make
 mkdir -p $PREFIX/bin
 cp -R rcorrector run_rcorrector.pl $PREFIX/bin
--- a/recipes/rcorrector/build.sh
+++ b/recipes/rcorrector/build.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 
+export C_INCLUDE_PATH=$PREFIX/include
 make
 mkdir -p $PREFIX/bin
-cp -R rcorrector run_rcorrector.pl jellyfish/ $PREFIX/bin
+cp -R rcorrector run_rcorrector.pl $PREFIX/bin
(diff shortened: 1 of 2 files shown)


bimbam23:
is someone able to help me with the error message? I thought I fix it by adding zlib as dependency.

daler:
Hi @bimbam23 -  

No easy answer that I can see here, but a couple changes that may possibly help:

1. It looks like this tool is wget-ing the jellyfish source and compiling it (https://github.com/mourisl/Rcorrector/blob/master/Makefile#L13), but only if it doesn't exist. Adding a `- jellyfish >2` to the build and run deps should skip this step, and if the zlib error is happening on the jellyfish side, this may fix your error.

2. If you do end up needing zlib, it should be pinned (https://bioconda.github.io/linting.html#not-pinned). Pretty sure this should have been picked up by the linter but I'm not sure why and will look into that.

bimbam23:
thanks for your help... It looks like there might be more dependencies... but zlib is still not working.

daler:
Not sure why zlib isn't being found . . . @bioconda/lib any suggestions?

(By the way, you'll likely need jellyfish as a run dependency as well, once the zlib issue is resolved)

dpryan79:
Add `export C_INCLUDE_PATH=$PREFIX/include` to the Makefile before `make`.

dpryan79:
You'll want to remove the `{{CONDA_JELLYFISH}}` bit, that's not a predefined value for the template. Just use `>=2` or whatever is appropriate.

bimbam23:
thanks Devon,
pinning everything does not help :-)

bimbam23:
by the way, is the order important for the build and run?

dpryan79:
Order doesn't seem to matter in any way.

dpryan79:
Hmm, I guess my `export` trick failed.

dpryan79:
You might also try setting CPLUS_INCLUDE_PATH or OBJC_INCLUDE_PATH. The one I listed is explicitly for C programs.

bimbam23:
should I add all of them?

dpryan79:
To see if any work, sure.

dpryan79:
Looks like you might need to set `LD_LIBRARY_PATH=$PATH/lib` too.

recipes/rcorrector/build.sh
```diff
@@ -2,7 +2,9 @@
 
 export C_INCLUDE_PATH=$PREFIX/include
 export OBJC_INCLUDE_PATH=$PREFIX/include
-export CPLUS_INCLUDE_PATH=$PREFIX/include
+export CPLUS_INCLUDE_PATH=$PREFIX/include 
+LD_LIBRARY_PATH=$PATH/lib
```
dpryan79: You forgot `export`. You might want `export LD_LIBRARY_PATH=$PATH/lib:$LD_LIBRARY_PATH`. Thankfully conda will end up making a bunch of relative links out of things.

bimbam23: see I know what I'm doing ;-)

bimbam23:
now it looks good.

epruesse:
Is there a reason you can't compile with LLVM on macos?

```
build:
  - gcc  # [lnux]
  - llvm  # [macos]
```

bimbam23:
I don't know "llvm". Should I change to llvm to test it?

bimbam23:
is there anything I can add/improve here?

jerowe:
Looks good!

PR #8311, opened 2018-03-19
Remove clever-toolkit from blacklist.

johanneskoester:

clever-toolkit, which was updated yesterday by @tobiasmarschall, silently passed all tests because it was blacklisted. We should definitely add a lint that fails if a blacklisted recipe is updated, see bioconda/bioconda-utils#280.


johanneskoester:
Ok, gcc is fine now. However, it cannot find libz, although I set the CMAKE_PREFIX_PATH in the build.sh. Are you overwriting the search path somewhere?

tobiasmarschall:
Just had a look at the [CMakeLists.txt](https://bitbucket.org/tobiasmarschall/clever-toolkit/src/d7ac15294e882f43b91983d1f1ae688e6ea439f8/CMakeLists.txt?at=master&fileviewer=file-view-default), which looks ok as far as I can tell. These lines are related to paths:

```
SET(CMAKE_SKIP_BUILD_RPATH  FALSE)
SET(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
SET(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib")
```

Not sure how CMAKE_PREFIX_PATH and CMAKE_INSTALL_PREFIX interact.

tobiasmarschall:
Ok, one thing to try might be to add ${ZLIB_LIBRARY} to target_link_libraries in [3rd-party/bamtools/src/bamtools/api/CMakeLists.txt](https://bitbucket.org/tobiasmarschall/clever-toolkit/src/d7ac15294e882f43b91983d1f1ae688e6ea439f8/3rd-party/bamtools/src/bamtools/api/CMakeLists.txt?at=master&fileviewer=file-view-default#CMakeLists.txt-51). What's the easiest way to test this without making a new release?

johanneskoester:
You can add a patch, and refer to it in the meta.yaml, see [here](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#patches). I have send you an invitation to the bioconda organization. This way you can contribute directly to my branch here.

tobiasmarschall:
Ok, that didn't help. I hope I will have time to take a more detailed look soon, but the next days are pretty dense.

johanneskoester:
@bioconda/lib can you please help us here? `ld` cannot find zlib, although cmake looks like it is configured correctly...

dpryan79:
I hate cmake. Anyway, switching to the conda-forge gcc is sufficient to deal with the zlib linking issue. However that leads to the `auto_ptr` issue.

johanneskoester:
You nailed it @dpryan79 :-(. Of course there are the new conda build 3 compilers, see [here](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/octopus/meta.yaml#L24). Their usage will be much easier with conda-build 3, and more verbose for the moment, but we could use them as a workaround if there is no easy fix by configuring cmake or fixing the build scripts in clever-toolkit.

tobiasmarschall:
Thanks. I see. I should fix the auto_ptr issue anyway. Will try to do that later this week.

tobiasmarschall:
Hi @dpryan79, hi @johanneskoester,

thanks for your help so far. I released a new clever-toolkit version (2.2.1) that replaces `auto_ptr` by `unique_ptr`. I successfully built it on three different systems with GCC versions 4.9.2, 5.4.0, and 6.3.0.

Still the bioconda build failed and my guess is that now the GCC used is too old (4.8.2). I replaced "gcc" with "gcc_linux-64" in the recipe in the hopes that this would give me a newer GCC, but apparently I got that part wrong. Can you help?

Best,
Tobias

dpryan79:
Can you tell cmake to use `$PREFIX/bin/x86_64-conda_cos6-linux-gnu-cpp` as the C++ compiler? Likely that'd solve the problem.

tobiasmarschall:
@dpryan79 Unfortunately, that didn't work: 
```
-- Check for working CXX compiler: /opt/conda/conda-bld/clever-toolkit_1521813232339/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/bin/x86_64-conda_cos6-linux-gnu-cpp -- broken
CMake Error at /opt/conda/conda-bld/clever-toolkit_1521813232339/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/share/cmake-3.10/Modules/CMakeTestCXXCompiler.cmake:45 (message):
  The C++ compiler

    "/opt/conda/conda-bld/clever-toolkit_1521813232339/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/bin/x86_64-conda_cos6-linux-gnu-cpp"

  is not able to compile a simple test program.

  It fails with the following output:

    Change Dir: /opt/conda/conda-bld/clever-toolkit_1521813232339/work/clever-toolkit-v2.2.1/CMakeFiles/CMakeTmp
    
    Run Build Command:"/usr/bin/gmake" "cmTC_9a31e/fast"
    /usr/bin/gmake -f CMakeFiles/cmTC_9a31e.dir/build.make CMakeFiles/cmTC_9a31e.dir/build
    gmake[1]: Entering directory `/opt/conda/conda-bld/clever-toolkit_1521813232339/work/clever-toolkit-v2.2.1/CMakeFiles/CMakeTmp'
    Building CXX object CMakeFiles/cmTC_9a31e.dir/testCXXCompiler.cxx.o
    /opt/conda/conda-bld/clever-toolkit_1521813232339/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/bin/x86_64-conda_cos6-linux-gnu-cpp     -o CMakeFiles/cmTC_9a31e.dir/testCXXCompiler.cxx.o -c /opt/conda/conda-bld/clever-toolkit_1521813232339/work/clever-toolkit-v2.2.1/CMakeFiles/CMakeTmp/testCXXCompiler.cxx
    x86_64-conda_cos6-linux-gnu-cpp: fatal error: '-c' is not a valid option to the preprocessor
    compilation terminated.
    gmake[1]: *** [CMakeFiles/cmTC_9a31e.dir/testCXXCompiler.cxx.o] Error 1
    gmake[1]: Leaving directory `/opt/conda/conda-bld/clever-toolkit_1521813232339/work/clever-toolkit-v2.2.1/CMakeFiles/CMakeTmp'
    gmake: *** [cmTC_9a31e/fast] Error 2
```

Any hints what I might try?

johanneskoester:
According to the docs, I think the binary has to end on g++, not cpp, see [here](https://conda.io/docs/user-guide/tasks/build-packages/compiler-tools.html#using-the-compiler-packages). However, I am suprised that is does not work automatically, because it should actually run the activate script delivered together with the gcc package. Maybe this only works with the latest conda build 3. Let's see if my manual fix helps.

johanneskoester:
Ok, that works, but now we have the problem that boost from conda-forge has been compiled with an older version of gcc. They want to migrate to the new gcc packages, but it is currently not yet clear when this will happen. Hence, I think we cannot avoid giving the system gcc another chance.

tobiasmarschall:
With version 2.2.1, setting `-std=c++11` should not be necessary since I've added `set (CMAKE_CXX_STANDARD 11)` to the cmake file, so does this automatically (I think for gcc it uses `gnu11`).

Anything I can do to help resolve this?

johanneskoester:
I don't really know how to go on. @tobiasmarschall and @bioconda/lib here is a summary:

1. clever-toolkit uses unique_ptr, which is apparently a c++11 feature.
2. I therefore tried to use the build-in devtoolset-2 version of gcc from our docker container, because that is supposed to support c++11. However, it still cannot find unique_ptr.
3. Alternatively, I thought we could use the new conda-build 3 compilers (gcc 7). While this works to solve the unique_ptr issue, it now fails to see boost stuff (which has been compiled with gcc 4.8 on conda-forge), although I specified to use the old c++11 ABI as suggested [here](https://stackoverflow.com/questions/36309021/how-to-compile-boost-with-gcc-5-using-old-abi).

I don't know enough about C++ to solve this, sorry. Any help is greatly appreciated, because we urgently need this to come online for an important project.

tobiasmarschall:
I should add that version 2.2 does not use `unique_ptr`, but the deprecated `auto_ptr`. I actually created 2.2.1 in the hopes that abandoning `auto_ptr` would fix some trouble with newer compilers (which it does). Getting any of the two versions 2.2 or 2.2.1 is equally helpful.

@johanneskoester I'm not too deeply into this, but I recall having trouble using boost compiled on a different compiler before in a different project (but don't remember the details).

mbargull:
No CMake expert either, but I think you'd need to use
`cmake -DCMAKE_CXX_FLAGS=-D_GLIBCXX_USE_CXX11_ABI=0 ...`
or just `export CXX_FLAGS="${CXX_FLAGS} -D_GLIBCXX_USE_CXX11_ABI=0"`.

johanneskoester:
@mbargull you are a hero! Now it works!

recipes/clever-toolkit/build.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash
 
-cmake -DCMAKE_INSTALL_PREFIX=$PREFIX -DBOOST_ROOT=$PREFIX -DBoost_NO_SYSTEM_PATHS=ON .
-make
+# TODO remove requirement for ABI=0 once conda-forge has switched to new compiler packages.
```
mbargull: Just for reference, this needs
https://github.com/bioconda/bioconda-utils/issues/284
and 
https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/27
to be done.

tobiasmarschall:
Awesome, thanks! (In general thank you all for the effort you put in bioconda!)

PR #8323, opened 2018-03-20
Circularmapper

apeltzer:


apeltzer:
@bioconda/core

daler:
@apeltzer I'm on the fence on whether or not the realign portion should be split out into a separate recipe. On one hand it has its own sha256 and version, and generally we try to avoid that, but these look like two tightly coupled tools that you're maintaining in the same repo so it's probably OK. Are the releases of the tools always synchronized?

apeltzer:
Yes, they are built automatically using TravisCI based on a grade build script and are thus always in sync. Can’t figure out why building doesn’t work but I guess it’s just a minor thing I missed. I can split things but these actually belong quite tightly together.

daler:
@apeltzer I think that's fine then. Looks like your issue here is that the sha256 doesn't match? (I'm getting b91e0a58761f9854a89c526c45e58f716ccb5bcb456d2e9707fc127865c74f79).

apeltzer:
Hm, nope. Thats not it :/

apeltzer:
This is weird - I can build this with `conda-build recipe/circularmapper` locally and the recipe works too

apeltzer:
Since @daler already had a look and things build fine now - I guess its okay to merge this in?

daler:
Yep, thanks!

PR #8338, opened 2018-03-21
initial r-perfmeas recipe

visze:

### DIFF
--- a/recipes/r-perfmeas/meta.yaml
+++ b/recipes/r-perfmeas/meta.yaml
@@ -1,0 +1,58 @@
+{% set version = '1.2.1' %}
+
+{% set posix = 'm2-' if win else '' %}
+{% set native = 'm2w64-' if win else '' %}
+
+package:
+  name: r-perfmeas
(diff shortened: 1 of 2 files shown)


visze:
package not directly related to biological science. But it includes bioconductor packages which are only available on bioconda. So conda-forge will not be possible.

However this package is needed for another R package of biological science `r-hemdag`

daler:
Thanks @visze, agreed it should be here due to the bioconductor deps.

Can you please remove the Windows-related lines, as we don't build for Windows in bioconda?

visze:
sure! I forgot to remove them because I tried it first on conda-forge :-)

recipes/r-perfmeas/meta.yaml
```diff
@@ -0,0 +1,57 @@
+{% set version = '1.2.1' %}
+
+{% set posix = 'm2-' if win else '' %}
+{% set native = 'm2w64-' if win else '' %}
```
daler: These too please?

PR #8351, opened 2018-03-22
Seqmagick

dturaev:

So, conda build worked, but the circleci build failed, and I didn't figure out why. It seemed to me that somehow the environment wasn't correct: the $PYTHON variable was empty, and if given explicitly ("python"), the package dependencies were missing.

### DIFF
--- a/recipes/seqmagick/0.6.1/build.sh
+++ b/recipes/seqmagick/0.6.1/build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+$PYTHON setup.py install 
+
--- a/recipes/seqmagick/0.6.1/build.sh
+++ b/recipes/seqmagick/0.6.1/build.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-$PYTHON setup.py install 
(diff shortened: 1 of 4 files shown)


daler:
@dturaev thanks. It looks like this new recipe was created with conda-build-3, which we don't yet support (though we would love to, and are working toward doing so). As a result, one thing that's happening is that the `host:` key is not recognized by conda-build-2, resulting in no build dependencies being installed. It also makes sense that it would work locally running conda-build-3, but fail on circle-ci.

In general, you can update versions in-place (see https://bioconda.github.io/guidelines.html#versions). Doing that here would be the easiest way forward -- update the 0.6.1 recipe with the changes needed to build 0.7.0.

dturaev:
Got it, thanks for the explanation! https://github.com/bioconda/bioconda-recipes/pull/8369

dturaev:
Btw, when you say that you are working toward doing so, is there any way I can help? I might have some time for that...

daler:
@dturaev thanks, a high-level overview is at https://github.com/bioconda/bioconda-utils/issues/284, but there are lots of little details in the code that will likely need to be changed to match the behavior of cb3.

PR #8354, opened 2018-03-22
Updating nanosv 1.1.2

wdecoster:

[NanoSV](https://github.com/mroosmalen/nanosv) has been rewritten in Python. I used conda skeleton pypi to create the meta.yaml


wdecoster:
I've just swapped out the "host" key by "build" and this solved an error, but now I have:

```
  - configparser -> python 2.7*
  - python 3.5*

Typically if a conflict is with the Python or R
packages, the other package needs to be rebuilt
(e.g., a conflict with 'python 3.5*' and 'x' means
'x' isn't build for Python 3.5 and needs to be rebuilt.
```

daler:
`configparser` is built-in module in py3; I think if you add a `# [py27]` to the respective dependency lines it should be OK.

wdecoster:
Thanks for the suggestion, I'll try that. Makes me realize that `# [py27]` is not just a comment...

wdecoster:
Okay looks like I need to change the test because NanoSV, when imported, starts checking arguments:

```
run_test.py: error: the following arguments are required: bam
import: 'nanosv'
```

PR #8359, opened 2018-03-22
Added amused and brockman 

Carldeboer:
These are two k-mer based DNA-analysis programs (brockman depends on amused).
As this is my first bioconda pull, I would appreciate someone double checking it @bioconda/core.
I tried a local build and it appeared to work:
```
conda build -c bioconda  -c conda-forge brockman-pipeline
conda create -n BrockmanEnv2
conda install -n BrockmanEnv2 --use-local brockman-pipeline
```
Also tried circleci and it said that it was successful, although I was not sure that it was doing things right; my changes had already been committed to the repo at that time and it said:
```
12:59:31 BIOCONDA INFO No recipe modified according to git, exiting.
Success!
```

### DIFF
--- a/recipes/amused/build.sh
+++ b/recipes/amused/build.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+libloc="$PREFIX/lib/ruby/site_ruby"
+mkdir -p $libloc
+mkdir -p $PREFIX/bin
+cp Ruby-DNA-Tools/*.rb $libloc/. 
+cp AMUSED/NMERTREE.rb $libloc/. 
+cp AMUSED/AMUSED $PREFIX/bin/. 
+cp AMUSED/AMUSED-KS $PREFIX/bin/. 
+cp AMUSED/alignKMers $PREFIX/bin/. 
+cp AMUSED/shuffleCodons.rb $PREFIX/bin/. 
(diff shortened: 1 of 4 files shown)


daler:
@Carldeboer looks like it's not getting through parsing the meta.yaml. Can you try removing the empty `build:` key?

Carldeboer:
@daler I think it's the lack of hash for the source(s). It's dying when looking for 'sha256' in missing_hash.  Reading more, I see that the hash is [required](https://bioconda.github.io/linting.html), but still trying to figure out how to get a sha265 hash from a github repo...

daler:
Ah, I missed that. In general, we prefer tarballs from github releases rather than git urls. This is because the Galaxy team archives source tarballs for long-term reproducibility, but can't do that for git repos. Luckily, it looks like you recently cut a release ;) https://github.com/Carldeboer/AMUSED/archive/1.0.tar.gz.

But there's another issue. This recipe is using conda-build-3 syntax which we're not supporting yet. In this case, it means you can't have multiple sources in the same recipe, so you'll have to split these out into two separate recipes. That also means you'll need to create a github release for Ruby-DNA-Tools so you'll have a tarball (and hash) for that recipe.

daler:
To summarize: 

1. convert this single recipe into two recipes, one per tool
2. Use github release tarballs instead of git_urls

Thanks for your contribution!

PR #8364, opened 2018-03-22
Mikado

lucventurini:

### DIFF
--- a/recipes/mikado/meta.yaml
+++ b/recipes/mikado/meta.yaml
@@ -10,9 +10,7 @@ package:
 
 source:
   fn: '{{ version }}.{{ file_ext }}'
-  # url: https://files.pythonhosted.org/packages/f0/31/64f3630dfe4ed06b36226cf48d25370e8ee9ad875f0fb82e1dd2f129bdc2/Mikado-1.0.2.tar.gz
-  # url: file:////usr/users/ga002/venturil/workspace/mikado/dist/Mikado-1.0.2.tar.gz
-  url: https://github.com/lucventurini/mikado/archive/1.0.2.tar.gz
+  url: https://github.com/lucventurini/mikado/archive/{{version}}.tar.gz
   '{{ hash_type }}': '{{ hash_value }}'
 
 build:
@@ -20,8 +18,7 @@ build:
   entry_points:
     - mikado = Mikado:main
     - daijin = Mikado.daijin:main
-  # noarch: python
-  skip: True # [py27 or osx or py33]
+  skip: True # [py27 or osx or py33 or py36]
   script: python setup.py install --single-version-externally-managed --record=record.txt
 
 requirements:
--- a/recipes/mikado/meta.yaml
+++ b/recipes/mikado/meta.yaml
@@ -43,7 +43,7 @@ requirements:
     - scikit-learn >=0.17.0
     - scipy >=0.15.0
     - frozendict
-    - libmagic
+    - libmagic >=5.32
     - python-magic
     - drmaa
     - snakemake
@@ -67,7 +67,7 @@ requirements:
     - scikit-learn >=0.17.0
     - scipy >=0.15.0
     - frozendict
-    - libmagic
+    - libmagic >=5.32
     - python-magic
     - drmaa
     - snakemake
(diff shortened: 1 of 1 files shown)


lucventurini:
Hi @daler, unfortunately I am unable to configure Mikado correctly. This is due to some mis-configuration in the libmagic library in the Conda installation machines; when Mikado tries to interface with libmagic, it cannot find it, and crashes. I have added both "libmagic" and "python-magic" as dependencies (for building and running) so I do not presume that the problem stems from there.

Would you have any suggestion on how to configure the environment so that my tool is capable of finding the libmagic library?

daler:
I don't see anything obvious, sorry. Seems to be a problem with the `python-magic` package in conda-forge, as the following recreates the error:

```bash
conda create -n magic python-magic
source activate magic
python -c 'import magic; magic.Magic()'
```

lucventurini:
Hi @daler, finally Mikado builds correctly :-) Would you be able to officially accept it into the bioconda repository?
Also, I will have an updated version to load, version v1.1 - the version specified by this recipe is the one I am using for the academic paper, but development has since moved on.

Thanks again for your help!

recipes/mikado/meta.yaml
```diff
@@ -0,0 +1,122 @@
...
+{% set hash_type = "sha256" %}
+{% set hash_value = "2e668e098d465ef1bff837e152732b49bbb61f74cc5fbb938be9d72281bd710a" %}
+
+package:
+  name: '{{ name|lower }}'
+  version: '{{ version }}'
+
+source:
+  fn: '{{ version }}.{{ file_ext }}'
+  # url: https://files.pythonhosted.org/packages/f0/31/64f3630dfe4ed06b36226cf48d25370e8ee9ad875f0fb82e1dd2f129bdc2/Mikado-1.0.2.tar.gz
```
daler: rm these comment lines please

recipes/mikado/meta.yaml
```diff
@@ -0,0 +1,122 @@
...
+  # url: file:////usr/users/ga002/venturil/workspace/mikado/dist/Mikado-1.0.2.tar.gz
+  url: https://github.com/lucventurini/mikado/archive/1.0.2.tar.gz
```
daler: here you can use the jinja2 variable `{{ version }}` in place of `1.0.2` for easier future maintenance

recipes/mikado/meta.yaml
```diff
@@ -0,0 +1,122 @@
...
+  '{{ hash_type }}': '{{ hash_value }}'
+
+build:
+  number: 0
+  entry_points:
+    - mikado = Mikado:main
+    - daijin = Mikado.daijin:main
+  # noarch: python
```
daler: rm the noarch line please

daler:
@lucventurini Great! It looks like the solution was that somehow it was pulling in older versions of libmagic and python-magic? I'm not sure why that was happening but I'm glad it's working now.

I added some minor comments on the recipe, can you address them please? Then we can merge this, and then you can open another PR for v1.1.

lucventurini:
Hi @daler, changes made as requested. I have also disabled the build for python 3.6 as this particular version has a bug triggered specifically with Python 3.6 (but not python 3.5). The updated version 1.1 and following will be available for python 3.6 as well.

How should I upload a new version of Mikado after this first official release is accepted? Should I just open a new pull request with the updated recipe?

Thank you again for your help.

lucventurini:
And to answer your question, yes, the problem was an incompatibility between the conda-forge up-to-date libmagic and the bioconda python-magic library. If I may, might it be a good idea to remove those two libraries from bioconda and rely on the conda-forge versions?

daler:
Thanks!

Now what we do is wait for the master branch to build. Once that's done, it will upload the bioconda package and v1.0.2 will be available, and it will continue to be available in the conda channel even as the recipe here changes. So you can now submit a PR that makes the changes to this recipe to update it to v1.1 (version, hash, py36).

As for python-libmagic, thanks for the info. I'll add the `broken` label to the bioconda versions of `libmagic` and `python-magic`. In your v1.1 recipe, can you remove the version pinning to test if it then correctly installs the conda-forge versions?

PR #8369, opened 2018-03-22
update seqmagick to 0.7.0

dturaev:


dturaev:
I'll be happy to modify the recipe, if someone gives me a hint what's wrong with it. Otherwise, what's the best way to contribute to the development of bioconda, which I feel is a really useful and promising project? This is my first pull request, and I'm not sure where to go from here.

daler:
Thanks @dturaev. In general, you can overwrite the recipe with the newer version (see [docs here](https://bioconda.github.io/guidelines.html#versions)) and the old version will continue to be hosted on anaconda.org. That means you can simplify this PR to just change the existing meta.yaml and build.sh to reflect the new version info.

Since there's no good way to autodetect when packages need updating, contributions like this are exactly what we need and a great way to contribute to bioconda!

dturaev:
Got it, thanks! I fixed it, hope this works now. Ok, if I need a package that's not on bioconda, I'll provide a recipe for it. And maybe also try to figure out what needs to be done in bioconda-utils e.g. for supporting cb-3...

daler:
Thanks, looks good!

PR #8382, opened 2018-03-23
Libmagic

lucventurini:

@daler Fixed libmagic. The Mikado recipe should compile using this updated libmagic, instead of the current 5.25.0.


daler:
Nice, thanks!

This is a non-biology-related package which made it in during the early days of bioconda, but has since moved to conda-forge. Would you please make these fixes to https://github.com/conda-forge/libmagic-feedstock? I'm happy to give a hand with that if you need.

PR #8383, opened 2018-03-23
Du Novo: Update to 2.14.

NickSto:

This makes a relatively small update. The only real change to the recipe is adding one executable script and removing another.


NickSto:
@bgruening Thanks! That was quick!

PR #8397, opened 2018-03-25
Add maintainer and DOI to CATCH recipe

haydenm:
Thanks @druvus for adding a recipe for [CATCH](https://github.com/broadinstitute/catch). This had been on my to-do list. I'm the author of it –– I added myself as a maintainer and the DOI of our preprint.

### DIFF
--- a/recipes/catch/meta.yaml
+++ b/recipes/catch/meta.yaml
@@ -35,3 +35,8 @@ about:
   license: MIT
   license_file: LICENSE
   summary: A package for designing compact and comprehensive probe sets.
+
+extra:
+  maintainers:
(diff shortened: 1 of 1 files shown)


druvus:
@bgruening what is the current recommendations for adding metadata to recipes. Could we use abyss as an example for adding identifiers (#7940)?

I was following bioconda/bioconda-utils#252 but I am not sure about what you finally agreed on.

bgruening:
@druvus yes following like we did it in abyss should be the way to go. The maintainers attribute is not used currently.

bgruening:
Nice, thanks!

PR #8405, opened 2018-03-26
IntaRNA v2.2.0

martin-mann:

### DIFF
--- a/recipes/intarna/build.sh
+++ b/recipes/intarna/build.sh
@@ -33,7 +33,7 @@ fi
             --with-boost=$PREFIX \
             $CONFIGURE_MULTITHREADING \
             --disable-log-coloring \
-            --with-boost-libdir=$PREFIX/lib64 \
+            --with-boost-libdir=$PREFIX/lib \
             ${extra_config_options}
             
 make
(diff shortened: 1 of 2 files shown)


martin-mann:
@bgruening seems the (gnu) m4 macros for boost-lib-detection are not finding the boost lib for linking (see according configure error 
```
checking whether the Boost::System library is available... yes
configure: error: Could not find a version of the library!
```
which I only see if the boost devel is not correctly installed.. :/  was working for boost-1.61 !?!

martin-mann:
seems it might get solved by setting `configure --with-boost-libdir=` seed [ laanwj commented on 11 Nov 2013 ](https://github.com/bitcoin/bitcoin/issues/3219)

martin-mann:
what is the boost lib install path in bioconda at the end?

recipes/intarna/build.sh
```diff
@@ -33,7 +33,8 @@ fi
             --with-boost=$PREFIX \
             $CONFIGURE_MULTITHREADING \
             --disable-log-coloring \
-            ${extra_config_options} \
+            --with-boost-libdir=$PREFIX/lib64 \
```
bgruening: $PREFIX/lib should be the path, you can look this up by simply untar the boost tarball.

martin-mann:
@bgruening have seen in my prelast try.. ☺

bgruening:
:+1:

PR #8422, opened 2018-03-27
IgDiscover fails when run with matplotlib 2.2

marcelm:


marcelm:
@bioconda/core Could someone have a look at the error this PR is producing? I added a file `clusterplot.tab.gz` and believe I correctly added a `files:` section to the `tests:` section, but the file cannot be found.

I found this comment by @pvanheus in [`repeatmasker/meta.yaml`]([https://github.com/bioconda/bioconda-recipes/blob/master/recipes/repeatmasker/meta.yaml):
```
# pvanheus - disable file based tests for
# now - they don't work with Dockerized
# containers - 13/1/2017
#  files:
#    - rm_input1.fasta
#    - sample_repeats.fasta
#    - test_md5.txt
```
Does this still apply?

If yes, I can just remove the test I added; it’s not that important.

rvalieris:
the syntax is correct, however the mulled-build tests only support test-commands right now, so the command needs to be self contained. you could use a test-command to create a test file on-the-fly, but you can't bring a existing file inside mulled-build.

marcelm:
Thanks, I’ll remove the test then.

PR #8432, opened 2018-03-27
Add dependencies required to run metfrag in batch

sneumann:

A possible way to use MetFrag is to pass a ZIP file of queries and run MetFrag on all of them. 
This requires the dependencies GNU Parallel and ZIP, which are available as conda packages.


bgruening:
@sneumann do you only need `zip` or also `unzip`?

sneumann:
Ah, correct, I missed the unzip. Thanks for catching. I would have only realised 
when the Galaxy wrappers would've failed in the resulting container.  PR amended.

PR #8439, opened 2018-03-28
update sra

sachalau:


sachalau:
Hello @bioconda/team !

I want to update the fastq-dump version from NCBI but I see that this recipe is a bit unusual : it builds other recipes from the NCBI at the same time (in build.sh).

I don't want to mess up everything but shouldn't we build separate recipes for NGS_SDK and NCBI_VDB ? Does someone know why the recipe was built like that ? 

Thanks a lot for your help !

sachalau:
No I think it will be better to build them in another recipes right ? I don't know why it was built like this before.

bgruening:
In another recipe?

sachalau:
Yes, starting now  https://github.com/bioconda/bioconda-recipes/pull/8467

rvalieris:
hello, I made a commit fixing some stuff, it should be alright now.

sachalau:
@rvalieris thanks a lot for your help !

PR #8440, opened 2018-03-28
include sqlite support in augustus 3.3 build

twrightsman:

Need some feedback from @bioconda/core regarding my change to the build string. I recently updated augustus to 3.3 and built it with [CGP support](http://bioinf.uni-greifswald.de/augustus/binaries/README-cgp.txt) and this pull request also adds sqlite support as well. I figured this should be noted in the build string instead of specifying the boost library and bamtools version, but I'm not sure. Maybe this is a job for [conda features](https://conda.io/docs/user-guide/tasks/build-packages/features.html).

### DIFF
--- a/recipes/augustus/build.sh
+++ b/recipes/augustus/build.sh
@@ -22,7 +22,11 @@ mv common.mk common.mk.orig
 sed 's/# COMPGENEPRED = true/COMPGENPRED = true/' < common.mk.orig | sed 's/# SQLITE = true/SQLITE = true/' > common.mk
 
 ## Make the software
-
+if [ "$(uname)" == Linux ] ; then
+    # TODO: remove this when switching to newer compilers
+    export CC=gcc
+    export CXX=g++
+fi
 make CC="${CC}" CXX="${CXX}"
 
 ## Build Perl
--- a/recipes/augustus/build.sh
+++ b/recipes/augustus/build.sh
@@ -22,6 +22,8 @@ mv common.mk common.mk.orig
 sed 's/# COMPGENEPRED = true/COMPGENPRED = true/' < common.mk.orig | sed 's/# SQLITE = true/SQLITE = true/' > common.mk
 
 ## Make the software
+sed -i.bak 's/\<CC\>/CXX/g' auxprogs/homGeneMapping/src/Makefile
+sed -i.bak 's/\<CC\>/CXX/g' auxprogs/joingenes/Makefile
 if [ "$(uname)" == Linux ] ; then
     # TODO: remove this when switching to newer compilers
     export CC=gcc
--- a/recipes/augustus/build.sh
+++ b/recipes/augustus/build.sh
@@ -22,7 +22,8 @@ if [ "$(uname)" == Darwin ] ; then
   # SQLITE disabled due to compile issue, see: https://svn.boost.org/trac10/ticket/13501
   make CC="${CC}" CXX="${CXX}" COMPGENPRED=true
 else
-  make CC="${CC}" CXX="${CXX}" COMPGENPRED=true SQLITE=true
+  # TODO: use 'CC="${CC}" CXX="${CXX}"' when switching to newer compilers
+  make CC=gcc CXX=g++ COMPGENPRED=true SQLITE=true
 fi
 
 ## Build Perl
@@ -52,4 +53,3 @@ chmod a+x $PREFIX/etc/conda/activate.d/augustus-confdir.sh
 mkdir -p $PREFIX/etc/conda/deactivate.d/
 echo "unset AUGUSTUS_CONFIG_PATH" > $PREFIX/etc/conda/deactivate.d/augustus-confdir.sh
 chmod a+x $PREFIX/etc/conda/deactivate.d/augustus-confdir.sh
-
--- a/recipes/augustus/build.sh
+++ b/recipes/augustus/build.sh
@@ -16,10 +16,10 @@ mkdir -p $PREFIX/bin
 mkdir -p $PREFIX/scripts
 mkdir -p $PREFIX/config
 
-## enable augustus CGP mode
+## enable augustus CGP mode and sqlite database support
 
 mv common.mk common.mk.orig
-sed 's/# COMPGENEPRED = true/COMPGENPRED = true/' < common.mk.orig > common.mk
+sed 's/# COMPGENEPRED = true/COMPGENPRED = true/' < common.mk.orig | sed 's/# SQLITE = true/SQLITE = true/' > common.mk
 
 ## Make the software
 
--- a/recipes/augustus/build.sh
+++ b/recipes/augustus/build.sh
@@ -18,8 +18,8 @@ mkdir -p $PREFIX/config
 
 ## Make the software
 
-sed -i.bak 's/\<CC\>/CXX/g' auxprogs/homGeneMapping/src/Makefile
-sed -i.bak 's/\<CC\>/CXX/g' auxprogs/joingenes/Makefile
+sed -i.bak -e 's/^CC *=/CXX=/' -e 's/\$(CC)/$(CXX)/g' auxprogs/homGeneMapping/src/Makefile
+sed -i.bak -e 's/^CC *=/CXX=/' -e 's/\$(CC)/$(CXX)/g' auxprogs/joingenes/Makefile
 # TODO: don't set CC/CXX here when switching to newer compilers
 CC=gcc
 CXX=g++
--- a/recipes/augustus/build.sh
+++ b/recipes/augustus/build.sh
@@ -16,20 +16,14 @@ mkdir -p $PREFIX/bin
 mkdir -p $PREFIX/scripts
 mkdir -p $PREFIX/config
 
-## enable augustus CGP mode and sqlite database support
-
-mv common.mk common.mk.orig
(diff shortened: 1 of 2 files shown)


twrightsman:
```
In file included from geneMSA.cc:16:
In file included from ../include/genomicMSA.hh:15:
In file included from ../include/geneMSA.hh:21:
In file included from /Users/distiller/project/miniconda/conda-bld/augustus_1522236874781/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/include/boost/functional/hash.hpp:6:
In file included from /Users/distiller/project/miniconda/conda-bld/augustus_1522236874781/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/include/boost/functional/hash/hash.hpp:572:
/Users/distiller/project/miniconda/conda-bld/augustus_1522236874781/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/include/boost/functional/hash/extensions.hpp:262:20: error: no matching function for call to 'hash_value'
            return hash_value(val);
                   ^~~~~~~~~~
```

bgruening:
@twrightsman I think you can leave this out of the build string.

mbargull:
Looks like the macOS build failed because of some `clang`/`boost` incompatibility that surfaced because previously `#ifdef SQLITE`-excluded code is now compiled. You could try using another version of `boost` (see the other `augustus` recipe in the subfolder) or try the a newer Clang version by replacing `llvm` by the `clangxx_osx-64` etc., see https://github.com/bioconda/bioconda-recipes/blob/20993e33e8dd146b981c067fc119b84027072db3/recipes/stacks/2.0/meta.yaml for an example.
Note that using another `boost` version might be the easier solution to get this built, but this could pose more difficulties for users if they want to install other packages which are pinned to `boost {{ CONDA_BOOST }}.*` into the same environment they install `augustus` into.

twrightsman:
https://svn.boost.org/trac10/ticket/13501 Submitted an issue to Boost developers, will have to wait

mbargull:
Looks like the `Makefile`s use hardcoded values for `CXX` etc., see:
```
grep -E 'g(cc|\+\+)|C(C|XX|PP)?(FLAGS)?\s*:?:?=' augustus/**/Makefile                                                                                                                                                                                                
augustus/auxprogs/aln2wig/Makefile:CFLAGS := -Wall -Wno-sign-compare -ansi -pedantic -O2 -ggdb ${CFLAGS}
augustus/auxprogs/aln2wig/Makefile:	gcc $(CFLAGS) $(CPPFLAGS) $(CXXFLAGS) $(LDFLAGS) -o aln2wig aln2wig.o;
augustus/auxprogs/aln2wig/Makefile:	gcc $(CPPFLAGS) $(CXXFLAGS) $(CFLAGS) $(LDFLAGS) -c aln2wig.c
augustus/auxprogs/bam2hints/Makefile:LINK.cc = g++
augustus/auxprogs/bam2wig/Makefile:CFLAGS=-Wall -O2 $(INCLUDES)
augustus/auxprogs/bam2wig/Makefile:CC=gcc
augustus/auxprogs/checkTargetSortedness/Makefile:CFLAGS := -g -Wall -O2 ${CFLAGS}
augustus/auxprogs/checkTargetSortedness/Makefile:CC=gcc
augustus/auxprogs/compileSpliceCands/Makefile:	gcc $(LDFLAGS) -o compileSpliceCands compileSpliceCands.o list.o;
augustus/auxprogs/compileSpliceCands/Makefile:	gcc -Wall -pedantic -ansi $(CPPFLAGS) -c compileSpliceCands.c 
augustus/auxprogs/filterBam/src/Makefile:CFLAGS = -std=c++0x
augustus/auxprogs/homGeneMapping/src/Makefile:CC      = g++
augustus/auxprogs/homGeneMapping/src/Makefile:CFLAGS := -Wall -Wno-sign-compare -ansi -pedantic -std=c++0x -pthread -O2 ${CFLAGS} # -DDEBUG -g -ggdb -pg
augustus/auxprogs/joingenes/Makefile:CC=g++
augustus/auxprogs/joingenes/Makefile:CFLAGS=-Wall -std=gnu++0x
augustus/src/Makefile:CXX	= g++
augustus/src/Makefile:CXXFLAGS := -Wall -Wno-sign-compare -Wno-strict-overflow -ansi -pedantic -O3 ${CXXFLAGS} # -DDEBUG -g -ggdb -pg -DDEBUG_STATES -DDEBUG 
```
. I added a commit which should help with most of those files. If there is still something else that uses `g++`/`gcc` instead of `clang`/`clang++`, we might need to patch those `Makefile`s.
I assume https://svn.boost.org/trac10/ticket/13501 to be invalid then.

mbargull:
Last build failed because it would've needed `CC`/`CXX` to be set to `gcc`/`g++` -- instead of doing this in `build.sh`, let's see if it builds fine with the newer compiler instead first.

mbargull:
Spits out:
```
bin/../x86_64-conda_cos6-linux-gnu/sysroot/usr/lib/../lib/gcrt1.o: relocation R_X86_64_32S against symbol `__libc_csu_fini' can not be used when making a shared object; recompile with -fPIC
/opt/conda/conda-bld/augustus_1522329096463/_b_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/bin/../lib/gcc/x86_64-conda_cos6-linux-gnu/7.2.0/../../../../x86_64-conda_cos6-linux-gnu/bin/ld: final link failed: Nonrepresentable section on output
make[3]: *** [filterBam] Error 1
make[3]: Leaving directory `/opt/conda/conda-bld/augustus_1522329096463/work/augustus/auxprogs/filterBam/src'
make[2]: *** [all] Error 2
make[2]: Leaving directory `/opt/conda/conda-bld/augustus_1522329096463/work/augustus/auxprogs/filterBam'
make[1]: *** [all] Error 2
make[1]: Leaving directory `/opt/conda/conda-bld/augustus_1522329096463/work/augustus/auxprogs'
make: *** [all] Error 2
```
with the newer GCC package -- I don't have time to look into this further currently, so let's just use the old GCC instead again.

mbargull:
```
augustus/auxprogs/homGeneMapping/src/Makefile:CC      = g++
```
:unamused:

twrightsman:
It seems the error persists... Thank you for all of the troubleshooting help @mbargull. Unfortunately, C/C++ compilation is not my area of expertise.

mbargull:
Yeah, I withdraw that
> I assume https://svn.boost.org/trac10/ticket/13501 to be invalid then.

statement. Can you update the Boost issue with a link to the last failed build, stating that this uses `clang 4.0.1`?

So, what should we do with this PR then? Do you think it is reasonable to only enable that functionality on Linux? The benefit is, of course, that failing macOS builds don't hold back feature support on Linux. The downside is that macOS and Linux users get different software/feature sets which can always lead to different behavior/output (speaking in general, no idea in which way this applies to `augustus`).
If we make this Linux-only, please append
```yaml

extra:
  notes: some mentioning/explanation which features/dependencies are Linux-only
```
to `meta.yaml`.

mbargull:
Oh, and instead of
```bash
mv common.mk common.mk.orig
sed 's/# COMPGENEPRED = true/COMPGENPRED = true/' < common.mk.orig | sed 's/# SQLITE = true/SQLITE = true/' > common.mk
...
make CC="${CC}" CXX="${CXX}"
```
I think we can make this
```bash
make CC="${CC}" CXX="${CXX}" COMPGENPRED=true SQLITE=true
```
or, with the latter being Linux-only, something like
```bash
if [ "$(uname)" == Darwin ] ; then
    # SQLITE disabled due to compile issue, see: https://svn.boost.org/trac10/ticket/13501
    make CC="${CC}" CXX="${CXX}" COMPGENPRED=true
else
    make CC="${CC}" CXX="${CXX}" COMPGENPRED=true SQLITE=true
fi
```

twrightsman:
I think you've hit the nail on the head here. I'll add the extra note and turn off SQLITE compilation for macOS for now in the way you've suggested. If/when boost 1.64 gets updated with a patch, that can be rebuilt on conda and then we can try to re-build `augustus` on macOS with SQLITE.

mbargull:
@twrightsman, if you think this is fine now, feel free to merge. I for one don't really know Augustus and thus can't asses the implications of this change, so I leave that up to you to decide :wink:.

mbargull:
Regarding the misnamed `CC`/`CXX` variables in the `auxprogs/**/Makefile` and the build failures in https://circleci.com/gh/bioconda/bioconda-recipes/6736 for the newer GCC 7.2 packages, this is something we could relay to the Augustus developers.
Looks like there is a @MarioStanke on GitHub -- maybe this is the one from https://math-inf.uni-greifswald.de/institut/ueber-uns/mitarbeitende/stanke/ we want? (If not, sorry for the bother @MarioStanke.)

twrightsman:
I will send an email to the developers about this; I think we can merge now. Since I'll be using this soon, I can report and hopefully fix issues :wink:

mbargull:
Great, thanks!

PR #8443, opened 2018-03-28
Added biotools identifiers

MathiasHaudgaard:

As we have discussed in the last Biotools IS meeting we are now starting to add the biotools identifiers to all bioconda packages according to this mapping: https://github.com/BioContainers/biotools-bioconda-ids/blob/master/mapping.csv

We're a bit in doubt about the doi tag. What if a package has more papers related, should they be included as well? or should we just take the first doi and add it?

@bgruening @ypriverol @ArneKr

### DIFF
--- a/recipes/bioconductor-a4base/meta.yaml
+++ b/recipes/bioconductor-a4base/meta.yaml
@@ -51,3 +51,7 @@ about:
   home: 'http://bioconductor.org/packages/{{ bioc }}/bioc/html/{{ name }}.html'
   license: GPL-3
   summary: 'Automated Affymetrix Array Analysis'
+extra:
+  identifiers:
+    - biotools:a4base
+    - doi:10.1038/nmeth.3252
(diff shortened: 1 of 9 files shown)


ypriverol:
In principle we should add only the main paper related with this tool, but you are allow to put more.

bgruening:
Awesome!
I agree with @ypriverol, I think the primary paper should be mentioned, if really really needed a additional can be added, but this should be an exception I guess.

ArneKr:
We asked about this mainly because we were generating the annotations automatically (There are over 1000 packages that needs to be modified after all) and there is as far as we know no way to identify which of the papers that are linked to a tool on bio.tools are the most relevant. If only the main paper has to be added to the meta data we need to manually check each tool, which makes our tasks alot more difficult.

bgruening:
Try to add them, we will help with curation, or this can be done incrementally during the package updates.
However, I would trust bioconductor annotated DOIs for example more then the hand.cureated biotools one. So if possible I would consider the upstream annotation (e.g. citations() for bioconductor) higher ...

PR #8455, opened 2018-03-28
Use Extended Regex (OSX compat); otherwise “exec #” quietly does nothing

kastman:

Without this PR, abyss quietly does nothing because of the makefile wrapper. This PR uses extended regex in the wrapper to actually exec instead of `exec # "comments"`. 

Also adds tests to make sure output is shown.


recipes/abyss/1.5.2/create-wrapper.sh
```diff
@@ -14,6 +14,6 @@ chmod a-x "$2"
 
 cat >"$1" <<EOF
 #!/bin/bash
-exec $(sed 's|^#!/\(usr/\)\?bin/||;q' "$2") '$2' "\$@"
+exec $(sed -E 's|^#!(/\usr)?\/bin/||;q' "$2") '$2' "\$@"
```
mbargull: There is a typo `/\`, but we don't have to escape the `/` anyway since we use `|` as the delimiter.
Maybe we better make this
```bash
interpreter="$(sed -nE '1 s|^#! *(/usr)?/bin/||p;q' "$2")"
[ -n "${interpreter}" ]
cat >"$1" <<EOF
#!/bin/bash
exec ${interpreter} '$2' "\$@"
EOF
```
to catch that failure already at build time.

mbargull: Or even `exec '${PREFIX}/'${interpreter} '$2' "\$@"` for the `exec` part.

recipes/abyss/1.5.2/meta.yaml
```diff
@@ -33,9 +33,9 @@ requirements:
 
 test:
   commands:
-    - ABYSS-P --version
-    - ABYSS --version
-    - abyss-pe help
+    - ABYSS-P --version | grep "{{ version }}" > /dev/null
```
mbargull: Instead of `grep ... > /dev/null` you can also just do `grep -q ...`.

recipes/abyss/1.5.2/create-wrapper.sh
kastman: Is `${PREFIX}` necessary, given the path and `/usr/bin/env ${interpreter}` should be working correctly? I'm not sure which is better here; implicit or explicit? I do like defining the interpreter on its own line though.

kastman: Also, one _possible_ reason not to use `-E` extended is that I saw somewhere it's slower. Since we're literally only searching 1 line I'm sure it's not even measureable but would conceivably be a difference.

recipes/abyss/1.5.2/meta.yaml
kastman: Ahh, I was just pulling the style from other recipes, but `grep -q` does seem better.

epruesse:
Naive question - why not just add `- sed` to build requirements?

Honestly, I even think that having the linter check for `sed` and enforce the package is installed might be useful. The differences in `coreutils`, `sed` and `grep` between their GNU Linux versions and the ones that come with e.g. macos are often subtle and the workarounds are regularly lengthy (e.g. 10 lines of bash to emulate `readlink -f`), tedious and error prone.

kastman:
@epruesse I like `brew`'s solution, which is to allow explicitly installing of the GNU versions with "g" appended, e.g. `gsed`, `gawk`, etc., So if you're writing your own scripts you can get access to the functionality.

There's a bit too much surprise for my taste in overwriting `sed` inside a conda environment though; as a user if I'm expecting BSD `sed` and get gnu `sed` without requesting it by installing some bio-related package, that seems more insidious than the workaround (which in this case is simply using an extended regex with `-E`)?

epruesse:
@kastman The only way to be reasonably sure what kind of `sed` you get is to actually require the package. On macos, the system one is BSD, but unless you call `/usr/bin/sed` you might get `$CONDA_ENV/bin/sed` installed by another package into your env, or `$MINICONDA/bin/sed` if the user installed into the root env (which is always still in the PATH), or even `/usr/local/bin/sed` if the user used `brew install gnu-sed --default-names`. 

So you are left using the common denominator only (as you did here), which can be painful and is a challenge to get right (how many people writing Bioinf tools have never used anything but Linux and Windows?)

The beauty of having `conda` available to specify your environment is that you don't have to waste tons of time on that kind of compatibility code and testing. Instead of maintaining compatibility with a gazillion libraries and tools, you can just require and pin them and have the system provide what you need. It's exactly that freedom that convinced me to join the party...

All of that said.... in this case, `-E` does the job just as well :-)

recipes/abyss/1.5.2/create-wrapper.sh
mbargull: And we are calling that `sed -E` only once at *build time* :wink:.
There are pros and cons for using the full path or not. Generally, I think we should prefer absolute paths. Often times we can't due to those massive shebang restrictions -- esp. on Linux, see https://www.in-ulm.de/~mascheck/various/shebang/#results (shebangs were just not engineered for today's relocatable environments...). As is, we're not consistent in that regard in Bioconda overall, so it's not that important, I guess. Currently, either way should be fine, feel free to decide.

mbargull:
> Naive question - why not just add - sed to build requirements?

IMHO, not naive at all. I already often thought it would be convenient to always have the `sed` (maybe also `core-utils`, `grep`) packages (i.e., GNU versions) in the build environment (and only build, unless the built packages themselves expect GNU behavior). Especially `sed`: people very often make mistakes in regards to `sed -i` since the GNU and macOS version are completely incompatible for that parameter.

epruesse:
>Especially sed: people very often make mistakes in regards to sed -i since the GNU and macOS version are completely incompatible for that parameter.

`sed -ibak` works with both. You just have to accept that there will be a `.bak` file.

mbargull:
Right, I've been imprecise, I meant incompatible in regards to not creating that file, i.e., `sed -i` vs. `sed -i ''`. People sometimes write one or the other expecting it to work everywhere. I've even seen the attempt to combine this to `sed -i''` (which obviously failed since it's just `sed -i`).

kastman:
I suppose that’s true - I totally see it eg in `gcc`, so it makes sense in this case. I guess when it’s as simple as `-E` I’m for fewer requirements, but probably wouldn’t object to putting it in everything. Good point!

recipes/abyss/1.5.2/create-wrapper.sh
```diff
@@ -16,7 +16,7 @@ interpreter="$(sed -nE '1 s|^#! *(/usr)?/bin/||p;q' "$2")"
 [ -n "${interpreter}" ]
 cat >"$1" <<EOF
 #!/bin/bash
-exec '${PREFIX}/'${interpreter} '$2' "\$@"
```
mbargull: Argh, my bad -- that needed to be `'${PREFIX}'/bin/${interpreter}`

mbargull:
Something went wrong with the merge. Can you try merging or rebasing again?

kastman:
Merge worked. Odd!

PR #8460, opened 2018-03-28
Bump transdecoder to version 5.1.0

nathanweeks:

First attempt. `circleci build` seems to fail without explanation in the test phase.

### DIFF
--- a/recipes/transdecoder/build.sh
+++ b/recipes/transdecoder/build.sh
@@ -2,7 +2,9 @@
 
 # install in opt/
 mkdir -p $PREFIX/opt/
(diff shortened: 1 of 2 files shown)


nathanweeks:
Turns out that perl-db-file is required, as while DB_File is in Perl core, it isn't installed if Berkeley DB isn't available when Perl is built.

nathanweeks:
Pinging to check the status on this. Would a bioconda maintainer be able to review this one? This transdecoder release is a prerequisite before I can create & submit a new bioconda recipe for PASA 2.3.2. Thanks!

PR #8461, opened 2018-03-28
update perl-dbi to 1.641

danielee0707:


danielee0707:
The perl environment in the test report shows CONDA_PERL=5.22.0. But I think it depends on 5.26. Is that the problem? This is my first time pulling request. Do you mind help me with this? @mbargull  Thanks.

recipes/perl-dbi/build.sh
```diff
@@ -17,3 +17,9 @@ else
     echo 'Unable to find Build.PL or Makefile.PL. You need to modify build.sh.'
     exit 1
 fi
+
```
bgruening: can you remove these comments from both files please

danielee0707: is this ok?

PR #8480, opened 2018-03-29
NewRecipe: AMPtk - amplicon toolkit

nextgenusfs:

This is my first PR for Bioconda and is for my amplicon toolkit (AMPtk) package.  All tests passed locally with `circleci build`.  I would love a review and any suggestions for improvement - thanks!


nextgenusfs:
So seems like this is error?:
```
[Mar 29 20:59:45] SERR /bin/sh: amptk: not found
[Mar 29 20:59:45] ERRO Task processing failed: Unexpected exit code [127] of container [a7d8d5d9cbae step-d873861761], container preserved
```

The executable file is 'amptk' which imports 'amptk.py' from the same working directory.  I made an install `build.sh` script in same way that `snippy`(https://github.com/bioconda/bioconda-recipes/tree/master/recipes/snippy) was being installed, but I must be missing something.  

The `build.sh` looks like (simply copying the files to the $PATH folder? - but there is probably a better way to do this?):
```
#!/bin/bash

cp -r ./* $PREFIX/
```

The local tests passed, but seems like they are failing because `amptk` is not in the path?  Although since I'm new at this, I might be missing some other obvious error. The folder structure needs to be maintained for everything to work correctly, if I install this tool locally you can either symlink `amptk` into the $PATH or export the amptk folder into $PATH.

recipes/amptk/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
+  url: https://github.com/nextgenusfs/{{ name }}/archive/{{ version }}.tar.gz
+  sha256: db9cc4d4754d462cea85b8d050bd25c535038b6bb5bdab562e0a0294c195e034
+
+build:
+  number: 1
```
bgruening: new recipes should start with 0

recipes/amptk/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+cp -r ./* $PREFIX/
```
bgruening: We do not need all of these folders/files in the recipe, isn't it? Can you please explicitly copy the needed files?

nextgenusfs: I see, I just did a local conda build and see that this currently "pollutes" the environment.  The `amptk` script is a wrapper for running several other scripts in the directories, so it does need to have the `lib`, `util`, `bin`, and `DB` folders.  But after seeing what `conda build` did locally that is not what we want to do.  Is it possible to keep the `amptk` files/package in subdirectory and then just softlink the `amptk` script into the `bin` folder?  That is how I install it on home-brew for example.

nextgenusfs:
Tried to setup similarly to `mummer` recipe where move the necessary folder structure to `opt/amptk-version` and then symlink amptk to the `bin` directory.

druvus:
The r-ape package file seems to be corrupt. Opened a PR to rebuild it here: conda-forge/r-ape-feedstock#5

druvus:
@nextgenusfs could you please merge in the current master.

nextgenusfs:
@druvus yes thanks, I was just testing your change locally - which seems to fix the issue, thanks for that. I also had a few bug fixes upstream so bumped to the updated release.

nextgenusfs:
Thanks @druvus !

druvus:
thanks @nextgenusfs! Hopefully you could release the r-ape pinning when you are updating next time.

Issue #3017, opened 2016-11-21
OSError: [Errno 13] Permission denied: '/MaSuRCA-3.2.1_09012016'


mictadlo:
Hi, I tried to write a recipe for MaSuRCA as shown below:

meta.yaml
```
{% set name = "masurca" %}
{% set version = "3.2.1_09012016" %}

package:
  name: {{ name|lower }}
  version: {{ version }}

source:
  fn: {{ name|lower }}_{{ version }}.tar.gz
  url: ftp://ftp.genome.umd.edu/pub/MaSuRCA/MaSuRCA-{{ version }}.tar.gz

build:
  skip: True # [not py27]

requirements:
  build:
    - perl
    - gcc # [not osx]
    - llvm # [osx]
    - boost

  run:
    - perl
    - boost
    - libgcc  # [not osx]

test:
  commands:
    - variantCaller

about:
  home: https://github.com/PacificBiosciences/GenomicConsensus
  license: BSD-3-Clause
  summary: 'PacBio variant and consensus caller'
```
build.sh
```
#!/bin/bash

./install.sh
```

Unfortunately, I recived the following error:
```
> conda build recipes/masurca/
BUILD START: masurca-3.2.1_09012016-pl5.20.3_0
    (actual version deferred until further download or env creation)
updating index in: /QRISdata/Q0231/apps/conda-bld/linux-64
updating index in: /QRISdata/Q0231/apps/conda-bld/noarch
Using Anaconda API: https://api.anaconda.org

The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    perl-5.20.3.1              |                1        14.0 MB  conda-forge
    xz-5.2.2                   |                0         853 KB  conda-forge
    mpfr-3.1.4                 |                2         402 KB  conda-forge
    python-3.5.2               |                2        31.4 MB  conda-forge
    boost-1.62.0               |           py35_0        17.5 MB  conda-forge
    certifi-2016.9.26          |           py35_0         217 KB  conda-forge
    setuptools-28.8.0          |           py35_0         366 KB  conda-forge
    wheel-0.29.0               |           py35_0          82 KB  conda-forge
    pip-9.0.1                  |           py35_0         1.8 MB  conda-forge
    ------------------------------------------------------------
                                           Total:        66.6 MB

The following NEW packages will be INSTALLED:

    boost:           1.62.0-py35_0    conda-forge
    ca-certificates: 2016.9.26-0      conda-forge
[...]

PR #8492, opened 2018-03-30
Added (not yet working) recipe for MaSuRCA

jamesabbott:

This PR is to allow investigation of a problem with a source tarball containing absolute paths:

I'm trying to create a recipe for masurca, which it looks like has been tried unsuccessfully before bioconda/bioconda-recipes#3017, and I'm hitting the same problem. The cause seems to be that the tarball contains paths prefixed with '/', but conda-build extracts these equivalent to 'tar -P' and tries to create an absolute path of /MaSuRCA-3.2.4

### DIFF
--- a/recipes/masurca/build.sh
+++ b/recipes/masurca/build.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -x -e
+
+export BOOST_ROOT=${PREFIX}
+
+#../install.sh
+
--- a/recipes/masurca/meta.yaml
+++ b/recipes/masurca/meta.yaml
@@ -1,0 +1,36 @@
+{% set name = "masurca" %}
+{% set version = "3.2.4" %}
+
+about:
+    home: http://masurca.blogspot.co.uk/
+    license: GNU General Public License, Version 3.0
+    summary: | 
+        MaSuRCA (Maryland Super-Read Celera Assembler) genome assembly software.
+        MaSuRCA requires Illimina data, and supports third-generation PacBio/Nanopore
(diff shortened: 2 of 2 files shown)


mbargull:
The initial upstream responses don't look promising in regards to expecting a fixed source or handling by `conda-build` anytime soon (well, for the latter, unless one would prep a PR for `conda-build`).

I moved the download/extract to `build.sh`. That way, you can at least start working on this build.
I also removed the `gcc` build dependency so that we get the system/`devtoolset-2` GCC, which we need for compatibility with `conda-forge`'s `perl` builds.

jamesabbott:
A patch has now been accepted in conda-build to resolve the underlying problem extracting the tarballs. I guess I either need to also back-port this to cb2 or wait for the bioconda cb3 migration to be finished. Is waiting for cb3 the best bet?

ysu20:
Hi guys, I am also trying to package MaSuRCA for one of our scientists.  I tired to make some progress based on this PR and here is what I learned and observed:

1. the `build.sh` included in this PR needs to set environment variable, `DEST` to `${PREFIX}`; otherwise the `install.sh` won't install built binaries in the correct location.
1. MaSuRCA seem to depend on older version of Perl that is shipped with the OS where it is developed.  I am able to build it successful on a CentOS 7 server by using the Perl (v5.16.3) shipped with it.  However the build would fail if I specify the BioConda's preferred Perl (5.22.0.0) in the recipe.

Happy to contribute a commit  to address first issue; however I am not sure what's the best way to address the second.  Any suggestions?

PR #8494, opened 2018-03-30
Bump perl-dbd-sqlite to version 1.58

nathanweeks:

### DIFF
--- a/recipes/perl-dbd-sqlite/meta.yaml
+++ b/recipes/perl-dbd-sqlite/meta.yaml
@@ -1,23 +1,24 @@
+{% set version = '1.58' %}
+{% set sha256 = '7120dd99d0338dea2802fda8bfe3fbf10077d5af559f6c67ae35e9270d1a1d3b' %}
 package:
   name: perl-dbd-sqlite
-  version: '1.50'
+  version: {{ version }}
 
 source:
-  fn: DBD-SQLite-1.50.tar.gz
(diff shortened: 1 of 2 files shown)


recipes/perl-dbd-sqlite/meta.yaml
```diff
@@ -1,23 +1,24 @@
...
 
 build:
-  number: 1
+  number: 0
 
 requirements:
   build:
-    - gcc
-    - perl-threaded
+    - perl
```
epruesse: pin to `{{CONDA_PERL}}*`

recipes/perl-dbd-sqlite/meta.yaml
```diff
@@ -1,23 +1,24 @@
...
     - perl-app-cpanminus
-    - perl-module-build
     - perl-dbi
   run:
-    - perl-threaded
+    - perl
```
epruesse: same as above

epruesse:
Since this isn't really bio specific, the package should actually go to conda-forge. Do you think you could create a PR there? Ping me for review there.

nathanweeks:
@epruesse : sure thing. I assume that the build/run dependency perl-dbi would also needed in conda-forge?

epruesse:
@nathanweeks Well, technically we should be moving all the legacy non-bio perl packages from bioconda to conda-forge... I just don't have the time (and I'm not too much of a perl expert either).

PR #8495, opened 2018-03-30
Trinity recipe: adding missing dependency.

tstrnity:

This dependency is needed by util/support_scripts/run_TMM_scale_matrix.pl

### DIFF
--- a/recipes/trinity/meta.yaml
+++ b/recipes/trinity/meta.yaml
@@ -1,5 +1,5 @@
 build:
-  number: 1
+  number: 2
 # OSX build skipped because of this error: https://gist.github.com/pvanheus/2fbce93b326a3edb1a57aaebc8ca749e
   skip: True # [osx]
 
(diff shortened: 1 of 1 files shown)


yesimon:
This is more of an optional package right? https://github.com/trinityrnaseq/trinityrnaseq/wiki/Trinity-Differential-Expression You can also choose between other DE programs so it doesn't particularly make sense to only add edgeR.

tstrnity:
But if EdgeR is not in there, then it does not work. If there are other "optional" packages that do not have all of thier requirements in there, then the packages don't work either. When this package is getting pulled into a Galaxy environment, if there are missing dependencies for any particular package, then that package will not be usable within the conda environment without adding the dependency as a requirement of the tool. Why wouldn't you want all of these "optional" packages to be supported?

tstrnity:
Hmmm, not sure what happened with the build test. It worked with circleci on my computer.

PR #8506, opened 2018-04-02
Added new recipe for perl-perl-ostype

jamesabbott:

Requires perl-extutils-makemaker, perl-constant and perl-exporter which are present in bioconda but not conda-forge, so can not be submitted there...

### DIFF
--- a/recipes/perl-perl-ostype/1.010/build.sh
+++ b/recipes/perl-perl-ostype/1.010/build.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+if [ -f Build.PL ]; then
+    perl Build.PL
+    perl ./Build
+    perl ./Build test
+    perl ./Build install --installdirs site
+elif [ -f Makefile.PL ]; then
+    perl Makefile.PL INSTALLDIRS=site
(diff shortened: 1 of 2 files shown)


epruesse:
shouldn't this go to conda-forge?

jamesabbott:
As indicated in the comment in the PR, this has dependencies within bioconda which are not fulfilled by conda-forge.

Looking at the number of perl modules in each channel:

LS27911:~ jabbott$ conda search perl- |grep  bioconda|cut -f1 -d' '|sort -u|wc -l
     493
LS27911:~ jabbott$ conda search perl- |grep  conda-forge|cut -f1 -d' '|sort -u|wc -l
       4

I would consider the chance of getting any perl module into conda-forge without hitting a load of missing prerequisites extremely unlikely. Even perl-extutils-makemaker and perl-module-build which are used to build the modules (and populated into the meta.yaml by conda-skeleton) are not present on conda-forge.

Unless the perl infrastructure is moved entirely into conda-forge then I don’t seen any way round this, and have not hit problems getting perl modules accepted before for this same reason.

PR #8509, opened 2018-04-02
Update PePr from v1.1.18 to v1.1.23

shawnzhangyx:
* x[ ] This PR updates an existing recipe.

### DIFF
--- a/recipes/pepr/meta.yaml
+++ b/recipes/pepr/meta.yaml
@@ -1,11 +1,11 @@
 package:
   name: pepr
-  version: "1.1.18"
+  version: "1.1.23"
 
 source:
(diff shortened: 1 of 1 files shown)


recipes/pepr/meta.yaml
```diff
@@ -1,11 +1,11 @@
 package:
   name: pepr
-  version: "1.1.18"
+  version: "1.1.24"
```
bgruening: Can you please use variables for your version.

shawnzhangyx:
How about this?

recipes/pepr/meta.yaml
shawnzhangyx: Is this what you meant?

PR #8517, opened 2018-04-03
Upgrade HTSLib suite

kyleabeauchamp:

### DIFF
--- a/recipes/bcftools/meta.yaml
+++ b/recipes/bcftools/meta.yaml
@@ -1,4 +1,4 @@
-{% set version="1.7" %}
+{% set version="1.8" %}
 about:
   home: https://github.com/samtools/bcftools
   license: MIT
@@ -35,7 +35,7 @@ requirements:
 source:
   fn: bcftools-{{ version }}.tar.bz2
   url: https://github.com/samtools/bcftools/releases/download/{{ version }}/bcftools-{{ version }}.tar.bz2
-  sha256: dd4f63d91b0dffb0f0ce88ac75c2387251930c8063f7799611265083f8d302d1
+  sha256: 4acbfd691f137742e0be63d09f516434f0faf617a5c60f466140e0677915fced
 
 test:
   commands:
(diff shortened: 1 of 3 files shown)


kyleabeauchamp:
The latest versions of HTSLib support a new compression backend (libdeflate) that offers improved performance.  If these benchmarks prove to be true, does anyone here have any objections to adding lib-deflate to the build for these libraries (and pysam, eventually)?

kyleabeauchamp:
PS: I will do the lib-deflate in a separate PR (with an incremented build number), just wanted to chat about it first.  I should probably open a separate issue.

kyleabeauchamp:
FYI I haven't updated the HTSLIB environment matrix yet

epruesse:
Is there any reason not to enable libdeflate right away? Why did they chose to just make it a compile time flag, rather than the new default?

Is the htslib so backwards compatible, or would we need to recompile?

kyleabeauchamp:
So for me, I want to make sure that downstream applications are not impacted.  I also want to wait until pysam supports libdeflate as well.

dpryan79:
Won't bumping the build number for libdeflate support obviate the need for a non-libdeflate build? Since the API is unchanged, I presume pysam will be able to use libdeflate by simply compiling against a libdeflate-enabled htslib.

kyleabeauchamp:
FWIW, I have preliminary evidence (below) that libdeflate-based gzip is faster for both compression and decompression, so I think this might be worthwhile:

```
$ time gzip -c ~/dat/genomes/hg19/bwa_ref/hg19.bwa_mem.fa > out.fa.gz

real	8m17.887s
user	8m12.304s
sys	0m3.715s
$ time ./gzip -c ~/dat/genomes/hg19/bwa_ref/hg19.bwa_mem.fa > out2.fa.gz

real	1m38.979s
user	1m32.337s
sys	0m3.826s

$ time gunzip -c out2.fa.gz  > /dev/null 

real	0m8.089s
user	0m7.719s
sys	0m0.338s
$ time ./gunzip -c out2.fa.gz  > /dev/null 

real	0m6.580s
user	0m5.096s
sys	0m1.452s
```

dpryan79:
This seems to jive with what was reported on the htslib repository :)

epruesse:
Wow. That's quite the difference. Who would have thought that zlib has that much room for improvement. How does the compression ratio hold up?

PR #8521, opened 2018-04-03
Add libdeflate, try building HTSLib with it

kyleabeauchamp:

### DIFF
--- a/recipes/bcftools/build.sh
+++ b/recipes/bcftools/build.sh
@@ -4,10 +4,10 @@ export CPPFLAGS="-I$PREFIX/include"
 export LDFLAGS="-L$PREFIX/lib"
 
 cd htslib*
-./configure --prefix=$PREFIX --enable-libcurl CPPFLAGS="-I$PREFIX/include" LDFLAGS="-L$PREFIX/lib"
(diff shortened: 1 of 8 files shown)


jtratner:
would it make sense to create a separate samtools-deflate package for this to allow folks to test it? (otherwise you're opting everyone in to libdeflate)?

Not sure level of robustness that samtools maintainers feel they have.

kyleabeauchamp:
Seems like the HTSLib developers are fairly confident, but that's a good thing to think about.

epruesse:
@jtratner On the htslib/samtools side, the coding would not have been very invasive, so unlikely to contain breakage. Libdeflate itself seems to be stable already judging from the recent releases, and since it produces compatible output, I wouldn't worry too much.

jtratner:
cool - sounds good to me - seems like we don't need a separate package :)

epruesse:
Technically, libdeflate is a candidate for conda-forge though. It's likely to appear there eventually, and then we have the duplication issue again.

kyleabeauchamp:
Any objections?  I tested that I do see substantial speedups when I use the builds from my local bioconda build of the same recipes.

kyleabeauchamp:
FYI: future todos on this story are:

1.  When CB3 / GCC7 is available, upgrade the compiler to see if performance improves
2.  Add additional compiler optimizations / -march targets as appropriate to further improve performance.

PR #8525, opened 2018-04-04
Adding NGS VDB for sra-tools

sachalau:


sachalau:
The make test failed for some reasons. I am not sure I can do something about it.

> Test: CacheTee_ReadOnly cache.dat.cache
/opt/conda/conda-bld/ncbi-vdb_1522845593863/work/ncbi-vdb-2.9.0-1/test/kfs/cacheteetest.cpp:562: fatal error in "CacheTee_ReadOnly": Unexpected return code from KDirectoryOpenFileRead( dir, &cache, "%s", CACHEFILE ): RC(rcNoErr)
Test: CacheTee_Multiple_Users_with_Promoting
Test: CacheTee_Multiple_Users_without_Promoting
*** 1 failure detected in test suite "CacheTeeTests"
and the result is: RC(INVALID,INVALID,INVALID,INVALID,INVALID)

epruesse:
Any chance you can run at least some of the tests? Something like `make -C tests/vdb` to get at least something run.

You might want to look at the package in debian-med. It appears Andreas Tille has been working on it recently (see https://salsa.debian.org/med-team/ncbi-vdb) . Judging from the patches (./debian/patches folder), he was also having trouble with plenty tests failing.

As it is, you only check that the library builds and that the files are present in the package archive. There is a lot that still can go wrong. We don't have to test that the library content is correct, but it should at least load. So if nothing else works, try a minimal program like the ones `configure` generates to verify that the lib can be linked. 

Say, `test.c` containing:
```C
#include <ncbi-vdb-header-file.h>

const char var[] = "blablub";

int main() {
  ncbi_vdb_api_call(arg);
  return 0;
}
```

And a test compiling and executing
```
gcc -o test test.c -lncbi-vdb -I$PREFIX/include -L$PREFIX/lib
./test
```

sachalau:
Thanks for your inputs ! I followed you advice and indeed the test specific to vdb seem to be working.

recipes/ncbi-vdb/build.sh
```diff
@@ -0,0 +1,5 @@
+export ROOT=$PREFIX
```
bgruening: is the ROOT needed here?

sachalau:
I will try but NCBI-NGS did not work without the root export.

sachalau:
I get the following error withouth `export ROOT`:

```
ln: creating symbolic link `/opt/conda/usr/include/ncbi-vdb': No such file or directory
```

I also see now that `make install` complains that `bc` is not present. Do you know which conda forge package ships it ?

kwrodarmer:
Rather than issue a PR or the like, could you explain more concretely what it is that you're trying to achieve so that we can analyze the problem in NCBI and provide a fix?

You shouldn't need to be setting the fs ROOT variable for this test: the reason it is there is so that WE can test our behavior in the case that we would have root account access (and we can't on Federal computers), so we manually can set ROOT to somewhere else to simulate root access.

You appear to be trying to set the install prefix?

- NCBI SRA Toolkit Team

sachalau:
If I don't set the ROOT variable to PREFIX (where we are actually building the package) the make install fails. If just export ROOT=$PREFIX, then I can build without any problem.

sachalau:
Hello ! Could I get a merge on this please, I think the `export ROOT=$PREFIX` is good enough no ?

recipes/ncbi-vdb/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - gcc # [not osx]
+    - llvm # [osx]
+    - perl
+    - libiconv
+    - libxml2
+    - ncbi-ngs-sdk
+    - bison
+
+  run:
+    #
```
epruesse: No runtime dependencies? Are these now filled automatically? I somehow doubt the library will work without at least `libxml2`, `libiconv` and `libgcc`.

recipes/ncbi-vdb/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+about:
+  home: https://github.com/ncbi/ncbi-vdb
+  license: Public Domain
+  summary: The SRA Toolkit and SDK from NCBI is a collection of tools and libraries for using data in the INSDC Sequence Read Archives.
```
epruesse: I know that's what it says on their Github page. What does `ncbi-vdb` do though? 

```
summary: NCBI libraries for accessing the INSDC Sequence Read Archive (SRA)
description: |
  The SRA Toolkit and SDK from NCBI is a collection of tools and libraries for using
  data in the INSDC Sequence Read Archives. ...
```

recipes/ncbi-vdb/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+test:
+  commands:
+    - test -e ${PREFIX}/lib64/libncbi-vdb.so # [not osx]
+    - test -e ${PREFIX}/lib64/libncbi-vdb.dylib # [osx]
+
```
epruesse: I still would like to see some binary using the library executed here. Very likely that'll show that the `run:` section needs some content, too.

epruesse:
@kwrodarmer Could you shed some light on why @sachalau is having trouble getting `make install` to comply without the `ROOT` variable?

recipes/ncbi-vdb/meta.yaml
kwrodarmer: We link against `libxml2` when found, but it is not required. In fact, its only use is as a relic for some very old software that is no longer supported. We also detect `libhdf5` and include some additions for that if found, in order to support a loader that is only run within NCBI. We link against `-ldl`, `-lm` and `-lpthreads`.

recipes/ncbi-vdb/meta.yaml
kwrodarmer: VDB is the database engine that all SRA tools use. SRA is built upon VDB.
VDB is a columnar database system with a number of unique features, but basically all SRA objects are stored in VDB.

kwrodarmer:
I have not yet understood the nature of the problem @sachalau is having. I have tried to explain the nature of the $ROOT variable hack within our installation, which is there simply because we need to test behavior of the script when run as root, without having the ability to actually become root on any government system. The use of the variable is not intended to be used by anyone for any other purpose.

kwrodarmer:
One of the things that we thought immediately upon reading your first post was that perhaps $ROOT was defined to be something else when you run the normal install, and redefining it to something compatible makes it work. Can you confirm that ROOT is undefined when you get the error?

sachalau:
I guess `ROOT` is set to `/opt/conda/usr/` by default in our environment.

epruesse:
@sachalau Any luck on producing a test?

sachalau:
Thanks a lot @epruesse for your help & reminder ! I started working on it again and here is where I am stuck: the test fails because there's an `#include <exception>`.

The exact error is : 
```
In file included from /opt/conda/conda-bld/ncbi-vdb_1526304859910/_t_env/include/ngs/ErrorMsg.hpp:31:0,
                 from /opt/conda/conda-bld/ncbi-vdb_1526304859910/_t_env/include/ngs/Fragment.hpp:31,
                 from /opt/conda/conda-bld/ncbi-vdb_1526304859910/_t_env/include/ngs/FragmentIterator.hpp:31,
                 from /opt/conda/conda-bld/ncbi-vdb_1526304859910/_t_env/include/ngs/Read.hpp:31,
                 from /opt/conda/conda-bld/ncbi-vdb_1526304859910/_t_env/include/ngs/ReadIterator.hpp:31,
                 from /opt/conda/conda-bld/ncbi-vdb_1526304859910/_t_env/include/ngs/ReadGroup.hpp:31,
                 from /opt/conda/conda-bld/ncbi-vdb_1526304859910/_t_env/include/ngs/ReadGroupIterator.hpp:31,
                 from /opt/conda/conda-bld/ncbi-vdb_1526304859910/_t_env/include/ngs/ReadCollection.hpp:31,
                 from /opt/conda/conda-bld/ncbi-vdb_1526304859910/_t_env/include/ncbi-vdb/NGS.hpp:31,
                 from test.c:1:
/opt/conda/conda-bld/ncbi-vdb_1526304859910/_t_env/include/ngs/itf/ErrorMsg.hpp:34:21: fatal error: exception: No such file or directory
 #include <exception>
                     ^
```
From what I have google exception should be part of the standard libraries (http://www.cplusplus.com/doc/tutorial/exceptions/) of C++ but somehow it is not found. I added gcc as a requirement of the test but without any change. Do you have any idea why exception is not found ?

rvalieris:
good work @sachalau, I managed to get this built by fixing 2 typos: `-m -lpthreads` -> `-lm -lpthread` and adding rpath: `-Wl,-rpath=$PREFIX/lib64`, usually conda deals with rpath for us automagically, but that not happen when we build stuff on the test section.

however the main problem here is mulled-build does not honor test-requires/test-files, we could echo test.c inline but we will have to move the test-requires to the run-requirements, which might be a deal breaker..

personally I would be ok with only running the make tests to avoid these extra dependencies.

sachalau:
Thank you so much for your help I was getting quite stuck here! I think the - lm and - lpthread are actually not needed but I was trying many different things to get this working... The rpath was probably the what got me stuck. I am not sure I am following your second argument though, would you rather build my home made test in the run section instead of the test section?

sachalau:
Ok now I understand your second comment, thanks @rvalieris ! Indeed adding gcc at run time dependency does not seem very useful except for the inhouse test, @epruesse what do you think, should we keep trying to get this test run or are you okay with the one already included ?

rvalieris:
libtool is not the problem, you can leave the default one.
the problem is:
>build prefix: UNKNOWN/ncbi-outdir/ncbi-vdb
[....]
>error: /Applications/Xcode-8.3.3.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/libtool: can't create output file: UNKNOWN/ncbi-outdir/ncbi-vdb/mac/clang/x86_64/rel/ilib/libz.a (No such file or directory)

this "UNKNOWN" is probably a undefined variable that should have a directory. the docs says that by default the build is output on $HOME (why ?!), for some reason I think $HOME is unset.

try adding this to configure: `--build-prefix $PWD/outdir`

rvalieris:
cool, I am merging it like this. in the future we should have a way to exclude some tests from running on mulled-build (xref: https://github.com/bioconda/bioconda-utils/issues/183), but for now I would rather avoid including gcc as run dependency. the real test will be getting sra-tools to link against these.

sachalau:
`the real test will be getting sra-tools to link against these`

Yes, working on it soon !

PR #8537, opened 2018-04-05
Enable easy local testing of mulled-build

daler:

After we moved to CircleCI, it got harder to do local tests with mulled-build due to docker not running nicely from within docker using `circleci build`. This PR makes some changes to `.circleci/setup.sh` to make it play nicer with the new driver script `bootstrap.py`. The driver script creates a custom `activate` script to activate the new conda installation.

Example usage:

```bash
$ ./bootstrap.py /tmp/minconda
$ source activate ~/.config/bioconda/activate
(BIOCONDA-UTILS) $ bioconda-utils build recipes config.yml --mulled-build --git-range master
(BIOCONDA-UTILS) $ source deactivate
```

### DIFF
--- a/.circleci/setup.sh
+++ b/.circleci/setup.sh
@@ -1,7 +1,11 @@
 #!/bin/bash
 set -eu
 
-WORKSPACE=`pwd`
+set +u
+[[ -z $WORKSPACE ]] && WORKSPACE=`pwd`
+[[ -z $BASH_ENV ]] && BASH_ENV=`tempfile`
+[[ -z $BOOTSTRAP ]] && BOOTSTRAP=false
+set -u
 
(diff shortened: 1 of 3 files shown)


bgruening:
@daler out of interest what was the error which you are seeing. It seemed to work for me really nicely.

daler:
@bgruening no error, it's just that I don't think `circleci build` runs the mulled-build tests (https://github.com/bioconda/bioconda-recipes/blob/master/.circleci/config.yml#L79). So there's not really a good way of locally debugging failures from the mulled-build stage using just the CircleCI client. Also I know folks have complained in the past about messing up their existing root env when installing bioconda-utils, so this tries to take care of all of those points.

daler:
(making sure a recipe builds correctly, will revert build number once it does)

PR #8541, opened 2018-04-05
New recipe for tardis.

tesujimath:

### DIFF
--- a/recipes/tardis/meta.yaml
+++ b/recipes/tardis/meta.yaml
@@ -1,0 +1,62 @@
+{% set name = "tardis" %}
+{% set version = "0.5.9" %}
+{% set sha256 = "887e198c988bb916122f87030f2f43caecf4e308563caef344117b52aa130e14" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
+  url: https://github.com/AgResearch/{{ name }}/archive/v{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  skip: True  # [not py27]
+  number: 0
+  entry_points:
+    - tardis = tardis.run:tardis_main
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - setuptools_scm
+    - setuptools_scm_git_archive
+    - make
+    - gcc  # [not osx]
+    - llvm # [osx]
+    - zlib {{CONDA_ZLIB}}*
+
+  run:
+    - python
+    - libgcc # [not osx]
+    - zlib {{CONDA_ZLIB}}*
+
+test:
+  imports:
+    - tardis
+  commands:
+    - tardis -h
+    - kseq_split -h
+
+about:
+  home: https://github.com/AgResearch/{{ name }}
+  license: GPL-2.0
+  license_family: GPL
+  license_file: LICENSE
+  summary: 'Pre-processor for bioinformatics cluster job submission'
+  description: |
+    Tardis is a pre-processor which reconditions (precompiles) a marked-up unix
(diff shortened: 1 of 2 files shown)


recipes/tardis/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+build:
+  skip: True  # [win]
+  number: 0
+  entry_points:
+    - tardis = tardis.run:tardis_main
+
+requirements:
+  build:
+    - python 2.7.*
```
daler: Change to `- python

recipes/tardis/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
+  url: https://github.com/AgResearch/{{ name }}/archive/v{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  skip: True  # [win]
```
daler: We don't build on windows, but it looks like you want to restrict to py27, so change `# [win]` to `# [not py27]`

recipes/tardis/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - setuptools
+    - setuptools_scm
+    - setuptools_scm_git_archive
+    - make
+    - gcc  # [not osx]
+    - llvm # [osx]
+    - zlib {{CONDA_ZLIB}}*
+
+  run:
+    - python 2.7.*
```
daler: change to `- python`

recipes/tardis/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - zlib {{CONDA_ZLIB}}*
+
+test:
+  imports:
+    - tardis
+  source_files:
+    - kseq_split/test.fastq
+  commands:
+    - tardis -h
+    - kseq_split -s 0.1 -o fastq test.fastq 10 split-%05d.fastq
```
daler: I'm surprised this works . . . @bgruening does mulled-build copy over test files from source now?

tesujimath: Ah, I see kseq_split is returning a success exit code even when it can't find its input files.  Fixed now.

recipes/tardis/meta.yaml
tesujimath: Done

recipes/tardis/meta.yaml
tesujimath: Thanks, yes.  Done.

recipes/tardis/meta.yaml
tesujimath: Done.

tesujimath:
@daler Thanks for the review.  I have addressed your comments, and pushed a new version.  Tardis absolutely is bioinformatics specific.  It does domain-aware job-splitting and output file recombining, for one thing.

daler:
@tesujimath thanks. I had to read through more of the readme to find the bioinformatics-specific stuff; for others following this, https://github.com/AgResearch/tardis says:

> - content-aware splitting of large input files. tardis includes a fast C based fasta and fastq splitter
> - handling multiple inputs which must be split in lockstep - for example pairs of paired-end sequence files

I also see:

> - there are a couple of features such as filtering sequences by length that require biopython - however for most use-cases tardis will run fine without biopython

Do you think it's worth it to include biopython as a dependency?

tesujimath:
@daler We really don't want to include BioPython as a dependency.  Tardis works find without it.  It only uses it for one corner case, which is quite rare.

Updating the README is in hand.

daler:
OK, thank you @tesujimath!

PR #8545, opened 2018-04-06
satsuma new recipe

mictadlo:

### DIFF
--- a/recipes/satsuma/build.sh
+++ b/recipes/satsuma/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+cmake -D CMAKE_INSTALL_PREFIX:PATH=$PREFIX .
+make
+mkdir -p $PREFIX/bin
+cp bin/* $PREFIX/bin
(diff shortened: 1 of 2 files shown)


recipes/satsuma/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+cmake -D CMAKE_INSTALL_PREFIX:PATH=$PREFIX .
+make
+mkdir -p $PREFIX/bin
```
bgruening: is `make install` not available?

recipes/satsuma/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  url: https://github.com/bioinfologics/satsuma2/archive/da694aeecf352e344b790bea4a7aaa529f5b69e6.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+  skip: True  # [osx]
+
+requirements:
+  build:
+    - gcc_linux-64  # [linux64]
```
bgruening: The normal `gcc` package is not working?

mictadlo: gcc is too old.

recipes/satsuma/build.sh
mictadlo: I tried `make install` but it failed.

PR #8552, opened 2018-04-06
Bump sicer

daler:

Not actually sure how the last recipe built, as the sha256 was commented out and the test used `$SRC_DIR` which shouldn't exist in the mulled-build container. @sebastian-luna-valero, do you recall anything about the sha256?

### DIFF
--- a/recipes/sicer/build.sh
+++ b/recipes/sicer/build.sh
@@ -10,10 +10,7 @@ OUTDIR=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
 mkdir -p $OUTDIR
 
 # install folders
-cp -r $SRC_DIR/SICER/ex $OUTDIR
-cp -r $SRC_DIR/SICER/lib $OUTDIR
-cp -r $SRC_DIR/SICER/src $OUTDIR
-cp -r $SRC_DIR/SICER/utility $OUTDIR
+cp -r $SRC_DIR/SICER $OUTDIR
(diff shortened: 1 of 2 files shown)


sebastian-luna-valero:
Thanks @daler 

Sorry, I don't fully remember the details about the sha256. I think this package was build using Travis-CI. 

What I remember is:
* According to my tests at that time, `sicer` only works in Python 2.6 and therefore the restriction in the runtime deps.
* I emailed the developers asking about the license, and we decided to go for MIT.

I hope that helps.

Best regards,
Sebastian

daler:
@sebastian-luna-valero thanks for the info.

The current tests appear to be OK with py27, but do you remember any details on what required py26? We've been running it locally in production on py27 with no issues.

In our hands the existing package is broken because sicer has lots of hard-coded links to other files which need to be in the `SICER` subdirectory of `$OUTDIR`. I don't know if the tarball was changed on the authors' server without a version bump or why otherwise travis-ci tests passed! 

I'll change back to MIT based on your conversation with the developers, and add a comment to the recipe.

sebastian-luna-valero:
I am very sorry, this is also a mystery to me!

I don't remember the details and I don't find any references in my notes. I think during one of the runs we found an import to a py26 module.... Anyway, it looks like it's properly running in py27 so I might have missed something.

daler:
OK I'll merge this and we'll test it on real data and make any adjustments as need. Thanks for the info!

Issue #8586, opened 2018-04-10
fastqc needs fonts, no fonts in container


charlesreid1:
Hi,

I am running a workflow that involves a step with fastqc. I'm running into a problem in the last step of fastqc, when it has completed the analysis and is attempting to generate a report. The Java engine is crashing because it cannot find font files it needs.

From inside container:

```
$ java -version
openjdk version "1.8.0_121"
OpenJDK Runtime Environment (Zulu 8.20.0.5-linux64) (build 1.8.0_121-b15)
OpenJDK 64-Bit Server VM (Zulu 8.20.0.5-linux64) (build 25.121-b15, mixed mode)
```

Here's the command I was running:

```
docker run \
    -u $(id -u):$(id -g) \
    -v /home/ubuntu/dahak-taco/data:/data/ \
    -it \
    quay.io/biocontainers/fastqc:0.11.7--pl5.22.0_0 \
    fastqc \
    -t 1 \
    /data/SRR606249_1_reads.fq.gz \
    -o /data/
```

And here's the output that was printed:

```
Approx 50% complete for SRR606249_1_reads.fq.gz
Approx 55% complete for SRR606249_1_reads.fq.gz
Approx 60% complete for SRR606249_1_reads.fq.gz
Approx 65% complete for SRR606249_1_reads.fq.gz
Approx 70% complete for SRR606249_1_reads.fq.gz
Approx 75% complete for SRR606249_1_reads.fq.gz
Approx 80% complete for SRR606249_1_reads.fq.gz
Approx 85% complete for SRR606249_1_reads.fq.gz
Approx 90% complete for SRR606249_1_reads.fq.gz
Approx 95% complete for SRR606249_1_reads.fq.gz
Analysis complete for SRR606249_1_reads.fq.gz
Exception in thread "Thread-1" java.lang.Error: Probable fatal error:No fonts found.
        at sun.font.SunFontManager.getDefaultPhysicalFont(SunFontManager.java:1236)
        at sun.font.SunFontManager.initialiseDeferredFont(SunFontManager.java:1100)
        at sun.font.SunFontManager.findOtherDeferredFont(SunFontManager.java:1037)
        at sun.font.SunFontManager.findDeferredFont(SunFontManager.java:1054)
        at sun.font.SunFontManager.findFont2D(SunFontManager.java:2256)
        at sun.font.SunFontManager.getDefaultPhysicalFont(SunFontManager.java:1220)
[...]

PR #8588, opened 2018-04-11
Fixing openjdk version for fastqc container

charlesreid1:
Updating openjdk to come from conda-forge explicitly, rather than default. The openjdk in default is old and does not include fonts; the conda-forge version is the more recent. This change is also discussed in #8586 and #5026.

* <s>This PR adds a new recipe.</s>
* <s>This PR does something else (explain below).</s>

### DIFF
--- a/recipes/fastqc/meta.yaml
+++ b/recipes/fastqc/meta.yaml
@@ -19,7 +19,11 @@ build:
 
 requirements:
   run:
-    - "conda-forge::openjdk"
+    # Enforce a version requirement on openjdk to ensure
+    # it comes from the conda-forge channel and not default.
+    # Many yaks were shaved to bring us this information.
+    # Version number reference: https://github.com/conda/conda/issues/6948#issuecomment-369360906
+    - openjdk>8.0.121
     - perl
 
 test:
--- a/recipes/fastqc/meta.yaml
+++ b/recipes/fastqc/meta.yaml
@@ -23,7 +23,7 @@ requirements:
     # it comes from the conda-forge channel and not default.
     # Many yaks were shaved to bring us this information.
     # Version number reference: https://github.com/conda/conda/issues/6948#issuecomment-369360906
-    - openjdk>8.0.121
+    - openjdk >8.0.121
     - perl
 
 test:
--- a/recipes/fastqc/meta.yaml
+++ b/recipes/fastqc/meta.yaml
@@ -19,7 +19,7 @@ build:
 
 requirements:
   run:
-    - openjdk
+    - "conda-forge::openjdk"
     - perl
 
 test:


bgruening:
@charlesreid1 this error should not happen if you have the channel order correctly set. Make sure conda-forge has a higher priority than defaults.

charlesreid1:
This was discussed in #5026, where @ewels stated:

> So the channel order should be correct I think, but I'm still getting the error.
> Note - that if I explicitly add conda-forge::openjdk to the Dockerfile conda
>  install command, everything is fine

Hence the PR - but it looks like it is making something unhappy.

ewels:
Agreed - there is something more than just the channel order happening here. See our [`environment.yml`](https://github.com/SciLifeLab/NGI-RNAseq/blob/master/environment.yml), where we do have the correct channel order, yet the explicit `conda-forge::` definition is still needed.

Phil

daler:
@ewels ah, that's the issue: the channel order in that [`environment.yaml`](https://github.com/SciLifeLab/NGI-RNAseq/blob/master/environment.yml) is flipped.

It's confusing. When adding things from the command line, you use:
```
conda config --add channels defaults
```

but upon doing that, the resulting order that shows up in `~/.condarc` is:

```yaml
channels:
  - bioconda
  - conda-forge
  - defaults
```

so the environment.yaml's channels section needs to have defaults last in the list.

daler:
@ewels whoops, while the channel order in that env yaml may be an issue, I can reproduce openjdk coming from defaults with what I'm pretty sure is a correct channel order:

```
 conda create -n tmp --dry-run fastqc                                                                                                                                                                                                                    
Fetching package metadata ...........
Solving package specifications: .

Package plan for installation in environment /anaconda/envs/tmp:

    fastqc:                    0.11.7-pl5.22.0_0 bioconda
    font-ttf-dejavu-sans-mono: 2.37-0
    openjdk:                   8.0.121-1
    perl:                      5.22.0.1-0        conda-forge

```

Unfortunately I think the `::` syntax in dependencies as used in this PR is a conda-build v3 feature which we have not yet moved to. Hence the CircleCI error. Will poke around with this some more.

ewels:
Thanks @daler - glad you also replicated manually. I was pretty sure that I'd done the same thing outside of just the `environment.yml` file. Will still update that as well though, thanks for the explanation 👍

daler:
reporting some more info: 

@caseywdunn removed the font dependency for 0.11.7 in #7443. Based on the commit message, this was because it was causing openjdk to come from defaults.

In #8600 I bumped the build number, keeping the fonts out. Note that in my [comment above](https://github.com/bioconda/bioconda-recipes/pull/8588#issuecomment-380475357) the fonts were being picked up, despite the previous recipe not including it. That seemed strange to me and I can't explain that yet.

Anyway, when trying to install the bumped-build-number FastQC, the fonts are not brought in as they were before. But openjdk is still coming from defaults:

```
$ conda create -n tmp --dry-run fastqc
Solving package specifications: .

Package plan for installation in environment /tmp/aaa/miniconda/envs/tmp:

    fastqc:  0.11.7-pl5.22.0_1 bioconda
    openjdk: 8.0.121-1
    perl:    5.22.0.1-0        conda-forge
```

I noticed that if I just specify openjdk when creating the environment, the proper one comes in:

```
$ conda create -n tmp --dry-run fastqc openjdk
Solving package specifications: .

Package plan for installation in environment /tmp/aaa/miniconda/envs/tmp:

    fastqc:  0.11.7-pl5.22.0_1      bioconda
    openjdk: 8.0.144-zulu8.23.0.3_1 conda-forge
    perl:    5.22.0.1-0             conda-forge
```

This also works when openjdk is specified in an environment yaml, for example with the following minimal environment.yaml:

```yaml
name: ngi-rnaseq
channels:
  - bioconda
  - conda-forge
  - defaults
dependencies:
  - openjdk
  - fastqc=0.11.7
```

```
$ conda env create --file env.yaml -v
Using Anaconda API: https://api.anaconda.org

Solving package specifications: .
()

===> LINKING PACKAGE: conda-forge::openjdk-8.0.144-zulu8.23.0.3_1 <===
  prefix=/tmp/aaa/miniconda/envs/ngi-rnaseq
  source=/tmp/aaa/miniconda/pkgs/openjdk-8.0.144-zulu8.23.0.3_1


===> LINKING PACKAGE: conda-forge::perl-5.22.0.1-0 <===
  prefix=/tmp/aaa/miniconda/envs/ngi-rnaseq
  source=/tmp/aaa/miniconda/pkgs/perl-5.22.0.1-0


===> LINKING PACKAGE: bioconda::fastqc-0.11.7-pl5.22.0_1 <===
  prefix=/tmp/aaa/miniconda/envs/ngi-rnaseq
  source=/tmp/aaa/miniconda/pkgs/fastqc-0.11.7-pl5.22.0_1


#
# To activate this environment, use:
# > source activate ngi-rnaseq
#
# To deactivate an active environment, use:
# > source deactivate
#
```

**So: currently, the solution seems to be to specify openjdk in addition to fastqc**, with no need to specify the conda-forge channel using the conda-build 3 syntax. However I am still super confused as to why this is happening in the first place.

@mbargull I seem to remember you chasing down some obscure dependency issues, do you have any hints for figuring this out?

daler:
@charlesreid1 thanks, I missed that comment. By the end of this we will have yak fur parkas for everyone!

epruesse:
++yak :)

(I didn't know that idiom... very nice... feel like that describes so much of what I do...)

mbargull:
> @mbargull I seem to remember you chasing down some obscure dependency issues, do you have any hints for figuring this out?

Didn't read the whole issue here. But if it's about `conda` picking up `openjdk` from `defaults`, that's another case of `conda`'s solver's inconsistent handling of requested and dependecy packages with `features`/`track_features`, see https://github.com/conda/conda/issues/6948#issuecomment-369360906.
There's not much one can do about this apart from manually adding `openjdk` to the *requested* packages, e.g., with `conda create -n tmp --dry-run fastqc openjdk` (given that `conda-forge` has a higher priority that `defaults`) or adding the `openjdk` feature to the environment.

charlesreid1:
If I understand correctly, it sounds like these solutions happen on the user side. But for the case of building a fastqc docker container image for uploading quay.io, everything must be specified in the recipe, so I don't know how to apply them.

The fastqc `meta.yaml` file lists openjdk under the recipe's requirements (link: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/fastqc/meta.yaml#L22). Is that equivalent to @mbargull's suggestion of adding openjdk into the environment, or running the `conda create` command with both fastqc openjdk specified?

Alternatively, in the linked thread another solution suggested was to constrain the installable openjdk using e.g., `conda config --add pinned_packages 'openjdk >8.0.121`. Is it possible/wise to do that in this case?

mbargull:
Yes, all of the given workarounds are to be applied by the user and can't be applied to the build process.
For the CI -- and thus for the container creation -- I think the only currently working workaround would be to add the version constraint to the requirement, i.e.:
```diff
 requirements:
   run:
-    - openjdk
+    # !! useful comment describing the obscure lower version constraint !!
+    - openjdk >8.0.121
     - perl
```

charlesreid1:
Is `8.0.121` the first openjdk version available  in the conda-forge channel and not in the default channel?

mbargull:
From https://github.com/conda/conda/issues/6948#issuecomment-369360906:
> `openjdk-8.0.121-1.tar.bz2` is the last version without `track_features`.

From https://github.com/conda/conda/issues/6948#issuecomment-369368081:
> The problem isn't conda-forge here, but that defaults::conda-forge=8.0.152 added track_features="openjdk" and existing features aren't handles properly.
If you want to avoid to always supply openjdk as part of the package list for conda install & co, you can constrain the installable openjdk versions, [...]

charlesreid1:
We cannot specify a version requirement because the `>` char is invalid. CI failing with:

```
Filtering 1 of 5 (20.0%) recipes/fastqc       
Error: bad character '>' in package name dependency 'openjdk>8.0.121'
Exited with code 1
```

charlesreid1:
::passes out yak fur parkas and cigars and caviar and shakes everyone's hand and slaps everyone on the back::

Thanks everyone! I will follow up if there are any issues with the quay.io biocontainer image, since that's where the issue was coming up originally.

ewels:
Fantastic! So it's safe to drop the `openjdk` dependency completely now in new environments?

charlesreid1:
@ewels That's 💯 % correct!

ewels:
Hmm, ok so things are not working quite as smoothly as I'd hoped. Removing the `openjdk` dependency makes my conda build process hang at the `Solving environment:` stage (at least - long enough to make Dockerhub builds fail and I gave up after waiting > 30 mins).

Interestingly, if I add `conda-forge::openjdk=8.0.144` back it works fine. So definitely seems to be something to do with resolving this package.

Presumably something related to https://github.com/conda/conda/issues/5536 ?

epruesse:
I've run into that one before as well. Usually more specific version requirements help. I do wonder whether it's a bug or just the potentially enormous complexity (NP?) resolving the dependencies...

mbargull:
> potentially enormous complexity (NP?) resolving the dependencies

This.

@ewels: Try adding a `r-base >=3.4` dependency or try `-c conda-forge -c bioconda`, i.e., letting `conda-forge` have a higher priority than `bioconda`. Both of those suggestions should work.

mbargull:
Nope, I have not. Which upstream do you mean?
For `conda` it's just another case of a known issue, i.e., `features` being problematic and have to be reworked with a breaking change (maybe for some `conda 5.x` release?? but nothing happening very soon necessarily).
Upstream as in https://github.com/AnacondaRecipes/openjdk-feedstock/ and https://github.com/conda-forge/openjdk-feedstock to request to remove the `track_features: [openjdk]` would make sense, though.

ewels:
@mbargull - can this be done in the FastQC recipe somehow? eg. Pinning the `conda-forge` channel to the `openjdk` dependency or something? Everything is working fine for me now that I'm specifying `openjdk` in my environment yaml files, but it would be nice to contain this fix within the fastqc recipe for people attempting `conda install fastqc`...

PR #8594, opened 2018-04-11
Added biotools identifiers

ArneKr:


ArneKr:
@bgruening At the meeting we talked about only making one pull request at a time for the annotations to not block other pull requests but it seems that circleci just skips them anyway as there is nothing new to do. Is it ok if i just merge them all at once or should i wait?

bgruening:
You can make a giant PR that we than all review. As long as you don't increase the build number or the package number CI will not test the recipe it will only lint it.

Thanks for your work on this!

PR #8602, opened 2018-04-11
Increase minimal blast version for plant_tribes_gene_family_classifier

gregvonkuster:

Older versions of blast throw the following exception on some Linux systems, and this PR should correct this problem.

```blastp: error while loading shared libraries: libgomp.so.1: cannot open shared object file: No such file or directory```


gregvonkuster:
Thanks @bgruening !

PR #8661, opened 2018-04-18
add akt

daler:

Several things I want to point out about this recipe:

1. The latest github release of akt is from 2016, and the tarball for it includes `htslib-1.3`. Similar to the samtools recipe, here I'm letting the build use that vendored `htslib` rather than patching it to use a pinned `CONDA_HTSLIB`.

2. There are also some R and Python [helper scripts](https://github.com/Illumina/akt/tree/master/scripts). As far as I can tell these only appear to be used for testing so I'm not including them here, nor am I including R or Python as run deps.

3. Licenses: akt is GPL3, the vendored Eigen is MPL2; the vendored htslib is MIT. I don't see any other recipes with multiple licenses indicated, so I've kept just the akt license in here for now.

This version corresponds to the version in the manuscript, doi.org/10.1093/bioinformatics/btw576. I'll ask upstream for a more recent github release and push a follow-up PR updating the recipe accordingly.

### DIFF
--- a/recipes/akt/build.sh
+++ b/recipes/akt/build.sh
@@ -1,3 +1,7 @@
 #!/bin/bash
-make
+if [[ "$OSTYPE" == "darwin"* ]]; then
+    make no_omp
+else
+    make
+fi
 cp akt $PREFIX/bin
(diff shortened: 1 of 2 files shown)


stephenturner:
@daler -- long time user of bioconda, trying to figure out how to best contribute here. There's a new release of akt with a critical bug fix, described at illumina/akt#33, release 0.3.3 https://github.com/Illumina/akt/releases. What's the least-friction path to getting this updated version into bioconda?

PR #8694, opened 2018-04-22
update genometools to 1.5.10

notestaff:

### DIFF
--- a/recipes/genometools-genometools/meta.yaml
+++ b/recipes/genometools-genometools/meta.yaml
@@ -10,7 +10,7 @@ package:
 source:
   fn: {{ name }}-{{ version }}.tar.gz
   url: https://github.com/{{ name }}/{{ name }}/archive/v{{ version }}.tar.gz
-  sha1: {{ sha256 }}
+  sha256: {{ sha256 }}
 
 build:
   number: 0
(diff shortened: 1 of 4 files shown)


recipes/genometools-genometools/meta.yaml
```diff
@@ -1,28 +1,36 @@
...
+  number: 0
 
 requirements:
   build:
-    - gcc
+    - gcc  # [not osx]
+    - llvm  # [osx]
     - python
-    - pango
+    - pango  # [not osx]
```
bgruening: out of interest why does cairo not work for OSX?

notestaff: Cairo works, but we'd also need xquartz, which isn't currently in conda-forge:
https://github.com/genometools/genometools/blob/master/INSTALL#L51

In general, is this the right way to provide build variants (here one not requiring cairo/pango/X) of recipes, as subdirs of the main recipe, with the same version but different build string?  Conda also has 'features' but hardly any recipe seems to use that.

bgruening: features are really broken and should not be used. I'm fine with merging it, was just curious.

standage:
It would be awesome if AnnotationSketch were available on the OSX version of this package. It looks like @epruesse has a list [here](https://github.com/conda-forge/staged-recipes/issues/2587#issuecomment-293111507) of remaining obstacles to integrating XQuartz into conda forge. Before I start getting my hands dirty with this, do you know if any progress has been made on that list since then, @epruesse?

PR #8701, opened 2018-04-23
bioconductor-bsgenome.hsapiens.ucsc.hg19: add build script and remove old linker files

visze:


bgruening:
@visze can you explain these changes? The idea behind the post-link scripts is that we do not need to store the data in the package manager, which is a terrible idea imho.

visze:
Hi Björn,

You are absolutely right. I have the problem that the library does not work
like it is used right now. Conda installs it but the library is not
available under R. when I use it with the build script it works.

I will close the PR without merging tomorrow. Sorry I was just
experimenting if it goes through...



Björn Grüning <notifications@github.com> schrieb am Mo., 23. Apr. 2018,
18:55:

visze:
large data should not be stored in the package manager! Sry for the mistake

PR #8707, opened 2018-04-23
amptk v1.2.0: fix sha256 and add DOI

nextgenusfs:

Adding DOI to recipe.

### DIFF
--- a/recipes/amptk/meta.yaml
+++ b/recipes/amptk/meta.yaml
@@ -11,7 +11,7 @@ source:
   sha256: 54c109237fc06bda36bb788b5ae1069593171f4a36db037a078e6736e04292cc
 
 build:
-  number: 1
+  number: 0
 
 requirements:
   run:
--- a/recipes/amptk/meta.yaml
+++ b/recipes/amptk/meta.yaml
@@ -8,10 +8,10 @@ package:
 source:
   fn: {{ name }}-{{ version }}.tar.gz
   url: https://github.com/nextgenusfs/{{ name }}/archive/{{ version }}.tar.gz
-  sha256: ba8fb8ba0bb440863031613b7ce2074ab356abbdf5931280e582261f4a0e30cc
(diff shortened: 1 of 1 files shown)


recipes/amptk/meta.yaml
```diff
@@ -8,10 +8,10 @@ package:
 source:
   fn: {{ name }}-{{ version }}.tar.gz
   url: https://github.com/nextgenusfs/{{ name }}/archive/{{ version }}.tar.gz
-  sha256: ba8fb8ba0bb440863031613b7ce2074ab356abbdf5931280e582261f4a0e30cc
+  sha256: 54c109237fc06bda36bb788b5ae1069593171f4a36db037a078e6736e04292cc
```
bgruening: why has the checksum changed?

nextgenusfs: I had a slight typo in code.

recipes/amptk/meta.yaml
```diff
@@ -40,3 +40,7 @@ about:
   license_file: LICENSE.md
   summary: 'AMPtk: Amplicon tool kit for processing high throughput amplicon sequencing data.'
   license_family: BSD
+
```
bgruening: Thanks for adding the DOI. If you only add this you don't need to bump the build number. I'm wondering why you needed to change the checksum.

nextgenusfs: source changed slightly due to a fixed typo, if bumping the version number is better I can do that as well.

recipes/amptk/meta.yaml
bgruening: This is not good practice. You don't change the version, but in the results can change. Better is to do released and minor releases.

bgruening: If you decrease the build number again to 0, I will remove the previous version from anaconda.org

PR #8713, opened 2018-04-24
Updated reference to SearchGUI version 3.2.24 and PeptideShaker 1.16.20

CarlosHorro:

### DIFF
--- a/recipes/peptide-shaker/meta.yaml
+++ b/recipes/peptide-shaker/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "PeptideShaker" %}
-{% set version = "1.16.17" %}
+{% set version = "1.16.20" %}
 # Do not forget to update the version string in the peptide-shaker.py file
 
 about:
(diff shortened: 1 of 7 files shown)


tomkinsc:
Thanks for the update. Is there a particular reason to preserve recipes for older versions of SearchGUI? (especially [patch-increment](https://semver.org/) versions) Just as an FYI, recipes can generally be updated in-place since built packages for the older versions will be retained on [anaconda.org](https://anaconda.org/bioconda/searchgui/files). It can be helpful to retain a copy of an older recipe with notes about what's different between versions if there are significant algorithmic differences (ex. [older trinity](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/trinity/date.2011_11_26)), but otherwise making subdirectories for older versions is largely unnecessary. That said, there's minimal harm.

CarlosHorro:
Hi tomkinsc,

Well, I though it could be useful for keeping some trace of the different versions here also, although of course you are right and it's completely optional. 

Greetings,
Carlos

CarlosHorro:
Sorry for the last-minute commits, they were the last ones for this PR.

bgruening:
Thanks @CarlosHorro!

PR #8737, opened 2018-04-25
[WIP]  Use new compilers and CB3 syntax in BWA recipe

kyleabeauchamp:

This PR is an attempt at updating our BWA recipe to use the latest config settings from conda-build 3, as well as the updated compilers that they rely on.  

AFAIK the libgcc runtime dependency is no longer explicitly necessary, as it is implicitly added by the GCC compiler (https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html).

### DIFF
--- a/recipes/bwa/meta.yaml
+++ b/recipes/bwa/meta.yaml
@@ -11,16 +11,15 @@ source:
   sha256: {{ sha256 }}
 
 build:
-  number: 2
+  number: 3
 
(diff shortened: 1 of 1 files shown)


kyleabeauchamp:
FWIW, this change works on my OSX and linux machines, although I think there were some build errors on OSX for the older 0.6X BWA versions.

PR #8739, opened 2018-04-25
Add miRge2.0 for small RNAseq analysis

lpantano:


jenzopr:
The current build fails because the recent version of `biopython` does not contain `Bio.Align.Generics` any more (see http://biopython.org/DIST/docs/api/Bio.Align.MultipleSeqAlignment-class.html). 

In principal, pinning an older release of biopython (`biopython==1.68`) should work, but best would be that upstream authors (https://github.com/mhalushka/miRge) change their code to work with the recent release of `biopython`.

lpantano:
Thanks @jenzopr , working on that!

PR #8742, opened 2018-04-25
Recipes for R packages RAPIDR and PropCIs.

CarlosBorroto:


daler:
@CarlosBorroto thanks for this. In the future, would you please submit non-biology-specific packages like `r-propcis` to conda-forge? That way it can be used by folks in other fields.

CarlosBorroto:
Will do. Sorry, I only looked at the description of the package (my main interest on this pull was rapidr). I assumed it was mostly for biostatistics. I can now see this is a more general stats package.

PR #8745, opened 2018-04-26
Frubino - remove from blacklist

frubino:

Removes the mgkit package from the blacklist

### DIFF
--- a/recipes/abawaca/meta.yaml
+++ b/recipes/abawaca/meta.yaml
@@ -21,6 +21,10 @@ requirements:
   run:
     - libgcc    # [not osx]
 
+test:
+  commands:
+    - abawaca 2>&1 | grep Usage
+
 about:
   home: 'https://github.com/CK7/abawaca'
   license: "open source"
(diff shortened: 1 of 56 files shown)


frubino:
@johanneskoester to be honest, I don't understand why the requirement is failing... I added matplotlib to the build requirements as well and it seems too fail. Any idea?

frubino:
I imagine the problem is due to the import of pyplot in one of the modules, one of those that are irrelevant at the moment.

frubino:
Can you try and merge it, now?

frubino:
@bgruening, the reason the exception is raised is most likely the import of `pyplot`. In some cases, like virtual environments, the import gets a backend that is not correct. It's the case I have when building documentation and autodoc in sphinx has problem with that.

Right now I'm changing references to pyplot, but these are changer along more substantial ones in the package (I'm the developer). I'd prefer to get accustomed with the process here and let a user use the command line tools (which don't use any plot function) for now.

bgruening:
Thanks @frubino!

PR #1951, opened 2016-07-11
Software from ISMB 2016

johanneskoester:
* [x] I have read the guidelines above.
* [x] This PR adds a new recipe.
* [ ] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

Greetings from the ISMB-2016 @bioconda/all.
Today, I went through all open access papers in the proceedings of the ISMB 2016 and determined if they provide mature, installable software packages. If that was the case, I have created a bioconda package. I have to say that the results are rather disappointing. From 47 publications, only 4 can go directly into Bioconda. For the rest:

status | count
-------- | --------
pubs without software implementation | 7
based on the non-free MATLAB | 4
either unclear, no, or errorneous installation instructions | 3
invalid links | 4
without any download link | 1
collections of scripts without proper installer | 12
R packages that are not yet in CRAN or Bioc | 3
docker container | 1
demo only although the README promises a release before the ISMB | 1
available only upon request | 2
already available in bioconda | 2
web service | 4
build error | 1

Of course, R packages that are not yet in Bioconductor or CRAN could still be packaged (and will eventually). Also, the scripts can be packaged, although I would like to see people think about the user and provide installable software right away. Also, web services or docker images are not criticized here. The disappointment rather stems from the rest.
And here the details:

publication	|	status	|	bioconda
------------- | ----------------- | -----------------
Agostinelli et al.	|	web service	|	-
Yuan et al.	|	web service	|	-
Brougard et al.	|	no software	|	-
He et al.	|	no software	|	-
Deng et al.	|	scripts	|	-
Kraus et al.	|	available upon request	|	-
Haydarlou et al.	|	scripts	|	-
Peng et al.	|	available upon request	|	-
Schissler et al.	|	scripts	|	-
Zitnik et al.	|	ok	|	medusa-data-fusion
Glicksberg et al.	|	no software	|	-
Hu et al.	|	scripts	|	-
Zeng et al.	|	docker	|	-
Jo et al.	|	scripts	|	-
Sun et al.	|	not yet in CRAN or Bioc	|	-
Topa et al.	|	scripts	|	-
Duong et al.	|	no software	|	-
Lee et al.	|	software not available	|	-
Liu et al.	|	ok	|	debwt
Liu et al.	|	matlab	|	-
Srivastava et al.	|	already available	|	rapmap
Chikhi et al.	|	[build error](https://s3.amazonaws.com/archive.travis-ci.org/jobs/144154593/log.txt)	|	bcalm
[...]

PR #8750, opened 2018-04-26
proposal to add bcalm software

rchikhi:

Attempts to re-vive bcalm from https://github.com/bioconda/bioconda-recipes/pull/1951


bgruening:
@rchikhi why is this a re-add?
Looks good to me, do you think we should create a new package for the subpackage which you are cloning here?

rchikhi:
"re-add" because it was previously attempted to be added then removed in this PR https://github.com/bioconda/bioconda-recipes/pull/1951

what do you mean by "subpackage that your are cloning"? I'm not familiar with conda terminology

bgruening:
This one here: https://github.com/bioconda/bioconda-recipes/pull/8750/files#diff-a324caac488461278b34a2d28d3d3123R9 do we need a extra package for it?

rchikhi:
oh I see, thanks for the clarification.

Due to ancien design choices, bcalm and its build system expects that `gatb-core` is precisely at the directory where it is cloned. I'm afraid that many non-trivial things in the CMakeFile would need to be changed in gatb-core was moved as an independent package,

rchikhi:
changed title

rchikhi:
Hmm I wanted to update bioconda-recipes with most recent commits, probably did mess up the merge since the PR included all those commits. Will close the pull request and open another one

PR #8752, opened 2018-04-27
Added GATK 4.0.4.0 Updated Recipe

apeltzer:

Note that GATK team decided to change the name of the wrapper script from "gatk-launch" to gatk, so I also changed it in here. I suppose that should be fine with everyone?

### DIFF
--- a/recipes/gatk4/4.0.3.0/build.sh
+++ b/recipes/gatk4/4.0.3.0/build.sh
@@ -8,9 +8,9 @@ mkdir -p $PACKAGE_HOME
 
 sed -i.bak 's#!/usr/bin/env python#!/opt/anaconda1anaconda2anaconda3/bin/python#' gatk
 chmod +x gatk
-cp gatk ${PACKAGE_HOME}/gatk
+cp gatk ${PACKAGE_HOME}/gatk-launch
 cp gatk-*-local.jar $PACKAGE_HOME
 # Does not yet install spark jar to save space.
 # cp gatk-*-spark.jar $PACKAGE_HOME
 
-ln -s $PACKAGE_HOME/gatk $PREFIX/bin
+ln -s $PACKAGE_HOME/gatk-launch $PREFIX/bin
--- a/recipes/gatk4/4.0.3.0/meta.yaml
+++ b/recipes/gatk4/4.0.3.0/meta.yaml
@@ -1,4 +1,4 @@
-{% set version="4.0.4.0" %}
+{% set version="4.0.3.0" %}
 
 about:
   home: https://www.broadinstitute.org/gatk/
@@ -26,6 +26,6 @@ requirements:
 
 test:
   commands:
-    - gatk -h
-    - gatk --list
-    - gatk HaplotypeCaller --help
+    - gatk-launch -h
+    - gatk-launch --list
+    - gatk-launch HaplotypeCaller --help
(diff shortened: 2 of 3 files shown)


bgruening:
Thanks @apeltzer!

chapmanb:
Alexander -- thanks for keeping gatk up to date. Swapping over the script name from `gatk-launch` to `gatk` is breaking my usage of gatk4 and I wonder if we could swap it back. The major issue is that `gatk` is already used in bioconda for the `gatk` (gatk3) recipe so it's not possible to support both installs concurrently.

If we're opposed to `gatk-launch`, which I just kept for back compatibility, I'd propose `gatk4` to at least differentiate.

What do y'all think?

apeltzer:
Had a discussion about this recently with @marchoeppner and we were just thinking to follow the recommendation that GATK developers made recently to "go back to just `gatk`" instead of `gatk-launch`. I see your problem though, so maybe its worth thinking about differentiating between `gatk`(legacy, 3.X) and `gatk4` for version 4.X as you suggested?

ewels:
@chapmanb - out of curiosity, why do you want to have both versions loaded at the same time?

bgruening:
I would stick to whatever upstream provides. If GATK developers did not consider changing the name we should also not change it, if not absolutely needed. Please keep in mind that people are using these packages/containers in automatic environments and do not tweak their scripts to be "bioconda" conform.

chapmanb:
Alexander, Phil and Björn;
Thanks so much for the thoughts. I agree that it would be best to stick with what upstream provides, but would like some way to tackle back compatibility and also supporting GATK 3 and 4. In bcbio we support users who want GATK 3 for back compatibility with previous runs, so I see that being a use case for the forseeable future. I also would like some measure of back compatibility: the Broad team were using `gatk-launch` and then swapped over to just `gatk` which created this disconnect with upstream.

Some choices to help resolve this:

- Keep the GATK 4 script as `gatk` and then switch GATK 3 to `gatk3`. The downside is that might create confusion from users who expected `gatk` to be GATK 3 and disrupts users of both packages.
- Rename the GATK4 script to `gatk4`. This creates a new naming scheme in bioconda only and requires a transition for gatk4 users using gatk-launch.
- Keep GATK4 as `gatk-launch`. The downside is that this does not match upstream and is the previous naming scheme.

I favored the last in trying to create less disruption (I like back compatibility) but understand coordination with upstream as well. What do you think about the first suggestion if we want to stay consistent with script naming?

kyleabeauchamp:
We could also make a metapackage that just builds a renamed version for dual-GATK environments, so that the "primary" packages use the upstream naming.

apeltzer:
I like @kyleabeauchamp s suggestion for the extreme case of having both gatk versions installed. However, I dislike having a separate scheme for bioconda only, as this would enforce people to use exactly this scheme and does not line up with upstream. So I also tend to your option 1, although that requires changes for both GATK3 and GATK4 users - however its not as bad, as some might even switch to GATK4 anyways (and don't really care about that) or can willingly change to call gatk3 in case they definitely need to be compatible with GATK3 (which was anyways very difficult to package with bioconda if I'm not wrong?).

johanneskoester:
I think it would be preferable to use `gatk` for the new launcher and `gatk3` for the old launcher. THis way people can use both without problems in one environment (this will be rare), but it is straightforward to just use the latest.

johanneskoester:
At some point, we should also consider to rename the old recipe to `gatk3` and the new to `gatk`.

PR #8755, opened 2018-04-27
Build umis for py3

jenzopr:

Currently, umis is restricted to py27. This renders a usage within snakemake impossible. Since I was able to install umis==0.7.0 in py3 via pip, building for py3 in general should not be a problem.

### DIFF
--- a/recipes/umis/meta.yaml
+++ b/recipes/umis/meta.yaml
@@ -36,7 +36,7 @@ requirements:
 test:
   commands:
     # click requires a unicode locale
-    - LANG=en_US.utf8 umis --help
+    - LANG=en_US.UTF-8 umis --help
 
 about:
   home: https://github.com/vals/umis
--- a/recipes/umis/meta.yaml
+++ b/recipes/umis/meta.yaml
@@ -36,7 +36,7 @@ requirements:
 test:
   commands:
     # click requires a unicode locale
-    - LANG=C.UTF-8 umis --help
+    - LANG=en_US.utf8 umis --help
 
 about:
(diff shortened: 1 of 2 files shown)


jenzopr:
Problem is described here http://click.pocoo.org/5/python3/#python3-surrogates , as well as solutions.

daler:
@jenzopr maybe try `en_US.utf8`? 

Also it's likely you'll also need the extended container for the docker image to work. See the [`multiqc` recipe](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/multiqc/meta.yaml#L132) for how to set this up, and [bioconda troubleshooting docs](https://bioconda.github.io/troubleshooting.html#using-the-extended-image) for background.

jenzopr:
Thank you @daler ! I think your hints brought me onto the right track 😃

recipes/umis/meta.yaml
```diff
@@ -37,9 +36,15 @@ requirements:
 test:
   commands:
     # click requires a unicode locale
-    - LANG=C.UTF-8 umis --help
+    - LANG=en_US.UTF-8 umis --help
```
jenzopr: Maybe exporting `LANG=en_US.UTF-8` is not even needed for the extended container.

bgruening: Yes, I don't think it is needed.

jenzopr:
Thanks, Björn. If no further comments arise, I'll merge 😃

bgruening:
Go for it - and thanks a lot!

PR #8763, opened 2018-04-27
CTAT lncrna

asmariyaz23:
* [Yes ] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [Yes ] This PR adds a new recipe.
* [Yes ] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).
* [No ] This PR updates an existing recipe.
* [No ] This PR does something else (explain below).

### DIFF
--- a/recipes/ctat-lncrna/build.sh
+++ b/recipes/ctat-lncrna/build.sh
@@ -10,25 +10,13 @@ mkdir -p $ctat_lncrna_INSTALL_PATH
 # The copy of everything does not work, because the build process puts a conda_build script
 # into the same directory. That file has hard coded paths in it which are not able to be
 # corrected, so the build fails.
-<<<<<<< HEAD
 cp -R $SRC_DIR/LICENSE.txt $SRC_DIR/lncrna $SRC_DIR/SciEDPipeR $ctat_lncrna_INSTALL_PATH
 chmod a+x $ctat_lncrna_INSTALL_PATH/lncrna
-=======
-cp -R $SRC_DIR/LICENSE.txt $SRC_DIR/lncrna_discovery.py $SRC_DIR/SciEDPipeR $SRC_DIR/slncky $ctat_lncrna_INSTALL_PATH
-cp -R $SRC_DIR/LICENSE.txt $SRC_DIR/lncrna_discovery.py $SRC_DIR/SciEDPipeR $SRC_DIR/slncky $PREFIX/bin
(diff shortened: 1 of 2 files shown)


recipes/ctat-lncrna/build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# ctat_lncrna_DIR_NAME="$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM"
+# The above is an error. The following is correct.
+ctat_lncrna_DIR_NAME="$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM"
+ctat_lncrna_INSTALL_PATH="$PREFIX/share/$ctat_lncrna_DIR_NAME"
+# Make the install directory and move the ctat_lncrna files to that location.
+mkdir -p $PREFIX/bin
+mkdir -p $ctat_lncrna_INSTALL_PATH
+#copy to INSTALL_PATH
+cp -R $SRC_DIR/LICENSE.txt $SRC_DIR/lncrna $SRC_DIR/SciEDPipeR $ctat_lncrna_INSTALL_PATH
```
bgruening: why do you copy the license as well? Conda will do this for you if you have specified it in the meta.yaml, which you have done :)
Why not copying everything into PREFIX/bin directly?

recipes/ctat-lncrna/meta.yaml
```diff
@@ -0,0 +1,37 @@
+{% set name = "ctat-lncrna" %}
+{% set version = "2.0" %}
+{% set build = "1" %}
```
bgruening: should be 0 if you have a new version

recipes/ctat-lncrna/build.sh
asmariyaz23: The reason for not copying into bin directly is as follows:
Most of the CTAT tools use different versions of SciEDPipeR, so if a user installs 2 of CTAT tools then version conflicts may occur hence my attempt at keeping everything in install path and invoking from there.

bgruening:
Ok, thanks!

recipes/ctat-lncrna/meta.yaml
asmariyaz23: I am making a new build including the changes you suggested.

PR #8792, opened 2018-05-02
added swipe

MathiasHaudgaard:


MathiasHaudgaard:
@bgruening the package fails due to the -h command returns 1. We have experienced more packages with the same issue. What should we do about it? :shipit:

bgruening:
@MathiasHaudgaard you run a command like this `foo --help | grep "some string from your help massage"`. This way the return code from grep is returned if successful.

PR #8813, opened 2018-05-03
Mikado

lucventurini:


recipes/mikado/meta.yaml
```diff
@@ -1,24 +1,23 @@
 {% set name = "mikado" %}
-{% set version = "1.1.1" %}
+{% set version = "1.2" %}
```
druvus: Why is version set to 1.2 and not 1.2.1?

lucventurini: My mistake, I will update it.

PR #8822, opened 2018-05-04
Ammended syntax for setting permission as "+mode" is now deprecated.

slacalle:

### DIFF
--- a/recipes/canu/build.sh
+++ b/recipes/canu/build.sh
@@ -6,5 +6,6 @@ mkdir -p "$PREFIX/libexec" "$PREFIX/bin"
 ( cd src; make TARGET_DIR=$PREFIX/libexec )
 
 # Link all executable files to bin
+find $PREFIX/libexec -type f -perm /111 -exec ln -s {} $PREFIX/bin \;
+#Added the above syntax change for setting permission bits as the "+mode" is now deprecated. Left it below just in case build env is different.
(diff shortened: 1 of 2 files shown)


daler:
@slacalle thanks, can you clarify the issue? Is this a change to how `find` works? What version is the change in? Does the previous syntax return an error?

By the way, you'll have to bump the build number to get trigger a new build.

slacalle:
Hi @daler , I was attempting to build the canu package with the default template and was not able to. This was due to the tests failing as it could not find the "canu" executable. Upon further investigation, it looks like the last command in the build script did not run correctly. When running manually, the following message appears:

"find: warning: you are using `-perm +MODE'.  The interpretation of `-perm +omode' changed in findutils-4.5.11.  The syntax `-perm +omode' was removed in findutils-4.5.12, in favour of `-perm /omode'."

more on this on the find manpages.

This is with findutils.x86_64 version 4.5.11-5.el7

daler:
@slacalle thanks. The docker container we're using, `bioconda/bioconda-utils-build-env`, has `find 4.4.2` in it. If you test locally using the "bootstrap" method (https://bioconda.github.io/contribute-a-recipe.html#test-locally) it works -- so I don't think there's a need for this PR.

Of course, once we migrate the build system to something later than CentOS 6 we'll need to apply this fix. But that likely won't be for quite some time.

slacalle:
@daler, ah now I understand. That makes sense. I will close this pull request for now then.

Thanks,

PR #8824, opened 2018-05-04
perl-vcf-vcftools: install perl scripts & remove unnecessary dependencies

nathanweeks:

This adds the perl scripts documented at: https://vcftools.github.io/perl_module.html
Note that the tests in src/perl/test.t apparently aren't being run. That would require some additional build dependencies (and developer time!)


nathanweeks:
Switching from single to double quotes in the test commands fixed the odd error. Is double-quotes-instead-of-single-quotes requirement for test commands documented somewhere? If not, it would be helpful...

PR #8825, opened 2018-05-04
Mikado

lucventurini:


lucventurini:
Hi @daler , sorry to disturb you again regarding my package, but something is going awry. The new version seems to build nicely (no problems with the setup or building the Cython extensions), yet, during the tests the basic "import" fails telling CircleCI that the library cannot be found.

I basically copied and updated the previous recipe, so I am stumped. I imagine that there might be a problem with the classic "python setup.py install " etc., but I cannot figure it out.

Thank you again for your time and patience.

daler:
@lucventurini I think the issue might be the pinning of python >=3.5. That worked locally; let's see how these tests go.

daler:
That was it, thanks!

PR #8830, opened 2018-05-06
Add recipe for cdna_cupcake

npavlovikj:

### DIFF
--- a/recipes/cdna_cupcake/build.sh
+++ b/recipes/cdna_cupcake/build.sh
@@ -1,0 +1,16 @@
+#!/bin/sh
+
+$PYTHON setup.py build
+$PYTHON setup.py install
+
+cp annotation/*.py ${PREFIX}/bin/
+cp bacteria/*.py ${PREFIX}/bin/
+cp phasing/[a-zA-Z]*.py ${PREFIX}/bin/
+cp phasing/io/[a-zA-Z]*.py ${PREFIX}/bin/
+cp post_isoseq_cluster/*.py ${PREFIX}/bin/
+cp SequelQC/*.py ${PREFIX}/bin
+cp sequence/*.py ${PREFIX}/bin/
+cp simulate/*.py ${PREFIX}/bin/
+cp targeted/* ${PREFIX}/bin/
+
+chmod +x ${PREFIX}/bin/*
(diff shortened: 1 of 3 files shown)


recipes/cdna_cupcake/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+  build:
+    - python
+    - biopython
+    - bx-python <=0.7.3
+
+  run:
+    - python
+    - biopython
+    - bx-python <=0.7.3
+    - r
```
bgruening: r is probably not needed here

recipes/cdna_cupcake/build.sh
```diff
@@ -0,0 +1,16 @@
...
+
+$PYTHON setup.py build
+$PYTHON setup.py install
+
+cp annotation/*.py ${PREFIX}/bin/
+cp bacteria/*.py ${PREFIX}/bin/
+cp phasing/[a-zA-Z]*.py ${PREFIX}/bin/
+cp phasing/io/[a-zA-Z]*.py ${PREFIX}/bin/
+cp post_isoseq_cluster/*.py ${PREFIX}/bin/
+cp SequelQC/*.py ${PREFIX}/bin
```
bgruening: Its in general not good to copy everything in /bin. Is this really needed, are all binaries? Why is setup.py install not taking care of this?

recipes/cdna_cupcake/meta.yaml
npavlovikj: Thanks for the feedback @bgruening ! There are some R scripts in the package, that's why R is there.

recipes/cdna_cupcake/build.sh
npavlovikj: cDNA_Cupcake is a collection of various scripts, organized in different sub-directories. That's why I am copying all of these individual scripts into bin/. Only one set of scripts requires compilation and uses setup.py. If there is a better way of handling this, please let me know.

recipes/cdna_cupcake/meta.yaml
bgruening: You already included r-base, which should be enough in most cases so `r` can be remove.

recipes/cdna_cupcake/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    commands:
+      - parse_matchAnnot.py -h
+      - make_file_for_subsampling_from_collapsed.py -h
+      - subsample.py -h
+      - get_seq_stats.py -h
+      - revcomp.py -h
+      - fa2fq.py -h
+      - fq2fa.py -h
+      - sort_fasta_by_len.py -h
+      - get_seqs_from_list.py -h
```
bgruening: can you add the r-scripts here as well?

PR #8834, opened 2018-05-06
Add files via upload

luketerry:

### DIFF
--- a/mirge/meta.yaml
+++ b/mirge/meta.yaml
@@ -9,9 +9,9 @@ about:
 
 source:
   #git_url: https://github.com/mhalushka/miRge.git
-  fn: miRge2.0.1.tar.gz
-  url: https://github.com/mhalushka/miRge/archive/miRge2.0.1.tar.gz
-  sha256: 88c1023cb005a2c01524b94276303cade5526b67c6652ec030b7c519addf36d3 
+  fn: miRge2.0.2.tar.gz
+  url: https://github.com/mhalushka/miRge/archive/miRge2.0.2.tar.gz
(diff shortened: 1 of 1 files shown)


daler:
@luketerry actually I just realized this recipe is at the top level of the bioconda repo. I will create a PR to revert this change.

luketerry:
@daler Hi, actually I just updated the mirge and the url in the source part of meta.yaml is changed. I found that mirge directory under  https://github.com/bioconda/bioconda-recipes/tree/master/mirge was removed. Can you revert it? Thank you.

daler:
@luketerry it looks like you initially added the recipe at the top level in https://github.com/bioconda/bioconda-recipes/pull/8818 (which was not caught by any reviewers) and then subsequently updated that in this PR. Neither recipe was picked up by the build system since recipes are expected to be in the `recipes` directory.

In general, new recipes go in the `recipes` directory. However in the specific case of `miRge`, @lpantano is (was?) working on it over in https://github.com/bioconda/bioconda-recipes/pull/8739. It appears that it expects an older version of Biopython. As noted there, you can try pinning the biopython version to 1.68 to see if that will pass tests.

luketerry:
@daler  I see. I  cooperated with @lpantano on the integration of mirge into bcbio. Initially, I set biopython ==1.68 then I revised the code to make it compatible with higher biopython version (1.71) according to @lpantano's request. Now mirge works well if biopython's version is 1.71. So in current meta.yaml, I set biopython >=1.68.
If I need to upload the current version of mirge, what should I do now? 
Thank you

luketerry:
@lpantano OK. I find mirge is available in biodonda and can be installed by conda now. 
Thank you.

PR #8843, opened 2018-05-07
Fix Genomedata recipe to use updated PyTables and fix dependency conflicts

EricR86:

Genomedata relies on a fix to the issue found here:
https://github.com/conda-forge/pytables-feedstock/issues/29
Fixed in this PR:
https://github.com/conda-forge/pytables-feedstock/pull/30


EricR86:
Not sure what is happening with the build machine in this case. The link that build says it cannot download from works for me when I copy it.

daler:
@EricR86 PyPI is having network issues today https://status.python.org; this failure it's likely an intermittent issue outside our control.

daler:
@EricR86 I restarted the tests, they now pass. Looks like you needed to exclude the hdf5 from this recipe to avoid conflicting with the [pinned version in conda-forge](https://github.com/tomkooij/pytables-feedstock/blob/ad09360603e9ae9db3bdf9d5bc045acef60f9ead/recipe/meta.yaml#L32), is that right?

EricR86:
I believe so.

More accurately I essentially need to keep up to date with the version PyTables uses since older conda builds (and releases) will cause Genomedata to segfault. The reason for that is in the issues I linked originally.

daler:
Got it, thanks!

PR #8855, opened 2018-05-08
secapr update

tobiashofmann88:

### DIFF
--- a/recipes/secapr/meta.yaml
+++ b/recipes/secapr/meta.yaml
@@ -42,7 +42,6 @@ requirements:
     - pandas ==0.22.0
     - numpy ==1.14
     - cogent ==1.5.3
-    - openjdk ==7.0.161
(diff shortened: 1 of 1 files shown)


tobiashofmann88:
Does anybody have an idea what the problem with my recipe is? It seems to almost build but then exits with this error "Error: bad character '=' in build/string: np114py27r==3.4.3_0", which I think is the error that makes the build crash. Not sure what the np114py27r package is, it's not speicifed in my recipe but most be part of one of the dependencies.
Any help is much appreciated!
Best,
Tobi

marcelm:
@bioconda/core Could anyone take a look at this? The error message is
```
Error: bad character '=' in build/string: np114py27r==3.4.3_0
```
I wonder if it works locally for @tobiashofmann88 because he is likely using a more recent `conda-build` version.

PR #8859, opened 2018-05-09
added esimsa

MathiasHaudgaard:


MathiasHaudgaard:
@bgruening the package didn't have any doi so I added the pmid instead. Is that fine?

bgruening:
Yes I think this is ok. We do not yet support it but the annotation is important. Thanks!

PR #8870, opened 2018-05-09
adding recipe for QCF software

pblischak:

### DIFF
--- a/recipes/QCF/build.sh
+++ b/recipes/QCF/build.sh
@@ -2,4 +2,5 @@
 
 cd $SRC_DIR
 make
-make install
+cp qcf ${PREFIX}/bin/qcf
+chmod +x ${PREFIX}/bin/qcf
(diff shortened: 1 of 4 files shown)


recipes/QCF/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+cd $SRC_DIR
```
epruesse: should not be necessary

pblischak: Thanks for the comments!

I have tried to do the rebase procedure several times without success (I'm not really sure I fully understand how it works). I'm not sure what the issue is, but if you have any general suggestions I would very much appreciate it. Here is what I tried:

```
git checkout QCF
git rebase master
```

I have a few changes to make to my code as well, so would it be easier to make those changes and start a new submission?

daler:

PR #8876, opened 2018-05-10
secapr update

tobiashofmann88:

### DIFF
--- a/recipes/secapr/meta.yaml
+++ b/recipes/secapr/meta.yaml
@@ -25,7 +25,6 @@ requirements:
 
   run:
     - python ==2.7.14
-    - r ==3.4.3
     - biopython ==1.68
     - emboss ==6.5.7
     - bcftools ==1.8
@@ -37,13 +36,11 @@ requirements:
     - mafft ==7.130
     - muscle ==3.8.31
     - trimmomatic ==0.32
-    - trinity ==2.0.6
     - abyss ==1.3.7
     - fastqc >=0.11,<=0.12
     - pandas ==0.22.0
     - numpy ==1.14
     - cogent ==1.5.3
-    - openjdk ==7.0.161
     
 test:
   imports:
--- a/recipes/secapr/meta.yaml
+++ b/recipes/secapr/meta.yaml
@@ -25,9 +25,10 @@ requirements:
     - setuptools
     - biopython
     - pandas
+    - r
 
   run:
-    - python ==2.7.14
+    - python
     - biopython ==1.68
     - emboss ==6.5.7
     - bcftools ==1.8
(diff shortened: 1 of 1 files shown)


tobiashofmann88:
Not sure what is going wrong, by now I tried everything I can come up with.
The current error is the following:
> Can't build /opt/recipe due to unsatisfiable dependencies:
> The following specifications were found to be in conflict:
>   - secapr 1.1.7 np114py27r3.4.1_0

marcelm:
Only guessing now, but try to remove the version specification from `numpy`.

Also, specify the same versions for `pandas` and `biopython` in both the `run:` and `build:` requirements sections.

Do only one of the above at a time.

tobiashofmann88:
okay its working now!! Thanks for your help @marcelm!

tobiashofmann88:
@bioconda/core: I think this recipe is now ready to be merged. Thanks!!

recipes/secapr/meta.yaml
```diff
@@ -22,27 +25,28 @@ requirements:
...
+    - samtools ==0.1.19
+    - picard ==1.126
+    - seqtk >=1.0.82,<=1.2
+    - bwa >=0.7
+    - lastz ==1.0.2
+    - mafft >=7.2
+    - muscle ==3.8.31
+    - trimmomatic ==0.33
+    - abyss ==1.5.2
+    - fastqc >=0.11,<=0.12
```
marcelm: You can express this as `fastqc 0.11*`

recipes/secapr/meta.yaml
```diff
@@ -22,27 +25,28 @@ requirements:
...
-    - muscle
-    - cogent
-    - trimmomatic
-    - trinity
-    - abyss
-    - fastqc
-    - pandas
-    - numpy 
-
+    - biopython ==1.68
```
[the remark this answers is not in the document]
marcelm: To Tobias defense: Yes, these are the versions that `secapr` has been tested with. I think this needs to be allowed for applications.

recipes/secapr/meta.yaml
mbargull: Yes, in the general case this should even be either `fastqc 0.11.*` or `fastqc >=0.11,<0.12.0a0` -- if (hypothetically) there where an alpha version `0.12.a1`, that one would be considered `<0.12`.

tobiashofmann88:
I implemented all suggested changes and I hope it's good to go now!
Thanks a lot for your help!

marcelm:
Tobias wants to use this package at a workshop where he’s teaching today, so I’m taking the liberty of merging this now as it seems all suggestions have been addressed. There can always be another PR for any remaining fixes.

PR #8891, opened 2018-05-11
fix SIMD instruction set to SSE2

kain88-de:

This fixes the instruction set of GROMACS to SSE2. Those settings are very conservative to allow the package to run on almost any x86_64 system after 2003. The next available option SSE4.1 is first supported by Intel in ~2007 and AMD ~2012.

This fixes some 'illegal instruction set' error we have noticed for GromacsWrapper.

https://github.com/Becksteinlab/GromacsWrapper/pull/130#issuecomment-388430601


epruesse:
>  Those settings are very conservative to allow the package to run on almost any x86_64 system after 2003. 

Actually, SSE2 will run on any x86_64 system as SSE2 is part of the x86_64 spec.

PR #8898, opened 2018-05-13
Migrate n

tiagoantao:

This PR is a new recipe for migrate-n. Can someone @bioconda/core have a look please?


recipes/migrate-n/run_test.py
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/python
+
+import os
+import sys
+
+if os.system('migrate-n -h') != 65280:
```
bgruening: can you migrate this into the yaml file and do something like this `migrate-n -h | grep "Foo"` ... with Foo a text included in the help message. This way the error code is taken from the grep step and it should be 0.

recipes/migrate-n/meta.yaml
```diff
@@ -0,0 +1,31 @@
+package:
+  name: migrate-n
+  version: "3.6.11"
```
bgruening: you could use jinja templates for the version string

recipes/migrate-n/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+build:
+  number: 0
+  skip: True  # [not linux]
+
+source:
+  fn: migrate-3.6.11.src.tar.gz
+  url: http://popgen.sc.fsu.edu/currentversions/migrate-3.6.11.src.tar.gz
+  md5: 4e85552db1e73dc394042176e710508c
```
bgruening: sha256 is possible.

recipes/migrate-n/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+    - zlib {{ CONDA_ZLIB }}*
+
+about:
+  home: http://popgen.sc.fsu.edu/Migrate/Migrate-n.html
+  license: MIT
+  summary: Population Genetics - Panmixia and Migration detection
+
+extra:
+  identifiers:
+    - doi:10.1093/bioinformatics/btu033
```
bgruening: Yeah! Thanks for adding the DOI!

recipes/migrate-n/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+
+cd src
+
+./configure
+sed -i 's/#include <xlocale.h>//' data.c
```
bgruening: `-i.bak` if you want to be compatible with OSX

tiagoantao:
Thanks @bgruening , as usual you are amazing. Could you have another look to see if it looks OK?

tiagoantao:
Thanks for the help @bgruening

PR #8919, opened 2018-05-14
MOB_suite

kbessonov1984:
Please add MOB-suite recipe to the bioconda channel. MOB-suite is a set of software tools for clustering, reconstruction and typing of plasmids from draft assemblies.


kbessonov1984:
This is a recipe to compile our plasmid tool. It works fine on my privite anaconda cloud and also using docker image produced by bioconda-utils. For some reason the test phase does not pass and mob_suite module can not be imported.

kbessonov1984:
Could you please merge the c4eb46f commit to the main branch? Finally all tests passed. The issue was the biopython 1.70 version strict requirement and runner was installing 1.68. Thus, the tests did not pass ...

bootstrap.py
```diff
@@ -1,5 +1,3 @@
-#!/usr/bin/env python
```
daler: Was there a reason you needed to remove this line?

recipes/mob_suite/meta.yaml
```diff
@@ -0,0 +1,39 @@
+package:
+    name: mob_suite
+    version: 1.4.1
+
+build:
+    number: 1
```
daler: Build number should start at zero

recipes/mob_suite/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    skip: True # [py27 or py34 or py35]
+    script: python setup.py install --single-version-externally-managed --record=record.txt
+    string: "py{{CONDA_PY}}"
```
daler: Don't need the string here; python versions are already handled in the build strings.

kbessonov1984: OK, will remove it

recipes/mob_suite/meta.yaml
kbessonov1984: OK, thank you. Will change to zero

bootstrap.py
kbessonov1984: It was my mistake. When I was installing the development tools I had several issues. I've thought that wrong python installation was used.

PR #8940, opened 2018-05-16
Add r-zerone package

mvdbeek:


recipes/r-zerone/meta.yaml
```diff
@@ -0,0 +1,39 @@
+{% set version = 'v1.0' %}
```
rvalieris: could you just leave the version as '1.0' ? these extra letters can confuse the version ordering of conda.

PR #8952, opened 2018-05-17
Use mktemp instead of tr ... < /dev/urandom for osx compatibility

nathanweeks:
* [ X] This PR does something else (explain below).

On my macOS 10.11.6 workstation, `./bootstrap.py --no-docker /tmp/miniconda` hangs during the `tr` command in .circleci/common.sh, with the `tr` command continuing to consume CPU cycles without terminating due to the SIGPIPE that should result from the subsequent `head -c10`. Executing the command manually at the command line results in an error:

```
$ tr -dc a-z < /dev/urandom | head -c10
tr: Illegal byte sequence
```

A workaround is noted here:
https://unix.stackexchange.com/questions/45404/why-cant-tr-read-from-dev-urandom-on-osx

However, even implementing the LC_ALL/LC_TYPE workaround, `tr` still seems to not terminate. Commenting-out the `set -e` and `set -u` allows the script to continue, so it must be some subtle interaction there.

One alternate approach is to generate 10 pseudorandom alphanumeric characters using `mktemp -u XXXXXXXXXX`, as `mktemp` is used elsewhere in the script and so is already assumed to be on the target system.

[edit: originally specified `mktemp XXXXXXXXXX`, but that leaves a temporary _file_ behind, so adding the `-u` option]

### DIFF
--- a/.circleci/setup.sh
+++ b/.circleci/setup.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 set -e
-
 [[ -z $WORKSPACE ]] && WORKSPACE=`pwd`
 [[ -z $BOOTSTRAP ]] && BOOTSTRAP=false
 [[ -z $BASH_ENV ]] && BASH_ENV=`mktemp`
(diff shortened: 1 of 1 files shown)


mvdbeek:
I had the same issue. Another approach would be `python3 -c "import os; import codecs; print(int(codecs.encode(os.urandom(8),'hex'), 16))"`

bgruening:
@daler is this ok for you as well?

mbargull:
Thanks!

PR #8983, opened 2018-05-21
R ecodist

nick-youngblut:

### DIFF
--- a/recipes/bioconductor-experimenthub/build.sh
+++ b/recipes/bioconductor-experimenthub/build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+mv DESCRIPTION DESCRIPTION.old
+grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
+$R CMD INSTALL --build .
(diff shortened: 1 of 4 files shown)


nick-youngblut:
I'm not sure why I'm getting the error: `jinja2.exceptions.UndefinedError: 'compiler' is undefined`. I created the recipe with `conda skeleton cran ecodist` (conda-build 3.10.4-py36_0). Shouldn't the compiler be automatically defined? If not, I'm not sure which compiler to use. I've been searching through other recipes to find a similar example (eg., a CRAN recipe using `{{ compiler('c') }}`), but I haven't been able to find one. 

Any ideas?

bgruening:
@nick-youngblut we are in a trasition to the new build infrastructure and this recipe relies on it. Either you can wait a week or two, or you need to crate a recipe following the conda-build 2 way.

nick-youngblut:
Thanks @bgruening for the clarification! From reading the docs, I thought that "use conda-build v2" only applied to pypi packages.

PR #8997, opened 2018-05-22
add r-ade4

nick-youngblut:

### DIFF
--- a/recipes/r-ade4/build.sh
+++ b/recipes/r-ade4/build.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+mv DESCRIPTION DESCRIPTION.old
+grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
+
+$R CMD INSTALL --build .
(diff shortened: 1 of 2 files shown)


daler:
@nick-youngblut this one should have gone into conda-forge I believe since it's not bio-specific.

In fact, now that I check it's already in conda-forge . . . the linting should have caught this I think.

nick-youngblut:
Sorry about that! I just saw that it was a dependency for another package that wasn't available on bioconda, so I added it. I should have checked conda-forge. Can I remove ade4?

daler:
I will remove the `main` tag from channel (admin has to do that); would you submit a PR that removes the recipe? Thanks.

nick-youngblut:
OK. I'll do that right now. Just as an FYI, the `r-adephylo` recipe will be broken once r-ade4 is removed. I see that conda-forge is lacking the `r-ade4` 1.7_10, which `r-adephylo` recipe lists as a dependency. I'll try to update r-ade4 on conda-forge ASAP. Sorry again for making this an issue

PR #9019, opened 2018-05-23
updated SPAdes to 3.12.0

notestaff:

### DIFF
--- a/recipes/spades/meta.yaml
+++ b/recipes/spades/meta.yaml
@@ -36,8 +36,7 @@ about:
   home: http://cab.spbu.ru/software/spades/
   license: GPLv2
   license_family: GPL
-  license_file: LICENSE  # [linux]
-  license_file: share/spades/LICENSE  # [osx]
+  license_file: share/spades/LICENSE
   summary: |
     SPAdes (St. Petersburg genome assembler) is intended for both standard isolates and single-cell MDA bacteria assemblies.
   dev_url: https://github.com/ablab/spades
--- a/recipes/spades/meta.yaml
+++ b/recipes/spades/meta.yaml
@@ -8,10 +8,10 @@ source:
   # We currently cannot build SPAdes from source, because it requires g++ 5.3.1 or later,
   # which bioconda does not yet support.  Therefore, we fetch the statically compiled binaries.
   fn: SPAdes-{{ version }}-Darwin.tar.gz  # [osx]
-  url: http://cab.spbu.ru/files/release{{ version }}/SPAdes-{{ version }}-Darwin.tar.gz  # [osx]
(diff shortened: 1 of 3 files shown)


notestaff:
@andrewprzh @snurk @asl spades developers -- if have time, maybe look over the recipe for installing spades through bioconda?   thanks a lot!

notestaff:
@SilasK here is another attempt at updating SPAdes to 3.12.0

SilasK:
Thank you, I should have asked you before starting my own PR. @notestaff  For the sha256 there is no simpler way than downloading the package and calculate the sha sum, isn't it?

nick-youngblut:
You can use the following for calculating the sha256 key: `wget -O- https://github.com/zdk123/SpiecEasi/archive/v0.1.4.tar.gz | shasum -a 256`

recipes/spades/meta.yaml
```diff
@@ -1,33 +1,25 @@
...
 
 package:
   name: spades
   version: {{ version }}
 
 source:
+  # We currently cannot build SPAdes from source, because it requires g++ 5.3.1 or later,
+  # which bioconda does not yet support.  Therefore, we fetch the statically compiled binaries.
   fn: SPAdes-{{ version }}-Darwin.tar.gz  # [osx]
   url: http://cab.spbu.ru/files/release{{ version }}/SPAdes-{{ version }}-Darwin.tar.gz  # [osx]
```
asl: You may want to download stuff from github as well. Is there a way to provide several URLs if the primary one is unavailable for one reason or another?

notestaff: @asl There is a way to fetch additional URLs, however only the primary one gets persistently cached by the bioconda setup.   What would the other URLs be?
Typically bioconda recipe URLs point to a tagged release file hosted on github, so that would be best.

notestaff: @asl thanks, changed the URL to github.

The URL in the recipe is normally only accessed once, when the bioconda binary package is built; the package is then uploaded to https://anaconda.org/bioconda/spades/files , and bioconda users fetch it from there.

recipes/spades/meta.yaml
```diff
@@ -38,22 +30,22 @@ test:
...
+  doc_url: http://cab.spbu.ru/files/release{{version}}/manual.html
 extra:
-  doi: 10.1089/cmb.2012.0021
   recipe-maintainers:
     - druvus
     - notestaff
   identifiers:
     - biotools:Spades
     - doi:10.1089/cmb.2012.0021
+    - doi:10.1101/gr.213959.116
```
bgruening: This is metaspades isn't it?

notestaff: @bgruening Yes, which is part of the spades package.

bgruening:
Thanks!

PR #9022, opened 2018-05-24
update KMC kmer counter to 3.1.0

notestaff:


notestaff:
@marekkokot here is a recipe for installing KMC through bioconda, maybe look it over?

marekkokot:
@notestaff 
I am not very familiar with bioconda, so I am not sure how I may if this pull request works, so I only looked at your commits and I think it looks OK.
From what I understand now conda install kmc will just download binaries from main kmc repository instead of compiling it from scratch, is that right? 
For me this is OK, but I have seen that in previous version there were also no-asmlib version because of its license. 
If you would like to make it possible to use KMC without asmlib (which seems to be reasonable) and without the need to compile it by bioconda pipeline (I don't know how it works) I may compile no-asmlib version of kmc 3.1.0 and add it to release on main KMC directory. Let me know.

notestaff:
@marekkokot yes for now the conda package will just include your static binaries.  Later when conda supports c++14 I'll change the recipe to compile from source.

Since asmlib is distributed under GPL, like kmc, I don't think a noasmlib version is required.

notestaff:
@marekkokot btw maybe update the page at 

http://sun.aei.polsl.pl/REFRESH/index.php?page=projects&project=kmc&subpage=download

to 3.1.0

marekkokot:
@notestaff I am not sure about this license compatibility (check https://github.com/refresh-bio/KMC/issues/44). I hoped that I will be able to drop both dependencies (asmlib and vcl) but there were some unexpected behaviors, and now it seems that I will not be able to do it quickly. On the other hand even if I would compile KMC without asmlib, there is still vcl that is used to choose the appropriate version of out sorting routine depending on the machine on which KMC was run. 

I just wanted to point this out to make you aware how things are, but I hope it is not that much relevant for bioconda. 

The page is updated by my supervisor, if he will not update it in the next week I will remind him, thanks.

mbargull:
IMO, distributing KMC builds that use asmlib and/or vectorclass is not possible without license violations. KMC is licensed under `GPL-2.0-only`, the other libraries are under `GPL-3.0-or-later`\*. Those are not compatible, see https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility.
The easiest thing would be to licence KMC under `GPL-2.0-or-later` which would mean it's `GPL-2.0-or-later` if not linked to asmlib/vectorclass and `GPL-3.0-or-later` otherwise. @marekkokot, would such a licence update be possible?
<sub>* The additional non-commercial terms from asmlib/vectorclass are just void and should play any role, IMHO, see https://github.com/refresh-bio/KMC/issues/44.</sub>

@notestaff: Does compiling from source not work anymore? Are the patches I wrote earlier (esp. `0003-kmc-c++11-compat.patch`) not sufficient anymore?

notestaff:
@mbargull sorry, I haven't tried to make the patches work on the new version.   Maybe you can try?   I just worry that without a full understanding of KMC code, I might break something and not know it (especially since it does not currently have a publicly available regression test suite).   I hope it won't be too long before bioconda supports C++14, there are other great packages that could then be compiled.

PR #9027, opened 2018-05-24
Bumped perl-set-intervaltree version

micknudsen:


rvalieris:
ah looks like a dependency is missing, ExtUtils::CppGuess.

you should be able to use "conda skeleton cpan  ExtUtils::CppGuess" to include it.

micknudsen:
Does ExtUtils::CppGuess belong in bioconda? Maybe conda-forge would be more appropriate?

rvalieris:
yes it would, but that would require migrating all the dependencies as well, we do have plans to migrate these packages at some point but no timeline yet.
I think its ok to include this package now so that star-fusion can be used.

rvalieris:
you can now add the dependency `perl-extutils-cppguess` on build and run requirements.

PR #9041, opened 2018-05-25
Migrate to conda-build 3

johanneskoester:
This PR rebuilds recipes with conda-build 3, and thereby also tests our update bioconda-utils.

# TODO before merging!
* [x] Remove all `CONDA_*` variables from recipes.
* [x] Remove build strings (conda-build 3 automatically generates hashes from the dependencies).
* [ ] Move build dependencies that are also run dependencies into the host section.
* [ ] Replace `{{ compiler('c') }}` with `{{ compiler('cxx') }}` for recipes that need C++.
* [ ] Use `noarch: python` and `noarch: generic` whenever possible.
* [ ] Bump build number.
* [ ] Wait for all recipes to build successfully.
* [ ] Merge bioconda/bioconda-utils#293.
* [ ] Release a new bioconda-utils version.
* [ ] Update https://github.com/bioconda/bioconda-common to new bioconda-utils and miniconda version.
* [ ] Revert [override](https://github.com/bioconda/bioconda-recipes/pull/9041/files#diff-1d37e48f9ceff6d8030570cd36286a61R18) of `common.sh` in `.circleci/config.yaml` in this PR.
* [ ] Adjust Docker Hub to create tags not only for branches but [also Git tags](https://github.com/bioconda/bioconda-utils/pull/293#issuecomment-392079490).
* [ ] Merge this PR.


johanneskoester:
@bioconda/core we are now migrating to conda-build 3 in this bulk branch. For the plan, see the checklist in the PR description.

bgruening:
Done with the first 2 items.

johanneskoester:
Oh no, we have done the same thing and you have been faster :-)

johanneskoester:
I wanted to push right now :-).

bgruening:
I will also remove libgcc is that ok?

bgruening:
I'm grepping some food now, take over ;)

johanneskoester:
For libgcc, we need to replace gcc with those new compiler functions.

johanneskoester:
Here, I am working on that now.

johanneskoester:
Ok, compilers have been replaced. For now, I took the c compiler by default. We need to manually change this to `{{ compiler('cxx') }}` wherever needed. I thought this is cleaner than always having both compilers in there. It seems like recipes will fail to build if the host section is not used for library dependencies that are also runtime, because the headers are not present if they are build dependencies. This is good for us, because it enforces us to convert to the host section, which ensures that automatic pinning works as expected.

johanneskoester:
I think everything that can be easily done automatically is done. Now we have to fix by hand.

johanneskoester:
Filtering is really slow with cb3, because it needs to solve the entire environment in order to calculate the build strings. We should probably deactivate it if a git range is provided.

johanneskoester:
Ok, I have changed bioconda-utils so that filtering is no longer done a priori. This should solve the problem, because we can decrease the load as packages are uploaded in bulk.

bgruening:
> We need to manually change this to {{ compiler('cxx') }} wherever needed. I thought this is cleaner than always having both compilers in there.

lol, I was thinking about this on my way home and came to the same conclusion. :+1:

mbargull:
> We need to manually change this to `{{ compiler('cxx') }}` wherever needed.

Replacing that with
```
  # TODO: check whether the build really needs both C and C++ compilers
  - {{ compiler('c') }}
  - {{ compiler('cxx') }}
```
would be the "conceptually safer" (i.e., better surplus than missing dependencies) version. But (for Linux builds) using only `{{ compiler('c') }}`, like you did, can potentially (not guaranteed) catch the missing `libstdcxx_ng`  when testing in the minimal container. Either way has its cons/pros. I think, although I usually prefer the "conceptually correct" way, I currently agree with your choice.

> It seems like recipes will fail to build if the host section is not used for library dependencies that are also runtime, because the headers are not present if they are build dependencies. This is good for us, because it enforces us to convert to the host section, which ensures that automatic pinning works as expected.

Yes, :100: :+1:, nice! 

> Filtering is really slow with cb3, because it needs to solve the entire environment in order to calculate the build strings. We should probably deactivate it if a git range is provided.

Grrr, I completely disregarded the filtering in https://github.com/bioconda/bioconda-utils/pull/293/commits/7079bb3ca4ca5bbf49ce25b820c863bc0413999b. The current code filters on the filenames of built packages, which is "correct", but as you noted infeasible wrt running time. Maybe we can get away by being "slightly less correct" and instead of comparing filenames we compare the counts of package builds with the same `(name, version, build_number)` key.

> https://github.com/bioconda/bioconda-utils/pull/293/commits/d3530ad2e0e47fc3064cbcdfe6f8b1a639fb562b#diff-877ea8e79f39898f92d9e5f6dea153fdR267
```python
    # TODO remove DAG building for now. It takes too long to read all the metadata
    # with CB3.
```

Do you think this is inherently slow with CB3 or maybe just because we inspect too many recipes because of the disabled filtering?

bgruening:
gitter @johanneskoester  ;)
I will start with python packages and migrate them to host?

bgruening:
Just double checking we are also migrating to:
```
run:
  - python 
  - {{ pin_compatible('numpy') }}
```

johanneskoester:
That is the replacement for numpy x.x I guess? Then, yes.

johanneskoester:
DAG building is activated again. We need it in order to sort the recipes topologically. In order to make it feasible, I have implemented a fast way of loading meta.yaml without conda API and just YAML. This should be fine if just used for build order determination.

@mbargull filtering recipes as we did before is not crucial I think. We mainly use git to filter recipes. Now, even cosmetic changes lead to a rebuild of the recipe although the resulting package might not be new. This is acceptable I think.

johanneskoester:
Ok, but now the DAG seems to contain a cycle... The DAG building code itself did not really change, so this is likely real. Unfortunately, this only occurs when doing the topological sort, but not when trying to print the cycle with `nx.find_cycle`. Any help would be greatly appreciated.

kyleabeauchamp:
I added an embed() right before the exception and saw this:

```
In [6]: for sc in nx.simple_cycles(subdag):
   ...:     print(sc)
   ...:     
['perl-extutils-makemaker', 'perl-data-dumper']
```

Could that be it?

kyleabeauchamp:
I then did `mv recipes/perl-data-dumper/ ./tmp` and `bioconda-utils build recipes config.yml` and we seem to be back in business.

johanneskoester:
Good catch! Seems like simple_cycles can find things that find_cycles cannot... I will add a proper check to bioconda utils.

johanneskoester:
@kyleabeauchamp the problem is that we currently use one node for the two versions of perl-extutils-makemaker. I will see if I can find a solution for the DAG problem that solves this issue.

.circleci/config.yml
```diff
@@ -2,20 +2,24 @@ version: 2
 
 variables:
   restore_cache: &restore_cache
-    restore_cache:
-      keys:
-        - miniconda-{{ checksum ".circleci/common.sh" }}-{{ checksum ".circleci/setup.sh" }}-{{ checksum ".circleci/config.yml" }}-{{ arch }}
+    # TODO remove this temporal override before merging the PR into master
+    {}
```
mbargull: Ah, CircleCI didn't like this to be empty

mbargull:
> I have implemented a fast way of loading meta.yaml without conda API and just YAML. This should be fine if just used for build order determination.

If it works, it could pose as a temporary solution. I'm pretty sure that there a many cases this might fail (`conda-build`'s `meta.yaml` format is just very complex, wrt parsing :frowning_face:).

> Do you think this is inherently slow with CB3 or maybe just because we inspect too many recipes because of the disabled filtering?

I did some profiling and we should be able to speed up the rendering a lot (=manyfold) with some very few changes upstream in `conda-build`. Though, it looks like rendering (with `finalize=False, bypass_env_check=True`) will still take 5-6 minutes for all recipes. I'll open a PR upstream (since that's out of scope for this PR currently, the "fast way of loading" is definitely appreciated).

mbargull:
Good news: Some builds for `macos` succeeded!
Bad news: Those builds got uploaded although they have the same build number as already existing ones. This is due to them having another build string (cb3 hash-based) and thus other file names.
These are the uploads from the log:
```
14:15:26 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/zeroc-ice-3.6.3-py27hd0a1c67_0.tar.bz2
14:15:30 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/zeroc-ice-3.6.3-py35hd0a1c67_0.tar.bz2
14:16:09 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/yass-1.14-h470a237_0.tar.bz2
14:17:26 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/yacrd-0.2-h470a237_0.tar.bz2
14:19:20 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/xtandem-15.12.15.2-hd869df4_1.tar.bz2
14:28:29 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/w4mclstrpeakpics-0.98.1-r341_0.tar.bz2
14:29:46 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/w4mclassfilter-0.98.3-r341_0.tar.bz2
14:31:01 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/w4mclassfilter-0.98.1-r341_0.tar.bz2
14:32:16 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/w4mclassfilter-0.98.6-r341_0.tar.bz2
14:33:32 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/w4mclassfilter-0.98.0-r341_0.tar.bz2
14:34:48 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/w4mclassfilter-0.98.2-r341_0.tar.bz2
14:36:05 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/w4mclassfilter-0.98.7-r341_0.tar.bz2
14:36:51 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/vphaser2-2.0-ha3694ff_6.tar.bz2
14:52:41 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/viennarna-2.3.0-h470a237_1.tar.bz2
14:58:32 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/viennarna-2.3.1-h470a237_1.tar.bz2
15:02:18 BIOCONDA INFO UPLOAD SUCCESS: uploaded package /Users/distiller/project/miniconda/conda-bld/osx-64/viennarna-2.3.2-h470a237_0.tar.bz2
```
We probably should label them as broken since we actually want them to have a newer build number.

I now disabled the upload (incl. BioContainer, of course) for the `bulk` branch temporarily until we fix that. I won't be able to look into this until tomorrow night, though.

bgruening:
Thanks @mbargull for your work here.
Can we simply bump all build numbers by one for all packages?

daler:
@bgruening bumping should work, but not sure the best way to do it. One option would be to prepare all recipes in bulk without bumping them (and keeping uploading disabled), then bump everything in one shot. Another option would be to incrementally bump each recipe as it's updated. Seems like the former would take a fair amount of coordination, and the latter leaves room for forgetting to bump something.

Alternatively when checking to see if a package exists in the channel in `get_package_paths()`, we can temporarily only care about `(packagename, version, build_number)` instead of checking the full tarball name.

E.g., 

```
bioconductor-limma-3.34.9-r3.4.1_0.tar.bz2
```

and

```
bioconductor-limma-3.34.9-r341h470a237_0.tar.bz2
```
are both `(bioconductor-limma, 3.34.9, 0)` and so would not be built until the build number is bumped.

bgruening:
I don't think disabling upload will work, we probably need more then one CI run to process everything. Why not bumping all recipes and proceeding fixing recipes that are failing?
One thing that could help dramatically is if we migrate as much recipes as possible no `noarch` builds.

johanneskoester:
@bgruening that sounds like a plan. I have updated the TODO list accordingly.

daler:
@johanneskoester @bgruening similarly I'll work on bioconductor and R in a separate branch targeting this.

bgruening:
@daler I tried to document my shell commands here: https://github.com/bioconda/bioconda-recipes/pull/9049
Maybe you can target that branch: https://github.com/bioconda/bioconda-recipes/pull/9049 I have also added build/numbers everywhere.

daler:
nice, thanks! I'll push to `bulk-noarch`.

mbargull:
> I don't think disabling upload will work, we probably need more then one CI run to process everything. 

Definitely! Disabling was only a temporary measure to avoid more duplicate builds in case someone pushed additional changes.

I think in general the following might work:
* During filtering (if we were to reinstate filtering again), i.e., without a finalized rendering, we could inspect the count of packages in `repodata.json` with the same `(name, version, build_number)` and filter out recipes for which the count in the repository is at least as high as the one reported by `conda-build`.
* When we have a recipe rendered with `finalize=True`, we know all build strings of the packages that would be built. We can filter `repodata.json` for `(name, version, build_number)` again and then:
  * **abort** the build completely whenever there is a package in the repository that as a build string not calculated by `conda-build`. In that case we know that the build setup somehow got changed and thus we need a build number bump.
  * **continue** the build if the set of build strings for packages in the repository is a subset of the ones reported by `conda-build`. In this case there are just some builds missing in the repo, e.g., due to previous upload errors, and as such, we can just upload those.

I'll try to add that functionality to https://github.com/bioconda/bioconda-utils/pull/293 later today.

bgruening:
:+1: for bumping and get this in!
@daler what about making all java packages (with this python wrapper) `noarch: python`?

daler:
@bgruening sure, if it works! I'm not sure of the implications of noarch:python with java though, and don't have a mac handy to test it.
 
Similarly, would it be possible to make BioC data packages `noarch`?

mbargull:
> I'm not sure of the implications of noarch:python with java though

I can't remember the details from the top of my head right now, but we should use `noarch: python` only for real Python packages. There is some special logic in `conda` regarding Python packages for that `noarch` setting.
We should use `noarch: generic` for all other packages.

> Similarly, would it be possible to make BioC data packages noarch?

I'm fairly, but not 100 % sure that `post-link` scripts are a no-no for Conda's `noarch` packages.

mbargull:
> I'm fairly, but not 100 % sure that `post-link` scripts are a no-no for Conda's `noarch` packages.

Though, come to think of it, I believe the old/deprecated `noarch_python` packages used post-link scripts :thinking: -- I'll have to look up what `conda`...

mbargull:
> `post-link.sh` for `noarch: generic` *might* be fine.

Seems to work fine: I just built a package with `CONDA_SUBDIR=osx-64 conda-build recipe_dir` with `conda-build 3.10.3` and installed it with `conda 4.5.4` and the `post-link.sh` ran successfully :thumbsup:

daler:
Thanks for investigating. So `noarch: generic` for both Java and BioC data packages sounds like it should work.

In all cases should we additionally add .bat files, or just leave it to Windows users to find out the hard way?

johanneskoester:
Closing in favor of #9404.

PR #9049, opened 2018-05-27
mass migration of all packages to new cb3 recommendations

bgruening:


bgruening:
ping @johanneskoester @daler 
This first round should be pretty save I hope.

I did it with this command:
```
grep -lZ "\- python" ./* -R | xargs -r0 grep -L "compiler" | xargs grep -L "cython" | xargs grep -L "py2k" | xargs grep -L "py2k" | xargs grep -L "py3k" | xargs grep -L "skip: True" | xargs grep -L "skip: true" | sort -u |  xargs sed -i 's/^build:.*$/build:\n  noarch: python/g'
```

Lets see, it can be that I mixed up indentations ... should we migrate all recipes to 2-spaces?

bgruening:
@johanneskoester @daler in the second commit I added build/number sections to all recipes that are missing them. We can now increase all build numbers with sed.

command used for it: ` grep -L '^build:' ./* -R | grep meta.yaml | xargs grep -L '^host:' | xargs sed -i 's/^$/\nbuild:\n  number: 0\n/g' biogridpy/meta.yaml`

daler:
R and BioC recipes should be OK to have their build number bumped when ready. With those last 3 commits:

- all non-compiler deps moved to `host`
- any recipes depending on `r-rcpp` or `r-rcpparmadillo` use `{{ compiler('cxx') }}` instead of `'c'`.

bgruening:
@bioconda/core let's work on this branch.
* fix the linting failures
* bump all build numbers by one
* merge into bulk, which will trigger a rebuild marathon

Ok?

daler:
@bgruening sounds great!

I noticed that many recipes built with conda-skeleton in the old days that still had the skeleton comments do not have build numbers. One such example is `py-graphviz`. Next on my list was to clean those up, but I don't have time today.

Also, there are lots of recipes that are redundant with conda-forge. What should we do with them?

bgruening:
>I noticed that many recipes built with conda-skeleton in the old days that still had the skeleton comments do not have build numbers. One such example is py-graphviz. Next on my list was to clean those up, but I don't have time today.

Upps, true. I just fixed all recipes that do not have any build-section at all. But the only with build but `number` commented out I missed.

> Also, there are lots of recipes that are redundant with conda-forge. What should we do with them?

Ideally remove them.

bgruening:
I will work on it now.

bgruening:
@daler done. I also changed `perl-threaded` to `perl`

mbargull:
I'll try to skim over those changes later, but nice work already: https://i.imgur.com/chzi8ol.jpg

bgruening:
@daler I will finish the linting hopefully today. How do we proceed with new recipes in master. Do we merge master into bulk-noarch before we increase all build-numbers?

daler:
@bgruening great, thanks.

It looks like the linting you're doing is to make sure meta.yaml can be read, rather than running our linting functions?

I'm working on linting too, starting alphabetically from the top, using `bioconda-utils lint` from `cb3-migration` branch . . . but maybe that'll be more appropriate for working directly on bulk.

Anyway, yes, I think a final merge from master just before bumping build numbers would be good. We'll have to do a round of cleanup on master on any recipes added while we're working here, but hopefully it won't be too many.

daler:
@bgruening (or possibly @mbargull) if a package has `setuptools` in `host:` can it be `noarch: python`?

bgruening:
@daler yes this should be possible.
> It looks like the linting you're doing is to make sure meta.yaml can be read, rather than running our linting functions?

Yes, I think anything else, could/should be done in a hackthon by involving more form the community. The perl recipes are a mess and we need to spend a lot of time on those I think.

mbargull:
> Do we merge master into bulk-noarch before we increase all build-numbers?

Yes, I think we should do that.

> if a package has `setuptools` in `host:` can it be `noarch: python`?

Yes, essentially every Python package that has no direct dependency on compilers or their runtimes (`gcc`, `llvm`, `libgcc`, `cython`, ...) can be `noarch: python`. That is, of course, given that the packages have well defined dependencies and didn't omit their compiler requirements previously.
I *think* (not completely sure) that `conda-build` has some functionality built-in to detect if `.so` files were created and fail (or maybe just warn? \[which would be bad for us\]) a `noarch: python` build then. **If** that's true and I'm not remembering that wrong, this would give us a failsafe before uploading non-`noarch: python` packages to the `noarch` subdir.

Otherwise, I think we should just hold off using `noarch` for packages that have `post-link`, etc., or `{,de}activate` scripts. In case of Bioconductor data packages I would be fine with adding a `post-link.bat` with something like (code not checked!):
```bat
echo "Data package not supported on Windows, yet." >>"%PREFIX%\.messages.txt"
exit /B 1
```

daler:
@bgruening 

> Yes, I think anything else, could/should be done in a hackthon by involving more form the community. The perl recipes are a mess and we need to spend a lot of time on those I think.

I'll make sure the infrastructure is in place to make this as easy as possible on folks helping out.

@mbargull thanks for the extra info.

johanneskoester:
@mbargull @daler @bgruening `noarch: python` also implies independency on the python version. Hence, packages that e.g. skip python 2 (like Snakemake) cannot be noarch.

bgruening:
With a little bit of luck, this should now pass the linting step.
So if anyone wants to bump all build-version we could use the weekend to rebuild :)
My plan would be to spend less time on fixing recipe but putting them on the blacklist, to merge this asap into master.

bgruening:
This can now read again all recipes. We can now skip the linting and start rebuilding.

daler:
@bgruening awesome!

I edited #9124 to aggregate by recipe and display in table form.

For everything in #9124, should they be added to blacklist? Or are you suggesting disabling linting altogether? I think we at least need to be careful about removing packages duplicated on conda-forge.

daler:
OK. I will add the packages with `in_other_channels` lints to the blacklist then.

daler:
Such a large pile of changes makes me nervous ;)

I am about to push changes to bioconda-utils that only lints a recipe just before it is built instead of all at once. That should save us a ton of time without having to disable lints.

daler:
Currently merging in master, and have a script for bumping all build numbers, stay tuned . . .

daler:
(there is a fair amount of merge conflicts to resolve when pulling in master)

.circleci/config.yml
```diff
@@ -151,7 +151,7 @@ jobs:
       # build only current commit (since we use squashed merging, this is safe)
       - run:
           name: Building, testing and uploading
-          command: bioconda-utils build recipes config.yml --anaconda-upload --git-range $CIRCLE_SHA1~1 $CIRCLE_SHA1
+          command: bioconda-utils build recipes config.yml --anaconda-upload --git-range $CIRCLE_SHA1~1 $CIRCLE_SHA1 --prelint
```
mbargull: Since we want to merge into `bulk`, I guess you also want to add this below to `bulk-linux`/`bulk-macos`, right?

daler:
@bgruening @mbargull @johanneskoester I've merged in master, linting happens only just before building a recipe to save lots of time but still allowing linting to happen, and I just bumped all build numbers. I'll let these tests go a little bit to make sure things are working OK, and then we can pull the trigger to merge with `bulk`.

.circleci/config.yml
daler: Thanks, good catch

daler:
@kyleabeauchamp not sure yet. I would lead toward no, because we don't know how long bulk is going to take to load everything. There will probably be a last merge from master to bulk, and maybe we can have a freeze while that last batch is building.

bgruening:
If we see bulk is in principle working, we could also merge into master and keep master running

mbargull:
The run exited due to https://github.com/conda/conda-build/issues/2945.

mbargull:
A few lines above it shows a build string of `h470a237_2` instead of `h14c3975_2` :thinking:

mbargull:
> A few lines above it shows a build string of `h470a237_2` instead of `h14c3975_2` :thinking: 

This happened due to the `conda-build ... --output` command in `bioconda_utils.docker_utils.BUILD_SCRIPT_TEMPLATE` not using the `conda-forge-pinning` config file. Also, I noticed the Docker-based build didn't use the `bioconda_utils-conda_build_config.yaml`.
I think that https://github.com/bioconda/bioconda-utils/pull/293/commits/41f39c06607fe142d47e5ba244cd88444e5367cd should have fixed both of those issues.

bgruening:
Thanks @mbargull! Bumping and merge? To master or bulk?

mbargull:
> Thanks @mbargull! Bumping and merge? To master or bulk?

Oh, I somehow missed that comment. IMO, **not** to `master`!
Rebuilding all those packages will take ages and likely block most new contributions to `master.`

I didn't think of this before, but even merging to `bulk` can be problematic for few cases:
The logic to check whether there are builds with divergent build strings is not available for `master` => if `bulk` created `pkg_newbuildstring_1` and after that `master` created `pkg_oldbuildstring_1`, we have uploaded builds with the same number but different content.
This only affects "build number bump"-only builds on `master` for packages already processed on `bulk`, of course, so isn't something too common.
*If* we want to properly handle those cases, we would have to backport that functionality and release a new `bioconda-utils 0.12.x` version, or be more daring and just put a `bioconda-utils 0.13.0` live on `master` by:
* polishing up https://github.com/bioconda/bioconda-utils/pull/293,
* tagging a new `bioconda-utils` release,
* making Docker Hub create an image for that tag,
* updating `bulk` and `master` for that new release:
  * transfer necessary the CI config changes over to `master` (e.g., `--pre-lint` flags)
  * bump `bioconda-common` to use the `Miniconda` and `bioconda-utils` version used on `bulk`

bgruening:
I'm not concerned about blocking master/bulk a few days, this is a massive change and I think people can wait or help us with this. My concern is that our work here is not mergeable at some point anymore :(

If we merge into bulk, people will see for a week or so new packages but nothing on master and we will have hard times to merge this back.

bgruening:
@daler @johanneskoester @mbargull we need to take a decision here. If BOSC/GCC is coming very close we probably have even less time :(

johanneskoester:
What about merging to bulk, but disallowing (via permissions) merges to master until we are done.

johanneskoester:
I must say that I think both approaches have an equal amount of pros and cons. @bgruening if you ask me, choose what you prefer.

bgruening:
> What about merging to bulk, but disallowing (via permissions) merges to master until we are done?

I guess this boils down to more or less the same as merging to master, but contributors probably need to fix there PRs after we merged bulk into master - merge conflicts will arise.

@daler can you run your script to bump build numbers?

johanneskoester:
Yes, good point. I agree with merging to master then.

daler:
@bgruening, https://github.com/bioconda/bioconda-recipes/pull/9049/commits/849cebef3af826bbfb2c57949ae927bd70611984 already bumped build numbers (just after the last time I had merged in master).

We can't merge into master until conflicts are resolved, right? There are currently a lot of conflicts. I'm on vacation on Monday, will have limited time to do this. Should be able to help a bit when I get back.

For future reference, the script is in https://github.com/bioconda/bioconda-recipes/blob/bulk-noarch/scripts/bump-build.py, which will need to be run on those recipes that changed in master since June 2 (time of last merge).

kyleabeauchamp:
If merge conflicts are an issue then maybe we freeze master until we
resolve and merge?

johanneskoester:
I am working on resolving the conflicts now.

johanneskoester:
Let's merge to master afterwards (see PR #9402). This is the most straightforward way to (a) avoid new conflicts, and (b) get the changes populated to all PRs ASAP.

johanneskoester:
Closing in favor of #9404.

PR #9062, opened 2018-05-29
Added recipe for spalter

tedil:


tedil:
@bioconda/lib @bioconda/rust
This is a rust binary(+library) crate, with at least two dependencies which in turn have either c or c++ dependencies (rust-htslib [htslib] and rustlearn [libsvm]). However, the build process fails due to the fact that `gcc` and `ld` versions are incompatible (at least, that's what I've gathered from the logs, though I'm no expert on cargo and gcc compilation issues). What I've tried so far:

- Specifying gcc 4.8.2 (or more precisely: forcing `gcc <4.8.5` in meta.yaml, because of the `cannot find Scrt1.o` bug):
  - fails because rustlearn depends on libsvm (which gets built from source) and needs a more recent version of gcc (see [log](https://pastebin.com/SaRpmekj))
- Specifying gcc 7.2.0 (using `gcc_linux-64`, `gxx_linux-64`) and ld 2.28 (from `binutils_linux-64`):
  - fails during the linking stage, since the version of `ld` *actually* used (4.8.2) does not match the version of `gcc` (7.2.0) , I think (see [log](https://pastebin.com/u5pEJeph))
- Specifying gcc 5.4.0:
  - fails with the same message as gcc 7.2.0

As I see it, the main question is why the linker is set to `/opt/rh/devtoolset-2/root/usr/libexec/gcc/x86_64-redhat-linux/4.8.2/ld` instead of the one from the respective binutils package. What am I doing wrong? Did I miss something, i.e. setting some environment variables or including a certain (version of a) package?

tedil:
Ping @bioconda/lib @bioconda/rust

dpryan79:
Looks like a missing `-fPIC` flag somewhere.

mbargull:
Hi Till :smile:!
The logs indicate that `cc` is being used instead of the compiler set by the compiler package's activation scripts. So it looks like the Rust/Cargo build does not honor `${CXX}`?? Sadly, I still have no experience with Rust and thus don't know the build process involved and how to modify/fix it. We'll have to look into it at a later point, I guess. Until then:

As for that `cannot find Scrt1.o` thing, that's an easy one. By setting `LIBRARY_PATH=$PREFIX/lib` you probably killed off the standard library paths (which include the `*crt*.o` files) of the compiler.
If you replace the newer compiler/binutils packages with
```diff
-    - gxx_linux-64
-    - gcc_linux-64
-    - binutils_linux-64
+    - toolchain_c_linux-64  # TODO: replace with compiler jinja function in cb3
+    - toolchain_cxx_linux-64  # TODO: replace with compiler jinja function in cb3
```
and the `libgcc` in `run` with
```diff
-    - libgcc
+    - libgcc-ng >=4.9  # TODO: replace with compiler jinja function in cb3
+    - libstdcxx-ng >=4.9  # TODO: replace with compiler jinja function in cb3
```
you'll get the older compilers again but also getting library/include paths in `CFLAGS`/`CXXFLAGS`/`LDFLAGS` set by the activation scripts from `toolchain_*`. With that you should be able to use
```diff
-C_INCLUDE_PATH=$PREFIX/include LIBRARY_PATH=$PREFIX/lib cargo install --root $PREFIX
+cargo install --root "${PREFIX}"
```
. If that doesn't work and the Rust build also does not honor those `*FLAGS`, you can try to use
```diff
+C_INCLUDE_PATH="${PREFIX}/include:${C_INCLUDE_PATH}" \
+    LIBRARY_PATH="${PREFIX}/lib:${LIBRARY_PATH}" \
+    cargo install --root "${PREFIX}"
```
maybe (i.e., adding to, but not overwriting the `*_PATH` variables).

mbargull:
Huh, `git merge master` says `fatal: refusing to merge unrelated histories`... strange...
Would you be fine with me rebasing on upstream `master` and force push to your branch?

mbargull:
Nvm, I had a shallow copy of `master` locally which (rightfully) made Git complain.

recipes/spalter/meta.yaml
```diff
@@ -1,25 +1,22 @@
...
 
 source:
-  fn: {{ name }}-{{version}}.tar.gz
-  url: https://bitbucket.org/tillux/{{ name }}/get/v{{ version }}.tar.gz
+  url: https://bitbucket.org/tillux/spalter/get/v{{ version }}.tar.gz
   sha256: {{ sha256 }}
 
 requirements:
   build:
+    - {{ compiler('c') }}
```
mbargull: So, this `- {{ compiler('c') }}` in `requirements/build` does some "magic" for us:
- Depending on the build configuration (i.e., just OS for Bioconda currently), that Jinja2 template evaluates to the compiler package for the given language (`'c'`), e.g., `toolchain_c_linux-64` (and likely `gcc_linux-64` once we update to newer compilers).
- That `toolchain_c_linux-64` does the rest of the magic then:
  - It has (via another package called `toolchain`) some "activation script" that automatically sets a bunch of compiler/linker variables (`LDFLAGS`, etc.).
  - It has a `run_export` definition for the `libgcc-ng` package which then automagically gets added to `requirements/run` to `spalter`'s recipe.

At some point we may (or rather will, I assume) add `{{ compiler('rust') }}`, too (via the `conda-forge-pinning` package). It's just not there yet :wink:.

dlaehnemann: Thanks for the elaboration, will be useful for other rust-related recipes!

daler:

(though I did learn a lot from Marcel's comments!)

Issue #9056, opened 2018-05-28
SAMtools - can't access remote sequence when reading CRAM


garrett-stevens:
Hello all,

When decoding CRAM files, if there are `@SQ` headers with MD5 values (`M5` tags) but the specified FASTA file can't be found/opened, SAMtools will by default try to use the MD5 to download the sequence from the EBI server (`http://www.ebi.ac.uk/ena/cram/md5/<md5>`) and save it to the REF_CACHE location.

Using SAMtools from bioconda, I am not able to get this to work:
```
$ conda create -y -p /opt/anaconda3/envs/test1 samtools
$ source activate test1
$ samtools --version
samtools 1.8
Using htslib 1.8
Copyright (C) 2018 Genome Research Ltd.
$ samtools view test.cram > /dev/null
[E::cram_get_ref] Failed to populate reference for id 0
[E::cram_decode_slice] Unable to fetch reference #0 11109669..16279328
[E::cram_next_slice] Failure to decode slice
[main_samview] truncated file.
```
However, using SAMtools I compile on my own, this works just fine (some output omitted for brevity):
```
$ wget https://github.com/samtools/samtools/releases/download/1.8/samtools-1.8.tar.bz2
$ tar xjf samtools-1.8.tar.bz2
$ cd samtools-1.8/
$ ./configure
$ make
$ cd ..
$ ./samtools-1.8/samtools --version
samtools 1.8
Using htslib 1.8
Copyright (C) 2018 Genome Research Ltd.
$ # This works fine
$ ./samtools-1.8/samtools view test.cram > /dev/null
$ # Also, you can see in my cache where the downloaded reference now is:
$ ls -R .cache/
.cache/:
hts-ref

.cache/hts-ref:
a7

.cache/hts-ref/a7:
18

.cache/hts-ref/a7/18:
acaa6135fdca8357d5bfe94211dd
```

I've tried manually setting `REF_PATH='http://www.ebi.ac.uk/ena/cram/md5/%s'` (also tried HTTPS) and `REF_CACHE='$HOME/.cache'` when using the bioconda version, but that doesn't help. I can see from the `build.sh` that libcurl is explicitly enabled, so I'm not sure what's preventing it from downloading those references via HTTP: https://github.com/bioconda/bioconda-recipes/blob/fa495c10ee8e421f12afe7ceb373eeb8e6487652/recipes/samtools/build.sh#L21

[...]

PR #9072, opened 2018-05-29
clean samtools recipe

rvalieris:

xref: https://github.com/bioconda/bioconda-recipes/issues/9056

* added https protocol test
* removed redundant  configure/make on htslib directory
* removed enable-plugins flags

still pending linking against the htslib package, but since conda-build 3 is close, I think we can wait for that to avoid pinning htslib.

the enable-plugins flag was removed, but note that since the plugins where never built no functionally was removed, if the plugins are desired we will have to include them on htslib recipe first and link here later.

### DIFF
--- a/recipes/bcftools/build.sh
+++ b/recipes/bcftools/build.sh
@@ -3,11 +3,6 @@
 export CPPFLAGS="-DHAVE_LIBDEFLATE -I$PREFIX/include"
 export LDFLAGS="-L$PREFIX/lib"
 
-cd htslib*
 ./configure --prefix=$PREFIX --enable-libcurl CPPFLAGS="$CPPFLAGS" LDFLAGS="-L$PREFIX/lib"
-make
-cd ..
-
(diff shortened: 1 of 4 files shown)


kyleabeauchamp:
I think we should make the same parallel changes to bcftools.  WDYT?

rvalieris:
osx build failed with 
> bgzf.c:39:10: fatal error: 'libdeflate.h' file not found

did I do something wrong ?

kyleabeauchamp:
Hmm, I *thought* I had libdeflate support working in both bcftools and samtools.  Perhaps there needs to be some additional compiler flags set.  ("-I" ?).  Not sure what the problem is.  If you need to disable the lib-deflate in bcftools, that's not the end of the world (we can always re-visit when we have CB3 + separate HTSLib package).

rvalieris:
cool, thanks for the help @kyleabeauchamp.

PR #9085, opened 2018-05-30
Whitespace fixes [WIP]

daler:

@bgruening this uses the new `scripts/indentation-fixer.py` to fix indentation to 2 spaces. Couldn't work out a more elegant way than tracking indentation levels.

Eventually I'm thinking we can use something like that in bioconda-utils for linting (I can add that later), but for now I'm running it on all recipes to get consistent, 2-space indentation and remove trailing whitespace. Figured since we're mucking about with the recipes for cb3, might as well clean 'em up a bit.

Method is to use 
```bash
for i in `find . -name "meta.yaml" | sort`; do
  echo $i
  scripts/indentation-fixer.py $i --inplace
done
```
and then manually check diffs before adding. The script sometimes has trouble with summaries, so sometimes you have to discard its edits in summary sections. For the most part it does a decent job though.


bgruening:
Awesome!

PR #9088, opened 2018-05-30
flash bump

mictadlo:

### DIFF
--- a/recipes/flash/meta.yaml
+++ b/recipes/flash/meta.yaml
@@ -1,14 +1,18 @@
+{% set name = "flash" %}
+{% set version = "2.2.00" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+
 build:
   number: 1
 
-package:
-  name: flash
-  version: '1.2.11'
-
 source:
-  fn: FLASH-1.2.11.tar.gz
-  url: https://sourceforge.net/projects/flashpage/files/FLASH-1.2.11.tar.gz
-  sha256: 685ca6f7fedda07434d8ee03c536f4763385671c4509c5bb48beb3055fd236ac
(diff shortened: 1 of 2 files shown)


recipes/flash/meta.yaml
```diff
@@ -21,13 +24,10 @@ requirements:
...
+    - flash2 -v
 
 about:
-  home: https://ccb.jhu.edu/software/FLASH/
+  home: https://github.com/dstreett/FLASH2
   license: GPLv3+
-  summary: 'Merge mates from fragments that are shorter than twice the read length'
+  summary: 'Merge paired-end reads from fragments that are shorter than twice the read length'
 
-extra:
```
bgruening: Why have you removed this? its still the same tool isn't it?
Is there a publication which we can reference?

recipes/flash/meta.yaml
```diff
@@ -1,14 +1,17 @@
...
+{% set version = "2.2.00" %}
 
 package:
-  name: flash
-  version: '1.2.11'
+  name: {{ name|lower }}
+  version: {{ version }}
+
+build:
+  number: 1
```
bgruening: start with 0

marcelm:
I’m uncertain as to the status of how FLASH2 relates to FLASH. Looking at the [GitHub repository](https://github.com/dstreett/FLASH2/), it seems that the project is an independent one (not from the original authors).

Perhaps @dstreett could explain?

According to dstreet/FLASH2#2, the behavior is also slightly different, so perhaps it woud be an idea to put this into a separate recipe named `flash2`.

bgruening:
It it is really a separate "tool" than a new recipe would be preferred, I agree.

mictadlo:
I created a new PR https://github.com/bioconda/bioconda-recipes/pull/9103

PR #9089, opened 2018-05-31
Spades 3.5.0

pjbriggs:

PR adding a legacy version of `SPAdes` (version 3.5.0).

(NB although MacOS binaries appear to be available for this version, MacOS support has been excluded as the CircleCI builds for this platform stall for unknown reasons and have to be killed. As this is a legacy version of the software it doesn't seem worth expending effort to investigate at this time.)

### DIFF
--- a/recipes/spades/3.5.0/meta.yaml
+++ b/recipes/spades/3.5.0/meta.yaml
@@ -1,0 +1,33 @@
+package:
+  name: spades
+  version: '3.5.0'
+
+source:
+  fn: SPAdes-3.5.0-Linux.tar.gz                                              # [linux]
+  url: http://spades.bioinf.spbau.ru/release3.5.0/SPAdes-3.5.0-Linux.tar.gz  # [linux]
+  md5: 1a4ef6e9c9da8894353fb76ed1a31c9d                                      # [linux]
+  fn: SPAdes-3.5.0-Darwin.tar.gz                                             # [osx]
+  url: http://spades.bioinf.spbau.ru/release3.5.0/SPAdes-3.5.0-Darwin.tar.gz # [osx]
+  md5: d3c4c89591436328f0680e1ce6f5e637                                      # [osx]
+
+build:
+  number: 0
+  skip: True  # [py36]
+
+requirements:
+  run:
+    - python <=3.3
+
+test:
+  commands:
+    - spades.py --test && rm -rf spades_test
+
+about:
+  home: http://cab.spbu.ru/software/spades/
+  license: GPLv2
+  summary: SPAdes (St. Petersburg genome assembler) is intended for both standard isolates and single-cell MDA bacteria assemblies.
+
+extra:
+  identifiers:
+    - biotools:Spades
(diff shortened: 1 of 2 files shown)


recipes/spades/3.5.0/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+  url: http://spades.bioinf.spbau.ru/release3.5.0/SPAdes-3.5.0-Linux.tar.gz  # [linux]
+  md5: 1a4ef6e9c9da8894353fb76ed1a31c9d                                      # [linux]
+
+build:
+  number: 0
+  skip: True  # [py36 or osx]
+
+requirements:
+  run:
+    - python <=3.3
```
bgruening: Can you remove this pinning and adjust the one above at build/skip?

pjbriggs:
Hi @bgruening thanks for the review - I've updated, and while the skip looks a bit ugly it seems to work. Does this seem okay to you?

recipes/spades/3.5.0/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+  version: '3.5.0'
+
+source:
+  fn: SPAdes-3.5.0-Linux.tar.gz                                              # [linux]
+  url: http://spades.bioinf.spbau.ru/release3.5.0/SPAdes-3.5.0-Linux.tar.gz  # [linux]
+  md5: 1a4ef6e9c9da8894353fb76ed1a31c9d                                      # [linux]
+
+build:
+  number: 0
+  skip: True  # [py34 or py35 or py36 or osx]
```
bgruening: [py3k or osx] would be more correct isn't it?

recipes/spades/3.5.0/meta.yaml
```diff
@@ -0,0 +1,30 @@
+package:
+  name: spades
+  version: '3.5.0'
+
+source:
+  fn: SPAdes-3.5.0-Linux.tar.gz                                              # [linux]
```
bgruening: if you skip osx you don't need to pin this here to linux

recipes/spades/3.5.0/meta.yaml
pjbriggs: This version of SPAdes can be built with Python 3.3 (hence the original pinning), which does appear to be supported by conda. Using `[py3k or ...]` would presumably disable the package for Python 3.3 within bioconda.
Is there a neater way to maintain support for Python 3.3? Or should it just drop support for all Python3 versions? (It's not needed for the pipeline I'm trying to support.)

bgruening: We do not build for 3.3 anymore. See here: https://github.com/bioconda/bioconda-recipes/blob/master/scripts/env_matrix.yml#L1

pjbriggs: Thanks - it's as you say then!

bgruening:
Cool thanks Peter!

pjbriggs:
Thanks Bjoern!

PR #9092, opened 2018-05-31
Blast legacy 2.2.26

pjbriggs:

This PR adds a recipe for `blast-legacy` version 2.2.26.

### DIFF
--- a/recipes/blast-legacy/2.2.26/build.sh
+++ b/recipes/blast-legacy/2.2.26/build.sh
@@ -1,7 +1,1 @@
 cp -R bin $PREFIX
-cp -R data $PREFIX
-cp -R doc $PREFIX
-#mkdir -p $PREFIX/etc/conda/activate.d/
-#echo "export BLASTMAT=$PREFIX/data/" > $PREFIX/etc/conda/activate.d/set_blastmat_env_variable.sh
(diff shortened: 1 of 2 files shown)


recipes/blast-legacy/2.2.26/build.sh
```diff
@@ -0,0 +1,7 @@
+cp -R bin $PREFIX
+cp -R data $PREFIX
+cp -R doc $PREFIX
```
bgruening: is doc and data needed?

pjbriggs: These were taken from the `build.sh` from the existing `blast-legacy` 2.2.22 recipe. The doc for standalone `blast` refers to the `data` directory but it needs the user to configure the location by creating a  `.ncbirc` file. I'm happy to remove `data` and `doc`if they're not appropriate in your opinion.

bgruening: I'm just wondering and would like to keep the package small if possible. If there is nothing that is relevant for the binaries, lets remove it.

pjbriggs: No problem, I understand. I'll remove them and if it turns out there's a problem then it's always possible to update in future. Thanks for taking a look!

pjbriggs:
Thanks Bjoern!

PR #9100, opened 2018-05-31
Add r-picante

nick-youngblut:

This R package is currently on [conda-forge](https://github.com/conda-forge/r-picante-feedstock) (I added the package), but since it's a bioinformatics R package, I think that it should belong on bioconda (see [Issue#1](https://github.com/conda-forge/r-picante-feedstock/issues/1))

### DIFF
--- a/r-picante/meta.yaml
+++ b/r-picante/meta.yaml
@@ -45,3 +45,6 @@ about:
     functions for data input/output and phylogeny plotting. A full description of package
     functionality and methods are provided by Kembel et al. (2010)
   license_family: GPL2
+
+extra:
+  doi: 10.1093/bioinformatics/btq166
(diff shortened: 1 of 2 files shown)


r-picante/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - r-base
+    - r-ape
+    - r-nlme
+    - r-vegan{indent}libgcc
+
+test:
+  commands:
+    - $R -e "library('picante')"
+
+about:
```
bgruening: can you add a doi to the extras section?

nick-youngblut:
@bgruening I've added the doi to the extras section. Note that the tests are currently skipping the package build because it is technically still available from conda-forge (see above)

PR #9128, opened 2018-06-02
PacBio: bump blasr_libcpp to 5.3.1 and take ownership

pb-dseifert:


pb-dseifert:
@bgruening would it be possible to bump up the HDF5 we use in Conda? At PacBio, we use 1.8.20 which works with libblasr. Alternatively, we could also use HDF5 1.10 (which also works with libblasr)?

bgruening:
@pb-dseifert if you have a Q please wait with merging :) It always good to get a review.

pb-dseifert:
@bgruening I decided that bumping the bioconda-wide HDF5 to 1.8.20 or even 1.10 would be way too invasive, given that there is some API churn involved, hence I decided to locally revert our API porting efforts and use that git patch in order to get build up. Once you're ready to bump the HDF5 version, fixing **blasr_libcpp** is as simple as just removing the patch.

PR #9139, opened 2018-06-04
Conda recipe for PheWAS R package - phenome wide association study

slacalle:

### DIFF
--- a/recipes/r-phewas/meta.yaml
+++ b/recipes/r-phewas/meta.yaml
@@ -1,0 +1,70 @@
+{% set version = '0.12.1' %}
+{% set name = 'PHeWAS' %}
+{% set hash = '7063af0c3759e37e44f4ee2b92e22e579d9c58e15f0a529de593540489878544' %}
+
+package:
+  name: 'r-{{ name|lower }}'
+  version: '{{ version }}'
+
+source:
+  fn: 'v0.12.tar.gz'
+  url:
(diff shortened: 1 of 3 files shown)


recipes/r-phewas/meta.yaml
```diff
@@ -24,7 +24,7 @@ build:
 requirements:
   build:
     - gcc_linux-64
-    - gofrtran_linux-64
+    - gfortran_linux-64
```
marcelotrevisani: it would be great if you change that to:

```
- {{ compiler('c') }}
- {{ compiler('fortran') }}
```

:)

slacalle: Hi @marcelotrevisani I tried with jinja functions but did not work. will try again.

recipes/r-phewas/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+  skip: True    # [win]
+  number: 0
+
+  rpaths:
+    - lib/R/lib/
+    - lib/
+
+requirements:
+  build:
+    - {{ compiler('c') }}
```
marcelotrevisani: I think we have to change that to:

```
- {{ compiler('r_c') }}
```

slacalle: Hi @marcelotrevisani ,

From what I gather from https://bioconda.github.io/troubleshooting.html it looks something to do with the build environment and conda-build=3 new features. i will look to ammend the recipe to be compatible with the former version.

Thanks,

PR #9149, opened 2018-06-05
Add r-ngsplot package

apeltzer:

So this should hopefully work and provide the base package of ngs.plot as a bioconda package, as requested [here](https://github.com/shenlab-sinai/ngsplot/issues/74) and  [here](https://github.com/nf-core/chipseq/issues/15).

One question regarding the packaging here remains open for me: This is NOT a "real" R package, but instead a bunch of R scripts that are bundled together, so I can't install these e.g. using `R CMD INSTALL --build .` similar to a standard CRAN or BIOCONDUCTOR package. I did this with symlinking the respective scripts (all of them!) in `bin` and that should probably work, however some feedback from @bioconda/core  would be great to be sure this is fine for everyone. Happy to do things differently, just nothing on the docs for such a case... 

Regarding the Databases that come with this, @bgruening suggested using `post-link` scripts. However we don't have a stable URL for these (GDrive only, no versioning): How do I drop these at the Galaxy Servers as you suggested @bgruening ? 
Thanks!

### DIFF
--- a/recipes/r-ngsplot/meta.yaml
+++ b/recipes/r-ngsplot/meta.yaml
@@ -5,41 +5,31 @@ package:
 
 source:
   url: https://github.com/shenlab-sinai/ngsplot/archive/{{ version }}.tar.gz
-  md5: bf0d6f42d9461aeb7a20b1ddb12b57bc42d11732ca6f69ec3eb2157364cefd5b
+  md5: 811815e5f5ef7bdc1cb028ecd59fb2a7
 
 build:
   number: 0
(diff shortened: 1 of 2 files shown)


recipes/r-ngsplot/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  name: r-ngsplot
+  version: {{ version }}
+
+source:
+  url: https://github.com/shenlab-sinai/ngsplot/archive/{{ version }}.tar.gz
+  md5: 811815e5f5ef7bdc1cb028ecd59fb2a7
+
+build:
+  number: 0
+  skip: True
```
mbargull: Whoops :wink:

PR #9157, opened 2018-06-05
Add NGSPlotDB-hg19-data recipe

apeltzer:

The GDrive URL is somewhat a bit crazy - submitted a [PR](https://github.com/galaxyproject/cargo-port/pull/148) to get this to Galaxy Cargo-Port too. Maybe that eases things... 

Referencing https://github.com/nf-core/chipseq/issues/2 
and https://github.com/nf-core/chipseq/issues/15 to keep things related.

If this one is fine, I'm trying to do the same for hg38 and mm10 to cover the most used DBs for ngsplot.


apeltzer:
@bioconda/core  The `r-ngsplot` package seems to be missing for this? I submitted that yesterday, @bgruening merged it in - so it should be there right?

mbargull:
You remove the `skip: True` line from the `r-ngsplot` recipe (see https://github.com/bioconda/bioconda-recipes/pull/9149/files#r193333485) and add that change to this PR.

apeltzer:
Whoopsies!

apeltzer:
Hmm, doesn't build that way. I will probably resubmit in a different PR then?!  @mbargull

recipes/r-ngsplot-hg19/post-link.sh
```diff
@@ -0,0 +1,44 @@
...
+done
+
+if [[ $SUCCESS != 1 ]]; then
+  echo "ERROR: post-link.sh was unable to download any of the following URLs with the md5sum $MD5:"
+  printf '%s\n' "${URLS[@]}"
+  exit 1
+fi
+
+# Install and clean up
+ngsplotdb.py install ngsplotdb_hg19_75_3.00.tar.gz
```
mbargull: You want to use "${TARBALL}" here.
Also, if this outputs anything (progress status and such), you'll want to write that to `.messages.txt`, see https://conda.io/docs/user-guide/tasks/build-packages/link-scripts.html, i.e.,
```bash
ngsplotdb.py install "${TARBALL}" > "${PREFIX}/.messages.txt"
```
maybe even
```bash
ngsplotdb.py install "${TARBALL}" > "${PREFIX}/.messages.txt" 2>&1
```
to also capture `stderr`.

recipes/r-ngsplot-hg19/pre-unlink.sh
```diff
@@ -0,0 +1 @@
+ngsplotdb.py remove hg19
```
mbargull: Redirect output to `.messages.txt`, see above.

recipes/r-ngsplot-hg19/post-link.sh
```diff
@@ -0,0 +1,44 @@
...
+      SUCCESS=1
+      break
+    fi
+  fi
+fi
+done
+
+if [[ $SUCCESS != 1 ]]; then
+  echo "ERROR: post-link.sh was unable to download any of the following URLs with the md5sum $MD5:"
+  printf '%s\n' "${URLS[@]}"
```
mbargull: I think even this should write to `.messages.txt` (see below), i.e., 
```bash
  echo "ERROR: post-link.sh was unable to download any of the following URLs with the md5sum $MD5:" > "${PREFIX}/.messages.txt" 2>&1
  printf '%s\n' "${URLS[@]}" > "${PREFIX}/.messages.txt" 2>&1
```

mbargull:
> Hmm, doesn't build that way. I will probably resubmit in a different PR then?! @mbargull

You can use this one, no problem.

```diff
+#!/bin/bash
+outdir=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+mkdir -p $outdir
+mkdir -p $PREFIX/bin
+cp -R * $outdir 
```
^-- please don't copy everything unconditionally but rather selectively the files you want.
```diff
+#Set up links for 
+for f in $outdir/*; do
+    ln -s $outdir/* $PREFIX/bin
```
^-- you'll want to use `ln -s "${f}" "${PREFIX}/bin"` here. But also, you should not loop over `$outdir/*` but just the files in `$outdir` that are the executables you want to link to.
```diff
+    fbname=$(basename "$f")
+    chmod 0755 ${PREFIX}/bin/$fbname
+done 
```

mbargull:
I took a quick peek at the file list in https://github.com/shenlab-sinai/ngsplot/archive/2.63.tar.gz. There is a `bin` folder in there => I assume you want to only link the files (or possibly only a subset of them?) in `"${outdir}"/bin/*` then, right?

mbargull:
And the files in `bin` seem to have the executable mode set already, to the `chmod` is likely not needed.

recipes/r-ngsplot-hg19/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  name: 'ngsplotdb-{{ name|lower }}'
+  version: '{{ version }}'
+source:
+  fn: '{{ name }}_{{ version }}.tar.gz'
+  url:
+    - 'https://doc-0g-a0-docs.googleusercontent.com/docs/securesc/ha0ro937gcuc7l7deffksulhg5h7mbp1/jiksn2m41cm3dbspq0nqt102h6m6eo1o/1528221600000/01382619737792242945/*/0B5hDZ2BucCI6SURYWW5XdUxnbW8?e=download'
+  sha256: a3ad6daceec383f88faf3d3ee899f2bef37b3be2658ee9afa01e86404c0c92bd
+build:
+  number: 0
+  rpaths:
```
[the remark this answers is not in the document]
apeltzer: Done!

apeltzer:
Hi @mbargull !
thanks for all the tips / hints. The problem is, that the respective tools I need `ngsplot.py` use system calls to respective tools in the `bin` folder and I don't want to try & error each of them until this works. Rather I'd say I restrict to simply add all tools in `bin` to the path, similar to what the authors suggest in their readme. What do you think?

recipes/r-ngsplot-hg19/post-link.sh
```diff
@@ -0,0 +1,44 @@
...
+TARBALL=$STAGING/$FN
+
+SUCCESS=0
+for URL in ${URLS[@]}; do
+  wget -O- -q $URL > $TARBALL
+  [[ $? == 0 ]] || continue
+
+  # Platform-specific md5sum checks.
+  if [[ $(uname -s) == "Linux" ]]; then
+    if md5sum -c <<<"$MD5  $TARBALL"; then
```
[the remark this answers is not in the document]
apeltzer: Changed back to md5

recipes/r-ngsplot-hg19/post-link.sh
```diff
@@ -0,0 +1,44 @@
...
+      SUCCESS=1
+      break
+    fi
+  else if [[ $(uname -s) == "Darwin" ]]; then
+    if [[ $(md5 $TARBALL | cut -f4 -d " ") == "$MD5" ]]; then
```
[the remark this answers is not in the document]
apeltzer: Changed to md5

recipes/r-ngsplot-hg19/meta.yaml
```diff
@@ -0,0 +1,31 @@
+{% set version = "3.00" %}
+{% set name = "ngsplotdb-hg19" %}
+
+package:
+  name: 'ngsplotdb-{{ name|lower }}'
+  version: '{{ version }}'
+source:
```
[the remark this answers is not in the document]
apeltzer: Ok, removed that - but the md5 is still required / good practice right?

recipes/r-ngsplot-hg19/meta.yaml
```diff
@@ -0,0 +1,21 @@
...
+build:
+  number: 0
+requirements:
+  run:
+    - 'r-ngsplot'
+    - r-base
+    - wget
+test:
+  commands:
+    - ngsplotdb.py install "${TARBALL}" > "${PREFIX}/.messages.txt" 2>&1
```
mbargull: The test is run after the package is installed, i.e., after the `post-link.sh` is run.
So, that `ngsplotdb.py install` is already done at that point. Rather than executing it again, you should run a test that checks if the database has been successfully installed (either with some (short-running!) command from `r-gnsplot` or if that's not available, at least something like `test -f "${PREFIX}/some-file-path"`.)

recipes/r-ngsplot-hg19/meta.yaml
```diff
@@ -0,0 +1,21 @@
+{% set version = "3.00" %}
+{% set name = "ngsplotdb-hg19" %}
+
+package:
+  name: 'ngsplotdb-{{ name|lower }}'
+  version: '{{ version }}'
+  md5: 172a6a93de3d1ae8a3ad6d6d79e911f7
```
mbargull: You can remove that `md5` here.

recipes/r-ngsplot-hg19/post-link.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+FN="ngsplotdb_hg19_75_3.00.tar.gz"
+URLS=(
+  "https://drive.google.com/uc?export=download&id=0B5hDZ2BucCI6SURYWW5XdUxnbW8"
+)
+sha256="a3ad6daceec383f88faf3d3ee899f2bef37b3be2658ee9afa01e86404c0c92bd"
```
mbargull:
```
MD5='172a6a93de3d1ae8a3ad6d6d79e911f7'
```
instead of `sha256`

recipes/r-ngsplot-hg19/post-link.sh
```diff
@@ -0,0 +1,44 @@
...
+# Use a staging area in the conda dir rather than temp dirs, both to avoid
+# permission issues as well as to have things downloaded in a predictable
+# manner.
+STAGING=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+mkdir -p $STAGING
+TARBALL=$STAGING/$FN
+
+SUCCESS=0
+for URL in ${URLS[@]}; do
+  wget -O- -q $URL > $TARBALL
```
mbargull: The CI choked on the unquoted `$URL`, I think.
```
wget -qO "${TARBALL}" "${URL}"
```
should do the trick.

recipes/r-ngsplot-hg19/post-link.sh
```diff
@@ -0,0 +1,44 @@
...
+    if [[ $(md5 $TARBALL | cut -f4 -d " ") == "$MD5" ]]; then
+      SUCCESS=1
+      break
+    fi
+  fi
+fi
+done
+
+if [[ $SUCCESS != 1 ]]; then
+  echo "ERROR: post-link.sh was unable to download any of the following URLs with the md5sum $MD5:" > "${PREFIX}/.messages.txt" 2>&1
```
mbargull: My bad, should be (appending to `.messages.txt`):
```bash
  echo "ERROR: post-link.sh was unable to download any of the following URLs with the md5sum $MD5:" >> "${PREFIX}/.messages.txt" 2>&1
```

recipes/r-ngsplot-hg19/post-link.sh
```diff
@@ -0,0 +1,44 @@
...
+  printf '%s\n' "${URLS[@]}" > "${PREFIX}/.messages.txt" 2>&1
```
mbargull: My bad, should be (appending to `.messages.txt`):
```bash
  printf '%s\n' "${URLS[@]}" >> "${PREFIX}/.messages.txt" 2>&1
```

recipes/r-ngsplot-hg19/post-link.sh
```diff
@@ -0,0 +1,44 @@
...
+  exit 1
+fi
+
+# Install and clean up
+ngsplotdb.py install "${TARBALL}" > "${PREFIX}/.messages.txt" 2>&1
```
mbargull: My bad, should be (appending to `.messages.txt`):
```bash
ngsplotdb.py install "${TARBALL}" >> "${PREFIX}/.messages.txt" 2>&1
```

recipes/r-ngsplot-hg19/pre-unlink.sh
```diff
@@ -0,0 +1 @@
+ngsplotdb.py remove hg19 > "${PREFIX}/.messages.txt" 2>&1
```
mbargull: My bad, should be (appending to `.messages.txt`):
```bash
ngsplotdb.py remove hg19 >> "${PREFIX}/.messages.txt" 2>&1
```

mbargull: Also, missing new line on EOF :wink:

recipes/r-ngsplot-hg19/post-link.sh
```diff
@@ -0,0 +1,44 @@
...
+rm $TARBALL
+rmdir $STAGING
```
mbargull: Missing new line on EOF :wink:

recipes/r-ngsplot/build.sh
```diff
@@ -4,8 +4,7 @@ mkdir -p $outdir
 mkdir -p $PREFIX/bin
 cp -R * $outdir 
 #Set up links for 
-for f in $outdir/*; do
-    ln -s $outdir/* $PREFIX/bin
+for f in ${outdir}/bin/*; do
+    ln -s ${f} ${PREFIX}/bin
     fbname=$(basename "$f")
```
mbargull: Can be removed now.

recipes/r-ngsplot/build.sh
```diff
@@ -4,8 +4,7 @@ mkdir -p $outdir
...
-    chmod 0755 ${PREFIX}/bin/$fbname
 done
```
mbargull: Also, missing new line on EOF :wink:

recipes/r-ngsplot/build.sh
```diff
@@ -4,8 +4,7 @@ mkdir -p $outdir
 mkdir -p $PREFIX/bin
 cp -R * $outdir
```
mbargull: I'm not sure if it's still the case, but I know `conda-build` sometimes wrote files to the cwd during building (I think `conda-build.sh` or `run-test.sh` or something). Those files will/might get copied too if you `cp -R *` here.

recipes/r-ngsplot-hg19/meta.yaml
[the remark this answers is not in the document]
mbargull: Nah, the `md5` in `meta.yaml` is only tied to the one in `source`, hence can be removed, too.

recipes/r-ngsplot-hg19/post-link.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 FN="ngsplotdb_hg19_75_3.00.tar.gz"
 URLS=(
-  "https://doc-0g-a0-docs.googleusercontent.com/docs/securesc/ha0ro937gcuc7l7deffksulhg5h7mbp1/jiksn2m41cm3dbspq0nqt102h6m6eo1o/1528221600000/01382619737792242945/*/0B5hDZ2BucCI6SURYWW5XdUxnbW8?e=download"
+  "https://drive.google.com/uc?export=download&id=0B5hDZ2BucCI6SURYWW5XdUxnbW8"
```
mbargull: :+1: looks much saner :smile:. https://github.com/galaxyproject/cargo-port/pull/148 is probably still preferred, I think, but if that isn't merged in a timely manner, we can always package a new build for `r-ngsplot-hg19` updating that URL later.

recipes/r-ngsplot/build.sh
apeltzer: I don't get it right now - why?

recipes/r-ngsplot/build.sh
apeltzer: Done

apeltzer:
So setting this environment variable as suggested in the [README](https://github.com/shenlab-sinai/ngsplot#local-installation) won't work. Therefore, the second test fails and probably also the recipe can't be called properly. 

Using something like this:
```
#Setting environment variable NGSPLOT as requested by authors readme
ln -s ${outdir} ${PREFIX}/bin/NGSPLOT
```
Doesn't work unfortunately... Setting something via `export NGSPLOT=/path/to...` doesn't work either. 

@mbargull any idea? This is the last thing required for the recipe...

apeltzer:
Ok, I don't have a clue what causes this test fail... 

@bioconda/core  anybody an idea?

recipes/r-ngsplot/build.sh
mbargull: Sorry, I've been imprecise. Just meant the `fbname=$(basename "$f")` part :wink:

recipes/r-ngsplot/build.sh
```diff
@@ -2,10 +2,17 @@
...
\ No newline at end of file
+done
+
+#Activate/Deactivate dir
+ACTIVATE_DIR=$PREFIX/etc/conda/activate.d
+DEACTIVATE_DIR=$PREFIX/etc/conda/deactivate.d
+mkdir -p $ACTIVATE_DIR
+mkdir -p $DEACTIVATE_DIR
+cp $RECIPE_DIR/scripts/activate.sh $ACTIVATE_DIR/ngsplot-activate.sh
+cp $RECIPE_DIR/scripts/deactivate.sh $DEACTIVATE_DIR/ngsplot-deactivate.sh
```
mbargull: To avoid conflicts, you should rename those to the exact name of the package, i.e., `r-ngsplot-...`. You can use the `${PKG_NAME}` variable for this, see, e.g.: https://github.com/conda-forge/toolchain-feedstock/blob/bf0bfa1f06f6592eb3e6ef8989339a005cfd93ef/recipe/install-toolchain.sh#L8

recipes/r-ngsplot-hg19/post-link.sh
```diff
@@ -0,0 +1,44 @@
...
+      SUCCESS=1
+      break
+    fi
+  fi
+fi
+done
+
+if [[ $SUCCESS != 1 ]]; then
+  echo "ERROR: post-link.sh was unable to download any of the following URLs with the md5sum $MD5:" >> "${PREFIX}/.messages.txt" 2>&1
+  printf '%s\n' "${URLS[@]}" > "${PREFIX}/.messages.txt" 2>&1
```
mbargull: Please fix my `>` mistake :wink:
```sh
printf '%s\n' "${URLS[@]}" >> "${PREFIX}/.messages.txt" 2>&1
```

recipes/r-ngsplot-hg19/post-link.sh
mbargull: Please fix my `>` mistake :wink:
```sh
ngsplotdb.py install "${TARBALL}" >> "${PREFIX}/.messages.txt" 2>&1
```

recipes/r-ngsplot-hg19/pre-unlink.sh
mbargull: Please fix my `>` mistake :wink:
```sh
ngsplotdb.py remove hg19 >> "${PREFIX}/.messages.txt" 2>&1
```

recipes/r-ngsplot/scripts/activate.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+export NGSPLOT=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
```
mbargull: Oh, and using `$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM` should not work here, I believe.
Those variable are only defined during the build/test phase or during `post-link` (and others), but not when activating an environment. If the tests passed for this, that's only because they were set in that case.
You'll want to create/modify the activation script in `build.sh`.

recipes/r-ngsplot/build.sh
```diff
@@ -2,10 +2,17 @@
...
-done
\ No newline at end of file
+done
+
+#Activate/Deactivate dir
+ACTIVATE_DIR=$PREFIX/etc/conda/activate.d
+DEACTIVATE_DIR=$PREFIX/etc/conda/deactivate.d
+mkdir -p $ACTIVATE_DIR
+mkdir -p $DEACTIVATE_DIR
+cp $RECIPE_DIR/scripts/activate.sh $ACTIVATE_DIR/ngsplot-activate.sh
```
mbargull: In regards to the comment below for `scripts/activate.sh`:
Replace
```sh
cp $RECIPE_DIR/scripts/activate.sh $ACTIVATE_DIR/ngsplot-activate.sh
```
by something like
```sh
echo > "${ACTIVATE_DIR}/${PKG_NAME}-activate.sh" <<EOF
#!/bin/bash
export NGSPLOT="${outdir}"
export _CONDA_SET_NGSPLOT_ENV=1
EOF
```
. `${outdir}` will then have been evaluated at build time (becaus of using `<<EOF` with `EOF` being unquoted) and replaced by the correct output path (which `conda` will in turn adjust to the environment's path at install time).

mbargull:
Ah, I feel the urge to clarify the

> Choices you have for this

The different approaches are of course not necessarily exclusive -- "hero" is always something you can try to achieve :smile:.

The "don't care" approach is not really preferable since it leaves the container in an incomplete/not directly usable state: The user would have to "guess"  the path and set the variable themselves to make use of the container.

The "defensive patch" approach *might* be overly wary, but it retain's the software's functionality to let the user choose/set an alternative path. Probably that variable is just used as a workaround to achieve relocatability so that functionality isn't needed if you patch and let `conda` do this job -- BUT I don't know the program/scripts/uses at all and might be wrong about this. (As a general note, patching should always be done carefully, especially if it's changing the sofware's behavior. But it could be the best solution in this case.)

recipes/r-ngsplot/build.sh
apeltzer: Done :-)

apeltzer:
Thanks for all these hints - never had that much trouble with a single recipe than this one (which is tbh in a bad shape for packaging...). 

I'll patch the call in the script I'm calling here in `build.sh` and the tool should then be able to function.
```
progpath <- Sys.getenv('NGSPLOT')
if(progpath == "") {
```
As far as I see it, this is the best way other than the hero approach, which I don't feel qualified to do.

apeltzer:
So replaced the NGSPLOT ENV call in the R script. Lets hope that works fine then :-)

mbargull:
Did a quick grep for `NGSPLOT` and it seems the code pattern for that is quite uniform. If the commit I added work, we can leave out the activation script entirely. Let's see how it goes

mbargull:
> Let's see how it goes

Didn't quite work yet... debugging...

> Thanks for all these hints - never had that much trouble with a single recipe than this one (which is tbh in a bad shape for packaging...).

(I must've missed that comment before.) You're welcome, happy to help. Yeah it's a bit troublesome that it's not packaged as usual, but hey, you got to learn a few new tricks which is never a bad things :wink: .

mbargull:
I think the last changes might make this pass. If it indeed does and you still think the patching is fine, feel free to merge!

recipes/r-ngsplot-hg19/meta.yaml
```diff
@@ -0,0 +1,20 @@
...
+  run:
+    - r-ngsplot
+    - r-base
+    - wget
+test:
+  commands:
+    - (ngsplotdb.py list | grep -qF "hg19" || ngsplotdb.py list)
+about:
+  home: 'https://github.com/shenlab-sinai/ngsplot'
+  license: GPL-2.0
```
mbargull: I wonder: does the database have another home URL? Is the database itself also licensed under the `GPL-2.0`?

apeltzer: No, they are on GDrive and linked from the same Github repository - so I think its fine to use the same licence (as they didn't state explicitly another one). 

Thanks for all the help - learned a lot here and will add two more databases today to finalize things here :-)

PR #9174, opened 2018-06-06
Remove abruijn; it was renamed Flye

corburn:
* [] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).

The developer of Abruijin renamed it Flye.
https://github.com/fenderglass/Flye/blob/flye/README.md#flye-assembler-successor-of-abruijn


peterjc:
Is there a new recipe for Flye planned on another pull request, or can we rename and update this recipe? Would that give a useful git history?

corburn:
Flye was introduced in this commit:

```
commit b04050f19e3f5b3b5ba4fa5cbca21ee09dca8e1f
Author: Andreas Sjödin <andreas.sjodin@gmail.com>
Date:   Sat Jan 6 00:54:57 2018 +0100

    Add flye (#7207)

    * Add flye

    * Add back llvm
```

peterjc:
Thanks - b04050f19e3f5b3b5ba4fa5cbca21ee09dca8e1f from #7207 - I should have searched.

It seems unlikely then that Abruijin v1.x will get any more updates; I'm unclear if that means BioConda should keep the recipe or not. I'm going to defer to the BioConda core team on this.

kyleabeauchamp:
Yeah, I'm not sure if we need to delete the recipe in this case...

druvus:
I added both packages and back then I decided to keep the abruijn recipe. However I don't have strong opinion so it fine to remove it for me. I think we should come up with some recommendation how to handle cases when software are renamed.

mmiladi:
@druvus @kyleabeauchamp I am in the favor of renaming/removing the recipe. Since the original repository has been renamed, not cloned/forked, even the url within the recipe https://github.com/fenderglass/ABruijn/ does not exist and redirects to Flye, the old versions of tarball as well.

corburn:
Rebased onto master.

PR #9180, opened 2018-06-06
Update hifive to v1.5.6

corburn:


peterjc:
Without the upstream project providing a change log this is hard to judge, but is there good reason it would be worth packaging v1.5.4 and 1.5.5 here rather than skipping them?

jxtx:
@peterjc I 1.5.5 was a bugfix release and 1.5.4 is mostly related to QuASAR. I don't think there is a reason to package them.

peterjc:
Thanks, if anyone does ever need 1.5.4 we can package that on request.

PR #9191, opened 2018-06-06
Update last to 941

npavlovikj:


npavlovikj:
@bioconda/core , there is one bash script in this package, "maf-sort", that uses "getopts". However, "getopts" is not available in the docker container created with "mulled-test", so the build fails. I was wondering how I can proceed with this, so any suggestions are appreciated.

daler:
@npavlovikj thanks for updating this. You can try using the extended image which has getopts: http://bioconda.github.io/troubleshooting.html#using-the-extended-image.

Also, in general when updating a recipe, it's fine to only keep the latest version. All earlier packages that have been built will remain in the channel, and their recipes are included in the conda package. So here you can just edit `recipes/last/meta.yaml` in place.

npavlovikj:
Thank you so much @daler , I will give the extended image a try. 
And just to verify your other comment - does this mean I don't need to create a directory for the older version, but just update the current meta.yaml ?

daler:
Looks like the extended container worked, good.

And right -- no need to create a subdirectory for the older version, just update the current meta.yaml to reflect the new version.

npavlovikj:
Thanks for the help and useful suggestions @daler !
I removed the directory now, and once the build passes, I will go ahead and merge the recipe if ok.

mbargull:
> @bioconda/core , there is one bash script in this package, "maf-sort", that uses "getopts". However, "getopts" is not available in the docker container created with "mulled-test", so the build fails.

That's a very strange thing, indeed. To give context for the issue: `getopt` is a normal binary, i.e., a program that can live on its own, whereas `getopts` is a shell builtin function. A shell might not implement that function but most moderately modern ones should.
Apparently `/bin/sh` from the smaller container -- which is `ash` provided by BusyBox -- is (for whatever reason...) built without support for some shell functions (even `help` is not available??). It's quite strange since BusyBox normally (even in very old versions) supports `getopts` & co.
A fix for this in the container would be to just to `rm /bin/sh && ln -s /bin/bash /bin/sh` since `bash` is also included in the container. But I think we would want to update the container at some point anyway. I started to put together containers that use more recent versions of `glibc`/`busybox`/`bash` a while back -- but I still have to do some final polishing up on those, yet.

PR #9198, opened 2018-06-07
Update bbmap to 38.06

npavlovikj:


npavlovikj:
@olgabot , I believe you can override the path by using the option "**path**", e.g. **removehuman.sh in=<input file> path=<path to an indexed human genome> outu=<clean output file>**. However, it looks like "**removehuman.sh**" requires some other files to be present in the "_path_" directory as well, so you may need to perform some previous steps of "_bbmap_" to get those files.
I believe these suggestions should work, but if that is not the case, please let me know and I can look into possibly modifying the "_bbmap_" recipe.

PR #9199, opened 2018-06-07
Update trinotate to 3.1.0 and 3.1.1

biocyberman:

### DIFF
--- a/recipes/trinotate/meta.yaml
+++ b/recipes/trinotate/meta.yaml
@@ -10,7 +10,7 @@ build:
 
 package:
   name: trinotate
-  version: '3.0.2'
+  version: '3.1.0'
 
 requirements:
   build:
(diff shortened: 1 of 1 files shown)


peterjc:
I believe that you need to do this as two pull request, first the 3.1.0 update, then once that is merged, built, and uploaded to anaconda.org you can do the 3.1.1 update.

If we just merge this as is, I believe the 3.1.0 update will never be processed.

biocyberman:
@peterjc Thanks for the explanation. I was not clear about this.  Will do a reset

peterjc:
Great - merging now for the 3.1.0 update...

peterjc:
You can prepare the new pull request now (I assume you kept the changes locally so that ought to be quick). Thanks.

PR #9202, opened 2018-06-07
Recipe for Trimmomatic 0.32, based on that for 0.33

peterjc:

Rather than an update per se, this adds a package for an older release (which I want to use to reproduce a published method).


peterjc:
Thank you - and thanks @tomkinsc too.

PR #9237, opened 2018-06-08
Update dDocent

jpuritz:

### DIFF
--- a/recipes/ddocent/build.sh
+++ b/recipes/ddocent/build.sh
@@ -6,7 +6,6 @@ git clone https://github.com/lh3/seqtk.git
 cd seqtk
 make
 cp seqtk $PREFIX/bin
-
-bash install_dDocent_requirements $PREFIX/bin
+bash ../install_dDocent_requirements $PREFIX/bin
 
 chmod +x $PREFIX/bin/dDocent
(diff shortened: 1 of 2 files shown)


bgruening:
@jpuritz can you explain why this `git clone` is necessary? Can we not update the seqtk recipes as well?

jpuritz:
I thought it was Bioconda policy to not update recipes for master branch
versions of software, and to only use stable releases.

dDocent now requires the latest version of seqtk, so I thought it was
better to update then build this way, so other seqtk users would not be
affected.

I can update seqtk if that is preferred. This version of seqtk is over a
year old, so it’s not exactly unstable.

-- 
Jon Puritz

bgruening:
Oh, so dDocent needs a unreleased version of seqtk? I saw your issue, lets see if we get a new release of seqtk in the next days.

jpuritz:
@bgruening Yes, that is correct.  I put in the issue, and I just submitted a pull request with the new version.  However, it seems like there was a previous request for a new version from January this year, so I am not sure waiting a few days will help.  In the mean time, I moved the download of the unreleased version to the internal `dDocent` install script.  This gives a passing build with only removing `seqtk` from the `meta.yaml`.  Hopefully, this is an acceptable, temporary workaround.  I will update the `dDocent` recipe as soon as a new version of `seqtk` is released.

bgruening:
Thanks!

PR #9238, opened 2018-06-08
atactk recipe

egeogretmen:

The test for the installed script called "trim_adapters" is "trim_adapters --help" which returns exit code 0 when installed, however it fails during CircleCI builds.

```
+ trim_adapters --help
/opt/conda/conda-bld/atactk_1528493079303/test_tmp/run_test.sh: /opt/conda/conda-bld/atactk_1528493079303/_t_env/bin/trim_adapters: /opt/conda/conda-bld/atactk_1528493079303/_t_env/bin/python: bad interpreter: No such file or directory
```

The shebang is `#!/usr/bin/env python` so it shouldn't be causing a problem. Does anybody have an idea what the problem might be?

### DIFF
--- a/recipes/atactk/build.sh
+++ b/recipes/atactk/build.sh
@@ -1,2 +1,0 @@
-#!/bin/bash
-pip install .
--- a/recipes/atactk/build.sh
+++ b/recipes/atactk/build.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+pip install ./atactk
--- a/recipes/atactk/build.sh
+++ b/recipes/atactk/build.sh
@@ -1,2 +1,2 @@
 #!/bin/bash
-pip install ./atactk
+pip install .
--- a/recipes/atactk/meta.yaml
+++ b/recipes/atactk/meta.yaml
@@ -27,6 +27,8 @@ test:
 extra:
     recipe-maintainers: 
         - egeogretmen
+    notes:
+        - Adds 3 scripts, namely "trim_adapters", "make_cut_matrix" and "plot_aggregate_matrix.R"
 
 about:
     home: http://theparkerlab.org/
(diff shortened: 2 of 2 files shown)


recipes/atactk/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+build:
+    number: 0
+
+requirements:
+    build:
+        - python
+        - pysam
+        - python-levenshtein
+        - sexpdata
```
bgruening: runtime dependencies are missing

recipes/atactk/build.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+pip install .
```
bgruening: Use that here: https://github.com/bioconda/bioconda-recipes/blob/eaf59a144191c3989ee0c2bd0d799fb0a9281f3e/recipes/bracken/build.sh#L5 for your scripts.

recipes/atactk/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+test:
+    commands:
+        - trim_adapters --help
+        - make_cut_matrix --help
+
+extra:
+    recipe-maintainers: 
+        - egeogretmen
+    notes:
+        - Adds 3 scripts, namely "trim_adapters", "make_cut_matrix" and "plot_aggregate_matrix.R"
```
bgruening: test all 3 scripts

egeogretmen: Is there a way to test Rscripts easily? They have to be explicitly coded to print a help text I guess. How can I test [this](https://github.com/ParkerLab/atactk/blob/master/scripts/plot_aggregate_matrix.R) Rscript?

bgruening: Just import the R modules: RColorBrewer, ggplot2, grid, gtools. That should be enough, you also need to depend on this as runtime dependency.

See here: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/r-gbm/meta.yaml#L37

jerowe:
Thanks!

PR #9244, opened 2018-06-10
Update: Cromwell to latest release (32)

chapmanb:

### DIFF
--- a/recipes/cromwell/meta.yaml
+++ b/recipes/cromwell/meta.yaml
@@ -1,5 +1,5 @@
 {% set version="32" %}
-{% set revision="8ef1196" %}
+{% set revision="32" %}
 about:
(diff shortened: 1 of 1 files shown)


bgruening:
@chapmanb should the package/version not also be 32 instead of 0.32?

chapmanb:
Björn -- I didn't initially set it in bioconda to be 0.32 (rather than 32) but like this approach in case they ever change the versioning scheme and have a 1.0 release. This would make it a pain for selects since 32 > 1. I've run into this in the past with tools switching from no versions (and us using year.mm.dd versions) then converting to versions. Is it okay to leave it like this or do you have strong objections?

bgruening:
Ok, makes sense, just wanted to double check.

PR #9268, opened 2018-06-11
Heinz: a gene expression network analysis tool, new to Bioconda

CicoZhang:

This is a new branch, but the same pull with #8926 which I have messed up with :-(.


recipes/heinz/meta.yaml
```diff
@@ -0,0 +1,34 @@
+{% set version = "2.0" %}
```
bgruening: 2.0.1?

CicoZhang: I define the version 0.0.1 as the changes of the dependent library due to the process of compilation, which is irrelevant to the functionality and performance of Heinz. Especially, 2.0.1 stands for CPLEX 12.7; whereas 2.0.0, 12.8.
2.0 is still its main version number of Heinz.

CicoZhang: @bgruening  To avoid the confusion in future, I have made the change to keep version number consistent today. :smiley:

recipes/heinz/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+  sha256: b3f94a703552387c18bbf6d42096ac858046b56b23b5792763d7cc98f27eeeb1
+
+build:
+  number: 0
+  skip: True # [osx]
+
+requirements:
+  build:
+    - gcc # [not osx]
+    - llvm  ==3.3 # [osx]
```
bgruening: is this strict pinning really needed?

CicoZhang: I have tried `gcc` on MacOS, but it does not work. I check the MakeFile, where the commands are for clang `-stdlib=libstdc++`; I guessed llvm might work, but llvm 4.0 does not, then I considered an older version, but I have not tested yet. But I decided to skip MacOS for now.

recipes/heinz/meta.yaml
CicoZhang: @bgruening  Still, it failed on MacOS with the following error message. Considering Heinz is mainly deployed via Galaxy, I will skip MacOS for now until the compilation issue is solved. 
I have just updated the recipe to skip MacOS and removed everything about it.
```
ld: symbol(s) not found for architecture x86_64
make[2]: *** [print] Error 1
make[1]: *** [CMakeFiles/print.dir/all] Error 2
make: *** [all] Error 2
BUILD START: heinz-2.0-0
    (actual version deferred until further download or env creation)
```

bgruening:
Ok, thanks!

PR #9336, opened 2018-06-12
Flexbar 3.3

bwlang:

### DIFF
--- a/recipes/flexbar/3.3.0/build.sh
+++ b/recipes/flexbar/3.3.0/build.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+cmake .
+make
+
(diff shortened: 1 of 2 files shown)


bwlang:
Is it a correct practice to add an explicit version as I've done here? or better to just maintain a single meta.yaml files (like cutadapt)?

bgruening:
Up to the maintainer. Keep in mind that if we rebuild the package, e.g bumping zlib, we will rebuild also all sub-packages.

bwlang:
given the api changes to the API, I see value in keeping the old version alive - so I think this is the right path and is ready to merge.

bgruening:
Feel free to merge with your new powers.

bwlang:
hah - thought that might be poor form to merge one's own PR

PR #9344, opened 2018-06-13
Update breseq to v0.32.1

corburn:


jerowe:
```
libtool: compile:  g++ -DHAVE_CONFIG_H -I. -I../../../aux_build -O2 -std=c++11 -I../../../extern/samtools-1.3.1 -I../../../extern/samtools-1.3.1/htslib-1.3.1 -I../../../aux_build -DDATADIR=\"../share/breseq/\" -g -O2 -MT alignment.lo -MD -MP -MF .deps/alignment.Tpo -c alignment.cpp  -fPIC -DPIC -o .libs/alignment.o
In file included from ./libbreseq/storable.h:25:0,
                 from ./libbreseq/settings.h:24,
                 from libbreseq/fastq.h:23,
                 from libbreseq/alignment.h:25,
                 from alignment.cpp:19:
./libbreseq/json.hpp:67:10: error: #error "unsupported GCC version - see https://github.com/nlohmann/json#supported-compilers"
         #error "unsupported GCC version - see https://github.com/nlohmann/json#supported-compilers"
          ^
make[1]: *** [alignment.lo] Error 1
make[1]: Leaving directory `/opt/conda/conda-bld/breseq_1528903419845/work/breseq-0.32.1/src/c/breseq'
```

There are a few different ways you can deal with recipes that require samtools and co, and it depends on what the recipe needs.

If it needs to just link against the library you can include samtools as a build/runtime dep, and patch the makefile to tell it not to try and compile samtools.

If it needs some file from samtools/htslib distros that are not included in the conda package (it happens sometimes), you will have to go through, find all the makefiles for samtools and co, and make sure they match the makefiles in bioconda.

The first option is less painful and preferable when it can be used. ;-)

jerowe:
https://github.com/nlohmann/json#supported-compilers

jerowe:
https://github.com/nlohmann/json/pull/212

The json library this require is not compatible with gcc 4.8, which is what we have. I think this version of gcc has some backports that may fix whatever it doesn't like, so maybe add a patch to the json.hpp so that it won't throw that error? That is a total hack, but it may work.

jerowe:
Possibly path libbreseq/json.hpp ?
```
./libbreseq/json.hpp:67:10: error: #error "unsupported GCC version - see https://github.com/nlohmann/json#supported-compilers"
         #error "unsupported GCC version - see https://github.com/nlohmann/json#supported-compilers"
```

jerowe:
This needs to be patched:

https://github.com/barricklab/breseq/blob/aa8f2595a244af15b721411475ba07bfa12c8634/src/c/breseq/libbreseq/json.hpp#L65

I think it might work with the linux-anvil gcc. It has some backports from 4.9 to fix stuff (I think).

PR #9356, opened 2018-06-13
Compile lorikeet-19 with scala 2.11.8 to match atk build

pvanheus:


pvanheus:
This is a workaround for a bug in the compilation of lorikeet-19: the new version of lorikeet relies on atk-127.jar that is compiled with scala 2.11.8. Since lorikeet-19 is compiled with scala 2.10.2, this causes a runtime error. The fix is to compile lorikeet-19 with scala 2.11.8 too. This PR provides a workaround at the cost of additional build logic.

PR #9390, opened 2018-06-14
Perl sys cpu

jupollet:


jupollet:
Hi , 
can I have some help with this package?
I have to modify an existing recipe find in anaconda cloud, because the "conda skeleton" method don't work to write this perl module recipe.

Thanks

PR #9419, opened 2018-06-17
fix zifa missing_hash

corburn:

### DIFF
--- a/recipes/zifa/meta.yaml
+++ b/recipes/zifa/meta.yaml
@@ -11,7 +11,7 @@ source:
 
 build:
   noarch: python
-  number: 2
+  number: 1
 
 requirements:
   host:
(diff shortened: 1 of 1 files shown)


recipes/zifa/meta.yaml
```diff
@@ -1,19 +1,22 @@
...
+  version: '{{version}}'
 
 source:
   url: https://github.com/epierson9/ZIFA/archive/e40fdbf6c41cd7e564ab03d242fc071f5ac543b7.zip
+  sha256: '{{sha256}}'
 
 build:
   noarch: python
-  number: 1
+  number: 2
```
bgruening: @corburn if this package wasn't build yet due to our huge backlog a increase of the build-number might be not needed. Also if you just fix linting errors, a rebuild of the package is imho also not needed, which also means not increasing the build number.

corburn: @bgruening I'll keep that in mind.

PR #9421, opened 2018-06-17
fix yaggo missing_hash

corburn:

### DIFF
--- a/recipes/yaggo/meta.yaml
+++ b/recipes/yaggo/meta.yaml
@@ -10,7 +10,7 @@ source:
   sha256: {{ sha256 }}
 
 build:
-  number: 0
+  number: 1
 
 requirements:
   build:
--- a/recipes/yaggo/meta.yaml
+++ b/recipes/yaggo/meta.yaml
@@ -3,14 +3,14 @@
 
 package:
   name: yaggo
-  version: '{{version}}'
+  version: {{ version }}
 
 source:
-  url: https://github.com/gmarcais/yaggo/releases/download/v{{version}}/yaggo
-  sha256: '{{sha256}}'
+  url: https://github.com/gmarcais/yaggo/releases/download/v{{ version }}/yaggo
+  sha256: {{ sha256 }}
 
 build:
-  number: 2
+  number: 0
 
 requirements:
   build:
@@ -19,7 +19,7 @@ requirements:
 
 test:
   commands:
-    - yaggo -h 2>&1 | grep Usage > /dev/null
+    - (yaggo -h 2>&1 | grep -qF Usage || yaggo -h)
 
 about:
   home: https://github.com/gmarcais/yaggo
--- a/recipes/yaggo/meta.yaml
+++ b/recipes/yaggo/meta.yaml
@@ -19,7 +19,7 @@ requirements:
 
 test:
   commands:
-    - (yaggo -h 2>&1 | grep -qF Usage || yaggo -h)
+    - (yaggo -h 2>&1 | grep -qF Usage || (yaggo -h && false))
 
 about:
   home: https://github.com/gmarcais/yaggo
--- a/recipes/yaggo/meta.yaml
+++ b/recipes/yaggo/meta.yaml
@@ -14,8 +14,8 @@ build:
 
 requirements:
   run:
-    - ruby
-    - jemalloc
+    # artificial version constraint to avoid the outdated bioconda::ruby
+    - ruby >2.2.3
 
 test:
   commands:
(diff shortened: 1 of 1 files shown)


mbargull:
Hm... I don't understand why the container test failed for the build of a5305b6ba5e4067e88bc3279b3c4008ad33a6051 but passed for 7f65ef935a5325bfaeef5a84d07f743ae542189c. The changes in the latter commit are only meant to show the output in case of a failure.

corburn:
@mbargull I don't think the tests are running at all since you reset the build number:
```
21:28:08 BIOCONDA INFO FILTER: not building yaggo-1.5.9-0.tar.bz2 because it is in channel(s) and it is not forced.
21:28:08 BIOCONDA INFO Nothing to be done for recipe recipes/yaggo
```

mbargull:
Yes, thanks! Meant to type a 1 but did a 0, probably out of habit... Now let's finally see what the failing test looks like

recipes/yaggo/meta.yaml
```diff
@@ -14,8 +14,8 @@ build:
 
 requirements:
   run:
-    - ruby
-    - jemalloc
+    # artificial version constraint to avoid the outdated bioconda::ruby
+    - ruby >2.2.3
```
mbargull: This is a ugly workaround but should be fine, IMO.
I'll work on a better solution either in Bioconda or in `conda` soon, I guess.

bgruening: This is what we recommend, if we can not remove the package from our channel. Thanks @mbargull!

mbargull: At some point we should discuss whether we want to change the channel priority from `[bioconda, conda-forge, defaults]` to `[conda-forge, bioconda, defaults]`. That would solve quite a few issues (including "solver getting stuck" ones). It of course has implications if we had package name conflicts with `conda-forge` (but thankfully we lint for that on `bioconda` as well as `conda-forge`).
The other possibility would be for me to implement (or persuade someone else to do so :wink:) some kind of selective channel priority, i.e., being able to specify `[[bioconda, conda-forge], defaults]` as a channel priority list (of lists) that expresses that `bioconda` and `conda-forge` have the same prio. Meaning like `--no-channel-priority`, but selectively only for `bioconda` and `conda-forge`.

bgruening: I'm all in for the first one, but you need to discuss this with @johanneskoester, he had objections last time against this :)
The second idea is problematic, as during install time you need to know this, which is not ideal for our users imho.
We could also just mark certain packages as deprecated ...

mbargull: Yeah, I'd definitely want some more opinions from the core team on this -- because I can get behind the reservations on this one. And it's furthermore a "hard to deploy" change since we are also talking about telling the user to yet again change their config...

Though, I'm not yet seeing how that "selective equal channel priority" thing is more problematic than telling the user to use `[bioconda, conda-forge, defaults]` or `[conda-forge, bioconda, defaults]`. All of those are a one-time configuration thing (or memorizing two CLI flags if they don't use a config).

bgruening: Ah your second comment was not about a per recipe change. Mh, have you tried equal priorities? What happens in the case of openjdk than? This sounds dangerous random to me :)

mbargull: > What happens in the case of openjdk than? This sounds dangerous random to me :)

Exactly because of that I wouldn't want to use `--no-channel-priority` :wink:. I was thinking about being able to set equal priorities for a *subset* of channels, i.e., `bioconda` and `conda-forge` only. This functionality is not yet available in `conda`, but I have some different use cases in mind (e.g., `(conda-canary, defaults)`) where it would make sense to have it.
Currently `conda`'s solver (roughly) sets priority via
```python
(channel_prio, version, build_number, ...)
```
or if you provide `--no-channel-priority`, it's
```python
(version, channel_prio, build_number, ...)
```
. What I want is something like
```python
(main_channel_prio, version, equi_channel_prio, build_number, ...)
```
with, e.g.,
```python
main_channel_prio = {('bioconda', 'conda-forge'): 0, ('defaults',): 1}
equi_channel_prio[('bioconda', 'conda-forge')] = {'bioconda':0, 'conda-forge': 1}
```
if that makes sense to you.
Deep down in the solver's code it should be fairly simple to implement. But I'm not sure what other parts of `conda`'s code are affected by this (e.g., `subdir`s being treated as subchannels or something). I only had the idea a while back but didn't look into it yet. I'll have to ask @kalefranz some time, I think.

PR #9430, opened 2018-06-17
merge bulk into master

bgruening:


kyleabeauchamp:
Quick question: in this PR, intervaltree was identified as something that was moved to conda-forge.  I understand that the recipe was therefore deleted.  However, I also noticed that the bioconda package was moved to https://anaconda.org/bioconda-legacy/intervaltree/files (and deleted from https://anaconda.org/bioconda/intervaltree/files).  Is this the intended workflow for any package that gets moved to conda-forge?  

I think I may recall that this is necessary to sidestep some conda DAG precendence issues, but it also does have the consequence of requiring older frozen environments to be updated with the new URL (as in below diff).

```
-https://conda.anaconda.org/bioconda/linux-64/intervaltree-2.1.0-py27_0.tar.bz2
+https://conda.anaconda.org/conda-forge/linux-64/intervaltree-2.1.0-py27_0.tar.bz2
```

kyleabeauchamp:
cc @jtratner

bgruening:
We try to avoid this, we can also mark the as deprecated, but sometime we need to move them over to the other channel.

kyleabeauchamp:
Thanks!  I wonder if we can come up with a simple guide to help folks who are looking for longer-term reproducibility.  E.g., perhaps we should add a section to the FAQ on long-term reproducibility and the various caveats + workarounds.

bgruening:
https://www.biorxiv.org/content/early/2017/10/11/200683 :)

But this is probably not what you are looking for. The correct solution would be to migrate the old versions also to conda-forge. I have done this yesterday with gawk (https://github.com/conda-forge/gawk-feedstock/pull/6), but its a lot of work :(

kyleabeauchamp:
Yeah, I guess I agree with virtualization/containerization for long-term reproducibility, so I guess I have to correct my previous statement and say I'm probably talking about "intermediate timescale" reproducibility, where we have pipelines that are evolving in time but still would still benefit from static behavior of upstream packages.

I think we might be able to address medium and long-term reproducibility in the Bioconda FAQ.  I might take a stab at it.

PR #9456, opened 2018-06-18
Try to fix wham build

kyleabeauchamp:


kyleabeauchamp:
I think the problem here is that the zlib / env INCLUDE path is not being propagated to the recursive make of bamtools.

corburn:
Resolved by #9365

PR #9464, opened 2018-06-19
Fix vispr

corburn:

### DIFF
--- a/recipes/vispr/meta.yaml
+++ b/recipes/vispr/meta.yaml
@@ -9,16 +9,16 @@ source:
   sha256: dc2cf6ac9c8930b0f1f1b3a2b7f57bfae180b2ac3674060123fd2d7fe10ccd82
 
 build:
-  skip: True # [py27]
+  noarch: python
   number: 2
   entry_points:
     - vispr = vispr.cli:main
 
 requirements:
   host:
-    - python
+    - python >=3
     - setuptools
-    - numpy
+    - numpy >=1.10
     - flask
     - pandas >=0.17.1,<0.20
     - pyyaml
@@ -26,8 +26,8 @@ requirements:
     - scikit-learn
     - appdirs
   run:
-    - python
-    - numpy
+    - python >=3
+    - numpy >=1.10
     - flask
     - pandas >=0.17.1,<0.20
     - pyyaml
@@ -41,5 +41,6 @@ test:
 
 about:
   home: https://bitbucket.org/liulab/vispr
-  license: MIT License
+  license: MIT
+  license_family: MIT
   summary: VISPR is a visualization framework and analysis workflow for CRISPR/Cas9 knockout screens. VISPR is designed to display results calculated by MAGeCK.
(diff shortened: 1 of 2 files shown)


recipes/vispr/meta.yaml
```diff
@@ -9,25 +9,25 @@ source:
   sha256: dc2cf6ac9c8930b0f1f1b3a2b7f57bfae180b2ac3674060123fd2d7fe10ccd82
 
 build:
-  skip: True # [py27]
+  noarch: python
```
bgruening: noarch python should only be used by pure python packages, numpy is not noarch and not pure python my understanding is that this is not working. Maybe @mbargull knows more?

mbargull: A `noarch` package can have non-`noarch` dependencies, no problem. A good example is the dependency on `python` itself, which is not `noarch` :wink:.
The package itself seems to be pure Python, so this looks good.

recipes/vispr/meta.yaml
```diff
@@ -9,25 +9,19 @@ source:
...
-    - pyyaml
-    - scipy
-    - scikit-learn
-    - appdirs
+    - pip
   run:
-    - python
-    - numpy
+    - python >=3
+    - numpy >=1.10
```
mbargull: @corburn, is there any reason you added those constraints?
Acc. to https://bitbucket.org/liulab/vispr/src/03dd84177d431f82943bec0b25e1540d4e674e6e/setup.py this should be Python 2 compatible (although it was `skip`ped before in the recipe) and no minimum `numpy` version is specified.

mbargull:
Oh and BTW, @corburn, I don't think I mentioned it before, but: Thanks for all the help! Really appreciated!

recipes/vispr/meta.yaml
mbargull: For Python projects I usually just take a look at the `setup.py` and `setup.cfg` (+ `requirement.txt` if that one is used in `setup.py` and the license file of course).
If `setup.py` has no platform-specific code (including in requirement specifiers) and doesn't build any extensions, the package should usually be pure Python and thus be likely eligible for `noarch: python`.
(Of course there can be cases where `noarch: python` isn't really true, e.g., when \[only for a subset of all covered systems\] some system libraries are dynamically loaded or system-specific file system accesses happen -- but that would require knowing/reading the whole source code ;P)

mbargull: And if the package's `setup.py` is nice (like it's the case for `vispr`) and only imports system modules but not itself or thirdparty packages, we can shrink the `build` and `requirements/host` section to
```yaml
build:

requirements:
  host:
   - python
   - pip

File: recipes/vispr/meta.yaml (hunk 2 above)
Comment (corburn) [reply recipes/vispr/meta.yaml]: @mbargull 

The python constraint replaced the `skip` in the recipe.
The numpy constraint resolved this error:

```
Returning non-final recipe for vispr-0.4.14-py36_2; one or more dependencies was unsatisfiable:^M                                                                                                                  
Build: None^M                                                                                                                                                                                                      
Host: The following specifications were found to be in conflict:^M                                                                                                                                                 
  - numpy=1.9^M                                                                                                                                                                                                    
  - pandas[version='>=0.17.1,<0.20'] -> numpy=1.10
```

PR #9466, opened 2018-06-19
multi-source orfm

corburn:


mbargull:
Nice one, thanks for trying out the multi-source thing! Looks much nicer this way :smile:.
I wonder why the linter didn't catch the missing checksum, though :thinking:.

mbargull:
> I wonder why the linter didn't catch the missing checksum, though :thinking:.

Ah, okay, because we didn't built yet :laughing:

PR #9471, opened 2018-06-19
New libMems recipe

jamesabbott:

This is a component of Mauve, but is available and distributed as a standalone library.

Mauve is a rather complex build which I have working (manually) under linux, however Clang++ does not like some definitions so osx builds are disabled for now. I'll try to sort these out once the whole linux build is working

Two patches are included with this recipe. An open bug report exists upstream for the first of these (https://sourceforge.net/p/mauve/bugs/46/), and I have opened a bug report for the second (https://sourceforge.net/p/mauve/bugs/53/). Bug 46 was opened nearly a year ago but has not been addressed.


bgruening:
Thanks James!

PR #9477, opened 2018-06-19
Add Schmutzi recipe

apeltzer:


apeltzer:
Almost impossible to do this:

The makefile pulls in `libgab` which has an own Makefile that relies on a hardcoded path to `bamtools`: 
`` 
BAMTOOLS= $(realpath ../bamtools/)
CXX      = g++ #-g  -pg
``
That would mean, I need to patch 

- the schmutzi Makefile
- the libgab Makefile, which is pulled inside the first schmutzi Makefile

@dpryan79 you were right, this is hard stuff....

grenaud:
Hi Alex,
The path to bamtools in libgab has to be one below. If this is problematic, we can always change things around. 

Gabriel

apeltzer:
If that would be possible Gabriel, that would make this feasible I guess.

Steps we'd need to have:

- [ ] Add a way to specify a custom bamtools path (e.g. one that is provided by the bioconda bamtools recipe (no need to recompile this in this case, slimmer in this case)
- [ ] Add a way to be able to specify the `libgab` dependency to also use the bamtools system-wide installation

If you could add these and make a new release on GitHub that would allow me to package this :)

PR #9493, opened 2018-06-20
Fix old builds of viennarna

dpryan79:


dpryan79:
@s-will This should fix building of the older versions of viennarna. Please note, however, that you don't need to keep every version around on github unless you expect to need to tweak them in the future. The packages will still be available via `conda install` regardless.

s-will:
Hey @dpryan79, thanks a lot for fixing the recipes!
Also thanks for the comment about the older recipes; you are right that most of them are likely not needed anymore.

dpryan79:
@bioconda/core Do we need to fix the `packages with divergent build strings` errors for `bulk` and `master`? In this case it was only the Linux build that was a problem, but of course fixing that is changing the build string on OSX too.

s-will:
If it helps / saves time, feel free to remove the failing older recipes

dpryan79:
@s-will Will do, that should make things easier :)

PR #9504, opened 2018-06-21
Vsearch v2.8.0 21

corburn:

Removed the 1.1.3 recipe because the CONDA_* variables triggered an error while building the 2.8.0 recipe locally.


colinbrislawn:
@corburn
@bgruening 

I feel like I might be asking a dumb question, but why was OSX skipped? What can we do to get OSX builds working again.

PR #9514, opened 2018-06-21
multiple fixes for build/host

bgruening:


bgruening:
@corburn, @dpryan79 can I get a review. I canceled the build I hope this PR improves recipes even if they do not compile and would like to merge it and build on bulk.

dpryan79:
@bgruening Yes! That should fix at least a quarter of the failures.

recipes/blat/meta.yaml
```diff
@@ -14,8 +14,6 @@ build:
 
 requirements:
   build:
-    - libpng
-    - zlib
   run:
```
dpryan79: Is this recipe even needed? It's probably identical to `ucsc-blat`.

bgruening: Probably not, I will remove it.

PR #9540, opened 2018-06-22
Update subread and fix real

dpryan79:


recipes/subread/build.sh
```diff
@@ -3,13 +3,21 @@
 mkdir -p "${PREFIX}/bin"
 mkdir -p "${PREFIX}/annotation"
 
+export C_INCLUDE_PATH="${PREFIX}/include"
 cd src
-MAKEFILE=Makefile.Linux
 if [ `uname` = "Darwin" ];
 then
   MAKEFILE=Makefile.MacOS
+  sed -i.bak "371" $MAKEFILE
```
bgruening: This does not seem to be proper OSX syntax.
Maybe we should just depend on sed as host-dependency.

dpryan79: Make that `31d`.

PR #9543, opened 2018-06-22
Libmems

jamesabbott:

There seem to be some problems linking against this library with boost 1.66, so pin to 1.65 which seems ok. There looks to be a bugfix in 1.67 which may resolve this once it is available.

There are also problems with linking if the boost libraries are not specified in a particular order, so modify the pkg-config setup to address this with a bit of sed...


jamesabbott:
merge pulled in commits from other branches...will revert and try again...

PR #9544, opened 2018-06-22
Update seqtk to version 1.3 and remove old builds

jpuritz:


kyleabeauchamp:
Do we need c++ (cxx) compilers for this?  Otherwise LGTM.

mbargull:
I changed this to use SHA256 instead of MD5 since the former is preferred.
But mostly I just wanted to issue a new build since the previous job ran for three hours... which is quite odd given that `seqtk` is a very small tool that should compile very fast.

jpuritz:
Thanks!  I’m not sure what the problem is, but my local tests stall as
well.

-- 
Jon Puritz, PhD

Assistant Professor
Department of Biological Sciences
University of Rhode Island
120 Flagg Road, Kingston, RI 02881

Webpage: MarineEvoEco.com

Email:
jpuritz@gmail.com

Cell:    401-338-8739
Work:  401-874-9020

"The most valuable of all talents is that of never using two words when one
will do.” -Thomas Jefferson


On June 25, 2018 at 2:00:44 PM, Marcel Bargull (notifications@github.com)
wrote:

I changed this to use SHA256 instead of MD5 since the former is preferred.
But mostly I just wanted to issue a new build since the previous job ran
for three hours... which is quite odd given that seqtk is a very small tool
that should compile very fast.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/bioconda/bioconda-recipes/pull/9544#issuecomment-400041536>,
or mute the thread

mbargull:
Ah, I see, thanks for confirming this locally. Okay, I'll see want to do, give me a sec

mbargull:
Hm... I can't see any obvious reason in https://github.com/lh3/seqtk/compare/v1.2...v1.3 why the tests would hang.. strange

jpuritz:
I really don't understand the problem.  Every time I have the compiler properly reference the ${PREFIX}/include and ${PREFIX}/lib directories, the build hangs forever.  I will let someone else with more C experience figure this out.

PR #9550, opened 2018-06-23
multiple fixes all over

bgruening:

### DIFF
--- a/recipes/abyss/meta.yaml
+++ b/recipes/abyss/meta.yaml
@@ -16,18 +16,19 @@ source:
 
 requirements:
   build:
-  - {{ compiler('c') }}
-  - boost
-  - sparsehash
-  - sqlite
-  - openmpi
+    - {{ compiler('cxx') }}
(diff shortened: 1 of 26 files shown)


recipes/abyss/meta.yaml
```diff
@@ -16,18 +16,19 @@ source:
...
-  - boost
-  - sparsehash
-  - sqlite
-  - openmpi
+    - {{ compiler('cxx') }}
+  host:
+    - boost
+    - sparsehash
+    - sqlite
+    - openmpi
```
sjackman: @bgruening Are these dependencies listed in `host` required to be installed on the end-user's computer that runs ABySS, or only on the machine that builds the package for ABySS? I read https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#host but wasn't quite clear on the answer to my question afterward.

sjackman:
Thanks for the explanation, Björn!

PR #9571, opened 2018-06-25
perl-sys-cpu

jupollet:


mbargull:
Looks like the problem was that a dependency, `perl-extutils-makemaker`, currently in `bioconda`, was not registered in https://github.com/conda-forge/staged-recipes/pull/6110/files#diff-e71e54a674d17ea23ea3ee6153720637R27 since it was commented out.

If I didn't miss anything else, only the following recipes need to be transferred to `conda-forge`:
```
perl-extutils-makemaker # dependency of perl-sys-cpu
perl-app-cpanminus      # dependency of perl-extutils-makemaker
perl-data-dumper        # dependency of perl-extutils-makemaker
perl-test-more          # dependency of perl-data-dumper
```
`perl` is already in `conda-forge` and `perl-threaded` is only an older/deprecated package for `perl`, so you needn't care about that one either.

So, all in all, looks like you're lucky and only need those four additional packages to fix https://github.com/conda-forge/perl-sys-cpu-feedstock :wink:.

jupollet:
Thank you for help.

PR #9582, opened 2018-06-26
ViennaRNA 2.4.8

martin-mann:

### DIFF
--- a/recipes/viennarna/2.4.7/meta.yaml
+++ b/recipes/viennarna/2.4.7/meta.yaml
@@ -1,0 +1,42 @@
+{% set version = "2.4.7" %}
+{% set sha256 = "3dbb6c6b54a372193d8106ee5f59442a0d7535838812a0aade09e632e599e4ee" %}
+
+package:
+  name: viennarna
+  version: {{ version }}
+
+build:
+  number: 5
+  string: "py{{CONDA_PY}}_{{PKG_BUILDNUM}}"
+
+source:
+  url: https://www.tbi.univie.ac.at/RNA/download/sourcecode/2_4_x/ViennaRNA-{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+  host:
+    - perl
+    - python
+    - zlib
+  run:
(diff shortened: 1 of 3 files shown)


recipes/viennarna/2.4.7/meta.yaml
```diff
@@ -0,0 +1,42 @@
+{% set version = "2.4.7" %}
```
mbargull: Do we need to retain the recipe for the older version? Any particular reason?

recipes/viennarna/meta.yaml
```diff
@@ -1,12 +1,12 @@
...
+{% set sha256 = "1f8a08b1c8e57e4c6610e0e48c66ae692caba4183867cde8eb58ec16c40f78fa" %}
 
 package:
   name: viennarna
   version: {{ version }}
 
 build:
-  number: 5
+  number: 0
   string: "py{{CONDA_PY}}_{{PKG_BUILDNUM}}"
```
mbargull: You can remove that `build/string` line.

recipes/viennarna/2.4.7/meta.yaml
martin-mann: mhh.. I am just updating (since I need the new version)

so far, the responsible people stored the old version's data in a subfolder, that's why I did the same.
ping: @s-will @jfallmann : is explicitely retaining the old version needed?

recipes/viennarna/meta.yaml
martin-mann: @s-will do yo need the package string extension for something?

s-will: Hi @mbargull, thanks for the review. The build string lets us build different packages for the different python versions, which we need to build python bindings for the respective version. Removing the build string thus breaks this functionality. Possibly this is quite a hack as well---would you suggest an alternative mechanism?

mbargull: `conda-build` will automatically take care of that. Either it creates build string with exactly that schema, e.g., `py36_0`, or something like `py36_h789abcd_0` when `python` is in `requirements/host` as well as `requirements/run`. With some older versions of `conda-build` there were few cases where the build strings had to be manually specified, but that shouldn't be needed for the current versions.

PR #9588, opened 2018-06-26
Update anvio to 5

corburn:


jmeppley:
@corburn, Was there a reason to drop the legacy builds (EG anvio 4) and the non-python nun dependencies (EG diamond)?

I don't have much of an opinion on the legacy builds, I'm just curious about the resoning.

I woiuld vote to restore the run dependencies. This let's users get up and running quicker as most of the dependencies are required to actually do anything. but maybe there is a good reason that I'm missing?

jmeppley:
Also...this update did nothing because Anvio is blacklisted because it uses conflicting numpy and pandas:

```
# conda.exceptions.UnsatisfiableError: The following specifications were found to be in conflict:
#   - numpy=1.9
#   - pandas==0.20.1 -> numpy=1.12
recipes/anvio
recipes/anvio/2.4.0
recipes/anvio/3.0.0
```

The circle ci tests never even tried to build the new anvio.

corburn:
@jmeppley I removed the legacy builds based on [discussions similar to this](https://gitter.im/bioconda/Lobby?at=5b274dbdc016730d86669c70):
> Usually the existing recipe gets updated when a new version of a package should be released. This will never mean a package on the channel is removed. It's just the recipe itself that gets updated.
With out current build infrastructure, having additional recipes per-version also mean we have to maintain those recipes, i.e., it introduces a maintenance burden.

You're right about the run dependencies. If no one gets to it before me, I'll fix it when I get home.

jmeppley:
OK, thanks.

Thanks for fixing the pip install. It make a TON more sense this way.

 I'm messing around with it right now. I loosend the pandas requirement to get around the blacklisting complaint. I should have a pull request up soon.

PR #9596, opened 2018-06-26
Add alignstats recipe

Lenbok:

This recipe currently fails the test section, but I don't understand why. If I disable the test section, the resulting package builds, installs, and runs fine. If I run the test command manually it runs fine with exit code 0. (I probably also have htslib specified as a requirement in more places than necessary)


Lenbok:
So it looks like the tests pass OK (not sure why it didn't work in my env). Please review the meta.yaml though, as it probably has some redundancy in how the htslib requirement is specified.

PR #9621, opened 2018-06-27
PeptideShaker updated to version 1.16.26. SearchGUI updated to version 3.3.3.

CarlosHorro:


CarlosHorro:
By the way, I see Travis is checking all old recipes and because of that the building takes a long time.

Would it be a good idea to remove very old recipe versions of both PeptideShaker and SearchGUI?
ie: PeptideShaker versions earlier than 1.16.0 and SearchGUI versions earlier than 3.2.3.

bgruening:
This is up to the recipe maintainer(s). So feel free to do so.

PR #9647, opened 2018-06-28
add suport for centroid rna package

SchwarzMarek:

This adds centroid RNA package (3 executables - centroid_fold, centroid_homfold and centroid_alifold).

I've build and installed it successfully, however I wasn't successful with running local tests due to docker problems.


dpryan79:
You'll either need to skip this on OSX or download the source code and actually build it (if you do that on OSX you might as well do it on Linux too).

bgruening:
This rebase went wrong can you please try again and clean the content of this PR?

SchwarzMarek:
Hello, I've done the rebase again and I hope that now it looks better.
Best regards

SchwarzMarek:
@bgruening I've added the doi identifiers.

recipes/centroid_rna_package/meta.yaml
```diff
@@ -0,0 +1,33 @@
+# building from source requiers also boost >= 1.4
+# this recipe uses working precompiled binaries
+package:
+  name: centroid_rna_package
+  version: 0.0.15
+
+build:
+  number: 0
+  skip: True # [osx]
```
bgruening: This should be `# [osx or py2k]`

recipes/centroid_rna_package/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+source:
+  sha1: e4318ffcd3bd4419a68c273fd924ec17a366c073
+  url: https://github.com/satoken/centroid-rna-package/releases/download/v0.0.15/centroid-rna-package-0.0.15-linux-x86_64.zip
+
+requirements:
+  run:
+    - viennarna >=1.8
```
bgruening: you need python as runtime requirement isn't it?

recipes/centroid_rna_package/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+test:
+  requires:
```
bgruening: remove the requirement and add a real test, for example exectuing your tool with `--help` or `--version`

recipes/centroid_rna_package/run_test.py
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env python3
```
bgruening: s/python3/python/

recipes/centroid_rna_package/meta.yaml
SchwarzMarek: Changed per request.

recipes/centroid_rna_package/meta.yaml
SchwarzMarek: No I don't need python at runtime.

recipes/centroid_rna_package/meta.yaml
SchwarzMarek: In conda documentation I found that the test are either executed with command inside meta.yaml or in form of run_test.[sh/py...].

I've initially not included the test in meta.yaml because the exit value of the --help is 1, so the test fails. I've searched on how this is usually handled and came with two options: 
1) check the exit value
I can't check exit value, because it is always 1, even when the program fails. 
2) check the return string
So I need to check the return string rather then exit code. I've written the check in python and that’s where the python test dependency comes in. The python is otherwise not needed.

I agree that it might not be the best idea and I propose to fix it with test in meta.yaml and checking the text with grep (assumes grep avalible).

recipes/centroid_rna_package/run_test.py
SchwarzMarek: Removed the whole file.

SchwarzMarek:
@bgruening I've done the requested changes. Comments are placed at the review comments.

PR #9655, opened 2018-06-29
fix swga 0.4.4

corburn:

My local build encountered the following error:
```
/usr/bin/g++ -o minia/SortingCount.o -c minia/SortingCount.cpp -O4 -fopenmp -DOMP=1
    minia/SortingCount.cpp: In function ‘void sorting_count(Bank*, char*, int, int, bool, int, bool)’:
    minia/SortingCount.cpp:547:57: error: ‘p’ has not been declared
     #pragma omp parallel for if (! separate_count) private (p)  num_threads(nb_threads)
                                                             ^
    make: *** [makefile:71: minia/SortingCount.o] Error 1
```

The patch file assumes the OpenMP private directive is not required because p is declared inside the pragma.

### DIFF
--- a/recipes/swga/meta.yaml
+++ b/recipes/swga/meta.yaml
@@ -15,7 +15,7 @@ build:
   entry_points:
     - swga = swga.main:main
   number: 1
-  script: python -m pip install --no-deps --ignore-installed . && cp swga/bin/* $PREFIX/bin/
+  script: python -m pip install --no-deps --ignore-installed .
 
 requirements:
   build:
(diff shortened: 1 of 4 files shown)


recipes/swga/meta.yaml
```diff
@@ -8,22 +8,26 @@ package:
...
   sha256: 28f6971234453f8ba91997c3203e71e94a4de5e37f7bcfd01096deb1fcb952c0
+  patches:
+    - SortingCount.cpp.patch
 
 build:
-  skip: true # [not py27]
+  skip: true # [osx]
   entry_points:
     - swga = swga.main:main
   number: 1
```
johanneskoester: Please also increase the build number here.

bgruening: @johanneskoester we increased all build numbers already, so in theory all packages should have a bump already included.

PR #9668, opened 2018-06-29
Remove transrate-tools from blacklist

luizirber:

Remove transrate-tools from blacklist. Also use `sha256` for hash, move the version to a template var, and add `zlib` as a build requirement.


bgruening:
Codefest PR merge count: 1
Awesome!

PR #9670, opened 2018-06-29
Fix rust-overlaps using the same fix from rust-bio-tools

### LINKED ISSUE #9266: unblacklist rust recipes
This is a tracking issue for progress with unblacklisting the two remaining rust recipes in the build blacklist in issue #7712, I.e. these lines (plus `rust-bio-tools`, which should probably need another look, as well, and `spalter` and `prosolo` can probably do with the same simultaneous changes):
https://github.com/bioconda/bioconda-recipes/blob/642ad05c02b9d23441108e09fee795390844aa90/build-fail-blacklist#L37-L39

The remaining issue boils down to some problem with the combination of `conda build` version `2.` and the `rust` compiler `cargo`, so the easiest is probably to just wait for the coming `bioconda-utils` update to `3.x` `conda build`:
https://github.com/bioconda/bioconda-utils/pull/293

The below infos are to document the previous attempt, but hopefully most of these workarounds will not be necessary any more. There is a branch I created with some attempts at getting these recipes unblacklisted, but it should probably be retired in favor of a clean new branch and should only cherry-pick relevant commits, if any:
https://github.com/bioconda/bioconda-recipes/tree/unblacklist_rust_recipes

The commits there incorporate the workarounds identified when unblacklisting prosolo (#7973) and further recommendations given for the new recipe for spalter that's on the way (#9062). But none of these got all the way and eventually we end up with an awkward interaction between the older version of `conda build` currently used (`2.1.18`) and `cargo`'s checking that certain paths are absolute paths:
```
thread 'main' panicked at 'assertion failed: src_path.is_absolute()', tools/cargo/src/cargo/core/manifest.rs:438:9
```

It goes away when I locally use `conda build` version `3.4.2`, hence the above suggestions to wait for the `bioconda-utils` update.

It does not go away when using older cargo versions, but the next future `rust`/`cargo` version (`1.27`) should include a change that would at least give a more informative panic message, should more debugging be needed in the future:
https://github.com/rust-lang/cargo/commit/dfdb9aab12c89257fbd529719265ac65609a55a3

luizirber:

Removes `rust-overlaps` from the blacklist, using the same fix from `rust-bio-tools`.

### DIFF
--- a/recipes/rust-overlaps/build.sh
+++ b/recipes/rust-overlaps/build.sh
@@ -1,4 +1,9 @@
 #!/bin/bash -e
 
+# circumvent a bug in conda-build >=2.1.18,<3.0.10
+# https://github.com/conda/conda-build/issues/2255
+[[ -z $REQUESTS_CA_BUNDLE && ${REQUESTS_CA_BUNDLE+x} ]] && unset REQUESTS_CA_BUNDLE
+[[ -z $SSL_CERT_FILE && ${SSL_CERT_FILE+x} ]] && unset SSL_CERT_FILE
+
 # build statically linked binary with Rust
 C_INCLUDE_PATH=$PREFIX/include LIBRARY_PATH=$PREFIX/lib cargo install --root $PREFIX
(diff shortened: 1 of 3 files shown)


luizirber:
I reverted the changes to `prosic` because it's failing for other reasons

recipes/rust-overlaps/build.sh
```diff
@@ -1,4 +1,9 @@
 #!/bin/bash -e
 
+# circumvent a bug in conda-build >=2.1.18,<3.0.10
```
mbargull: We should be able to remove these workarounds by now, actually. `conda-build >=3.0.10` shouldn't be adding those empty-value environment variables anymore.

PR #9675, opened 2018-06-29
FastQC: Update the openjdk dependency.

cbrueffer:

The "defaults" build was bumped, leading to the defaults version of openjdk be installed, instead of the conda-forge one.  See #5026. To prevent the next build bump causing this issue again, increment the minimum version by one.

### DIFF
--- a/recipes/fastqc/meta.yaml
+++ b/recipes/fastqc/meta.yaml
@@ -14,7 +14,7 @@ source:
 
 build:
   detect_binary_files_with_prefix: true
-  number: 3
+  number: 4
 
 requirements:
   run:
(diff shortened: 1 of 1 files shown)


recipes/fastqc/meta.yaml
```diff
@@ -14,15 +14,15 @@ source:
...
+  number: 4
 
 requirements:
   run:
     # Enforce a version requirement on openjdk to ensure
     # it comes from the conda-forge channel and not default.
     # Many yaks were shaved to bring us this information.
     # Version number reference: https://github.com/conda/conda/issues/6948#issuecomment-369360906
-    - openjdk >8.0.121
+    - openjdk >=8.0.122
```
mbargull: That won't work. We just need https://github.com/bioconda/bioconda-recipes/issues/5026#issuecomment-401472985.

cbrueffer: Yeah, I realized that as well after this didn't work ;-) Thanks for following up!

PR #9679, opened 2018-06-29
Maker: patch to allow using REPEATMASKER_LIB_DIR

abretaud:


recipes/maker/repeatmasker_check.patch
```diff
@@ -0,0 +1,11 @@
+--- lib/GI.pm	2018-06-29 20:53:50.113085044 +0200
++++ lib/GI.pm.new	2018-06-29 21:04:05.041345063 +0200
```
dpryan79: I think you want to change this to `lib/GI.pm`.

abretaud: ah yep, thanks, I will make another pr

PR #9686, opened 2018-06-29
sentieon: update to 201711.04

chapmanb:
Builds off of #9653 to avoid lint and build problems with cb3

### DIFF
--- a/recipes/sentieon/meta.yaml
+++ b/recipes/sentieon/meta.yaml
@@ -1,11 +1,11 @@
-{% set version="201711.03" %}
+{% set version="201711.04" %}
 
 package:
   name: sentieon
   version: {{ version }}
 source:
-  url: https://s3.amazonaws.com/sentieon-release/software/sentieon-genomics-{{ version }}.tar.gz  # [linux]
-  md5: d8f4fd7020d72917f5bdd15eb4e8f2a7  # [linux]
+  url: https://s3.amazonaws.com/sentieon-release/software/sentieon-genomics-{{ version }}.tar.gz
+  md5: 7ee8678c2c2ada3d9c34ccb277c6912a
   #url: https://sentieon.sharepoint.com/Release/_layouts/15/guestaccess.aspx?guestaccesstoken=xV%2bluck4cIhDqLGmeVzLT5p1%2b67SiarhwZqRUx%2fOK5w%3d&docid=0b3cf62ebe0fc4cc78615b5a1989b0108 # [osx]
   #md5: notavailable # [osx]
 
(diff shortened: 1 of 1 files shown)


Lenbok:
@chapmanb conda isn't showing this as available yet - is package building delayed or something?

chapmanb:
Len -- thanks for checking in on this, I'm not sure what happened here but guess that this somehow got delayed because of the big cb3 build going on. @bioconda/core -- does anyone have a clue how to kick this one to get this latest PR built?

rvalieris:
I am not sure why but, the logs say:
```
BIOCONDA INFO FILTER: not building recipe recipes/sentieon because the same number of builds are in channel(s) and it is not forced.
BIOCONDA INFO Nothing to be done for recipe recipes/sentieon
BIOCONDA INFO BUILD SUMMARY: successfully built 0 of 1 recipes
```

chapmanb:
I tried to bump the Sentieon build number to get this new package built but that still didn't do it. Bummed, I'm not sure what is going on here and would need help from Johannes or Bjorn or someone who understands what is going on with the cb3 build. My guess is that this is in the bulk rebuild somehow but I have no idea why it's not getting done and pushed.

mbargull:
TLDR; remove `or not py27` from the selector for `skip: True` and replace the `python` requirement by `python >=2.7,<3.0a0`.

----
The important build message is
```
BIOCONDA INFO Nothing to be done for recipe recipes/sentieon
```
which means `conda-build 3` didn't pick up a valid configuration for the build.
You can ignore
```
BIOCONDA INFO FILTER: not building recipe recipes/sentieon because the same number of builds are in channel(s) and it is not forced.
```
in this case (fixed that redundant message by https://github.com/bioconda/bioconda-utils/pull/313/files#r200166101).

What happened here is that `conda-build 3` is smarter in selecting the required build matrices. As `python` is only a `run` requirement here, `conda-build` does not build for all different `python` versions since the resulting package isn't affected by `python` at build time (`python` not being in `build`/`host`).
So it'll likely choose `python 3.6.*` currently. But due to `skip: True # [not linux or not py27]`, `[not py27]` evaluates to `True` now so that the build is skipped => `Nothing to be done for recipe recipes/sentieon`.
The gist is that the `[py*]` selectors can be mostly replaced by appropriate version constraints with `conda-build 3`.

chapmanb:
Marcel;
Awesome, thanks so much for the tip. That worked perfectly. Len, the latest build is now available in bioconda and should hopefully be working for you with an update. Let us know if you run into any other problems.

Lenbok:
@chapmanb It's close now! When running sentieon-bwa, I get:

````
/opt/conda/bin/sentieon-bwa: 6: /opt/conda/bin/sentieon-bwa: Bad substitution
````
It looks like the bwa_symlinks.patch has introduced bashisms without also changing the shebang.

PR #9694, opened 2018-06-30
update ea-utils and wgsim

jvolkening:

**NOTE**: I found the actual ea-utils version number in the `ea-utils.spec` file within the tarball


jvolkening:
Isn't
```
[Jun 30 06:35:44] SERR fastq-stats: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory
```
a runtime error? Does adding `{{ compiler('c') }}` to `build` automatically pull the libraries into the runtime environment as well?

I must still be doing something wrong -- both tools passed building and testing using bioconda-utils and Docker on my local machine.

bgruening:
I do think libgcc would solve this for you. But it is not the new cb3 way to do it. Indeed to can annotate recipes/packages with an automatic runtime dependency. This means that the compiler pkgs, once specified will automatically include a runtime dependency - without you explicitly specifying it.

This will also happen for zlib at some point and then we do not need to specify zlib as runtime dependency as well.

* libgcc should never be build dependency
* the current error looks like we need a c++ compiler and not just c.

jvolkening:
Thanks for the explanation.

recipes/ea-utils/meta.yaml
```diff
@@ -1,33 +1,35 @@
 build:
-  number: 2
+  number: 0
   skip: True # [osx]
 
 package:
   name: ea-utils
-  version: 1.1.2.537
+  version: 1.1.2.779
```
epruesse: 1.1.2 here

recipes/ea-utils/meta.yaml
```diff
@@ -1,33 +1,35 @@
...
 
 source:
-  url: https://ea-utils.googlecode.com/files/ea-utils.1.1.2-537.tar.gz
-  sha1: 688bddb1891ed186be0070d0d581816a35f7eb4e
+  url: https://github.com/ExpressionAnalysis/ea-utils/archive/1.04.807.tar.gz
```
epruesse: but 1.04 here?

epruesse:
ping @druvus @jvolkening  - can you shed any light on the version number here?

jvolkening:
@epruesse There was confusion on the version numbering system here when the upstream package was apparently adopted by a different maintainer. If you look at the source code, in fact each individual tool has a different version. The `1.04.807` actually seemed to be the specific version of the `fastq-mcf` tool, and I speculated that the GitHub tag was incorrectly named when the package was migrated from Google Code. For the _package_ version I used the value from the file `ea-utils.spec`, which was the only one that made sense in the context of existing numbering schemes (e.g. previous recipes).

In short, at the time it seemed like kind of a mess and we tried to choose the most appropriate value.

PR #9697, opened 2018-06-30
Mauve

jamesabbott:

New recipe for the final component of Mauve


bgruening:
Oh it is green!!!

bgruening:
Ready to merge?

jamesabbott:
Not quite...The test is overly basic and I don't actually now it works. It's a java gui app which has some command-line only methods but even with -Djava.awt.headless=true functional tests fail since it looks for X11. A bit more playing to do yet....If I can't get a headless test to work I guess it will have to be merged and then tested manually when the build is available

jamesabbott:
My Attempte to improve the tests have not succeeded. Testing with openjdk fails with errors accessing fonts even when using conda-forge::openjdk and freetype as discussed in issue #8586. Installing openjdk from conda-forge and testing a build against this works ok, so hopefully it is working...

Think it will have to be merged as-is, unless anyone has some ideas as to how to fix the java font issue...

corburn:
If xvfb is available, `xvfb-run [command]` could be used to run the program in a virtual xserver.

recipes/mauve/meta.yaml
```diff
@@ -16,11 +16,11 @@ source:
 requirements:
   build:
     - ant
-    - conda-forge::openjdk
+    - "conda-forge::openjdk"
   host:
     - mauvealigner
   run:
-    - conda-forge::openjdk
+    - "conda-forge::openjdk"
```
bgruening: If its still installing from `defaults` which is already very wired, I think another dependency has a pinning to the default one or something like that.
By default openjdk should come from conda-forge. Or maybe the conda-forge one is broken and we should fix this one.

jamesabbott: I've tried quoting 'conda-forge::openjdk' as well but it still comes from defaults. Pinning the version number as fastqc does (issue #8586) solves the problem, but strikes me as being a little fragile.

From the previous issues, the conda-forge jdk is the one that works, and the one in defaults is broken but is tending to get pulled in.

Anyway...I _finally_ have a functioning test, so it's ready to merge if it looks ok. Try not to laugh at the number of commits it took to get to this point...

bgruening:
Thanks @jamesabbott for all your effort!

PR #9698, opened 2018-06-30
genomeinfodbdata: use https for download

feuerbach:

The specific reason I care about this is that my cluster environment seems to block the tcp port 80 (plain http) but allows 443/https.

While that reason may not be widely applicable, using https is a generally good idea and I see very few downsides to this change.

### DIFF
--- a/recipes/bioconductor-genomeinfodbdata/meta.yaml
+++ b/recipes/bioconductor-genomeinfodbdata/meta.yaml
@@ -7,11 +7,11 @@ package:
   version: '{{ version }}'
 source:
   url:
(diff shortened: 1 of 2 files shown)


feuerbach:
Thanks for the tip @bgruening, done now. I also took the chance to upgrade URLs in meta.yaml to https too.

Netzach:
Hi!

I have made a local mirror of bioconda for our offline research cluster and bioconductor-genomeinfodbdata does not work. It tries to download stuff from the Internet via wget instead of using conda. Could this be changed? It is a prereq for some bioconductor packages and the only one that fails due to having hardcoded URLs.

PR #9720, opened 2018-07-02
Pathwaymatcher: updated to v1.8.1

CarlosHorro:

### DIFF
--- a/recipes/pathwaymatcher/meta.yaml
+++ b/recipes/pathwaymatcher/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "PathwayMatcher" %}
-{% set version = "1.8" %}
+{% set version = "1.8.1" %}
 # Do not forget to update the version string in the pathwaymatcher.py file
 
 about:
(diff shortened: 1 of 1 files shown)


CarlosHorro:
I was testing v1.8 package (with noarch:generic) and I realized that it is giving an error when installing on Windows:

CondaVerificationError: The package for pathwaymatcher located at C:\Users\Standard\Anaconda3\pkgs\pathwaymatcher-1.8-1
appears to be corrupted. The path 'share/pathwaymatcher'
specified in the package manifest cannot be found.

Actually, I realized that my package is installed here:
C:\Users\Standard\Anaconda3\pkgs\pathwaymatcher-1.8-1\share\pathwaymatcher-1.8-1
And there should be a link there from 
C:\Users\Standard\Anaconda3\pkgs\pathwaymatcher-1.8-1\share\pathwaymatcher
but it does not exist on Windows.

Do you know how it can be created only when installing on Windows, please? (on Linux "ln -s $PACKAGE_HOME $PREFIX/share/$PKG_NAME" must still be executed).

PR #9744, opened 2018-07-04
Fixing scrappie blas dependency

wdecoster:

The previous recipe had an issue. It listed 'blas' as a dependency which should be 'openblas'. Apparently, that passed the testing the previous time? I reran the test today and now it failed. I replaced 'blas' with 'openblas', which seems to solve the issue.

See also https://github.com/nanoporetech/scrappie/issues/19


bgruening:
Thanks!

PR #9757, opened 2018-07-04
abyss 2.1.0

sjackman:

- Remove the patch for macOS. It's no longer needed.
- Remove the sqlite and Pandoc requirements. They're not required.
- Remove the BWA requirement. It's optional.
- Add the DOI of the 2017 ABySS 2.0 paper.


sjackman:
Is there an appropriate place in the `meta.yml` file to list the GitHub page? https://github.com/bcgsc/abyss A secondary `about.home` perhaps? Is it okay to include two DOI? Are the requirements specified under `host` required to be installed on the user's system, or only on the system used to build the binary package? I read https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#host but wasn't entirely clear on the answer afterward.

sjackman:
@bgruening wrote…

> No, only runs are required to be installed on the users-computer. The name host is just a very bad name imho. See here for more details: https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#host

https://github.com/bioconda/bioconda-recipes/pull/9550#issuecomment-402546722

bgruening:
@sjackman does this help? https://github.com/conda-forge/diskcache-feedstock/blob/8c10281e8196500d10f0e171b70d2aa000c4b824/recipe/meta.yaml#L39

multiple DOI should be ok, I'm not sure we will use them appropriate in all places, but we are just in the beginning of interlinking the resources. So it does not hurt to add them.

sjackman:
I've added `dev_url` and `doc_url`. Thanks!

bgruening:
Awesome thanks!

npavlovikj:
@bgruening , ABySS 2.1.0 is still not available in the bioconda channel. Do you know why so?

druvus:
@npavlovikj @sjackman abyss is listed in `build-fail-blacklist` so it was never tested or built. You need to remove it from the list to get it built.

sjackman:
How unfortunate. Was this PR tested by CI, and was it successful?

npavlovikj:
Thanks for the figuring this out @druvus ! @sjackman, the recipe wasn't tested at all - since it was blacklisted, CircleCI ignored it and if you check the build message it says "BIOCONDA INFO Nothing to be done."...

sjackman:
I've opened PR https://github.com/bioconda/bioconda-recipes/pull/9856 to remove `abyss` from `build-fail-blacklist`.

PR #9768, opened 2018-07-05
Various fixes (clustalo and clustalw)

dpryan79:

### DIFF
--- a/recipes/clustalo/build.sh
+++ b/recipes/clustalo/build.sh
@@ -3,7 +3,12 @@
 export CPPFLAGS="-I$PREFIX/include"
 export LDFLAGS="-L$PREFIX/lib"
 
-./configure --prefix=$SRC_DIR OPENMP_CFLAGS='-fopenmp' CFLAGS='-DHAVE_OPENMP'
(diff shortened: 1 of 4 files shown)


recipes/clustalo/build.sh
```diff
@@ -3,7 +3,12 @@
 export CPPFLAGS="-I$PREFIX/include"
 export LDFLAGS="-L$PREFIX/lib"
 
-./configure --prefix=$SRC_DIR OPENMP_CFLAGS='-fopenmp' CFLAGS='-DHAVE_OPENMP'
+if [ "$(uname)" == "Darwin" ]; then
+    # clang doesn't accept -fopenmp and there's no clear way around that
+    ./configure --prefix=$SRC_DIR
```
corburn: Shouldn't prefix be the install path?
```bash
./configure --prefix=$PREFIX
```

dpryan79: One would presume, but there's no `make install` so it doesn't really matter what you put there. Truth be told I was just lazy and never changed that from the original recipe, but you're correct that that should either be removed or changed to `$PREFIX` for the sake of clarity/consistency :)

Issue #9783, opened 2018-07-06
Trying to use a .jar in windows (PathwayMatcher)


CarlosHorro:
Hi,

Reading the guidelines for bioconda recipes related to java programs:
https://bioconda.github.io/guidelines.html#java
I read:
"JAR files should go in $PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM. A wrapper script should be placed here as well, and symlinked to $PREFIX/bin."

I see also an example of the wrapper script of PeptideShaker: 
https://github.com/bioconda/bioconda-recipes/tree/master/recipes/peptide-shaker

which is using build.sh file. Trying to do the same with PathwayMatcher (https://github.com/bioconda/bioconda-recipes/tree/master/recipes/pathwaymatcher)

after v1.8 package (with noarch:generic) I realized that it is giving an error when installing on Windows (the application itself works perfectly on Windows):

CondaVerificationError: The package for pathwaymatcher located at C:\Users\Standard\Anaconda3\pkgs\pathwaymatcher-1.8-1
appears to be corrupted. The path 'share/pathwaymatcher'
specified in the package manifest cannot be found.

Actually, I realized that my package is installed here:
C:\Users\Standard\Anaconda3\pkgs\pathwaymatcher-1.8-1\share\pathwaymatcher-1.8-1
And there should be a link there from
C:\Users\Standard\Anaconda3\pkgs\pathwaymatcher-1.8-1\share\pathwaymatcher
but it does not exist on Windows as I think a normal user doesn't usually have permissions for creating links on windows (and, however, the syntax on Windows is different:
 https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/mklink)

Do you know how the topic related to symlinks should be managed when installing packages on Windows, please? I didn't find any example of it...

Greetings,
Carlos

PR #9784, opened 2018-07-06
Fix wdltool, womtool and prinseq

dpryan79:

### DIFF
--- a/recipes/prinseq/build.sh
+++ b/recipes/prinseq/build.sh
@@ -1,10 +1,5 @@
 #!/bin/bash
 
-binaries="\
-prinseq-lite.pl \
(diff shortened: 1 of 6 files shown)


dpryan79:
@bioconda/core The method I used to fix wdltool and womtool would work on other packages and fix #9783 . Since mulitple versions of the same package aren't going to get installed at the same time in the same environment anyway, I think this is a little clearer. I would suggest that the java recommendations be changed to:

> JAR files should go in $PREFIX/share/$PKG_NAME/. A wrapper script should be placed in $PREFIX/bin and the PKG_NAME and JAR_NAME fields edited appropriately.

And that the script referenced there be changed.

PR #9788, opened 2018-07-06
G ri d

asafpr:
This recipe installs a tool for a paper coming out, using DNA sequencing for measuring growth rate in a metagenomics sample


asafpr:
The test fails on lofreq recipe which I didn't touch, and compiler 'c' which I added to meta.yaml file. I can't make the test pass

asafpr:
@bioconda/core could you please review? My first recipe. Thanks

recipes/grid/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+set -eu -o pipefail
+
+outdir=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+mkdir -p $outdir
+mkdir -p $PREFIX/bin
+
+cd $SRC_DIR
+
+cp -R * $outdir
```
bgruening: Do we need the test data? The package is big and I wonder if everything is need what you are copying here.

asafpr: It's not necessary but it's nice to have in the installation in order to test it. I can remove it.

PR #9820, opened 2018-07-09
PathwayMatcher: java8 required, package not available in windows

CarlosHorro:


CarlosHorro:
Could this PR be merged, please? :-)

recipes/pathwaymatcher/meta.yaml
```diff
@@ -11,9 +10,10 @@ package:
     name: pathwaymatcher
     version: {{ version }}
 
+
 build:
-  noarch: generic
-  number: 1
+  skip: True  # [ win ]
```
bgruening: We do not build for win so this has no effect.

bgruening:
I'm hesitant and have hoped for others to chime in. I think the root problem should be solved and the correct way would be to use noarch here. But to solve the root problem someone needs to be familiar with Windows, but we do not support Windows ... its a trap :(

ping @bioconda/core

rvalieris:
its not clear to me what the issue is here, was the package working before on windows without noarch ?
if thats the case I guess we could revert the change as long as it doesn't affect linux and osx.

anyhow this is tricky since we don't build for windows there no guarantee it will keep working on future builds even if its fixed now.

CarlosHorro:
Hi @rvalieris , when I specified "noarch", the package is generated for all platforms, but when people tried to install it into Windows there is an error related to one missing symlink.
Without noarch, I tried to generate the package only for Mac and Linux because of the windows problem.
I think "noarch" is currently misleading as it generates generic packages supposedly working on all platforms including Windows, but it is not the actually case when you try to install them on windows...

rvalieris:
I see, from what I searched symlinks on windows are tricky, this should be solved on the conda level   (there is a [issue here](https://github.com/conda/conda/issues/7384)), I don't think there's anything bioconda can do short of not using symlinks at all.

CarlosHorro:
Um... what do I do now, then? Is it ok to use the noarch?

rvalieris:
for now yes, I will merge this as it is.
if you can think of a way to avoid using symlinks on this recipe (without affecting usage on linux/osx) you can propose the change on another PR.

but its worth bearing in mind what @bgruening said, we don't support windows so even if the package is installable on windows (because of noarch) it doesn't mean its gonna work because we have no tests in place to guarantee that.

I understand this can be confusing for windows users but the alternative would be dropping noarch for all packages until we start having windows builds.

PR #9828, opened 2018-07-09
ngseqbasic VS20.0 to pair with bioRxiv release

telenius:


recipes/ngseqbasic/meta.yaml
```diff
@@ -21,8 +21,10 @@ requirements:
     - ucsc-bedclip ==332
     - ucsc-bedgraphpack ==332
     - ucsc-bedtobigbed ==332
+    - ucsc-bigbedtobed ==332
```
bgruening: I would remove those strict pinnings in general. To guarantee a certain environment you should use environment.yaml files instead.

PR #9839, opened 2018-07-09
Add MentaLiST-0.2.3

dfornika:


dfornika:
@bgruening could I get some advice on why this is failing? It looks to me like `conda-build` is having trouble with dependency conflicts. I tried to fix it with #9842 but I'm not sure if that's the correct solution.

This is the same way that I've updated MentaLiST versions in the past so I'm a bit surprised that it's failing now.

dfornika:
@bgruening Thanks very much for your help, I'm sure there are other recipes that are asking for your attention. Please suggest or tag other maintainers who could help with this.

I don't understand why the CI system is complaining about failed linting on several other recipes for this pull request. I'm seeing:

```
	recipe	check	info	depends_on_java-jdk	fix	missing_hash	missing_home	no_tests	should_not_use_fn	should_use_compilers	uses_git_url
0	recipes/discovar-denovo	missing_tests	{'no_tests': True, 'fix': 'add basic tests'}		add basic tests			True			
1	recipes/discovar-denovo	missing_hash	"{'missing_hash': True, 'fix': 'add md5, sha1, or sha256 hash to ""source"" section'}"		"add md5, sha1, or sha256 hash to ""source"" section"	True					
2	recipes/fiji/20151222	uses_javajdk	"{'depends_on_java-jdk': True, 'fix': 'use ""openjdk"" instead of ""java-jdk""'}"	True	"use ""openjdk"" instead of ""java-jdk"""						
3	recipes/flexbar/3.3.0	should_not_use_fn	{'should_not_use_fn': True, 'fix': 'URL should specify path to file, which will be used as the filename'}		URL should specify path to file, which will be used as the filename				True		
4	recipes/flexbar/3.3.0	should_use_compilers	"{'should_use_compilers': True, 'fix': 'use {{ compiler(""c"") }} or other new-style compilers'}"		"use {{ compiler(""c"") }} or other new-style compilers"					True	
5	recipes/gfinisher	uses_javajdk	"{'depends_on_java-jdk': True, 'fix': 'use ""openjdk"" instead of ""java-jdk""'}"	True	"use ""openjdk"" instead of ""java-jdk"""						
6	recipes/hurry.filesize	missing_home	{'missing_home': True, 'fix': 'add about:home'}		add about:home		True				
7	recipes/icqsol	missing_hash	"{'missing_hash': True, 'fix': 'add md5, sha1, or sha256 hash to ""source"" section'}"		"add md5, sha1, or sha256 hash to ""source"" section"	True					
8	recipes/icqsol	uses_git_url	{'uses_git_url': True, 'fix': 'use tarballs whenever possible'}		use tarballs whenever possible						True
...
```

bgruening:
You need to merge master into your branch.

recipes/mentalist/0.1.4/meta.yaml
```diff
@@ -10,20 +10,20 @@ source:
...
 
 requirements:
-  build:
-    - julia 0.5.2
+  host:
+    - julia
     - hdf5
-    - mpfr
-    - libxml2 2.9.*
+    - {{ pin_compatible('mpfr') }}
```
bgruening: I don't think this will work. `mpfr` should be fine in both host/run.

dfornika: Could you help me interpret this error message?

```
conda.exceptions.UnsatisfiableError: The following specifications were found to be in conflict:
  - julia -> mpfr=3.1
  - mpfr=4
```

bgruening: it means the Julia package that was installed was linked against a older version than what we want to have today. This is most likely because you have used the bioconda Julia version, but you want to have the most recent conda-forge vesion. I have fixed this for you hopefully.

bgruening: And this error now means we need to rebuild julia in the conda-forge feedstock :(
You are maintainer, so let do this together :)

dfornika: `mentalist-0.1.x` is designed to run on `julia-0.5.x` and `mentalist-0.2.x` is designed to run on `julia-0.6.x`.

dfornika: What do you mean by 're-build'? What (if anything) needs to change in the `julia` recipe?

bgruening:
This was needed: https://github.com/conda-forge/julia-feedstock/pull/50
I wonder if you really need to keep the older versions of recipes?

dfornika:
I'm not sure. I suppose I've just been leaving them in there in case I need to issue a new build of an older version. Should I remove them?

bgruening:
This is really up to you. If you want to maintain older version keep them, if not remove them. Fixing old recipes can also work by just extract the old recipes from the git history.

PR #9844, opened 2018-07-10
Add recipe for SSAKE 4.0

npavlovikj:

### DIFF
--- a/recipes/ssake/build.sh
+++ b/recipes/ssake/build.sh
@@ -2,11 +2,6 @@
 
 mkdir -p ${PREFIX}/bin
 
-sed -i.bak "s:/usr/bin/perl:/usr/bin/env perl:" ssake/SSAKE
-sed -i.bak "s:/usr/bin/perl:/usr/bin/env perl:" ssake/tools/*.pl
-sed -i.bak "s:/home/martink/bin/perl:/usr/bin/env perl:" ssake/tools/getStats.pl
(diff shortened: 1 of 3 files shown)


recipes/ssake/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  sha256: {{ sha256 }}
+  url: https://github.com/warrenlr/{{ name }}/archive/v{{ version }}.tar.gz
+
+requirements:
+  run:
+    - perl-statistics-descriptive
```
bgruening: Do you also need perl here?

recipes/ssake/build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+mkdir -p ${PREFIX}/bin
+
+sed -i.bak "s:/usr/bin/perl:/usr/bin/env perl:" ssake/SSAKE
+sed -i.bak "s:/usr/bin/perl:/usr/bin/env perl:" ssake/tools/*.pl
```
bgruening: Can you try to patch this upstream as well?

npavlovikj: Done :)

recipes/ssake/meta.yaml
npavlovikj: "perl" is installed as part of "perl-statistics-descriptive". Isn't this enough? If I specify "perl" is an independent requirement, there is a version incompatibility with the other packages, and the build fails.

PR #9853, opened 2018-07-10
Update upstream tarball url and remove Augustus from blacklist

kastman:

The author doesn't seem to have a stable url to the most recent release; moving the tarballs to a directory called "old" after an updated version is released, so this was causing a break during build. Fixed the path here, and will then update to the newest version once the blacklist is cleared.


kastman:
Augustus was taken off the blacklist by #10364 and presumably the perl issues had resolved themselves by then (e.g. conflicting packages depending on threaded were removed; see the PR comments); closing this PR.

PR #9856, opened 2018-07-10
build-fail-blacklist: Remove abyss

sjackman:

Remove `abyss` from `build-fail-blacklist`.

### DIFF
--- a/recipes/abyss/meta.yaml
+++ b/recipes/abyss/meta.yaml
@@ -9,7 +9,7 @@ build:
   number: 0
 
 source:
-  url: http://www.bcgsc.ca/platform/bioinfo/software/{{ name|lower }}/releases/{{ version }}/{{ name|lower }}-{{ version }}.tar.gz
+  url: https://github.com/bcgsc/abyss/releases/download/{{ version }}/abyss-{{ version }}.tar.gz
   sha256: fe28aee5e2ee24ea2c550518bfeb65577f8f9900831bf4fb0858ac1829fd86fa
 
 requirements:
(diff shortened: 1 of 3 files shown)


sjackman:
Is this PR sufficient to trigger a CI build of `abyss`?

npavlovikj:
That is enough @sjackman  - however, the download url is not existing, so the build fails with "Error: HTTP 404 NOT FOUND for url <http://www.bcgsc.ca/platform/bioinfo/software/abyss/releases/2.1.0/abyss-2.1.0.tar.gz>". @druvus did some fixes in #9855, but the recipe still needs some more tweaking to build properly.

sjackman:
```
06:08:14 BIOCONDA INFO BUILD SUCCESS abyss-2.1.0-h09b0a5c_0.tar.bz2
…
[Jul 11 06:09:20] SERR ABYSS-P: error while loading shared libraries: libsqlite3.so.0: cannot open shared object file: No such file or directory
```
https://circleci.com/gh/bioconda/bioconda-recipes/17500

`sqlite` is not listed a dependency of the recipe, yet `configure` appears to have picked up sqlite opportunistically. I'll add an explicit `./configure --without-sqlite`.

sjackman:
💚 and life is good.

recipes/abyss/meta.yaml
```diff
@@ -9,7 +9,7 @@ build:
   number: 0
 
 source:
-  url: http://www.bcgsc.ca/platform/bioinfo/software/{{ name|lower }}/releases/{{ version }}/{{ name|lower }}-{{ version }}.tar.gz
```
bgruening: @sjackman if you do not increase the build number (or the version number and reset the build number to 0) the pkg will not be build.

druvus: Abyss version 2.1 were never previously built due to blacklisted so I think this is good to go.

sjackman:
Thanks for merging, Andreas!

PR #9885, opened 2018-07-12
update recipe for libseq 1.9.3

molpopgen:

### DIFF
--- a/recipes/libsequence/meta.yaml
+++ b/recipes/libsequence/meta.yaml
@@ -12,7 +12,7 @@ build:
 
 requirements:
   build:
-    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
   host:
     - zlib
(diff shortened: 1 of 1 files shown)


molpopgen:
@bgruening -- just an FYI.  It appears that a recent set of bulk recipe added - {{ compiler('c') }} to many recipes that are C++.  It should be - {{ compiler('cxx') }}.  This PR failed at first because libstdc++ couldn't be loaded.

I'll update recipes for some of my other projects over the next several days.

molpopgen:
> I've also seen people add both compilers in a few cases, but I have no idea
> if that's necessary or desirable...

For a C++ project, it is probably totally fine.  For a C-only project, it is harmless overkill?  But missing the cxx for a C++-only project is going to lead to build failures when projects get bumped.

PR #9887, opened 2018-07-13
Added prototype for ribohmm recipe

eggzilla:


recipes/ribohmm/build.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+cp -r data/kozak_model.npz $PREFIX/share/kozak_model.npz
+mkdir -p $PREFIX/bin/data/
+cp -r data/kozak_model.npz $PREFIX/bin/data/kozak_model.npz
```
bgruening: This line can be removed?

recipes/ribohmm/build.sh
```diff
@@ -0,0 +1,20 @@
...
+sed -i 's@data/kozak_model.npz@'"$PREFIX"'/share/kozak_model.npz@' seq.pyx
+$PYTHON setup.py build_ext --inplace --force
```
bgruening: Is this really needed?

recipes/ribohmm/meta.yaml
```diff
@@ -0,0 +1,41 @@
+package:
+  name: ribohmm
+  version: "1.0.0"
+
+source:
+  url: https://github.com/rajanil/riboHMM/archive/57a34d5afbd329d9641c895d12d1b0cefea0735b.zip
+  sha256: b0e182e0e07ede199aafef1936e2eab21fa21c6d15ffb1e00386bc0db167b436
+
+build:
+  number: 1
```
bgruening: 0

recipes/ribohmm/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  skip: True [osx or py3k]
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('fortran') }}
+  host:
+    - python
+    - numpy 1.11
```
bgruening: I don't think you should pin numpy here.
Is this tool compiling against the numpy ABI?

eggzilla: Problem is that it failed with newer numpy versions, so i downgraded

bgruening:
The issue is probably upstream. We are trying to fix this upstream.

eggzilla:
Thanks for the help! Could you ping me as soon as it is fixed?

PR #9893, opened 2018-07-13
Adds recipes for barriers 1.6. and 1.7.

j-4:
Adds recipes for barriers 1.6. and 1.7. as well as for the dependency gengetopt


bgruening:
@j-4 the gengetopt should be in conda-forge imho.

j-4:
@bgruening I searched for it under various names, but could not find it... Maybe they use a uncommon name? The only thing I could find was `Yaggo`, but this seems to be a fork/reimplementation of gengetopt.

nsoranzo:
I think what @bgruening means is that the `gengetopt` recipe should be added to conda-forge instead of bioconda.

nsoranzo:
@j-4 Can you remove the version from the recipe paths? E.g. `git mv recipes/barriers/1.6.0/build.sh recipes/barriers/build.sh`

j-4:
@nsoranzo I included the versions, as I need both 1.6 and 1.7 and they come from different sources (github and tbi homepage). Do you think, the versions are messing up the recipes?

nsoranzo:
@j-4 Yeah, I noticed after sending the comment, sorry. I don't think the versions are messing up the recipes, but not using them (when possible) makes it easier to maintain the repository in the long term.
So, I'd suggest to remove the version for `barriers` 1.7 and `gengetopt`.

recipes/barriers/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+./autogen.sh && \
```
nsoranzo: You don't need the `&& \`

recipes/barriers/build.sh
```diff
@@ -0,0 +1,6 @@
...
+./configure --prefix=$PREFIX --with-hash-bits=27 CFLAGS='-mcmodel=large' && \
```
nsoranzo: To help debugging, you can replace the `&& \` with `|| (cat config.log ; exit 1)`

recipes/barriers/meta.yaml
```diff
@@ -0,0 +1,36 @@
+{% set name = "barriers" %}
+{% set version = "1.7.0" %}
+{% set sha256 = "7061db4974b91d8dbace21256a7f3a9776f45358cc473e197d2f2dc7f236013a" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+about:
+  home: https://www.tbi.univie.ac.at/RNA/Barriers/
```
nsoranzo: Maybe replace with https://github.com/ViennaRNA/Barriers ? Up to you.

recipes/barriers/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+  license: GPL
+  license_file: COPYING
+  summary: "Compute local minima and energy barriers of a landscape."
+
+build:
+  number: 0
+
+source:
+  url: https://github.com/ViennaRNA/Barriers/archive/v1.7.0.tar.gz
```
nsoranzo: Use `{{ version }}` instead of `1.7.0` ?

j-4:
@nsoranzo Thanks for reviewing, I included all your suggestions!

nsoranzo:
@j-4 Thank you, now we need to wait for the conda-forge PR.

nsoranzo:
@j-4 Can you rebase the PR (or merge master)?

PR #9899, opened 2018-07-13
Bump build number b/c we are rebuilding against libseq 1.9.3.  Also

molpopgen:
change to use cxx as the compiler.


molpopgen:
@johanneskoester -- Requesting help/review here.  Linting tests are failing for most/all open PRs.  Unable to parse out the issue from the Circle CI output.

kyleabeauchamp:
@bgruening  did someone un-blacklist a bunch of recipes that fail linting?  

Do we just need to re-add those recipes to https://github.com/bioconda/bioconda-recipes/blob/master/build-fail-blacklist ?

molpopgen:
> Do we just need to re-add those recipes to https://github.com/bioconda/bioconda-recipes/blob/master/build-fail-blacklist ?

That may be the case.  However, I don't think this package was blacklisted.  If it was, I want to fix it!

My other open PR did end up on the blacklist, but I was never contacted about that, and the issue is almost certainly that the compiler was mis-specified as 'c' (when it should be 'cxx') by whomever made those updates.

bgruening:
You just need to merge master into your branch this will fix the linting problem.

bgruening:
> My other open PR did end up on the blacklist, but I was never contacted about that, and the issue is almost certainly that the compiler was mis-specified as 'c' (when it should be 'cxx') by whomever made those updates.

That was most likely me - as I try to update all recipes to new cb3 syntax. Bare with me that I'm not trying to read all 4000 recipes and do this semi-automatic and fix packages on bulk when I see errors.

PR #9919, opened 2018-07-16
Salmon 0.11.0 buildbump

rob-p:


bgruening:
@rob-p what was the Docker build issue?
I think the correct way is to reset the build number to 0 if you release a complete new version. Sorry, have not spotted this in your last PR.

rob-p:
@bgruening : The issue actually did't show up at all in bioconda's build (or on my local CI).  It showed up only when building the Docker image for salmon (on Ubuntu 18.04 stock).  It's complicated, but basically, on that system, the inclusion of boost_thread forced `-lpthread` into the linker flags, which breaks the build (because pthread should be linked statically by passing `-pthread` to the compiler and linker).  The solution was to simply stop linking against `boost_pthread`, since I wasn't using it anyway.

rob-p:
The only reason I re-submitted the bioconda build is because I had to update the CMakeLists.txt file, which changed the md5sum.  Since I didn't tag the release on Docker or on our github releases page yet, I simply moved the 0.11.0 tag to this new version.  I wanted to make sure the md5sum for the bioconda build matched what we have tagged as 0.11.0, so I updated the build.

bgruening:
Ok! Thanks for the explanation. Lets reset the build number to 0 for this new version. Feel free to merge, bedtime for me.

PR #9937, opened 2018-07-17
Gem3 mapper

karl616:


karl616:
This is gem3-mapper: https://github.com/smarco/gem3-mapper
I think it would be a worthy addition to bioconda

recipes/gem3-mapper/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+#adding include and library paths in order to find bzip2
+sed -i.bak '
+    /^PATH_INCLUDE=/ s@$@ -I'$PREFIX'/include@
+    /^PATH_LIB=/ s@$@ -L'$PREFIX'/lib@
+  ' Makefile.mk.in
+
+
+./configure
```
bgruening: ./configure --prefix=$PREFIX?

recipes/gem3-mapper/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+  host:
+    - bzip2
+    - zlib
+  run:
+    - bzip2
+    - zlib
+
+test:
+  commands:
+          #    - gem-mappability -h 2>&1 |grep -e "-I <index_prefix>.*(mandatory)"
```
bgruening: anything wrong with this test?

recipes/gem3-mapper/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - gem-mapper 2>&1 | grep 'index|-I <index_file.gem>'
+    - gem-indexer -h 2>&1 | grep 'input|-i <input_file> (Multi-FASTA)'
+    - gem-retriever -h 2>&1 | grep 'index|-I <index_file.gem> (GEM archive)'
+
+about:
+  home: https://github.com/smarco/gem3-mapper
+  license: GPL3
```
bgruening: license_file: LICENSE could be added.
Can you maybe also include a DOI or bio.tools ID?

recipes/gem3-mapper/build.sh
karl616: Nope, it is not enough. I also tried with CFLAGS and LDFLAGS, but they are not respected. Thanks for the suggestion though, I didn't think of it earlier.

recipes/gem3-mapper/meta.yaml
karl616: Will do.
I'm adding the original paper. There is an entry in bio.tools/gemmapper, but it is currently pointing to an older version, not updated, version of the software. I will add it for now and ask the developer to try and update it.

recipes/gem3-mapper/meta.yaml
karl616: the particular executable is not included in this version. I started with an older version, but changed along the way. I will remove it.

karl616:
@bgruening Thanks for the feedback, always appreciated

PR #9943, opened 2018-07-17
Salmon update

bgruening:

Let's see if this works. Sorry @rob-p, the last month exhausting, we rebuild most of Bioconda and we had there and there some issues. Hope things got better. I got the bulk branch to build successfully on Linux at least.


rob-p:
Thanks @bgruening --- I loosely followed the Gitter channel last month, so I observed a bit of the Herculean effort.  Great job!  Thanks for looking into this 👍.

rob-p:
@bgruening --- looks like it's not building again.  However, it also looks like `salmon` appears in the package blacklist (which is, I presume, the reason why).  Shall we try and remove it (I'm not sure of the procedure for this)?

bgruening:
Sorry, forgot to add one commit.

bgruening:
@rob-p this is the error I got:

```
/opt/conda/conda-bld/salmon_1531831047676/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/lib/libboost_regex.a(icu.o):icu.cpp:(.text._ZN5boost16re_detail_10660019basic_regex_creatorIiNS_16icu_regex_traitsEE16create_startmapsEPNS0_14re_syntax_baseE[_ZN5boost16re_detail_10660019basic_regex_creatorIiNS_16icu_regex_traitsEE16create_startmapsEPNS0_14re_syntax_baseE]+0xe55): more undefined references to `u_tolower_58' follow
```

rob-p:
Looks boost/icu related?

rob-p:
@bgruening : how can I request the linux test to run again?

rob-p:
Trying to pin ICU didn't work, but the issue is definitely related to failure to find icu-related symbols (perhaps as a result of how boost was compiled?).  Was boost built with the `--std=c++11` flags? I'm not sure exactly what else the issue may be.

bgruening:
Boost was most likely not compiled with c++11, this will still take a few weeks :(

rob-p:
Im not even sure that is what is wrong, just trying to think what we did to fix this last time this error popped up...

recipes/salmon/salmon_cmake.patch
```diff
@@ -0,0 +1,20 @@
+--- CMakeLists.txt
++++ CMakeLists.txt
+@@ -339,7 +339,7 @@ if (BOOST_RECONFIGURE)
+     set(CMAKE_PREFIX_PATH ${CMAKE_CURRENT_SOURCE_DIR}/external/install)
+     set(Boost_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/external/install/include)
+     set(Boost_LIBRARY_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/external/install/lib)
+-    find_package(Boost 1.59.0 COMPONENTS iostreams filesystem system timer chrono program_options REQUIRED)
```
bgruening: We are also using boost 1.67, could this be a problem as well? https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml#L243

rob-p: Hi @bgruening,

  Newer should not be a problem, 1.59 is a lower bound.  If the user doesn't have boost available, we fetch 1.66.  I *believe* that CMake implements a >= policy for these versions, so the user must have version >= 1.59.0.  The issue is clearly related to finding these icu58-related symbols during linking.  I recall having an issue like this before, where the solution was to pin a specific version of icu & boost --- but now conda is using the newer version of icu (58) that we'd pinned by default, so I think the issue must be different.

bgruening: No current boost is not compiled with a recent GCC. conda-forge is still rebuilding :(
Sorry.

rob-p: ahhhh ok --- so I think this might make more sense now.  I wonder if this is the cause of a number of other failures, as any package that is building on cb3 (and hence, the newer gcc), but not _explicitly_ turning on the old ABI, should have issues.

bgruening: The newer GCC just comes in for salmon by adding this file to the salmon dir: https://github.com/bioconda/bioconda-recipes/blob/salmon/recipes/salmon/conda_build_config.yaml
So all (there are a handful other exceptions) other recipes do still build against old GCC until conda-forge has rebuild first.

rob-p: Ahh, so, salmon only need c++11 right now.  Could we revert to gcc 4.8.5 or whatever for this release? Would that fix things?

rob-p: Id be happy to give it a try but removing the file seems to anger the linter.

rob-p: @bgruening , 

  Any idea why I might be seeing this from the linter?

```
0	recipes/salmon	should_use_compilers	"{'should_use_compilers': True, 'fix': 'use {{ compiler(""c"") }} or other new-style compilers'}"	"use {{ compiler(""c"") }} or other new-style compilers"	True
```

bgruening:
Now we are back to the gcc error :(

recipes/salmon/salmon_cmake.patch
```diff
@@ -0,0 +1,11 @@
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -45,7 +45,7 @@ if(CONDA_BUILD)
+   set(CMAKE_CXX_STANDARD_REQUIRED ON)
+   set(CMAKE_CXX_EXTENSIONS OFF)
+   set(CXXSTDFLAG "-std=c++11")
+-  set(GCCVERSION "4.8.5")
++  set(GCCVERSION "4.8.2")
```
bgruening: Good to have upstream developer support :)
Thanks @rob-p! Seems to work for Linux now. If OSX passes I will clean up the recipe a little bit and we can merge it.

rob-p: Great!  I'm not sure if OSX will work or not yet though --- there was a reason we had all of that `llvm_dev` jazz there before ;P.

rob-p: Ok, so i think the osx error is due to the compiler version.  Is there anything we can do at this point to get llvm_dev, or do we have to wait for something regarding cb3?

bgruening:
@bioconda/osx can you help us here?
One possibility is to overwrite the compiler for OSX using the conda_build_config.yaml file.

In the meantime we could skip OSX is this is desired?

lparsons:
Shall we go ahead and skip build on OSX for now?

epruesse:
I‘ll have a look at OSX later today. 

Elmar

--
message sent from mobile device

acaprez:
I have no objection to skipping OS X for now if needed.

rob-p:
Since we aren't making progress on OSX yet, im in favor of skipping it for now as well.

lparsons:
OK, updated the PR to skip building on OSX for now. @bioconda/osx If you could try to check/fix this recipe again once the new build system is fully in place, that would be much appreciated. I think we're ready to merge?

PR #9944, opened 2018-07-17
Remove salmon from blacklist

rob-p:

This PR removes salmon from the blacklist after the "bulk" update.


martin-mann:
seems this PR is obsolete, since salmon was removed from the blacklist somewhere else...

PR #9952, opened 2018-07-17
updating KMC to version 3.1.1rc1

notestaff:


recipes/kmc/3.1.1rc1/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  patches:
+    - 0001-fixes-for-conda.patch
+
+build:
+  number: 0
+  # this recipe builds kmc without the asmlib speedup
+
+requirements:
+  build:
+    # C++14 features are required
```
bgruening: Are you sure about this comment? You are not using the new compilers as it seems.

notestaff: @bgruening the makefile does specify C++14: https://github.com/bioconda/bioconda-recipes/pull/9952/files#diff-10ea6eb45d548abe3d44bba556156971R32

bgruening:
I see, it is blacklisted. Can you please remove kmc from the blacklist in this branch. Thanks.

notestaff:
@bgruening so, is it possible to build with C++14?  if so, how?

bgruening:
Yes, include this file https://github.com/bioconda/bioconda-recipes/blob/master/recipes/stacks/2.1/conda_build_config.yaml into your folder. It will overwrite the default settings.

bgruening:
Both version will be build now. The older one fails because of the missing new compiler, the newer one fails because of a missing patch as it seems.
If you like you can remove the older recipe.

bgruening:
Thanks!!!

PR #9956, opened 2018-07-17
Unblacklist perl packages

druvus:

### DIFF
--- a/recipes/perl-module-implementation/build.sh
+++ b/recipes/perl-module-implementation/build.sh
@@ -3,10 +3,10 @@
 # If it has Build.PL use that, otherwise use Makefile.PL
 if [ -f Build.PL ]; then
     perl Build.PL
-    ./Build
-    ./Build test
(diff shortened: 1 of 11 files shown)


recipes/perl-module-implementation/build.sh
```diff
@@ -3,10 +3,10 @@
 # If it has Build.PL use that, otherwise use Makefile.PL
 if [ -f Build.PL ]; then
     perl Build.PL
-    ./Build
-    ./Build test
+    perl ./Build
```
bgruening: I have done this now already multiple time, if you think we should do this for all packages I would be ok to merge a PR for `sed`ding all perl pkgs :)

druvus:
Seems that perl-module-implementation not picking up the newly built perl-module-runtime so I will split it in multiple PR.

bgruening:
If you think its easier for you please apply against the bulk branch and I will merge it into bulk.

PR #9976, opened 2018-07-18
adding vkmz recipe

eslerm:

Since conda creates an env/info directory I added the README and LICENSE there. Is this compliant with guidelines?


eslerm:
circleci lint failed with: 
```
recipes/vkmz	should_be_noarch	"{'should_be_noarch': True, 'fix': 'add ""build: noarch"" section'}"	"add ""build: noarch"" section"	True
```
But tool is python3 only and guidelines state: 
>> Update 7 Feb 2018: Previously we had recommended that if the recipe is a pure Python package, it should be marked as a “noarch” package (details). However due to technical incompatibilies we can’t do this – so please DO NOT use "noarch" for now.

eslerm:
A big thanks to @bgruening for helping me write the recipe!

bgruening:
Thanks!

marcelotrevisani:
@bgruening , is it better to use in build section
```
skip: true  # [not py27]
```
or just let
```
python >=3
```
in requirements?

bgruening:
I don't think `skip: true  # [not py27]` will work with noarch pkgs.

PR #9982, opened 2018-07-19
New recipe: ichorCNA for cfDNA tumor estimation

chapmanb:

### DIFF
--- a/recipes/r-ichorcna/meta.yaml
+++ b/recipes/r-ichorcna/meta.yaml
@@ -16,12 +16,13 @@ build:
     - lib/R/lib/
     - lib/
 requirements:
-  build:
-    - perl # for adding in correct path references to scripts
+  host:
     - 'bioconductor-genomeinfodb >=1.8.7'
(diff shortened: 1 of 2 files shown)


recipes/r-ichorcna/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+build:
+  number: 0
+  rpaths:
+    - lib/R/lib/
+    - lib/
+requirements:
+  host:
+    - r-base
+    - libgfortran-ng
```
bgruening: This does not belong into the host section - run should be enough.
Also please note that this is a bug, which we are working on in the r-base-feedstock.

chapmanb:
Björn -- thanks so much for working on the general fix to the gfortran issue. I ran into it yesterday during an install for the first time and thought it was something miswired with Rcpp. I ended up needing libgfortran in both the host and run, but can hopefully remove on the next iteration after it's fixed. Thanks again.

PR #9983, opened 2018-07-19
bs_call

karl616:


karl616:
I'm not sure my solution for fopenmp and macos is correct and I don't have the possibility to test it

recipes/bs_call/build.sh
```diff
@@ -0,0 +1,28 @@
...
+    /^GSL_INC = -I\/apps\/GSL\/2.4\/include\// s@= -I.*@= -I'$PREFIX'/include@
+  ' Gsl.mk
+
+sed -i.bak '
+    /^INCLUDE_FLAGS=/ s@$@ -I'$PREFIX'/include@
+    /^LIB_PATH_FLAGS=/ s@$@ -L'$PREFIX'/lib@
+  ' GEMTools/Makefile.mk
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  #remove openmp for macos
```
bgruening: This does not look like you are removing openmp support.

karl616:
Hi @bioconda/core,
this here recipe would need a review. It adds a new package, bs_call, which is a fast implementation for CpG methylation calling. I think the package as such is worthy of inclusion. It was used by the Blueprint consortia (http://www.blueprint-epigenome.eu/).

Where I'm a bit hesitant is in the way I handled the macosx build. The package makes use of openmp and I found a solution that passes the checks, but I'm not sure if it really works... if there is a better way to do it... I have no mac-access to test it...

recipes/bs_call/build.sh
karl616: right, that is a comment from earlier. I should change that. 

Removing openmp from the Makefile doesn't help as there are files requiring omp.h. As the code is written is written it is a forced inclusion. What I found instead is a workaround:

https://iscinumpy.gitlab.io/post/omp-on-high-sierra/

But I feel I'm in deep water here. Would it be better to disable the macos build?

PR #9984, opened 2018-07-19
Mothur1.40

apeltzer:

### DIFF
--- a/recipes/mothur/patch
+++ b/recipes/mothur/patch
@@ -1,0 +1,22 @@
+--- Makefile	2018-06-19 16:16:02.000000000 +0200
++++ Makefile_new	2018-07-19 12:40:55.822132011 +0200
+@@ -22,8 +22,8 @@
+ USEREADLINE ?= yes
+ USEBOOST ?= yes
+ LOGFILE_NAME ?= yes
+-BOOST_LIBRARY_DIR="\"Enter_your_boost_library_path_here\""
+-BOOST_INCLUDE_DIR="\"Enter_your_boost_include_path_here\""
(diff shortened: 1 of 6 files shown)


apeltzer:
Sorry, @shiltemann - maybe you have an idea why this fails? Just updated the patches and it still doesn't run through.

recipes/mothur/patch
```diff
@@ -0,0 +1,22 @@
+--- Makefile	2018-06-19 16:16:02.000000000 +0200
++++ Makefile	2018-07-19 12:40:55.822132011 +0200
+@@ -22,8 +22,8 @@
+ USEREADLINE ?= yes
+ USEBOOST ?= yes
+ LOGFILE_NAME ?= yes
+-BOOST_LIBRARY_DIR="\"Enter_your_boost_library_path_here\""
+-BOOST_INCLUDE_DIR="\"Enter_your_boost_include_path_here\""
++BOOST_LIBRARY_DIR=${PREFIX}/lib
```
bgruening: It seems that this does not work. PREFIX evaluates to "". Better to set this in the build.sh script.

acaprez:
It's the `PREFIX := ${CURDIR}` line in the `Makefile` file that's the issue.  `CURDIR` is a built-in variable for `make`, so that line overrides the `PREFIX` value from the environment.  I updated the `patch` file ([patch.txt](https://github.com/bioconda/bioconda-recipes/files/2211476/patch.txt))
to remove that line, and it built ok then.

apeltzer:
Looks good to me, too! Just added your patch and it builds locally.

PR #9991, opened 2018-07-19
Chado tools

puethe:

command line tool to access CHADO databases

### DIFF
--- a/recipes/chado-tools/meta.yaml
+++ b/recipes/chado-tools/meta.yaml
@@ -1,0 +1,36 @@
+{% set name = "chado-tools" %}
+{% set version = "0.0.3" %}
+{% set sha256 = "e8b7407013eee815288589f07314655d1e6ffc7c688f7daf58080fc4369a3668" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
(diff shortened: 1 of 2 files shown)


recipes/chado-tools/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/sanger-pathogens/{{ name }}/archive/v{{ version }}.{{ file_ext }}
+  {{ hash_type }}: {{ hash_value }}
+
+build:
+  number: 0
+  noarch: generic
```
bgruening: I guess this can be `noarch: python`?

puethe: I'm not sure. It's a command line tool, and according to [the documentation](https://bioconda.github.io/linting.html#should-be-noarch), it should then be "generic" - if I understood correctly.

puethe: Tests on my machine pass flawlessly with either setting. Let me know what you think is better.

bgruening: python is preferred I guess.

puethe: Ok, thanks. Changed it.

PR #10015, opened 2018-07-20
update to new version

macmanes:
the old version had a bug whereby some reads were dropped. See https://github.com/mourisl/Rcorrector/issues/12

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1895,8 +1895,6 @@ recipes/pyvcf
 # /usr/bin/env: python -Es: No such file or directory
 recipes/pizzly
 
-# zlib
-recipes/rcorrector
 # compiles its own zlib
 recipes/plink2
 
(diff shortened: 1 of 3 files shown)


jpuritz:
Might have to update `build.sh` with some of the flags noted [here](https://bioconda.github.io/troubleshooting.html#zlib-errors)

macmanes:
Hi @jpuritz I don't see errors related to `zlib`, so not sure if adding stuff in that link are going to help. In general, this is a very minor bump in version, but a critically important one. 1.0.2 has a bug which drops reads from R2. This bug is fixed in `1.0.3`. I'm not sure why `build.sh` should change, but then again, I know literally *nothing* about this.

macmanes:
OK, I'm stumped. Reading the CircleCI log does not help me understand what is wrong here. Again, a minor fix, should not require substantial changes to the build process.

jpuritz:
I’ll take a look again today. Even though this is a minor code update,
Bioconda has begun switching to the new Conda build system which requires
some setting of environmental variable for proper compilation.

-- 
Jon Puritz

jpuritz:
So, the makefile had some non-standard flags, so I had to modify it to add the `-L$PREFIX/lib` to the compiler.

PR #10019, opened 2018-07-20
Squid

Congm12:

Hi @bioconda/core ,

I am requesting a pull for a new recipe, SQUID, which is a command line tool for transcriptomic structural variation detection.

This is my first recipe to bioconda. When tested with bootstrap method, compilation is successful for linux-64 environment, but there seems to be something wrong with my osx-64 enviroment that even for existing recipe outside blacklist. The error message says:
```
BUILD FAILED: the built package /tmp/miniconda/miniconda/conda-bld/osx-64/*.tar.bz2 cannot be found
```

Since the existing successful recipe also has this error, I suspect there may be something wrong with my test system. Please review and let me know if there is error on your end.

Thanks.
Cong


recipes/squid/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  license: BSD 3
+  summary: Detector for fusion-gene and non-fusion-gene transcriptomic structural variations from RNA-seq data
+
+package:
+  name: squid
+  version: {{ version }}
+
+source:
+  url: https://github.com/Kingsford-Group/squid/archive/v{{ version }}.tar.gz
+  md5: 0786119db2ce764dc55fdb407e5ddd7b
```
bgruening: You could use a sha256 here.

recipes/squid/meta.yaml
```diff
@@ -0,0 +1,35 @@
+{% set version="1.4" %}
+
+build:
+  number: 0
+
+about:
+  home: https://github.com/Kingsford-Group/squid
+  license: BSD 3
```
bgruening: `license_file:` could be added as well

bgruening:
Thanks!

PR #10021, opened 2018-07-20
Update RGI recipe to fix dependency problems

LarsDu:

Though my own testing, I found that using a newer version of blast like 2.7.1, or a newer version of diamond like 9.2.2 will break RGI version 4.0.3 when running resistance gene detection using the `--alignment_mode DIAMOND` flag of `rgi main`

These exact matches should fix the aforementioned dependency issues.

### DIFF
--- a/recipes/rgi/meta.yaml
+++ b/recipes/rgi/meta.yaml
@@ -13,7 +13,7 @@ source:
 
 build:
   skip: True # [not py27]
-  number: 1
+  number: 2
 
 requirements:
   build:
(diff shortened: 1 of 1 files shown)


LarsDu:
I'm not changing the version of the package. I'm fixing the recipe. Does this still require a change in build number?

bgruening:
Yes, every change that you want to see on anaconda.org needs to be build and this is only triggered via a build-bump.

LarsDu:
Where is the build number located?

Issue #9794, opened 2018-07-06
problem with linking to bamtools


grenaud:
I am trying to build my little C++ package called libgab which contains some functions to rebuild the reference bases of a read from a BAM file. I use bamtools. I ran:

`conda config --add channels defaults`
`conda config --add channels conda-forge`
`conda config --add channels bioconda`
`conda install bamtools`
`git clone https://github.com/grenaud/libgab`
`cd libgab`
`make BAMTOOLSINC="/home/alex/.conda/envs/test_gabriel/include/bamtools"
BAMTOOLSLIB="/home/alex/.conda/envs/test_gabriel/lib/"`

this is all fine, I can compile but not link:
undefined reference to BamTools::BamAlignment::FindTag(std::__cxx11::basic_string<char, std::char_traits, std::allocator > const&, char*&, unsigned int const&, unsigned int&) const'

It seems that there is a discrepancy between the names in the libbamtools.a file and in the one I compile:

nm -A ReconsReferenceBAM.o |grep -i findtag ReconsReferenceBAM.o: U _ZNK8BamTools12BamAlignment7FindTagERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEERPcRKjRj

But this symbol is there:

nm -A /home/gabriel/.conda/envs/test_gabriel/lib/libbamtools.a |grep -i findtag /home/gabriel/.conda/envs/test_gabriel/lib/libbamtools.a:BamAlignment.cpp.o:00000000000012d0 T _ZNK8BamTools12BamAlignment7FindTagERKSsRPcRKjRj

Notice the missing basic_stringIcSt11char
However, if I compile the program using my local build, it links
without any problem because the symbol has the basic_stringIcSt11char:

nm -A /tmp/bamtools/build/src/api//libbamtools.a |grep -i findtag /tmp/bamtools/build/src/api//libbamtools.a:BamAlignment.cpp.o:0000000000001690 T _ZNK8BamTools12BamAlignment7FindTagERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEERPcRKjRj

I think this is due to the version of the compiler.

apeltzer:
Hi! @dpryan79 , this is the packaging issue I mentioned in the Gitter channel recently. You updated the bamtools package but still, that didn't resolve the issue if I'm not mistaken (the update was installed in a clean environment). 

[...]

PR #10026, opened 2018-07-22
Schmutzi recipe

apeltzer:

Implemented all the changes for 1.5.5.5 now, bamtools is still somewhat the issue. `make install` does something now as per the changes that @grenaud added in https://github.com/bioconda/bioconda-recipes/issues/9794


apeltzer:
As `libgab` uses bamtools, I had to change the order of the build process there. However, there were missing dependencies to `jsoncpp` missing for the bamtools recipe. Probably also the case in the main bamtools recipe?

dpryan79:
@apeltzer Regarding jsoncpp, that's distributed with bamtools I think. I don't know whether that's included in the compilation by default. If not presumably it just requires a configure option change.

apeltzer:
Hi @dpryan79 , it looks like I found the missing bits and pieces. Had that issue in my local env when testing locally but on CircleCI thats gone. Just patching the last issue (some restructuring of the archive in 1.5.5.5, then things should pass and run through...)

apeltzer:
Hm, tests fail because of missing `utils.h` for example:

```
Executing transaction: ...working... done
+ contDeam
utils.h getCWD failed on  c
Tests failed for schmutzi-1.5.5.5-pl526h1c23ad8_0.tar.bz2 - moving package to /opt/conda/conda-bld/broken
```

apeltzer:
Same story as before @dpryan79 .... not a single step further :-(  Maybe @grenaud knows something about it too? 

```
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
+ contDeam
utils.h getCWD failed on  c
Tests failed for schmutzi-1.5.5.5-pl526h1c23ad8_0.tar.bz2 - moving package to /opt/conda/conda-bld/broken
WARNING:conda_build.build:Tests failed for schmutzi-1.5.5.5-pl526h1c23ad8_0.tar.bz2 - moving package to /opt/conda/conda-bld/broken
removing: schmutzi-1.5.5.5-pl526h1c23ad8_0.tar.bz2
TESTS FAILED: schmutzi-1.5.5.5-pl526h1c23ad8_0.tar.bz2
```

apeltzer:
@grenaud looks like this line here is the "issue"... 

https://github.com/grenaud/libgab/blob/e55806c1f4df93fd516968ce70c9f10bf383f5e0/utils.h#L1357

grenaud:
@apeltzer  I added an exception for a user who complained that under his MacOS, there was no PATH_MAX which is bizarre. Line 1357 should work, I redirect a char * to a stream, what is the specific problem?

apeltzer:
Hi @grenaud, 

compilation of the recipe works fine and then we try to test it which fails unfortunately, with the error message mentioned above
```
utils.h getCWD failed on  c
```
Apparently, when we compile it the way its done here in the compile script, the individual binaries don't work.... even though that should be the case. Seems to look after `utils.h`...

grenaud:
so can I replicate these tests someone?

apeltzer:
Yeah! Just check out the branch, and install conda build in a conda environment and run something like:

`conda build meta.yaml` on the schmutzi recipe. That should run through till the test fails...

Or use one of the approaches described here: https://bioconda.github.io/contribute-a-recipe.html#test-locally

grenaud:
@apeltzer  was a solution found, I still need to address this once my workload goes down a bit.

apeltzer:
Not that I know of, but I'd rather re-open the PR once we have some time to figure out what's going on :-) (otherwise need to merge in `master` from time to time as this repository moves quite quickly ;-)

grenaud:
I am increasingly thinking that I should just remove bamtools and put htslib instead, it would be more time efficient.

apeltzer:
Would probably save you hours of work in the mid and long term...

Issue #9930, opened 2018-07-16
QIIME 2 on bioconda


ebolyen:
Hey all,

I just wanted to outline our rough plan for getting QIIME 2 on bioconda:

Right now a key flaw in how we distribute on conda is we don't exactly know what our compatible version ranges are for the dependencies. What we do is let conda figure out *a* dependency resolution, and then we verify it.

What we're going to be moving towards in the very near future is create some internal pipelines for testing version bounds. This will attempt to increment the max version for a given library and verify that the tests pass, and then commit that new max to the repo. This will allow us to trust dependency resolution at install time which will play better with bioconda and allow users to pick and choose which plugins they want installed (this ties in well with some other near-term goals).

Once that is in place (more or less). We'll create a post-release pipeline that issues a PR (via our helper bot @q2d2) updating the recipes here. 

Does it make sense to issue a single PR touching all related recipes, or are multiple PRs preferred?

Also, I suspect we may need to add a few dependencies to conda-forge. But ideally after this is done our channel order will just be `-c bioconda -c conda-forge`. 

cc @thermokarst and @bgruening

bgruening:
@ebolyen I would like to introduce you to @rhpvorderman. Ruben is creating automatically PRs for his software, like this one: https://github.com/bioconda/bioconda-recipes/pull/9941. Maybe you can share some code.

> Right now a key flaw in how we distribute on conda is we don't exactly know what our compatible version ranges are for the dependencies.

Why is this such a big problem for Qiime? Are you using some non-public APIs that often break? 

> Does it make sense to issue a single PR touching all related recipes, or are multiple PRs preferred?

As you like and what is more convenient for you.

> Also, I suspect we may need to add a few dependencies to conda-forge. But ideally after this is done our channel order will just be -c bioconda -c conda-forge.

Ping me on conda-forge, I can review and merge PRs there as well.

Thanks @ebolyen you rock!

ebolyen:
> Ruben is creating automatically PRs for his software, like this one: #9941. Maybe you can share some code.

Awesome! 

> Why is this such a big problem for Qiime? Are you using some non-public APIs that often break?

[...]

PR #10039, opened 2018-07-23
Added the itsxpress QIIME2 plugin to the itsxpress  build recipe

arivers:


recipes/itsxpress/meta.yaml
```diff
@@ -13,10 +13,12 @@ source:
...
 
 build:
-  number: 0
+  number: 1
   entry_points:
     - itsxpress=itsxpress.main:main
-  script: python -m pip install --no-deps --ignore-installed .
+  script:
+    - python -m pip install --no-deps --ignore-installed .
+    - python -m pip install q2-itsxpress
```
bgruening: This should be a real package. Please see here: https://github.com/bioconda/bioconda-recipes/issues/9930

arivers: Sorry, @bgruening  I don't quite know what you mean.  Q2-itsxpress is a QIIME2 plugin that makes the functionality of itsxpress available in QIIME2. Some plugins are in the core QIIME2 software distribution but mine in not. I had initially tried to create another conda recipe to install q2-itsxpress but it requires QIIME2 as a depedency to run so it can't be installed by bioconda.  I started a discussion on the QIIME forum about this. https://forum.qiime2.org/t/conda-repositories-for-contributed-plugin-packages/4956. 

Then I realized that the plugin could be bundled with itsxpress so when people install itsxpress within a conda environment containing QIIME2 itsxpress is available from within QIIME as well.

Do you have advice on the best way to distribute q2-itsxpress?

arivers:
I am closing this request because I needed to push out a bug fix with itsxpress v1.6.2.  @bgruening  if you could still provide your thoughts on how to release q2-itspress that would be very helpful.

PR #10041, opened 2018-07-23
Remove salmon from the blacklist

lparsons:
Remove salmon from the blacklist

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -188,9 +188,6 @@ recipes/sprinkles
 # link not founds
 recipes/soapcoverage
 
-# newer GCC, local GCC overwrite
-recipes/salmon
-
 # download error
 recipes/soapaligner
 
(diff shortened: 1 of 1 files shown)


lparsons:
Hmmm, for some reason the CircleCI checks didn't seem to realize that `salmon` was removed from the blacklist. Am I missing something or is something broken?

bgruening:
See here: https://github.com/bioconda/bioconda-recipes/pull/9943

lparsons:
Thanks for the reference @bgruening, I was looking for that... I really just wanted to see why it was blacklisted to see if I could help. I suppose I can just leave that all to you guys. I'm working on some updates to the galaxy wrapper, so I guess I'll target 0.10.2 for now and hold off on 0.11 unless #9943 gets resolved.

bgruening:
I'm not sure I can help @rob-p much with the OSX error. Maybe we skip OSX for now?

lparsons:
I'm fine with skipping OSX for now, it would be useful to have 0.11 on Linux at least. I'm not really sure where the limitation is, but it seems maybe bioconda's build system isn't quite able to be flexible enough atm to meet the specific needs of salmon?

rob-p:
I agree it would be useful to have a linux build for 0.11.0, since it's a non-trivial upgrade from 0.10.2.  I'm not quite sure what the issue with building on OSX is.  Previously (before the migration to cb3), there was an OSX related issue that was resolved by telling the build it needed llvm >= 4.0, which resolved the issue.  However, it seems that in the midst of the migration, such a specification isn't quite possible?

PR #10043, opened 2018-07-23
remove universal binaries for ViennaRNA

s22chan:

### DIFF
--- a/recipes/viennarna/2.4.5/build.sh
+++ b/recipes/viennarna/2.4.5/build.sh
@@ -5,7 +5,6 @@
 ##
 if [ `uname` == Darwin ] ; then
     extra_config_options="--disable-openmp 
-                          --enable-universal-binary
                           --without-python3
                           LDFLAGS=-Wl,-headerpad_max_install_names"
 fi
--- a/recipes/viennarna/2.4.5/meta.yaml
+++ b/recipes/viennarna/2.4.5/meta.yaml
@@ -3,7 +3,7 @@ package:
   version: 2.4.5
 
 build:
-  number: 2
+  number: 3
 
 source:
   url: https://www.tbi.univie.ac.at/RNA/download/sourcecode/2_4_x/ViennaRNA-2.4.5.tar.gz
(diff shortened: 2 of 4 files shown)


s22chan:
* the flag in configure causes a 1.33-2.00x slowdown (docs say it
interferes with some link-time optimizations)
* the flag is unnecessary since its intent was to bind with Apple's
stock python distribution (rather than conda)

```
echo 'GCGCUUCGCCGCGCGCC&GCGCUUCGCCGCGCGCA\nGCGCUUCGCCGCGCGCC&GCGCUUCGCCGCGCGCA\nGCGCUUCGCCGCGCGCC&GCGCUUCGCCGCGCGCA\nGCGCUUCGCCGCGCGCC&GCGCUUCGCCGCGCGCA\nGCGCUUCGCCGCGCGCC&GCGCUUCGCCGCGCGCA' | time RNAcofold --noPS -p
```

## before ##
RNAcofold --noPS -p  0.02s user 0.00s system 89% cpu 0.030 total

## after ##
RNAcofold --noPS -p  0.01s user 0.00s system 74% cpu 0.022 tota

bgruening:
You need to bump the build/number if you want to trigger a new build and pkg.

bgruening:
ping @s-will 
Thanks!

s22chan:
@s-will Hi, that was a smaller repro string, but for 50k of dimers, we can see execution time drop from 2mins to 1min.

LTO is disabled, so I'm unsure why the generated code is faster (icache? other linker optimizations?)

s-will:
Thanks for the quick response (and of course, many thanks for fixing this---apparently quite significant---issue).
It seems that LTO for ViennaRNA is still deactivated on Mac (due to some other problems) --- maybe @raumzeit has some idea what causes this speedup.

s22chan:
FYI, enabling LTO (or OpenMP) doesn't seem to make RNAcofold or RNAplex run any faster; not sure about other utilities.

bgruening:
Thanks all!

PR #10044, opened 2018-07-23
Update pysam to 0.15

kyleabeauchamp:

I am testing the bioconda build for pysam 0.15, as well as linked against libdeflate.


kyleabeauchamp:
OK, there's something strange about the dependencies here in that when I didn't pin HTSlib to 1.9, it grabbed 1.7 for some reason (and erroring out during pysam compilation).  So I've chosen to pin to 1.9...

kyleabeauchamp:
I've tested building the same branch locally with via docker/circleci and I can confirm (via `ldd`) that the SO file are linking to libdeflate.

kyleabeauchamp:
Separately (https://github.com/samtools/htslib/pull/744), I've discovered that HTSLib links to both libdeflate AND (vanilla) zlib, so I've bumped the HTSLib recipe to include that explicit zlib dependence.

recipes/pysam/meta.yaml
```diff
@@ -20,22 +20,24 @@ requirements:
   build:
     - {{ compiler('c') }}
   host:
-    - htslib
-    - samtools
-    - bcftools
+    - htslib 1.9
```
bgruening: We should increase the version here: https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/bioconda_utils-conda_build_config.yaml#L12

PR #10089, opened 2018-07-26
Add GeneNoteBook recipe

holmrenser:


holmrenser:
@jpuritz thanks for trying, but https://github.com/bioconda/bioconda-recipes/pull/10089/commits/83ac7f7d69d9e78c02090398f4570b4a1f1c29a0 is actually not a feasible solution. I should probably have clarified a bit more in #10095, but the symlink is necessary for the genenotebook executable to work and find dependencies. (It is using nodejs [commander](https://github.com/tj/commander.js) for git-style sub-commands, which requires the executable to be in the same folder as the sub-executables)

PR #10104, opened 2018-07-26
mac recipe and added bioperl

mshakya:


mshakya:
Hi @bgruening, just updated the bump (not bill) number to 1 for the main phame recipe. Thanks!

mshakya:
Hi, can this be merged now?

rvalieris:
why is there a separate recipe for mac ? can't the existing recipe build for both platforms ?

mshakya:
I separated the recipe to mac and linux because there is no Hyphy (one the dependencies for the tool) recipe for mac in conda yet.

rvalieris:
but it still works without it ?
in that case you can use a selector like this:
```
 - hyphy >=2.3.12 # [not osx]
```

rvalieris:
also, are you the developer of this package ?
I noticed the version of the package on conda is different from the version on the git repo, please try to sync both versions if possible.

mshakya:
I still works, but will not have all the functionality. Thanks for the tip about not using separate recipe. I am the developer and will make sure that the version number matches in my next push. Thanks again.

mshakya:
I addressed all the issues above. I think this can be merged now. Thank you

PR #10109, opened 2018-07-27
Gembs

karl616:


recipes/gembs/meta.yaml
```diff
@@ -0,0 +1,61 @@
+{% set version = "conda_v2" %}
```
rvalieris: it would be best if the version of the package reflected the actual version of the software.

recipes/gembs/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+{% set sha256 = "1ead3f20a36a40c0737ec58c2e783b9c1cb6a9de0317c0009964d1b164c598a0" %}
+{% set bcftoolsversion = "1.8" %}
+{% set bcftoolssha256 = "4acbfd691f137742e0be63d09f516434f0faf617a5c60f466140e0677915fced"%}
+
+package:
+  name: gembs
+  version: {{ version }}
+
+source:
+  - url: https://github.com/karl616/gemBS/archive/{{ version }}.tar.gz
```
rvalieris: is this a fork from the original repo ? is the original one not being maintained anymore ?

recipes/gembs/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    sha256: {{ sha256 }}
+  - url: https://github.com/samtools/bcftools/releases/download/{{ bcftoolsversion }}/bcftools-{{ bcftoolsversion }}.tar.bz2
```
rvalieris: is the bcftools source files being used ? its not obvious from the build.sh.

karl616: Yes, there is a custom compiled bcftools version installed to the python library location. It makes use of the plugin system of bcftools.
The installation script downloads bcftools, but I found it better to refer to it like this.

recipes/gembs/meta.yaml
karl616: Thank you for pointing this out. I'll have to ask for an up to date release. I'll definitely change this.

recipes/gembs/meta.yaml
karl616: This is leftovers from testing, connected to the funky version above. Should be changed.

recipes/gembs/meta.yaml
rvalieris: but is the script going to use this one if its there ? 
because if its going to download another one either way I think its best not to put this here.

karl616: The script will prioritize the instance installed to the library folder, with a fall-back to the main-path.

The download process is part of a Makefile. If the folder is already populated by conda, the download step during installation will be skipped. So yes, the file referred here is the one being used.

karl616:
@rvalieris Thanks for the feedback and help!

PR #10130, opened 2018-07-29
tcoffee 11.00.8cbe486

pditommaso:


pditommaso:
@bioconda/core Does anybody have any clue what this error means ?

```
08:32:56 BIOCONDA DEBUG lint recipes/tcoffee
	recipe	check	info	fix	should_use_compilers
0	recipes/tcoffee	should_use_compilers	"{'should_use_compilers': True, 'fix': 'use {{ compiler(""c"") }} or other new-style compilers'}"	"use {{ compiler(""c"") }} or other new-style compilers"	True
Exited with code 1
```

apcamargo:
I think this error is caused because your recipe is not formatted for Conda-build 3. I think this will be a easy fix. Look here:

https://bioconda.github.io/cb3.html#compiler-tools

apcamargo:
This may be more complicated than I though. This is the recipe from Biobuilds:

https://github.com/biobuilds/biobuilds/tree/master/t-coffee/11.00

Can anyone from @bioconda/core help here?

rvalieris:
when the build.sh starts you will already be inside a directory of the tar specified in source.url, so you can skip those commands.

apcamargo:
@pditommaso Please, pull the changes to your local repository. Your latest commit reverted the linting fixes.

pditommaso:
> Please, pull the changes to your local repository. Your latest commit reverted the linting fixes. 

Oops, you are right.

rvalieris:
> t_coffee: error while loading shared libraries: libstdc++.so.6: cannot open shared object file: No such file or directory

looks like it uses c++ stuff as well, add `- {{ compiler('cxx') }}` to the build section.

apcamargo:
@rvalieris I was working on this already. Thanks!

apcamargo:
From the T-COFFEE documentation:

> If no valid location can be found for HOME_4_TCOFFEE, the program exits. If you are running T-Coffee on a server, we recommend to hard set the following locations, where your scratch is a valid location. 
> 
> ```
> HOME_4_TCOFFEE=”your scratch”
> TMP_4_TCOFFEE=”your scratch”
> DIR_4_TCOFFEE=”your scratch”
> CACHE_4_TCOFFEE=”your scratch”
> NO_ERROR_REPORT_4_TCOFFEE=1
> ```

@rvalieris What you suggest here? I imagine this problem won't happen in a real life situation, but we need to do something to pass this test.

rvalieris:
not sure I understand, are these needed for the program to run normally ?

read the conda docs on link scripts to execute things after the package is installed: https://conda.io/docs/user-guide/tasks/build-packages/link-scripts.html

but only if its really needed, otherwise leave it for the user decide if he wants to set this.

pditommaso:
If I'm understanding well, the problem is only on Mac because of some missing env vars. Conda does not allow the definition of custom vars, right?

rvalieris:
you can set vars on the post-link.sh script,  but since this only happens on the test maybe try changing the test command to:
`HOME=/tmp/ t_coffee -version`

pditommaso:
Not even `HOME` is defined in the build environment?

pditommaso:
Nice, it passes the tests. 

The only thing I'm not happy this recipe has been renamed to `t-coffee` instead of `tcoffee`. The project [github repo](https://github.com/nextflow-io/patterns/tree/master/optional-output) does not contain the hyphen and (as maintainer of the tcoffee) I would prefer to not have the `-` in the name.

Is there anything preventing that?

apcamargo:
Thanks @rvalieris and @pditommaso !

pditommaso:
Fair enough.

apcamargo:
@pditommaso Are those packages really necessary to run T-COFFEE? That's a lot of stuff.

As far as I know, T-COFFEE don't need any of these packages in order to run.

pditommaso:
That's not even the complete list of external packages used by T-Coffee. See at [this link](https://github.com/cbcrg/tcoffee/blob/f7bb7f8dd2c259dfde535195ab821889fb508cba/lib/data_headers/tclinkdb.txt) all of them.

apcamargo:
According to the link below, those packages are not required and are used only in some specific cases (I suppose they're necessary only to run M-COFFEE):

http://tcoffee.readthedocs.io/en/latest/tcoffee_installation.html#third-party-packages

I'm reluctant to include all these dependencies only because this can make the installation quite heavy (mainly because of BLAST) and I think the user can install by himself if he wants to use it. As I don't use T-COFFEE heavily, I don't know the needs of most users, so I'll leave that decision up to you.

In case we include those dependencies, I think we now need to discover why OSX is blacklisted in 'poa'.

pditommaso:
I'm in the team of Cedric Notredame, the author of T-Coffee. One of main advantage of T-Coffee over other aligns is integrate and recombine pairwise alignments provide by other tools. Our custom installation package do install all that tools. 

I don't really understand what's the problem.

apcamargo:
There's no problem! I'm sorry if I was rude, but I wanted to know if the installation of those extra packages is really necessary.

pditommaso:
No problem :) 

Yes, most important T-Coffee modes like Expresso, M-Coffee, TM-Coffee do require that packages. Giving the user the ability to install all of them with Conda would be a big bonus. Our plan is to progressively add the missing tools as Conda recipes. 

Is `poa` is failing it can be removed for now, I guess.

apcamargo:
Unfortunately, I think that unless we find out why OSX is blacklisted in the `poa` recipe, this is the only thing we can do for now.

pditommaso:
OK, I've removed `poa`. Thanks for your feedback.

apcamargo:
Hey @pditommaso , I was taking a look at [selectors](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#preprocessing-selectors) and I think you could use it here. I’ve never used them, but I think you could use it as:

`- poa # [not osx]`

By doing that `poa` would be a requirement for Linux, but not macOS. What do you think?

pditommaso:
Sounds nice, I've added that.

apcamargo:
I had to merge master into the branch because to resolve a problem with an unrelated package that was failing to build. I think everything is done now

rvalieris:
this looks great, I will go ahead and merge this

mbargull:
The squashed commit's message on `master` is
```
* tcoffee 11.00.8cbe486
* Revert meta.yaml
* Update folder name
* add DOI [ci skip]
* Fixing test
* Added t-coffee required runtime deps
* Exclude poa from MacOS build
* Add poa dependency
```
and thus includes `[ci skip]` which made the CI skip the build :rofl: :man_facepalming:.
Nothing to worry about though, since the nightly builds should pick it up in the next few hours :wink:.

rvalieris:
doh !! I didn't noticed that. I will keep an eye out for it next time.

PR #10154, opened 2018-07-31
Added recipe for pybio-utils

RickGelhausen:


bgruening:
@RickGelhausen you are missing dependencies according to `RuntimeError: Setuptools downloading is disabled in conda build. Be sure to add all dependencies in the meta.yaml  url=https://pypi.org/simple/misc/`

bgruening:
It seems to be this dependency https://github.com/dieterich-lab/pybio-utils/blob/master/setup.py#L71

RickGelhausen:
@bgruening Hi Björn, thank you for your help.
I am currently working on a "workaround" as the creators of this package seem to have messed up their naming conventions. Egg and I were able to find the problem and will fix it as soon as possible. I just wanted to check whether my local circleci does its job.

bgruening:
You could include a patch and fix the setup.py file for example.

bgruening:
@RickGelhausen do you need any help here?

RickGelhausen:
yes, I am currently waiting for Florian to return, as this package is currently not that important for our work. But I want to get it up and running as soon as possible. I think, the problem lies with the pymisc-utils package though.
The thing is that the creators of these packages, created the pymisc-utils package as "misc" and import it like this in python. So what I tried is to replace these imports by:
```
pymisc_utils = __import__("pymisc-utils") 
import pymisc_utils.<module>
```
but I get:

```
ImportError: No module named 'pymisc-utils'
```
which seems to be installed now:

```
...
    pymisc-utils:         0.2.10-py35h37e70f2_0                 bioconda   
...
```
That makes me think that the "pymisc-utils" package needs some extra attention. But its just as likely, that I made a mistake somewhere. I will push it again, so you can see the full error message.

PR #10156, opened 2018-07-31
forgi: loosened the dependencies

croth1:

Loosenes dependencies of forgi according to https://github.com/ViennaRNA/forgi/issues/32


dpryan79:
FYI I’m out of town so if it passes testing then go for it.

croth1:
@bioconda/core can somebody quickly have a look at the changes? Is the conda-forge style pip installation ok for bioconda?

rvalieris:
yes thats fine, the important part is to avoid bundling dependencies on the same package (which `--no-deps` does).

PR #10167, opened 2018-07-31
Add deepsvr

bainscou:

### DIFF
--- a/recipes/deepsvr/meta.yaml
+++ b/recipes/deepsvr/meta.yaml
@@ -49,10 +49,10 @@ requirements:
     - h5py
 
 test:
-  requires:
-    - nose
   imports:
     - deepsvr
+  commands:
+    - deepsvr --version
 
 
 about:
--- a/recipes/deepsvr/meta.yaml
+++ b/recipes/deepsvr/meta.yaml
@@ -51,6 +51,15 @@ requirements:
 test:
   imports:
     - deepsvr
+  comands:
+    - LC_ALL=C.UTF-8 LANG=C.UTF-8 deepsvr --version # [linux]
+    - LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 deepsvr --version # [osx]
+    - LC_ALL=C.UTF-8 LANG=C.UTF-8 deepsvr --help # [linux]
+    - LC_ALL=en_US.UTF-8 LANG=en_US.UTF-8 deepsvr --help # [osx]
+  extra:
+    container:
+      extended-base: true
+    
 
 about:
   home: https://github.com/griffithlab/deepsvr
--- a/recipes/deepsvr/meta.yaml
+++ b/recipes/deepsvr/meta.yaml
@@ -4,7 +4,7 @@ package:
 
 source:
   url: https://github.com/griffithlab/DeepSVR/releases/download/v0.1.0/deepsvr-0.1.0.tar.gz
-  sha256: d13725430676dba28f10140fb0f85cb2ca7ba4a899f8d555dff60dcac64b4292 
+  sha256: c9f394ffbdeb0fbe56b519c3aaa8e59ad00c06675b903abeeb1a4e21b4f04c54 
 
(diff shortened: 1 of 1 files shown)


bainscou:
@bioconda/core can you please review?

recipes/deepsvr/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - scikit-learn
+    - matplotlib
+    - seaborn
+    - bam-readcount
+    - convert_zero_one_based
+    - click
+    - tensorflow<=1.0.1
+    - keras==2.0.4
+    - h5py
+  run:
```
corburn: Do the `run` dependencies need to be locked to these exact versions?

recipes/deepsvr/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - convert_zero_one_based
+    - click
+    - tensorflow<=1.0.1
+    - keras==2.0.4
+    - scikit-learn
+    - h5py
+
+test:
+  requires:
+    - nose
```
corburn:
```yaml
imports:
  - deepsvr
commands:
  - deepsvr --version
```

recipes/deepsvr/meta.yaml
```diff
@@ -0,0 +1,60 @@
+package:
+  name: deepsvr
+  version: "0.1.0"
+
+source:
+  url: https://github.com/griffithlab/DeepSVR/releases/download/v0.1.0/deepsvr-0.1.0.tar.gz
+  sha256: b95763c9fdfc718938ecf537a94714e2b4cbb1f083859faa7c86c2d90d22ef01 
+
+build:
```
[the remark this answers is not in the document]
bainscou: Sure, I can try this!

recipes/deepsvr/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  noarch: python
+  script:
+    - flake8 . -v --config=.flake8rc
+    - nosetests --verbose
+    - python setup.py install
+  entry_points:
+    - deepsvr = deepsvr.cli:main
+
+requirements:
+  build:
```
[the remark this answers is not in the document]
bainscou: What is the difference?

recipes/deepsvr/meta.yaml
bainscou: Unfortunately, yes. Tensorflow is under such active development that the APIs are constantly changing. It took me a lot of time to find the exact combination of versions that would install on mac and linux on circleci. If there is a better way to do this I am open to suggestions.

recipes/deepsvr/meta.yaml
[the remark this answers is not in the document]
corburn: Not much difference for recipes that do not link against shared libraries; mainly following the conda [documentation](https://github.com/conda/conda-docs/blob/master/docs/source/user-guide/tasks/build-packages/define-metadata.rst#requirements-section).

recipes/deepsvr/meta.yaml
bainscou: When I add a test of the commands I get the following error.

`SERR FileNotFoundError: [Errno 2] No such file or directory: 'locale'`

I am not sure why this is happening because it runs fine locally and on the local circleci.

For now I am going to remove the check. @corburn any ideas on how I can fix this?

corburn: @bainscou based on the error, you may need to use the extended-base image #5842.

bainscou: That solved my issue! Thanks!

recipes/deepsvr/meta.yaml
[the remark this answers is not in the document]
bainscou: I think the only consideration there is tensorflow, which uses the cuda library for GPU support. However, I can't see a use case where the GPU acceleration will make an appreciable difference in the use of this particular software.

PR #10170, opened 2018-08-01
Dropseq tools

cgirardot:

### DIFF
--- a/recipes/Drop-seq_tools/meta.yaml
+++ b/recipes/Drop-seq_tools/meta.yaml
@@ -20,7 +20,6 @@ source:
 
 requirements:
   build:
-    - {{ compiler('cxx') }}
   run:
     - openjdk >=7
 
--- a/recipes/Drop-seq_tools/meta.yaml
+++ b/recipes/Drop-seq_tools/meta.yaml
@@ -19,7 +19,6 @@ source:
   md5: 28533edd82867ed75867b05785cf8834
 
 requirements:
-  build:
   run:
     - openjdk >=7
 
--- a/recipes/Drop-seq_tools/meta.yaml
+++ b/recipes/Drop-seq_tools/meta.yaml
@@ -26,7 +26,7 @@ requirements:
 
 test:
   commands:
-    - [[ $(TrimStartingSequence -- --version 2>&1) == "1.13(7bed8f4_15130080339)" ]]
+    - '[[ $(TrimStartingSequence -- --version 2>&1) == "1.13(7bed8f4_1513008033)" ]]'
 
 extra:
   notes: |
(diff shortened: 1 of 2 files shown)


cgirardot:
@bioconda/core this PR adds a new package. This is Drop-seq_tools suite from the Mc Carroll lab http://mccarrolllab.com/dropseq/. The suite basically ships as a jar and shell wrappers for each sub-command. The recipe is very simple and is basically an adaptation of the Je-suite one already in bioconda. I could run CircleCI locally successfully.

recipes/Drop-seq_tools/meta.yaml
```diff
@@ -0,0 +1,31 @@
+{% set name = "Drop-seq_tools" %}
```
bgruening: Can we have small letters here?

recipes/Drop-seq_tools/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  license: MIT License
+  license_family: MIT
+  summary: |
+    Package fot the analysis of Drop-seq data developed by Jim Nemesh in the McCarroll Lab
+package:
+  name: dropseq_tools
+  version: {{ version }}
+
+build:
+  number: 1
```
bgruening: Please start with 0.

recipes/Drop-seq_tools/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  url: http://mccarrolllab.com/download/1276/{{ name }}-{{ version }}.zip
+  md5: 28533edd82867ed75867b05785cf8834
+
+requirements:
+  run:
+    - openjdk >=7
+
+test:
+  commands:
+    - '[[ $(TrimStartingSequence -- --version 2>&1) == "1.13(7bed8f4_1513008033)" ]]'
```
bgruening: Include a few more tests e.g. the --help from the nodes?

recipes/Drop-seq_tools/meta.yaml
cgirardot: the name is used to build the file name to download from their web site. I'd rather keep it like this if possible.

recipes/Drop-seq_tools/meta.yaml
cgirardot: sure

recipes/Drop-seq_tools/meta.yaml
cgirardot: ok can do this, give a sec

cgirardot:
@bgruening , just pushed a new version in which : 
- I added a test on every sub-command of the package
- build number is 0
- kept the upper case in name though, lemme know if thats OK

cgirardot:
@bgruening I am not sure if I should do something for this macOS plan ?

PR #10174, opened 2018-08-01
Me de com recipe

schmic05:

### DIFF
--- a/recipes/MeDeCom/build.sh
+++ b/recipes/MeDeCom/build.sh
@@ -1,0 +1,4 @@
+#!/bin/bash
+mv DESCRIPTION DESCRIPTION.old
+grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
+$R CMD INSTALL --build .
--- a/recipes/MeDeCom/meta.yaml
+++ b/recipes/MeDeCom/meta.yaml
@@ -20,7 +20,6 @@ requirements:
     - r-pracma
     - r-gtools
     - r-gplots
-    - r-parallel
     - r-RcppEigen >0.3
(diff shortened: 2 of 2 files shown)


schmic05:
Hey @bioconda/core !
I'd like to add MeDeCom as an R package to bioconda. MeDeCom is not yet on bioconductor and is only available as a GitHub package [(github.com/lutsik/MeDeCom](https://github.com/lutsik/MeDeCom)). When building MeDeCom, the system needs around 6GB of memory, since major preprocessing steps and object are conducted during package building. I noticed that circleci only provides up to 4GB of memory. Therefore, the build fails due to running out of memory. Is there an easy way to work around that? Would appreciate any help here.

bgruening:
@schmic05 wow, why are 6GB needed? We build Qt and stuff like that and it uses less :)
There is no good way to build such a package. The only way I can imagine is that an admin is building it locally and pushes it. But ideally the build procedure should be optimized :(

schmic05:
@bgruening Thanks for the fast reply. MeDeCom builds an index-like structure during the build process to improve running time. We will further work on reducing this, but this won't be easily possible. Just wanted to know, if there is an easy solution. We'll let you know as soon as we found a solution.

bgruening:
@schmic05 I haven't looked at the tool but the index-like structure needs to be rebuild everytime? Why not building it offline and ship it externally?

schmic05:
@bgruening I'm still pretty new to the tool and don't know everything about it, so I'll need some time to figure it out. But I guess you're right, it should be possible to build it only once.

daler:

PR #10177, opened 2018-08-01
Fasttreecondbuild3

zachcp:

Fix FastTree build.

Because FastTree is a single file it gets put into the `SRC_DIR` and we must change the build script accordingly. IN doing so I can see that CLANG will not build FastTreeMP so I moved that into a linux specif portion of the build

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1513,9 +1513,6 @@ recipes/discovar-denovo
 recipes/blast/2.2.21
 recipes/blast-legacy/2.2.22
 
-# clang: error: no such file or directory: 'FastTree.c'
-recipes/fasttree
-
 # seqan errors
 recipes/bowtie/1.0.0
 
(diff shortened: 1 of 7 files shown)


thermokarst:
Hey there @zachcp!

> IN doing so I can see that CLANG will not build FastTreeMP so I moved that into a linux specif portion of the build

So that means this recipe no longer builds `FastTreeMP` for darwin, right? Besides the `clang` issue, is there any other reason why `FastTreeMP` couldn't be built for darwin? I am happy to lend a hand with adding this back in for darwin, just want to check if there are any unforeseen problems with this. I am happy to open an issue, too, just wanted to check in with you, first. Thanks!

zachcp:
@thermokarst ,  yes `fasttreeMP` is  currently not building on darwin. If you can get a workaround to get the builds working, by all means do so. I believe  `conda-build-3` is flexible in regards to compilers so it may be simpler than I imagined but I am note sure.  I am happy to take a look at any PRs you generate.

zach cp

PR #10182, opened 2018-08-01
Updated ont-tombo to version 1.4.

marcus1487:

### DIFF
--- a/recipes/ont-tombo/meta.yaml
+++ b/recipes/ont-tombo/meta.yaml
@@ -13,7 +13,7 @@ source:
 build:
   entry_points:
     - tombo = tombo.__main__:main
-  number: 1
+  number: 0
 
 requirements:
   host:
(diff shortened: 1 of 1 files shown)


recipes/ont-tombo/meta.yaml
```diff
@@ -36,7 +36,7 @@ requirements:
   run:
     - python
     - setuptools >=18.0
```
bgruening: are you sure you need setuptools in run?

marcus1487:
Probably don’t need setuptools. Is there a specifications page for extras/identifiers? I think I tried to add some a while back and couldn’t find them.

marcus1487:
Found this [section](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#extra-section) but where do these extra bits end up?

bgruening:
@marcus1487 for example here: https://github.com/bioconda/bioconda-recipes/blob/2c0ca47b12eba1fa06d0c3378670d3b50bfdfbf3/recipes/fastspar/meta.yaml#L53

PR #10199, opened 2018-08-02
New recipe: Slamdunk

t-neumann:


recipes/slamdunk/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - cmake
+    - wget
+   
+  host:
+    - python=2.7
```
bgruening: please skip python 3 build in the build/skip section.

recipes/slamdunk/build.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+python setup.py install --single-version-externally-managed --record=record.txt
```
bgruening: please use pip install, have a look here: https://github.com/bioconda/bioconda-recipes/blob/31f2bc37fa14a7a1b51451383bcb39909f50f741/recipes/pymzml/0.7.5/meta.yaml#L12

recipes/slamdunk/meta.yaml
t-neumann: This should suffice right?

```
build:
  skip: True  # [not py27]
```

recipes/slamdunk/build.sh
t-neumann: It won't work unfortunately. When I run this build command, it will just skip with

```
07:34:27 BIOCONDA INFO Nothing to be done for recipe recipes/slamdunk
```

recipes/slamdunk/meta.yaml
t-neumann: Whenever I skip python3 build, it won't build my package anymore.

Neither
```
skip: True  # [not py27]
```

nor

```
skip: True  # [py3k]
```

works. So I'm leaving this for now.

t-neumann:
@bioconda/core I have followed @bgruening suggestions and deactivated the py3 build, but that will stop building the package alltogether ("nothing to do"). In addition, all local tests (CircleCI and via bootstrap.py) pass, but the ci/circleci: test-macos test fails which is hard for me to reproduce locally.

It basically says 
```
Successfully installed slamdunk-0.3.1
```

and then during packaging there's a lot of errors like this
```
Fixing linking of /System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation in /Users/distiller/project/miniconda/conda-bld/slamdunk_1533199159813/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/lib/python2.7/site-packages/slamdunk/plot/Rslamdunk/sp/libs/sp.so
New link location is None
```

eventually resulting in 

```
install_name_tool failed with exit status 1
```

Any ideas on that?

rvalieris:
when using the skip on py3, you need to leave the python dependency without any version restrictions. both on host and run sections. that might be why.

rvalieris:
for the osx error you can try adding on the build.sh:
```
if [[ $OSTYPE == darwin* ]]; then
     export LDFLAGS="${LDFLAGS} -headerpad_max_install_names"
fi
```

rvalieris:
also note that nextgenmap, samtools and varscan are already present in bioconda as packages, it would be much better for your package to depend on them instead of building them on `setup.py`

t-neumann:
@rvalieris We need to deliver those packages via PyPI to make the package self-contained and it would be quite a hassle to rewrite everything for bioconda. In addition, we're running a customized Nextgenmap version, Varscan2 will go with the next major release, so it's basically only samtools that's left. I hope that's allright and I'm fully aware that ideally these packages would be installed as dependencies.

recipes/slamdunk/meta.yaml
```diff
@@ -0,0 +1,83 @@
...
+  sha256: '{{sha256}}'
+  url: "https://github.com/t-neumann/slamdunk/archive/v0.3.1.tar.gz"
+
+build:
+  entry_points:
+    - slamdunk=slamdunk.slamdunk:run
+    - alleyoop=slamdunk.alleyoop:run
+    - splash=slamdunk.splash:run
+  number: 0
+  skip: True # [py3k]
```
bgruening: this should be `# [py3k or osx]`

t-neumann:
@bioconda/core All checks passing and all comments addresse, .buildkite restored. Ready for merge.

bgruening:
@t-neumann thanks a lot!

johanneskoester:
What a cool name :-)

rvalieris:
@t-neumann  I understand this isn't possible for pypi, but I noticed the setup.py check for `external*` vars before building samtools, etc. would it not be possible to set these vars on the conda recipe to avoid bundling them here ?

t-neumann:
@rvalieris That was the initial plan. Turns out pip passes those install-options also to all required modules, crashing the whole install process and there's no workaround. This issue has been already listed for some time without any solution here:

https://github.com/pypa/pip/issues/1883

Would have loved to include that, but right now impossible.

t-neumann:
Hi @bioconda/core Team,

there are some discrepancies between the bioconda-utils-build-env Docker container (https://hub.docker.com/r/bioconda/bioconda-utils-build-env/) build and what you guys are doing on your side to build the conda packages, causing my package build to be corrupt, even though the local build works without issues.

This is what I do to setup the local build:

First start the Docker container:

```
docker run -ti bioconda/bioconda-utils-build-env /bin/bash
```

Then checkout the bioconda-recipes and build slamdunk locally:

```
git clone https://github.com/bioconda/bioconda-recipes.git
cd bioconda-recipes/
bioconda-utils build recipes config.yml --force --packages slamdunk
```

Then I create one environment with the local build and one with the bioconda build:

```
conda create -n slamdunkLocal -c local slamdunk
conda create -n slamdunkBioconda -c bioconda slamdunk
```

Now the local environment perfectly fine creates the symlink to the custom Nextgenmap build and Nextgenmap runs fine:

```
source activate slamdunkLocal
(slamdunkLocal) [root@58fdb0c78e9e bioconda-recipes]# ls -lah /opt/conda/envs/slamdunkLocal/lib/python2.7/site-packages/slamdunk/contrib/ngm
lrwxrwxrwx 1 root root 28 Aug  4 06:43 /opt/conda/envs/slamdunkLocal/lib/python2.7/site-packages/slamdunk/contrib/ngm -> nextgenmap/bin/ngm-0.5.2/ngm
source deactivate
```

The bioconda build however replaces this symlink with a copy apparently, causing ngm to crash with an error.

```
source activate slamdunkBioconda
(slamdunkBioconda) [root@58fdb0c78e9e bioconda-recipes]# ls -lah /opt/conda/envs/slamdunkBioconda/lib/python2.7/site-packages/slamdunk/contrib/ngm
-rwxrwxr-x 2 root root 466 Aug  3 13:48 /opt/conda/envs/slamdunkBioconda/lib/python2.7/site-packages/slamdunk/contrib/ngm
(slamdunkBioconda) [root@58fdb0c78e9e bioconda-recipes]# /opt/conda/envs/slamdunkBioconda/lib/python2.7/site-packages/slamdunk/contrib/ngm
/opt/conda/envs/slamdunkBioconda/lib/python2.7/site-packages/slamdunk/contrib/ngm: line 16: /opt/conda/envs/slamdunkBioconda/lib/python2.7/site-packages/slamdunk/contrib/ngm-core: No such file or directory
/opt/conda/envs/slamdunkBioconda/lib/python2.7/site-packages/slamdunk/contrib/ngm: line 16: exec: /opt/conda/envs/slamdunkBioconda/lib/python2.7/site-packages/slamdunk/contrib/ngm-core: cannot execute: No such file or directory
source deactivate
```

What is going on here - I thought the bioconda-utils build from bioconda-utils-build-env container is exactly what is used to build the final package?
And is there a workaround for this?

t-neumann:
Addressed in #10290

PR #10210, opened 2018-08-02
Bump libsequence to 1.9.4.  Update dependencies, which are build-only!

molpopgen:
Compiler and run unit tests after build.

I also fixed the LICENSE bit, which was out of date.

### DIFF
--- a/recipes/libsequence/build.sh
+++ b/recipes/libsequence/build.sh
@@ -2,5 +2,6 @@
 
 CXXFLAGS="-O3 -DNDEBUG" CPPFLAGS="-I$PREFIX/include $CPPFLAGS" LDFLAGS="-L$PREFIX/lib -Wl,-rpath,$PREFIX/lib $LDFLAGS" ./configure --prefix=$PREFIX
 make
+make check
 make install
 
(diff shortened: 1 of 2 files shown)


molpopgen:
@bioconda/core The OS X build failed, but apparently for reasons other than failing to build the package.  There's a message about "issuing a restart could fix this", but there's not a restart button, so I need help...

bgruening:
I restarted it. One question, is zlib needed during build-phase or not. Because currently you have not defined it.

molpopgen:
> I restarted it. One question, is zlib needed during build-phase or not. Because currently you have not defined it.

Thanks!  

No, there are no longer any dependencies at run-time.  The only build dependency is boost, and that is only for compiling/running the unit tests, which I've added to this PR.

Is there a way to trigger a rebuild that I missed?

bgruening:
Try to find this link:

grafik

And rerun the failed step in the circle-ci workflow.

molpopgen:
So, this is ready to merge, with the caveat that we have to skip OS X for a while.  The case-insensitive file system there means that there are some name conflicts to to subdirs that only differ by capitalization.  That'll go away as deprecated features are removed.

bgruening:
Thanks @molpopgen!

PR #10212, opened 2018-08-02
New recipe for SolexaQA.

tesujimath:


tesujimath:
This builds just fine locally with conda build 3.  I think I will wait until the dust settles on the conda build 3 migration.

rvalieris:
mulled build only support the test::command part, the test::files won't be available inside the container, so the test could be failing because of that.

tesujimath:
Is there a recommended workaround?  I can think of echoing the test data into the file before running the command.  That's pretty lame though.

recipes/solexaqa/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+  number: 0
+  skip: True # [osx or win]
+
+source:
+  url: https://downloads.sourceforge.net/project/solexaqa/src/SolexaQA++_v{{ version }}.zip
+  sha256: {{ sha256 }}
+
+requirements:
+  build:
+    - gxx_linux-64 >=7.3.0
```
bgruening: Please don't do this. Instead, overwrite the compiler config by using conda_build_config.yaml as shown here: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/salmon/conda_build_config.yaml

recipes/solexaqa/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - make
+  host:
+    - boost
+    - icu
+    - libstdcxx-ng
```
bgruening: not needed if you follow above tip

recipes/solexaqa/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - zlib
+    # R is not a strict dependency, but can provide enhanced functionality,
+    # so we leave it as an option for the user, not a hard requirement.
+    # - r
+
+test:
+  files:
+    - tiny.fastq
+  commands:
+    - SolexaQA++ dynamictrim tiny.fastq >/dev/null
```
bgruening: /dev/null can be removed.
Is the tiny.fastq file part of the pkg or only part of the tarball from which you install? If its part of the package you could `which SolexaQA++`\share\tiny.fastq the file maybe.

recipes/solexaqa/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - libstdcxx-ng
+    - zlib
+  run:
+    - boost
+    - icu
+    - libstdcxx-ng
+    - zlib
+    # R is not a strict dependency, but can provide enhanced functionality,
+    # so we leave it as an option for the user, not a hard requirement.
+    # - r
```
bgruening: r-base would be one alternative. I think we agreed to include optional dependencies by default - also to make the containers work by default.

recipes/solexaqa/meta.yaml
tesujimath: Not part of the package.  It's a trivial test file for exercising the application.  I'm currently struggling to get it into the test environment, because of the lack of support from mulled-build for test::files.  My current thinking is 4 echo commands, since it's a 4 line file.  Ugh!

recipes/solexaqa/meta.yaml
tesujimath: I think in 99% of the uses of SolexaQA, R is not used - maybe even 5 9s ;-). Even r-base is not a small package.  My strong preference is to omit it.

recipes/solexaqa/meta.yaml
mbargull: Never tried it, but in theory
```yaml
test:
  commands:
    - |
      SolexaQA++ dynamictrim <(echo -n '@HWI-ST871:91:C0JGJACXX:8:1302:10862:157183 1:N:0:TAGCTT
      TGTCTCTGCTTTTGAATATGCTATCTAGGTTGGTCATAACTTTCTTTCCAAGGAGTAATCATCTTTTAATTTCATGGCTACAATCACCATTTGCAGTTATT
      +
      CCCCFFFFFHHHHHJIJIJJJIIJJIGJJIJIJHIGHIJJJJJJJJIJJJJIIIGFEGIGGIJIIIJI>FHIJIHHJIHIHG>CEFFFFFED@C:>AC@>;
      ')
```
could work.
Or a little simpler
```yaml
test:
  commands:
    - |
      echo -n '@HWI-ST871:91:C0JGJACXX:8:1302:10862:157183 1:N:0:TAGCTT
      TGTCTCTGCTTTTGAATATGCTATCTAGGTTGGTCATAACTTTCTTTCCAAGGAGTAATCATCTTTTAATTTCATGGCTACAATCACCATTTGCAGTTATT
      +
      CCCCFFFFFHHHHHJIJIJJJIIJJIGJJIJIJHIGHIJJJJJJJJIJJJJIIIGFEGIGGIJIIIJI>FHIJIHHJIHIHG>CEFFFFFED@C:>AC@>;
      ' | SolexaQA++ dynamictrim
```
if `SolexaQA++ dynamictrim` is also able to read from stdin.
If that doesn't work, a less elegant way would be to put this all into one line and use `printf` instead of `echo` and replace all linebreaks by `\n`.

recipes/solexaqa/meta.yaml
tesujimath: Done

recipes/solexaqa/meta.yaml
tesujimath: OK

recipes/solexaqa/meta.yaml
tesujimath: Solved with an inline printf.  (But support for test::files would be a really useful addition!)

tesujimath:
@bgruening I addressed all the comments except the R dependency, and it builds cleanly now.  I think a dependency on R would be a bad idea.  I suspect almost no-one cares for that, and it adds a lot of bloat.  So I prefer to omit it.  Are you OK to merge it like it is now?

recipes/solexaqa/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - boost
+    - icu
+    - zlib
+    # R is not a strict dependency, but can provide enhanced functionality,
+    # so we leave it as an option for the user, not a hard requirement.
+    # - r
+
+# apparently bioconda mulled-build doesn't support test files, so we work around it like this:
+test:
+  files:
```
bgruening: This can now be removed as well as the file.

lmrodriguezr:
Hello. Is there a reason why this package won't build in macOS? I think all the requirements are available, and it is possible to build on macOS since the original release distributes macOS binaries.

Thanks!

tesujimath:
@lmrodriguezr Perhaps it will.  Can you test it and make a PR?

lmrodriguezr:
@tesujimath Thanks! I'll give this a try.

I'm getting this very strange environmental variable set:
```bash
BOOST_PATH=/usr/local/Caskroom/miniconda/base/envs/testenv/conda-bld/solexaqa_1567549192865/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho
```

Have you seen anything like that? I just started a fresh new conda environment, removed the osx from Skip, and tried building. If fails because it cannot find the ICU libraries, and I suspect this might be part of the issue. 

This is the line in the context of the full output: [gist](https://gist.github.com/lmrodriguezr/a99065fc07eb986bc37d8c0a772e54dc#file-gistfile1-txt-L140).

tesujimath:
The placeholder stuff in the path is key to the way conda works.  It's a grotesque hack, which is required so that conda can edit in the installation paths into any binaries it creates, and know that there is enough space for pretty long pathnames.  Truly, ugh!

I'm sorry I don't have any time to assist with this, really.

PR #10228, opened 2018-08-03
Update plant_tribes_gene_family_phylogeny_builder to v1.0.3

corburn:

### DIFF
--- a/recipes/plant_tribes_gene_family_phylogeny_builder/build.sh
+++ b/recipes/plant_tribes_gene_family_phylogeny_builder/build.sh
@@ -1,0 +1,14 @@
+#!/bin/sh
+
+mkdir -p $PREFIX/bin
+mkdir -p $PREFIX/config
+mkdir -p $PREFIX/data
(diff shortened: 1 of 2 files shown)


gregvonkuster:
@corburn it looks like you're updating the content of these PlantTribes recipes while keeping the versions the same (actually the source versions have not changed).  Can you clarify your intent?  Your new ```build.sh``` scripts will install some rather large files (e.g., ~/config directory) which is ok, just wanted to alert you.  

I wrote these recipes to install the PlantTribes pipelines into Galaxy environments - I work closely with the lab that authored the PlantTribes repo.  The Galaxy environment does not require anything beyond the actual pipeline and this is why my version of the ```build.sh``` script moved only the pipeline script into the ```bin``` directory.

Of course, these PlantTribes pipelines can all be used outside of Galaxy as well, although doing so require manual intervention (e.g., manually installing PlantTribes scaffolds into the ~/data directory.  Are you updating these recipes so that they can be installed with conda, but used outside of a Galaxy environment?

PR #10247, opened 2018-08-03
Unblacklist r-wgcna

apcamargo:

This PR removes r-wgcna from the blacklist.


apcamargo:
@druvus I've removed WGCNA from the blacklist, as you've suggested, but it can't be built for some reason. The error that that happens in the test is the same as [another PR](https://github.com/bioconda/bioconda-recipes/pull/10142) that I've submitted some days ago.

apcamargo:
Closed this PR because r-wgcna will be removed from the blacklist once we switch to R 3.5.1

PR #10250, opened 2018-08-03
NovaSplice

aryakaul:

### DIFF
--- a/recipes/novasplice/build.sh
+++ b/recipes/novasplice/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
--- a/recipes/novasplice/meta.yaml
+++ b/recipes/novasplice/meta.yaml
@@ -3,8 +3,7 @@ package:
   version: "0.0.3"
 
 build:
-  #noarch: True
-  skip: True # [py27]
+  noarch: python
   number: 0
   entry_points:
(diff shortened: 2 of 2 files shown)


aryakaul:
Please let me know if there's some way I can improve the recipe. Some stuff didn't make sense to me, but I followed other examples and it worked. For example, I would like a command-line version of the tool to be installed when the user runs `conda install novasplice` and I have a console entry point listed both in my `meta.yaml` file and in the `setup.py` file within my repository. Are one of those redundant? Also, my only test is checking to see if ```novasplice --help``` throws an error. Are there other recommended ways to test CL tools?

Thanks!
Arya

bgruening:
@aryakaul please feel free to ask any question if things are not clear.

aryakaul:
Thanks for the help! Here are my questions so far:
1. Why do I need to have an entry point described in my setup.py and in the meta.yaml? Shouldn't conda be able to access the entry point described in the setup.py without reproducing it in meta.yaml?
2. My only test is checking to see if the dependencies are installed, and if the command line version is working `novasplice --help`. Are there other types of tests I should do to ensure it's installed correctly? My tool requires a reference fasta to work, and I don't think it makes sense to upload such a large file to the bioconda servers just for the purpose of testing the algorithm.
3. What exactly does the value in `build: number` do within the meta.yaml? The conda documentation says: `The build number should be incremented for new builds of the same version. The number defaults to 0` But isn't this effectively the version number?

Thanks again!

Issue #9404, opened 2018-06-14
Conda build 3 migration


johanneskoester:
@bioconda/all: Via PR #9402, we have migrated all recipes to Conda Build 3, and at the same time updated our build system to conda build 3. It will take some time until all the recipes have been rebuilt. We will take care of that, but you can expect massive slowdowns of our build system until this is done. In order to let the migration go as smooth as possible, I would ask you to only create and merge PRs that are absolutely necessary until the migration is complete.

Once the migration is complete, I will close this issue and let you know.

# TODO
* [ ] fix failing recipes
* [ ] once all recipes have been built successfully uncomment `--git-range` in upload-linux and upload-macos jobs in circleci
* [ ] release new version of bioconda-utils

# Workflow
1. Look for the last [bulk build](https://github.com/bioconda/bioconda-recipes/commits/bulk) and select recipes to fix
2. Create a new branch from master, fix the selected failing recipes and PR the changes to master.
3. Once **all** recipes in bulk have been fixed this way, merge master into bulk again to trigger building eventually remaining recipes (this is only allowed for admins): 
    ```
    git checkout master
    git pull
    git checkout bulk
    git pull
    git merge --squash -X theirs master
    # fix remaining conflicts
    git commit -a
    git push
    ```

johanneskoester:
@bioconda/core: git-range based filtering on master is enabled again. In order to enforce a bulk build, you have to add `[bulk]` to the commit message.

johanneskoester:
This also means that we can easily fix failed recipes via committing just a fix to an individual recipe without adding `[bulk]` to the commit message. The ideal workflow is:

1. Look for the last `[bulk]` build in the commit history of the master branch
2. Fix all failing recipes and commit/pr the changes to master.
3. Once the fixes are uploaded, push a new `[bulk]` commit to master, in order to build the remaining recipes.

johanneskoester:
@bioconda/lib we get quite some compiler errors because the new gcc7 is less permissive regarding C/C++ style. Is it possible to set it to some kind of "old behavior mode" for such recipes?

SoapZA:
-fpermissive is an option, but this really is a feature and not a bug. Improving the strictness imposed on C++ code is an important part of making code more portable, I dont think there‘s a quick for this.

johanneskoester:
Thanks! I agree. But I don't think it should be our job to fix the sources. 
Seems that this does not solve all problems. With xtandem, we now get:
```
[...]

PR #10263, opened 2018-08-05
Update mlst to 2.11

tseemann:

Also removes it from blacklist.

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -2114,7 +2114,6 @@ recipes/perl-bio-db-sam
 recipes/perl-datetime-format-strptime/1.73
 recipes/perl-bio-eutilities
 recipes/perl-moose/2.2009
-recipes/mlst
 recipes/mlst/2.9
 recipes/perl-sanger-cgp-vagrent
 recipes/ensembl-vep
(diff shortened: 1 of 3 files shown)


druvus:
@tseemann mlst is currently blacklisted. You need to remove it from the `build-fail-blacklist` file. Still some perl dependencies (i.e. bioperl and deps) that needs to be rebuilt before it will pass.

tseemann:
@druvus i don't understand what any of that means. is there any documentation? i was surprised when it downgraded my Perl to 5.22 from 5.26.  Is this because of the modules it uses?  Some of those are core models and shouldn't be conda packages anyway.  @Slugger70 ?

corburn:
@tseemann bioconda was updated recently updated to [use conda 3](https://github.com/bioconda/bioconda-recipes/issues/9404). To expedite the transition, some problematic recipes were put on a [blacklist](https://github.com/bioconda/bioconda-recipes/blob/8eae705b3740499ee67ffed3c3130be1d8426d83/build-fail-blacklist#L2134) until their particular issues were resolved. In your pull request, you will need to remove your recipe from the [blacklist](https://github.com/bioconda/bioconda-recipes/blob/8eae705b3740499ee67ffed3c3130be1d8426d83/build-fail-blacklist#L2134) so it will build. I do not know anything about the perl issue.

tseemann:
Thanks @corburn - i'm guessing it was a perl dep that was the cause of the blacklist.

PR #10288, opened 2018-08-07
Pin R builds to 3.5.1

dpryan79:

The bumps the R version that packages are built against from 3.4.1 to 3.5.1. @bioconda/core Please have a look and discuss when you get a chance. I know a bunch of R packages have been blacklisted while waiting for this change.


bgruening:
@dpryan79 we are not using the nev matrix anymore. Instead we are using this file https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/bioconda_utils-conda_build_config.yaml and the corresponding one from conda-forge. This means we need to wait for conda-forge to make this bump. Which kind of makes sense, because we also need to rebuild all r-* pkgs on conda.forge.

We migrated in the last week all r-* pkgs on conda-forge to cb3. Now we need to rebuild against new compilers and r3.5.1. Then we can start here :(

PR #10326, opened 2018-08-08
add recipe for PlasFlow

shiltemann:
@willemdek11 here a start for a plasflow bioconda recipe

### DIFF
--- a/recipes/plasflow/build.sh
+++ b/recipes/plasflow/build.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# install PlasFlow
+$PYTHON -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv
+
+# install models
+git clone https://github.com/smaegol/PlasFlow.git
+cp -r PlasFlow/models $PREFIX/bin/
--- a/recipes/plasflow/build.sh
+++ b/recipes/plasflow/build.sh
@@ -1,7 +1,11 @@
 #!/bin/bash
 
+# install tensorflow version 0.10.0rc0
+pip install --no-deps --ignore-installed --no-cache-dir  tensorflow/tensorflow-0.10.0rc0-cp35-cp35m-linux_x86_64.whl
+
+
 # install PlasFlow
 $PYTHON -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv
 
-# install models
+# copy models
 cp -r PlasFlow/models $PREFIX/bin/
--- a/recipes/plasflow/build.sh
+++ b/recipes/plasflow/build.sh
@@ -4,5 +4,4 @@
 $PYTHON -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv
 
 # install models
-git clone https://github.com/smaegol/PlasFlow.git
 cp -r PlasFlow/models $PREFIX/bin/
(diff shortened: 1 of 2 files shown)


recipes/plasflow/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 352d5d7cd73672130e32f3f61c3f2ad95aa684d7aebf7668ab3f2594996c338b
+
+build:
+  number: 0
+  skip: True # [not py35 or osx]
+
+requirements:
+  host:
+    - python
```
bgruening: I assume everything besides python & pip can be removed from the host section.

recipes/plasflow/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+test:
+  imports:
+    - plasflow
+  commands:
+    - PlasFlow.py --help
+
+about:
+  home: https://github.com/smaegol/PlasFlow
+  license: GNU General Public v3 or later (GPLv3+)
+  license_family: GPL3
```
bgruening: license_file: LICENSE

recipes/plasflow/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# install PlasFlow
+$PYTHON -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv
+
+# install models
+git clone https://github.com/smaegol/PlasFlow.git
```
bgruening: Can we checkout a specific commit?

corburn: [Define metedata from multiple sources](https://github.com/conda/conda-docs/blob/f5eb0fe9206809216b08100d07249bfc970a4098/docs/source/user-guide/tasks/build-packages/define-metadata.rst#source-from-multiple-sources)

```yaml
source:
  - url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
    sha256: 352d5d7cd73672130e32f3f61c3f2ad95aa684d7aebf7668ab3f2594996c338b
  - url: https://github.com/smaegol/PlasFlow/archive/v1.1.tar.gz
    sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
    folder: PlasFlow
```

shiltemann:
thanks for the help @bgruening and @corburn :)

recipes/plasflow/meta.yaml
shiltemann: this fails with `ValueError: License file given in about/license_file does not exist in source root dir or in recipe root dir (with meta.yaml)` so I've commented it out for now

bgruening: Interesting, maybe because the source root dir is not clear in your case? Does PlasFlow/LICENSE work?

recipes/plasflow/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# install tensorflow version 0.10.0rc0
+pip install --no-deps --ignore-installed --no-cache-dir  tensorflow/tensorflow-0.10.0rc0-cp35-cp35m-linux_x86_64.whl
```
bgruening: Is there any reason you need this exact version? And if so is the runtime dependency on tensorflow still needed?

shiltemann: yes, plasflow doesnt seem to work properly with the conda version, and its install instructions say to use this exact version (and I left tensorflow in the yaml file so that its dependencies would still be installed by conda and I could use `--no-deps`, but maybe better to just lists those explicitly then yeah?)

PR #10338, opened 2018-08-08
lumpy: trying to fix skipped build

mbargull:

The build for gh-10337 gave
```
20:12:37 BIOCONDA INFO Nothing to be done for recipe recipes/lumpy-sv
```
which is not correct of course. Trying to fix it here.
Best guess so far: The run-only `python` requirement picked up `python 3.*` and thus triggered `skip: True # [not py27]`?

### DIFF
--- a/recipes/lumpy-sv/build.sh
+++ b/recipes/lumpy-sv/build.sh
@@ -1,18 +1,21 @@
 #!/bin/bash
 set -eu -o pipefail
 
-export CPPFLAGS="-I$PREFIX/include"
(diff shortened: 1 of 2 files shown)


mbargull:
In most cases (apart from OS-specific selectors) the `skip: True  # [...]` can be replaced by appropriate version restrictions on `python` & co -- and sometimes, like in this case, it even has to be done that way, apparently.

Other than that, a few tweak were needed to make it build, but all good to go now. Feel free to merge if it looks good to you, @bgruening!

PR #10344, opened 2018-08-09
Salmon gcc7

rob-p:

This is a test to see if we can build salmon using the new compiler (gcc7) ... at least on linux.  This works locally using [the "bootstrap" method](https://bioconda.github.io/contribute-a-recipe.html), so let's see what happens on Circle CI.


rob-p:
Hrmm, this seemed to work .... not so sure what's different between this and before when it failed.  Going to try uncommenting OSX now too.

rob-p:
Hrmmm .. looks like this icu58 issue is still popping up on OSX, but has gone away under linux on the newer compiler.  I'd really like to figure this out so I can provide a mac executable through bioconda.

recipes/salmon/build_config.yaml
```diff
@@ -0,0 +1,6 @@
+c_compiler:
+  - gcc      # [linux]
+  - clang    # [osx]
```
bgruening: Rob can you try to replace clang here with GCC as well?

rob-p: Ok; just pushed.  I didn't even know that building with GCC on OSX was an option here :).

rob-p:
It used clang anyway, it seems ....

rob-p:
It used clang anyway, it seems ....

bgruening:
Not really as we do not build many pkgs with the new compilers. My assumption is that we need to rebuild iuc with the new compilers before we build Salmon :(

rob-p:
Can that be done here, or is that coming from conda-forge?  Can we submit PRs there?

bgruening:
Its a larger effort we are working on. The migration to the new compilers should happen soon. But it will take time, nothing we can influence right now. Its mainly CI build time dependent ... If someone want to buy more CI build-time - more then welcome :)

I will keep you posted, sorry.

rob-p:
Ok, so I get it to build locally *if* I force the boost from anaconda cloud and not the one from conda-forge.  The conda-forge ppl aren't building a c++11 enabled boost :(.  Perhaps there is a more elegant way to do this.

rob-p:
ahh, so no we are getting something different popping up on OSX.  Perhaps this one is solvable.

rob-p:
Now we're stuck on this error under OSX:

`ld: library not found for -lrt`

is there a dependency I need for this using gcc on OSX, or should the entire linker flag go away?

acaprez:
`librt` is part of `glibc`, so yeah, you'd need to remove that flag entirely.  Hopefully you won't get an unresolved symbol error then...

rob-p:
Now it's back at a damn icu58 linker issue under OSX ...

rob-p:
*Note* : This is currently pulling from develop.  I will be cutting a new release (tagged) imminently.  Please don't merge this yet.

recipes/salmon/meta.yaml
```diff
@@ -24,32 +25,32 @@ package:
...
-    #- llvm # [osx]
     - {{ compiler('cxx') }}
     - {{ compiler('c') }}
-    #- clangdev =={{ llvm_version }}  # [osx]
-    #- libcxx =={{ llvm_version }}  # [osx]
+    - libcxx 
   host:
-    - boost
-    - icu
+    - libboost # [linux]
```
bgruening: the selector is now not needed here ...

recipes/salmon/meta.yaml
```diff
@@ -24,32 +25,32 @@ package:
...
     - unzip
     - jemalloc >=5.1.0
     - tbb
   run:
-    #- libgcc # [linux]
-    #- libcxx =={{ llvm_version }}  # [osx]
-    - boost
-    - icu
+    - libcxx 
+    - libboost # [linux]
```
bgruening: same here

recipes/salmon/meta.yaml
```diff
@@ -24,32 +25,32 @@ package:
...
   build:
     - autoconf
     - cmake
-    #- gcc ==4.8.5 # [linux]
-    #- llvm # [osx]
     - {{ compiler('cxx') }}
     - {{ compiler('c') }}
-    #- clangdev =={{ llvm_version }}  # [osx]
-    #- libcxx =={{ llvm_version }}  # [osx]
+    - libcxx
```
bgruening: is this needed?

recipes/salmon/meta.yaml
rob-p: Thanks @bgruening --- yes, I agree that this is not needed (and probably the same with the libcxx below).  Should I try to bump the version and remove these and see if it compiles?

bgruening: That would be great, yes. Thanks.

PR #10359, opened 2018-08-09
Initial edlib-python

pb-cdunn:
* Depends on edlib C++.
* Created via `conda skeleton pypi`, but renamed to edlib-python
  b/c the C++ conda module is already called `edlib`.


pb-cdunn:
I need some help with this recipe. `lint` passes for me locally, but the CircleCI `lint` message says:
```
18:18:30 BIOCONDA DEBUG lint recipes/edlib-python
    recipe    check    info    fix    should_be_noarch    should_not_use_fn
0    recipes/edlib-python    should_be_noarch    "{'should_be_noarch': True, 'fix': 'add ""build: noarch"" section'}"    "add ""build: noarch"" section"    True    
1    recipes/edlib-python    should_not_use_fn    {'should_not_use_fn': True, 'fix': 'URL should specify path to file, which will be used as the filename'}    URL should specify path to file, which will be used as the filename        True
```
(I'll deal with `fn` separately.)

Why does it care about `noarch`? According to the [Bioconda Python Recipe Guidelines](https://bioconda.github.io/guidelines.html):

> Update 7 Feb 2018: Previously we had recommended that if the recipe is a pure Python package, it should be marked as a “noarch” package (details). However due to technical incompatibilies we can’t do this – so please DO NOT use "noarch" for now.

pb-cdunn:
Actually, it's not even building properly locally. (My test must have re-used a build.) This is a PyPI package. It should be *simple*!

pb-cdunn:
`conda skeleton` did not work.

Trying directions from the module itself now. Oddly, this works for me in a
**virtualenv**, but it seems to fail in **bioconda** because `python setup.py`
is using the ext-module name `edlib.cpython-35m-darwin.so` instead
of `edlib.so`.

corburn:
@pb-cdunn the linter cares about noarch because the build system has been updated (https://github.com/bioconda/bioconda-recipes/issues/9404), but the documentation has not been updated.

PR #10364, opened 2018-08-10
Unblacklist a few recipes

abretaud:


abretaud:
@bioconda/core would it be possible to delete the perl-text-soundex package from bioconda channel on anaconda.org (https://anaconda.org/bioconda/perl-text-soundex/files)? The recipe as been deleted here as it's now in conda-forge (https://anaconda.org/conda-forge/perl-text-soundex/files).
As they have the same version number, repeatmasker takes the bioconda one which still depends on perl-threaded, and the build fails

PR #10365, opened 2018-08-10
Bumped Abricate build number for new perl and bioperl

Slugger70:


druvus:
@Slugger70 I think it is entrez-direct that is causing the problem for osx. It it depending on perl but it is not rebuilt yet for 5.26 and then causing perl-threaded to be pulled in looking at the 5.22 path.

dpryan79:
@Slugger70 Can this be closed now that version 0.8.10 has been incorporated?

Slugger70:
yes

PR #10368, opened 2018-08-10
nanopolish out of blacklist

mmiladi:

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -2191,6 +2191,6 @@ recipes/abricate/0.7
 recipes/abricate
 
 # clang
-recipes/nanopolish
+# recipes/nanopolish
 
 ####################### OSX errors above ###############
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -2191,6 +2191,5 @@ recipes/abricate/0.7
 recipes/abricate
 
 # clang
-# recipes/nanopolish
 
 ####################### OSX errors above ###############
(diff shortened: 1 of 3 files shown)


bgruening:
You should probably use `https://anaconda.org/conda-forge/toolchain_c_linux-64` and `toolschain_cxx_linux-64` on OSX as well.

bgruening:
For that add https://github.com/bioconda/bioconda-recipes/blob/master/recipes/mawk/conda_build_config.yaml into the repo and specify `toolchain_c_linux-64` and  `toolchain_cxx_linux-64`. Not sure if this will work :(

mmiladi:
Thanks for the hints@bgruening. So do you mean to not use clang?

`https://anaconda.org/conda-forge/toolchain_c_linux-64` has files only for linux. How can they be deployed on OSX? :-|

rob-p:
I was seeing the same thing [here](https://github.com/bioconda/bioconda-recipes/pull/10344).  I'd love to use GCC/G++ on OSX if it is possible.  The clang is from the OSX 10.9 SDK, so it's pretty old and a bit quirky.

bgruening:
@rob-p I don't love it. But can you do as I did here for salmon as well?

rob-p:
@bgruening : So is there a way to use the new gcc on linux and 4.8.5 on OSX?  If not, I guess we can temporarily revert and force to use 4.8.5 on both platforms.

bgruening:
Oh yes, sorry. This is using old GCC on Linux and OSX. Old not new, sorry.

mmiladi:
@bgruening Thanks a bunch!

Issue #10297, opened 2018-08-07
cromwell requires gnu find


karl616:
Hi,
cromwell requires findutils rather than the busybox version

https://github.com/bioconda/bioconda-recipes/blob/a2414a163fc43a0bc607659a14180c8afa4522c9/recipes/cromwell/meta.yaml#L21-L23

some of the implementations of wdl functionality requires the `find -empty` flag. In a normal conda installation this isn't a problem as you can fall back on the host version of find, but it fails when converted to a biocontainer.

I'm trying to add findutils to conda-forge for this reason:
 https://github.com/conda-forge/staged-recipes/pull/6414

PR #10383, opened 2018-08-11
cromwell uses find during runtime. With a normal conda installation i…

karl616:
…t can fall back on the host installation, but when running in a biocontainer it only has access to the busybox version which lacks some necessary options.

### DIFF
--- a/recipes/cromwell/meta.yaml
+++ b/recipes/cromwell/meta.yaml
@@ -7,7 +7,7 @@ package:
 
 build:
   noarch: python
-  number: 0
+  number: 1
 
 source:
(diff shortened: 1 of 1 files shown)


karl616:
In reference to this:
https://github.com/bioconda/bioconda-recipes/issues/10297

karl616:
@notestaff @tomkinsc 
Hi cromwell maintainers,
I would like to add findutils as a runtime dependency of cromwell, but before I do it I thought I'd give you the chance to react/object.
Best,
Karl

tomkinsc:
👍

PR #10403, opened 2018-08-13
Sumaclust recipe

JTFouquier:
* [X ] This PR adds a new recipe.
* [X ] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).


JTFouquier:
It said it was able to be merged, but now it's saying checks aren't completed yet. Happy to wait. :) Let me know if there's anything I can improve. I'm not really familiar with this process. Thanks!

JTFouquier:
Hi there, sorry, I'm only able to see the "fail" when I submit a PR. It looked good before. Any thoughts? Thank you!

screen shot 2018-08-13 at 6 02 35 pm

bgruening:
Mh, the error here is `sumaclust: command not found` :(

recipes/sumaclust/meta.yaml
```diff
@@ -0,0 +1,24 @@
...
+  url: https://git.metabarcoding.org/obitools/sumaclust/uploads/59ff189079b9e318f07b9ff9d5fee54b/sumaclust_v1.0.31.tar.gz
+  '{{ hash_type }}': '{{ hash_value }}'
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+
+test:
+  commands:
+    - sumaclust -h > /dev/null
```
bgruening: You can remove the `> /dev/null`

recipes/sumaclust/meta.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+about:
+  summary: Sumaclust. For more information see url.
```
bgruening: A short summery would be nice.

recipes/sumaclust/conda_build.sh
```diff
@@ -0,0 +1,4 @@
+
+bash conda_build.sh
```
bgruening: conda_build.sh is not part of this PR.
I have checked the tarball and there is a makefile included. Does `make install` work instead?

JTFouquier: `make` by itself is what is used to install it locally.

recipes/sumaclust/meta.yaml
JTFouquier: They have the longest 'summary'. Didn't want to butcher it but will try to do it justice. :) Thanks.

recipes/sumaclust/conda_build.sh
JTFouquier: Thanks for your help @bgruening!  I tried naming the file `build.sh` and it fails (wish some sort of hint to rename it as `conda_build.sh`, and when I do name it `conda_build.sh` it works for me locally. Everything is green before I change things. I attempted to add `make` into the bash script because that is how to install it locally. BTW I'm definitely more bioinformatics than CS if I sound confused. :grin: I have a few more tools I need to get into Bioconda, so any help is great so I can repeat the process. 

Thank you!

bgruening:
Oh ok. But this one is normal. This is created internally. What you see here is that your tool is actually working unter Linux but crashing under OSX.
This is because the makefile wants to use openmp as you can see in the error message: `clang: error: unsupported option '-fopenmp'`. And this is because we try to build with clang under OSX and not GCC.

bgruening:
@JTFouquier all green. My last commits are not really recommended its a hack to get your tool build on OSX. Alternatively, you could also patch your tool to be able to use LLVM (clang). Let the other tools coming :)

JTFouquier:
Thanks so much for your help and for being so fast!!

PR #10438, opened 2018-08-15
Update IQ-TREE to 1.6.6

apcamargo:


recipes/iqtree/meta.yaml
```diff
@@ -19,15 +19,18 @@ requirements:
...
     - eigen
     - zlib
+    - openmp # [osx]
   run:
     - zlib
+    - openmp # [osx]
 
 test:
   commands:
-    - iqtree-omp
```
bgruening: Was this dropped from the tool?

apcamargo:
Yes. Since version 1.6.0, the two versions of IQ-TREE  were merged.

http://www.iqtree.org/release/

apcamargo:
I can't get IQ-TREE to build on macOS. I tried to add OpenMP as a dependency but it didn't work.

The compilation guide states: 

> To compile the multicore version, the default installed Clang unfortunately does not support OpenMP (which might change in the near future). However, the latest Clang 3.7 supports OpenMP, which can be downloaded from http://clang.llvm.org.
http://www.iqtree.org/doc/Compilation-Guide

Could the Clang version used in the CI not support OpenMP?

recipes/iqtree/meta.yaml
apcamargo: Yes. Since version 1.6.0, the two versions of IQ-TREE were merged.

http://www.iqtree.org/release/

tseemann:
@apcamargo i did the brew formula for iqtree, i had to patch it even for gcc!
```
    if OS.mac?
      inreplace "CMakeLists.txt",
        "${CMAKE_EXE_LINKER_FLAGS_RELEASE} -Wl,--gc-sections",
        "${CMAKE_EXE_LINKER_FLAGS_RELEASE}"
    end
```

apcamargo:
Hi @tseemann . I'm sorry, I don't think I understood your suggestion. As far as I know, the cmake command you suggested is eqivalent to the one in the build file. Am I missing something?

tseemann:
As openmp isn't supported on MacOS, you might need to remove the `IQTREE_FLAGS=omp` to build on Mac.  Or make it linux only.

bgruening:
@apcamargo LGTM! ping @bioconda/osx for help regarding the OSX if needed.

bgruening:
We could use GCC on OSX as well if @bioconda/osx does not have a better idea.

tseemann:
@bgruening how do you force OSX to use gcc instead of clang for C++ code?
 `compiler("gcc")` ?

bgruening:
@tseemann you can overwrite the globals by putting a https://github.com/bioconda/bioconda-recipes/blob/master/recipes/mawk/conda_build_config.yaml next to your meta.yaml file and replace clang with gcc.

apcamargo:
@tseemann @bgruening I've submitted [another PR](https://github.com/bioconda/bioconda-recipes/pull/10583). I think we can continue our discussion there.

PR #10461, opened 2018-08-16
Update QualiMap to most up2date version with bed support

apeltzer:


corburn:
The build number should start from zero. The 2.2.2a directory is not required. I'm not sure if this is a beta release or date versioning.

apeltzer:
Hey! Will take care of that - the previous release was also from such a date release and used 2.2.2a, so I followed that guideline in here as well.

apeltzer:
That building takes ages - is that normal?

corburn:
@apeltzer probably, yes.

apeltzer:
Build Timeout after 5hours.... hmm...

chapmanb:
Thanks for the his update. I don't appear to have permissions to re-try builds but given nothing changed in the build.sh and it doesn't look too complex, I'm hoping that a restart will build cleanly. Johannes, any chance of re-kicking the build?

apeltzer:
I can also bump it by simply adding a newline somewhere in the build, thus retriggering it. Its a bit weird, that this timed out after 5 hours ?

apeltzer:
Same issue again @chapmanb / @johanneskoester .... will run into a timeout...

chapmanb:
Thanks for re-kicking this. I've been testing locally and it looks like the issue is that the bioconductor dependencies for qualimap are pretty heavyweight. I know you added these as well so probably have strong feelings about them but I commented out `bioconductor-repitools`, which is the worst offender and brings in a whole load of packages. This will at least tell us if it can finish in reasonable time without this and then decide how best to proceed.

apeltzer:
Its a bit weird though, as we'll need these for some steps in qualimap. As I'm building my containers using a conda environment, I can add that as external dependencies but some people might not know whats the problem here - though it seems that helped resolve the issues!

PR #10463, opened 2018-08-17
add clairvoyante v1

aquaskyline:
[\*] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
[\*] This PR adds a new recipe.
[\*] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).
[ ] This PR updates an existing recipe.
[ ] This PR does something else (explain below).


aquaskyline:
@bioconda/core Please review the receipt and merge. Thank you!

clairvoyante/build.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+python2.7 -m pip install intervaltree
```
corburn: Instead of installing with pip, could intervaltree and blosc be moved to the meta.yaml requirements run section?

aquaskyline: I was doing that in an earlier version. However, there exists no intervaltree and blosc for pypy in conda, so I need to pip install intervaltree and blosc using for pypy anyhow. For a better consistency, I also used pip to install the two libs for python.

aquaskyline: @corburn Just a courtesy reminder, please advise how to proceed, thank you.

corburn: @aquaskyline you could try pinging bioconda/core again.

Personally I'd expect a conda package to add commands to the PATH:
```bash
conda install clairvoyante
runDemo
```

From the recipe it looks like users would be expected to run:
```bash
conda install clairvoyante
pypa /path/to/conda/env/bin/clairvoyante/runDemo.py
```

I do not know where a conda recipe should copy a python package that is not using setuptools so it will be found on import. The following is still not correct, but I got as far as trying:
```yaml
package:
  name: clairvoyante
  version: 1.0

source:
  url: https://github.com/aquaskyline/Clairvoyante/releases/download/v1/Clairvoyante_v1.0.zip
  md5: 74b22669edcc08ee7be01b09f52569ee

build:
  noarch: python
  number: 0
  script:
    - mkdir -pv $STDLIB_DIR/site-packages
    - cp -rv clairvoyante dataPrepScripts $STDLIB_DIR/site-packages
  entry_points:
    - demoRun = clairvoyante.demoRun:Run
    #...

requirements:
  run:
    - python =2.7
    - pypy2.7
    - intervaltree
    - blosc
    - tensorflow
    - numpy
    - zlib

test:
  imports:
    - clairvoyante
  commands:
    - demoRun

about:
  home: 'https://github.com/aquaskyline/Clairvoyante'
  license: AGPLv3
  license_file: LICENSE.md
  summary: "A deep neural network based variant caller"
```

aquaskyline: @bioconda/core. Thanks for your suggestions,I've spent a few hours tweaking my tool as well as the recipe to make the best out of both, the main points are as follows:
1. In my tool, I have added an entry point that enables access to all the submodules, named `clairvoyante.py`. It's command-line accessible without a path.
2. I insist on using `build.sh` instead of putting everything in the `meta.yaml`.
3. I've tried installing `blosc` and `intervaltree` using conda. The libraries can be successfully installed, but not accessible. One way or another, I will have to install the two libraries for `pypy2.7`, so for consistency, I choose to do it the same for `python2.7`.
4. @corburn has suggested to put the scripts in $STDLIB_DIR, but I put them in $PREFIX/bin instead because they are programs instead of libraries.

I've tested locally and committed the updated recipe.

aquaskyline:
@bioconda/core Please review the receipt and merge.

aquaskyline:
@corburn Please review and merge. Thank you!

corburn:
@aquaskyline at a glance the recipe looks good. If someone from @bioconda/review doesn't spot an issue I'll merge it tonight local time.

The tests are good with possible alternatives being directly running `clairvoyante.py`, `clairvoyante.py --help`, or verifying `clairvoyante.py [submodules]` works.

It sounds like the tweaks you made addressed my concern about the commands not being executable without a path. I would remove the `channels` and requirements `build` sections; they are not required.

For the sake of mentioning it, an alternative to my attempt at adding `entry_points` and your approach of adding a submodule command could have been adding `#!/bin/env python` to the top of the programs and setting the executable bit. They would still need to be directly under the `$PREFIX/bin/` instead of `$PREFIX/bin/clairvoyante/` or `$PREFIX/bin/dataPrepScripts/` to be found on the `$PATH`

PR #10470, opened 2018-08-17
Update MAnorm to v1.1.4

haydensun:

### DIFF
--- a/recipes/manorm/meta.yaml
+++ b/recipes/manorm/meta.yaml
@@ -1,19 +1,19 @@
 package:
   name: manorm
-  version: 1.1.3
+  version: 1.1.4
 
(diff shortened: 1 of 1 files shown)


recipes/manorm/meta.yaml
```diff
@@ -1,19 +1,19 @@
...
 build:
-  number: 1
+  number: 0
   skip: True  # [not py27]
   entry_points:
     - manorm=manorm.cmdline:main
 
 requirements:
-  host:
+  build:
```
corburn: `host` was the correct section name.

Could replace `skip: true` with `noarch: python` in the build section by adding `python >=3` to the requirements host and run sections.

PR #10475, opened 2018-08-17
Add SomaticSeq

rhpvorderman:

Hi my attempt at adding somaticseq. Unfortunately testing locally did not work. As the tests were skipped. Some help is needed here.


rhpvorderman:
It will succeed at all the tests because it skips them.
@bioconda/core any idea how to fix it?

recipes/somaticseq/meta.yml
```diff
@@ -0,0 +1,36 @@
...
+  url: https://github.com/bioinform/somaticseq/archive/v{{version}}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+
+requirements:
+  host: []
+  run:
+    - python=3
```
rvalieris: remove the `=3` and add `skip: True # [py27]` on the build: section.

rhpvorderman:
Thank you @rvalieris for your comments. Unfortunately the package is still skipped in testing.
I would like to be able to test this package before it is merged into bioconda. But I cannot do so if the utils do not allow me to do so.

rhpvorderman:
> try renaming meta.yml to meta.yaml

@rvalieris  That did the trick. Thanks. The recipe should now be ready for reviewing.

recipes/somaticseq/build.sh
```diff
@@ -0,0 +1,24 @@
...
+cd $prog_dir
+for script in *.sh *.py
+do
+  ln -s $prog_dir/$script $PREFIX/bin/$script
+done
+
+mkdir $PREFIX/bin/utilities
+for file in $prog_dir/utilities/*
+do
+  ln -s $prog_dir/utilities/$file $PREFIX/bin/utilities/$file
```
rvalieris: the output changed but it still is not building.. 
$file here includes the "$prog_dir/utilities/" path so this won't work. you could use `f2=$(basename $file)`. same for the other for.

try adding python to the host section.

rvalieris:
cool, looks like its building now.

rhpvorderman:
Thanks @rvalieris ! It seems like its failing. But this is a good thing. I can now actually start testing this recipe locally. Thanks for pointing me in the right direction. I will start fixing the recipe and ping you when it is finished. Thanks again!

rhpvorderman:
@rvalieris It is fixed now. Had to skip OS X unfortunately since it gave too many problems, and I cannot truly debug this not owning one. It builds fine on linux.

PR #10478, opened 2018-08-17
MethylExtract recipe

phue:

As this is my first recipe, I kindly ask for review
@bioconda/core

### DIFF
--- a/recipes/methylextract/meta.yaml
+++ b/recipes/methylextract/meta.yaml
@@ -1,0 +1,35 @@
+{% set version = "1.9.1" %}
+{% set sha256 = "f9c1cb9d01814745cd37384790ee6bf71dfa69030414a11eda728fcdcc505d40" %}
+
+package:
+  name: methylextract
+  version: {{ version }}
+
+build:
+  number: 0
+
+source:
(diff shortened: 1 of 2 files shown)


recipes/methylextract/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+about:
+  home: http://bioinfo2.ugr.es/MethylExtract/
+  license: GNU General Public License v3 (GPLv3)
+  license_family: GPL3
+  summary: High-Quality methylation maps and SNV calling from whole genome bisulfite sequencing data
+
+test:
+  commands:
+    - which MethylExtract.pl
```
bgruening: Can you run the scripts as well? With some `--version` or `--help` flag?

phue: Thanks for your review. 
Unfortunately, the scripts don't have such flags. And running them without arguments results in exitcode 255, which makes the test fail.

I guess another approach would be to grep for some string in the usage information that gets printed when calling the script, if you prefer that over 'which'.

phue:
done :)

bgruening:
Cool! Thanks a lot!

PR #10479, opened 2018-08-17
New: python-intervaltree @ 2.1.0

pb-cdunn:
Actual Python package is called 'intervaltree'. Note that bioconda
already has 'intervaltree_bio'

### DIFF
--- a/recipes/python-intervaltree/meta.yaml
+++ b/recipes/python-intervaltree/meta.yaml
@@ -1,0 +1,26 @@
+{% set version = "2.1.0" %}
+
+package:
+  name: "python-intervaltree"
+  version: {{ version }}
+source:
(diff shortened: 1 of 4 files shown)


recipes/python-intervaltree/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - python
+test:
+  imports:
+    - intervaltree
+about:
+  home: https://pypi.org/project/intervaltree/#description
+  license: Apache
+  summary: "Faster than intervaltree_bio"
+extra:
+  web: https://github.com/chaimleib/intervaltree
```
corburn:
```
about:
    dev_uri: https://github.com/chaimleib/intervaltree
```

pb-cdunn: Thanks. I'll add that. I have to add `python-sortedcontainers` too.

corburn: Both packages have current versions in conda-forge.

PR #10508, opened 2018-08-18
Update pypeflow to 2.0.4

corburn:


cdunn2001:
I was about to update this. Thanks. Note that `2.0.4` is technically a "pre-release", but it's stable aside from attachments.

Note also that `pb-falcon` now includes the contents of this. (Same version. Different sha256 b/c the top-dir-name differs in the tarball.) We may eventually make `pypeflow` be a dependency of `pb-falcon`, but for now it's easier for us to release several Python repos together.

Also, `pypeflow` now works fine with `networkx-2`, so that constraint can be relaxed. Soon, it should be python-3 compatible.

PR #10525, opened 2018-08-19
Add showit

joshmoore:

A prerequisite of gh-10457


recipes/starfish/showit/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  name: showit
+  version: {{ SHOWIT_VERSION }}
+
+source:
+  url:  https://github.com/freeman-lab/showit/archive/ef76425797c71fbe3795b4302c49ab5be6b0bacb.zip
+  sha1: 64484b77fba4fb641daa964ff28b6cb151810222
+
+build:
+  noarch: python
+  number: 1
```
bgruening: please start with 0

recipes/starfish/showit/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+cd ./lib/python
```
bgruening: please use popd and pushd to change directories

recipes/starfish/showit/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
bgruening: Showit should move up one directory level I think.

joshmoore:
@bgruening : definitely, thanks for all the pointers. Turns out I was struggling with the `cd lib/python` step that had been created by skeleton(?).

recipes/showit/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  number: 0
+
+requirements:
+  host:
+    - setuptools
+    - python
+  run:
+    - python
+    - matplotlib
+    - numpy !=1.13.0
```
bgruening: last question, this numpy version is not allowed but any later one is ok?

joshmoore: This is from their requirements.txt file, due to the version in travis leading to:

> TypeError: numpy boolean subtract, the `-` operator, is deprecated, use the bitwise_xor, the `^` operator, or the logical_xor function instead.

See https://github.com/spacetx/starfish/pull/17

joshmoore:
Re-opening due to:

```
ssh: connect to host github.com port 22: Operation timed out
```

joshmoore:
Re-testing

bgruening:
restarted

recipes/showit/meta.yaml
bgruening: I read it like only <1.13 is supported as the operator is deprecated :(

bgruening:
Green!

recipes/showit/meta.yaml
ambrosejcarr: We only intended to disallow that particular version.

bgruening: @ambrosejcarr numpy 1.13.1 does support the operator again?

bgruening:
Ok, upstream confirmed. Good to go!

PR #10532, opened 2018-08-20
Bump pyBigWig to 0.3.12

dpryan79:

### DIFF
--- a/recipes/pybigwig/meta.yaml
+++ b/recipes/pybigwig/meta.yaml
@@ -19,7 +19,7 @@ requirements:
     - curl
     - libcurl
     - zlib
-    - numpy
+    - {{ pin_compatible('numpy') }}
   run:
     - python
     - libcurl
--- a/recipes/pybigwig/meta.yaml
+++ b/recipes/pybigwig/meta.yaml
@@ -1,15 +1,15 @@
-{% set version = "0.3.11" %}
+{% set version = "0.3.12" %}
 
 package:
   name: pybigwig
   version: {{ version }}
 
 source:
-  url: https://files.pythonhosted.org/packages/7b/b7/e135efa69cd146d22920c196e4575d2ff5a0b19c8f3a29634f5b4c639764/pyBigWig-0.3.11.tar.gz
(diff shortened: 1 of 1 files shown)


recipes/pybigwig/meta.yaml
```diff
@@ -19,7 +19,7 @@ requirements:
     - curl
     - libcurl
     - zlib
-    - numpy
+    - {{ pin_compatible('numpy') }}
```
bgruening: I don't think this is correct. Any reason why you have changed this? 
xref: https://conda-forge.org/docs/meta.html#building-against-numpy

dpryan79: Click over to the circleci error from the previous commit. For whatever reason it appears to have been building against one version (1.9.3 for some reason) and then testing/running against another (1.15.0). That won't work in C unless things are ABI compatible. This change seemed odd, but it got everything to the same version.

dpryan79: https://circleci.com/gh/bioconda/bioconda-recipes/23223?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

bgruening:
Mh ... https://github.com/conda-forge/conda-forge-pinning-feedstock/blame/master/recipe/conda_build_config.yaml#L424 the idea is really to compile against an old version that is ABI compatible and install the latest version.

dpryan79:
It's probably a bug in the header of that old version of numpy (I recall looking for the cause locally). `PyInt_Check()` is only defined in python2, since the size of integers changed in python 3. It appears that this bug was simply fixed in the more recent version that I compiled against.

dpryan79:
Here's the relevant bug in numpy: https://github.com/numpy/numpy/blob/maintenance/1.9.x/numpy/core/include/numpy/ndarrayobject.h#L68

That was fixed in numpy 1.10 (as best I can tell), which appears to have been the first to really support python 3 properly in the headers.

dpryan79:
I should note that I never ran into this before because I didn't use the `PyArray_IsIntegerScalar()` function before, since I didn't have any users using pandas before.

PR #10536, opened 2018-08-20
bcbio_monitor + eden: update depedency of graphviz python package

mbargull:

Follow up to
https://github.com/bioconda/bioconda-recipes/pull/10527
and
https://github.com/conda-forge/staged-recipes/pull/6532#issuecomment-414265097

gh-10527 removed the recipe for `bioconda::py-graphviz` which is the same package as `conda-forge::python-graphviz` so we should use the latter as a dependency.


mbargull:
ty!

PR #10556, opened 2018-08-20
Remove schema-salad from build-fail-blacklist

dfornika:

Remove `schema-salad` from build fail blacklist so it can be updated (#10485)

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1688,7 +1688,6 @@ recipes/qcumber
 
 recipes/saffrontree
 recipes/slclust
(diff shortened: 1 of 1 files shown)


dfornika:
Build seems to be failing because python library `past` isn't available: but is imported (here)[https://github.com/common-workflow-language/schema_salad/blob/6505d0a56a9c6f2c0aaa0b05a29593f9826d6cfc/schema_salad/__init__.py#L35]

dfornika:
Oops, I'll do this directly in #10485

PR #10583, opened 2018-08-21
Compile IQ-TREE in macOS

apcamargo:


apcamargo:
@bioconda/core Should I increase the build number here? The Linux build is exactly the same

bgruening:
Yes, otherwise it will not be built. You could increase the build number also just for OSX if you like.

bgruening:
cool, thanks!

Issue #10100, opened 2018-07-26
osx perl-json-xs (and other packages) install perl-threaded


nathanweeks:
On macOS, perl-json-xs (at least) depends on the deprecated perl-threaded, which is Perl 5.22.0. When perl-threaded is installed, it causes $CONDA_PREFIX/bin/perl to refer to its version of perl, causing problems when other perl modules that depend on the conda-forge 5.26.2 are installed (e.g., perl-perlio-gzip).

Pristine Miniconda3 installation w/ bioconda channels configured per http://bioconda.github.io/:

```
$ conda install perl-json
...
The following NEW packages will be INSTALLED:

    perl:              5.26.2-h16c6ff1_0 conda-forge
    perl-common-sense: 3.74-pl526_1      bioconda
    perl-json:         2.90-pl526_2      bioconda
    perl-json-xs:      2.34-0            bioconda
    perl-threaded:     5.22.0-10         bioconda
...
$ conda install perl-perlio-gzip
...
The following NEW packages will be INSTALLED:

    perl:             5.26.2-h16c6ff1_0    conda-forge
    perl-perlio-gzip: 0.19-pl526h1341992_3 bioconda
```

Installing both:

```
$ conda install perl-json perl-perlio-gzip
The following NEW packages will be INSTALLED:

    perl:              5.26.2-h16c6ff1_0    conda-forge
    perl-common-sense: 3.74-pl526_1         bioconda
    perl-json:         2.90-pl526_2         bioconda
    perl-json-xs:      2.34-0               bioconda
    perl-perlio-gzip:  0.19-pl526h1341992_3 bioconda
    perl-threaded:     5.22.0-10            bioconda
```

$CONDA_PREFIX/bin/perl (perl-threaded) can use the installed perl-json module, but not perl-perlio-gzip. Conversely, $CONDA_PREFIX/bin/perl5.26.2 can use perl-perlio-gzip, but not perl-json:

```
$ perl -v | head -n 2

This is perl 5, version 22, subversion 0 (v5.22.0) built for darwin-thread-multi-2level
$ perl5.26.2 -MPerlIO::gzip -e ''
$ perl -MPerlIO::gzip -e ''
[...]

PR #10596, opened 2018-08-22
QUAST 5.0.0: build number to 1 for forced rebuild

alexeigurevich:

### DIFF
--- a/recipes/quast/meta.yaml
+++ b/recipes/quast/meta.yaml
@@ -7,7 +7,7 @@ package:
   version: {{ version }}
 
 build:
-  number: 0
+  number: 1
   # skip Python 3 builds due to test failures in vendored libs
   skip: True  # [py3k]
 


alexeigurevich:
Hi @bgruening and bioconda team! Could you assist us with this update? Quast now can use Circos to build additional plots, so we added it to the dependency list, however, this causes an issue on macOS which we don't know how to solve properly. You may see in the log:

>conda.exceptions.UnsatisfiableError: The following specifications were found to be in conflict:
> - circos -> perl=5.22.0
> - perl=5.26

Quast requires Perl and by default bioconda assigns Perl 5.26 while Circos requires some Perl modules which are compiled on macOS only using deprecated 5.22.0 (see https://github.com/bioconda/bioconda-recipes/issues/10100 for more details).

Few notes: 
* If we don't specify `circos` in dependencies, Quast recipe passes all the tests and installation conda package is created. However, when a macOS user will try to install Circos after installing Quast 5.0.0, he/she gets an error (`UnsatisfiableError`) or gets a downgrade to Quast 4.6.3.
* If we specify `circos` in dependencies and upload Quast package to our own conda channel, everything seems to be working fine (Quast is associated with Perl 5.22 as Circos). Probably Perl 5.26 is somehow implicitly required by the bioconda channel. 

P.S. Sorry, I've created a new `alexeigurevich-patch-1` branch on `bioconda/bioconda-recipes` -- should create such branch in my own repo next time to prevent messing of the bioconda repo.

alexeigurevich:
It looks like the circos package was rebuild a month ago on both OSX and Linux (see https://anaconda.org/bioconda/circos/files). Do you think a newer rebuild can help? 

Just clarifying: am I right that for rebuilding circos I simply need to create a pull request with the incremented build number in `meta.yaml`?

PR #10613, opened 2018-08-22
Update MAFFT

apcamargo:


recipes/mafft/7.407/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+source:
+  url: https://mafft.cbrc.jp/alignment/software/mafft-{{ version }}-without-extensions-src.tgz
+  sha256: {{ hash }}
+  patches:
+    - osx-makefile.patch # [osx]
+    - linux-makefile.patch # [linux]
+
+build:
+  number: 0
+  skip: False
```
bgruening: This will not produce any package with skip: False. Can you remove this and see how it goes please?

recipes/mafft/7.407/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  commands:
+    - mafft --version
+
+about:
+  home: 'http://mafft.cbrc.jp/alignment/software/'
+  license: BSD
+  summary: Multiple alignment program for amino acid or nucleotide sequences based on fast Fourier transform
+
+extra:
+  identifiers:
```
bgruening: Thanks for adding the DOI!

recipes/mafft/7.407/meta.yaml
apcamargo: Sure!

apcamargo:
@bgruening Is there any reason to maintain one recipe for each version?

bgruening:
Only if the maintainer wants to support older versions as well.

PR #10621, opened 2018-08-22
* Fix command for iqtree and raxml

pvanheus:
* Add bzip2 runtime dependency for blast


pvanheus:
The `-nt` change for the `iqtree` command is proposed upstream at https://github.com/davidemms/OrthoFinder/pull/197 . The other change - to use the AVX2 instead of AVX raxmlHPC depends on what options users have and is not as easy to recommend in a general sense. I have thus not pushed it upstream at this stage.

PR #10633, opened 2018-08-23
add crux-toolkit

jvolkening:

### DIFF
--- a/recipes/crux-toolkit/meta.yaml
+++ b/recipes/crux-toolkit/meta.yaml
@@ -21,7 +21,7 @@ source:
 
 build:
   number: 0
-  script: cp bin/crux $PREFIX/bin
+  script: cp bin/crux $PREFIX/bin/crux
 
 requirements:
   build:
--- a/recipes/crux-toolkit/meta.yaml
+++ b/recipes/crux-toolkit/meta.yaml
@@ -21,7 +21,7 @@ source:
 
 build:
   number: 0
-  script: cp bin/crux $PREFIX/bin/crux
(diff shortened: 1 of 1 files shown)


recipes/crux-toolkit/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+  summary: A cross-platform suite of analysis tools for interpreting protein mass spectrometry data
+  license: Apache-2.0
+  license_family: Apache
+  dev_url: https://github.com/crux-toolkit/crux-toolkit
+
+source:
+  url: https://noble.gs.washington.edu/crux-downloads/crux-{{ version }}/crux-{{ version }}.Linux.i686.zip    # [linux32]
+  url: https://noble.gs.washington.edu/crux-downloads/crux-{{ version }}/crux-{{ version }}.Linux.x86_64.zip  # [linux64]
+  url: https://noble.gs.washington.edu/crux-downloads/crux-{{ version }}/crux-{{ version }}.Darwin.x86_64.zip # [osx]
+  sha256: 0cef62d754f432e78ee4df4e5a0e197617ecd337f68eacc16bddda0a8275a5d8 # [linux32]
```
bgruening: Linux32 is not needed. I assume there is no easy way to compile it?

jvolkening: The source is available -- I will revise it to build from that.

jvolkening: There is some weirdness where the build script expects a git branch/revision but it's not being built from a git repo, which throws an exception. In short, building from the source tarball appears non-trivial without some patching.

Is it worth the effort?

bgruening:
Thanks @jvolkening!

PR #10643, opened 2018-08-23
fix starfish

bgruening:


joshmoore:
Thanks, @bgruening. Pinning to py36 is definitely what's worked for me in the dependency PRs. I'm clearly missing something when it comes to testing these PRs though, having assumed that green would lead to deployment. If there's something else I should be doing, do let me know.

PR #10646, opened 2018-08-23
circos rebuild

bgruening:


druvus:
I have also seen this bug that forced me to split some PR. If we submit packages depending on the other and one is a broken package then the second package that depends on it takes the old build instead of the new build.

druvus:
Fixed circos in #10701 and quast in #10722

PR #10677, opened 2018-08-25
update superfocus

metageni:

I don't think the recipe was using the SUPER-FOCUS version that was refactored (p27 instead). I changed SUPER-FOCUS' setup.py so no need to have `disable_setuptools_scm.patch`. Also, I suspect `post-link.sh` is not needed.

Moreover, I updated it and also included one more test.


corburn:
@metageni the build number should be reset to 0 for a new version.

There have been 8 commits to super-focus between the version 31 recipe and the version 32 recipe all updating either setup.py or the readme.

Did you observe a package downgrade?:

```
$ conda install super-focus=31
$ superfocus --help
raise VersionConflict(dist, req).with_context(dependent_req)
pkg_resources.ContextualVersionConflict: (numpy 1.11.3 (/home/jtravis/miniconda3/envs/sf/lib/python3.7/site-packages), Requirement.parse('numpy>=1.12.1'), {'superfocus'})

$ conda install numpy=1.12.1
The following packages will be DOWNGRADED:

    python:      3.7.0-h5001a0f_1      conda-forge --> 2.7.15-h9fef7bc_0                    conda-forge
    super-focus: 0.31-py37_0           bioconda    --> 0.30-py27_0                          bioconda 
```

I am wondering if adding `noarch: python` would fix the version 31 recipe.

metageni:
@corburn I have reset the build to 0. I don't remember creating any recipe for 0.31. Maybe that is why it is happening?

corburn:
@metageni  in regards to your [question](https://github.com/bioconda/bioconda-recipes/commit/4d0981245051faa5ca2295536e1ff1b3cd166edf#commitcomment-30330800) why is it still downgrading to super-focus: 0.30-py27_0, this is what I found:

I was able to install super-focus=0.32 on Linux, but not on OSX. The reason is because `rapsearch` is not available on OSX. Adding `noarch: python` resolved the numpy dependency conflict in the super-focus=0.31 recipe, but introduced a `package not available` conflict when installing on OSX.

The following is from the conda [define-metadata](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html) documentation:
>Warning
>
>At the time of this writing, noarch packages should not make use of preprocess-selectors: noarch packages are built with the directives which evaluate to True in the platform it was built, which probably will result in incorrect/incomplete installation in other platforms.

A solution could be to remove the `rapsearch` dependency and bump the build number. I would also do the same for the 0.31 recipe adding `noarch: python` as well so it's not broken. Users could still opt to install rapsearch and you could prompt them to do so with a post-link.sh script.

PR #10683, opened 2018-08-26
update focus to download database

metageni:

Bioconda was not copying FOCUS database. This version is supposed to copy the database and uncompress it for the user.


recipes/focus/meta.yaml
```diff
@@ -11,7 +11,7 @@ source:
 
 build:
   noarch: python
-  number: 0
+  number: 1
```
bgruening: This needs to be 0 if you have a new version.

bgruening:
Thanks!

PR #10705, opened 2018-08-27
change jellyfish on focus

metageni:

Sorry @bgruening. I believe this may be the last update until FOCUS a new release comes up. The current recipe works fine, but when i try to create a conda enviroment (e.g. `conda create -n focus_run -c bioconda focus`), Jellyfish `2.2.6` is giving me a `Segmentation fault: 11` error when counting k-mers on the enviroment I created (outside of the enviroment it works fine). So I tested it with Jellyfish `2.2.3`, and it worked fine.

### DIFF
--- a/recipes/focus/meta.yaml
+++ b/recipes/focus/meta.yaml
@@ -23,7 +23,7 @@ requirements:
     - python >=3
     - numpy >=1.12.1
     - scipy >=0.19.0
-    - jellyfish
+    - jellyfish ==2.2.3
     - unzip
 
 test:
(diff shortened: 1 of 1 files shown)


bgruening:
Mh, did jellyfish changed their API/ABI? Can you report the error upstream?
To create a new version of the pkg, please increaes the build number. Also a test that triggers the failure would be good, so we make sure this does not happen again. 

Thanks!

metageni:
I just double checked, and the version of jellyfish which is giving the problem is `2.2.10`. I just tested `2.2.6`, and it worked fine. I will change the code.

Also, the jellyfish team have not answered any issue-question for the past month. I will create an issue for the problem, and if what comes up.

PR #10707, opened 2018-08-27
Try building phyml against openmpi 3

epruesse:
openmpi 2 is no longer available on bioconda or conda-forge


recipes/phyml/meta.yaml
```diff
@@ -16,18 +16,18 @@ source:
...
+  number: 0
 
 requirements:
   build:
     - {{ compiler('c') }}
     - autoconf
     - automake
     - pkg-config
-    - openmpi 2.*
+    - openmpi 3.*
```
bgruening: If you want to build against latest openmpi, just remove it. You can also remove the run-time deps I think as openmpi has run_exports defined: https://github.com/conda-forge/openmpi-feedstock/blob/master/recipe/meta.yaml#L18

PR #10720, opened 2018-08-28
MentaLiST: Remove old versions and pin dependencies in current versions

dfornika:


recipes/mentalist/0.1.9/meta.yaml
```diff
@@ -10,19 +10,19 @@ source:
...
+  number: 3
   skip: True  # [osx]
 
 requirements:
   host:
-    - julia
-    - hdf5
-    - mpfr
+    - julia 0.5.2
+    - hdf5 1.8.17
```
bgruening: This pinning is not needed anymore if you can use the global pinning that are in place. E.g this one: https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml#L343

If you think julia needs to be pinned, we should add julia as well to the above linked file.

dfornika: Sorry, I feel like I'm missing some understanding of how pinning is supposed to work under conda-build 3.

We have two versions of MentaLiST (a 0.1 series and a 0.2 series). The 0.1 series is written against `julia-0.5.2` and the 0.2 series is written against `julia-0.6.1`. So I need to ensure that the correct version of Julia is installed for each version of MentaLiST.

bgruening: Yes, but hdf5 is for example globally pinned. Julia 1.0 should probably be added as a global pin as well.

dfornika: Ah, I'm beginning to understand. Sorry, I feel like I've been a bit dense while adjusting to some of the new cb3 stuff. The docs are good, but it's taken me some time to absorb them.

dfornika:
Would it be possible to merge this despite the pinned `hdf5` & `mpfr` versions? The `julia-feedstock` repo is a bit of a complicated beast, so I'm not sure how quickly we can adopt global pinning for its dependencies (or if it's even possible).

I'd love to get to the point where MentaLiST is compatible with Julia 1.0, but it's an academic research project and we don't have a ton of resources for updating. I'm also not sure if its dependencies are Julia-1.0 compatible yet.

dfornika:
@bgruening @druvus Could I ask for some feedback on this pull request? We have downstream projects that depend on this package working and it's been broken for a  while now.

I'm not certain that I fully understand the global pinning situation, so if there's a better solution that we could implement without major changes to the julia conda-forge repo or to MentaLiST then I'd be open to accomodating that. But as I see it, this is the only way that I know of to get this package working.

dfornika:
Could I please get some feedback on this pull-request? It's been passing the test suite for some time now. I'm willing to work towards alternate solutions if the current pull-request isn't acceptable but I'm not sure what my viable options are.

@bgruening @druvus @johanneskoester @corburn Is there some sort of meta-user that I can tag to get attention from whichever maintainer is available to help? Otherwise I'm sort of tagging people in at random based on the most active contributors that I see on the repo.

bgruening:
@dfornika sorry that I missed the last ping. Try to catch me on gitter as well next time.

Please consider to add julia as a major pin to all recipes in the future.

PR #10725, opened 2018-08-28
Rebuild feelnc

cmonjeau:

### DIFF
--- a/recipes/feelnc/meta.yaml
+++ b/recipes/feelnc/meta.yaml
@@ -15,7 +15,6 @@ build:
 
 requirements:
   build:
-    - perl ==5.22.0.1
     - perl-bioperl
     - perl-db-file
     - perl-bio-featureio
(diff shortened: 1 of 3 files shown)


druvus:
Bioconda has moved to build against perl 5.26 so I think it would be better to fix dependencies that are not working. What issue did you have with perl 5.26 and feelnc?

recipes/feelnc/meta.yaml
```diff
@@ -21,6 +21,7 @@ requirements:
     - perl-bio-featureio
 
   run:
+    - perl ==5.22.0.1
```
abretaud: my dear @cmonjeau, can you left it as it was before? I think just rebuilding should be enough
BUT you'll need to remove feelnc from the blacklist https://github.com/bioconda/bioconda-recipes/blob/master/build-fail-blacklist#L1964

recipes/feelnc/meta.yaml
```diff
@@ -15,7 +15,6 @@ build:
 
 requirements:
   build:
-    - perl ==5.22.0.1
```
bgruening: perl should be given if perl is needed. This is a special case afaik for all scripting languages. Some for R or python.

abretaud: ok, good to know, I wasn't 100% sure

PR #10735, opened 2018-08-28
add smeg

aemiol:

### DIFF
--- a/recipes/smeg/build.sh
+++ b/recipes/smeg/build.sh
@@ -16,3 +16,4 @@ cp smeg $outdir
 chmod +x $outdir/smeg
 
 ln -s $outdir/smeg $PREFIX/bin/smeg
+cp $outdir/smeg $PREFIX/bin/.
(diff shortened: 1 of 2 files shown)


micknudsen:
It would be helpful if you modified the PR title to include the package name. If all PR were simple called "add new recipe", navigating PRs would be nearly impossible.

aemiol:
@micknudsen modified PR title

martin-mann:
something I noted (which might not solve the problem): you are explicitely requesting gcc.. needed?

If so, better replace with the recent `{{ compiler('c') }}` directive, which provides the system specific compiler.. afaik, `gcc` is not recommended anymore..

best,
Martin

aemiol:
@martin-mann  still getting same build time error

martin-mann:
jup I've seen.. but no idea why your build.sh goes ot hell... looks like it is eventually just copying things..

reping @bgruening any idea?

martin-mann:
maybe just start uncommenting your build.sh from back to front to see if any specific call/setup in there causes the timeout

PR #10742, opened 2018-08-29
Add pyega3

AlexanderVi:
* [ *] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [ *] This PR adds a new recipe.
* [ *] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).

### DIFF
--- a/recipes/pyega3/meta.yaml
+++ b/recipes/pyega3/meta.yaml
@@ -15,14 +15,12 @@ build:
   entry_points:
     - pyega3 = pyega3.pyega3:main
   script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv "
+  skip: True  # [not py3k]
 
 requirements:
   host:
     - pip
-    - pip
     - python
-    - requests
-    - tqdm
   run:
     - python
     - requests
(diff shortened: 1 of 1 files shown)


recipes/pyega3/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+build:
+  noarch: python
+  number: 0
+  entry_points:
+    - pyega3 = pyega3.pyega3:main
+  script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv "
+
+requirements:
+  host:
+    - pip
```
bgruening: only once :)

recipes/pyega3/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - pip
+    - python
+    - requests
```
bgruening: I don't think requests and tqdm are needed in the host env, isn't it?

AlexanderVi:
Thanks for your comments,Björn! I've fixed the host section. Also I had to remove 'noarch' (otherwise tests fail).
PS. Do you know ,by chance, is there any way to specify not exact version number, but some kind of wildcard saying "take the latest available on PyPi repo" ?

johanneskoester:
The explicit version is necessary, because it might be that dependencies change in future versions. We do not regularly rebuild the packages, but rather rely on PRs for version updates. However, we actually plan an update bot that tracks pypi, so that recipes are updated automatically for new versions.

PR #10747, opened 2018-08-29
R seurat scripts

pinin4fjords:

This PR adds a set of scripts for calling individual steps of the Seurat workflow (and one for testing the whole sequence). See summary in meta.yaml for details.


recipes/r-seurat-scripts/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+  url: https://github.com/ebi-gene-expression-group/r-seurat-scripts/archive/v{{ version }}.tar.gz
+  sha256: 580770b74194b41656e39af542af01250326f41a311038249b84dede8a57f76c
+
+build:
+  number: 0
+  noarch: generic
+
+requirements:
+    run:
+        - r-seurat>=2.3.1
```
bgruening: please add r-base as well.

recipes/r-seurat-scripts/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+        - r-optparse
+        - r-workflowscriptscommon>=0.0.1
```
bgruening: the pinning should not be needed

recipes/r-seurat-scripts/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+test:
+    commands:
+        - which seurat-read-10x.R
```
bgruening: A `seurat-read-10x.R --help` or `--version` would be nice. As this would also check the R imports.

pinin4fjords:
Thanks @bgruening, requested changes added.

recipes/r-seurat-scripts/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+        - which seurat-find-variable-genes.R
+        - which seurat-get-random-genes.R
+        - which seurat-run-pca.R
+        - which seurat-scale-data.R
+        - which seurat-dim-plot.R
+        - which seurat-find-clusters.R
+        - which seurat-run-tsne.R
+        - which seurat-find-markers.R
+        - which r-seurat-scripts-post-install-tests.sh
+        - seurat-read-10x.R --help
```
bgruening: Do you want to add this `--help` for the other tools as well? If help works, you don't need to `which` :)

pinin4fjords: Since all the scripts load the same libraries I figured a single help would do the job of checking those, but I can change everything to helps if you think that's better- I guess it is neater.

bgruening: It's also more future-proof. Assuming you add new libraries in the future to single files.

pcm32:
Thanks @pinin4fjords @bgruening! 

One question  @bgruening, is there a delay in the new container being generated? I cannot see https://quay.io/repository/biocontainers/r-seurat-scripts 

Or is it maybe that it is getting created as private initially?

Thanks!

bgruening:
> Or is it maybe that it is getting created as private initially?

exactly :( check again :)

pcm32:
mmmm... still unable to see it (https://quay.io/repository/biocontainers/r-seurat-scripts says "repository not found"). It doesn't show up in the search either. It is probably still private @bgruening :-(.

bgruening:
@pcm32 not sure what I did yesterday, but it seems it missed it. Sorry. Now it is.

PR #10759, opened 2018-08-30
Lordec 0.9

abretaud:

Added a recipe for gatb and prevent lordec from compiling its own boost and gatb versions


abretaud:
@bioconda/core do you see why the linux test is failing for lordec? It works fine on my laptop, and I see nothing wrong in the log

martin-mann:
seems your `-lrt` linker flag fails since the RT library is not found... Maybe the `compiler-rt` package will help, I will add and give it a try..

abretaud:
ah thanks @martin-mann, I missed this lib!

martin-mann:
seems this package doesnt do the trick.. :cry: maybe `-lrt` is not needed for osx?!

martin-mann:
@abretaud you might want to change your makefile.patch to something like this
```
ifeq ($(UNAME_S),Darwin)
LDFLAGS = -L$(GATB)/lib/ -lgatbcore -lhdf5$(HDFSUFFIX) -ldl -lz -lpthread -std=c++0x -O3 $(EXTRALDFLAGS)
else
# on NON-MacOS, we need to add RT library
LDFLAGS = -L$(GATB)/lib/ -lgatbcore -lhdf5$(HDFSUFFIX) -lrt -ldl -lz -lpthread -std=c++0x -O3 $(EXTRALDFLAGS)
endif
```
hope this helps

final solution: skip OSX build for this package....

abretaud:
and it... works! thanks @martin-mann!

PR #10769, opened 2018-08-31
wisecondorX: bumped subversion

matthdsm:

### DIFF
--- a/recipes/wisecondorX/meta.yaml
+++ b/recipes/wisecondorX/meta.yaml
@@ -1,11 +1,11 @@
-{% set version="0.2.0" %}
+{% set version="0.2.1" %}
 package:
   name: wisecondorx
   version:  {{ version }}
 
 source:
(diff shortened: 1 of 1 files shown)


recipes/wisecondorX/meta.yaml
```diff
@@ -1,11 +1,11 @@
...
   name: wisecondorx
   version:  {{ version }}
 
 source:
   url: https://github.com/CenterForMedicalGeneticsGhent/wisecondorX/archive/v{{ version }}.tar.gz
-  sha256: a03133935bf78234009a62aa9349d7cf67b192d6a611bb9c202a329d00addbc5
+  sha256: 048f95292d6854c6ac2335333ff9ac7d0cf86e2bdee12647e9a4642aa21046c4
 
 build:
   number: 1
```
corburn: The build number should be reset to zero after a version bump and the `skip: true` could be replaced with `noarch: python`.

matthdsm: Hi Jason
Thanks for catching this. I missed the build number. 

Cheers
M

PR #10791, opened 2018-09-03
Rebuild roary

druvus:

Rebuilding roary for perl 5.26 by restricting parallel and bioperl to newer version not pulling in perl-threaded.

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1972,7 +1972,6 @@ recipes/perl-hpc-runner-command-plugin-logger-sqlite/0.0.3
 
 # real long build-time
 recipes/perl-bio-viennangs
-recipes/roary
 
 #   - perl-bio-cigar -> perl=5.22.0
 recipes/rad_haplotyper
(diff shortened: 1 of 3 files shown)


druvus:
Build takes still very long time. Locally it seems that build get stuck at following test  `t/Bio/Roary/CommandLine/Roary.t ......................... 5/?`

tseemann:
It just says `Build timedout`

druvus:
yes, circleci just show timeout but running the build locally in docker shows that it hangs at `t/Bio/Roary/CommandLine/Roary.t`. My guess it is that a used conda package is broken but my perl knowledge is a little bit rusty.

aemiol:
My recipe has Roary as a dependency and I am having similar "build timeout" during test. Initially, I couldn't trace the source but from this thread, it appears roary might be the culprit.

tseemann:
You should be able to disable tests some how with an option in `Build.pl` maybe?
@andrewjpage

druvus:
Roary was rebuilt in #11017

druvus:
@tseemann and @aemiol Roary is now rebuilt. Please check if the issues remains.

tseemann:
Thanks @druvus  - I'll inform the people having trouble.

PR #10796, opened 2018-09-04
viennarna 2.4.9

martin-mann:

### DIFF
--- a/recipes/viennarna/meta.yaml
+++ b/recipes/viennarna/meta.yaml
@@ -1,5 +1,5 @@
 {% set version = "2.4.9" %}
-{% set sha256 = "268c4adef272a960ab178f2e05d57c10f317de2b214960b3ac7d5831b759a213" %}
+{% set sha256 = "a95cbcab730449a5df3edcc2b71d5126b436ac84ac9c0efa53a91896af142a1f" %}
 
 package:
   name: viennarna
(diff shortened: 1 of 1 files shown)


martin-mann:
want to ensure that pkgconfig lib information is present/installed for downstream configure checks (intarna). we got intarna travis build issues without enforcing pkgconfig to be installed, so i put it here.

no good?

martin-mann:
i guess pkgconfig came so far with some other dependency but havent tested/checked yet

s-will:
did it fail without pkgconfig? If you want to use pkgconfig downstream, there should be no need to specify it as build requirement here. Pkgconfig would then be needed only when building the downstream packages.

martin-mann:
not sure, since afaik pkgconfig is used to install the according subfolder with the lib information... !?!

martin-mann:
the vrna build most likely does not fail without pkgconfig but not sure whether or not the needed lib information is available if build without...

s-will:
nope, one does not need pkgconfig to build a package that defines its dependencies via pkgconfig (like the vienna rna package). Concretely, this works based on the file
RNAlib2.pc.in
which is translated by the usual autotools mechanism to some RNAlib2.pc, which is then installed to the correct directory. All of this works without having pkgconfig installed

martin-mann:
ok, thanks for this... will remove the pkgconfig dependency

martin-mann:
macos build pending anyway...

PR #10805, opened 2018-09-04
PacBio: Add disclaimer

pb-dseifert:

### DIFF
--- a/recipes/pbccs/meta.yaml
+++ b/recipes/pbccs/meta.yaml
@@ -7,12 +7,14 @@ package:
   version: {{ version }}
 
 source:
-  url: https://github.com/PacificBiosciences/unanimity/releases/download/v{{ version }}/ccs # [linux]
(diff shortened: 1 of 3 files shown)


pb-dseifert:
@bgruening any ideas what's going on here? The linter error is totally bizarre, given that the SHA256 is still there and untouched.

recipes/pbccs/meta.yaml
```diff
@@ -9,10 +9,12 @@ package:
 source:
   url: https://github.com/PacificBiosciences/unanimity/releases/download/v{{ version }}/ccs # [linux]
```
bgruening: Just remove the `# [linux]` here and the lines below. You are skipping OSX below.

pb-dseifert: Thanks, that did it. Is this a change in behaviour from CB2?

PR #10819, opened 2018-09-05
Reparation_blast recipe

RickGelhausen:

### DIFF
--- a/recipes/reparation_blast/meta.yaml
+++ b/recipes/reparation_blast/meta.yaml
@@ -28,9 +28,10 @@ requirements:
     - r-prroc
     - r-rocr
     - r-randomforest
+    - r-sizer
     - prodigal
     - glimmer
-    - blast >2.7.1
+    - blast >=2.7.1
 
 test:
   commands:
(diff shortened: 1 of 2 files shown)


recipes/reparation_blast/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - r-rocr
+    - r-randomforest
+    - r-sizer
+    - prodigal
+    - glimmer
+    - blast >=2.7.1
+
+test:
+  commands:
+    - reparation.pl || true
```
peterjc: Doesn't this ``|| true`` mean the command will always return success - making it useless as a test?

If the tool doesn't have a ``--version`` or ``--help`` option or similar, can you use ``grep`` on the output at least to confirm the script is running and producing a meaningful output?

RickGelhausen: Hi, thanks for the tip. 
You are right, this was a work-around for having no `--version` or `--help` option.
I have added a version option now.

peterjc: That's a better solution, and will be useful outside of BioConda too 👍

PR #10863, opened 2018-09-07
Add package for scanpy.

roryk:

This adds scanpy (https://github.com/theislab/scanpy) a single-cell analysis package in python. scanpy uses conda packages in its installation instructions but didn't have a recipe yet. Now it will.

### DIFF
--- a/recipes/scanpy/meta.yaml
+++ b/recipes/scanpy/meta.yaml
@@ -15,20 +15,6 @@ requirements:
   host:
     - python
     - setuptools
-    - seaborn
-    - 'scikit-learn>=0.19.1'
-    - statsmodels
-    - numba
-    - python-igraph
-    - louvain
-    - 'anndata >=0.6.10'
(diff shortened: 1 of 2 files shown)


bgruening:
I was starting to write one just now, good that I checked my inbox before :)

Thanks @roryk!

recipes/scanpy/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  skip: True # [not py36]
+source:
+  url: https://github.com/theislab/scanpy/archive/{{ version }}.tar.gz
+  sha256: ae29eb6e0998cb8eb613f72800a97265362c95bbe3cd4019d4127e006eef6453
+
+requirements:
+  host:
+    - python
+    - setuptools
+    - seaborn
```
bgruening: I guess most of these dependencies, besides setuptools and python, can be removed from the host section, isn't it?

Issue #10671, opened 2018-08-25
pyBigWig dependency update


joachimwolff:
pyBigWig lists as dependency numpy version 1.12, however, numpy version 1.15 is the newest one and 1.14 is missing too. 
Furthermore, a version for Python 3.7 is missing. Is there any way to enforce an update to the mentioned versions (assuming nothing will break and would need a fix of pybigwig from @dpryan79)?

Thanks a lot.

dpryan79:
pyBigWig doesn't depend on any particular version of numpy, it's just that when you build it you need to then not upgrade to a non-ABI compatible library. bioconda and conda-forge build against a couple different versions of numpy and nothing else, so that's what pyBigWig and other packages that need pinning get pinned against.

I don't think anything in bioconda is building against python 3.7 yet. It's in the conda-forge-pinning repository, but requires an environment variable be set appropriately to be used. Presumably this change is then waiting for everything in conda-forge to be built, one would need to ask /core.

bgruening:
> pyBigWig lists as dependency numpy version 1.12

Can you explain this a little bit more? Where is this listed?

`conda create -n pbw pybigwig` gives me numpy 1.14 and I can get 1.15 with `conda create -n pbw pybigwig numpy=1.15`

joachimwolff:
Yesterday I did:
`conda install --file requirements.txt -c bioconda -c conda-forge`

and got 
```
UnsatisfiableError: The following specifications were found to be in conflict:
  - numpy  1.14.*
  - pybigwig  0.3.* -> numpy 1.12*
```
Requirements.txt:
```
numpy = 1.14.*
scipy = 1.1.*
matplotlib = 2.2*
pysam = 0.14.*
intervaltree = 2.1.*
biopython = 1.72.*
pytables = 3.4.*
pandas = 0.23.*
pybigwig = 0.3.*
cooler = 0.7.10
jinja2 = 2.10
six = 1.11.*
future = 0.16.*
unidecode = 1.0.*
hic2cool = 0.4.*
scikit-learn = 0.19.*
```

Today I am doing the same, and there is no conflict and it can be installed. I am a bit confused about what is going on.

dpryan79:
Sounds like the universe is trying to remind you what weekends are :)

dpryan79:
@joachimwolff Can this be closed or is there still a problem?

joachimwolff:
Hi, 

[...]

PR #10876, opened 2018-09-09
pin_compatible does not pin API compatible versions of numpy. Pin to …

dpryan79:
…the version used for building to avoid problems.

@joachimwolff This should resolve numpy version problems in the future, since `pin_compatible()` isn't working properly.

### DIFF
--- a/recipes/pybigwig/meta.yaml
+++ b/recipes/pybigwig/meta.yaml
@@ -19,12 +19,12 @@ requirements:
     - curl
     - libcurl
     - zlib
-    - numpy {{ numpy }}
+    - numpy 1.15.*
   run:
     - python
     - libcurl
     - zlib
-    - numpy {{ numpy }}
+    - numpy 1.15.*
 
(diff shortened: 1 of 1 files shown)


johanneskoester:
@dpryan79 if there is a bug in pin_compatible, did you report it upstream? I am asking because this kind of workaround won't scale if we do a bulk update of numpy.

dpryan79:
@johanneskoester I haven't reported this upstream yet, this first came up in #10671 . I should add that this turns out to be mentioned in the conda build documentation, so it's not a bug in `pin_compatible()`, it's more of a question of whether the current minimum build version is appropriate. I'll raise this is conda-forge-pinning, in case it hasn't already been discussed there.

I agree that the solution in this PR is less than ideal. In my opinion, the versions from `pin_compatible()` should at least be ABI compatible, in which case whoever updates the numpy versions used by conda-forge should check things and update as appropriate.

PR #10883, opened 2018-09-10
add fastqz

MathiasHaudgaard:


MathiasHaudgaard:
@bgruening I'm getting an error in the test. I've tried several approaches but none of them seem to work. I get the error message: .Passing program arguments as a string may be a security hazard if combined with untrusted input.

micknudsen:
@MathiasHaudgaard Are you sure the test fails due to "Passing program arguments as a string may be a security hazard if combined with untrusted input"? It looks a like a warning (and not something that would fail). Have you tried printing the output of your test commands and see if it really is as expected?

MathiasHaudgaard:
@micknudsen yes I have. It seems like when i'm using _2>&1 | grep_  the error occurs. I'm not sure what causes this problem.

MathiasHaudgaard:
@bgruening maybe it's a bit of a hack, but i'm downloading libzpaq with wget and the use it for compilation. You can see it in build.sh
@micknudsen the problem is that the -h flag returns 1 which causes the error. That's why i'm using **2>&1**. It used to work but i've experienced this error several times now.

daler:

PR #10895, opened 2018-09-11
Update roary to bioperl 1.7

tseemann:

### DIFF
--- a/recipes/roary/meta.yaml
+++ b/recipes/roary/meta.yaml
@@ -7,7 +7,7 @@ package:
   version: {{ version }}
 
 build:
-  number: 1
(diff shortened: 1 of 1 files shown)


druvus:
@tseemann roary is listed in the build-fail-blacklist so it need to removed from it to be built and tested. The linting should be updated to show this issue. 

The build is getting stuck at a test for me both locally and on circleci. Maybe you could understand why it is not working? xref #10791

tseemann:
How does testing succeed if it is in a blacklist?
Anyway, that explains why it built so fast.
Thanks - i'll try and have a look.

PR #10902, opened 2018-09-11
added basic epa-ng recipe

gavinmdouglas:


recipes/epa-ng/meta.yaml
```diff
@@ -0,0 +1,36 @@
+package:
+  name: epa-ng
+  version: "0.3.3"
+
+build:
+  number: 1
```
epruesse: new recipe, should be 0

recipes/epa-ng/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+  build:
+    - {{ compiler('cxx') }}
+    - libtool
+    - flex
+    - bison
+    - m4
+    - cmake==3.11
+    - automake
+    - autoconf
+  run:
```
epruesse: zlib should be a host dependency. The run dependency will be injected automatically.

recipes/epa-ng/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - zlib
+
+test:
+  commands:
+    - epa-ng -h > /dev/null
+
+about:
+  home: https://github.com/Pbdas/epa-ng
+  license: GNU Affero General Public License v3.0
```
epruesse: GPL requires that the license be distributed with the packages. Please add `license_file:`

recipes/epa-ng/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+  summary: Massively parallel phylogenetic placement of genetic sequences
+
+extra:
+  identifiers:
+    - https://doi.org/10.1093/sysbio/syy054
```
epruesse: should be `- doi:10.1093/sysbio/syy054`

recipes/epa-ng/meta.yaml
```diff
@@ -0,0 +1,36 @@
+package:
+  name: epa-ng
+  version: "0.3.3"
+
+build:
+  number: 1
+
+source:
+  url: https://github.com/Pbdas/epa-ng/archive/v0.3.3.tar.gz
+  md5: 881d024877f3cfa632d344373e119f9d
```
epruesse: we prefer sha256.

recipes/epa-ng/conda_build_config.yaml
```diff
@@ -0,0 +1,6 @@
+c_compiler:
+  - gcc      # [linux]
+  - clang    # [osx]
+cxx_compiler:
+  - gxx      # [linux]
+  - clangxx  # [osx]
```
epruesse: I presume it doesn't work without this file?

recipes/epa-ng/meta.yaml
epruesse: The build/host/run is a little confusing. It's for cross compiling. Everything that is build machinery, so compiler and build tools go into `build`. Everything that is dependent on the build target, so libraries you need to link against, goes into `host`. Things that will be needed at run time go in to `run`. Some packages have what conda calls `run exports`. Those will add things to your `run` automatically. Zlib is one of them. The `run: - zlib` is therefore redundant.

gavinmdouglas: Yes I was confused by this; however, without adding zlib as a run requirement this error is causes: ```SERR epa-ng: error while loading shared libraries: libz.so.1: cannot open shared object file: No such file or directory```, which you can see here: https://circleci.com/gh/bioconda/bioconda-recipes/38521?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification

recipes/epa-ng/meta.yaml
gavinmdouglas: Changed.

recipes/epa-ng/conda_build_config.yaml
gavinmdouglas: Yes this resulted from extensive troubleshooting, which you can read about here: https://forum.qiime2.org/t/issue-with-cmake-when-making-epa-ng-conda-recipe/6430/9

recipes/epa-ng/meta.yaml
epruesse: Yes. It needs to be distributed with the binaries as well. By specifying `license_file: LICENSE` conda-build knows which file to add to the package (ends up in the `info` folder).

recipes/epa-ng/meta.yaml
epruesse: It needs to be in `host`.

epruesse: I added that - should work.

recipes/epa-ng/meta.yaml
gavinmdouglas: Ok thanks - I added it

gavinmdouglas:
Hi @epruesse, it looks like all of the minor edits have been addressed.

gavinmdouglas:
Thanks for your help @epruesse.

PR #10914, opened 2018-09-12
MEME: update to 5.0.2 + compile with mpi support

abretaud:

### DIFF
--- a/recipes/meme/meta.yaml
+++ b/recipes/meme/meta.yaml
@@ -73,3 +73,8 @@ about:
   license: Custom
   license_file: COPYING
   summary: Motif based sequence Analysis tools
+
+extra:
+  container:
+    # openmpi needs ssh/rsh
+    extended-base: true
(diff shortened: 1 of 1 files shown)


james-chuang:
In this update, many of the "support" tools (e.g. `fasta-get-markov` and `meme2meme`) from the MEME Suite are not added to `$PATH` by conda. I think this is because the executables for these scripts are now in `/libexec`, whereas they were in `/bin` in previous versions (4.x.x) of MEME Suite. My understanding is that `/libexec` is _supposed_ to be for executables called by other programs and not by the end user ([https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html)), hence why it's not added to `$PATH`. However, many of these clearly should be in `$PATH` as they were in previous versions of MEME. Is there some way to resolve this?

abretaud:
Thanks for reporting the problem @james-chuang, I've made #11379 which should fix the problem

james-chuang:
Great, thanks for the quick fix!

PR #10924, opened 2018-09-12
change channel priority

bgruening:

xref: https://github.com/bioconda/bioconda-utils/pull/334

### DIFF
--- a/.circleci/setup.sh
+++ b/.circleci/setup.sh
@@ -51,8 +51,8 @@ if ! type bioconda-utils > /dev/null || [[ $BOOTSTRAP == "true" ]]; then
 
     # step 2: setup channels
     $WORKSPACE/miniconda/bin/conda config --system --add channels defaults
-    $WORKSPACE/miniconda/bin/conda config --system --add channels conda-forge
     $WORKSPACE/miniconda/bin/conda config --system --add channels bioconda
+    $WORKSPACE/miniconda/bin/conda config --system --add channels conda-forge
 
     # step 3: install bioconda-utils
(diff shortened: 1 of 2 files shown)


mbargull:
cc @bioconda/all for awareness:
We just changed the recommended channel priority order of `bioconda` and `conda-forge` (giving the latter higher priority) in the docs as well as using that order in our build system. (Discussion at https://github.com/bioconda/bioconda-utils/issues/330.)

Please note the following `bioconda-utils` release notes as well as the changed documentation:

From https://github.com/bioconda/bioconda-utils/releases/tag/v0.14.5:
> This version changes the channel order such that conda-forge has highest priority. The rationale is that Bioconda shall provide bio-specific packages that are non general-purpose, hence it complements conda-forge in this area. By giving Bioconda a lower priority than conda-forge, non-bio packages that occur in both channels for historical reasons will be taken from conda-forge without having to remove them from bioconda, which would break reproducibility. A side effect of this strategy is that for some reason this order speeds up the dependency solver in certain cases.

From http://bioconda.github.io/#set-up-channels:
> ### 2. Set up channels
> 
> After installing conda you will need to add the bioconda channel as well as the other channels bioconda depends on. **It is important to add them in this order** so that the priority is set correctly (that is, `conda-forge` is highest priority).
> 
> The `conda-forge` channel contains many general-purpose packages not already found in the `defaults channel`.
> ```
> conda config --add channels defaults
> conda config --add channels bioconda
> conda config --add channels conda-forge
> ```

janxkoci:
What if I'm existing user? Should I re-add the channels in new order in all my installations?

PertuyF:
You should manage the order using `.condarc`, either master one in your base environment, or a user one in your $HOME, or one for each environment.
Priority is base < user < env, from lowest to highest.

PR #10931, opened 2018-09-13
Rebuild perl-inline-c

abretaud:

Just triggering a rebuild of perl-inline-c as the osx package is broken (installed in perl 5.22 lib dir, but depends on 5.26)
Needed by #10907

### DIFF
--- a/recipes/perl-inline-c/meta.yaml
+++ b/recipes/perl-inline-c/meta.yaml
@@ -30,7 +30,7 @@ requirements:
 
 test:
   imports:
-    - Inline::C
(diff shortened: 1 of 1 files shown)


abretaud:
Ah cool, thanks! Can you ping me when it's ready?

druvus:
Added PR #10935

The issue seems to have been fixed in newest version of Inline-C

MathiasHaudgaard:
This package messes with my builds. Somehow the linter tries to lint this package in all of my builds.

PR #10939, opened 2018-09-13
Bump  perl-inline-c

druvus:

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -2040,4 +2040,3 @@ recipes/perl-moosex-app/1.39
 
 ####################### OSX errors above ###############
 
-recipes/perl-inline-c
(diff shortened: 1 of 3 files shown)


abretaud:
Maybe we could just ignore the `t/08taint.t` tests ? it looks like they are kind of broken: https://www.nntp.perl.org/group/perl.inline/2014/06/msg4639.html

bgruening:
Mh ... ` /bin/sh: gcc: not found` it needs gcc as run-time dependency?

abretaud:
Yes, it's a module to compile C code from perl code (https://media.giphy.com/media/3osxYbFwfSPZ8k76cU/giphy.gif)
And adding gcc to the run time deps does not help as it requires a newer version (>4.9 I guess)

druvus:
thanks @abretaud  for fixing it!!

PR #10947, opened 2018-09-13
Add smeg

aemiol:

### DIFF
--- a/.circleci/setup.sh
+++ b/.circleci/setup.sh
@@ -59,7 +59,8 @@ if ! type bioconda-utils > /dev/null || [[ $BOOTSTRAP == "true" ]]; then
     $WORKSPACE/miniconda/bin/pip install git+https://github.com/bioconda/bioconda-utils.git@$BIOCONDA_UTILS_TAG
 
     # step 4: configure local channel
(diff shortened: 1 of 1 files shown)


mbargull:
I cancelled the build since it's taking way too long. Please check locally which test command is taking so long to finish. Also, please note that it is highly preferred that the test commands in `meta.yaml` check as much as possible but in as few time as possible, too. Ideally, those tests should be finished in seconds, and at most they should take few minutes.

mbargull:
I just now saw that this has a few R dependencies, which could also mean that `conda` is just slow in creating the environment here. If that is the case, then maybe after we implement https://github.com/bioconda/bioconda-utils/pull/334 thing will be faster.

mbargull:
> Does the conda-forge channel now have a higher priority than bioconda?

Not quite yet, but soon. We need to (in that order)
1. merge https://github.com/bioconda/bioconda-utils/pull/334
2. tag a new `bioconda-utils` release
3. merge https://github.com/bioconda/bioconda-recipes/pull/10924

Your PR is currently blocked by that process, sorry for the inconvenience.

aemiol:
Ok. Thanks

aemiol:
@mbargull I traced the problem to a single package (r-wgcna) which is in the bioconda channel so I dont think the new changes should matter to my recipe. The build-test was completed in about 30 min

PR #11039, opened 2018-09-20
Weeder2 fix wrapper script

pjbriggs:

PR to update recipe for the `weeder` package, specifically it updates the Python wrapper script to fix an egregious bug whereby any valid non-trivial command line (i.e. anything except the `-h`/`--help` or any empty argument list) resulted in the wrapper calling itself with the same arguments, and so on *ad infinitum*.

As an aside: when I first encountered this bug, the infinite recursion ultimately consumed all available memory and effectively crashed the workstation I was running on. AFAIK this would happen for any non-trivial usage of the program, which would suggest that the original changes were never tested by their author.

There are a couple of other changes:

* Updates `meta.yml` to include C++ compiler as a build requirement (C compiler didn't seem to be enough when I tried to build locally)
* Updates the wrapper script to capture and return the exit status of the underlying `weeder` executable.

### DIFF
--- a/recipes/weeder/meta.yaml
+++ b/recipes/weeder/meta.yaml
@@ -11,7 +11,7 @@ build:
 
 requirements:
   build:
-    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
   run:
     - python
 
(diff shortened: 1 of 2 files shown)


pjbriggs:
Thanks @johanneskoester!

PR #11042, opened 2018-09-20
Mtbseq recipe

000CU000:


recipes/mtbseq/meta.yaml
```diff
@@ -0,0 +1,40 @@
+package:
+  name: mtbseq
+  version: "1.0.2"
```
johanneskoester: Please move the version into a jinja2 variable and use it in the source url. This makes an update of the recipe easier.

000CU000: Changed to jinja2 variable.

johanneskoester:
This PR made me reinvestigate templating in conda-build 3. It turned out that PKG_VERSION is now a supported jinja variable that can be used in the meta.yaml. This makes using self-defined jinja2 variables superfluous. I took the liberty to adapt your PR accordingly.

000CU000:
Thank you for the changes. Still learning.

johanneskoester:
All fine, the variable handling is actually in flow here at bioconda. Hence, there was not even a chance for you to do it right ;-).

PR #11052, opened 2018-09-21
metaxa

bernt-matthias:
Adds metaxa2 http://microbiology.se/software/metaxa2/

### DIFF
--- a/recipes/metaxa/build.sh
+++ b/recipes/metaxa/build.sh
@@ -1,7 +1,12 @@
 #!/bin/bash
 
-echo "
-dquote> no
+# cat <<- EOF || ./install_metaxa2
+# no
+# $PREFIX/bin
+# no
+# EOF
+
+echo "no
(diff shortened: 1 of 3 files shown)


apcamargo:
The build is failing because the installations script looks for `/usr/bin/perl`. Maybe a patch can fix that.

bernt-matthias:
@apcamargo thanks for the tip.

should work now. 

instead of executing the install (perl script) one could cp and chmod the scripts of this package manually. then we could remove the perl build requirement. what du you think?

apcamargo:
That's seems fine to me, but I'm no specialist. A member of the core team will look into this before merging.

Now, about the macOS build problem, this is happening because [the required vsearch version isn't available for macOS](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/vsearch/meta.yaml). The latest vsearch version for macOS is 2.7.0.

PR #11073, opened 2018-09-24
PacBio: Require at least numpy 1.15

pb-dseifert:
Fixes PacificBiosciences/pbbioconda#16


peterjc:
In my testing when reporting the issue, NumPy 1.14 seemed to work too - but if you want to be cautious and require NumPy 1.15 onwards, that seems fine too.

Assuming the automated tests pass, I'm happy. Thank you.

pb-dseifert:
@peterjc I saw the 1.14 comment, but when it comes to python hackery, I am against the "most-liberal dependency" set, i.e. the largest possible range of compatible versions, mainly because numpy is quite a beast and the CC2 module was compiled against 1.15.

peterjc:
If you compiled against NumPy 1.15, than I agree - be cautious and specify that as the minimum NumPy required.

PR #11113, opened 2018-09-26
Added initial htstream recipe

jenzopr:

Here I started packaging the HTStream tool set for fast quality control of Hight Throughput Sequencing data.


jenzopr:
It seems that the linux check did actually not do anything... ?

jenzopr:
Can someone from @bioconda/core have a glimpse into that? ☺️

rvalieris:
the meta file must the called `meta.yaml`
also cmake should go in the host section and the build section needs to have `{{ compiler('cxx') }}`

jenzopr:
Argh, so small an still so obvious! Thanks, Renan! 🎉

jenzopr:
Rationale for the commit above is the comment in https://github.com/ibest/HTStream/issues/131#issuecomment-425497146

jenzopr:
Ping @bioconda/lib 
I'm pinging you, because the CI-check fail is related to the boost library and looks like this:
```
CMakeFiles/hts_common_test.dir/src/utils.cpp.o: In function `setInputOptions()':

utils.cpp:(.text+0x203): undefined reference to `boost::program_options::options_description::options_description(std::string const&, unsigned int, unsigned int)'

...

/opt/rh/devtoolset-2/root/usr/libexec/gcc/x86_64-redhat-linux/4.8.2/ld: final link failed: Bad value
```

Upstream developers assume that `boost-dev` libraries are missing (see [here](https://github.com/ibest/HTStream/issues/131#issuecomment-425828882)). 
Locally, I tried to explicitly export `CFLAGS` and `LDFLAGS` as suggested for a missing `zlib` in the bioconda [troubleshooting](https://bioconda.github.io/troubleshooting.html#zlib).

Any ideas on how we could proceed are welcome 😊 
Best and thanks,
Jens

dpryan79:
Try exporting `LIBRARY_PATH=${PREFIX}/lib`.

rvalieris:
looks like the includes you declared on the CFLAGS are not being included on the compiler line, try CXXFLAGS/CPPFLAGS or check with the authors, there is probably a different way to do it with cmake.

acaprez:
We had an existing recipe for an earlier `htstream` version in our local repo.  I updated it to 1.0 just now and it built ok with a few tweaks.  The recipe is [here](https://github.com/unlhcc/hcc-conda-recipes/tree/master/recipes/htstream/1.0).  Feel free to copy or use it as you like.

jenzopr:
Dear Adam and Elmar, 
many many thanks for the recipe and the valuable hint to exclude the `-static` flag! It seems to work here as well. I will try to add the OSX build now!

jenzopr:
The OSX build failed because of a HTTP error while downloading cmake for OSX 😭  . Can someone with write access trigger a rebuild or should I bump the build number?

druvus:
@jenzopr restarted osx check

recipes/htstream/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
+
+mkdir build && cd build
+
+cmake -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_OSX_DEPLOYMENT_TARGET=10.9 -DCMAKE_INSTALL_PREFIX=${PREFIX} -DCMAKE_LIBRARY_PATH=${PREFIX}/lib -DCMAKE_INCLUDE_PATH=${PREFIX}/include ..
```
acaprez: It's a minor thing, but you might want to split the `cmake` command up over multiple lines for readability.

recipes/htstream/cmake.patch
```diff
@@ -0,0 +1,13 @@
+--- CMakeLists.txt	2018-09-22 18:12:34.000000000 +0000
++++ CMakeLists.txt.new	2018-10-02 16:54:44.438693120 +0000
```
dpryan79: You're going to want to fix this to `CMakeLists.txt`.

jenzopr:
Thanks a lot, guys - You helped a lot!! 🎉 🎆

dpryan79:
Nicely done @jenzopr !

PR #11155, opened 2018-09-30
Dinucleotide pattern analysis tools v1.0

erinijapranckeviciene:

@bioconda/core  Dear bioconda team, this is my first contribution, please review.  I am adding 4 recipes of programs to analyze dinucleotide patterns in nucleosomes DNA fasta sequences.  I tested these recipes locally with bootstrap and they have passed the test. I am submitting to bioconda channel because I need to have wrappers for these tools and their derivatives in Galaxy ToolShed  for an extended project.  I also hope that they may have a wider application. Thank you and  with kind regards, Erinija.


recipes/dnp-binstrings/meta.yaml
```diff
@@ -0,0 +1,33 @@
+package:
+  name: binstrings
+  version: "1.0"
+
+source:
+  url: https://github.com/erinijapranckeviciene/dnpatterntools/files/2431226/v1.0.tar.gz
+  sha256: bb38cf23132445d9c2ffec04229a578207eb932c38a5cfc8168d5c9a79f94a39
+
+build:
+  number: 1
```
bgruening: Please do start new version of a recipe always with 0.

recipes/dnp-binstrings/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+svn checkout https://github.com/seqan/seqan/trunk/include >/dev/null 2>&1
```
bgruening: This is discouraged to do. Instead you could depend on the seqan library `seqan-library`.

recipes/dnp-binstrings/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+
+  host:
+    - svn
```
bgruening: if you depend on seqan you can hopefully get rid of the svn dependency

recipes/dnp-corrprofile/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+svn checkout https://github.com/seqan/seqan/trunk/include >/dev/null 2>&1
```
bgruening: see above

recipes/dnp-corrprofile/meta.yaml
```diff
@@ -0,0 +1,33 @@
+package:
+  name: corrprofile
+  version: "1.0"
+
+source:
+  url: https://github.com/erinijapranckeviciene/dnpatterntools/files/2431226/v1.0.tar.gz
+  sha256: bb38cf23132445d9c2ffec04229a578207eb932c38a5cfc8168d5c9a79f94a39
+
+build:
+  number: 1
```
bgruening: 0 please

recipes/dnp-binstrings/meta.yaml
erinijapranckeviciene: Thank you, yes, will start with 0 in all.

recipes/dnp-binstrings/build.sh
erinijapranckeviciene: I agree. Is there some documentation on a list of available dependencies and their proper names that I could consult. I am not sure that I have very clear understanding how to reference the dependencies.  Thank you for the comments.

recipes/dnp-binstrings/meta.yaml
erinijapranckeviciene: Yes, I agree the dependences would be better. Tidier. Thank you.

recipes/dnp-binstrings/build.sh
bgruening: You can search here: https://anaconda.org/bgruening/dashboard

Everything which is in the conda-forge channel and the Bioconda channel.

erinijapranckeviciene: Thank you, I found it.  I will change  recipes , test locally and will do a new push.  

I have another question, if it is appropriate here.  I did my local testing with bootstrap without docker and it was a success. Now with  CircleCI through github  I have a fail. The reason of fail is:

gcc -std=c++11 -std=c++14 -pthread -I include/ binstrings.cpp -lstdc++ -lrt -lm -o binstrings  
gcc: error: unrecognized command line option ‘-std=c++14’

I am not sure that I will be able to do testing on Ubuntu with docker (short of  resources). It would be helpful if I had a hint what should be changed - why  the gcc -std=c++14  option is not recognized (seems that -std=c++11 passed)  in the test docker environment , what dependency / syntax  I should use in the recipes? 

Many thanks in advance.

bgruening: We are currently building with older compilers by default, and those are not knowing about this option. Is c++14 required by your pkg? If so please add this file to your pkgs https://github.com/bioconda/bioconda-recipes/blob/master/recipes/kmc/conda_build_config.yaml.

It will overwrite the defaults and use a newer compiler stack - hopefully.

erinijapranckeviciene:
Hi @bgruening , this is just to let you know that dnpatterntools passed the tests. If you have time to review, then I would like to ask that you would review and tell the next steps. Many thank in advance.

PR #11177, opened 2018-10-01
Add tinyfasta-0.1.0

dfornika:

### DIFF
--- a/recipes/tinyfasta/meta.yaml
+++ b/recipes/tinyfasta/meta.yaml
@@ -13,7 +13,8 @@ source:
 build:
   number: 0
   noarch: python
-  script: python -m pip install --no-deps --ignore-installed .
+  build:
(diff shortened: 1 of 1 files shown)


recipes/tinyfasta/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/tjelvar-olsson/{{ name }}/archive/{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+  noarch: python
+  script: python -m pip install --no-deps --ignore-installed .
```
bgruening: I guess we should use this now: https://conda-forge.org/docs/meta.html#use-pip

dfornika: Hmmm.... not sure how that broke the build...

PR #11185, opened 2018-10-02
Add r-conflicted package.

grst:
" An alternative conflict resolution strategy for
R" - https://github.com/r-lib/conflicted

@bioconda/core, please review.

### DIFF
--- a/recipes/r-conflicted/build.sh
+++ b/recipes/r-conflicted/build.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+if [[ $target_platform =~ linux.* ]] || [[ $target_platform == win-32 ]] || [[ $target_platform == win-64 ]] || [[ $target_platform == osx-64 ]]; then
+  export DISABLE_AUTOBREW=1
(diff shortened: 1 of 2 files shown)


recipes/r-conflicted/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if [[ $target_platform =~ linux.* ]] || [[ $target_platform == win-32 ]] || [[ $target_platform == win-64 ]] || [[ $target_platform == osx-64 ]]; then
```
bgruening: We do not build for Windows, so I think this file can be made less complex.

grst:
Is this paragraph from the docs outdated then?

>The exception to this is with R packages. We are still coordinating with Anaconda and conda-forge about the best place to keep general R packages. In the meantime, R packages that are not specific to bioinformatics and that aren’t already in the conda-forge channel can be added to bioconda.

(https://bioconda.github.io/guidelines.html)

PR #11189, opened 2018-10-02
vcfSampleCompare

hepcat72:
I have resolved all issues that had previously made me tag this as WIP.  Please consider this PR for review.

The main utility of this script is to be able to compare sample groups contained in a single VCF file.  Think of it as the SNP/SV version of DESeq2, because it answers the basic question a researcher has, which is "What's different between my wildtype samples/replicates and my mutant samples?".  It works on freeBayes output and svTyper output.

My goal is to submit this as a galaxy tool wrapper so I can finalize a galaxy workflow version of my current shell script pipeline for finding prokaryotic variants.


hepcat72:
I cancelled the circleci tests because this is where I already know it hangs.

recipes/vcfsamplecompare/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+requirements:
+  build:
+    - perl
+    - make
+  run:
+    - perl
+
+source:
+  url: https://github.com/hepcat72/{{ name }}/archive/{{ version }}.tar.gz
+  md5: 7be40cdfa8ab7a18666ab7244ff7c040
```
nsoranzo: You can drop md5 if you provide sha256.

hepcat72:
This PR is ready for full review now - it's no longer a WIP.

recipes/vcfsamplecompare/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+    variant/row.
+
+build:
+  number: 0
+package:
+  name: {{ name }}
+  version: {{ version }}
+requirements:
+  build:
+    - perl
```
nsoranzo: `perl` should go in the `host:` section, see https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#host

recipes/vcfsamplecompare/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+source:
+  url: https://github.com/hepcat72/{{ realname }}/archive/{{ version }}.tar.gz
+  sha256: 646515313907441ea8540c359baf552682fbab27396abadac3ff8b6c4402d71c
+
+test:
+  commands:
+    - {{ realname }}.pl --usage
+    - {{ realname }}.pl --version
+    - {{ realname }}.pl --help
```
nsoranzo: I think that just `--version` should be enough.

recipes/vcfsamplecompare/meta.yaml
hepcat72: OK.  I'm confused.  I added:

```
  host:
    - perl
```

`conda build` succeeds, but when I do a local `circleci build`, I get:

```
+ vcfSampleCompare.pl --version
/opt/conda/conda-bld/vcfsamplecompare_1538760954997/test_tmp/run_test.sh: line 7: vcfSampleCompare.pl: command not found
```

I did it twice.  Failed twice.  I took the host back out and it succeeded again.

My guess is that the perl path on the she-bang line at the top of the script is different on the "host" compared to what I have locally.  I had noticed that the tests in the meta.yaml file were using my manually installed version of the script instead of a conda-built one.  The Makefile.pl script should be configuring things to edit that line for the "user's" perl path (the one they used to run Makefile.PL on the line `perl Makefile.PL` in the build.sh script).

So what do you think is going on here?  Is there no install performed on the host or is it not configuring the she-bang line correctly?

hepcat72: Or do I need to add something to the PATH from build.sh?

hepcat72: OK, I've been looking at build.sh scripts containing `Makefile.PL` and it looks like I need to supply $PREFIX in some arguments to Makefile.PL...  That makes sense.  I'll give it a go and update if/when I get it figured out.

hepcat72: I feel like I'm closer, but I must be supplying the wrong environment variable?

I now get this error:

```
+ vcfSampleCompare.pl --version
/Users/rleach/local/miniconda3/conda-bld/vcfsamplecompare_1538766140354/test_tmp/run_test.sh: /Users/rleach/local/miniconda3/conda-bld/vcfsamplecompare_1538766140354/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/bin/vcfSampleCompare.pl: /Users/rleach/local/miniconda3/conda-bld/vcfsamplecompare_1538766140354/_build: bad interpreter: No such file or directory
```

So at least it's *trying* to run a conda version of the script instead of using the one I had installed manually, but the perl interpreter path is incorrect?  Not sure why the entire interpreter path isn't shown, so I double-checked the she-bang line:

```
>head /Users/rleach/local/miniconda3/conda-bld/vcfsamplecompare_1538766140354/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/bin/vcfSampleCompare.pl
#!/Users/rleach/local/miniconda3/conda-bld/vcfsamplecompare_1538766140354/_build_env/bin/perl
```

The build directory was moved because of the failed build, but the perl path on the she-bang appears to have been valid: `/Users/rleach/local/miniconda3/conda-bld/vcfsamplecompare_1538766140354/build_prefix_moved_vcfsamplecompare-v2.005-pl526_0_osx-64/bin/perl`.  I can run it, NP.

I tried manually editing the she-bang to reflect the moved build dir, and it runs no problem:

```
>/Users/rleach/local/miniconda3/conda-bld/vcfsamplecompare_1538766140354/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/bin/vcfSampleCompare.pl
vcfSampleCompare.pl -i <file*...>... [OPTIONS]

* <file*...>...       VCF input file.
  -o <sfx>            VCF outfile suffix (appended to -i).
  -u <sfx>            [STDOUT] Summary outfile suffix (appended to -i).
  -s <str ...>...     [any^] A group of sample names for difference comparisons.
                      ^ See --extended usage.
  -d <int,...>...     [all] Minimum number of samples to use in a group to
                      determine difference with its partner.
  -a <flt>            [0.6] Allelic frequency difference threshold [0-1].
  --no-g              Do not use genotype calls (use allelic frequency).
  --no-f              Do not filter variant rows whose sample groups do not
                      differ (enough).
  --no-w              Do not add as many samples to sample groups as possible.
  --help              Print general info and file formats.
  --extended [<cnt>]  Print detailed usage.

* Required.
```

So I don't know why I get the error:

`/Users/rleach/local/miniconda3/conda-bld/vcfsamplecompare_1538766140354/_build: bad interpreter: No such file or directory`

Perhaps the parsed perl path is getting truncated (as shown in the error)?  What do you think?

hepcat72: *sigh* I'm stumped again. I can't seem to satisfy the setup on the host. I'm going to head over to gitter and ask some questions...

hepcat72: OK, I finally figured it out.

recipes/vcfsamplecompare/meta.yaml
hepcat72: Done.

recipes/vcfsamplecompare/meta.yaml
hepcat72: It wasn't obvious/clear to me from the doc that perl shouldn't also be in the 'build:' section, since I was using a perl call in the build.sh script. Lance caught that in short order. I banged my head against this for a couple days.

hepcat72:
Hey great. Thank you @nsoranzo.

PR #11205, opened 2018-10-03
New recipe for RFMix

ryanraaum:

### DIFF
--- a/recipes/rfmix/meta.yaml
+++ b/recipes/rfmix/meta.yaml
@@ -1,6 +1,8 @@
 package:
   name: rfmix
-  version: '2.9505bfa'
+  # Version number in github for this recipe is 2.03-r0, but there are no release tarballs. 
+  # To ensure stability, this recipe links to the git commit with prefix '9505bfa'
+  version: '2.03-r0.9505bfa'
 
 source:
   url: https://github.com/slowkoni/rfmix/archive/9505bfae51ea57314d98060e6d09f6759cda8e8d.zip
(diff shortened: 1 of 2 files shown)


ryanraaum:
@bioconda/core - requesting review of a new package recipe - thanks!

recipes/rfmix/meta.yaml
```diff
@@ -0,0 +1,36 @@
+package:
+  name: rfmix
+  version: '2.9505bfa'
```
rvalieris: where did you got this version ? the rfmix help says:
`RFMIX v2.03-r0`

ryanraaum: There are no formal releases (only the repository), so I used the most recent git commit prefix as the specifier and linked the download to that specific commit so that it is stable. I could change it to 2.03-r0, but that's not a stable, set, tarball release.

ryanraaum: Maybe 2.03-r0.9505bfa ?

rvalieris: oh I see, thats the commit prefix. yes I think its a good idea. 
how about: `2.03-r0-9505bfa`
and add a comment around the version code indicating the commit prefix was added.

ryanraaum: Sounds good. Will do.

rvalieris: uh oh, I forgot about the linting, use dots instead.

PR #11245, opened 2018-10-06
Dinucleotide patterns analysis tools v1.0

erinijapranckeviciene:

### DIFF
--- a/recipes/dnp-binstrings/meta.yaml
+++ b/recipes/dnp-binstrings/meta.yaml
@@ -13,13 +13,11 @@ requirements:
   build:
     - {{ compiler('c') }}
     - {{ compiler('cxx') }}
-    - seqan-library
     - cmake
-
   host:
     - zlib
     - bzip2
-
+    - seqan-library
   run:
     - zlib
     - bzip2
--- a/recipes/dnp-binstrings/meta.yaml
+++ b/recipes/dnp-binstrings/meta.yaml
@@ -1,0 +1,34 @@
+package:
+  name: binstrings
+  version: "1.0"
+
+source:
+  url: https://github.com/erinijapranckeviciene/dnpatterntools/files/2450741/v1.0.tar.gz
+  sha256: 19c46d6bc6a1af70e5a2b4a051aab334026dc38c84b367ea9c445da3fdb21cfe
+
+build:
+  number: 0
+
+requirements:
+  build:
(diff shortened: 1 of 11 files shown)


recipes/dnp-binstrings/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+  sha256: 19c46d6bc6a1af70e5a2b4a051aab334026dc38c84b367ea9c445da3fdb21cfe
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - seqan-library
```
bgruening: Seqan should probably in the host section.

erinijapranckeviciene: Thank you for the comment. It was a journey... No. Seqan is C++ header only library. It matters only in compilation.  But it depends on zlib and bzip2. The two should be in the host and run.

bgruening:
Thanks @erinijapranckeviciene!

erinijapranckeviciene:
Thank you @bgruening !  Does it mean I can try to install my tools with conda install?

bgruening:
Yes. You can see it here: https://anaconda.org/bioconda/ :)

erinijapranckeviciene:
Thank you. This is only half of the journey.  Now I will be submitting tools that depend on the dnpatterntools to the Galaxy ToolShed.  Thank you and bioconda team for being a driving force of bioinformatics open source  community. With best regards !

bgruening:
@erinijapranckeviciene your are welcome!

Looking at the bioconda page, your packages are not named `dnp-*`. This is wrong isn't it? They should be labeled with dnp as prefix I guess. If so please create a PR with `name: ` change to include the prefix and I will take care of the rest.

erinijapranckeviciene:
Thank you for the notice. I don't know if I am breaking some naming rules
with this naming. The packages don't need the dnp- prefix because they are
individual packages. I made dnp- subdirs for myself to be able to search
them as a group in the recipes directory and  did not rename in PR. Shall I
drop this dnp- subdir prefix and create a pull request to make subdirs
named after the packages to have things tidy, or it  is OK to have them
like this?

bgruening:
I would slightly prefer the prefix solution. As your names are a little bit generic and I fear name-collisions with other recipes/programs.

PR #11252, opened 2018-10-07
Rrrr bump all the rrr things

dpryan79:

I'm hoping that ~20% will build right away.


dpryan79:
@bioconda/core This is the main PR for R-3.5.1 support in most of the R packages. Can someone confirm that this should be merged into `bulk`? I presume so, due to it inevitably timing out and there being a good percent of packages that will fail due to packages needing updated on conda-forge, but I'd like to double check before clicking merge.

bgruening:
@dpryan79 thanks for working on this.
Yes this is still the correct way. We should merge it into bulk. One change we discussed recently with @johanneskoester is if we want to pin bioconductor dependencies to `>=X.Y,<=X.Y+2` to have indirect pinning to bioconductor releases. Do you think this is something we should explore here or with the next release?

dpryan79:
I guess there's no time like the present. Am I correct in understanding that the `<=X.Y+2` would be only for those packages already specifying versions in their dependencies? Or do you also want to try and figure out what version is currently going to get installed with `- bioconductor-something` and add an appropriate max value?

bgruening:
> Am I correct in understanding that the <=X.Y+2 would be only for those packages already specifying versions in their dependencies? 

Sorry, it should be `<X.Y+2` and for all packages. The assumption is that bioconductor has odd version numbers as developent versions between bioc releasesn and even numbers for every bioc release. With `>=X.Y,<X.Y+2` we could pin to exact releases supported for one particular bioc release.

Rational behind this is that bioc maintainers are worried that packages between bioc releases gets mixed.

bgruening:
I don't know, but this is the feedback we got from bioc people as we wanted to start a more close collaboration with them.

dpryan79:
Anything that facilitates that would be a positive. I'll update this PR Tuesday when I'm back in the office then. At some point the bioconductor-skeleton should get updated too, but that can wait I presume.

bgruening:
Cool thanks. 

I' wondering how you want to update this PR without adding magic to the skeleton wrapper?

dpryan79:
I'll just write a script to rewrite the recipes and add the extra pinning for bioconductor dependencies that have versions. One thing I wonder about is how many recipes set a minimum version to something relatively old (e.g., from when a package was added to bioconductor), which means that maybe naively adding `<X.Y+2` might not really pin to a bioconductor release. I guess that more of a question of how strict the are with ensuring the minimum versions get bumped all them time.

bgruening:
>  extra pinning for bioconductor dependencies that have versions

I think we want to pin all `bioconductor-*` versions to the current version that is currently in 3.7 with an upper bound of `<+2` to not get the next stable version. Means replacing all available pins.
That means the change should probably be happen on the skeleton level with a list of all versions that are currently available.

dpryan79:
@bgruening Have a look, everything should now be pinned as described to a given BioC release (interesting that they do indeed enforce minor version bumps as you described, I never knew that). I have not yet updated the skeleton to produce this, though should do so before the next BioC release.

dpryan79:
FYI, there's one package `DO.db` that doesn't follow the recommended bioconductor versioning scheme. It only has major and minor versions but no patch levels. I treated it the same as all of the others and just gave it a `.0` patch level, I hope that doesn't cause issues.

recipes/bioconductor-a4base/meta.yaml
```diff
@@ -10,35 +10,35 @@ source:
...
-    - bioconductor-genefilter
-    - bioconductor-limma
-    - bioconductor-multtest
+    - 'bioconductor-a4core >=1.28.0,<1.30.0'
+    - 'bioconductor-a4preproc >=1.28.0,<1.30.0'
+    - 'bioconductor-annaffy >=1.52.0,<1.54.0'
+    - 'bioconductor-annotationdbi >=1.42.1,<1.44.0'
+    - 'bioconductor-biobase >=2.40.0,<2.42.0'
+    - 'bioconductor-genefilter >=1.62.0,<1.64.0'
+    - 'bioconductor-limma >=3.36.5,<3.38.0'
```
bgruening: In theory the this should probably `>=3.36.0`, in practice, it does not matter I guess.

recipes/bioconductor-a4base/meta.yaml
```diff
@@ -50,7 +50,5 @@ about:
   home: 'http://bioconductor.org/packages/{{ bioc }}/bioc/html/{{ name }}.html'
   license: GPL-3
   summary: 'Automated Affymetrix Array Analysis'
-extra:
```
bgruening: @dpryan79 please ignore the `extra` section and do not overwrite it. This section is manually added with curation and we probably would like to keep it.
Indeed it would be cool if we could extract this information at some point from the package with the skeleton.

recipes/bioconductor-abaenrichment/meta.yaml
```diff
@@ -10,37 +10,37 @@ source:
...
   run:
-    - 'bioconductor-abadata >=0.99.2'
+    - 'bioconductor-abadata >=1.10.0,<1.12.0'
     - r-base
     - 'r-data.table >=1.10.4'
     - 'r-gplots >=2.14.2'
     - 'r-gtools >=3.5.0'
     - 'r-rcpp >=0.11.5'
-    - libcxx # [osx]
+  build:
```
bgruening: The ordering is wired here for the sections, but I guess we do not care atm

recipes/bioconductor-a4base/meta.yaml
dpryan79: D'oh, I bet I accidentally did that when I removed the skip-lints stuff. I'll add them all back in tomorrow.

dpryan79:
@bgruening I've hopefully added all of the indentifier and other such information back in now.

recipes/bioconductor-a4base/meta.yaml
dpryan79: Yeah, that can wait for the skeleton to get updated.

dpryan79:
I'll merge in now, we'll see how many we can get built by the time bioconductor 3.8 is released. At the very least this process will facilitate that due to showing which conda-forge packages need updating.

PR #11263, opened 2018-10-08
Add recipe for seer

johnlees:

This is an attempt to resurrect #5005 from @Slugger70 now that the newer gcc libraries are available.

It almost works, but I am getting a little out of my depth so was hoping to solicit some advice from @bioconda/core here.

The compiling is failing at the linking stage, I think because the gcc version used here is different from that used to compile the boost dependency. 
Would it be best to try and recompile the necessary boost library as part of this recipe?
Also, would it be best to perform static linking to avoid these issues on the running machines?

Many thanks!

### DIFF
--- a/recipes/seer/build.sh
+++ b/recipes/seer/build.sh
@@ -1,0 +1,8 @@
+pushd gzstream
+make clean
+make CXX="$CXX $CXXFLAGS" CPPFLAGS="$CPPFLAGS -I. -I$PREFIX/include" AR="$AR cr"
+popd
+pushd src
+make CXXFLAGS="$CXXFLAGS"
+make install
+popd
--- a/recipes/seer/build.sh
+++ b/recipes/seer/build.sh
@@ -1,13 +1,54 @@
+#!/bin/bash
+
+# build gzstream for static linking
 pushd gzstream
 make clean
 make CXX="$CXX $CXXFLAGS" CPPFLAGS="$CPPFLAGS -I. -I$PREFIX/include" AR="$AR cr"
 popd
+
+# build boost program options for static linking
+# see
+# https://github.com/conda-forge/boost-cpp-feedstock/blob/master/recipe/build.sh
+# https://github.com/boostorg/program_options/blob/develop/.travis.yml
 pushd boost
 python2 tools/boostdep/depinst/depinst.py program_options --include example
-./bootstrap.sh --with-libraries=program_options
-./b2 install toolset=gcc
+
+CXXFLAGS="${CXXFLAGS} -fPIC"
+INCLUDE_PATH="${PREFIX}/include"
+LIBRARY_PATH="${PREFIX}/lib"
+LINKFLAGS="${LINKFLAGS} -L${LIBRARY_PATH}"
+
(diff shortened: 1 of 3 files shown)


johnlees:
Also I'm not sure I understand the second linting error - I have the line:
```
  build:
    - {{ compiler('cxx') }}
```
in `meta.yaml`

johnlees:
Ok, reading the docs I see this is because of the `libgcc` dependency.
While seer can be compiled with gcc 4.8, to run it requires libgcc > 4.9, as it uses the regex functions (part of the standard used by gcc 4.8, but not implemented in the library until 4.9).

Would it be best to turn off this linting, or try with the standard compiler?

bgruening:
Hi @johnlees! `fn` needed to be removed and libgcc.

bgruening:
Consider using the new compiler toolchain by adding this file to your recipe: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/kmc/conda_build_config.yaml

johnlees:
@bioconda/core any idea why this worked in c66d8cbccf089476dd3a15da6e94394d4f37c67b (https://circleci.com/gh/bioconda/bioconda-recipes/29475) but is now failing now the build is not being run.
The error seems to be an issue with an undefined symbol from lapack

johnlees:
Is there a way to force this to rebuild?

johnlees:
(continuing to make notes to self here)
I think this could be due to the armadillo wrapper. Try with `-DARMA_DONT_USE_WRAPPER`

johnlees:
I've finally got this to work! @bioconda/core could you please review?

This was hard, but I learnt a lot about conda!

recipes/seer/build.sh
```diff
@@ -0,0 +1,54 @@
...
+    runtime-link=shared \
+    link=static,shared \
+    toolset=gcc-custom \
+    include="${INCLUDE_PATH}" \
+    cxxflags="${CXXFLAGS}" \
+    linkflags="${LINKFLAGS}" \
+    --layout=system \
+    -j"${CPU_COUNT}" \
+    install
+popd
```
bgruening: +1 for using popd and pushd :)

recipes/seer/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  sha256: db35dc61374ebd593093f0dbf07305a8d97e41307562c3858608c28f3829a41e
+
+build:
+  skip: True #  [osx]
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - python ==2.7.14
```
bgruening: I think `<3` should be better.

recipes/seer/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  host:
+    - zlib
+    - armadillo
+    - hdf5
+    - pthread-stubs
+  run:
+    - zlib
```
bgruening: We have now run_exports in zlib ... so this can be removed from the run-section: https://github.com/conda-forge/zlib-feedstock/blob/master/recipe/meta.yaml#L14

PR #11267, opened 2018-10-09
Genomepy 0.5.2

simonvh:

Bumped the build number to trigger build.


druvus:
@simonvh genomepy is blacklisted so you need to remove it from `build-fail-blacklist` file to get it tested and built.

simonvh:
Thanks for the notification @druvus! Have to say it's kind of a weird situation when a recipe is blacklisted but there is no notification or any sign at all that something is wrong.

PR #11268, opened 2018-10-09
New version for biopet-sampleconfig

ffinfo:
***

This pull request was automatically generated by [sbt-bioconda](https://github.com/biopet/sbt-bioconda). A release of [biopet-sampleconfig](https://github.com/biopet/sampleconfig) is automatically generated if the `sbt biocondaRelease` command is issued. As part of this process the recipes are tested using CircleCi as described [here](https://bioconda.github.io/contribute-a-recipe.html#test-locally). 

***

#### Tools - ExtractTsv

This mean can extract samples, libraries and readgroups from a sample config file. This meant as a supporting tool inside wdl pipelines.
It can also output a single layer as tsv file.

#### Tools - ReadFromTsv

This tool enables a user to create a full sample sheet in JSON format or
YAML format, suitable for all Biopet Queue pipelines, from TSV file(s).

#### Tools - CromwellArrays

This tool will convert the sample configs to a array based format that can be used inside wdl pipelines.
This tool is only to support biowdl pipelines.

#### Tools - CaseControl

 This tool will extract the case-control pairs from a sample config file.
 It will read the headers of the bam files to confirm that samples do exist.

For documentation and manuals visit our github.io page: https://biopet.github.io/sampleconfig


ffinfo:
oke this also not fixing the cpu load, I'm trying t debug this but can't see where the load comes from. Maybe the calander call?

Right now I'm trying to create a extra poll queue just for `isAlive`. Altough with the the status is not passed anymore to the normal polling. I think this is because the handle stays in memory till the jobActor is finished, starting a new queue mean a copy of the handle.

PR #11274, opened 2018-10-09
Add DCA recipe

bebatut:

### DIFF
--- a/recipes/dca/meta.yaml
+++ b/recipes/dca/meta.yaml
@@ -12,29 +12,22 @@ source:
 build:
   skip: True # [not py36]
   number: 0
+  noarch: python
   entry_points:
     - dca = dca.__main__:main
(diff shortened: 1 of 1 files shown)


recipes/dca/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: f3b979c3e11de0dfba6503631db5e1abd5e926b9233ec956ab7da4badb31c4e6
+
+build:
+  skip: True # [not py36]
```
bgruening: This could also be a noarch package, isn't it? You can skip the python version on the pinning level.

PR #11278, opened 2018-10-09
FGMP-1.0.1

hyphaltip:

### DIFF
--- a/recipes/fgmp/meta.yaml
+++ b/recipes/fgmp/meta.yaml
@@ -1,0 +1,64 @@
+package:
+  name: fgmp
+  version: 1.0.1
+
+source:
+  url: https://github.com/stajichlab/FGMP/archive/v1.0.1.tar.gz
+  sha256: 18a95eacaa96ea24bfcd5b848355c4b1b6845ea1d1e86c014ffdf8927ecc6385
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - perl >=5.10
+    - perl-ipc-run
+    - perl-bioperl
+    - augustus >=3.0
+    - blast ==2.2.31
+    - emboss >=6.5.7
+    - exonerate >=2.2.0
+    - hmmer >=3.0
+
+  host:
+    - perl >=5.10
+    - perl-ipc-run 
+    - perl-bioperl
+    - augustus >=3.0
+    - blast ==2.2.31
+    - emboss >=6.5.7
+    - exonerate >=2.2.0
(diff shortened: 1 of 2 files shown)


recipes/fgmp/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  url: https://github.com/stajichlab/FGMP/archive/v1.0.1.tar.gz
+  sha256: 18a95eacaa96ea24bfcd5b848355c4b1b6845ea1d1e86c014ffdf8927ecc6385
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - perl >=5.10
```
bgruening: I think everything besides the compiler can be removed in build.

recipes/fgmp/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+source:
+  url: https://github.com/stajichlab/FGMP/archive/v1.0.1.tar.gz
+  sha256: 18a95eacaa96ea24bfcd5b848355c4b1b6845ea1d1e86c014ffdf8927ecc6385
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
```
bgruening: You are not compiling at all isn't it? Do you really need the compiler?

recipes/fgmp/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+#    - perl $FGMP/src/fgmp.pl -h
+
+#commands:
+#    - perl $FGMP/src/fgmp.pl -g sample.dna 2> log
+#    - blast ?
+
+about:
+  home: https://github.com/stajichlab/FGMP
+  license: 'MIT'
+  summary: 'FGMP: assessing fungal genome completeness and gene content. bioRxiv 049619'
```
bgruening: you can remove the bioRxiv ID here ...

recipes/fgmp/meta.yaml
hyphaltip: That’s correct. Let’s remove

recipes/fgmp/meta.yaml
hyphaltip: Ok no citations needed. We can remove.

hyphaltip:
To followup I pushed these suggested removals on the branch so I guess check after tests run and see if it can be merged?

hyphaltip:
we are re-submitting paper so would like this to be available for installation as soon as possible.

PR #11288, opened 2018-10-10
Unicycler downgrade seqan patch from 2.4.0 to 2.3.2

PertuyF:

Try using Seqan 2.3.2 instead of 2.4.0.
This version should have [improved gcc7 compliance](https://github.com/seqan/seqan/blob/master/CHANGELOG.rst#release-232) compare to 2.3.1 that led to issues building Unicycler 0.4.6 with PR #10510 , but may solve a [blocking issue](https://github.com/rrwick/Unicycler/issues/140) encountered since use of Seqan 2.4.0.

### DIFF
--- a/recipes/unicycler/meta.yaml
+++ b/recipes/unicycler/meta.yaml
@@ -16,7 +16,7 @@ source:
     folder: 'seqan'
 
 build:
-  number: 0
+  number: 1
 
 requirements:
   build:
--- a/recipes/unicycler/meta.yaml
+++ b/recipes/unicycler/meta.yaml
@@ -11,8 +11,8 @@ source:
     patches:
       - Makefile.patch  #[linux]
       - misc.py.patch
-  - url: https://github.com/seqan/seqan/archive/seqan-v2.4.0.tar.gz
-    sha256: 'd7084d17729214003e84818e0280a16f223c8f1c6a30eeef040c27e0c0047bd7'
+  - url: https://github.com/seqan/seqan/archive/seqan-v2.3.2.tar.gz
+    sha256: '73ce4b344e6a259db1518547544c730f548bc0b32292b30e3a027146e158085e'
     folder: 'seqan'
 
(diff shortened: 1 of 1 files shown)


PertuyF:
@corburn @bgruening @sjackman What do you think about merging this one? I suspect Seqan v0.2.4 broke something for the hybrid assembly.

sjackman:
@PertuyF Are you able to reproduce the issue yourself, and does this PR resolve the issue?

PertuyF:
I know people that reproduced the issue with 0.4.7, but they confirmed 0.4.4 works fine. 
They should be able to test this PR very shortly.

bgruening:
I would be fine maintaining two versions next to each other. Does this work for you?

sjackman:
It'd be better to resolve the issue with Unicycler 0.4.7 if possible. @PertuyF Indicates that he knows people who can reproduce the issue, and verify whether this PR fixes the issue.
If it does resolve the issue, I suggest opening an issue upstream to ask that the version of SeqAn included in Unicycler be updated. https://github.com/rrwick/Unicycler/issues

PertuyF:
I agree with Shaun, I would prefer fixing the issue, and suggesting the fix to the Unicycler repo.
Due to a unexpected delay, testing have just started. I should know if this fixes the issue early next week.

PertuyF:
Looks like this PR does not solve the issue 😞 
I am conducting some more testing and will keep you updated.

PertuyF:
My tests confirmed that this PR does not change anything compared to the current build. On the other hand, it is not clear whether the issue mentioned above is related to the patch for `seqan 2.4.0`, as I got hybrid assemblies working with the current package. Asking for upstream integration of a gcc7-compatible library will likely the best way to go for future versions.

Directly using `seqan-library` from conda-forge in this recipe could also be an option. Any though about this @sjackman @bgruening ?

Closing.

Issue #11303, opened 2018-10-11
umi_tools: Pinned depenency versions


IanSudbery:
Hi,

I am one of the authors of `umi_tools`, however, I didn't build the biconda recipe (and we are super greatful to those maintaining this for us). The bioconda recipe includes a dependency for `certifi` that was not in the `meta.yml` from our own channel (. The version in the bioconda recipe is pinned at `certifi=2016.9.26`. However one of the pacakges that `umi_tools` is often used with is pinned at `certifi=2018.8.24`. 

Can someone explain to me:
1) Why this dependency is present
2) Why it is pinned at this version?

Cheers.

IanSudbery:
I notice that this requirement was added with the recipe was updated by @davebx  for `v0.5.0` (PR #5875), and that it was queried at the time by @croth1 .

croth1:
continued in: https://github.com/bioconda/bioconda-recipes/pull/11376

PR #11312, opened 2018-10-11
umi_tools:  clean dependencies

croth1:

removed dependency on certifi. Not sure why it was needed in the first place - also the hard pinning prevents updating a lot of stuff.

### DIFF
--- a/recipes/umi_tools/meta.yaml
+++ b/recipes/umi_tools/meta.yaml
@@ -11,7 +11,7 @@ source:
   sha256: {{ sha256hash }}
 
 build:
-  number: 1
+  number: 2
   script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv"
 
 requirements:
--- a/recipes/umi_tools/meta.yaml
+++ b/recipes/umi_tools/meta.yaml
@@ -13,6 +13,8 @@ source:
 build:
   number: 2
   script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv"
+  entry_points:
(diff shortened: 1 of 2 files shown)


recipes/umi_tools/meta.yaml
```diff
@@ -12,17 +12,18 @@ source:
 
 build:
   number: 1
```
bgruening: you need to bump the build number here

croth1: good catch, thanks!

croth1:
Fixes https://github.com/bioconda/bioconda-recipes/issues/11303.

@IanSudbery, if you have some spare time, could you please review the dependencies in the recipe? I am a bit surprised about `freetype` and `pysam`. Do you know whether these are indeed required?

IanSudbery:
Thanks for the fast turnaround! I believe that `freetype` is a dependency of `matplotlib`, but we certainly don't use it directly, and its not on our dependency list. We used to depend on `cython` because we cythonized and complied our extension on the fly with `pyximport`, but we've long since switched to provided the cythonized c file and compiling on install, so it *should* work without it.  I'm happy to run any new recipe through our testing boxes if you tell me how to use a recipe not merged into master.

croth1:
@bgruening, is everything fine with the checks? Why does it not show the CI?

croth1:
@bioconda/core any idea why the CI checks seem to be broken here?

rvalieris:
try merging changes from master perhaps ?

croth1:
@rvalieris is that required? CI ran for the first commit without problems, but now it's stuck.

rvalieris:
its just a guess really, I don't know why its happening either.

PR #11320, opened 2018-10-12
Various fixes, at least 4 packages (and their dependencies) should st…

dpryan79:
…ill fail


recipes/r-speaq/meta.yaml
```diff
@@ -29,14 +29,14 @@ requirements:
 
   host:
     - r-base
-    - r-massspecwavelet
+    - bioconductor-massspecwavelet
```
bgruening: I think we need to pin those as well, isn't it?

PR #11321, opened 2018-10-12
Add nullarbor 1.41

druvus:
Continuation of PR #7947 now when we have perl 5.26 dependencies in order.


druvus:
Forgot that we still need to fix kraken for osx

dpryan79:
This has since been added

PR #11391, opened 2018-10-18
Expansionhunter 2.5.5 (unblacklist)

apeltzer:

Apparently expansionhunter was on the blacklist thus the 2.5.5 update yesterday didn't get uploaded. Since its both testing and building just fine, I just removed it from the blacklist and created this PR again, bumping build version too.

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1734,10 +1734,6 @@ recipes/arvados-cli
 # takes to long to install
 recipes/bcbio-rnaseq
 
-# cannot open file
-
-recipes/expansionhunter
-
 recipes/biomaj
 # missing python requirements?
 recipes/illuminate
--- a/recipes/expansionhunter/meta.yaml
+++ b/recipes/expansionhunter/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "expansionhunter" %}
-{% set version = "2.0.8" %}
+{% set version = "2.5.5" %}
 
 
 package:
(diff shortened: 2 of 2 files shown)


dpryan79:
I have a feeling that the system gcc is getting used instead of the one that gets installed by conda. Likely this is incorrectly hard-coded some where in the build script.

apeltzer:
> I have a feeling that the system gcc is getting used instead of the one that gets installed by conda. Likely this is incorrectly hard-coded some where in the build script.

Maybe in the imported JSON library? I'll have a look

apeltzer:
Looks like they added JSON output in versions > 2.5, thus the library they use for this wasn't there in the (functional) 2.0.8 recipe. I'll have a look at the `json.hpp` file I found, this looks like it could be the reason for the breaking build.

apeltzer:
Its this probably causing the error:
```
// exclude unsupported compilers
#if defined(__clang__)
    #if (__clang_major__ * 10000 + __clang_minor__ * 100 + __clang_patchlevel__) < 30400
        #error "unsupported Clang version - see https://github.com/nlohmann/json#supported-compilers"
    #endif
#elif defined(__GNUC__)
    #if (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__) < 40900
        #error "unsupported GCC version - see https://github.com/nlohmann/json#supported-compilers"
    #endif
#endif
```
I couldn't find a simple way to add this to the `Cmakelists.txt` unfortunately - any idea @dpryan79 ?

dpryan79:
Don't know, cmake issues are always a pain :(

apeltzer:
Ok, I'll try setting up the circleci local client and will figure something out hopefully :-/

PR #11393, opened 2018-10-18
Bump IgBLAST to 1.9.0

marcelm:
Re-package the binary version this time. There are problems building the
source version when the configure script reaches this step:
"trying to build the NCBI SRA/VDB Toolkit from GitHub"

The previous problem about the binaries linking against libbz2.so, for
which we had no conda package, is now solved.

Also, this saves about 45 minutes compilation time.

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1886,7 +1886,6 @@ recipes/jali
 recipes/intervalstats
 
 # failing
-recipes/igblast
 recipes/igblast/1.4.0
 
 # missing fortran lib, will fix this upstream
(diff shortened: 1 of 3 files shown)


marcelm:
I’m giving up on the macOS build. Here’s the error (arises during testing), for reference:
```
+ igblastn -h
dyld: Library not loaded: /netopt/ncbi_tools/gnutls-3.4.0/Clang350-ReleaseMT64/lib/libgnutls.30.dylib
  Referenced from: /Users/distiller/project/miniconda/conda-bld/igblast_1539868286197/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/bin/../share/igblast/bin/igblastn
  Reason: image not found
```
I have no macOS development experience, but it appears as if a hard-coded path to the gnutls library is embedded into the binary. I had to repair a problem like this in the Linux binaries as well with `patchelf`, but have no idea how to do this in macOS.

PR #11440, opened 2018-10-20
gseapy bump version 0.9.8

zqfang:

### DIFF
--- a/recipes/gseapy/meta.yaml
+++ b/recipes/gseapy/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: gseapy
-  version: "0.9.7"
+  version: "0.9.8"
 
 source:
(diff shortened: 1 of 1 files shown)


recipes/gseapy/meta.yaml
```diff
@@ -3,8 +3,8 @@ package:
   version: "0.9.7"
 
 source:
-  url: https://files.pythonhosted.org/packages/d8/c3/0f0c932e37671d53eb00b825f23de9d8e489a72f50af1f1152fa9f41d457/gseapy-0.9.7.tar.gz
```
bgruening: This one is wried. We should at least bump the build string. But I would say such cases we should just skip with the auto-updater.

zqfang: the version number is wrong, corrected now

PR #11449, opened 2018-10-20
Pygtftk

guillaumecharbonnier:

This recipe is adding [pygtftk](https://github.com/dputhier/pygtftk).

@bioconda/core
I had to add this line ```ln -s $CC `dirname $CC`/gcc``` into the build.sh ([details](https://github.com/dputhier/pygtftk/issues/21)) to successfully compile but it does not look like something other packages need so maybe there is a more elegant way to do with the setup.py.


recipes/pygtftk/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - pandas >=0.23.3
+    - scipy >=1.1.0
+    - pybedtools >=0.7.8
+    - pybigwig >=0.3
+    - pyyaml >=3.12
+    - requests >=2.13.0
+    - cffi >=1.10.0
+    - pyparsing >=2.2.0
+    - biopython >=1.69
+    - zlib
```
bgruening: zlib can be removed here as it is a run_export nowadays.

recipes/pygtftk/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+  sha1: 23862a12605a02de001d540c4e2fa5688294d662
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - python >=3.6,<3.7
```
bgruening: can you `skip: True` builds instead of this pinning here

recipes/pygtftk/build.sh
```diff
@@ -0,0 +1,3 @@
+ln -s $CC `dirname $CC`/gcc
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
```
bgruening: pip install is recommend instead of setup.py usage

bgruening:
Thanks a lot!

recipes/pygtftk/build.sh
guillaumecharbonnier: I have tried using `script: python -m pip install --no-deps --ignore-installed --no-cache-dir -vvv .` in the meta.yaml but it fails for the same reason explained in my first message.
It works if I replace `$PYTHON setup.py install --single-version-externally-managed --record=record.txt`  by `$PYTHON -m pip install --no-deps --ignore-installed --no-cache-dir -vvv .` in build.sh, but I guess this is not the recommended way.

recipes/pygtftk/meta.yaml
guillaumecharbonnier: If I understand well, I need `skip: true # [not py35 and not py36]`, keep `- python` in host but still `- python >=3.5,<3.7` in run. Am I right? 
By the way, py35 is also ok hence the difference with my commit.

PR #11452, opened 2018-10-20
[BULK] Don't use r-base >=3.5, use a conda_build_config.yaml instead

dpryan79:

This has been breaking `bulk`. Ideally whatever is broken in the conda-build API will get fixed instead.

Alternatively, we could globally set an R 3.5.1 pinning.


dpryan79:
It at least gets past bioconductor-scnorm.

dpryan79:
@bgruening If you have a second today, please merge.

bgruening:
Mh, my fear is a little bit that we do not remember to clean up this hack for the next version.

bgruening:
Can we create an issue to track this? Not that we are good in tracking issues, but at least we have a chance :)

dpryan79:
Yeah, for the next release maybe we just change the pinning for the whole project...

Anyway, I'll create an issue.

msarahan:
@bgruening The problem here is basically the same as people see with python.  Specifying any kind of constraint for python or R breaks conda-build's handling of it.  The core languages are very special cases.  It is best to use

```
build:
  skip: True  # [some r version limit]
```

instead.  PRs to improve this part of conda-build would be most welcome.

PR #11453, opened 2018-10-20
Updating bioconductor packages

epruesse:
Updating all 86 Bioconductor packages with new versions at once as I assume they'll all build cleanly. (Hope that doesn't jinx this...)

### DIFF
--- a/recipes/bioconductor-absseq/meta.yaml
+++ b/recipes/bioconductor-absseq/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "1.32.0" %}
+{% set version = "1.32.3" %}
 {% set name = "ABSSeq" %}
 {% set bioc = "3.6" %}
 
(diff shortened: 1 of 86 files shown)


epruesse:
That went well... not.

Out of the 86 packages that have updates, only 2 are currently not blacklisted. Didn't picture the blacklist to be that large. @bgruening What to do with this?

bgruening:
@epruesse we already rebuild all R package on bulk and on bulk they are also not blacklisted anymore.
Also please have a look at the bioconductor updated on bulk. We decided to pin every bioc pkg to `>=x.y,<x.y+1` and a specific R version of this bioc release.

bgruening:
I will close this in favor of the bulk work which we already do.

@epruesse we might also consider splitting these kinds of updates into smaller chunks of the DAG. Such a PR will never succeed at once and will timeout. So either we commit those giant PRs (like ucsc ...) to bulk or we need to be smarter in chunkìng them up.

PR #11454, opened 2018-10-21
Update bioconda-utils to 0.14.9

epruesse:
This PR was created automatically. See bioconda/bioconda-utils#348


epruesse:
Accidental, but apparently we can self-update as well ;*)

bgruening:
Nice :)

PR #11464, opened 2018-10-21
Update cyvcf2 to 0.10.2

epruesse:
Update `cyvcf2`: **0.10.0** &rarr; **0.10.2**

***

[Package Info](https://bioconda.github.io/recipes/cyvcf2/README.html) | [Recipe Folder](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/cyvcf2) | [Upstream Home](https://github.com/brentp/cyvcf2) | [Upstream Releases](https://github.com/vladsaveliev/cyvcf2/releases)  

***

This pull request was automatically generated by [bioconda-utils](https://github.com/bioconda/bioconda-utils).


epruesse:
MacOS fails with

```
warning: cyvcf2/cyvcf2.pxd:26:16: Declarations should not be declared inline.
warning: cyvcf2/cyvcf2.pxd:188:14: 'BCF_HT_FLAG' redeclared 
warning: cyvcf2/cyvcf2.pxd:276:30: Declarations should not be declared inline.
error: command 'clang' failed with exit status 1
```

ping @vladsaveliev  - can you take a look at this?

vladsaveliev:
Hi @epruesse,

Looks like the `crypt` is not available on macos. Removing it helps on my mac and I don't see any other problems with the installation. Following up on the cyvcf2 issue: https://github.com/brentp/cyvcf2/issues/91

Vlad

brentp:
I just tagged v0.10.2 which makes the change suggested by @vladsaveliev please give that a try.
thanks for updating the recipe!

brentp:
cyvcf2 recently changed to include htslib as a submodule. 
not sure how to add that to the recipe: 
` git submodule update --init` should do it, but I thnk it downloads a tag...

vladsaveliev:
Right, `git submodule update --init` should help, but should we try to add `htslib` as a conda dependency? Will try

vladsaveliev:
Nope, `htslib` as a dependency errors:

```
ImportError: dlopen(/Users/vsaveliev/tmp/cyvcf2/miniconda/envs/cyvcf2_0102/lib/python3.6/site-packages/cyvcf2/cyvcf2.cpython-36m-darwin.so, 2): Symbol not found: _bcf_float_missing
```

By the way @brentp, the recipe downloads the release tarball from https://github.com/brentp/cyvcf2/releases which doesn't contain htslib source. Can switch to pulling a tag here, but you might want to add htslib source into the tarball anyway?

brentp:
> By the way @brentp, the recipe downloads the release tarball from https://github.com/brentp/cyvcf2/releases which doesn't contain htslib source. Can switch to pulling a tag here, but you might want to add htslib source into the tarball anyway?

that is automtically generated by github. is there a way to adjust it?

vladsaveliev:
Can't figure out a way to make GitHub clone recursively in releases... I think the only way is to generate a tarball with:

```
python setup.py sdist
```

which will go into `dist/cyvcf2-0.10.2.tar.gz`, and upload it instead

epruesse:
@brentp You can build and upload artifacts from Travis using the releases deployment provider. See e.g. https://github.com/epruesse/SINA/blob/master/.travis.yml

I’m not sure about htslib as part of your sources. It seems common, but not really pretty.  Can’t we pull that in as a package?

brentp:
I include htslib to ease installation outside of conda, so it would probably be best to use the bioconda version of conda if possible. I'm happy to help facilitate that if I can, but haven't had time to dig into improving the packaging on my end.

vladsaveliev:
Hmm, wondering why CI has stopped running after new commits.

bgruening:
This can happen once in a while. Hope it works now. Thanks!

epruesse:
It’s been happening a lot lately. In particular when I auto bump. I can’t find anything in the circle API clearly meant to trigger a build. Technically, I could have something check whether the lint is pending and automate the close-reopen. Not very pretty though.

vladsaveliev:
My version that's because we are kind of installing htslib twice: first explicitly with configure/make, then implicitly when linking cyvcf2 source together with htslib within `setup.py`. But if I drop htslib installation from build.sh, it'll still fail at some point.

ccwang002:
I tried a different approach in PR #13726 by patching the `setup.py` to link to the conda htslib instead of compiling one inside cyvcf2. All the tests passed for 0.10.8 and lets see if this approach can work for the future releases as well.

PR #11469, opened 2018-10-21
FRED2 as a conda package

apeltzer:


recipes/fred/meta.yaml
```diff
@@ -0,0 +1,49 @@
+{% set name = "fred" %}
```
bgruening: should that be fred2?

recipes/fred/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - python -c "from Fred2.Core import Allele, Peptide, Protein,generate_peptides_from_proteins"
+
+about:
+  home: http://nf-co.re/
+  license: MIT
+  summary: 'Python package with helper tools for the nf-core community.'
+
+extra:
+  container:
+    # click requires a unicode locale when used with Python 3
```
bgruening: But you are not using click here isn't it?

recipes/fred/meta.yaml
apeltzer: Hm, could be - should I rename it ?

recipes/fred/meta.yaml
apeltzer: Whoops, thats a relict of my template. I'll remove it.

apeltzer: Changed it, renamed the package as well to fred2

PR #11481, opened 2018-10-22
Add vcf-annotator recipe

rpetit3:

### DIFF
--- a/recipes/vcf-annotator/build.sh
+++ b/recipes/vcf-annotator/build.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+mkdir -p $PREFIX/bin
+pip install PyVCF==0.6.8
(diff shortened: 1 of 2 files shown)


recipes/vcf-annotator/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+mkdir -p $PREFIX/bin
+pip install PyVCF==0.6.8
```
bgruening: please depend on pyvcf instead.

rpetit3: Thank you for the update. Should I submit a recipe for PyVCF to bioconda?

rpetit3: Disregard that, just saw pyvcf was moved to conda-forge, will update

PR #11506, opened 2018-10-23
Update SVDB to version 1.3.0

J35P312:
renamed:    reipes/svdb/svdb.patch -> recipes/svdb/1.1.2/svdb.patch
	modified:   recipes/svdb/meta.yaml


epruesse:
@J35P312 Thank you!

We don't need the 1.1.2 directory though - could you delete it?

Background: those sub directories (which you've seen in other recipes) exist primarily to be able to build  e.g. 1.1.3 **after** 1.3.0 has been build so that we don't jump back and forth on the primary meta.yaml. There was a little confusion in the beginning, leading to plenty of those sub-recipes lying about which we do not actually need.

J35P312:
Hello there!
Thanks for the background! I have deleted the 1.1.2 directory now.
Have a nice weekend!
//Jesper

PR #11553, opened 2018-10-24
Add multicore-tsne

grst:

Adds a recipe for [multicore-TSNE](https://github.com/DmitryUlyanov/Multicore-TSNE/), which is an accelerated TSNE implementation for single cell analysis.

### DIFF
--- a/recipes/multicore-tsne/meta.yaml
+++ b/recipes/multicore-tsne/meta.yaml
@@ -33,7 +33,7 @@ test:
   source_files:
     - MulticoreTSNE/tests/test_base.py
   commands:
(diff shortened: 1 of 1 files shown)


bgruening:
@grst I think this package should go to conda-forge as it is not strictly bio* related, isn't it?

grst:
I believe it's only used for single cell RNA-seq and is a (optional) dependency of e.g. [scanpy](https://github.com/theislab/scanpy/) or [scanorama](https://github.com/brianhie/scanorama)

bgruening:
TSNE is a very general algorithm and other implementations are already on conda-forge as fasttsne for example.

grst:
of, fair enough.

PR #11570, opened 2018-10-25
Fix and unblacklist TwoPaCo

npavlovikj:


build-fail-blacklist
```diff
@@ -147,7 +147,7 @@ recipes/spingo
 #cd /opt/conda/conda-bld/twopaco_1529606958569/work/build/graphdump && /opt/rh/devtoolset-2/root/usr/bin/g++   -I/opt/conda/conda-bld/twopaco_1529606958569/work/src/graphdump/../common  -DBOOST_MATH_DISABLE_FLOAT128 -m64 -fPIC;-std=c++0x -O3 -DNDEBUG   -o CMakeFiles/graphdump.dir/graphdump.cpp.o -c /opt/conda/conda-bld/twopaco_1529606958569/work/src/graphdump/graphdump.cpp
 #g++: fatal error: no input files
 
-recipes/twopaco
+# recipes/twopaco
```
epruesse: you can just delete here, rather than comment the line
```suggestion
```

recipes/twopaco/meta.yaml
```diff
@@ -1,27 +1,32 @@
...
 requirements:
   build:
     - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
     - cmake
-    - tbb
-
   run:
     - tbb
+    - tbb-devel
```
epruesse: Probably don't need `tbb-devel` as runtime dependency
```suggestion
```

recipes/twopaco/meta.yaml
```diff
@@ -1,27 +1,32 @@
...
+  host:
+    - tbb
```
epruesse: Pretty sure `tbb-devel` requires `tbb`  (actually, I think they are built from the same recipe at CF)
```suggestion
```

PR #11616, opened 2018-10-26
Update ShoRAH recipe to version 1.9.95

davebx:


recipes/shorah/1.9.95/conda_build_config.yaml
```diff
@@ -0,0 +1,2 @@
+blas:
+  - 1.1 openblas
```
bgruening: If you use this, I don't think you need {{ blas }} anymore.

PR #11701, opened 2018-10-29
Add recipe for transit version 2.3.1

davebx:
* [x] This recipe **is directly relevant to the biological sciences**

### DIFF
--- a/recipes/transit/meta.yaml
+++ b/recipes/transit/meta.yaml
@@ -1,0 +1,46 @@
+{% set name = "transit" %}
+{% set version = "2.3.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/mad-lab/{{ name }}/archive/v{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


recipes/transit/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  url: https://github.com/mad-lab/{{ name }}/archive/v{{ version }}.tar.gz
+  sha256: ca2dd3543789acf5f4f7327ddac0a27b7ddb2ff7966959a148a84382ccddb486
+
+build:
+  number: 0
+  script: "python setup.py install --single-version-externally-managed --record=record.txt"
+
+requirements:
+  build:
+    - python==2.7.*
```
bgruening: please skip the build instead of pin it like that

recipes/transit/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - numpy>=1.15
```
bgruening: you probably need a space here

Issue #1314, opened 2016-04-13
OS X stdlib future direction and conda-recipes compatibility for C++ packages


yesimon:
**Warning: Only relevant to C++ packages**

I talked about this briefly in #393, and have a fuller picture of OS X build environment in conda/conda-recipes#505.

They are linking against libstdc++ and setting a minimum OS X version of 10.7. Meanwhile, the homebrew package manager is linking against libc++ and requiring minimum OS X 10.9. I think that keeping in sync with conda-recipes might be problematic for OS X, since libstdc++ is now unsupported and now starting to lack many recent features in C++11/14 and beyond, which developers are now starting to utilize in their applications. 

I don't think having a 10.9 version minimum is unreasonable since OS X updates are now free. For a scientific software perspective, we should target more compatibility on the developer coding side than old operating system side of things.

Concretely, for the most part the conda-recipes packages should present no problems as long as they are binary packages, although you might be able to gain a bit from compiling with `MACOSX_DEPLOYMENT_TARGET=10.9`. For library packages, I propose rebuilding all with `libc++` to standardize our base.

johanneskoester:
I would like to delay changes to the minimum dependencies as long as possible. I have seen many old osx installations on peoples desktops. Especially if bioconda shall be used by non-experts, requiring an os update for installing a package is a big hurdle. Maybe we should use this thread to collect a list of packages that won't build on the old setup. If that list contains enough important packages, we switch?

yesimon:
Let's maintain max compatibility with and closely follow `conda-recipes`.

Similarly to #1332, I would strongly support a separate channel for newer OS X supporting 10.9 (what's available on travis)., `libc`, and newer C++ features because there are already some cases where `libstdc++` is not new enough.

johanneskoester:
Having a separate channel is problematic. I believe our funding from Continuum is only for the primary bioconda channel. I wonder if there is a way to have such packages in the main channel, and have them automatically selected if a newer osx is available... We could use the "feature" functionality for that, similar to the mkl builds of numpy in the main channel.

epruesse:
Closing - outdated.

PR #11702, opened 2018-10-30
Restores vsearch builds on OSX

colinbrislawn:

Work in Progress!
Restores vsearch builds on OSX. Related to #9404

### DIFF
--- a/recipes/vsearch/build.sh
+++ b/recipes/vsearch/build.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -euo pipefail
 
+CFLAGS="-lm" 
+
 ./autogen.sh
 ./configure --prefix=$PREFIX
 make
--- a/recipes/vsearch/meta.yaml
+++ b/recipes/vsearch/meta.yaml
@@ -10,8 +10,7 @@ source:
   sha256: '{{ sha256 }}'
 
 build:
-  number: 0
-  skip: true # [osx]
+  number: 1
 
 requirements:
   build:


dpryan79:
You can try setting `CFLAGS` to `-lm` and see if that helps.

epruesse:
@torognes Can you provide any help here?

This looks almost like some kind of STL linking issue. Have you seen this before?

This came up in #12918 again as well.

colinbrislawn:
Any updates? Who is getting builds working on OSX?

torognes:
Sorry for the late reply.

It appears that the problem is related to linking to the C++ standard library. On macOS, VSEARCH is currently usually built using the older and depreciated `libstdc++` library and not the newer and recommended `libc++` library from Apple.

I have looked at the test build report on macOS and compared it to building it on my local mac (macOS Mojave 10.14.3). The test build is made using Xcode version 8.3.3 while I use the newer Xcode version 10.1.  The test build is using the compiler/linker command `clang++` instead of `c++`, but it seems like `c++` is just a softlink to `clang++` anyway, so it probably doesn't matter.

I get the same error messages as in the test build if I add the option `-stdlib=libc++` to the linker command line. It works well if I rather specify `-stdlib=libstdc++` on the command line (or nothing). (Except for the usual warning `clang: warning: libstdc++ is deprecated; move to libc++ [-Wdeprecated]`)

The default C++ standard library for clang++ may have changed at some point from `libstdc++` to `libc++`, probably from macOS 10.9.

The `configure.ac` file of VSEARCH specifies `MACOSX_DEPLOYMENT_TARGET="10.7"` that should indicate the use of the older C++ standard library, but perhaps this is not taken into account for some reason.

If possible, building with the `-stdlib=libstdc++` option to the linker could potentially solve the problem. I could alternatively add this option to the upstream vsearch Makefile.

The best solution may be to change MACOSX_DEPLOYMENT_TARGET to 10.9 (OSX Mavericks from October 2013) in VSEARCH, which may be about time, as extremely few users should be using something older.

See also issue #1314.

dpryan79:
Xref #13931

torognes:
It seems like the change to `MACOSX_DEPLOYMENT_TARGET="10.9"` did the trick. I'll make that change in the next VSEARCH version, coming soon.

dpryan79:
This has since been done in another PR, so I'll close this.

PR #11706, opened 2018-10-30
[BULK] update 30.10.18

dpryan79:

Removes a number of packages from the blacklist. I added these previously because of the issues on conda-forge staged-recipes, but everything exists now.

It's quite possible that everything except bioconductor-mzr will be finished today. There are a couple build issues on OSX and we should discuss exactly how those should be dealt with (namely, do we blacklist them or just leave them be and make a PR on bulk when/if they can be fixed).

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -789,6 +789,7 @@ recipes/r-intego
 recipes/r-lme4qtl
 recipes/r-mmgenome
 recipes/r-nastiseq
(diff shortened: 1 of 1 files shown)


dpryan79:
There's no particular rush on this PR, bulk can run for another 7-10 hours before it runs out of packages to build.

dpryan79:
@bgruening When you have a chance, please merge.

PR #11727, opened 2018-10-31
update frogsv2.0.1: Manage rdptools all versions

ValentinMarcon:

(See also https://github.com/oinizan/FROGS/commit/a3fbce06cf1e021248180aa312d1c098fdd8b679 and https://github.com/oinizan/FROGS/commit/d737609359bc1f433f2155f23739c125871b1287 to understand v2.0.1b of FROGS)

### DIFF
--- a/recipes/frogs/build.sh
+++ b/recipes/frogs/build.sh
@@ -11,7 +11,6 @@ mkdir -p $BINARY_HOME
 mkdir -p $LIB_HOME
 mkdir -p $LIBEXEC_HOME
 
-
 mv $SRC_DIR/* $FROGS_HOME
 
 #Make sure all css and js requests are https so reports work on https servers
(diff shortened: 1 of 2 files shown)


ValentinMarcon:
I dont think so. 
Because our recipe dont work anymore. 
In fact, their is a new version of rdptools, our old recipe will install it (rdptools >2.1.2 on meta. yaml) and exist

bgruening:
@ValentinMarcon but then we should fix the old recipe and not overwrite it with the new version isn't it?
You can fix the old recipe by pinning rdptools to `<2.1.2`.

ValentinMarcon:
I'm sorry, I'm not sure to understand. 

Is the version 2.0.1b of FROGS archive that is a problem?

We would like to have these corrections (rdptools and FROGS 2.0.1b patch) installed when we choose to install FROGS 2.0.1 on Bioconda. That is to replace the older version (not working anymore) by this one, still 2.0.1.

dpryan79:
No longer needed

PR #11733, opened 2018-10-31
Added sortmerna 3

eggzilla:


recipes/sortmerna/3.0.0/build.sh
```diff
@@ -0,0 +1,4 @@
+mkdir -pv $PREFIX/bin
+bash build.sh
```
bgruening: It seems there is no build.sh file in the source dir?

recipes/sortmerna/3.0.0/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+  host:
+    - zlib
+  run:
+    - zlib
```
bgruening: zlib is exporting themself now, this is a new feature of the zlib package, so ideally you do not need to specify it anymore in the run dependency

recipes/sortmerna/3.0.0/build.sh
eggzilla: yeah, building now depends on cmake, i will make the necessary modifications :-)

recipes/sortmerna/3.0.0/meta.yaml
eggzilla: removed, thanks for reviewing :-)

daler:

PR #11737, opened 2018-11-01
adding the new recipe flowcatchR

federicomarini:

### DIFF
--- a/recipes/bioconductor-flowcatchr/meta.yaml
+++ b/recipes/bioconductor-flowcatchr/meta.yaml
@@ -1,0 +1,42 @@
+{% set version = "1.16.0" %}
+{% set name = "flowcatchR" %}
+{% set bioc = "3.8" %}
+
+package:
+  name: 'bioconductor-{{ name|lower }}'
+  version: '{{ version }}'
+source:
+  url:
+    - 'http://bioconductor.org/packages/{{ bioc }}/bioc/src/contrib/{{ name }}_{{ version }}.tar.gz'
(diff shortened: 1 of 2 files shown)


recipes/bioconductor-flowcatchr/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - 'https://depot.galaxyproject.org/software/bioconductor-{{ name|lower }}/bioconductor-{{ name|lower }}_{{ version }}_src_all.tar.gz'
+  sha256: 3f3880f090f9889a6c4eeb78ce19625ffbaa5825a8bb0faa9bdb1e3d1aaa8e33
+build:
+  number: 0
+  rpaths:
+    - lib/R/lib/
+    - lib/
+requirements:
+  host:
+    - bioconductor-biocparallel
```
bgruening: add pinnings here

federicomarini: Naive question, but do the specified packages also need to receive an updated recipe? (my guess is yes)

bgruening: Yes :) That's why we rebuild all packages at once. Otherwise, you mix versions of different bioc versions. That might work or might not work ... no one knows.

federicomarini: Are the existing packages already under a bulk-rebuilding process after 3.8 release (Bioc), or do they need manual updates?

federicomarini:
Same as for https://github.com/bioconda/bioconda-recipes/pull/11735.
@bgruening @dpryan79 I guess this one is to close as well, I see the recipe currently built 👍

PR #11741, opened 2018-11-01
Macs2 update

jenzopr:

I rewrote the recipe using `conda skeleton pypi --python-version 2.7 macs`.


dpryan79:
Does 2.1.2 no longer need R for producing PDFs? The older versions needed that, which is why R is was a runtime dependency.

recipes/macs2/meta.yaml
```diff
@@ -1,31 +1,39 @@
...
-  number: 3
+  number: 0
   skip: True  # [py3k]
+  script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed -vvv "
 
 requirements:
-  build:
-    - {{ compiler('c') }}
   host:
+    - numpy >=1.6
```
bgruening: not sure numpy is needed here.

recipes/macs2/meta.yaml
```diff
@@ -1,31 +1,39 @@
...
     - python
-    - numpy
+    - pip
   run:
+    - numpy >=1.6
     - python
-    - numpy
-    - r-base
```
bgruening: same question as @dpryan79 ... r-base is probably needed isn't it?

recipes/macs2/meta.yaml
```diff
@@ -1,31 +1,39 @@
...
+    - macs2 -h
 
 about:
-  home: https://github.com/taoliu/MACS/
-  license: BSD New
-  summary: MACS -- Model-based Analysis of ChIP-Seq
+  home: http://github.com/taoliu/MACS/
+  license: BSD
+  license_family: BSD
+  license_file:
```
bgruening: COPYING is the license file

recipes/macs2/meta.yaml
jenzopr: Not needed *per se*, the `README` states that a call to `RScript` is necessary. We can, however, include r-base. I also checked the generated Rscript, there are no other dependencies. Let me know :-)

recipes/macs2/meta.yaml
jenzopr: Thanks! I missed that one..

recipes/macs2/meta.yaml
jenzopr: Probably true, I will remove it from the `host` section.

recipes/macs2/meta.yaml
dpryan79: I used to see a LOT of errors in Galaxy due to R not being present and users specifying to produce a PDF. I expect this happens with users on the CLI too, so I'd very much be in favor of including r-base.

jenzopr:
Weird. All tests were successful, just the last `cp: cannot stat /opt/conda/conda-bld/linux-64/macs2-2.1.2-py27_0.tar.bz2: No such file or directory` command failed. 
Can someone rerun the test?

jenzopr:
Could increasing the build number help?

bgruening:
reruning! Thanks @jenzopr!

dpryan79:
It looks like this is caused by having `r-base` in the `run` but not `host` section. In essence, we need the equivalent of `noarch: python` for R. The work-around is to just add `r-base` to `host`, even though it's not actually needed there.

jenzopr:
Ping @bgruening @dpryan79

jenzopr:
Great hint, @dpryan79 ! Thanks a lot 🎉

dpryan79:
Glad that took care of it, great work!

PR #11767, opened 2018-11-02
add new recipe for wgfast

tfursten:


recipes/wgfast/install.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash -e -x
+cd ext
+wget https://sourceforge.net/projects/bbmap/files/BBMap_38.08.tar.gz
+tar -xzf BBMap_38.08.tar.gz
+chmod +x ./bbmap/bbduk.sh
+
```
npavlovikj: both "bbmap" and "raxml" are available as bioconda packages, so you should add them as host/run dependencies instead of installing them from scratch here. Also, the bioconda "raxml" package contains "raxmlHPC" and "raxmlHPC-AVX" - if you specifically need the SSE3 executables, you will probably need to modify the current "raxml" recipe.

tfursten: @npavlovikj  I have added bbmap and raxml as run dependencies. As far as I can tell, the raxml package provides the necessary executables in /bin/. I have also added a recipe for mtsv which depends on wgfast.

npavlovikj: Thanks for updating the recipe @tfursten ! 

Since "mtsv" depends on "wgfast", this recipe won't be built because "wgfast" is not yet available in "bioconda". Therefore, please remove the "mtsv" recipe from this PR. Once "wgfast" is merged, you can create a new PR only for "mtsv".

For "wgfasat", it is a good practice to include the LICENCE information in the recipe as well, https://bioconda.github.io/guidelines.html, to show that redistribution is allowed. Since you are the developer of the tool, I would suggest you to include a licence in the distribution, and include the "license", "license_family" and "license_file" sections in meta.yaml. You might find this link useful for that, https://help.github.com/articles/adding-a-license-to-a-repository/.

Also, if you check the Linux and OSX build logs, they say "BIOCONDA INFO Nothing to be done.". I am really not sure why this happens since all your requirements are available and not blacklisted. However, let's first remove "mtsv" and see if the recipe gets build afterwards.

npavlovikj:
Hey @tfursten, I don't know why your recipe build is pending, so can you please rebase your branch with master and re-trigger another build?

recipes/wgfast/install.sh
tfursten: Thanks @npavlovikj! It looks like the checks have passed

npavlovikj: That is great @tfursten! Before I merge the request, I have one more suggestion :) 
When installing pure python packages, the practice is to use "python -m pip install" instead of "python setup.py install". Also, since the build script has only one line, you can completely remove "build.sh" and the add the script in "meta.yaml" and add "pip" in "host". Please see https://github.com/conda-forge/staged-recipes/blob/master/recipes/example/meta.yaml for an example of this.

tfursten: I have made these changes @npavlovikj

npavlovikj: Everything looks great @tfursten - thank you for contributing this recipe!

PR #11782, opened 2018-11-05
ARB: Pin glib

epruesse:

@bgruening - I hate to ask, but the previous builds need to be removed from the channel. Otherwise `conda` will install glib `2.55.1` while `2.56.1` is needed (just because it was linked to 2.56, not because of ARB).  It breaks on OSX, Linux is more tolerant I think.

I've created issue conda-forge/glib-feedstock#38 and PR conda-forge/glib-feedstock#37 at CF to add `run_exports` to `glib`.

### DIFF
--- a/recipes/arb-bio/meta.yaml
+++ b/recipes/arb-bio/meta.yaml
@@ -25,7 +25,7 @@ source:
     - gsed.patch
 
 build:
-  number: 4
+  number: 5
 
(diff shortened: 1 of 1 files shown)


epruesse:
**sigh**

Why does utils think this has noarch set????

epruesse:
fingers crossed...

PR #11802, opened 2018-11-06
bump pylibseq

molpopgen:
* [] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).

### DIFF
--- a/recipes/pylibseq/meta.yaml
+++ b/recipes/pylibseq/meta.yaml
@@ -21,6 +21,8 @@ requirements:
     - libsequence ==1.9.6
     - pybind11 >=2.2.3
     - msprime >=0.6.0
+    - svgwrite # for msprime
+
 
   run:
     - python
--- a/recipes/pylibseq/meta.yaml
+++ b/recipes/pylibseq/meta.yaml
@@ -21,6 +21,7 @@ requirements:
     - libsequence ==1.9.6
     - pybind11 >=2.2.3
     - msprime >=0.6.0
+    - openblas # workaround for current msprime issue
     - svgwrite # for msprime
 
 
@@ -28,6 +29,8 @@ requirements:
     - python
     - libsequence
     - msprime >=0.6.0
+    - openblas # See above
+    - svgwrite # See above
 
 test:
   # Python imports
--- a/recipes/pylibseq/meta.yaml
+++ b/recipes/pylibseq/meta.yaml
@@ -1,24 +1,24 @@
 package:
   name: pylibseq
-  version: "0.2.0"
+  version: "0.2.1"
 
 source:
-  url: https://files.pythonhosted.org/packages/ac/22/06172cf9ff7e62a9c83cd8574c3994acaf0d3e16d9472f236e4a71a9692b/pylibseq-0.2.0.tar.gz
-  sha256: 04c9ba91e79876c6c31229eeda70e49600afb81fa78270ae4f6f58b847b8b72f
+  url: https://files.pythonhosted.org/packages/86/f9/0da11046680d74ceba0aade1bca859951479eacc4bda2bf062a568216cb4/pylibseq-0.2.1.tar.gz 
+  sha256: 6bfc509f90832f38e9c9ca862b8f57e35a68ba2b218a1092e3c18affef41f6dd
 
 build:
-  skip: True # [py27]
+  skip: True # [py27] [osx]
   # If this is a new build for the same version, increment the build
   # number. If you do not include this key, it defaults to 0.
-  number: 2
+  number: 1
 
 requirements:
   build:
     - {{ compiler('cxx') }}
   host:
     - python
-    - libsequence ==1.9.3
-    - pybind11 >=2.2.0
+    - libsequence ==1.9.6
+    - pybind11 >=2.2.3
 
   run:
     - python
--- a/recipes/pylibseq/meta.yaml
+++ b/recipes/pylibseq/meta.yaml
@@ -20,17 +20,11 @@ requirements:
     - python
     - libsequence ==1.9.6
     - pybind11 >=2.2.3
-    - msprime >=0.6.0
-    - openblas # workaround for current msprime issue
-    - svgwrite # for msprime
 
 
   run:
     - python
     - libsequence
-    - msprime >=0.6.0
-    - openblas # See above
-    - svgwrite # See above
 
 test:
   # Python imports
--- a/recipes/pylibseq/meta.yaml
+++ b/recipes/pylibseq/meta.yaml
@@ -7,7 +7,8 @@ source:
   sha256: 6bfc509f90832f38e9c9ca862b8f57e35a68ba2b218a1092e3c18affef41f6dd
 
 build:
-  skip: True # [py27] [osx]
+  skip: True # [py27] 
+  skip: True # [osx]
   # If this is a new build for the same version, increment the build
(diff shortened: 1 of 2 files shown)


molpopgen:
@bgruening It seems like something odd is happening here.  There is a complaint about svgwrite and trying to install via setuptools.  But, that doesn't seem write.  svgwrite is a dependency of msprime, which is installed in the meta.yaml and svgwrite is in conda-forge.  The recipe for msprime does show that svgwrite is a run dependency: https://github.com/conda-forge/msprime-feedstock/blob/master/recipe/meta.yaml

recipes/pylibseq/meta.yaml
```diff
@@ -1,29 +1,31 @@
...
+  sha256: 6bfc509f90832f38e9c9ca862b8f57e35a68ba2b218a1092e3c18affef41f6dd
 
 build:
-  skip: True # [py27]
+  skip: True # [py27] 
+  skip: True # [osx]
   # If this is a new build for the same version, increment the build
   # number. If you do not include this key, it defaults to 0.
-  number: 2
+  number: 1
```
bgruening: this should be 0

molpopgen: Thanks. I'll change the build number.

svgwrite is a dependency of msprime, which is a listed dependency of my project.

molpopgen: @bgruening This is the error being reported:

```
RuntimeError: Setuptools downloading is disabled in conda build. Be sure to add all dependencies in the meta.yaml  url=https://pypi.org/simple/svgwrite/
```

bgruening:
circles everywhere :)

molpopgen:
@bgruening @bioconda/core 
Any ideas on how to debug this?  The build system is claiming that I am attempting to install svgwrite via setuptools, which is not the case.

svgwrite is a dependency of msprime, which is in turn a dependency of pylibseq.

The relevant section of the meta.yaml for pylibseq is currently:

```
  host:
    - python
    - libsequence ==1.9.6
    - pybind11 >=2.2.3
    - msprime >=0.6.0
    - svgwrite # for msprime, but should not be needed


  run:
    - python
    - libsequence
    - msprime >=0.6.0
```

bgruening:
@molpopgen `- svgwrite`is missing from the run section of `pylibseq` as it seems.

molpopgen:
> @molpopgen `- svgwrite`is missing from the run section of `pylibseq` as it seems.

Do you think that will matter?  This error is occurring even if svgwrite is not listed as a build dependency (you can see previous build logs). svgwrite is being pulled in via marking msprime as a depdendency.

bgruening:
There is something fishy not in your repo but in msprime etc, this comment does not look good ` - svgwrite # for msprime, but should not be needed` :) So I think we should fix that at first and hopefully your problem goes away with it as well.

molpopgen:
> There is something fishy not in your repo but in msprime etc, this comment does not look good ` - svgwrite # for msprime, but should not be needed` :) So I think we should fix that at first and hopefully your problem goes away with it as well.

@bgruening OK, so if I remove that, we will be back to a failing build.  I'll ping the msprime author, who is also the maintainer of the conda-forge recipe.

bgruening:
@molpopgen you can also create a PR against the conda-forge one if you like.

molpopgen:
@bgruening I've been in contact with the msprime recipe maintainer, who thinks that some of the problem is due to a massive rebuild taking place on conda-forge.  I seem unable to restart failed jobs.  I tried following your instructions from [here](https://github.com/bioconda/bioconda-recipes/pull/10210#issuecomment-410089121), but I do not have permission it seems?

bgruening:
I restarted the tests for you. You should have permissions. What do you see on top of circle-cI?

molpopgen:
> I restarted the tests for you. You should have permissions. What do you see on top of circle-cI?

The "rerun workflow" button is grayed out and says "You need write permissions..." when I mouse over it.

When I click the "bioconda-workflow" link, I see nothing obvious about how to restart.

I am not logged in to Circle CI in any way--could that be the issue?  I've never used it outside of these PRs.

molpopgen:
It does seem to be the logging in issue.  I just logged in via GitHub and can now see the restart button is activated.  Duh!

molpopgen:
@bgruening This is the puzzling thing from the build log.  After building, there is another scan for dependencies.  I'm unable to track down why this is happening.  Shouldn't all dependency resolution have already happened?  In fact, you can see in the CircleCI logs that everything has been installed, including the svgwrite that is being complained about.

```
Installed /opt/conda/conda-bld/pylibseq_1543365247619/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/python3.5/site-packages/pylibseq-0.2.1-py3.5-linux-x86_64.egg

Processing dependencies for pylibseq==0.2.1

Searching for svgwrite

  File "setup.py", line 193, in <module>

    zip_safe=False,

  File "/opt/conda/conda-bld/pylibseq_1543365247619/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/python3.5/site-packages/setuptools/__init__.py", line 140, in setup
```

molpopgen:
@jeromekelleher --Yes, I think I will try removing the build dependency.

Regarding openblas.  Conda-forge made an odd decision, IMO.  The compile GSL against openblas, but do so by deleting libgslcblas and using a symlink to openblas as a replacement. See [here](https://github.com/conda-forge/gsl-feedstock/blob/master/recipe/build.sh).  This "works" but is not how GSL recommend using an alternate blas lib.  We ran into issue with this when updating the minimum GSL in bioconda.

jeromekelleher:
> Regarding openblas. Conda-forge made an odd decision, IMO. The compile GSL against openblas, but do so by deleting libgslcblas and using a symlink to openblas as a replacement. See here. This "works" but is not how GSL recommend using an alternate blas lib. We ran into issue with this when updating the minimum GSL in bioconda.

This is what's blocking me on the updated msprime package:
https://travis-ci.org/conda-forge/msprime-feedstock/jobs/459904966#L611

molpopgen:
@jeromekelleher unfortunately removing msprime as a build dependency did not solve the problem

jeromekelleher:
> @jeromekelleher unfortunately removing msprime as a build dependency did not solve the problem

<headscratch emoji> No idea what's going on here @molpopgen, sorry.

molpopgen:
I think I may remove it as both build and run dependencies.  Doing so is
technically correct.  There is never an attempted import of msprime except
in one C++ function, and that will error out with an exception if msprime
is not installed, which is correct behavior.
__
Kevin Thornton
Associate Professor
Ecology and Evolutionary Biology
UC Irvine
www.molpopgen.org
github.com/molpopgen
github.com/ThorntonLab

jeromekelleher:
> I think I may remove it as both build and run dependencies.  

Sounds like a plan.

molpopgen:
Closing.  To remove msprime as a dependency, `setup.py` needs fixing, so I need to push a new version to PyPi first and try again.

Issue #2354, opened 2016-09-19
gcc: compilation optimization march/mtune level


yesimon:
I'm opening this issue so that we can standardize on instruction level support. Typically compilation with gcc will default to `-march=x86-64` which is the most generic x86 support available and should work on all processors. However, newer processors have newer instruction sets such as SSSE3, AVX, etc that are only supported on a subset of processors.

Some packages have Makefiles that include parameters such as `-march=native`, which optimizes for the current architecture but may only work on the cpu level that the package was compiled on (in our case whatever Travis provides), but this software wouldn't work on older cpus.

One approach would be to standardize on a level such as `-march=x86-64` - this precludes optimization in favor of compatibility and is how OS system packages are built. For OS X, Homebrew standardizes on `-march=core2` which I think is around the time that Apple switched to Intel so that level of support should be universal.

Another approach may be to add yet more to the build string suffix. In this case, there can be suffixes for differing levels of CPU  such as `sandybridge` or `skylake` https://gcc.gnu.org/onlinedocs/gcc-6.2.0/gcc/x86-Options.html#x86-Options

kyleabeauchamp:
We may wish to check in with conda-forge and Continuum about best practices.

Clearly a lot of bioinformatics can benefit from having some level of optimization so it would be great if we had systems in place to maximize performance.

yesimon:
Another approach, such as by RAxML is to compile multiple binaries with suffixes in a single package: 
https://github.com/bioconda/bioconda-recipes/blob/master/recipes/raxml/build.sh

kyleabeauchamp:
As food for thought, intel has a whitepaper that claims having AVX instructions may improve BWA:

http://www.intel.com/content/www/us/en/healthcare-it/solutions/documents/performance-benchmarking-of-burrow-wheeler-aligner.html

I'm not sure if their results apply to the current versions of BWA, but I for one would be thrilled to have faster alignment.

kyleabeauchamp:
Looks like this might be in a non-standard branch of bwa, sadly: https://sourceforge.net/p/bio-bwa/mailman/message/34618535/

yesimon:
Yes the whitepaper is very misleading by because it claims bwa was compiled with gcc flags `-O2 -funroll-loops`, which should still default to `-march=generic`. It's also comparing two completely different CPUs.

I have actually tested compiling with `icc` with ipo optimizations and even then the performance difference was negligible (0-5% improvement).

[...]

PR #11822, opened 2018-11-08
Fix octopus binary "invalid instruction" (-march too new)

holtgrewe:

@chapmanb @kyleabeauchamp What do you think?


holtgrewe:
@chapmanb Ah, sorry for not giving much context here. We are hitting an issue here with that the binaries for octopus actually require a Haswell architecture (one or two generations after Sandybridge) and we get an "invalid instruction" error on our Sandybridge cluster nodes. At another point, I asked about policies of this but there really was none somehow.

I think #2354 contains some relevant discussion and links. Maybe @kyleabeauchamp can give some insights into this. I would be fine with "going back" further or even somehow shipping multiple architectures at the same time.

Right now, the package is broken for us and I could imagine that this is the case for people who have even older cluster nodes (e.g., high memory nodes tend to be kept alive longer than the average compute node as they are pricy).

chapmanb:
Manuel -- thanks much for the context. I'd definitely be keen to make this as portable as possible. I don't know enough about this to comment usefully, but it would be helpful to have @dancooke in on the conversation. Dan, we're working on making the octopus binary more portable. Do you know the potential impact of compiling against older or multiple architectures on runtime? Any tips/tricks for us doing this in the best way?

holtgrewe:
@epruesse I agree. But how do we achieve this in Bioconda for this package in a short timeframe?

epruesse:
@holtgrewe Just add a patch removing the `-march=native` bit from `src/CMakeLists.txt` and open an issue upstream, notifying them that their `Release` builds code that won't run on a CPU older than the build machine.

epruesse:
@dancooke All 64 bit CPUs have SSE2, so we always get that. Depending on what's being done, SSE4.2 brings significant extra performance, while AVX and AVX2 aren't worth it in my experience. AVX-512 again brings a lot, but it's actually a collection of extensions of which CPUs only implement some.

Options:
 - follow the `RAxML` approach and compile a bunch of variants of the binary. 
 - go above that and have a wrapper script selecting the right one. 
 - go the `vsearch` route and do this inside of your binary with hand optimized SIMD instructions (although `vsearch` doesn't get this 100% right and breaks on a small subset of the Travis nodes - testing is really hard). 
 - build the compute kernels in a separate DSO, build that multiple times for a range of `-march` targets and `dlopen` the right one at run time
- create a channel, or multiple channels, of your own, in which you place copies of the packages built for specific architectures. You can then layer the appropriate channel above Bioconda to override packages with custom optimized ones.
  Bear in mind though, that this needs a separate miniconda folder for each CPU architecture.

epruesse:
There you go. Let's see if that works.

lh3:
Intel developers suggest that the right strategy is to use the so-called [CPU dispatch](https://attractivechaos.wordpress.com/2017/09/04/on-cpu-dispatch/). The idea is to compile different versions of the same functionality at the compile time. During run time, the program dynamically chooses the fastest version based on the host CPU. GATK uses this approach.

[ksw2_dispatch.c](https://github.com/lh3/minimap2/blob/master/ksw2_dispatch.c) from minimap2 shows a working example. It prefers sse4.1 if present, or falls back to sse2 otherwise.

epruesse:
@lh3 Thanks for the links!

In the end, dynamically selecting code paths suited to the CPUs capabilities at run time is the way to go. 

It can happen at whole-binary level (e.g. `raxml-AVX` - though it's missing a dispatcher), or library level (e.g. Intel MKL, where `libmkl_core` contains a dispatcher and `libmkl_avx2` and friends are optimized), or function level (e.g. `minimap2` or `vsearch`). 

The higher levels are easier if you want to let the compiler optimize for features, the lower if you want to do it by hand.

The catch with hand optimized SIMD code, such as `vsearch` and `minmap2` are using, is that it's not just hard to write, it's also hard to test without having a large selection of computers with different CPU generations on hand. (We have an issue here regarding a vsearch problem on *some* Travis instances...)

The catch with letting the compiler do the optimization is that the tree vectorizers are somewhat fickle with regard to the way your loops are written. So you have to monitor with diagnostic flags that your code is actually still vectorized. And, as happened to DADA2, you may find that users with older compilers suddenly loose a lot of speed (Bioconda builds with GCC 4.8 were orders of magnitude slower).

No free lunch anymore... :)

lh3:
When vectorized, some algorithms take a distinct form. Automatically vectorizing such algorithms is beyond the capability of compilers. Hand written intrinsics is the only option. Debugging CPU dispatch is actually easy. Because dispatching is dynamic, you can implement a command-line option to target a particular instruction set at run time. You only need one recent computer to test all older architectures.

CPU dispatch does pose a challenge to conda because it may require fairly recent CPU and compilers. Purely from an enduser point of view, the best solution is to provide a portable binary precompiled on a recent machine (GATK follows this route). The conda recipe only downloads the binary, not compiling from the source. This way, all users get the maximal performance on their machines. From a developer point of view, however, this can be very complicated to implement.

lh3:
On paired-HMM, you are explicitly using SSE2. Keeping it that way requires no changes to the code. On inner products etc, you can create different versions of sdot() and choose the best version at runtime. Another option is to use a BLAS library and let the library deal with the rest. That is one extra dependency, though. Note that the point of CPU dispatch is to let users conveniently and consistently get the best performance out of our tools. CPU dispatch does make development more difficult. That is the price we developers pay for better user experiences.

If you go for the wrapper approach, the wrapper should choose the right version at runtime. In a cluster environment, we install a tool on one machine and run it on other machines that may have different specs. The launcher script has to choose the right version at runtime, or the tool may cause the invalid instruction error (if the install machine has a higher spec) or underperform (if the install machine has a lower spec).

By the way, back to this PR, I fully agree that we should remove "-march=native". CPU dispatch is more for future considerations.

epruesse:
> I presume this means the script selects the correct binary on installation, rather than at runtime?
@dancooke You can't assume that the CPU features available at installation time are also available at run time. Small clusters in which new generations of compute nodes are added over time are probably pretty common. So I'd definitely detect the CPU at run time: the cost in storage and CPU cycles is negligible, whereas explaining to each new Bioconda user how CPUs are different is anything but.

epruesse:
(Build failed because conda.anaconda.org had a 5xx error - let's retry that)

epruesse:
[off topic]

> When vectorized, some algorithms take a distinct form. Automatically vectorizing such algorithms is beyond the capability of compilers. Hand written intrinsics is the only option. 

@lh3 My experience is very limited, so not contesting your claim here. When I was working on ARB's parsimony implementation, I found it possible to write that "distinct form" in C, allowing the compiler to vectorize while avoiding intrinsics. The code ended up looking rather unusual, and required checks in the build system to prevent minor edits from breaking vectorization. The hope was that this would be more future proof than assembly style direct use of intrinsics.

> Debugging CPU dispatch is actually easy. Because dispatching is dynamic, you can implement a command-line option to target a particular instruction set at run time. You only need one recent computer to test all older architectures.

While that allows you test each implementation, it doesn't allow you to test whether an implementation works on a CPU with a smaller instruction set. You could still have an AVX2 instruction in your Pentium4 implementation and wouldn't catch it (in reality it's going to be more subtle than that of course). I think that's what's causing the vsearch issue.

I wish there was some kind of test framework that allowed checking this. Never found anything though.
 
> CPU dispatch does pose a challenge to conda because it may require fairly recent CPU and compilers. Purely from an enduser point of view, the best solution is to provide a portable binary precompiled on a recent machine (GATK follows this route). The conda recipe only downloads the binary, not compiling from the source. This way, all users get the maximal performance on their machines. From a developer point of view, however, this can be very complicated to implement.

Why does the compile host matter? You may not be able to run tests, but "cross compiling" for more capable CPUs should only be a function of the compiler.

BTW: There is a WG21 proposal for extending C++ with standard templates abstracting from SIMD intrinsics (P0214R9, partial implementations in dimsum, vc, pik/simd).

lh3:
We are talking about slightly different things, but anyway I agree with you. If you know what assembly or intrinsics-based code should look like, you can write unusual but pure C code and let the compiler vectorize for you. I know some Intel developers are doing this. However, their implementation at the time only worked with Intel compiler.

On debugging, you can disable advanced instruction sets at the compile time, such that calling those instructions leads to a compiling error.

> Why does the compile host matter? You may not be able to run tests, but "cross compiling" for more capable CPUs should only be a function of the compiler.

That is host dependent. Gcc on several of my machines are recent enough but they don't have advanced instruction sets built in.

PR #11823, opened 2018-11-08
Added build file for itsxpress 1.7.2

arivers:

### DIFF
--- a/recipes/itsxpress/1.7.1/meta.yaml
+++ b/recipes/itsxpress/1.7.1/meta.yaml
@@ -1,8 +1,8 @@
 {% set name = "itsxpress" %}
-{% set version = "1.7.2" %}
+{% set version = "1.7.1" %}
 {% set file_ext = "tar.gz" %}
 {% set hash_type = "sha256" %}
-{% set hash_value = "1f3359d416e1b840be5aead25915b152456b1201c2fc9a9665162d4442887665" %}
+{% set hash_value = "2e04503cb1c4261e75672dfb7c2999ce27fac539d8918ec5ebd53392216d2164" %}
 
 package:
   name: '{{ name|lower }}'
--- a/recipes/itsxpress/meta.yaml
+++ b/recipes/itsxpress/meta.yaml
@@ -39,14 +39,14 @@ test:
 
 about:
   home: http://github.com/usda-ars-gbru/itsxpress
-  license: 'License :: CC0 1.0 Universal (CC0 1.0) Public Domain Dedication'
+  license: 'License :: CC0 1.0'
   license_family: 'PUBLIC-DOMAIN'
-  license_file: ''
+  license_file: 'LICENSE.txt'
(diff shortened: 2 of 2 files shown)


recipes/itsxpress/1.7.1/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+about:
+  home: http://github.com/usda-ars-gbru/itsxpress
+  license: 'License :: CC0 1.0 Universal (CC0 1.0) Public Domain Dedication'
+  license_family: 'PUBLIC-DOMAIN'
+  license_file: ''
+  summary: 'ITSxpress: Software to rapidly trim the Internally Transcribed Spacer (ITS) region from FASTQ files'
+
+
+extra:
+  recipe-maintainers: 'Adam R. Rivers'
```
bgruening: please add here your github username

recipes/itsxpress/1.7.1/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+test:
+  imports:
+    - itsxpress
+  commands:
+    - itsxpress --help
+
+about:
+  home: http://github.com/usda-ars-gbru/itsxpress
+  license: 'License :: CC0 1.0 Universal (CC0 1.0) Public Domain Dedication'
```
bgruening: Can you just use CC0 1.0

recipes/itsxpress/1.7.1/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  license_family: 'PUBLIC-DOMAIN'
+  license_file: ''
```
bgruening: LICENSE.txt

bgruening (commented):
@arivers can you please add my comments to the newer recipe. Thanks!

recipes/itsxpress/1.7.1/meta.yaml
arivers: I wasn't quite sure the correct format but after looking at some other meta.yml examples I think this is correct:

```
 recipe-maintainers:
    - arivers
```

recipes/itsxpress/1.7.1/meta.yaml
arivers: Changed to 

```
license: 'CC0 1.0'
```

recipes/itsxpress/1.7.1/meta.yaml
arivers: Would it be possible to do this on the next update?

I had omitted LICENSE.txt from the MANIFEST.in file so it was not in the Python package distribution. My conda recipe builds from Pypi and I've already submitted v1.7.2 to it. To include the LICENCE.txt file I would have to bump the software version to 1.7.3.  I've aready made the suggested changes on github https://github.com/USDA-ARS-GBRU/itsxpress/commit/f89fc5775fea709ca9e7db0d319815ffeb9014d7 so it should be in there next time.

PR #11825, opened 2018-11-08
Add macOS build for comet-ms

lkuchenb:

Adds support for macOS by building from source. This requires modifying the `Makefile` as it is not compatible with macOS.

### DIFF
--- a/recipes/comet/build.sh
+++ b/recipes/comet/build.sh
@@ -1,6 +1,17 @@
 #!/usr/bin/env bash
+set -e
+set -x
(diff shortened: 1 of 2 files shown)


lkuchenb:
@bioconda/core I'm not quite sure why the build on macOS fails. It appears as if `$PREFIX/bin` doesn't exist.

lkuchenb:
Is that typically necessary? Shouldn't the `$PREFIX` directory exist beforehand?

PR #11871, opened 2018-11-09
Update Percolator: pin xercesc and boost version

Leon-Bichmann:

### DIFF
--- a/recipes/percolator/meta.yaml
+++ b/recipes/percolator/meta.yaml
@@ -16,11 +16,11 @@ requirements:
     - {{ compiler('c') }}
   host:
     - zlib
-    - boost
+    - boost = 1.64
     - tokyocabinet
     - sqlite
     - bzip2
-    - xerces-c
+    - xerces-c > 3.1
     - xsd
   run:
     - zlib


Leon-Bichmann:
@lkuchenb

Leon-Bichmann:
Hey,
I was trying to set openms2.4 and percolator3.1 as bioconda packages in a docker file
unfortunately I was getting a dependency conflict:

`conflict:
openms=2.4 -> boost[version='>=1.64.0,<1.64.1.0a0']
percolator -> boost==1.62
percolator -> xerces-c==3.1.2`

So I was hoping pining boost would solve that conflict.

recipes/percolator/meta.yaml
```diff
@@ -16,11 +16,11 @@ requirements:
     - {{ compiler('c') }}
   host:
     - zlib
-    - boost
+    - boost = 1.64
```
bgruening: I'm wondering if we can use boost-cpp here ...

lkuchenb:
I actually didn't have any problem running `conda install percolator openms` - the requirements aren't actually conflicting, are they? I mean percolator would take any boost version, right?

epruesse:
@bgruening What's actually the difference between `boost`, `boost-cpp` and `libboost` (and potentially other packages I haven't found yet)? I realize some of that is naming legacy, but that's the one to use if you normally would say `apt-get install libboost-program-options-devel` (i.e. compiling, need headers and static lib for boost program options) and what's the one for `libboost-program-options.1.32` (i.e. compiled, need that partiular DSO)?

epruesse:
@lkuchenb @Leon-Bichmann 

> I actually didn't have any problem running conda install percolator openms

That's probably because you didn't get `openms=2.4`, but a newer version. 

The big catch with boost is that it's ABI is highly unstable. If you compiled a tool with 1.67, the binaries (hence A**B**I) will be incompatible with any other version. To make matters worse, they release very frequently (quarterly).

The approach taken by the major Linux distros - split boost into dozens of small packages and give each release a new package (i.e. `libboost-program-options1.64`) so that multiple versions can be installed at the same time - was rejected by CF out of hand.

So all we can do is pin to limit the frequency of updates, and rebuild everything depending on boost every time the pin changes.

We don't have a mechanism to do the rebuild yet, so there may be packages that have been built with an older boost around, and hence won't work with the current 1.67.

You can trigger a rebuild by just bumping the build number. If you need a specific version rebuilt (2.4), you'd need to create a sub folder (e.g. openms/2.4) and put the resurrected state of the recipe at version 2.4 in there.

Leon-Bichmann:
@MatthewThe do you have an idea how to fix this conflict?

lkuchenb:
@epruesse Yup, actually an older one, gave me `openms==2.3` - which is missing the `xerces-c` version constraint (it does actually require `3.2` though)

MatthewThe:
I don't think upgrading the boost version should be a problem for percolator. However, @glormph created the percolator recipe on bioconda, so he probably knows better.

glormph:
@Leon-Bichmann this is also in my interest since I (also) have to do something similar for for nf-core who prefer one-docker-image for a whole pipeline :). But it's been a long time since I worked on the bioconda recipe so I'm a bit lost.

As for the diff between boost and boost-cpp, the only thing I could quickly find is that the boost feedstock passes this to its `bootstrap.sh`:
```
    --with-python="${PYTHON}" \
    --with-python-root="${PREFIX} : ${PREFIX}/include/python${PY_VER}m ${PREFIX}/include/python${PY_VER}" \
```
As opposed to the boost-cpp feedstock (builds without python).

But what is the actual problem here? I agree with @bgruening that it would be ideal to make the openms recipe use a later boost, but if that fails, I guess it works to do what you created a PR for? The PR checks out, right?

Leon-Bichmann:
Ok I think in any case this pull request is not usefull

PR #12058, opened 2018-11-12
DO NOT MERGE, example of updated bioconductor skeleton

dpryan79:

Xref: https://github.com/bioconda/bioconda-utils/pull/353


nturaga:
@dpryan79 Just confirming with you once, these packages are coming from the `manifest` repository with branch `RELEASE_3_8` correct?

dpryan79:
They're from three `VIEWS` files for the 3.8 release (I haven't added workflows). I presume those are updated along with `manifest`.

nturaga:
I wouldn't count on the VIEWS file for a manifest of all the packages. Because the views file represents the packages that have successfully built and it keeps updating. The manifest however, is a list of packages set of that release( RELEASE_3_8). 

```
$$ git clone https://git@git.bioconductor.org/admin/manifest

$$ git checkout -b RELEASE_3_8 origin/RELEASE_3_8

$$ ls

data-annotation.txt	data-experiment.txt	software.txt		workflows.txt
``` 

Getting the list from "software.txt" for software packages, and so on, is the best way.

dpryan79:
Do we want to include failed packages? That seems like a recipe for errors.

nturaga:
There should "never" be any failed packages in the manifest **theoretically**, since we make sure at release time. 

Theoretically, the manifest for RELEASE_3_8 should also be the same as the VIEWS file. But I think this can be sorted out within a week. I will coordinate with some of the team from Bioconductor Lori Shepherd and Valerie Obenchain to get this resolved. 

But for right now, going with the VIEWS file is fine.

dpryan79:
It's the path of least resistance at this point, since (A) it's implemented and (B) doesn't involve extra steps :) Anyway, thanks for coordinating all of that!

PR #12062, opened 2018-11-12
scanpy-scripts: pin scanpy version requirement to 1.3.2

nh3:


recipes/scanpy-scripts/meta.yaml
```diff
@@ -19,7 +19,7 @@ requirements:
         - python>=3.6
         - pandas>=0.21
         - matplotlib>=3.0.0
-        - scanpy>=1.3.2
+        - scanpy==1.3.2
```
epruesse: @bgruening I believe this would require moving the old build to `bioconda-legacy` or relabelling it. Lest conda just install the old build. Or did that get fixed in conda?

bgruening: No you are right. I can mark the old build as broken as soon as this one here is merged.

nh3: Thank you! This is a temporary fix necessary because certain existing code wouldn't work with scanpy v1.3.3. I'll work on another fix to hopefully make it forward compatible.

PR #12084, opened 2018-11-14
Gmapr

karl616:


karl616:
This adds the gmapR package from bioconductor. It is just the bioconda-utils skeleton code.
@bioconda/core Please review

bgruening:
In principle, this looks good according to our current guideline. We are also working together with the Bioconductor people to build all packages with slightly different and better recipes. Can you maybe wait a week or so ... than we might start the bioc rebuild including your recipe here.

dpryan79:
This has since been added in our twice yearly bioconductor updates.

PR #12096, opened 2018-11-15
Add humann2 v. 0.11.2 recipe

uvesten:


uvesten:
Could someone merge this, please?

recipes/humann2/0.11.2/meta.yaml
```diff
@@ -0,0 +1,84 @@
+package:
+  name: humann2
+  version: "0.11.2"
+
+build:
+  number: 3
```
bgruening: this should be 0

uvesten: Oops, will fix. But why update existing files? Isn't it better to keep the old recipes too?

bgruening: This is up the the maintainers, but if you want to do this, you move the old version into a subdir and keep the latest in root.

uvesten: Ah, idk who is maintaining this. I just need the newest version to be able to avoid a division by zero error when running this on a cluster, thought I'd might as well update the conda package. 

Didn't see in the guidlines that the most recent version should be in root. Moving the recipe then.

druvus: I restarted the test so lets hope it works now.

bgruening:
Nice work. Thanks.

PR #12101, opened 2018-11-16
First attempt at a recipe for ncrf

rsharris:

### DIFF
--- a/recipes/ncrf/build.sh
+++ b/recipes/ncrf/build.sh
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+mkdir -p $PREFIX/bin
+
+make 
+
+chmod +x NCRF
+
+mv NCRF $PREFIX/bin
+
+chmod +x *.py
+cp *.py $PREFIX/bin/
+cp ncrf_to_bed.py $PREFIX/bin/ncrf_to_bed
+cp ncrf_summary.py $PREFIX/bin/ncrf_summary
+cp ncrf_sort.py $PREFIX/bin/ncrf_sort
+cp ncrf_cat.py $PREFIX/bin/ncrf_cat
+cp ncrf_resolve_overlaps.py $PREFIX/bin/ncrf_resolve_overlaps
+cp ncrf_consensus_filter.py $PREFIX/bin/ncrf_consensus_filter
--- a/recipes/ncrf/meta.yaml
+++ b/recipes/ncrf/meta.yaml
@@ -1,0 +1,26 @@
+package:
+  name: ncrf
+  version: "1.0.4"
+
+build:
+  number: 0
+  skip: True  # [not py27]
+
+source:
+  url: https://github.com/makovalab-psu/NoiseCancellingRepeatFinder/archive/v1.00.04.tar.gz
+  sha256: fae5700393100736ad0d69cd8aa5d4c967674b12c8536bfb7fd647960397aba9
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  run:
+    - python
+test:
+  commands:
+    - NCRF --version 2>&1 | grep 1.00.04 > /dev/null
+
+about:
+  home: https://github.com/makovalab-psu/NoiseCancellingRepeatFinder
+  license: MIT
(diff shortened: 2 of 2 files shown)


recipes/ncrf/build.sh
```diff
@@ -0,0 +1,18 @@
...
+
+make 
+
+chmod +x NCRF
+
+mv NCRF $PREFIX/bin
+
+chmod +x *.py
+cp *.py $PREFIX/bin/
+cp ncrf_to_bed.py $PREFIX/bin/ncrf_to_bed
```
bgruening: you copied them already above, should this maybe a symlink?

rsharris: Thanks.  I had tried using symlinks instead of all the copies, but in my local test phase that failed. I was unable to figure out why (I am a newbie to bioconda), so I resorted to doing all the copies.

So, e.g. "ln -s ncrf_to_bed.py $PREFIX/bin/ncrf_to_bed" ought to work, right?

rsharris: I'm not clear how to proceed.  Should I (a) cancel this PR, (b) change my second copies to symlinks, (c) commit, and (d) do another PR?

martenson: @rsharris I think (b) is what @bgruening is asking for. Make a commit that changes the extra copies to symlinking and push it to `makovalab-psu:ncrf`, this PR will automatically update.

recipes/ncrf/build.sh
```diff
@@ -0,0 +1,18 @@
...
+
+make 
+
+chmod +x NCRF
+
+mv NCRF $PREFIX/bin
+
+chmod +x *.py
+cp *.py $PREFIX/bin/
+ln -s ncrf_to_bed.py $PREFIX/bin/ncrf_to_bed
```
dpryan79: These symlinks won't end up in the the final tarball, since they're being created in the source directory. Did you mean `ln -s ${PREFIX}/bin/ncrf_to_bed.py ${PREFIX}/bin/ncrf_to_bed`? You could probably just rename the files at that point (I don't know if users are expecting the .py extensions or not, though if they are then don't bother creating the symlinks).

rsharris: Yes, that is what I meant (ln -s ${PREFIX}/bin/ncrf_to_bed.py ${PREFIX}/bin/ncrf_to_bed).  I will change that.  Thanks.

Rename wouldn't satisfy my intent.  The goal is to satify two types of users -- those who want to run a python script with "python path/bin/ncrf_to_bed.py" and those who don't want to have to know what language it was written in with "path/bin/ncrf_to_bed". That's why I have two names for the same thing.

rsharris:
I've changed the symlinks, using absolute paths.

recipes/ncrf/build.sh
martenson: Just to follow up, the previous commit would have worked I think, since the relative target of the symlink would resolve relatively to the source of the symlink which would be correct in this case.

rsharris:
@bgruening @dpryan79 @martenson  Thanks all for helping this bioconda newbie.

Any estimate for how long it takes for deployment?  "conda search ncrf" isn't finding anything. (Using a miniconda install on OSX that is able to find other packages in bioconda, such as lastz). I ask only because this failure may indicate there's some additional step I need to take to make this publicly available.

dpryan79:
Hmm, it skipped building ncrf and I'm not sure why.

dpryan79:
Oh, I wonder if this happened because it lists a build skip for python 3 but then doesn't list python as a build dependency.

rsharris:
I probably specified that incorrectly.  I don't need any python as part of the build.  But the scripts, when run, are only python2 compatible.  How do I specify that?

dpryan79:
I'm testing that in https://github.com/bioconda/bioconda-recipes/pull/12264

dpryan79:
That appears to be working correctly in the PR, so I'll merge it once OSX passes testing. It should be online within an hour then.

rsharris:
A colleague here suggests I need
  host:
    - python
after my
  run
    - python

rsharris:
@dpryan   Cool, thanks.

PR #12111, opened 2018-11-16
blockclust

pavanvidem:
blockclust 1.1.0 recipe


recipes/blockclust/meta.yaml
```diff
@@ -0,0 +1,39 @@
+package:
+  name: blockclust
+  version: 1.1.0
+
+source:
+  git_url: https://github.com/pavanvidem/blockclust
+  git_rev: c4b8a985fcb615b2b4356a242ff1bc60b672f8e1
```
druvus: it is recommended to use following url format: `https://github.com/pavanvidem/blockclust/archive/c4b8a985fcb615b2b4356a242ff1bc60b672f8e1.tar.gz` instead of git_url.

druvus: It is of course even better if you could tag a release.

recipes/blockclust/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+build:
+  number: 0
+  skip: True  # [osx]
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+  run:
+    - python
+    - scikit-learn >=0.20.0
```
bgruening: you probably also need wget as runtime dependency

PR #12113, opened 2018-11-16
Strict rec

ursky:

### DIFF
--- a/recipes/metawrap/meta.yaml
+++ b/recipes/metawrap/meta.yaml
@@ -24,7 +24,6 @@ requirements:
     - bmtagger 3.101
     - bmtool 3.101
     - boost 1.64.0
-    - boost-cpp 1.64.0
     - bowtie2 2.3.*
     - brewer2mpl 1.4.1
     - bwa 0.7.15
--- a/recipes/metawrap/meta.yaml
+++ b/recipes/metawrap/meta.yaml
@@ -17,60 +17,56 @@ requirements:
   build: 
 
   run:
-    - bcftools
-    - biopython
-    - blast >=2.7.1 
-    - bmfilter >=3.101 1
-    - bmtagger >=3.101 3
-    - bmtool >=3.101 1
-    - boost >=1.64.0
+    - bcftools 1.6
+    - biopython 1.69
+    - blast 2.7.*
+    - bmfilter 3.101
+    - bmtagger 3.101
+    - bmtool 3.101
+    - boost 1.64.0
     - boost-cpp 1.64.0
-    - bowtie2 >=2.3.0
-    - bwa >=0.7.15
-    - cairo >=1.14.8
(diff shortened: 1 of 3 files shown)


recipes/metawrap/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+build:
+  number: 0
+  skip: True # [osx]
+
+requirements:
+  build: 
+
+  run:
+    - bcftools 1.6
```
bgruening: we should not pin all these packages so strict imho.

recipes/metawrap/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+    - biopython 1.69
+    - blast 2.7.*
+    - bmfilter 3.101
+    - bmtagger 3.101
+    - bmtool 3.101
+    - boost 1.64.0
+    - boost-cpp 1.64.0
```
bgruening: why are both boost dependencies needed

recipes/metawrap/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+    - bowtie2 2.3.*
+    - brewer2mpl 1.4.1
+    - bwa 0.7.15
+    - cairo 1.14.8
```
bgruening: are you sure cairo is needed?

recipes/metawrap/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+    - checkm-genome 1.*
+    - concoct 0.4.0
+    - fastqc 0.11.5
+    - gnutls 3.5.15
+    - icu 58.2
```
bgruening: From where did you get this list? Is this an environment file originally.

ursky: No, this is not an environment file. MetaWRAP relies on dozens of software to work, as it workflow incorporates many aspects of metagenomic analysis.

recipes/metawrap/meta.yaml
ursky: This is a good question. This is due to the high number of dependencies for the software. I cannot relax too many software dependencies because then the conda installation takes too long, as there are too many permutations that conda has to recursively consider when building the environment. See issue https://github.com/bxlab/metaWRAP/issues/74. I tried more relaxed settings previously but the Bioconda test build step just timed out. Additionally, I want to keep the dependencies somewhat strict to make it future-proof. Because of the number of pieces working together, it is likely that any change in them may break the pipeline.

recipes/metawrap/meta.yaml
ursky: I should also add that this list has been curated and worked over dozens of iterations and is well refined at this point. See the MetaWRAP conda distribution: https://github.com/bxlab/metaWRAP/tree/master/conda_pkg and https://anaconda.org/ursky/metawrap-mg/files.

PR #12136, opened 2018-11-18
peptideshaker, searchgui: version bump

bernt-matthias:
solving https://github.com/compomics/searchgui/issues/192


bernt-matthias:
just trying to test locally. but circleci seems to test everything ...

bernt-matthias:
@bgruening removing noarch and adding the c compile was really successful. Question is if we should ask the developers to change the package? On the other hand it might also help to use the binaries of the search engines (some of which did not work before).

bgruening:
The only downside with this currently is that we get for every python version a different pkg and the container is bigger. I guess we can life with that?

PR #12236, opened 2018-11-19
update AMPtk to version 1.2.5

nextgenusfs:

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -417,9 +417,6 @@ recipes/snvphyl-tools
 # missing perl module
 recipes/snippy/3.0
 
-# is running forever?
-recipes/amptk
-
 # hashFunction.c:94: Error: invalid instruction suffix for `push'
 recipes/soapdenovo2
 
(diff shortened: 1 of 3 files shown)


nextgenusfs:
@druvus was wondering if you would be able to help me out -- I'm trying to update AMPtk recipe to v1.2.5 -- I had to fix a problem with a dependency version check that was causing scripts to die.  But when I went to update the recipe, I noticed it was on the blacklist with a note saying "taking forever".  I'm not sure what is wrong.  I was able to get the `circleci build` locally to work, but in order to do that I needed to remove it from blacklist.  The checks seem to be stuck (i.e. the blacklisted "taking forever" comment).  One error I got during linting was a `should_be_noarch`, however, when I looked at the docs it says specifically to not add this if your program is a command line tool (which is what AMPtk is).

nextgenusfs:
Okay, so looks like if I use the skip no_arch method then it passes linting.  However, looks like the first build failed, although I don't see why -- the logs seem to indicate it built successfully but that the tar.bz2 was missing??
```
cp: cannot stat `/opt/conda/conda-bld/linux-64/amptk-1.2.5-0.tar.bz2': No such file or directory
```

druvus:
@nextgenusfs I am now back after some travelling so I will look into this issue.

druvus:
Previous attempt to get it working #11133

nextgenusfs:
Here is what is happening to me locally with bootstrap.py, not sure if this helps or not..... https://github.com/bioconda/bioconda-recipes/issues/12290

nextgenusfs:
I reorganized code and pushed a new recipe https://github.com/bioconda/bioconda-recipes/pull/13553

PR #12242, opened 2018-11-19
removed cloudpickle requirement for blockclust

pavanvidem:


epruesse:
@pavanvidem Please create a release for this. Just create a tag "v1.1.1" or "v1.1.0post1" and Github will make friendly versioned tar.gz for you on your releases page. You create tags online (https://stackoverflow.com/questions/18216991/create-a-tag-in-github-repository), or from the shell with `git tag v1.1.1`, or using a tool like `bumpversion` to update version numbers inside of files as you release a new version.

pavanvidem:
@epruesse thanks for the advice! Added a release to the source. Please check this out.

pavanvidem:
thanks @bgruening

PR #12250, opened 2018-11-20
Fix Jinja2 templated version must be quoted

epruesse:

---

ping @bioconda/core - please review as this changes 1184 recipes

Pathological case:

```yaml
{% set version = "1.10" }
package:
   name: bla
   version: {{ version }}
```

becomes

```yaml
package:
  name: bla
  version: 1.10
```

becomes

```
{'package:' { 'name': 'bla', 'version': 1.1 } }
```

which is a different version.

This PR fixes the existing recipes. We'll need a lint for this, too, though.


epruesse:
@johanneskoester I know you've wanted to use the implicit `{{PKG_VERSION}}`  instead of the Jinja2 variable, and I can see the benefits.

It does feel ugly to me though, because it mixes Jinja2 parsing with YAML parsing. The `package: version:` value isn't known before the YAML is parsed from the Jinja2 processed recipe text. So I'm guessing conda does two rounds of parsing: Line-Selectors -> Jinja2 -> YAML -> Jinja2 -> YAML. Not really pretty.

epruesse:
bump... :)

@johanneskoester wanted to use the implicit `{{PKG_VERSION}}`  instead of the Jinja2 variable, and I can see the benefits. For now this is a bug though that needs fixing... 

Comments?

The linting fails because some 30 packages have things no longer accepted. If I must, I can go through them, just not sure that this is the right place to fix legacy lint.

PR #12265, opened 2018-11-20
Add recipe for SNVer 0.5.3

cwuensch:

@bioconda/core

### DIFF
--- a/recipes/snver/meta.yaml
+++ b/recipes/snver/meta.yaml
@@ -16,7 +16,7 @@ package:
 
 build:
   noarch: generic
-  number: 1
+  number: 0
 
 source:
   url: https://sourceforge.net/projects/snver/files/SNVer-{{ version }}.tar.gz
--- a/recipes/snver/meta.yaml
+++ b/recipes/snver/meta.yaml
@@ -1,0 +1,43 @@
+{% set name = "snver" %}
+{% set version = "0.5.3" %}
+
+about:
+  home: http://snver.sourceforge.net/
+  license: GPL
+  summary: |
+    SNVer is a statistical tool for calling common and rare variants in analysis of pool or individual next-generation sequencing data.
+    It reports one single overall p-value for evaluating the significance of a candidate locus being a variant, based on which multiplicity control can be obtained.
+    Loci with any (low) coverage can be tested and depth of coverage will be quantitatively factored into final significance calculation.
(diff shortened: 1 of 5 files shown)


cwuensch:
Hello stain,

1.) I have changed the names of the installed commands as you suggested.
2.) I am not the author of SNVer. I just wanted to contribute this tool, because we are using it a lot. But since I am not the author, I am not able and do not want to make any changes to the upstream-package that is provided by the authors. So I would prefer to keep the lib-folder as is provided in the published download archive.
3.) I could add some tests with actual data, but we will not be able to find out, whether the output is correct. So it seems to be as good to just test it the programm can be run. Right?
4.) License is "GNU General Public License version 3.0 (GPLv3)"

Does this answer all your questions sufficiently?

cwuensch:
@stain Please let me know, whether my changes / answers to your review do solve your change request, or what should still be done.

cwuensch:
Dear bioconda team @bioconda/core,
what about this recipe now? I tried to address all changes that have been requested by stain as good as possible.
Does my recipe fit your requirements now?
Am I allowed to just push the Merge-Button?

recipes/snver/meta.yaml
```diff
@@ -0,0 +1,43 @@
+{% set name = "snver" %}
+{% set version = "0.5.3" %}
+
+about:
+  home: http://snver.sourceforge.net/
+  license: GNU General Public License version 3.0 (GPLv3)
```
epruesse: GPL requires that distributions be accompanied by a copy of the license, so `license_file` must be set. Should be somewhere in the tarball.

cwuensch:
@epruesse There is - unfortunately - no license file included in the author's tarball.
Only on their website they are stating GPL 3.0.
What shall I do about this? I cannot change the source tarball, because I am not the author of this tool.

epruesse:
Could you try to contact the authors? Manual, manuscript and code all do not mention a license. The only place I see it is on the "browse" page. 

@bioconda/core - how do we handle situations like this? Has there been "precedence"?

bgruening:
We are currently not that strict in bioconda, which we should change. First and foremost the linter should complain. conda-forge is fixing this issue by putting the license file into the recipe folder.

cwuensch:
Did what @bgruening suggested. Added license file to recipe archive.
Now everything okay from your point, @epruesse?

epruesse:
If you're certain the license is OK now, then yes.

PR #12272, opened 2018-11-20
Create recipe for SVIM

eldariont:


recipes/svim/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  url: https://files.pythonhosted.org/packages/ba/1b/80a7542b27637ee85ff9239b10ac14283be3c2fcf52a9659d3cc82b34929/svim-{{ version }}.tar.gz
+  sha256: bf4aae8e774bb2621d8ad649148152a73ad07952a04b672aaf6201546b900d76
+
+build:
+  number: 0
+  skip: True # [py27]
+
+requirements:
+  host:
+    - python 3.6
```
epruesse: Make this `3.6.*` - patch releases should be ok

recipes/svim/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  name: svim
+  version: {{ version }}
+
+source:
+  url: https://files.pythonhosted.org/packages/ba/1b/80a7542b27637ee85ff9239b10ac14283be3c2fcf52a9659d3cc82b34929/svim-{{ version }}.tar.gz
+  sha256: bf4aae8e774bb2621d8ad649148152a73ad07952a04b672aaf6201546b900d76
+
+build:
+  number: 0
+  skip: True # [py27]
```
epruesse: This isn't needed if we have the python requirement below.

recipes/svim/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - ngmlr
+    - samtools
+
+test:
+  commands:
+    - svim -h
+
+about:
+  home: https://github.com/eldariont/svim
+  license: GPL-3.0
```
epruesse: please add the license_file as well. GPL requires it.

recipes/svim/meta.yaml
eldariont: Okay, is it necessary to remove it? I'm unsure because it is requested in the bioconda guidelines: https://bioconda.github.io/guidelines.html#python

epruesse: If you don't have need for a `{{compiler('xyz')}}`, your package should be `noarch:`. If it is noarch, it should not have `skip: True`.

eldariont: Alright, thanks for the feedback. I replaced `skip: True` with `noarch: python`.

recipes/svim/meta.yaml
eldariont: Thanks, done.

recipes/svim/meta.yaml
eldariont: Added the license file.

PR #12274, opened 2018-11-21
Make snakemake a noarch package.

johanneskoester:

### DIFF
--- a/recipes/snakemake-minimal/meta.yaml
+++ b/recipes/snakemake-minimal/meta.yaml
@@ -9,8 +9,8 @@ source:
   sha256: 612a16f2bba580c47183e4f752ed40b034852541ca3e96e480d01fb30ad9570c
 
 build:
-  number: 1
-  skip: True # [py27]
+  number: 2
+  noarch: python
   script: python -m pip install --no-deps --ignore-installed .
   entry_points:
     - snakemake = snakemake:main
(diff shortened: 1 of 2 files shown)


recipes/snakemake-minimal/meta.yaml
```diff
@@ -28,7 +28,7 @@ requirements:
     - appdirs
     - datrie
   run:
```
melund: Since this is the "minimal" package I think it should only include the packages also listed in `Setup.py`. So maybe drop `psutil`

johanneskoester: Good catch. However, psutil will soon become a requirement (a future version of Snakemake will do benchmarking by default). Hence, I think it should be fine to keep it.

recipes/snakemake-minimal/meta.yaml
```diff
@@ -9,16 +9,16 @@ source:
...
-  skip: True # [py27]
+  number: 2
+  noarch: python
   script: python -m pip install --no-deps --ignore-installed .
   entry_points:
     - snakemake = snakemake:main
     - snakemake-bash-completion = snakemake:bash_completion
 
 requirements:
   host:
```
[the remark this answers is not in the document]
johanneskoester: Yes indeed. It was necessary in some previous release. But not anymore.

PR #12275, opened 2018-11-21
bioconductor-decipher: restrict r-rsqlite version

richjstark:
r-rsqlite version dependency added


apcamargo:
The recipe is not building because the build number is the same. You need to change it to 1

stain:
Thanks @richjstark!

PR #12282, opened 2018-11-21
update rgi4.2.2 recipe

mjram0s:


recipes/rgi/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
+cd card-data
+mv card.json ..
+cd ..
+rm -rf card-data
+./rgi load -i card.json
```
apetkau: You should be able to condense the above `cd`, `mv` statements into something like `./rgi load -i card-data/card.json`.  Also, maybe add a comment saying this is loading up the database files.

recipes/rgi/meta.yaml
```diff
@@ -1,29 +1,43 @@
 {% set name = "rgi" %}
 {% set version = "4.2.2" %}
-{% set sha256 = "ab4a5614e359884129064ec318d8dda2516052b26aa38cfd480245d24e0c96e7" %}
```
apetkau: Could you actually keep this variable set up here at the top?  You can also add another variable `sha256database` for the database.

apetkau:
@raphenya this update should distribute the CARD database along with rgi in bioconda.

apetkau:
Hooray, thanks @mjram0s

PR #12286, opened 2018-11-21
R tigger

PertuyF:


stain:
Sorry as I'm a newbie for the `r-*` packages, but is it a convention for new packages to include a long list of maintainers like @bgruening and @johanneskoester ?

bgruening:
@stain this is probably generated with a helper script that we use for conda-forge where this is necessary. I don't mind if we leave them included. It might be nice to keep the r-* recipes consistent but other then that I don't have any preference.

PR #12314, opened 2018-11-23
Bioc3.8

dpryan79:


dpryan79:
@bgruening When you get a chance, please merge.

bgruening:
@dpryan79 I got pretty far today with conda-forge and r-* packages. If we get the gcc7 tag re-labeled as main in the next weeks, should we wait with this PR? Should we also adopt many of those packages to `noarch: generic`? This would decrease also the build time.

dpryan79:
You're more optimistic about finishing the conda-forge migration this year than I am.

Yes, we should implement the `generic` change. The skeleton will need to be updated appropriately. I don't think there's anything preventing us from at least preparing that part now, so I'll try to get it done this week.

PR #12330, opened 2018-11-25
MSGFplus does not work with newer Java

glormph:

This is hardly definite, but the conda-forge openJDK recipe got updated from 1.8 to 11.0 3 or so days ago. This led apparently to it dropping some classes that MSGF needs for XML reading. A later version of MSGF+ works fine apparently (so I will also update the MSGF version in this recipe to a newer version later). The problem is referenced here: https://github.com/MSGFPlus/msgfplus/issues/18.
I tried to use the `--add-modules java.xml.bind` but it hasnt worked (yet).

### DIFF
--- a/recipes/msgf_plus/meta.yaml
+++ b/recipes/msgf_plus/meta.yaml
@@ -12,7 +12,7 @@ source:
 
 requirements:
   run:
-    - openjdk
+    - openjdk<11
     - python
 
 test:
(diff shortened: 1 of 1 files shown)


glormph:
It works. Thanks @bgruening for the quick merge!

glormph:
Whoops I touched the revert button (while figuring out how to change this PR). Pls ignore.

PR #12340, opened 2018-11-26
New MOB-suite built of the 1.4.9 version with Python req downgraded to >=3.4

kbessonov1984:
* [x ] This PR does something else (explain below).


recipes/mob_suite/meta.yaml
```diff
@@ -16,11 +16,11 @@ source:
 
 requirements:
   host:
-    - python
+    - python >=3.4
```
bgruening: you already have `skip: True  # [not py36]` above ... are you sure this solves your problem?

kbessonov1984:
Can our branch be merged. The only change is that this built should have package versions for python 3.4, 3.5 and 3.6

PR #12342, opened 2018-11-26
Add recipe for libstatgen-1.0.5

dfornika:

### DIFF
--- a/recipes/libstatgen/build.sh
+++ b/recipes/libstatgen/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+make
+cp libStatGen.a "${PREFIX}"/lib
+
--- a/recipes/libstatgen/build.sh
+++ b/recipes/libstatgen/build.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+export CFLAGS="-I$PREFIX/include -I${SRC_DIR}/include"
+export LDFLAGS="-L$PREFIX/lib"
+export CPATH=${PREFIX}/include:${SRC_DIR}/include
+
 make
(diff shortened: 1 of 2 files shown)


dfornika:
Build is failing with this message:

```
SamReferenceInfo.h:21:25: fatal error: StringArray.h: No such file or directory

 #include "StringArray.h"
```

I think the problem is that during the build, symbolic links to `StringArray.h` and other headers are created in the `include` directory of the source repo. But the compiler isn't looking them there while compiling `SamFileHeader.cpp`.

npavlovikj:
@dfornika , is there a reason you are not building the newest 1.0.14 version of libstatgen ?

dfornika:
Thanks @npavlovikj. I'm interested in this pipeline: https://github.com/CPTR-ReSeqTB/UVP, which lists 'Fastqvalidator Version 1.0.5' as a dependency. There is no 1.0.5 release of fastQValidator, but there is a 1.0.5 release of the libStatGen library that fastQValidator uses.

But I guess that isn't really a safe assumption, since fastQValidator pulls in libStatGen with a simple `git clone` of the master branch during its build process.

@mezewudo Could you comment on exactly what you're referring to when you list 'Fastqvalidator Version 1.0.5' as a dependency of https://github.com/CPTR-ReSeqTB/UVP?

mezewudo:
You are right, there is no FastqValidator Version 1.0.5, there is only a
libStatGen library release of 1.0.5. I will update the UVP GitHub readme
document to reflect the accurate version of fastQvalidator implemented ---
Version 0.1.1

dfornika:
This recipe was added in #12348

PR #12405, opened 2018-11-28
Fix duplicate build section in various recipes

epruesse:

### DIFF
--- a/recipes/avro-cwl/meta.yaml
+++ b/recipes/avro-cwl/meta.yaml
@@ -1,8 +1,5 @@
 {% set version = "1.8.9" %}
 
-build:
-  number: 0
-
 package:
   name: avro-cwl
   version: {{ version }}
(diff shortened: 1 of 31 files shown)


dpryan79:
Looks good, hopefully the additional tests would have passed had they been run.

On a related note, is there a reason so many generic perl packages are in bioconda rather than conda-forge?

bgruening:
Also I must admit that we don't have a solid plan for Perl. Like noarch Perl or which version to pin. Major vs minor.

epruesse:
IIRC Perl needs pinning at x.x if not x.x.x for binary modules. For pure Perl the language seems to be more stable, but I’m not a Perl master so might be wrong.

In any case - a lot of the Perl modules should be moved to CF. Last time I was asking about this I was very novice and messing with X11 on the CF side to get ARB to work. I was told to wait until after R was done. Guess that’s over with now...

Migration issues will be the dependency graph and the windows side I’m guessing.

PR #12407, opened 2018-11-28
gffcompare: unblacklist and update (plus remove gclib)

bernt-matthias:

this unblacklists gffcompare


bernt-matthias:
I guess this is the way to go and we can close https://github.com/bioconda/bioconda-recipes/pull/12416 ..

bernt-matthias:
Q: should we blacklist older versions?

bgruening:
>Q: should we blacklist older versions?

Remove them :)

>Q: should we delete the gclib recipe?

Yes please.

bernt-matthias:
Removed gclib. 

Currently I would only remove 0.10.1 since 0.9.8 is used in a galaxy tool. Question is how? :)

bernt-matthias:
Also there seems to be an issue with bioconda linting at the moment. I also observed this for other recipes

bernt-matthias:
@jfallmann could you have a look ... maybe you also know an answer for the question from above:

> how to blacklist old versions

but, maybe this is not necessary anyway?

jfallmann:
@bernt-matthias No I'm not sure what to do for deleting only a specific version, could we just blacklist that specific version and keep the older version unblacklisted?

bgruening:
Which version do you need to be blacklisted?

jfallmann:
@bgruening 
> Currently I would only remove 0.10.1 since 0.9.8 is used in a galaxy tool. Question is how? :)
So 0.10.1 remove and keep 0.9.8

bgruening:
done

bernt-matthias:
> Which version do you need to be marked as broken?

Unsure. According to https://anaconda.org/bioconda/gffcompare/files there were only 0.9.8  and 0.10.1. Both have been downloaded often and I guess they are functional. 

My idea was just born from the fact that gffcompare was blacklisted (from here https://github.com/bioconda/bioconda-recipes/commit/7f2d1f0bf092e68c15fee5fd7b6fe3262d0e188a .. this was seemingly during 0.10.1 - but I don't know why it was blacklisted). Version 0.10.5 which came later apparently never went online. 

So I guess it's fine now?

PR #12413, opened 2018-11-29
Metaquantome

caleb-easterly:
@bgruening, @jj-umn, do you have pointers about how to fix this? I'm not sure if it's the program structure (see https://github.com/galaxyproteomics/metaquantome), the setup.py file, or the conda recipe that's causing the error.

I'm getting the following error when the build fails:
```python
ImportError: No module named 'metaquantome.cli'
```

### DIFF
--- a/recipes/metaquantome/LICENSE
+++ b/recipes/metaquantome/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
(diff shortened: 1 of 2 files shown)


caleb-easterly:
I think the problem is that `cli.py` is not in the directory with the other module code (the inner metaquantome directory), and is not getting packaged in setup.py.

However, when I move `cli.py` into the inner metaquantome directory, I get the following import error:

```
  File "metaquantome/metaquantome/cli.py", line 5, in <module>
    from metaquantome.analysis.expand import expand
ImportError: No module named 'metaquantome'
```
I don't understand why `cli.py` can't import the module it's in, which is why I moved it out in the first place.

Is there a way to fix how setup.py deals with the entry point?

bgruening:
Cool @caleb-easterly!

PR #12416, opened 2018-11-29
gclib: update to new version and unblacklist

bernt-matthias:

Needed to get gffcompare running https://github.com/bioconda/bioconda-recipes/pull/12407 (I will remove the gclib changes from the other RP)

Test did not run locally for me since g++ was not found. But compiler is in build requirements.

### DIFF
--- a/recipes/gclib/chrono.patch
+++ b/recipes/gclib/chrono.patch
@@ -1,0 +1,63 @@
+diff -ruN /tmp/gclib-0.10.3/GThreads.cpp ./GThreads.cpp
+--- ./GThreads.cpp	2018-06-26 00:26:41.000000000 +0200
++++ ./GThreads.cpp	2018-11-29 13:55:31.212906749 +0100
+@@ -375,7 +375,7 @@
+ 
+ void GThread::wait_all() {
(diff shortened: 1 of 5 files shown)


recipes/gclib/chrono.patch
```diff
@@ -0,0 +1,63 @@
+diff -ruN /tmp/gclib-0.10.3/GThreads.cpp ./GThreads.cpp
```
epruesse: Can you put a comment at the top of this file, explaining briefly what it does? (Yep, that's allowed in patch files)

epruesse: (And the other patches as well ... this one seems to need a lot of patching...)

bernt-matthias:
I reported upstream: https://github.com/gpertea/gclib/issues/3

bernt-matthias:
I added the comments. But @gpertea suggested to build gffcompare differently and NOT to distribute gclib as library: https://github.com/gpertea/gclib/issues/4

I will give it a try. Maybe we leave the PR open until this is done.

bernt-matthias:
Superseded by https://github.com/bioconda/bioconda-recipes/pull/12407

PR #12420, opened 2018-11-29
stacks: fix for fix 'fixing' @ in exe_path

bernt-matthias:
the previous fix https://github.com/bioconda/bioconda-recipes/pull/11580
tried to solve the problem with the @ in the exe_path by setting this
variable empty. But the perl scripts
- append a / to the empty string and (i.e. the bin path gets /binary)
- check for file presence (problem: neither binary nor /binary are
  present)

Now I quote the @ in the exe_path.


epruesse:
ok ... let's give it a try. Can you ping Bjoern if the old build number is broken enough to warrant deletion?

bernt-matthias:
ping @bgruening should we delete build 1? The perl pipelines were broken, but since I'm just working on the Galaxy wrapper I guess it wasn't installed anywhere.

bernt-matthias:
hi @epruesse . I just tried the new recipe and it seems that quoting the `@` does not work. I just tried the sed command once more and it works. Any idea?

I installed 
+bioconda::stacks-2.2-he860b03_2 

this should be the most recent version, ie this PR?

PR #12426, opened 2018-11-29
pin against latest boost and friends

bgruening:


epruesse:
These shouldn't even need a build number bump, should they?

No matter though...

bgruening:
@epruesse I think we do need to build number bump. I was also confused and ask in conda-forge and they said we should bump. This makes all our rebuilding more complicated as we need to bump all buildnumbers as soon as we update conda-forge-pinnings.

PR #12441, opened 2018-11-30
Salmon v0.12.0 --- version bump and fix OSX build

rob-p:


rob-p:
Well, that failure in TBB is new ...

dpryan79:
You probably want to ping /lib rather than /all next time :)

rob-p:
Good call (fixed, but too bad I already pinged too many ppl).  Are there other packages using libtbb?  Seemingly, they should all suffer from this problem if using the new compilers.

dpryan79:
There are only 8 bioconda packages (including salmon and sailfish) using tbb. I think flexbar is using new compilers with tbb too, but I'm not overly familiar with it.

epruesse:
You should only need `tbb-devel` in `requirements/build`. It `run_exports` the `tbb` package, so like zlib, it'll be included and pinned automatically.

rob-p:
Is there a way to skip linting?

epruesse:
No ... I wonder what's going on.

dpryan79:
It passed this time at least

rob-p:
man, this linter is really a PITA.

PertuyF:
Looks like it is possible to [skip linting](https://bioconda.github.io/linting.html#skipping-persistently-on-a-per-recipe-basis), at least part of it.
Requires some investigation on how to use it though.

PertuyF:
Error comes from [there](https://github.com/bioconda/bioconda-utils/blob/7cbde1cd8bdb61b0f4d7beff4b3cbe999d31a29e/bioconda_utils/linting.py#L123), and `bioconda-utils.linting.channel_dataframe()` is used by `[bioconda-utils.cli.lint()](https://github.com/bioconda/bioconda-utils/blob/d54a6d8bbd9f23e2bc7967030170c4e12369dc58/bioconda_utils/cli.py#L228)`. So I guess there is no specific linter function we can skip here.
There may be some problem in JSON objects returned from `bioconda-utils.utils.get_channel_repodata()`?

Could it be a global problem from `bioconda-utils`?

apcamargo:
https://github.com/bioconda/bioconda-utils/pull/362

This PR should fix the linter problems.

daler:
Thanks to @dpryan79 in bioconda/bioconda-utils#362 the lint is now passing!

rob-p:
The linting (and linux build) succeed now.  However, I noticed the following warnings after a successful build:

```
The update_index function has changed to index all subdirs at once.  You're pointing it at a single subdir.  Please update your code to point it at the channel root, rather than a subdir.
WARNING:conda_build.index:The update_index function has changed to index all subdirs at once.  You're pointing it at a single subdir.  Please update your code to point it at the channel root, rather than a subdir.
The update_index function has changed to index all subdirs at once.  You're pointing it at a single subdir.  Please update your code to point it at the channel root, rather than a subdir.
WARNING:conda_build.index:The update_index function has changed to index all subdirs at once.  You're pointing it at a single subdir.  Please update your code to point it at the channel root, rather than a subdir.
..
```

should we be concerned about these?

dpryan79:
The warning can be ignored, it’s just due to the directory passed to conda index.

PR #12454, opened 2018-12-01
stacks: another try for a fix for @ in exe_path

bernt-matthias:
really sorry, but the last fix did not work. I guess that in the constructed package there is no `@`, but only placeholders. Hence I guess this needs to be fixed during runtime.  


bernt-matthias:
Hi @bgruening could you have a look at this or suggest a reviewer?

PR #12459, opened 2018-12-03
Recipe for sparcc

pvanheus:


druvus:
@pvanheus we should probably ask upstream to package the python code properly so it is not only a set of files copied to bin folder.

pvanheus:
Thanks for the fixes @druvus . I've made a [PR](https://bitbucket.org/yonatanf/sparcc/pull-requests/3/provide-packaging-for-sparcc/diff) that adds packaging to the upstream package. When/if it gets merged the recipe can be updated. For now I think this should be merged.

PR #12474, opened 2018-12-04
Added UCLUST recipe

ValentinMarcon:

poke @mariabernard


epruesse:
I'm sorry. We can't distribute this because of the license.

From http://www.drive5.com/uclust/downloads1_2_22q.html

> **These binaries are licensed only for use in PyNAST and QIIME.**
> **The license does not permit stand-alone use.**

mariabernard:
That's right, we read it!
@ValentinMarcon made this pull request because uclust is not included in the pynast recipe, and we use pynast in FROGS.

What about a pull request on the pynast recipe ? would it be ok ?

PR #12488, opened 2018-12-05
For VarDict-Java, install utility scripts.

cbrueffer:
Utility script that were previously only shipped with VarDict are not
part of VarDict-Java as well, making installation of both packages
unnecessary.

### DIFF
--- a/recipes/vardict-java/build.sh
+++ b/recipes/vardict-java/build.sh
@@ -7,5 +7,8 @@ mkdir -p $PREFIX/bin
 
 cp lib/*.jar $outdir/lib
 cp bin/VarDict $outdir/bin/vardict-java
-chmod +x $outdir/bin/vardict-java
-ln -s $outdir/bin/vardict-java $PREFIX/bin
+cp bin/*.{R,pl} $outdir/bin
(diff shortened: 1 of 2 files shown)


cbrueffer:
@chapmanb this may make installation of the `vardict` package in bcbio-nextgen unnecessary.

chapmanb:
Christian;
Thanks for this. Having these available without vardict (perl) would be great, but my main worry is that this causes the projects to conflict since they're both installing all the accessory scripts. The regular vardict is often useful as it's more of a development version and sometimes has features that aren't yet ported to vardict-java.

What do you think about only installing the accessory scripts here, removing them from `vardict` and then making `vardict` depend on `vardict-java`? That would avoid the conflicts and also make `vardict-java` easier to use without `vardict`, fulfilling your goal.

cbrueffer:
Good point; the alternative would be installing the VarDict-Java versions with a prefix, although I like your suggestion better.

cbrueffer:
@chapmanb vardict depends on vardict-java now; does this look OK to merge?

chapmanb:
Brilliant, looks perfect. Thanks so much.

Issue #12487, opened 2018-12-05
Recipe gets blacklisted during upload and new version doesn't get pushed to r-seurat-scripts


pcm32:
While tests and merging went fine, posterior upload-linux and upload-mac doesn't do anything: 

https://circleci.com/gh/bioconda/bioconda-recipes/36244

In the Building, testing and uploading part it can be seen that r-seurat-script (the package in question) gets blacklisted and then it outputs:

```
12:11:29 BIOCONDA INFO Nothing to be done.
```

We tried re-sending the job, but the same behaviour happens. As a result, the package is not available in conda/bioconda.

Thanks!

pcm32:
ping @bioconda/core

daler:
@pcm32 `r-seurat-scripts` is in https://github.com/bioconda/bioconda-recipes/blob/master/build-fail-blacklist#L788. It looks to be there due to the recent R upgrade. You'll need to submit a PR with that line removed to get the package built.

pcm32:
Thanks @daler! I have made the PR now, including a few of our other packages.

bgruening:
I will investigate this here: https://github.com/bioconda/bioconda-recipes/pull/12491

bgruening:
Upps, ok I had the tab open and haven't seen the updates. I checked the blacklist file and seurat was not there, clearly because it was fixed already :)

Great! :)

PR #12489, opened 2018-12-05
Removes r-seurat-scripts from blacklisting

pcm32:

Remove package from blacklist. Sorts out #12487


pcm32:
I wonder if we should change the names of this packages to avoid this confusion with normal r and bioconductor packages?

bgruening:
> I wonder if we should change the names of these packages to avoid this confusion with normal r and bioconductor packages?

Yes, `seurat-scripts` should be enough.

PR #12490, opened 2018-12-05
Pin compatible numpy for older HTSEQ versions

lparsons:
This is required to keep older versions of HTSEQ functional

### DIFF
--- a/recipes/htseq/0.6.1/meta.yaml
+++ b/recipes/htseq/0.6.1/meta.yaml
@@ -12,6 +12,8 @@ package:
   version: 0.6.1
 
 requirements:
+  build:
+  - {{ compiler('c') }}
   host:
   - python
   - cython
--- a/recipes/htseq/0.6.1/meta.yaml
+++ b/recipes/htseq/0.6.1/meta.yaml
@@ -4,7 +4,7 @@ about:
   summary: 'A framework to process and analyze data from high-throughput sequencing (HTS) assays'
 
 build:
-  number: 1
+  number: 2
   skip: True # [py3k or osx]
 
 package:
--- a/recipes/htseq/0.6.1.post1/meta.yaml
+++ b/recipes/htseq/0.6.1.post1/meta.yaml
@@ -8,7 +8,7 @@ source:
 
 build:
   preserve_egg_dir: True
-  number: 2
+  number: 3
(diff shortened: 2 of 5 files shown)


lparsons:
This appears to be working, but I have one question: Is there some way to test that this fixes the issue of HTSeq getting installed with an incompatible numpy before merging? I have been unable to reproduce the error in testing since the tests rebuild the packages with the new numpy.

bgruening:
> This appears to be working, but I have one question: Is there some way to test that this fixes the issue of HTSeq getting installed with an incompatible numpy before merging? I have been unable to reproduce the error in testing since the tests rebuild the packages with the new numpy.

You mean htseq as a dependency of an other package? You could include in this PR to recipes one depends on the other, this should work.

recipes/htseq/0.6.1/meta.yaml
```diff
@@ -12,6 +12,8 @@ package:
   version: 0.6.1
 
 requirements:
+  build:
```
bgruening: Build number increase in this recipe please.

lparsons: No changes to that recipe.

lparsons: Oh, wait, I see there is one... Sorry. I'm actually not clear if that's needed. I put that in attempting to fix `circleci build` errors, but I never got that working.

lparsons:
@bgruening I don't see how this tests anything. If I understand the problem correctly, the issue is that HTSeq was built with an older numpy, but when it's installed now, a newer, incompatible numpy is installed with it. This update to the recipe triggers a new build, which is with the new numpy, and thus works. The question is, will it break in the future? I guess no way to know.

recipes/htseq/0.6.1/meta.yaml
```diff
@@ -12,6 +12,8 @@ package:
...
+  - {{ compiler('c') }}
```
lparsons: Is this required/recommended? I added it in an attempt to get `circleci build` tests to pass, but they never did. Not sure if I should keep it or not.

bgruening: yes it is, if a compiler is needed :)

bgruening:
Thanks!

PR #12500, opened 2018-12-06
snakemake 5.3.1

johanneskoester:

### DIFF
--- a/recipes/snakemake-minimal/meta.yaml
+++ b/recipes/snakemake-minimal/meta.yaml
@@ -41,8 +41,9 @@ test:
   imports:
     - snakemake
   commands:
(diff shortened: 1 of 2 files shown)


johanneskoester:
Problem found. You cannot have a string starting with `GIT_` anywhere in the recipe, because then conda thinks that the recipe contains a git source definition...

epruesse:
Was it conda itself or bioconda utils?
BTW: While you are updating the recipe - if you use the outputs thing you can have Snakemake and Snakemake-minimal in one recipe and autobump can catch updates you release on bit bucket.

johanneskoester:
@epruesse thanks for the hint! I will definitely do that soon. No time for it now though.

PR #12553, opened 2018-12-10
added build files for pathogist

realseanla:

Please and thank you!

### DIFF
--- a/recipes/adam/meta.yaml
+++ b/recipes/adam/meta.yaml
@@ -1,4 +1,4 @@
-{% set version="0.24.0" %}
+{% set version="0.25.0" %}
 
 about:
   home: https://github.com/bigdatagenomics/adam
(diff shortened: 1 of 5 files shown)


druvus:
cplex is a missing dependency

realseanla:
I think the reason is that cplex is taken from the conda channel `ibmdecisionoptimization`, and the build tests might not have access to that channel, so it's throwing a dependency error.
@druvus do you think this could be the issue?

bgruening:
@realseanla yes. You need to add cplex to conda-forge at first.

realseanla:
@bgruening I don't own cplex, IBM does. I'm not sure whether I have the legal right to add a private organization's proprietary software to conda-forge. Is there a way to link the `ibmdecisionoptimization` to the circleci tests?

PR #12581, opened 2018-12-11
Try to use GCC7 for hap.py

kyleabeauchamp:

This tries to overcome https://github.com/Illumina/hap.py/issues/66 by using the latest GCC7 compilers.

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -328,8 +328,6 @@ recipes/koeken
 recipes/nucleoatac/0.3.1
 
 
-# no clue currently
-recipes/hap.py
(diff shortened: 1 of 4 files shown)


kyleabeauchamp:
It looks like the latest commit fails due to hard-coded `gcc` calls in external dependency makefiles, which are included in tarballs.  It would be hard to patch those because they live inside tarballs, rather than as text in the git source tree.  :(

bgruening:
@kyleabeauchamp try to symlink conda GCC to the expected gcc. E.g. with `ln -s $GCC $PREFIX/bin/gcc`

kyleabeauchamp:
Good idea, that solved the hard-coded gcc.  Now I just have to fix the hardcoded zlib paths in the vendored package building.

PR #12618, opened 2018-12-13
Bump topas

druvus:


recipes/topas/build.sh
```diff
@@ -4,7 +4,7 @@ set -eu -o pipefail
 outdir=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
 mkdir -p $outdir
 mkdir -p $PREFIX/bin
-cp -R $SRC_DIR/executable/TOPAS.jar  $outdir/
+cp -R $SRC_DIR/TOPAS-1.0.1.jar  $outdir/
```
epruesse: If you use `${PKG_VERSION}`, autobump might be able to handle the next version of this.

epruesse:
@druvus - Theoretically, I could make autobump look for the version in `build.sh` and referenced patch files and replace it in there as well. It's not going to work perfectly though - the change for a false positive on e.g. `1.0` is going to be way too high. You've dealt with more updates than I. What do you think?

recipes/topas/build.sh
druvus: thanks @epruesse! Updated it in latest commit.

druvus:
@epruesse an issue with all java wrappers around jar-files is hard-coded file names. It this case we have this row`jar_file = 'TOPAS-1.0.1.jar'`in topas.py. Do you have any suggestion how we could make it more generic so it doesn't need to be manually updated.

druvus:
I guess you are right about the high risk of false positives. You might be able to do it for java-wrappers by looking the additional `packagename.py` file.

epruesse:
We could just rename the jars in the `build.sh`. Program names don't usually include the version.

epruesse:
> I guess you are right about the high risk of false positives.
Experience. During early development of autobump, I was replacing the version across the `meta.yaml`, but that lead to problems with the `doi`, which will often contain something like `1.0`.

recipes/topas/build.sh
```diff
@@ -4,7 +4,7 @@ set -eu -o pipefail
 outdir=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
 mkdir -p $outdir
 mkdir -p $PREFIX/bin
-cp -R $SRC_DIR/executable/TOPAS.jar  $outdir/
+cp -R $SRC_DIR/TOPAS-${PKG_VERSION}.jar  $outdir/
```
epruesse: If you write 
```
cp -R $SRC_DIR/TOPAS-${PKG_VERSION}.jar  $outdir/TOPAS.har
```

recipes/topas/topas.py
```diff
@@ -11,7 +11,7 @@
 import sys
 import subprocess
 from os import access, getenv, X_OK
-jar_file = 'TOPAS.jar'
+jar_file = 'TOPAS-1.0.1.jar'
```
epruesse: ... then this can stay `jar_file = 'TOPAS.jar'

recipes/topas/build.sh
epruesse: Actually, the `-R` is not needed - it's just one file and we don't want recursion here.

PR #12662, opened 2018-12-14
Use quotes when extending PATH.

pkrog:

Correct PATH definition, when extending it for miniconda.
When extending PATH env var, use quotes in order to handle potential space characters in PATH.

### DIFF
--- a/.circleci/setup.sh
+++ b/.circleci/setup.sh
@@ -12,7 +12,7 @@ set -u
 source .circleci/common.sh
 
 # Set path
-echo "export PATH=$WORKSPACE/miniconda/bin:$PATH" >> $BASH_ENV
+echo "export PATH=\"$WORKSPACE/miniconda/bin:$PATH\"" >> $BASH_ENV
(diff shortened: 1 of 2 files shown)


bgruening:
You need to increase the build number.

pkrog:
Which build number ? This PR does not modify any recipe.
It just fixes an issue in the `setup.sh` script.
Do you mean the checks were not run and I need to force a run ?

bgruening:
Yes, we only build recipes if the version+build number changes. So you need to increase the build/number by one.

epruesse:
What brought this up? Did someone actually use spaces in the path?

epruesse:
(I mean, sure, do that. Might even work with most Linux Desktop software these days. But with bioinf software? I'm guessing at least half of everything that includes a shell script will break. Probably more.)

pkrog:
Yes, I actually have a space in my `PATH` env var, due to some application. Anyway, since space is allowed in file names, it must be handled.

PR #12670, opened 2018-12-15
Use findtools to install find

pvanheus:


pvanheus:
BUSCO uses `find` internally and relies on the `-delete` flag that is not present in the BusyBox version of `find`. This recipe adds a dependency on GNU findtools to ensure that an appropriate `find` is present for BUSCO to use.

recipes/busco/meta.yaml
```diff
@@ -29,10 +30,10 @@ test:
   requires:
     - wget
   commands:
-    - run_BUSCO.py -h
-    - run_busco -h
-    - generate_plot.py -h
-    - generate_plot -h
+    - run_BUSCO.py -h >/dev/null
```
bgruening: Why is this needed @pvanheus?

recipes/busco/run_test.sh
```diff
@@ -2,6 +2,9 @@
 
 set -euo pipefail
 
+# test that find has the -delete flag, which BUSCO uses
+mkdir findtest ; cd findtest ; touch file ; find . -delete ; cd ..
```
bgruening: Why can this not be in the meta.yaml?

recipes/busco/meta.yaml
pvanheus: It makes for less noisy test output - not needed, but it makes spotting the source of errors easier.

recipes/busco/run_test.sh
pvanheus: This whole shell command? I originally had it in meta.yaml, but this is rather verbose and thus since there was already a run_test.sh, I moved it there.

bgruening: The current Docker tests, do not use external run_test files. So all test that you also want to run in Docker should be in meta.yaml.

recipes/busco/meta.yaml
bgruening: But if the error message is "missing foobar" that is printed to stdout you are not seeing it, right?
Your call. I find more output good in tests.

bgruening:
Thanks!

PR #12725, opened 2018-12-18
add cromshell, a command-line interface to the cromwell workflow runner

notestaff:


recipes/cromshell/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -eu -o pipefail
+
+mkdir -p $PREFIX/bin
+cp cromshell $PREFIX/bin/
+
+outdir=$PREFIX/share/$PKG_NAME
```
epruesse: FHS would suggest `$PREFIX/share/doc/$PKG_NAME` - but do we even need this?

pvanheus:
@epruesse Ah you're right about the doc path. I also don't feel we need it though.

Issue #12586, opened 2018-12-12
bioconductor rebuild todo list


bgruening:
The following are packages that either need to be created or updated to have the gcc7 label for the bioconductor 3.8 rebuild to finish. Please note that packages in bioconda also need to have the `gcc7` label attached to them. The only way to do this currently is by building them in the `bulk` branch, so please direct any PRs there. Help either adding recipes to the `bulk` branch or on conda-forge would be greatly appreciated.

# Missing packages
  * recipes/bioconductor-hilbertvisgui needs gtkmm
  * libsbml package is missing

# R packages that need to be rebuilt
 - [ ] r-animation (depends on r-magick)
 - [ ] r-clustersim https://github.com/conda-forge/staged-recipes/pull/7567
 - [ ] r-dagitty (depends on r-v8 and, this, libv8)
 - [ ] r-evmix (needs an r-gsl rebuild)
 - [ ] r-feature (needs an r-rgl rebuild)
 - [ ] r-hiver (needs an r-rgl rebuild)
 - [ ] r-magick (needs an imagemagick rebuild)
 - [ ] r-pca3d (needs an r-rgl rebuild)
 - [x] r-rcdklibs https://github.com/conda-forge/staged-recipes/pull/7243
 - [ ] r-v8

dpryan79:
Thanks @bgruening, this will help a lot!

federicomarini:
Hey @bgruening I'll be happy to try to throw in some time before Christmas comes, provided no colleagues go on rampage-submission mode in the last days 😨

jenzopr:
Hi Björn,
for the packages intended to go to bioconda (r-gap, r-grain ,r-pcalg), the following dependencies are missing as well: `r-grbase`
Best,
Jens

brainstorm:
As mentioned in https://github.com/bioconda/bioconda-recipes/pull/12732#issuecomment-448897849, `r-bezier` would be nice to have, if https://github.com/conda-forge/staged-recipes/pull/7342 doesn't get it first.

bgruening:
@bioconda/r any help appreciated.

bgruening:
@nturaga do you have any insight in how often the bioc data packages are updated. They take a large amount of time and I'm wondering if we need to pin them to a specific R version. How many of them contain R version specific code. E.g. this one https://bioconductor.org/packages/release/data/annotation/html/Homo.sapiens.html could be probably unpinned to any R version, isn't it?

[...]

PR #12764, opened 2018-12-20
Added R package r-grbase

jenzopr:

xref #12586 and conda-forge/staged-recipes#7305

### DIFF
--- a/recipes/r-grbase/build.sh
+++ b/recipes/r-grbase/build.sh
@@ -1,0 +1,9 @@
+if [[ $target_platform =~ linux.* ]] || [[ $target_platform == win-32 ]] || [[ $target_platform == win-64 ]] || [[ $target_platform == osx-64 ]]; then
+  export DISABLE_AUTOBREW=1
+  mv DESCRIPTION DESCRIPTION.old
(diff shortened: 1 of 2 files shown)


dpryan79:
FYI, we've built this in `bulk` using gcc7 already if you want to use that instead: https://anaconda.org/bioconda/r-grbase/files

jenzopr:
I see! I can wait until `bulk` is merged then. Just thought that the list in #12586 is somehow relevant and can be worked on?

dpryan79:
Totally, if you could help getting the various CRAN packages that aren't appropriate for bioconda into conda-forge that'd be a HUGE help. I'll update #12586 to make this clearer, but the packages need to get a gcc7 label somewhere (either here or on conda-forge) in order for them to get used by the bioconductor 3.8 recipes. The only way to do that on the bioconda side is by building things on `bulk`, which is currently heavily tweaked to use conda-forge packages with the gcc7 label for dependency resolution and to then upload to anaconda.org with the gcc7 label. On the conda-forge side there's [a list](https://conda-forge.org/status/) of packages that still need to be rebuilt for gcc7, but that's thankfully close to being finished.

PR #12765, opened 2018-12-20
Add rrbgen

rwdavies:

### DIFF
--- a/recipes/rrbgen/meta.yaml
+++ b/recipes/rrbgen/meta.yaml
@@ -15,23 +15,21 @@ build:
 
 requirements:
   build:
-    - r-base >=3.5.0
   host:
-    - r-base >=3.5.0
+    - r-base
     - r-rcpp >=0.12.18
     - r-rcpparmadillo >=0.8.600.0.0
     - r-testthat >=2.0.0    
   run:
     - r-rcpp >=0.12.18
     - r-rcpparmadillo >=0.8.600.0.0
-    - r-base >=3.5.0
+    - r-base
     - r-testthat >=2.0.0
     
 test:
   commands:
     - $R -e "library('rrbgen')"  
     - $R -e "testthat::test_package('rrbgen', stop_on_failure = TRUE, stop_on_warning = TRUE)"
-
 
 about:
   home: https://github.com/rwdavies/rrbgen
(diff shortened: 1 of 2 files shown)


recipes/rrbgen/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  url: https://github.com/rwdavies/rrbgen/releases/download/{{ version }}/rrbgen_{{ version }}.tar.gz
+  sha256: '{{ sha256 }}'
+
+build:
+  number: 0
+  script: '$R CMD INSTALL --build --install-tests .'
+
+requirements:
+  build:
+    - r-base >=3.5.0
```
bgruening: This looks wired, have you used the skeleton generator for this?

rwdavies: I started with an existing meta.yaml from a different recipe and then modified as seemed appropriate until it passed the tests. I can re-do from a skeleton if this looks too weird

bgruening: That would be nice and please let me know which template you used, maybe we also need to fix this.

rwdavies: Unfortunately it seems CRAN package submission is offline until the New Year and this isn't currently in CRAN, so I can't seem to start directly from a skeleton.

I used a combination of https://github.com/bioconda/bioconda-recipes/blob/master/recipes/r-spp/meta.yaml (in the docs, suggested as a recipe for R package not on CRAN) and https://github.com/bioconda/bioconda-recipes/blob/master/recipes/abeona/meta.yaml (written by a friend). On re-review, neither have anything like this in the build section. I think I just added R in every part of the requirements section and it worked so I figured it was OK

Are these tests (the circleci ones) not comprehensive? My thought process when submitting this pull recipe was if anything was wrong, the tests would fail, and I could read the logs / re-submit until I understood what was supposed to go in every section

bgruening:
One last Q, since it is a R only package, should we name it `r-rrbgen`?

rwdavies:
Thanks for the changes, and yup, re-naming would be fine with me

bgruening:
@rwdavies can you rename it?

rwdavies:
Sure thing, changed now

bgruening:
Thanks!

PR #12780, opened 2018-12-21
Ensure zlib support in bioconductor-rhdf5.

johanneskoester:


recipes/bioconductor-rhdf5lib/meta.yaml
```diff
@@ -21,8 +23,12 @@ build:
 requirements:
   host:
     - r-base
+    - zlib
+    - libgfortran
```
bgruening: is `- {{ compiler('fortran') }}` not working?

johanneskoester: Didn't know this is equivalent. I thought it just wants to link against fortran, not building fortran code. But I might be wrong.

recipes/bioconductor-rhdf5lib/configure.patch
```diff
@@ -0,0 +1,11 @@
+--- configure	2018-12-20 17:09:38.001376321 +0100
++++ configure.patched	2018-12-20 17:25:42.834031679 +0100
```
dpryan79: You probably want `configure` rather than `configure.patched` there.

bgruening:
@johanneskoester do you know about a test that can ensure zlib support.

johanneskoester:
@bgruening unfortunately not. We could generate an hdf5 file that uses gz compression, but this package, as far as I know, only provides the C library, so it would be some effort to write a test.

recipes/bioconductor-rhdf5lib/configure.patch
johanneskoester: Isn't that only cosmetic?

johanneskoester:
@dpryan79 I don't think this is a problem in the way conda applies patches. The changes are definitely applied when looking at the logging output.

johanneskoester:
I have added a test case, let's see if it works like that.

johanneskoester:
@dpryan79 @bgruening  the test case works. Unfortunately, it reveals that although we now link rhdf5lib against zlib, there is still a problem opening that file. The requested dataset is definitely gzipped, so there should be no other filter causing a problem. One explanation could be that rhdf5lib still fails to open the right zlib.

saketkc:
Is this good to go @johanneskoester after @enneamer's suggested changes? I just bumped into https://github.com/pachterlab/sleuth/issues/120

PR #12801, opened 2018-12-26
Update ncbi-genome-download to 0.2.8

kblin:
Signed-off-by: Kai Blin <kblin@biosustain.dtu.dk>

### DIFF
--- a/recipes/ncbi-genome-download/build.sh
+++ b/recipes/ncbi-genome-download/build.sh
@@ -1,9 +1,0 @@
-#!/bin/bash
-
-$PYTHON setup.py install --single-version-externally-managed --record=record.txt
-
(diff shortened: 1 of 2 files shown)


pvanheus:
Hi @kblin, three minor comments that you might or might not want to address before merge:

1. The `sha256` can be moved to a variable (to be consistent with the version number, etc)
2. There are unnecessary comments in the `build.sh`. In fact the `build.sh` is a good candidate for removal with the single line command moved to `script:` within `meta.yaml`.
3. There are dependencies (e.g. `coverage`) in the test section that seem unnecessary given the tests that are actually run.

kblin:
Hi @pvanheus, I think this version should address all the comments.

PR #12809, opened 2018-12-26
Update libdeflate to 1.1

epruesse:
Update `libdeflate`: **1.0** &rarr; **1.1**

***

[Package Info](https://bioconda.github.io/recipes/libdeflate/README.html) | [Recipe Folder](https://github.com//bioconda/bioconda-recipes/tree/bump/libdeflate/recipes/libdeflate) | [Upstream Home](https://github.com/ebiggers/libdeflate) | [Upstream Releases](https://github.com/ebiggers/libdeflate/tags)

Upstream Author: `@ebiggers`

***


bgruening:
ping @bioconda/osx - can you please help?

epruesse:
Just out of curiosity - isn't this one for CF?

bgruening:
Yes, see my comment here: https://github.com/bioconda/bioconda-recipes/pull/8831

@kyleabeauchamp needs to decide I guess.

kyleabeauchamp:
My mild inclination is to keep it here for now because we haven't actually had any requests from conda-forge to migrate.  WDYT?

kyleabeauchamp:
Also, the pysam + htslib + samtools + libdeflate build chain is already somewhat challenging, so keeping libdeflate here might help us keep that thread from unravelling.

epruesse:
It just means that things will get interesting if anything else using `libdeflate` gets added to CF.

kyleabeauchamp:
Totally agree, I guess I'm betting on the possibility that no one ever asks for it in CF and thus we save on the up-front cost of migrating it :)

kyleabeauchamp:
Note to all: I've updated the title and body of this PR to note that the upgrade now targets version 1.2.

epruesse:
We pin this package. Do we need to update the pin?

bgruening:
We do not need to but it will effectively not be used if everything works es expected.

epruesse:
It won't.

**If** the package was pinning to `>=1.0, <2.0.0a0` we would see it used even without the pinning change. Since it pins to `>=1.0<1.1.0a0`, this package here will not get used.

PR #12817, opened 2018-12-26
Build perl-business-isbn-data with perl=5.22.0 (needed for prokka=1.12)

npavlovikj:

### DIFF
--- a/recipes/perl-business-isbn-data/conda_build_config.yaml
+++ b/recipes/perl-business-isbn-data/conda_build_config.yaml
@@ -1,0 +1,5 @@
+perl:
+  - 5.22.0
+pin_run_as_build:
+  perl:
+    max_pin: x
--- a/recipes/perl-business-isbn-data/meta.yaml
+++ b/recipes/perl-business-isbn-data/meta.yaml
@@ -11,16 +11,16 @@ source:
   sha256: {{ sha256 }}
 
 build:
-  number: 0
+  number: 1
 
 requirements:
   host:
-    - perl
+    - perl {{ perl }}
     - perl-extutils-makemaker
     - perl-carp
 
   run:
-    - perl
+    - {{ pin_compatible('perl') }}
     - perl-carp
 
 test:
--- a/recipes/perl-business-isbn-data/meta.yaml
+++ b/recipes/perl-business-isbn-data/meta.yaml
@@ -15,12 +15,12 @@ build:
 
 requirements:
   host:
-    - perl ==5.22.0
+    - perl 5.22.0
     - perl-extutils-makemaker
     - perl-carp
 
   run:
-    - perl ==5.22.0
+    - perl 5.22.0
     - perl-carp
 
 test:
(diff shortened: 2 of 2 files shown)


druvus:
I would recommend to rebuild prokka 1.12 to use Perl 5.26. Any specific reason you need an older prokka version?

npavlovikj:
Thanks @druvus , that sounds like a better idea. Should I create a subfolder 1.12 in prokka's recipe dir so I don't modify the 1.13 recipe?

druvus:
yes, create a subfolder 1.12 in prokka's recipe folder to built it using perl 5.26.

druvus:
merged in prokka 1.12 so closing this PR

npavlovikj:
@epruesse , Andreas suggested to re-build "prokka-1.12" with perl 5.26 and that fixed the dependency issues.

I am just curious about 2 things, so I would appreciate your input when you can.

1. If I really, really needed to use perl 5.22, is the content in this "conda_build_config.yaml" file what I am supposed to do? (the 1st commit is the most accurate, the others are just attempts).

2. Looks like in the "build" step of this PR, the right perl 5.22 dependencies are used:
**perl:                    5.22.0.1-0**          conda-forge
**perl-app-cpanminus:      1.7043-pl5.22.0_0**   bioconda   
**perl-carp:               1.38-pl5.22.0_0**     bioconda   
**perl-data-dumper:        2.161-pl5.22.0_0**    bioconda   
**perl-exporter:           5.72-pl5.22.0_0**     bioconda   
**perl-extutils-makemaker: 7.24-pl5.22.0_1**     bioconda   
**perl-test-more:          1.001002-pl5.22.0_0** bioconda 
However, in the "test" stage the used dependencies are from perl 5.26 instead:
libgcc-ng:               7.2.0-hdf63c60_3     conda-forge
**perl:                    5.26.2-h470a237_0**    conda-forge
**perl-business-isbn-data: 20140910.003-pl522_1** local      
**perl-carp:               1.38-pl526_1**         bioconda   
**perl-exporter:           5.72-pl526_1**         bioconda   
**perl-extutils-makemaker: 7.34-pl526_3**         bioconda 
and there is where the recipe fails.
Do you know what may be the reason for this?

As I said, I just want to know this out of curiosity, so don't feel obligated to answer :)

epruesse:
@npavlovikj Yes, the `conda_build_config.yaml` would probably be the way to go. The catch is that all binary modules for interpreted languages have to be (re-)built for the specific interpreter version, though, so setting the interpreter you use to something other than our current default will likely lead to issues elsewhere. You'd have to ask the conda-build experts on the conda-forge gitter channel to learn how something like this would have to be done.

The main interpreters, Perl, Lua, R and Python, are also treated specially by `conda` and `conda-build`. I can't tell you exactly how - the code is a little messy. This might explain why you see 5.26 being used by the tests (or mulled-test - we have two test stages).

In the end, it's probably quicker to patch upstream code to work with 5.26 than to make the 5.22 thing work out right. Even if you get it to pass the tests, we might still end up with users getting broken environment installations.

npavlovikj:
@epruesse , many thanks for the detailed response - this is really useful to know.

PR #12847, opened 2018-12-28
RNAz fix

jfallmann:
Added fixes from https://github.com/wash/rnaz/issues/7#issue-30859115


jfallmann:
@eggzilla Yiiiihaaaa, ready to merge when you are ;)

recipes/rnaz/meta.yaml
```diff
@@ -7,12 +7,17 @@ source:
...
   skip: True # [osx]
 
+patches:
+  - "rnaz.patch"
+  
 requirements:
   build:
     - {{ compiler('cxx') }}
+  host:
+    - libsvm =3.21
```
bgruening: Is the pinning needer here?

bgruening: I see it is, because it is hard coded in your patch.

recipes/rnaz/build.sh
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
-./configure --prefix=$PREFIX --datadir=$PREFIX/share
+./configure CFLAGS="-std=gnu89 -g -O2" --prefix=$PREFIX --datadir=$PREFIX/share
 make
 make install
-cp $PREFIX/share/RNAz/perl/*.pl $PREFIX
```
bgruening: we could symlink the perl scripts that are needed into $PREFIX/bin isn't it?

recipes/rnaz/meta.yaml
jfallmann: Yeah, has to be that way, its hard coded into RNAz, if you not provide the exact version it doesn't run, also the compiler options have to be set, but it would need a major overhaul of the source code to fix that

recipes/rnaz/build.sh
jfallmann: but then we also need to make the RNAz.pm available in the $PERL5LIB variable, or copy that as well to $PREFIX/bin, or have perl in a pinned version as dependency so I know what $PREFIX/lib/perl5/site-perl version to use

jfallmann:
@eggzilla @bgruening Works now, also perl scripts, should be good to go

PR #12890, opened 2019-01-03
MultiQC: add DOI in identifiers

ewels:

Adding a DOI identifier to the MultiQC recipe, as discussed in https://github.com/nf-core/tools/issues/236#issuecomment-451091256


bgruening:
@ewels this will not create a new package as you have not increased the build number. So it will not end up on anaconda.org. As soon as you release a new version or increase the build number it will be uploaded.

ewels:
Ah good point - I'll push a new commit with a bumped build number then if that's ok.

PR #12901, opened 2019-01-03
Bump plant_tribes_assembly_post_processor to version 1.0.4

gregvonkuster:


recipes/plant_tribes_assembly_post_processor/1.0.4/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  skip: True # [osx]
+
+requirements:
+  run:
+    - cap3
+    - genometools-genometools >=1,<2
+    - hmmer >=3
+    - mafft >=7,<8
+    - perl >=5.22
+    - perl-estscan2 2.1 pl5.22.0_1
```
bgruening: @gregvonkuster is this really needed with the strict pinning? We try to get rid of those old perl version.

PR #12923, opened 2019-01-05
a few updates

dpryan79:

### DIFF
--- a/recipes/r-mess/meta.yaml
+++ b/recipes/r-mess/meta.yaml
@@ -1,0 +1,59 @@
+{% set version = '0.5.4' %}
+{% set posix = 'm2-' if win else '' %}
+{% set native = 'm2w64-' if win else '' %}
+
+package:
+  name: r-mess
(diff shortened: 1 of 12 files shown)


recipes/r-mess/meta.yaml
```diff
@@ -0,0 +1,59 @@
+{% set version = '0.5.4' %}
```
bgruening: conda-forge?

dpryan79: It's from the wrapper script, yes.

bgruening: No I mean should this go to conda-forge :)

dpryan79: It has requirements in bioconda (r-kinship2), so no.

bgruening: Mh, this will bite us sooner or later. Another example where it would be better to move all r-* packages which have no bioconductor dependencies into conda-forge.

PR #12924, opened 2019-01-05
bumb build-numbers and keep bulk busy

bgruening:
extracted from the last OSX build errors
@dpryan79 will kill Linux and just keep OSX running for the moment.


dpryan79:
:+1:

dpryan79:
Please let bioconductor-champ finish before merging, since it takes about an hour to build it.

bgruening:
I'm not going to kill the previous Linux run, but the run that is triggered by this PR. `champ` is also available now :)

PR #12943, opened 2019-01-07
Added SvABA recipe 

micknudsen:


micknudsen:
SVaBA is an excellent tool, which I think deserves to be in Bioconda. However, I cleary do not have the required skills to write a working recipe (it works perfectly if I repeat the steps locally). If anybody could give me a hint on how to move on (see https://github.com/walaj/svaba/issues/43), I would be very happy.

How does one apply labels? I see that some PRs have a "Help Wanted" label, but I am not able to add it.

recipes/svaba/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+source:
+  url: https://github.com/walaj/svaba/archive/{{ version }}.tar.gz
+  sha256: d2381389a4e2420ad55d38c40bd57b91ad5619db08a0bdf0ec2321e9979bd12c
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - zlib
```
bgruening: that is not necessary, to have zlib in host should be enough.

recipes/svaba/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  host:
+    - backports.lzma
+    - bzip2
+    - zlib
+  run:
+
+about:
+  home: https://github.com/walaj/svaba
+  license: GPL-3
```
bgruening: please add a license_file

recipes/svaba/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+rm -fr SeqLib
+git clone --recursive https://github.com/walaj/SeqLib.git
```
bgruening: why do you need this special seqlib version?

micknudsen: When downloading the [release](https://github.com/walaj/svaba/archive/1.1.0.tar.gz), SeqLib is not included. There is just an empty folder. I just wanted to make sure that the version matched the one at the time the release was created. Maybe that is not necessary.

recipes/svaba/meta.yaml
micknudsen: Done

bgruening:
I'm not sure, but it can be that you need to patch the Makefile to convince it to find zlib. Maybe the path is hardcoded. That is sadly a common case :(

PR #12989, opened 2019-01-08
Update PopDel to version 1.0.8

serosko:
* [yes] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [no] This PR adds a new recipe.
* [yes] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).
* [yes] This PR updates an existing recipe.
* [no] This PR does something else (explain below).

### DIFF
--- a/recipes/bioconductor-deqms/meta.yaml
+++ b/recipes/bioconductor-deqms/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "1.0.0" %}
+{% set version = "1.0.1" %}
 {% set name = "DEqMS" %}
 {% set bioc = "3.8" %}
 
@@ -10,7 +10,7 @@ source:
     - 'https://bioconductor.org/packages/{{ bioc }}/bioc/src/contrib/{{ name }}_{{ version }}.tar.gz'
     - 'https://bioarchive.galaxyproject.org/{{ name }}_{{ version }}.tar.gz'
     - 'https://depot.galaxyproject.org/software/bioconductor-{{ name|lower }}/bioconductor-{{ name|lower }}_{{ version }}_src_all.tar.gz'
-  md5: 43415cf52b11a83771f6eb5b308cce83
+  sha256: e1cc1f30d361427422477667d5f6e8c4d55fefcc5357903df009f603bcaec9a0
 build:
   number: 0
   rpaths:
--- a/recipes/bioconductor-rhdf5/meta.yaml
+++ b/recipes/bioconductor-rhdf5/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "2.26.1" %}
+{% set version = "2.26.2" %}
 {% set name = "rhdf5" %}
 {% set bioc = "3.8" %}
 
@@ -10,7 +10,7 @@ source:
     - 'https://bioconductor.org/packages/{{ bioc }}/bioc/src/contrib/{{ name }}_{{ version }}.tar.gz'
     - 'https://bioarchive.galaxyproject.org/{{ name }}_{{ version }}.tar.gz'
     - 'https://depot.galaxyproject.org/software/bioconductor-{{ name }}/bioconductor-{{ name }}_{{ version }}_src_all.tar.gz'
-  sha256: e3c53bddf5f892ea8e51363bca750a09c08b90de0a44b51d261e46f80545b0e2
+  sha256: 34f99cc588c989cd213d896b25e77de2f26c9ca7f79d405cf4bbf211e6e0f383
 build:
   number: 0
   rpaths:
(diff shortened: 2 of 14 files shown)


serosko:
Before merging I have an adeitional request for advice:
I observed that the conda binary of PopDel is 50% slower compared to the binary I get from compiling the code from the github repo. The apparent reasons for this is the different flags the conda build environment applies. 
Compiling from github happens whith the following command line:
`g++ -I. -DSEQAN_HAS_ZLIB=1 -std=c++14 -DSEQAN_DISABLE_VERSION_CHECK -DDATE=\""on 2019-01-08"\" -DVERSION=\""1.0.8"\" -W -Wall -Wno-long-long -pedantic -Wno-variadic-macros -Wno-unused-result -O3 -DSEQAN_ENABLE_TESTING=0 -DSEQAN_ENABLE_DEBUG=0   -c -o popdel.o popdel.cpp`
While compilation withon conda-build looks like this:
`/home/sroskosch/miniconda3/conda-bld/popdel_1546945217822/_build_env/bin/x86_64-conda_cos6-linux-gnu-c++ -fvisibility-inlines-hidden -std=c++17 -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -I/home/sroskosch/miniconda3/conda-bld/popdel_1546945217822/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_/include -fdebug-prefix-map=/home/sroskosch/miniconda3/conda-bld/popdel_1546945217822/work=/usr/local/src/conda/popdel-1.0.8 -fdebug-prefix-map=/home/sroskosch/miniconda3/conda-bld/popdel_1546945217822/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_=/usr/local/src/conda-prefix -I. -DSEQAN_HAS_ZLIB=1 -std=c++14 -DSEQAN_DISABLE_VERSION_CHECK -DDATE=\""on 2019-01-08"\" -DVERSION=\""1.0.8"\" -W -Wall -Wno-long-long -pedantic -Wno-variadic-macros -Wno-unused-result -O3 -DSEQAN_ENABLE_TESTING=0 -DSEQAN_ENABLE_DEBUG=0 -DNDEBUG -D_FORTIFY_SOURCE=2 -O2  -c -o popdel.o popdel.cpp`
Is there any way I can prevent conda from appending -D_FORTIFY_SOURCE=2 and -O2, as I believe this to be the cause of the massive performance loss.

epruesse:
@serosko I think the update went through in another PR already. The issue of speed is one that affects all of Bioconda. It may actually be the compiler version and not the specific flags that cause the difference in performance. We have not yet migrated to gcc7 and the older gcc we use isn't as good at vectorizing as the newer versions.

epruesse:
You could in theory override the flags from `build.sh`. You'd have to reproduce the build environment locally though to test this and benchmark the package. I doubt it's FORTIFY_SOURCE - that just adds tiny checks around `strcpy` and friends to guard against buffer overflows. The O3 may make a difference, but it's not 50% usually. Hence my thought that it's the vectorization in gcc4.8 not understanding the relevant loops properly.

serosko:
Thank you for your advice. But are you sure about the compiler version? The program requires gcc4.9 or higher. gcc4.8 should not work at all when compiling it. Additionally, when I spin up "conda-build popdel" it looks like gcc 7.3 is beeing used:

\## Package Plan \##

  environment location: /home/sroskosch/miniconda3/conda-bld/popdel_1547020560015/_build_env


    binutils_impl_linux-64: 2.31.1-h6176602_1
    binutils_linux-64:      2.31.1-h6176602_6
    gcc_impl_linux-64:      7.3.0-habb00fd_1 
    gcc_linux-64:           7.3.0-h553295d_6 
    gxx_impl_linux-64:      7.3.0-hdf63c60_1 
    gxx_linux-64:           7.3.0-h553295d_6 
    libgcc-ng:              8.2.0-hdf63c60_1 
    libstdcxx-ng:           8.2.0-hdf63c60_1 

Also the compiler flags hint to a recent version of gcc, otherwise the flag "-std=c++17" would not work. I would like to try the approach of overwriting the flags in build.sh. I tried to replace the call to make with "make CXXFLAGS=..." but this only seems to overwrite the flags the are defined in the makefile. The last 3 flags ("-DNDEBUG -D_FORTIFY_SOURCE=2 -O2") get appended nevertheless. Do you have any idea how to get rid of them or where I can read more about this?

bgruening:
This needs a rebase against master.

serosko:
This should take care of the rebasing. But the originial problem remains.

PR #13000, opened 2019-01-09
Frogs3.0.0

ValentinMarcon:

(Poke @mariabernard @oinizan )


recipes/frogs/meta.yaml
```diff
@@ -1,53 +1,67 @@
...
+    md5: 1a72f44e82517b6839a82cd677b98858 
+  - url: https://github.com/mahendra-mariadassou/phyloseq-extended/archive/v{{ phyloseq_extended_version }}.tar.gz
+    md5: 1ec072751503eebb791e876a60433120
+    folder: lib/external-lib
 
 requirements:
   host:
     - python
-    - rdptools >=2.0.2
+    - rdptools 2.0.2
```
bgruening: if possible do not pin that strict and do not pin mpi or openmpi we do this globally

ValentinMarcon: Ok so I removed that dependencies

recipes/frogs/meta.yaml
```diff
@@ -1,53 +1,62 @@
...
   name: {{ name|lower }}
   version: {{ version }}
 
 build:
-  number: 3
+  number: 0
   skip: true  # [py3k]
 
+# 1) Uplaod FROGS source code
+# 2) Add phyloseq-extented lib from Mahendra Mariadassou to FROGS (lib/external-lib) :
```
epruesse: Do we need those comments?

recipes/frogs/meta.yaml
```diff
@@ -1,53 +1,62 @@
...
 source:
-  url: https://github.com/oinizan/FROGS/archive/v{{ version }}.tar.gz
-  md5: da41b0950a750be2038cefa717edf9df
+  - url: https://github.com/geraldinepascal/FROGS/archive/v{{ version }}.tar.gz 
+    md5: 1a72f44e82517b6839a82cd677b98858 
+  - url: https://github.com/mahendra-mariadassou/phyloseq-extended/archive/v{{ phyloseq_extended_version }}.tar.gz
+    md5: 1ec072751503eebb791e876a60433120
+    folder: lib/external-lib
```
epruesse: Is there a way to have phyloseq-extended as a separate package?

recipes/frogs/meta.yaml
```diff
@@ -1,53 +1,62 @@
...
-    - swarm >=2.1.13
-    - vsearch >=2.4.3
-    - blast >=2.2.31
-    - rdptools >=2.0.2
+    - perl-io-gzip 0.20
+    - flash 1.2.11
+    - cutadapt 1.18
+    - swarm 2.2.2
+    - vsearch 2.9.1
+    - blast 2.7.1
```
epruesse: Those tight pins can make it difficult to find a solution later on. Wouldn't e.g. `blast 2.7.*` suffice?

recipes/frogs/meta.yaml
ValentinMarcon: Not really because this is really specific to Frogs

recipes/frogs/meta.yaml
ValentinMarcon: We have chosen to pin **all** the dependencies version to ensure we will **always** have the same results with this version of FROGS

recipes/frogs/meta.yaml
ValentinMarcon: Why not (except for the mistyping) ?

ValentinMarcon:
Hello,

According to your advices @bgruening we decided to unpin the dependencies.
Hope this time all the recipe will be fine 🤞 
Cheers

( @epruesse  @mariabernard )

PR #13002, opened 2019-01-09
Bulk update 090119b

dpryan79:


recipes/bioconductor-dexseq/meta.yaml
```diff
@@ -57,6 +57,8 @@ requirements:
     - r-rcolorbrewer
     - r-statmod
     - r-stringr
+    - python <3
+    - 'htseq >=0.6.1,<0.10.0'  # dexseq_count.py performes an overly simplistic version check
 test:
```
dpryan79: @nturaga Can you relay `python 2` and `htseq` as system requirements upstream for the dexseq package? We noticed this during the 3.7 release as well. Those are both needed for the counting script to work, which is pretty integral for actually using the package.

PR #13029, opened 2019-01-11
Islandpath

innovate-invent:

### DIFF
--- a/recipes/islandpath/build.sh
+++ b/recipes/islandpath/build.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+mkdir -p $PREFIX/islandpath
+cp -r . $PREFIX/islandpath
+mkdir -p $PREFIX/bin
+ln -s $PREFIX/islandpath/Dimob.pl $PREFIX/bin/islandpath
+chmod +x $PREFIX/bin/*
--- a/recipes/islandpath/build.sh
+++ b/recipes/islandpath/build.sh
@@ -4,3 +4,5 @@ cp -r . $PREFIX/islandpath
 mkdir -p $PREFIX/bin
 ln -s $PREFIX/islandpath/Dimob.pl $PREFIX/bin/islandpath
 chmod +x $PREFIX/bin/*
+cpanm --force TAP::Harness::Env ExtUtils::Helpers
+cpanm MooseX::Singleton
--- a/recipes/islandpath/meta.yaml
+++ b/recipes/islandpath/meta.yaml
@@ -1,28 +1,34 @@
-{% set version_num = "1.0.0" %}
-{% set version = "v{{ version_num }}" %}
+{% set version = "v1.0.3" %}
 
 package:
   name: islandpath
-  version: {{ version }}
+  version: {{version}}
 
 source:
-  fn: islandpath-{{ version_num }}.tar.gz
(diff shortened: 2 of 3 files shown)


recipes/islandpath/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+  name: islandpath
+  version: {{version}}
+
+source:
+  #  git_url: https://github.com/brinkmanlab/islandpath
+  url: https://github.com/brinkmanlab/islandpath/archive/{{version}}.tar.gz
+  sha256: ac9198240fe1674b876fc0ab4c461e5a34799ddba521481eb2bb15324c2845bd
+  
+requirements:
+  build:
```
bgruening: This needs to be `host`

recipes/islandpath/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - hmmer
+    - perl-bioperl 
+    - perl-data-dumper 
+    - perl-log-log4perl 
+    - perl-config-simple 
+    - perl-moose 
+    #- perl-moosex-singleton
+
+test:
+  source_files:
```
bgruening: This is currently not supported in Bioconda, you can not user a file as test.

innovate-invent: Removed

innovate-invent:
I am not sure how to correct the error for macos. 
`Can't write to cpanm home 'UNKNOWN/.cpanm': You should fix it with chown/chmod first.`
This almost seems like a build environment issue.

recipes/islandpath/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+mkdir -p $PREFIX/islandpath
+cp -r . $PREFIX/islandpath
+mkdir -p $PREFIX/bin
+ln -s $PREFIX/islandpath/Dimob.pl $PREFIX/bin/islandpath
+chmod +x $PREFIX/bin/*
+export HOME=/tmp 
+cpanm --force TAP::Harness::Env ExtUtils::Helpers
+cpanm MooseX::Singleton
```
bgruening: ideally this should be a separate perl package which you depend on

innovate-invent:
@druvus Did you create the Singleton package? How did you resolve the dependency issues I was having?

druvus:
The missing perl-moosex-singleton package was added in #13099. Also added perl as a dependency with same reqs in host and build.

recipes/islandpath/build.sh
druvus: perl-moosex-singleton  is now added as a bioconda package

innovate-invent:
What about TAP::Harness::Env and ExtUtils::Helpers failing?

druvus:
The main issue was that you didn't listed perl as dependency in host and run. It then used perl 5.22 instead of 5.26 and started to pull in older builds.

innovate-invent:
Thats odd, I originally listed it and it forced perl 5.22 ¯\_(ツ)_/¯
Thanks for the help!

druvus:
Happy to help! 

perl packages from early days of Bioconda (2015ish) does not have perl version tagged so those packages sometimes crashes the dependencies (i.e. older bioperl). An easy way to spot issues is when perl-threaded got pulled in.

PR #13081, opened 2019-01-15
Update plantcv recipe to v3.0.4

nfahlgren:


recipes/plantcv/meta.yaml
```diff
@@ -2,7 +2,7 @@
 {% set version = "3.0.3" %}
```
bgruening: should that be 3.0.4?

nfahlgren: Yes, good catch!

nfahlgren: Fixed now, look good otherwise?

bgruening:
@nfahlgren what I was wondering is why did if pass before? The URL is changing but the checksum stays?

nfahlgren:
@bgruening hmm, not sure. The only thing I changed in the YAML file, in addition to adding the `LICENSE` file, was the `{% set version = "3.0.4" %}` and the `{% set hash_value = "2a120bc5923cd0d273e762bbdc21fef88fa384fc2d5427b671f08d1b71e93b87" %}`. That sha256 checksum is definitely the checksum found for the source distribution at https://pypi.org/project/plantcv/#files and is different than the checksum for 3.0.3 `e8524b622379cf3df03fa1a9a87b7f754b1942329581cc7d94e9d93a3793ca4f`

epruesse:
@nfahlgren Can you delete the 3.0.3 folder? I don't think we need that.

nfahlgren:
@epruesse for sure, I removed the 3.0.3 folder. I updated the recipe to PlantCV v3.0.5 because we were having issues with the PyPI 3.0.4 version. I didn't mean to push these updates yet though because local tests were giving me trouble, so not sure the circleci tests are going to pass yet.

nfahlgren:
There was just a problem with my local environment, tests pass locally again without changing the branch, and looks like it passes on CircleCI too.

PR #13085, opened 2019-01-15
pb-assembly avro-python2==1.8.0

pb-cdunn:
https://github.com/gconcepcion/pb-assembly/issues/1

### DIFF
--- a/recipes/pb-assembly/meta.yaml
+++ b/recipes/pb-assembly/meta.yaml
@@ -2,7 +2,7 @@ package:
   name: pb-assembly
   version: "0.0.3"
 build:
-  number: 1
+  number: 2
   script: echo noop
   skip: True # [not py27 or osx]
 requirements:
@@ -25,6 +25,7 @@ requirements:
     - bedtools
     - bwa
     - numpy
+    - avro-python2==1.8.0
 test:
   imports:
     - falcon_unzip


pb-dseifert:
@bgruening 
```
conda.exceptions.ResolvePackageNotFound: 
  - toolchain_c_linux-64
```
That seems like a very weird exception and something being very off. Any ideas?

pb-dseifert:
@epruesse could the channel order in https://github.com/bioconda/bioconda-utils/commit/b589da8fbebdfde9ebfcdeaa637c1ae95a5337e1#r31952703 explain these weird toolchain related failures?

epruesse:
@pb-cdunn @pb-dseifert The problem is very likely due to the label switch at conda-forge. 

@bgruening - thoughts?

pb-cdunn:
Same issue as #13102

pb-cdunn:
@pb-dseifert , switching from `c-compiler` to `cxx-compiler` produced *almost* the same error:
```
conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform linux-64: {'toolchain_cxx_linux-64'}
```

epruesse:
@pb-cdunn @pb-dseifert  conda-forge just did their "label switch". They had been rebuilding for gcc7 for a while now and just changed the labels so that everything now installs gcc7 dependencies. At the same time, they also changed the labels for conda-forge-pinnings, which brings the conda build config we also use. Your issue here is very likely caused by the change to conda-forge-pinnings, which changes the compiler packages. 

See https://github.com/bioconda/bioconda-utils/pull/390

PR #13093, opened 2019-01-15
Remove htslib/samtools/bcftools dependencies

mvdbeek:
This fixes https://github.com/pysam-developers/pysam/issues/640
which is an issue with the bioconda build of pysam.
Removing those dependencies will make pysam use the bundled
subset of htslib.
A side-effect of this is that pysam and samtools/bcftools/htslib
versions are now independent and can be mixed at will.


mvdbeek:
> HTSLib built on Bioconda uses a faster ZLIB implementation (libdeflate) and I'm not sure if the vanilla pysam build will grab it.

I think it does, it's unfortunate that we don't get the full build log, but you can see libdeflate being used in the failed previous build: https://circleci.com/gh/bioconda/bioconda-recipes/40770?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

bgruening:
@mvdbeek is this still needed? Want to update the PR?

mvdbeek:
Not for the bug, which should be fixed, but I don't think it's reasonable to ship samtools with pysam -- the wheels don't come with samtools either.

bgruening:
Then please merge it with master. @kyleabeauchamp ok?

kyleabeauchamp:
Fine with me, but ping @AndreasHeger to confirm,

AndreasHeger:
Yes, removing samtools/bcftools should be fine - they are test requirements only.

I don't fully understand the htslib removal but if you are confident it will work I have no objections.

dpryan79:
If I didn't bork anything while handling the merge conflict then please merge :)

kyleabeauchamp:
Seems good still, as long as the build is green

PR #13095, opened 2019-01-15
Stacks 2.3

lecorguille:


lecorguille:
Oups, I did it again

lecorguille:
I was lazy and removed 1.47 but if for any reason someone want that old versions of stacks survive, I can make the effort.
On the contrary, should we keep the other versions?

bgruening:
>  On the contrary, should we keep the other versions?

That is entirely up the maintainers.

PR #13098, opened 2019-01-15
Add PICRUSt2

apcamargo:

### DIFF
--- a/recipes/picrust2/meta.yaml
+++ b/recipes/picrust2/meta.yaml
@@ -12,6 +12,7 @@ source:
 build:
   number: 0
   script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed -vv "
+  noarch: python
 
 requirements:
   host:
@@ -27,6 +28,7 @@ requirements:
     - glpk
     - hmmer
     - epa-ng
+    - gappa
     - biom-format
     - pytest
     - pytest-cov
--- a/recipes/picrust2/meta.yaml
+++ b/recipes/picrust2/meta.yaml
@@ -17,7 +17,7 @@ build:
 requirements:
   host:
     - pip
-    - python <3.7
+    - python >=3.0.0,<3.7
   run:
     - python <3.7
     - r-base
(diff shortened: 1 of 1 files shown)


apcamargo:
@gavinmdouglas Are you ok with this recipe?

gavinmdouglas:
Thanks for writing this recipe @apcamargo ! 

Could you please change the test section to be:

```
test:
  source_files:
    - test
  commands:
    - picrust2_pipeline.py -h
    - pytest
```

Thanks again and I'm pretty amazed that this was recipe was produced so quickly after GAPPA was added to bioconda...

apcamargo:
You're welcome!

It seems `mp_pred_out.tsv` and `test_predicted_func.tsv` are not being found. Do you know what's going on?

gavinmdouglas:
Ah I see now - the test scripts were identifying the path to the testfiles based on where the code was located. I can fix this to be compatible with conda with a future release. For now it would be easiest to remove that extra test.

apcamargo:
Ok. I'll revert the test section and merge the recipe. Thanks!

gavinmdouglas:
Hey @apcamargo , after testing the conda install it seems none of the database files or accompanying Rscripts were installed in ```lib/python3.6/site-packages/picrust2``` as I expected. Is there an easy way to specify that certain directories need to be kept as well? Based on my googling it seems like a manifest file could be used.

apcamargo:
I guess there is, but I'm unsure if the database files should be included in the recipe. I'd prefer to include them (and the R scripts) to make it easier for the user to install, but this may be against bioconda's practices.

@bioconda/core what should be done in this situation?

dpryan79:
How big are the files in question? For larger R packages (think all of the big bioconductor annotation packages) we use a trick where the database is installed as a post-link script, that way it's doesn't make a 100+MB tarball that we have to host on anaconda.org. See the following for an example: https://github.com/bioconda/bioconda-recipes/blob/bulk/recipes/bioconductor-txdb.mmusculus.ucsc.mm10.ensgene/post-link.sh

gavinmdouglas:
The additional files are ~150 MB. Ok that's good to know about, thanks.

apcamargo:
@gavinmdouglas I'll be unable to create this PR in the next few weeks (lots of deadlines approaching). How do you think we should proceed? Can you create the PR?

PR #13104, opened 2019-01-16
Add recipe for GraphAligner

maickrau:

@bioconda/core I would like a review as this is my first recipe and the pull request adds a new package.

### DIFF
--- a/recipes/graphaligner/meta.yaml
+++ b/recipes/graphaligner/meta.yaml
@@ -9,7 +9,6 @@ source:
 
 build:
   number: 0
-  skip: False # [linux]
   skip: True # [osx]
 
 requirements:
@@ -25,8 +24,6 @@ requirements:
     - pkg-config
     - libdivsufsort
     - protobuf
-  run:
-    - libdivsufsort
 
 test:
   commands:


recipes/graphaligner/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+  version: 1.0.0
+
+source:
+  url: https://github.com/maickrau/GraphAligner/files/2764381/GraphAligner.tar.gz
+  sha256: 6bbebd3c7373c2018245391c8ecdc97cc58e2afa82d3523e08afd0407df1bf63
+  patches: version.patch
+
+build:
+  number: 0
+  skip: False # [linux]
```
bgruening: you can remove this line

recipes/graphaligner/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - {{ compiler('cxx') }}
+  host:
+    - jemalloc
+    - zlib
+    - mummer4
+    - boost
+    - libboost
+    - sparsehash
+    - pkg-config
+    - libdivsufsort
```
bgruening: Do you need to pin this? Is the library API stable?

recipes/graphaligner/meta.yaml
maickrau: Thanks, removed

recipes/graphaligner/meta.yaml
maickrau: Turns out that libdivsufsort was getting statically linked even though the makefile suggests that it should be dynamically linked. I removed it from the run: section and it still seems to work.

PR #13121, opened 2019-01-17
[WIP] Jbrowse: build with plugins

abretaud:

Here's an attempt to build JBrowse with a few plugins needed for [Galaxy tool](https://github.com/galaxyproject/tools-iuc/pull/1683)
Not tested yet, not sure it's the right approach until we have a solution for https://github.com/GMOD/jbrowse/issues/1288

@erasche I could add Combotrack and themes too, but I guess if I do they will be enabled by default?

### DIFF
--- a/recipes/jbrowse/build.sh
+++ b/recipes/jbrowse/build.sh
@@ -10,10 +10,20 @@ chmod a+x $PREFIX/etc/conda/deactivate.d/jbrowse-sourcedir.sh
 
 cd $SRC_DIR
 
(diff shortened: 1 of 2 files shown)


erasche:
Nah no need to add the themes. Combo view would be useful since it's a
really nice feature to expose the dataset provenance in the jbrowse instance

Op do 17 jan. 2019 17:55 schreef Anthony Bretaudeau <
notifications@github.com:

PR #13129, opened 2019-01-18
Bump Transabyss to v2.0.1 fix missing version bump in meta.yml

winni2k:

### DIFF
--- a/recipes/transabyss/meta.yaml
+++ b/recipes/transabyss/meta.yaml
@@ -9,7 +9,7 @@ build:
 
 package:
   name: transabyss
-  version: '1.5.5'
+  version: '2.0.1'
 
 requirements:
   host:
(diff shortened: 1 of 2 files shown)


recipes/transabyss/meta.yaml
```diff
@@ -9,24 +9,24 @@ build:
...
 package:
   name: transabyss
-  version: '1.5.5'
+  version: '2.0.1'
 
 requirements:
   host:
     - python
   run:
     - zlib
```
bgruening: why is zlib needed if we do not compile anything, same for libiconv

winni2k: No idea. I am not a maintainer of this code, just a user. 

Happy to remove these and see how things go.

PR #13139, opened 2019-01-19
Update nanopolish to v0.11.0

dfornika:


dfornika:
I'm not completely sure that the patch for this recipe won't need to be updated.

dfornika:
Looks like the patch will need to be updated.

druvus:
Updated the patch in #13169 and it seems to pass now.

PR #13173, opened 2019-01-22
PICRUSt2 install fix

gavinmdouglas:


gavinmdouglas:
@apcamargo quick note that I've added a PR that should address the issue we were discussing in PR https://github.com/bioconda/bioconda-recipes/pull/13098

apcamargo:
Hey @gavinmdouglas , I see that you're working in the installation fix. Thanks!

I think we should also submit a recipe of an older version of `epa-ng`, to avoid the segmentation fault issue.

gavinmdouglas:
Hi @apcamargo, yes I think I'm close to getting it working now.

That would be difficult to add an older version of EPA-NG to bioconda since several code fixes (added in later versions) were required for it to build correctly on macos (see https://github.com/Pbdas/epa-ng/issues/22).  I think it will be better to simply bump the version of EPA-NG (or PICRUSt2) once the cause of the segmentation fault is identified and fixed.

gavinmdouglas:
Thanks for the help @dpryan79 !

PR #13191, opened 2019-01-23
update wtdbg2 to v2.3

mjsull:

### DIFF
--- a/recipes/wtdbg/build.sh
+++ b/recipes/wtdbg/build.sh
@@ -2,5 +2,8 @@
 
 mkdir -p $PREFIX/bin
 
+export CFLAGS="-I$PREFIX/include"
+export LDFLAGS="-L$PREFIX/lib"
+
 make
 cp wtdbg2 wtdbg-cns wtpoa-cns $PREFIX/bin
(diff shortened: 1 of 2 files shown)


mjsull:
@bioconda I am getting a fatal error: zlib.h: No such file or directory, I've tried the suggestions in troubleshooting but it hasn't fixed the problem.

recipes/wtdbg/meta.yaml
```diff
@@ -15,6 +15,8 @@ build:
 requirements:
   build:
     - {{ compiler('c') }}
+  host:
+    - zlib
```
thanhleviet: You missed 
```
run:
- zlib 
```

recipes/wtdbg/build.sh
```diff
@@ -2,5 +2,8 @@
 
 mkdir -p $PREFIX/bin
 
+export CFLAGS="-I$PREFIX/include"
+export LDFLAGS="-L$PREFIX/lib"
```
thanhleviet: Replace line 5,6 with these:

export C_INCLUDE_PATH=${PREFIX}/include
export LIBRARY_PATH=${PREFIX}/lib

recipes/wtdbg/meta.yaml
thanhleviet: Ah no, you don't need it. Just replace 2 lines in build.sh as the tool does not use ./configure

lh3:
Didn't notice this PR an hour earlier than my #13194. My PR has passed the linux test.

PR #13196, opened 2019-01-23
Update Picard to 2.18.25.

cbrueffer:


epruesse:
Does anything inside of Bioconda depend on the old version that won't build anymore?

epruesse:
Looking at the depends:
- ctat-mutations is hard pinned to 2.18.14
- cap-mirseq needs >=1.97

The others all just require picard. So IMHO, all subfolders but 2.18.14 can go. And even for that one I have the feeling that ctat-mutations is overpinned.

epruesse:
Do you know how upstream keeps their versions? I vote for removing all sub recipes and making ctat-mutations depend on >=2.18.14,<2.19 - unless upstream has a history of changing API at will, which I sort of doubt.

cbrueffer:
No idea, but the packages themselves remain available, so the 2.18.14 dependency will be satisfied. Thanks for having a look!

PR #13199, opened 2019-01-23
bump nanopolish

mmiladi:


jts:
Thanks. Happy to help provide some extra flags in the Makefile to remove the need for the patch.

If I change `LIBS=-lz` to `LIBS+=-lz` you should be able to pass in the extra libraries in the `make` command. 

Is removing the HTS check in the patch necessary since you pass `HTS=`?

PR #13221, opened 2019-01-25
RawTools: new recipe

CarlosHorro:

The recipe seems to work but I am not being able (yet) to execute full tests on it.


recipes/rawtools/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+    license: Apache License, Version 2.0
+    summary: RawTools is an open-source and freely available package designed to perform scan data parsing and quantification, and quality control analysis of Thermo Orbitrap raw mass spectrometer files.
+
+package:
+    name: rawtools
+    version: {{ version }}
+
+build:
+  noarch: generic
+  number: 1
```
bgruening: 0

recipes/rawtools/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+requirements:
+  build:
+  run:
+    - mono >=5.0.0
+
+
+test:
+    commands:
+#     - bash -c 'rawtools.sh "$@"' parse --help
```
bgruening: can you include this and then lets see what the error is

CarlosHorro:
I do not understand why it shows this error message

`System.ArgumentOutOfRangeException: Length cannot be less than zero.
`
as if it is not receiving params at all. Testing on my local something similar:

`bash -c './rawtools.sh "$@"' parse --help
`

params are properly received...

CarlosHorro:
Hi @kevinkovalchik , do you know which could be the cause related to: System.ArgumentOutOfRangeException: Length cannot be less than zero.
?

kevinkovalchik:
Looking through the exception message from the failed test, I'm not sure exactly. It is something with the command line parse library, but the fact that it usually runs fine and doesn't crash when parsing the arguments makes me wonder if it is a mono issue. Can you specify mono-complete for the mono requirement? This has solved some other issues people have had in the past.

CarlosHorro:
Um... I think that solution would not be feasible as there is no mono-complete package in Conda or different packages for every one of its dependencies:
    - mono-runtime
    - mono-runtime-sgen
    - mono-llvm-support
    - libmono-2.0-1
    - libmono-profiler
    - mono-utils
    - mono-jay
    - mono-devel
    - mono-mcs
    - mono-roslyn
    - mono-csharp-shell
    - mono-4.0-gac
    - mono-4.0-service
    - monodoc-base
    - monodoc-manual
    - libmono-cil-dev
    - referenceassemblies-pcl
    - msbuild
    - ca-certificates-mono

kevinkovalchik:
Ah, okay. Well, I didn't look into it thoroughly so it was mostly a guess. I will look into the error some more.

kevinkovalchik:
I am new to working with Conda. Just to make sure I know what's going on, is this what is happening here?

1. Conda downloads the release from the specified location
2. Conda installs the release, along with the specified dependencies
3. Conda runs a test to make sure things are working. In this case, the test is:
```
bash -c 'RawTools.sh "$@"' parse --help
```

And step 3 is where things are failing. These tests are hosted elsewhere? Or is this happening on your system?

I tracked down the specific location where the exception is being thrown. It is inside a condition like this: `if (!string.IsNullOrEmpty(optionHelpText)) { do stuff }`, so I'm pretty sure the problem is not that no parameters are being passed. Even if no parameters are passed, a default help screen should still pop up.

That command works fine on my system (and yours too, it sounds like), so it still seems like something with the environment to me.

I can also try upgrading our command line parser library to the latest version. I suspect that won't help, but maybe...

CarlosHorro:
Hi Kevin,

Thank you for your testing. Your steps are correct, If I am not wrong, tests are run over freshly created docker containers with the specified dependencies. But we all have also the possibility to run these tests on our local machine, downloading this entire repo and doing:

`./bootstrap.py /tmp/miniconda
`
(This generates a new conda environment with all testing tools.)
(To activate the environment:)

`source ~/.config/bioconda/activate
`
(And lastly for testing, ie: rawtools:)

`bioconda-utils build recipes/rawtools/ config.yml --docker --mulled-test --git-range master
`

I think you are right about the cause of the problem (the environment), but I am not sure about how to solve this problem.
Reviewing
https://anaconda.org/conda-forge/mono/files
It seems that the most advanced Mono version we have available as a Conda package is the 5.14, but there are also v4.8.1.0 and v4.6.2.6, so I am going to do more tests with older Mono versions to check if this problem still happens...

CarlosHorro:
No surprises, the same error is shown on all mono versions available in Conda...

CarlosHorro:
Well, even replacing in the .sh file:
`mono $DIR/RawTools.exe "$@"`
by
`mono $DIR/RawTools.exe parse --help`

the exception is still shown, so this problem is not produced by a bad management of the params. 
Now the question is: how do we solve this type of environment problem? what library or library version could CommandLineParser be missing on standar Mono installations (without extra libraries from mono-complete?
Most of the suggested fixes for that type of "System.ArgumentOutOfRangeException: Length cannot be less than zero" problem make reference to the usual error of passing as the second param to the substring function the position of the last character instead of the total length, but I guess this is not the problem here.
Maybe we could ask them directly at
https://github.com/commandlineparser/commandline
?

kevinkovalchik:
It probably is that substring is being passed a negative length, but why that happens is a mystery... The length is calculated a couple method calls back, but I didn't try to follow things all the way back to the arguments. I wouldn't know which step is the issue anyway.

If I get a chance this weekend I'll try testing with the latest command line parser version, and if that doesn't resolve it I'll contact them and ask if them if they have any ideas. For testing, it looks like I can get the recipe from your GitHub?

CarlosHorro:
Yes, you can get the recipe from my GitHub:
https://github.com/CarlosHorro/bioconda-recipes

CarlosHorro:
Hi @kevinkovalchik , any news about this? :-)

kevinkovalchik:
No news yet. I updated the CL parser in a new branch, but haven't tested
it. Do you know if the tests can be run on a local build or will it only
work on a release hosted elsewhere, e.g. on GitHub?

CarlosHorro:
Hi,

Yes, as I explained earlier, you can run the tests on your local machine. The basic steps are:
1) Download the repo with the changes to be tested from Github (in this case, my bioconda-recipes fork)
2) Generate a new conda environment with all testing tools
`./bootstrap.py /tmp/miniconda `
3) Activate the environment
`source ~/.config/bioconda/activate `
4) And lastly, for testing, ie: rawtools, execute from the biodonda-recipes folder:

`bioconda-utils build recipes/rawtools/ config.yml --docker --mulled-test --git-range master `

Greetings,
Carlos

kevinkovalchik:
Sorry, my question was poorly phrased. I was wondering if a recipe run on a
local system could also point toward a local source for the compiled code
(e.g. an archive on my hard drive rather than a release binary on the
GitHub page). But I guess that isn't really necessary, so I'll just make a
temporary release with the change and test it. I will try this today and
let you know how it goes.

Kevin

kevinkovalchik:
It still fails after updating commandlineparser. But I found something else!

I looked into the code more where the error is being thrown in commandlineparser, and it is part of a function which helps generate the command line help text. The "length" argument that is causing the issue (it is negative, but substring can't take a negative value) is related to the width of the console window. They have some checks that are supposed to avoid errors here (like if the window is too narrow or the width is defined as negative) but they are broken because I can recreate the error, even in Windows with .NET, by making the console window as narrow as possible and calling up a help text window.

So probably in the testing environment the console window width is undefined or is 0 or something and command line parse is crashing. I'll trying manually changing the console settings when RawTools starts up and see if that fixes it.

kevinkovalchik:
Manually setting the width did not work. The maximum console width is determined by the current font size and screen resolution, so whatever this value is still leads to problems. I found a similar issue open on the commandlineparse github repository. It has been open for a long time, with a number of people having the same issue, so not sure they are going to fix this in a hurry.

The test just needs to return an exit status of 0, right? I can add a testing method to RawTools that bypasses the command line parser and executes as soon as RawTools loads. Something like:
```
static void Main(string[] args)
{
    if (args[0] == "voidTest") return;
}
```
and the program should exit with a status of 0. I tried this, and it still isn't working, but I can't see the actual exception that caused the test to fail, so maybe you should try.

There is a pre-release in the RawTools repository that has this added to it. When the program receives a first argument of "voidTest" it simply exits.

Another possible option would be to change the test to something with does not trigger the helptext to appear. Something like "RawTools parse -f void -p" would probably do the trick, as we check for invalid files. In such cases RawTools prints it's own error message and exits. This would be backward compatible with most versions of RawTools.

CarlosHorro:
Great job kevin, I'm going to do some testing on my local with both alternatives you propose. 
What worries me the most is the possibility of having the same problem when executing the galaxy tool. We will then need to set params in the command line, so I hope this width problem will not be reproduced in that real scenario...

kevinkovalchik:
Well, I suspect this is only an issue with the auto-generated help text,
which wouldn't be something accessed through Galaxy, right? Though if
Galaxy passed something that isn't part of the CLI, then the help-text
would want to pop up. If it is an issue, we might be able to disable it
when running through Galaxy.

Or I could fix the bug myself, either in the official repository if they
let me or in a fork.

CarlosHorro:
It makes sense :-)

CarlosHorro:
It seems it actually worked without that help text :-)

CarlosHorro:
Could this PR be merged, please? :-)
We have started to work on the Galaxy Tool and we would need to begin to test the functionality...

PR #13363, opened 2019-01-26
add Intlim version 1.1.0

abotzki:


recipes/r-intlim/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+  number: 0
+  skip: true  # [win32]
+
+  rpaths:
+    - lib/R/lib/
+    - lib/
+
+requirements:
+  build:
+    - {{posix}}zip               # [win]
```
bgruening: if this package needs to be compiled, can you please add compilers as well into the build section.

johanneskoester: I think the compiler `{{ compiler('c') }}` should be explicitly added, right @bgruening?

abotzki: Sorry to ask these maybe simple question: I've checked other comparable recipes and it seems that we could live without having any build step for Windows. The authors of the software indicate that there is no compilation necessary. I can confirm this on my Windows machine when installing it via devtools.
So, my idea was to completely remove the build step.

dpryan79:
@bioconda-bot please update

dpryan79:

dpryan79:
@bioconda-bot please merge

PR #13396, opened 2019-01-29
new package: r-misha

matthdsm:

@bioconda/r 
Hi guys,
One of our researchers insists on using this package, but I've got no idea what I'm doing with this recipe. Any and all comments are welcome, as I'm sure this really isn't up to par.

Thanks for the help
Cheers
M

### DIFF
--- a/recipes/r-shaman/build.sh
+++ b/recipes/r-shaman/build.sh
@@ -1,0 +1,1 @@
+$R CMD INSTALL --build .
--- a/recipes/r-shaman/meta.yaml
+++ b/recipes/r-shaman/meta.yaml
@@ -1,0 +1,30 @@
+{% set version = "180109" %}
+{% set git_hash = "52d5edefc170" %}
+package:
+  name: 'r-shaman'
+  version: '{{ version }}'
+source:
+  url: 'https://bitbucket.org/tanaylab/shaman/get/{{ git_hash }}.zip'
+  sha256: ad868c40c54bbcac5b38776bc56afca95bc514f091ea042b4d08303c83c4fe27
+build:
+  number: 0
+  rpaths:
+    - lib/R/lib/
+    - lib/
+requirements:
+  host:
+    - r-base
+    - bioconductor-genomeinfodb
+    - bioconductor-gviz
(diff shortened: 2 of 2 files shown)


dlaehnemann:
Hi @matthdsm,

don't worry, this will probably take some iterations, but is usually a pretty straightforward process with proper testing ensured by bioconda. And it looks like you found a good template to base your recipe on, although the following recipe (installing an R package from Bitbucket that is only available through `devtools` installation) is probably even closer, so maybe check it out:
https://github.com/bioconda/bioconda-recipes/tree/master/recipes/r-prestor

Looking at the circleci test of the current version (clicking on the red `X` next to the latest commit `6afb7c0` will get you there), it currently fails due to missing R package dependencies:
```
ERROR: dependencies ‘misha’, ‘RANN’, ‘doMC’ are not available for package ‘shaman’
```

Having the `r-devtools` dependency in the `build` environment might already do the trick. If there are still missing dependencies, you can check if they are available as [conda-forge packages](https://conda-forge.org/feedstocks/) or as [bioconda packages](https://bioconda.github.io/recipes.html) and then include them in `build` section if they are only needed at build time or the `host` section, if they are needed at build and at run time.

One thing to maybe try, is to ask the authors of the software to create a [semantic versioning](https://semver.org/) tag number of a specific commit that they deem stable, e.g. by [filing an issue in the misha repository](https://bitbucket.org/tanaylab/shaman/issues/new) -- the authors seem to be responsive. This way you avoid having the commit hash as a version number and confusing with future versions of the package, and it probably more stably generates the required package which will then be available from:
https://bitbucket.org/tanaylab/shaman/downloads/?tab=tags

And if there are further roadblocks, don't hesitate to ask and ping again!

cheers,
david

matthdsm:
Wow, thank for the comprehensive explanation! Really helped me a lot.

I hope the fixed I added will be enough.

Thanks again
Cheers
M

dlaehnemann:
Congrats, this was a tricky one with the dependencies and not on CRAN or bioconductor!

And I also learned things, so thanks as well.

PR #13399, opened 2019-01-29
Added irap-components

pinin4fjords:

This PR adds a recipe for providing some of the scripts etc from the iRAP package, without installation. The purpose of this is to allow use of some of those components outside of iRAP's Make-based workflows without a full install - which involves lengthy installation of dependencies. 

The recipe includes activation/ deactivation to set the necessary variables for iRAP to function.

This is based on a branch of a fork of iRAP, from which software packages have been removed, which would otherwise have made iRAP painful to download in this recipe. Future versions of this recipe may provide those dependencies via Conda for a fully-functioning Bioconda iRAP install.

### DIFF
--- a/recipes/irap-components/activate.sh
+++ b/recipes/irap-components/activate.sh
@@ -1,5 +1,3 @@
-#!/bin/sh
-  
 # Unset the passed env var <myvar>, backing up its value in BAK_<myvar>.
 backup_env_var() {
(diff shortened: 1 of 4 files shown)


pinin4fjords:
Abandoning for now in favour of an Anaconda org, just while I'm working on things.

PR #13407, opened 2019-01-30
updated dxpy to 0.273.0

notestaff:

### DIFF
--- a/recipes/dxpy/build.sh
+++ b/recipes/dxpy/build.sh
@@ -1,3 +1,0 @@
-#!/bin/bash
-sed -i.bak 's/==/>=/' requirements.txt
-$PYTHON setup.py install --single-version-externally-managed --root=/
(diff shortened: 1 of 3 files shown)


chapmanb:
Ilya;
Thanks much for modernizing this recipe. All of the updates look great but I'm not sure why the linux test is failing. It looks to be hanging indefinitely; I've tried re-starting it but it's still failing. Are you successfully building the updates locally? Thanks again.

notestaff:
@chapmanb locally, the circleci build succeeds.  i'm now running the bootstrap-based mulled build locally.  what is it doing as it's printing the dots?  is there a way to get it to print more details?

chapmanb:
I'm not sure what is making it take so long here if it builds fine locally. Pinging @bioconda/core to see if anyone else has an idea of how best to debug this.

dpryan79:
The trick is to run `top` when the mulled test is running locally. Sometimes the lack of a normal environment (no proper terminal, etc.) mucks up programs and they wait forever for input.

notestaff:
changing install from pip to setuptools seems to help, at least on linux...

PR #13408, opened 2019-01-30
recipe: r-misha

matthdsm:

### DIFF
--- a/recipes/r-misha/meta.yaml
+++ b/recipes/r-misha/meta.yaml
@@ -9,6 +9,7 @@ source:
 
 build:
   number: 0
+  skip: true # [osx)
 
 requirements:
   host:


recipes/r-misha/meta.yaml
```diff
@@ -0,0 +1,28 @@
+{% set version = "4.0.4" %}
+package:
+  name: 'r-misha'
+  version: '{{ version }}'
+
+source:
+  url: 'https://bitbucket.org/tanaylab/misha-package/get/96b4e57cfe08.zip'
```
dlaehnemann: From what I gather, the following link should also work: <https://bitbucket.org/tanaylab/misha-package/get/4.0.4.zip>

So to avoid missing to update the link in future package updates, I would probably turn the url entry into:
```
  url: 'https://bitbucket.org/tanaylab/misha-package/get/{{ version }}.zip'
```

recipes/r-misha/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+  sha256: f15608e7597085d62cbfc731b7936589e5d1f9938149104bd00cdc7d5f619c06
+
+build:
+  number: 0
+  skip: true # [osx)
```
dlaehnemann: I think the typo in the closing bracket seems top have triggered a "silent" bug in the bioconda testing framework. I just had a quick glance at the circleci output and noticed that the last lines read:
```
10:01:21 BIOCONDA INFO Nothing to be done for recipe recipes/r-misha
10:01:21 BIOCONDA INFO BUILD SUMMARY: successfully built 0 of 1 recipes
```

So it seems like the test skipped all builds. As it's already merged, I would suggest fixing by correcting to the `]` bracket and updating the `build` `number` to `1`.

dlaehnemann:
Nice find, that there's a tagged version of `misha`!

However, I think you inadvertently short-circuited the linux-test, and the dependent package will probably not build... ;)

recipes/r-misha/meta.yaml
matthdsm: thanks for catching this! Should be fixed now.

M

dlaehnemann:
Ah, and looking at the options on this PR, I think you'll have to file a new PR for the fix. But that should be possible from the same branch?!

recipes/r-misha/meta.yaml
matthdsm: Hi, 

I used the commit tag to get the latest versions. the repo has no tags, but I got the version number from the description in the source code. Hence, the proposed url wouldn't work, sadly.

Cheers
M

matthdsm:
I opened a new PR: #13410
Cheers
M

recipes/r-misha/meta.yaml
dlaehnemann: If you check, there is a tag for this and the link I posted works. A bit hard to find in the download section:
https://bitbucket.org/tanaylab/misha-package/downloads/?tab=tags

But the commit tag works, as well. So this is just to prevent misses in the future when updating.

matthdsm: I just merged the previous fix. If there's ever a need to update to a new version, I can update the URL.

m

dlaehnemann: Yeah, no worries. If it builds, it should be fine.

Good luck with the downstream packages, and should they run into awkward troubles, keep in mind that it might have to do with the `rpaths:` entry in the `meta.yaml`:
https://github.com/bioconda/bioconda-recipes/pull/13410#issuecomment-458904991

matthdsm: Thanks,

I know of the Rlibs, but in the template I used as shown in the downstream PR the libs weren't added, so I left them out too.

PR #13410, opened 2019-01-30
r-misha fix

matthdsm:

### DIFF
--- a/recipes/r-misha/meta.yaml
+++ b/recipes/r-misha/meta.yaml
@@ -8,8 +8,8 @@ source:
   sha256: f15608e7597085d62cbfc731b7936589e5d1f9938149104bd00cdc7d5f619c06
 
 build:
-  number: 0
-  skip: true # [osx)
+  number: 1
+  skip: true # [osx]
 
 requirements:
   host:


dlaehnemann:
I haven't built any `R` packages, yet, just used a lot of them and packaged a bunch of other stuff. But from looking at the [R recommendations and examples in the Contribution guidelines](https://bioconda.github.io/guidelines.html#r-cran), I realised that R packages seem to all have `rpaths` defined in their `build` section, i.e. something like:

```
build:
  number: 1
  rpaths:
    - lib/R/lib/
    - lib/
```

@bioconda/r, any comments if this is always needed and on what it does? Should this info be added to the contribution guidelines?

jdblischak:
> any comments if this is always needed and on what it does?

@dlaehnemann The [rpath](https://en.wikipedia.org/wiki/Rpath) is a Linux setting that instructs a program where to search for shared libraries.

From the conda recipe documentation on the [rpaths section](https://docs.conda.io/projects/conda-build/en/latest/source/define-metadata.html#rpaths): "Set which RPATHs are used when making executables relocatable on Linux. This is a Linux feature that is ignored on other systems. The default is `lib/`."

From the [conda cran skeleton](https://github.com/conda/conda-build/blob/bdbd7de3d61befb62e99a1c6e8731e9cab04882e/conda_build/skeletons/cran.py#L74): "This is required to make R link correctly on Linux."

Is this always required? It is definitely required if the R package uses any of the shared libraries available in `lib/R/lib` or `lib/`. It's not clear to me what would happen if it was removed for an R package that doesn't rely on shared libraries. It'd be an interesting experiment if you'd like to give it a try.

> Should this info be added to the contribution guidelines?

It is automatically added by the skeleton and doesn't need to be customized, so I assume that is why it is not discussed in the documentation.

jdblischak:
> the recommendation is for them to also start from the conda CRAN skeleton.

@dlaehnemann My general advice would be the following:

1. CRAN - use `conda skeleton cran`
1. Bioconductor - use `bioconda-utils bioconductor-skeleton`
1. GitHub - use `conda skeleton cran <URL of GitHub repository>` (requires that the package has at least one tagged version)
1. Bitbucket - copy-paste an existing recipe of a package on Bitbucket (ideally) or GitHub

I should also note that the first step for a package on GitHub or Bitbucket is to ask the authors if they intend to submit the package to CRAN/Bioconductor. Packages that are submitted to official repositories pass basic QC tests and the download URLs are stable (a Git repo can disappear at any time).

> What do you think? Could open a PR over at the bioconda.github.io repo...

I think more documentation is always welcome! If you send a PR, please tag the bioconda R team.

PR #13420, opened 2019-01-30
Add imctools

drpatelh:

Addition of `imctools` to `bioconda`. Local tests with `bioconda-utils` passing on Linux. 

This is my first Python-based submission so it would be great if you can just double-check everything is ok. 

I was also wondering what the difference would be in using:
```
 script: pip install guzzle_sphinx_theme && $PYTHON setup.py install --single-version-externally-managed --record=record.txt
```

as opposed to :
```
script: pip install guzzle_sphinx_theme && python -m pip install --no-deps --ignore-installed .
```

Thanks in advance!

### DIFF
--- a/recipes/imctools/meta.yaml
+++ b/recipes/imctools/meta.yaml
@@ -1,0 +1,42 @@
+{% set name = "imctools" %}
+{% set version = "0.2" %}
+{% set sha256  = "f105c2b725500937ff712845db2431fad6c2971e97c6f287869e507d74110233" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+build:
+  number: 0
+  script: pip install guzzle_sphinx_theme && $PYTHON setup.py install --single-version-externally-managed --record=record.txt
+
+source:
+  url: https://github.com/BodenmillerGroup/imctools/archive/v{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+requirements:
+  host:
+    - python
+    - pip
+  run:
+    - python
+    - numpy
+    - tifffile >=0.13.5
+    - scikit-image
+    - scipy
+    - pandas
+    - flake8
+    - pytest
+    - sphinx
+
+test:
+  imports:
+    - imctools
+
+about:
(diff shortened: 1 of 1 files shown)


drpatelh:
Sorry. Forgoto to ping @bioconda/core

epruesse:
Please don't install `guzzle_sphinx_theme` into the `imctools` package. It should be a separate package (it is on pypi after all). If it is not available as conda package yet, please use the `conda skeleton pypi` tool to create a package and submit it at conda-forge. You may ping `@bgruening` there to expedite the merge of your PR so that this one here can proceed.

Yes - this is a little more circuitous than you likely hoped. But packages should not contain things other than themselves - otherwise we run into clashes with file names already taken.

dpryan79:
None of the sphinx stuff is needed (likewise, pip and flake8 aren't needed) and I've removed them. That's all for generating documentation in the package, which isn't part of the installation usually.

drpatelh:
Apologies @epruesse. This was more out of naivety than anything else. I saw this being done for another package so I thought it would be acceptable.

Thanks @dpryan79 . I see now that the requirements come from `setup.py` and not `requirements.txt`. I did try using `noarch: python` as recently recommended but I was getting some build errors suggesting that a couple of `.so` libraries couldnt be found. I then found a found a post where you suggested to remove that line. However, I strongly suspect that was just down to the state of the recipe at the time.

Thank you very much.

recipes/imctools/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - scipy
+    - pandas
+
+test:
+  imports:
+    - imctools
+
+about:
+  home: https://github.com/BodenmillerGroup/imctools
+  license: UNKNOWN
```
bgruening: This seems to be a MIT license.

epruesse:
Ah - must be some caching issue - the `changed files` link showed me the old `pip install ...` line. Thanks @dpryan79

drpatelh:
Am I ok to merge this?

bgruening:
Yes you can! :)

PR #13430, opened 2019-01-31
 Add python-bioformats from pypi

drpatelh:

Added `python-bioformats` package required by CellProfiler. Recipe file was built using:
```
conda skeleton pypi python-bioformats
```

### DIFF
--- a/recipes/python-bioformats/meta.yaml
+++ b/recipes/python-bioformats/meta.yaml
@@ -21,14 +21,17 @@ requirements:
   host:
     - python
     - setuptools
+    - setuptools_scm
     - boto3
     - future
     - javabridge >=1.0
   run:
     - python
+    - setuptools_scm
     - boto3
     - future
     - javabridge >=1.0
+    - openjdk
 
 test:
   imports:
--- a/recipes/python-bioformats/meta.yaml
+++ b/recipes/python-bioformats/meta.yaml
@@ -1,0 +1,46 @@
+{% set name = "python-bioformats" %}
+{% set version = "1.5.2" %}
+{% set file_ext = "tar.gz" %}
+{% set hash_type = "sha256" %}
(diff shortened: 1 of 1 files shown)


drpatelh:
Added in `setuptools_scm` because build was failing, and `java-jdk` because tests were failing.

drpatelh:
@bioconda/core 

I generated the recipe using `conda skeleton` and I fixed this error:
```
RuntimeError: Setuptools downloading is disabled in conda build. Be sure to add all dependencies in the meta.yaml  url=https://pypi.org/simple/setuptools_scm/
```
by um adding `setuptools_scm` to the recipe.

I then started getting `java` errors where the tests were failing because java was required. I added `java-jdk` but realised thats now been deprecated so replaced it with `openjdk`. The build is now failing with:

```
ImportError: libjvm.so: cannot open shared object file: No such file or directory
``` 

Any ideas?

dpryan79:
It looks like conda's javabridge recipe is old enough that it wasn't built against a particular version of java, but instead uses various tricks to try and find any installed version. Those tricks seem to not always work, such as in this case. A better solution would be to put javabridge in conda-forge and have it build against a given range of openjdk versions (or at least 8.0).

drpatelh:
@dpryan79 It looks like `anaconda/ javabridge` contains the latest release i.e. `1.0.14`. This itself contains a patch called `0001-Dynamically-load-libjvm.so-before-importing-_javabri.patch`, and the recipe doesnt explicitly `build` or `run` `openjdk` as you suggested.
https://github.com/LeeKamentsky/python-javabridge

Installing CellProfiler will be the bain of my existence...

drpatelh:
Adding `javabridge` to `conda-forge` is currently failing on OSX without an obvious solution.
https://github.com/conda-forge/staged-recipes/pull/7701

I am closing this PR with the hope of revisiting in the near future.

PR #13441, opened 2019-02-01
Update sistr

apetkau:

@bioconda/core I am trying to pin the version of `blast` to a specific version (in order to address https://github.com/phac-nml/sistr_cmd/issues/34), but I'm not sure if this is the best way of doing this.

I also noticed that **sistr_cmd** had been added to `build-fail-blacklist`, but after fixing up the script again it seems to be working so I removed it.


bgruening:
The pinning looks good, I think you could also do `-blast 2.5.*`

dpryan79:
Blast is going to be problematic. One of its dependencies needs to get updated for the new pinnings but has a circular dependency.

apetkau:
Thanks @bgruening I've changed to `-blast 2.5.*`.

I took a look at newer way of of pinning dependencies (https://bioconda.github.io/faqs.html#how-are-dependencies-pinned-to-particular-versions), but wasn't sure how it all worked. Is it okay to use this older style of pinning dependencies for now? Especially since it seems that it won't be so straightforward with blast?

PR #13444, opened 2019-02-02
Update pinnings for bedtools (97 dependants) and bowtie2 (56 dependants)

dpryan79:


dpryan79:
The version of TBB that comes with the most recent bowtie2 doesn't play well with the new compilers (specifically, it has different settings for different compilers and throws its hands up if you aren't using a compiler named `gcc`/`icc`/`clang`). I'm going to try and switch it over to using tbb from conda-forge instead.

Issue #13155, opened 2019-01-21
plan for the GCC7 migration


bgruening:
* [x] label all main packages with an additional label (as conda-forge)
* [x] merge bulk into master
* [x] add `main` label to all packages with a `gcc7` label
* [x] communicate clearly to our community that now everything gets (re)build against gcc7
* [x] communicate that we need to bump the build-number to all packages (and its dependents) if you want to have a rebuild (https://bioconda.github.io/updating.html#updating-recipes-for-a-pinning-change)
* [ ] create a bot to assist users to bump builds, potentially in a single-issue
* [ ] make the bot smart enough to (re)build packages it pinning are bumped globally

bgruening:
Thanks to @dpryan79 we have now https://bioconda.github.io/updating.html#updating-recipes-for-a-pinning-change  ... the bot in progress by @epruesse afaik. That leaves us to the point of how we communicate this.

Creating a new issue and ping /all/?

bgruening:
I added an issue  (https://github.com/bioconda/bioconda-recipes/issues/13578) and update the PR template (https://github.com/bioconda/bioconda-recipes/pull/13579) to reach more people.

kyleabeauchamp:
Is the plan for us to wait for the old compilers to be removed and then
start bumping the build number on our favorite recipes?

epruesse:
You can already start bumping. That's actually what the new `bioconda-utils update-pinning` command does - you hand it packages of interest and it'll bump the dependency chain as necessary.

epruesse:
@bgruening - Does this mean I can restart `autobump` now?

bgruening:
@epruesse if `autobump` also considers `update-pinning` then yes.

Let's wait for @dpryan79 to confirm and feel free to merge https://github.com/bioconda/bioconda-recipes/pull/13579 so we update our users.

@kyleabeauchamp: unrelated but could you add `run_exports` to htslib. I think this would help a lot. Would be great if we could use `run_exports` more in our stack.

kyleabeauchamp:
Will take a look tonight

dpryan79:
Go for it, no need to wait for me!

kyleabeauchamp:
I ran the update-pinning script and I get a lot of errors:

```
bioconda-utils update-pinning --bump-only-python recipes/ config.yml --packages deeptools
[...]

Attempting to finalize metadata for pysam
Solving environment: ...working... done
Solving environment: ...working... failed
[...]

PR #13448, opened 2019-02-02
samtools w/ openssl-1.1.1 (but not pinned)

pb-cdunn:
Another attempt to deal with python2.7.15/openssl issues. E.g.
https://github.com/PacificBiosciences/pbbioconda/issues/85

### DIFF
--- a/recipes/samtools/meta.yaml
+++ b/recipes/samtools/meta.yaml
@@ -5,10 +5,10 @@ package:
   version: {{ version }}
 
 build:
-  number: 2
+  number: 3
 
 source:
-  url: https://github.com/samtools/samtools/releases/download/{{ version }}/samtools-{{ version }}.tar.bz2
+  url: https://github.com/samtools/samtools/releases/download/{{ version }}/samtools-1.9.tar.bz2
   sha256: 083f688d7070082411c72c27372104ed472ed7a620591d06f928e653ebc23482
 
 requirements:


kyleabeauchamp:
Could this be fixed by bumping the build number and leaving the version unchanged?

pb-cdunn:
@kyleabeauchamp , Actually, I just realized something. The current build needs to specify the current openssl, to catch the problem earlier. I will do what you suggest first, but with the old `openssl` specified, since that's what people were actually using.

Then, we can talk about how to *fix* the problem...

pb-cdunn:
`Malformed version string '>=1.0.2,<1.1.1'`

I'll try simply rebuilding, without specify the openssl version. That's not quite right, but it might be most likely to let current users keep what they have.

kyleabeauchamp:
Cool.  We'll be rebuilding with GCC7 soon (https://github.com/bioconda/bioconda-recipes/issues/13155), which might be another opportunity for further cleanup.

pb-cdunn:
That didn't work. `conda update samtools` updated to that build, but it still dynamically links with openssl-1.0.2. I will require openssl >= 1.1.1. Since we don't want a version bump of samtools, I think some people will break, but I don't see a better set of options. Things are already pretty broken.

bgruening:
@kyleabeauchamp we are ready. Everything is in place.

pb-cdunn:
Maybe related: https://github.com/conda-forge/conda-forge.github.io/issues/701

I hope the GCC7 update fixes everything. I'm out of ideas.

PR #13463, opened 2019-02-03
update pinning for bwa and paml

dpryan79:

### DIFF
--- a/recipes/bwa/0.6.2/meta.yaml
+++ b/recipes/bwa/0.6.2/meta.yaml
@@ -6,7 +6,7 @@ package:
   version: {{ version }}
 
 build:
-  number: 2
+  number: 3
 
 source:
   url: http://downloads.sourceforge.net/project/bio-bwa/bwa-{{ version }}.tar.bz2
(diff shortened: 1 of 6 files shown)


kyleabeauchamp:
@dpryan79  do we expect this pin upgrade to move BWA to the new compilers?  I think we might need to patch the makefiles to ensure that the docker image GCC isn't leaking in.  WDYT?  

```
strings -a bwa |grep gcc
```

Gives some stuff that includes 4.8.2.  Not sure if that's definitive or not.

kyleabeauchamp:
See, for example, https://github.com/bioconda/bioconda-recipes/pull/13594

dpryan79:
It does use gcc, it's just named differently on the command line. For C programs it doesn't actually matter, they're binary compatible, it's only an issue for C++ (from what I understand at least). We do often have to patch makefiles or explicitly inject the compilers, but those are caught during mulled testing at least.

kyleabeauchamp:
I'm more concerned about which version of GCC got used---how can we tell that the new versions got used?

bgruening:
@dpryan79 should we try to use the latest Docker image and conda 4.6.3 again?

dpryan79:
@bgruening It's worth a shot, that should lack a system gcc.

kyleabeauchamp:
I think my PR #13594  is definitely using the GCC 7, as it's now failing due to lack of ability to link to zlib.

PR #13464, opened 2019-02-03
samtools: Do not specify openssl version

pb-cdunn:
This does not solve our problem in all cases, and it may
hinder bioconda-wide changes to the openssl version.

See https://github.com/bioconda/bioconda-recipes/pull/13448

### DIFF
--- a/recipes/samtools/meta.yaml
+++ b/recipes/samtools/meta.yaml
@@ -5,7 +5,7 @@ package:
   version: {{ version }}
 
 build:
-  number: 4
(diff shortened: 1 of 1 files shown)


s22chan:
@pb-cdunn openssl is a build dependency, it needs to be put in the `build` section and it'll get auto exported into `build.run_exports` section

s22chan:
but more importantly, can samtools just disable ssl? as far as I can tell it's just used to securely download test scripts

PR #13468, opened 2019-02-04
Add recipe for SpydrPick

johnlees:

Need to build boost compiled libraries, but this may not be the most stable way.


johnlees:
Request for review @bioconda/core. 

There are three parts of this recipe I'd like to ask about:
* it clones parts of the boost repo before building it
* a single file from another github is downloaded with curl
* it clones a dependency (apegrunt) and then does a checkout of a specific commit
Are these things ok? A way to work around these requirements would be if I added a repo of my own which has these files prepared. Would that be preferable?

My main concern is the boost repo, which as it stands will always checkout the master branch, rather than a specific commit/archive.

Thanks!

dpryan79:
1. Can't it just use the boost package?
2. It's simpler to just include that one (assuming the license allows that)
3. That's not so bad, though it's generally preferable to make dependencies their own packages.

You shouldn't need `conda_build_config.yaml`, its contents are the defaults as of Friday.

johnlees:
1. Unfortunately not - while the header only boost libs are fine, the compiled files always lead to errors at the linker stage, I think due to them being built with different compiler options. I've run into this before, and I think the best solution is to make a partial boost build of these modules with consistent options as I've done here. I've modified the recipe so that a specific boost tag is checked out (as in 3).
2. I've now included this with the recipe as suggested
3. Noted - I will try and do this once this dependency has been properly released

I've removed `conda_build_config.yaml`

johnlees:
Is this ok to merge now I have fixed those issues?

dpryan79:
I mostly worry about someone trying to install this in an environment with boost and then getting cryptic error due to boost getting replaced on them without their knowledge. Is there an actual .so from boost that gets compiled and shipped and where is it ending up? A hacky solution would be to put it in a non-standard location and adding that to the rpath. Then at least multiple boosts could coexist.

johnlees:
The boost that's compiled here isn't installed (it's just in a work directory `boost_built` in `${SRC_DIR}`) and the needed .a libraries are linked statically. So all that is installed are the two binary targets -- no libraries. I ran `ldd` and confirmed there is no dynamic linking to any boost .so libs.

dpryan79:
Then I needn't worry :) I think you're fine to merge then.

PR #13473, opened 2019-02-04
Added fastq_utils recipe

pinin4fjords:


pinin4fjords:
@bioconda/core can I get a review? Guidelines say I should ask if it's a new recipe https://bioconda.github.io/contribute-a-recipe.html

dpryan79:
Can you get rid of the `conda_build_config.yaml`? That's trying to specify the old compilers, which we really don't want to use any more

dpryan79:
BTW, I expect much of the problems come with building with samtools 0.1.19 and then specifying samtools 1.3 as a runtime dependency.

pinin4fjords:
Thanks @dpryan79 . The fastq_utils software (written by someone else) requires bam.h etc from Samtools 0.1.19 at compile time, but depends on 1.3 for running- hence the discrepancy.

I tried the new compilers, but they didn't work with the Samtools 0.1.19 files- got errors to do with compiler options like "relocation R_X86_64_32S against `.rodata' can not be used when making a PIE object; recompile with -fPIC". Only downgrading the compiler fixed things (see last couple of commits). 

I guess would have needed to re-compile Samtools to address this without recourse to the old compiler. Is that what you suggest or is there a better solution I'm missing?

dpryan79:
Yeah, we need to get at least samtools 0.1.19 recompiled. I'm working on cutadapt, htseq, and bbmap now, so I'll try samtools 0.1.19 after that (I already recompiled samtools 1.9, it was in the first batch of updated packages). I assume the runtime dependency could be switched to a newer samtools, since otherwise we'd need to rebuild 1.3 as well (it's easy enough to add I assume, but I doubt it's really needed unless the tools uses the old `samtools sort` syntax).

pinin4fjords:
Okay great thanks @dpryan79 . So I'll wait for the recompile to happen before amending the recipe and merging? 

I can certainly switch to 1.9 initially, I doubt it will cause issues.

dpryan79:
If you can do a quick search through the fastq_utils code to see if it's using `samtools sort` that'd help. I think 1.3 is the last one that supported the old syntax (or maybe it was 1.6, I can't recall). Either way, if it's not using that then any version will work.

Yes, please wait until I can get 0.1.19 building. If you don't hear from me by this time tomorrow then please ping me :)

pinin4fjords:
I did check the code - I don't see any evidence of the old sort syntax. I will quickly identify if the higher version causes any other issues (though I know it should't). 

Thanks for the work- see you tomorrow ;-)

dpryan79:
I just merged in an updated build of samtools-0.1.19. It should be available within the next hour or so.

pinin4fjords:
Awesome, thanks @dpryan79 - seems to have done the trick. I took out conda_build_config.yaml and bumped runtime Samtools. Happy for me to merge once the tests finish running?

dpryan79:
:+1:Go for it!

PR #13490, opened 2019-02-04
pourRNA recipe

entzian:


recipes/pourRNA/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+./configure --prefix=$PREFIX
```
j-4: Can you please do a "${PREFIX}" to be sure, that nothing is messed up with spaces etc

recipes/pourRNA/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+test:
+  commands:
+    - pourRNA --version
+
+about:
+  home: https://github.com/ViennaRNA/pourRNA/
+  license: GPL
+  license_file: COPYING
+  summary: "Compute local minima and energy barriers of a landscape."
+
```
j-4: Is this tool published?
If so, you could add something like:
```
extra:
  identifiers:
- doi: xx/xxx.xx.xxx
```

bgruening:
@entzian @j-4 @eggzilla question for the viennarna experts. Would it make sense to include a `run_exports` into the viennarna package (https://docs.conda.io/projects/conda-build/en/latest/source/define-metadata.html?highlight=run_exports#export-runtime-requirements). This would ensure that the correct viennarna package is automatically added as run-time dependency.
If not we probably need to pin viennarna here?

j-4:
`run_exports` sounds like a valid thing to do here! viennarna acually is just a shared libarary with the swig python wrapper, so I guess this is an ideal use case for that!

eggzilla:
I am always pinning viennarna in my recipes, gives the package authors more freedom to make
breaking API changes.

recipes/pourRNA/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+    - viennarna
+  run:
+    - viennarna
+test:
+  commands:
+    - pourRNA --version
+
+about:
+  home: https://github.com/ViennaRNA/pourRNA/
+  license: GPL
```
eggzilla: Please specify the GPL version

bgruening:
Pinning is good if there are ABI changes, but keep in mind that this makes it impossible to mix environments (if tools have different VRNA pinnings). So the pinning should be ideally as loose as possible. Is there any guarantee from Ronny to only break ABI/API in x. or x.x. releases?

j-4:
Yes, Ronny only breaks ABI on major releases, so earliest with verions 3.0 of the package!

recipes/pourRNA/meta.yaml
entzian: It is submitted. I will add it in the next version.

bgruening:
@j-4 your call :)
I would pin viennarna here to <3 for the time being and include run_exports in viennarna at some point.

Thanks @entzian !

entzian:
This should be correct: >=2.4.11,<3.0.0. I don't know why lint fails.

martin-raden:
@entzian you have to rebase your branch to the current state of the master branch to solve the linting issue (it is not about your package but some other)

bgruening:
Merci!!!

PR #13520, opened 2019-02-06
rapid v0.8

skarunan:


skarunan:
My local tests also fail for the same reason below:

"Attempting to finalize metadata for rapid
Solving environment: ...working... Too long with no output (exceeded 10m0s)."

To be specific:
while trying 
- circleci build: timeout, as above
- bioconda-utils build (at the same step waited for more than 1 hour, and I aborted it)

My previous versions never ran in to such issues. I have only added one additional requirement (r-viridis) compared to my previous lists of requirements. I have already ensured the availability of this conda package. And, I've also ensured that there are no version dependency conflicts within my requirements list. 

@bioconda/build-system  Could someone please guide me in solving this issue?

recipes/rapid/v0.8/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+source:
+  url: https://github.com/SchulzLab/RAPID/archive/v{{ version }}.tar.gz
+  sha256: caddaa79d9be8012bc1ab09bdc431b863c80414eb471ac3a79ce5c940a03d4b5
+build:
+  number: 0
+
+requirements:
+  build:
+    - r-base >=3.4.1
```
bgruening: Please remove all the pinnings if not really needed.
Also with conda-build 3, this should be the `host` section and not the build section.

So you really need perl in the host section?

skarunan: Thanks for pointing out @bgruening . I modified the section to `host`, and removed Perl. 
I tried building locally, and unfortunately, I still face the same issue. 

And, I would prefer to have the other pinnings, in order to avoid issues from deprecated functionalities/incompatabilities across different versions.

skarunan: I removed all the pinning, and it successfully passed the tests. Thanks @bgruening .
 Now will test it out if the functionalities of my pipeline works fine.

PR #13523, opened 2019-02-06
Update to Gromacs 2019

stain:
http://manual.gromacs.org/documentation/2019/download.html


stain:
I think newer hwloc2 should work. Tell me again @bgruening, what is the difference between `host` and `run_exports`? Why would only hwloc go into `run_exports`?

If we compile with hwloc it will need to be there at runtime -- it can be compiled without it, but then I think GROMACS can't detect all host features (cpu type etc) at runtime.

I want to move this to have multiple AVX/SSE binaries built like in https://github.com/bioexcel/gromacs-docker/blob/master/Dockerfile#L79 which should isolate us from what the build host has vs what install host has.

bgruening:
`run_exports` are sitting in the libhwloc2 recipe and specifying that a specific version (dependent on what you use at compile-time) of libhwloc2 is __automatically___ included as a run-time dependency without you specifying it.

This means you, as gromacs maintainer, only include libhwloc2 as host-dependency, without pinning. Whatever was used during package-build-time will be specified as run-time dep automatically. Does this make sense?

stain:
I think that makes sense, @bgruening - so we need to suggest a change to https://github.com/conda-forge/libhwloc-feedstock/blob/master/recipe/meta.yaml to add something like this to restrict by major version?

```
build:
  run_exports:
    - {{ pin_subpackage('libhwloc', max_pin='x') }}
```

Am I right that all my `host` dependencies now in https://github.com/bioconda/bioconda-recipes/pull/13523/files are sufficient and I no longer need anything under `run`?

stain:
I'm also trying to resolve this [compilation error](https://circleci.com/gh/bioconda/bioconda-recipes/43396) conveniently at 94%:

```
In file included from /opt/conda/conda-bld/gromacs_1549463962759/work/src/gromacs/gpu_utils/gpu_utils_ocl.cpp:63:0:

/opt/conda/conda-bld/gromacs_1549463962759/work/src/gromacs/gpu_utils/oclraii.h:67:46: error: field initializer is not constant

     static constexpr ReleaserType releaser = clReleaseContext;
```

and

```
-- The C compiler identification is GNU 7.3.0
-- The CXX compiler identification is GNU 4.8.2
```

@acmnpv says I need to update cxx to >=5.1 for gcc as the versions need to match.

Perhaps change to `compiler('cxx')` is sufficient, or can we force the version while still using the `compiler` macro?

bgruening:
We will soon remove all compilers from the base image, so these confusions will go away. Good from my side!

ping @simonbray

acmnpv:
> I'm also trying to resolve this [compilation error](https://circleci.com/gh/bioconda/bioconda-recipes/43396) conveniently at 94%:
> 
> ```
> In file included from /opt/conda/conda-bld/gromacs_1549463962759/work/src/gromacs/gpu_utils/gpu_utils_ocl.cpp:63:0:
> 
> /opt/conda/conda-bld/gromacs_1549463962759/work/src/gromacs/gpu_utils/oclraii.h:67:46: error: field initializer is not constant
> 
>      static constexpr ReleaserType releaser = clReleaseContext;
> ```
> 
> and
> 
> ```
> -- The C compiler identification is GNU 7.3.0
> -- The CXX compiler identification is GNU 4.8.2
> ```
> 
> @acmnpv says I need to update cxx to >=5.1 for gcc as the versions need to match.
> 
> Perhaps change to `compiler('cxx')` is sufficient, or can we force the version while still using the `compiler` macro?

In general it would be best to have the compilers always match, but if this is not possible then the CXX compiler needs to have at least version 5.1 for gcc, 3.6 for clang and 17.0.1 for icc (as well as a standard library that is compatible with those versions)

stain:
..as [agreed in the chat](https://gitter.im/bioconda/Lobby?at=5c5d8162604f233ab6d55d41) I added both `2018` and `2019` versions in subfolders as they maintain both series concurrently (similar to Debian stable/testing)

stain:
```
+ gmx -version
gmx: error while loading shared libraries: libhwloc.so.5: cannot open shared object file: No such file or directory
Tests failed for gromacs-2018.5-h6bb024c_0.tar.bz2 - moving package to /opt/conda/conda-bld/broken
```

Looks like I got that `host` thing wrong again.. my fault - hoping 9e4d59abfb10eb7599d0b4620fb420e47c90274c will make it ;)

PR #13530, opened 2019-02-06
gottcha2 recipe

poeli:


recipes/gottcha2/2.1.4_beta/meta.yaml
```diff
@@ -0,0 +1,31 @@
+{% set name = "GOTTCHA2" %}
+{% set version = "2.1.4_beta" %}
```
bgruening: Do we really need the beta version? Can we wait for a stable version.

druvus: @poeli would you as the upstream author consider to package GOTTCHA2 using a setup.py file? Another improvement would be to include a download function to retrieve the databases automatically.

poeli: @bgruening The version is quite stable -- I'm comfortable to remove the "beta".
@druvus Thanks for the suggestions. I will make one.

stain: I would keep this PR open until @druvus have requested/added a non-beta version upstream with the database download function

druvus:
@bioconda-bot please update

PR #13541, opened 2019-02-07
moFF : Enforced python v3.4 or later as pymzml requires it.

CarlosHorro:


bgruening:
@Maux82 is moff still not working under py35? So is the only supported version 36 and 37?

Maux82:
@bgruening moFF should work in py35. If we force to py36 it will be better also moFF and pymzl

CarlosHorro:
Um... documentation here
https://github.com/compomics/moFF

says that moFF requires Python3.6+?

CarlosHorro:
I suppose new pymzml version was going to be 2.0.6.1 but it is 2.0.6-py_1, so I think just referencing to 2.0.6 we will use the last .1 build...

bgruening:
Cool thanks!

PR #13543, opened 2019-02-07
pymzml only for python3

bgruening:


CarlosHorro:
http://pymzml.github.io/intro.html#download

says it is compatible with python 2.6.5+ and python 3, but 

https://github.com/pymzml/pymzML

says it is just compatible with python 3.4+. 
I do not know if it would be (easily) possible to make it work actually on python 2.6.5+ or not...

PR #13545, opened 2019-02-07
Pinning bedtools version

skarunan:


recipes/rapid/v0.8/meta.yaml
```diff
@@ -23,7 +23,7 @@ requirements:
     - r-viridis
     - bioconductor-deseq2
     - bowtie2
-    - bedtools
+    - bedtools ==2.23.0
```
bgruening: can you add a comment why this is needed?

skarunan: Sure. The output fields of the 'bedtools coverage' command, in combination with the parameters I've employed, has changed. That's affecting some downstream analysis. So, I preferred to pin this version for the moment. I shall work on updating my downstream analysis accordingly in the future.

PR #13560, opened 2019-02-08
bioconductor-xcms-3.4.3 update from github

lecorguille:

@bioconda/r 

This update don't use bioconductor sources but GitHub. The version 3.4.3, I need, isn't release yet.
If you think that it's not the policy of bioconda, I can understand.
In the meanwhile, I have pushed temporary the package in a project conda channel: https://anaconda.org/workflow4metabolomics/bioconductor-xcms. So no pressure :)

### DIFF
--- a/recipes/bioconductor-xcms/meta.yaml
+++ b/recipes/bioconductor-xcms/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "3.4.2" %}
+{% set version = "3.4.3" %}
 {% set name = "xcms" %}
 {% set bioc = "3.8" %}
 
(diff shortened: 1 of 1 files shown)


dlaehnemann:
Hi @lecorguille,

thanks for the contribution!

For guidelines on non-CRAN and non-bioconductor sources for R packages, I would point you to the respective section in the guidelines:
https://bioconda.github.io/guidelines.html#r-other-sources

So it would be best, to first check with the authors, if they can release a new version via the standard channels. Otherwise, you might also generate conflicting version numbers, as the authors might release a version `3.4.3` in the future, that points to a different commit / version of the repository.

Another option for you might be to create a local custom anaconda channel and just use the respective recipe privately without publishing it on bioconda:
https://stackoverflow.com/questions/35359147/how-can-i-host-my-own-private-conda-repository

I hope this info helps?

cheers,
david

jdblischak:
> In the meanwhile, I have pushed temporary the package in a project conda channel: https://anaconda.org/workflow4metabolomics/bioconductor-xcms. So no pressure :)

@lecorguille I think this is a good use case for a personal channel. That way you can use the bleeding edge GitHub version until the official Bioconductor version is released and made available on bioconda.

Another thing to note is that [RELEASE_3_8](https://github.com/sneumann/xcms/tree/RELEASE_3_8) is only a branch. It'd be better (even just for the package on your personal channel) if it was tagged/released so that it wasn't a potentially moving target.

lecorguille:
Clear 👍 
Thanks for the review.
I close this PR!

PR #13566, opened 2019-02-08
update mykrobe to v0.6.1

mbhall88:

As this is my first recipe (technically an update, but I effectively had to write from scratch as the program has changed quite a lot) it would be great to have a review from the core team.


recipes/mykrobe/build.sh
```diff
@@ -1,3 +1,5 @@
 #!/bin/bash
-
-$PYTHON setup.py install --single-version-externally-managed --record=record.txt
+wget https://bit.ly/2H9HKTU -O - | tar -vxzf  -
```
bgruening: please use a post-installation hook to download this data after installation and do not include it in this package.

mbhall88: Sorry I don't know what you mean by that @bgruening

bgruening: Something like that for example: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/ensembl-vep/post-link.sh

Or a more complicated one: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/r-ngsplot-hg19/post-link.sh

mbhall88: May I ask why? It is a panel of variants that program *needs* in order to function correctly. Therefore should it not be included in the package?

bgruening: We want to keep the package size small. All of this needs to be stored and backuped. Data should be distributed separately. You download would add 30MB for every version of your package to the conda channel. Because you have no arch-depending python package, bioconda will create for you 4 packages (py27,35,36,37) - means 4*30MB - just for one singel version.

If you need a permanent storage for your data consider using https://github.com/galaxyproject/cargo-port or Zenodo.

mbhall88:
OK. So I have tried all night to get this post-link script to work but I can't get it to work and it doesn't even give me any indication as to why in the error message. See https://github.com/bioconda/bioconda-recipes/pull/13566/commits/e62adc1dff55da078714268102bb7c1067467fe0

https://github.com/bioconda/bioconda-recipes/pull/13566/commits/60bc3cbfaed58820d89070b820118265624a7679 was fairly straight-forward to construct and passed all the tests but as @bgruening stated, this is not going to be accepted.   

I understand the concern surrounding package size but given how difficult it's becoming to create a bioconda recipe (this was my first experience with creating a bioconda recipe) I think I might just abandon the whole thing.

recipes/mykrobe/build.sh
mbhall88: Another question about this post-link script. Will the data I am trying to add in the post-link script be accessible to mykrobe at runtime? 

Also I see in the [conda docs](https://conda.io/projects/conda-build/en/latest/source/link-scripts.html) that they strongly discourage the use of these scripts.

mbhall88:
@bgruening  The linting error says `use {{ compiler(""c"") }} or other new-style compilers`. What is meant by "other new-style compilers"?

mbhall88:
I don't get why this package won't compile with the latest GCC. It compiles fine on GCC 7.3 on my ubuntu 18.04 VM...

bgruening:
Let's use the new compiler and then try to understand the error if there is nothing inherent that forces us to use old GCC-stack - which you can do with {{ compiler }} and a local conda_build_config.yaml overwrite file.

mbhall88:
See the error log, but the crux of the problem is these errors when trying to use htslib object files.
```
/opt/conda/conda-bld/mykrobe_1550167954216/_build_env/bin/../lib/gcc/x86_64-conda_cos6-linux-gnu/7.3.0/../../../../x86_64-conda_cos6-linux-gnu/bin/ld: ../htslib/libhts.a(knetfile.o): relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIC

/opt/conda/conda-bld/mykrobe_1550167954216/_build_env/bin/../lib/gcc/x86_64-conda_cos6-linux-gnu/7.3.0/../../../../x86_64-conda_cos6-linux-gnu/bin/ld: final link failed: nonrepresentable section on output

```

recipes/mykrobe/build.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -x
 
-export CFLAGS="-I$PREFIX/include"
+export CFLAGS="-I$PREFIX/include -fPIC"
```
mbhall88: I've tried this locally and it still gave the same error. I'm going a little crazy staring at c compilation errors 😅

bgruening: The error indicates its using htslib.a - can it be that you try to compile statically and not dynamically?

mbhall88: Not quite sure. The htslib compiling is done by mccortex as it is one of it's dependencies. It's strange because there is a [bioconda recipe for mccortex](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/mccortex) already and I seem to have copied what they do.  
Admittedly, the mccortex required by mykrobe here is a fork that is ever so slightly modified, but it hasn't diverged that much from the original mccortex.  
FWIW, if I require `gcc <5.0` instead of the `{{ compiler('c') }}` require it builds and tests ok locally.

mbhall88:
Ok. Everything finally passes. Did you want to review it before merging @bgruening ?

PR #13580, opened 2019-02-10
Slamdunk

t-neumann:
* [ x ] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).


t-neumann:
Hi @bgruening 

the build is failing and I have a general bioconda question:

This is the error I get when building (I did **not** change any dependencies since the last version):

```
conda.exceptions.UnsatisfiableError: The following specifications were found to be in conflict:

  - curl=7.59

  - r-base=3.5.1 -> krb5[version='>=1.16.2,<1.17.0a0']

  - r-base=3.5.1 -> libcurl[version='>=7.62.0,<8.0a0']
```

The meta.yaml however does not set any fixed version for either R or curl:

```
...
    - r-base
    - r-tidyverse
    - ncurses
    - zlib
    - bzip2
    - curl
...
```

Isn't conda supposed to resolve any conflicts between dependencies automatically?

With regards to the DOI, there should be something citable soon, can this wait until then?

bgruening:
Conda has a gloabl pinning schema. E.g. here: https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml#L316

You can also find r-base there.

> With regards to the DOI, there should be something citable soon, can this wait until then?

Sure!

Let me try something.

t-neumann:
@bgruening Could you elaborate a bit more what that global pinning scheme exactly is and how this caused that issue? Sorry I'm a bit lost and there's not much on the help page.

And why you only had to fix the ```curl``` version.

bgruening:
Tobias, the global pinning file is taken by conda-build to determine if a given library should be pinned. Best example boost. You want to only build against a specific boost version and you want to be sure that the user gets the correct boost library during installation - that is the same version as you used during building.
Now you could do this by manually pinning `- boost 1.2.3`. That would work.
But what happens if tool A depends on boost-1.2.3 and tool B on boost-1.3.0 AND a user wants to have both tools in one environment? This will not work. Hence, we try to build the entire stack against a given fixed boost version - that does not need to be the latest - but its consistent across the stack.

Same foes for curl, as curl has a libcurl component. And there was an error imho which I tried to fix in https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/157 ... so not your fault. The reason was that `r-base` was compiled against a very recent `libcurl` version, and because `libcurl` is not pinned globally (`curl` is) you get a older curl version than the `libcurl` version was used in `r-base`. So in the end `curl` and `libcurl` do not match.

At least this is my interpretation of the error.

t-neumann:
@bgruening Thanks for the explanation - now I get it. Is there a lot of test jobs queued in ```ci/circleci: test-macos``` or is there another hickup why these tests have not run yet?

bgruening:
We are not sure what's going on but have writte the support team. It's not the queue.

t-neumann:
@bgruening Should I restart the macos tests or wait?

bgruening:
restarted it.

t-neumann:
Hi @bgruening 

I have a serious issue when running the containers produced from the bioconda slamdunk module:

When I simply create a slamdunk environment like this:

```
conda create --name slamdunk slamdunk
```
and then run commands:
```
source activate slamdunk
slamdunk all -r slamdunk/test/data/ref.fa -b slamdunk/test/data/actb.bed -o slamdunk/test/data/output -rl 100 -mbq 27 -5 0 slamdunk/test/data/reads.fq
```
all runs correctly.

However, when I use the biocontainers container created for the Bioconda release:
```
singularity exec docker://quay.io/biocontainers/slamdunk:0.3.4--py27r351h469c4b5_0 slamdunk all -r slamdunk/test/data/ref.fa -b slamdunk/test/data/actb.bed -o slamdunk/test/data/output -rl 100 -mbq 27 -5 0 slamdunk/test/data/reads.fq
```

It always crashes with some weird OpenCL compilation error. How is that possible? If the conda module wouldn't work, I would have heard by now because no matter what input data, I always end up with the OpenCL error.

```
[OPENCL] Available platforms: 1
[OPENCL] AMD Accelerated Parallel Processing
[OPENCL] Selecting OpenCl platform: AMD Accelerated Parallel Processing
[OPENCL] Platform: OpenCL 1.2 AMD-APP (1214.3)
[OPENCL] 1 CPU device found.
[OPENCL] Device 0: Intel(R) Xeon(R) CPU E5-2690 v4 @ 2.60GHz (Driver: 1214.3 (sse2,avx))
[OPENCL] 56 CPU cores available.
[OPENCL] Build failed: Program build failure
[OPENCL] Build status: build failed
[OPENCL] Build log:
[OPENCL] Internal Error:  as failed
[OPENCL] Codegen phase failed compilation.
[OPENCL] Unable to build program end.
Error: Program build failure (-11)
terminate called without an active exception
Aborted
```

It is also not a general container issue, because the Docker containers I build for my Slamdunk releases work fine.

I'm a bit lost here - and without a proper conda based container, any progress on my nf-core pipeline is stalled. Do you have any ideas?

t-neumann:
I played a bit around and it must have to do with the Docker base image: What are you using on Bioconda?
`ubuntu:18.04` works when I install Miniconda and then create a conda environment with slamdunk

bgruening:
I will look into this, but downloading takes some time. So you see the same for Docker? We are using a minimal busybox image.

bgruening:
@t-neumann can you try something like that:

```
docker run -i -t quay.io/biocontainers/slamdunk:0.3.4--py27r351h469c4b5_0 slamdunk all -r slamdunk/test/data/ref.fa -b slamdunk/test/data/actb.bed -o slamdunk/test/data/output -rl 100 -mbq 27 -5 0 slamdunk/test/data/reads.fq
```

You might need to mount fastq files into the container.

t-neumann:
@bgruening Unfortunately that gives me the same error. It looks like there's something missing from busybox to ubuntu. These are the packages I install when setting up my Docker image from the Ubuntu base image - can you spot anything?

```
FROM ubuntu:18.04

MAINTAINER Tobias Neumann <tobias.neumann.at@gmail.com>

ARG VERSION_ARG

ENV DEBIAN_FRONTEND noninteractive
ENV DEBCONF_NONINTERACTIVE_SEEN true

# binutils is required to run opencl programs
RUN buildDeps='git wget gcc g++ libc6-dev make cmake zlib1g-dev python-pip python-dev python-distribute python-pip bzip2 libncurses-dev libbz2-dev liblzma-dev' \
    runDeps='python default-jre binutils r-base unzip' \
    && echo "tzdata tzdata/Areas select Europe" > /tmp/preseed.txt; \
    echo "tzdata tzdata/Zones/Europe select Berlin" >> /tmp/preseed.txt; \
    debconf-set-selections /tmp/preseed.txt \
    && set -x \
    && apt-get update && apt-get install -y $buildDeps $runDeps --no-install-recommends \
    && rm -rf /var/lib/apt/lists/* \
    && pip install --upgrade pip==9.0.3 \
    && pip install git+https://github.com/t-neumann/slamdunk.git@${VERSION_ARG} \
    && apt-get purge -y --auto-remove $buildDeps
```

bgruening:
@t-neumann yes all of this is not included. The idea is that you define all dependencies as conda dependencies and with that, they are installed in the Container as well.

So maybe you need to add binutils to the container as well. And maybe you can come-up with a test that triggers this error here on bioconda as well.

t-neumann:
@bgruening Ok I will play around a bit. It's just a bit counter-intuitive that a native conda installation works and the resulting Docker container does not because the underlying OS is different. You see what I mean? But I guess there's no way around that.

bgruening:
The reason is that the underlying OS ships with dependencies that you have not defined. E.g. in the very beginning we build all packages without specifying `zlib` as conda deps. This worked well ... because inside the Docker container zlib was always installed. But it crashed on HPC envrionments where no zlib was installed.

In the end we need to define as many dependencies as possible in conda and not trust the underlying OS. Because you want your package to be deployable everywhere independent of the underlying OS.

PR #13583, opened 2019-02-10
Octopus: update to 0.6.2-beta

dancooke:

### DIFF
--- a/recipes/octopus/meta.yaml
+++ b/recipes/octopus/meta.yaml
@@ -21,7 +21,7 @@ requirements:
     - {{ compiler('cxx') }}
     - cmake
   host:
-    - boost-cpp
+    - libboost
     - htslib
     - wget
     - icu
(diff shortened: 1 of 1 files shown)


dancooke:
@bgruening This causes a build error that indicates Boost wasn't compiled with C++14. How was this boost library built?

bgruening:
`    boost-cpp:       1.68.0-h11c811c_1000       conda-forge` the one with the build-number 1000 was build with the new compilers as far as I can tell. `boost` is also just `boost-cpp` + `python` + `numpy`. https://github.com/conda-forge/boost-feedstock/blob/master/recipe/meta.yaml#L22

dancooke:
Doesn't look like this Boost package was built using C++14. This may not be an issue with the newer build of the boost-cpp package (`boost-cpp: 1.69.0-h6a4c333_1000`) using Boost 1.69 (https://github.com/boostorg/system/issues/28). Is it possible to use this build? Otherwise I'd suggest reverting to libboost as this built without issue.

dancooke:
@bgruening I couldn't get `boost-cpp` to work so reverted to `libboost`. It looks like Octopus built successfully on Linux but there was some issue with Bioconda. Any ideas?

Issue #13578, opened 2019-02-10
GCC7 migration is done


bgruening:
Thanks to @dpryan79, @epruesse and the conda-forge community we finished the GCC7/py37 migration. Thanks everyone.

### What does this mean?

* New Python packages, or updates, will get a new build against Python 3.7. Please note that it is still preferred to use [`noarch: python`](https://conda-forge.org/docs/meta.html#building-noarch-packages) for all pure Python packages (run-time dependencies do not count).

* If you rebuild or build a (C/C++) package you will now use the new compilers that are coming from the conda default channel. The compilers in the container are not used anymore and will be removed from the container in the next days.

* It can be that some dependencies in Bioconda are not yet compiled against GCC7, so that your program fails unexpectedly. To fix this please also re-build all the dependencies that you need. Bioconda utils gained an [`update-pinning`](https://bioconda.github.io/updating.html#updating-recipes-for-a-pinning-change) command which you could use to figure out what needs to be rebuilt.

* We already rebuilt all Bioconductor packages against GCC7 and the latest BioConductor release. Enjoy a clean and up-to-date R stack!

* Watch out for the [skip-list](https://github.com/bioconda/bioconda-recipes/blob/master/build-fail-blacklist). If packages are failing during a rebuild or produce major problems we put them on this list. Please feel free to help us "fixing" those recipes to bring them back.

Enjoy the new compilers, we hope that the Conda stack gets more stable with it.

SilasK:
I cannot install the bioconda-utils

`pip install https://github.com/bioconda/bioconda-utils`

gives me the following error:

```
Collecting https://github.com/bioconda/bioconda-utils
  Downloading https://github.com/bioconda/bioconda-utils
     \ 184kB 901kB/s
  Cannot unpack file /private/var/folders/bt/97hqkwdn0nv0lwjfmtmsk20c0000gn/T/pip-unpack-522j387s/bioconda-utils (downloaded from /private/var/folders/bt/97hqkwdn0nv0lwjfmtmsk20c0000gn/T/pip-req-build-_c9mgh3g, content-type: text/html; charset=utf-8); cannot detect archive format
Cannot determine archive format of /private/var/folders/bt/97hqkwdn0nv0lwjfmtmsk20c0000gn/T/pip-req-build-_c9mgh3g
```

dpryan79:
`pip install git+https://github.com/bioconda/bioconda-utils`

unode:
Hi @bgruening 

[...]

PR #13584, opened 2019-02-10
I haven't updated stuff from the WGBS pipeline yet, but this takes ca…

dpryan79:
…re of the rest of snakePipes

:information_source:

----------------


dpryan79:
I've updated the blacklist a bit to add some old recipes that aren't used within bioconda at least.

bgruening:
> I've updated the blacklist a bit to add some old recipes that aren't used within bioconda at least.

Feel free to remove them.

dpryan79:
I'm a bit hesitant to remove them, since I don't know if someone actually needs them around for some reason (using them in a pipeline and maintaining the versions). I would suggest we revisit the blacklist this summer and purge anything in it that's been blacklisted for at least 6 months.

PR #13585, opened 2019-02-10
Added tximport-scripts recipe

pinin4fjords:

This is another R package CLI along the lines of e.g. [r-seurat-scripts](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/r-seurat-scripts). This is basically a single rich wrapper script for the central tximport() function of the Bioconductor tximport package. It uses DropletUtils to provide optional sparse matrix output. Source package at https://github.com/ebi-gene-expression-group/tximport-scripts. I've dropped the 'r-' prefix convention to prevent interference with build systems recognising r packages (to be remedied in r-seurat-scripts in the future).

### DIFF
--- a/recipes/tximport-scripts/meta.yaml
+++ b/recipes/tximport-scripts/meta.yaml
@@ -1,0 +1,42 @@
+{% set version = "0.0.1" %}
+
+package:
+  name: tximport-scripts
+  version: {{ version }}
+
+source:
+  url: https://github.com/ebi-gene-expression-group/tximport-scripts/archive/v{{ version }}.tar.gz 
+  sha256: 7910a7519dd6655d8a2007019639d4204829aeb8d1a9f6a94d0c80055003ea10 
+
+build:
+  number: 0
+  skip: true  # [win32]
+  noarch: generic
+
+requirements:
+    host:
+        - r-base 3.5.1
+    run:
+        - r-base 3.5.1
+        - bioconductor-tximport 1.10.0
+        - r-optparse
+        - r-workflowscriptscommon
+
+test:
+    commands:
+        - tximport.R --help
+
+about:
+  home: https://github.com/ebi-gene-expression-group/tximport-scripts 
+  dev_url: https://github.com/ebi-gene-expression-group/tximport-scripts
+  license: GPL-3
+  summary: A set of wrappers for individual components of the tximport package.
+      Functions R packages are hard to call when building workflows outside of R,
+      so this package adds a set of simple wrappers with robust argument parsing.
+      Intermediate steps are currently mainly serialized R objects, but the
+      ultimate objective is to have language-agnostic intermediate formats allowing
+      composite workflows using a variety of software packages.
+  license_family: GPL
+extra:
+  recipe-maintainers:
+    - pinin4fjords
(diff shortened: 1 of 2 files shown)


pinin4fjords:
ping @bioconda/core for review due to new recipe

recipes/tximport-scripts/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  name: tximport-scripts
+  version: {{ version }}
+
+source:
+  url: https://github.com/ebi-gene-expression-group/tximport-scripts/archive/v{{ version }}.tar.gz 
+  sha256: 7910a7519dd6655d8a2007019639d4204829aeb8d1a9f6a94d0c80055003ea10 
+
+build:
+  number: 0
+  skip: true  # [win32]
```
dpryan79: You can remove this, it's a hold-over from conda-forge.

recipes/tximport-scripts/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  noarch: generic
+
+requirements:
+    host:
+        - r-base 3.5.1
```
dpryan79: You should be able to just use `r-base`, since (A) we only build for R 3.5.1 at the moment and anyway the bioconductor packages that are run time dependencies are themselves pinned to specific R versions.

recipes/tximport-scripts/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    run:
+        - r-base 3.5.1
+        - bioconductor-tximport 1.10.0
```
dpryan79: Can you make this `1.10.*`? That will allow for bug fix releases of tximport to be used (but not those from bioconductor 3.9).

recipes/tximport-scripts/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  summary: A set of wrappers for individual components of the tximport package.
+      Functions R packages are hard to call when building workflows outside of R,
+      so this package adds a set of simple wrappers with robust argument parsing.
+      Intermediate steps are currently mainly serialized R objects, but the
+      ultimate objective is to have language-agnostic intermediate formats allowing
+      composite workflows using a variety of software packages.
+  license_family: GPL
+extra:
+  recipe-maintainers:
+    - pinin4fjords
```
dpryan79: It doesn't hurt anything, but you can get rid of the `extra:` section.

pinin4fjords:
Thanks @dpryan79 ! Have made the requested changes, will merge when the tests pass

recipes/tximport-scripts/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+  url: https://github.com/ebi-gene-expression-group/tximport-scripts/archive/v{{ version }}.tar.gz 
+  sha256: 7910a7519dd6655d8a2007019639d4204829aeb8d1a9f6a94d0c80055003ea10 
+
+build:
+  number: 0
+  noarch: generic
+
+requirements:
+    run:
+        - r-base 3.5.1
```
bgruening: you can remove the pinning here, we do this automatically now.

pinin4fjords: Okay, thanks @bgruening , I've changed that. BTW, any idea why the CI is getting stuck in the queue for the macos part but not the linux?

dpryan79: We've contacted circleci about that, we're waiting to hear back from them.

PR #13586, opened 2019-02-11
[WIP] Add run_exports for htslib

kyleabeauchamp:
:information_source:

----------------


recipes/htslib/meta.yaml
```diff
@@ -6,6 +6,8 @@ package:
 
 build:
   number: 5
+  run_exports:
+    - {{ pin_subpackage(name, max_pin='x.x') }}
```
bgruening: @kyleabeauchamp the syntax looks like that and it needs to be part of /build/run_exports/.

Do you know which versions are API stable? x.x. or x.x.x or x?

kyleabeauchamp: I'm not sure how well they abide by versioning, but I would guess x.x, i.e., I would trust anything in the 1.9.X series for this build...

kyleabeauchamp:
PS: we should probably wait until https://github.com/bioconda/bioconda-utils/issues/379 is resolved before merging this.

bgruening:
@kyleabeauchamp can you keep track of this and merge if you think its ok. Thanks!

PR #13592, opened 2019-02-11
Add bioverbs recipe

pinin4fjords:

This PR adds the bioverbs component of basejump. See discussion at https://github.com/steinbaugh/basejump/issues/141.


mjsteinbaugh:
Thanks @pinin4fjords. Let's pin it to v0.1.6 instead, which includes an additional `plotCounts()` generic that is now in use in a few dependency packages, including basejump. I added a release for that update.

mjsteinbaugh:
Perfect thanks that looks good

pinin4fjords:
Thanks @mjsteinbaugh . There's a known CI issue right now which prevents merge, hopefully will be resolved soon.

ping @bioconda/core for review of new package

dpryan79:
Is there a reason this is pinned to an out-dated R version?

pinin4fjords:
Was at request of package author @mjsteinbaugh

mjsteinbaugh:
A number of the Bioconductor packages on bioconda aren't updated to R 3.5/BioC 3.8 yet, so we have to maintain support for R 3.4. I'm not sure how we need to pin this in the recipe.

Here's the current approach for bcbioRNASeq, which depends on basejump:
https://github.com/bioconda/bioconda-recipes/tree/master/recipes/r-bcbiornaseq

Maybe @roryk has some advice on how to update the recipe?

mjsteinbaugh:
Sorry about that -- yeah looking at the recipe, we shouldn't need to pin to R 3.4.

dpryan79:
I've updated all of the CRAN and bioconductor packages already, a bunch of non-CRAN/bioconductor packages were blacklisted in the process, but you can feel free to unblacklist them as needed. In general we should remove the `r-` and `bioconductor-` prefixes from the names, though, as they're not CRAN/bioconductor packages.

mjsteinbaugh:
@dpryan79 Is there a recommended prefix if the package is on GitHub? I'm in the process of getting these submitted to Bioconductor, but that takes a while to go through review.

epruesse:
How long does it take? Package names are "forever", so if this will be pushed to bioconductor, it would be best if it does get that name.

pinin4fjords:
Is okay to give it a Bioconductor prefix even if the source is temporarily coming from GitHub @epruesse ?

epruesse:
Ideally, we'd wait for acceptance by bioconductor. That's why I was asking - how long will that take?

epruesse:
You can use `r-bioverbs`.

mjsteinbaugh:
@epruesse I'll try to submit these to Bioconductor ASAP. Apologies for the naming issue.

PR #13618, opened 2019-02-13
omero: install bin/omero as well

joshmoore:
see: https://www.openmicroscopy.org/community/viewtopic.php?f=5&t=8678

### DIFF
--- a/recipes/python-omero/build.sh
+++ b/recipes/python-omero/build.sh
@@ -3,3 +3,6 @@
 cd ./lib/python
 cp $RECIPE_DIR/setup.py .
 $PYTHON setup.py install --single-version-externally-managed --record=record.txt
+
+cd ../..
+cp bin/omero $PREFIX/bin/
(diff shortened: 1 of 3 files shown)


recipes/python-omero/build.sh
```diff
@@ -3,3 +3,6 @@
 cd ./lib/python
 cp $RECIPE_DIR/setup.py .
 $PYTHON setup.py install --single-version-externally-managed --record=record.txt
+
```
bgruening: I don't understand why setup.py is copied in the first place. But can it be removed after installation?

The omero binary is not compiled in this recipe, for compatibility we recommend to compile everything from scratch.

joshmoore: Hi @bgruening 

> I don't understand why setup.py is copied in the first place.

I don't know what issues @ThomasWollmann ran into originally with the mainline setup.py and I've failed to try it out for myself. I'll put that on the list.


> But can it be removed after installation?

I've added a commit to delete on this PR.

> The omero binary is not compiled in this recipe...

`bin/omero` is just a launcher script: https://github.com/openmicroscopy/openmicroscopy/blob/develop/components/tools/OmeroPy/bin/omero

bgruening:
Gracias!

PR #13629, opened 2019-02-13
Update r-bcbiobase recipe to pin r-basejump 0.7.2

mjsteinbaugh:
:information_source:

----------------

----------------

The recipe needs to be updated to pin [`r-basejump`](https://github.com/steinbaugh/basejump) 0.7.2 specifically, because the current release (0.9.9) isn't yet fully backward compatible with `r-bcbiobase` 0.4.1 and/or `r-bcbiornaseq` 0.2.8. An updated version of `r-basejump` has been requested on bioconda, so this fix is needed temporarily until I can make `r-basejump` fully backward compatible.

### DIFF
--- a/recipes/r-bcbiobase/meta.yaml
+++ b/recipes/r-bcbiobase/meta.yaml
@@ -25,7 +25,7 @@ requirements:
     - bioconductor-iranges
     - bioconductor-s4vectors
     - r-base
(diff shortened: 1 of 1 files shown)


roryk:
Looks good to me-- I think you don't actually need the package rebuilt since its already built against 0.7.2 so not incrementing the build number makes sense.

PR #13633, opened 2019-02-13
Update pinnings for methyldackel, metilene, and ucsc-facount

dpryan79:
:information_source:

----------------

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -670,7 +670,6 @@ recipes/gnu-getopt
 recipes/ghmm
 recipes/domclust
 recipes/bwa/0.6.2
-recipes/metilene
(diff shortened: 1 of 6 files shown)


bgruening:
@daler we need to rebuild all UCSC packages against the new compiler-stack. Do you think a general upgrade would be worthwhile instead of rebuilding the old once?

daler:
I'm slow, sorry -- yes, I think an upgrade would be much better. I don't think they've been upgraded in a while.

dpryan79:
The last time was probably during the CB3 change. I'll merge an update into bulk.

dpryan79:
Now that the UCSC tools have been updated I'll start this PR fresh.

PR #13635, opened 2019-02-13
Add compare-reads

mvdbeek:
----------------

### DIFF
--- a/recipes/compare-reads/build.sh
+++ b/recipes/compare-reads/build.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+$PYTHON setup.py install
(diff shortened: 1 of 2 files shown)


mvdbeek:
Hmm, that's a little weird:
```
cp: cannot stat `/opt/conda/conda-bld/linux-64/compare-reads-0.0.1-h14c3975_0.tar.bz2': No such file or directory
```
which seems to be the only error I can spot. It looks like there's a mix of apostrophe and single quote in there, but that may be a red herring.

mvdbeek:
OK, the rerun was more informative.

mvdbeek:
That's rather surprising that you need to explicitly list python in the run requirements if I already list a python package (pysam), but OK, it works.

stain:
Thank you, @mvdbeek!

PR #13638, opened 2019-02-14
Add draft wg-blimp-v0.9.0

MarWoes:

This is a first test version, things will likely change until final PR.

### DIFF
--- a/recipes/wg-blimp/build.sh
+++ b/recipes/wg-blimp/build.sh
@@ -1,0 +1,2 @@
+#!/bin/bash
+$PYTHON setup.py install --single-version-externally-managed --record=/tmp/record.txt
--- a/recipes/wg-blimp/build.sh
+++ b/recipes/wg-blimp/build.sh
@@ -1,2 +1,0 @@
-#!/bin/bash
-$PYTHON setup.py install --single-version-externally-managed --record=/tmp/record.txt
--- a/recipes/wg-blimp/meta.yaml
+++ b/recipes/wg-blimp/meta.yaml
@@ -18,6 +18,7 @@ requirements:
     - click
     - ruamel.yaml
     - snakemake
+    - git
 
   run:
     - python=3.5
@@ -29,6 +30,7 @@ requirements:
     - bwa
     - snakemake
     - fastqc
+    - git
     - multiqc
     - methyldackel
     - qualimap
--- a/recipes/wg-blimp/meta.yaml
+++ b/recipes/wg-blimp/meta.yaml
@@ -11,13 +11,12 @@ source:
 build:
   number: 0
   noarch: python
+  script: "{{ PYTHON }} -m pip install . --no-deps -vv"
 
 requirements:
   host:
     - python=3.5
-    - click
-    - ruamel.yaml
-    - snakemake
+    - pip
     - git
 
   run:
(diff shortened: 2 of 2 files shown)


MarWoes:
Hi,

Since this is my first PR in bioconda, would someone have a look at it? @bioconda/core 

Thanks in advance!

recipes/wg-blimp/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+  url: https://github.com/MarWoes/wg-blimp/releases/download/v0.9.0/wg-blimp-v0.9.0.tar.gz
+  sha256: 7508b014dab5674f4b49745c487230fdecbb2b065ef08280234bee1e02604f58
+
+build:
+  number: 0
+  noarch: python
+
+requirements:
+  host:
+    - python=3.5
```
bgruening: is there any reason you pin it to 3.5?

recipes/wg-blimp/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - ruamel.yaml
+
+test:
+
+  commands:
+    - wg-blimp --help
+
+about:
+  home: https://github.com/MarWoes/wg-blimp
+  license: AGPL-3.0
```
bgruening: is there a license file?

recipes/wg-blimp/build.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
```
bgruening: you can get rid of this file by using https://conda-forge.org/docs/meta.html?highlight=pip%20install#use-pip

recipes/wg-blimp/meta.yaml
MarWoes: License file was added

recipes/wg-blimp/build.sh
MarWoes: Updated to remove build.sh

recipes/wg-blimp/meta.yaml
MarWoes: Changed so 3.5 is only used as minimum. 3.5 is required as minimum because of some python features are used that require at least 3.5. Also, not stating a version seems to cause conflicts (2.7 is too low for snakemake), is this sufficient for defining versions or is there a better way to state the dependency on python >= 3.5?

MarWoes:
Thank you very much for your review! @bgruening 
I changed the code based on your suggestions, is there anything else that should be changed?

bgruening:
@MarWoes why is git needed?

MarWoes:
@bgruening There were some errors regarding git missing, but I think it is indeed sufficient to have git in the run section.

bgruening:
Do you know why this is needed? Please make sure you are not missing any deps or this tool is not downloading anything that is not tracked. That would be bad for reproducibility. 

Thanks @MarWoes!

MarWoes:
@bgruening snakemake seems to require a working git version (build crashes otherwise), however it is only used for certain features (such as --archive) and not downloading anything untracked, so I think it would not affect reproducibility.

bgruening:
Thanks @MarWoes!

MarWoes:
Great, thank you very much for your review! @bgruening

PR #13639, opened 2019-02-14
Update bioconductor-loomexperiment

nh3:
This PR updates bioconductor-loomexperiment to 1.1.5. Changes from 1.0.1 to 1.15 include recent fixes to correctly export rowData and features to handle sparse matrix, therefore making its availability in bioconda highly desirable.


bgruening:
@nh3 this version is not yet released, right? We currently have no strict guideline for that, but I know people recommended to use your own channel for beta packages.

Otherwise, this looks good.

nh3:
You are right @bgruening, this is in devel. I'll leave it open as it is for now.

dpryan79:
We can probably close this. The next bioconductor release is schedule for the end of April and we'll start on a mass upgrade of all bioconductor packages then. That will allow us to keep all of the dependency pinnings appropriate.

nh3:
Hi @dpryan79, I can't find bionconductor-loomexperiment in #15160, should I reopen this PR and update the recipes?

dpryan79:
@nh3 It's in there, it's getting bumped to version 1.2.0. We do all bioconductor packages at once due to make the pinning easier.

nh3:
Thank you @dpryan79! Sorry that I was just searching in the commit logs rather than the changed files. When do you think `bulk` is going to be merged into `master`?

dpryan79:
It will probably take another week or two for everything in bulk to get built and uploaded, so it will get merged after that. We're also now dealing with the issue that bioconductor 3.9 is supposed to be used with R 3.6, which won't be used on the conda side until R 3.6.1 is released. That's normally not a problem, except a few core Bioconductor packages (BiocGenerics and S4Vectors chief among them) refuse to install under R 3.5. I've contacted the bioconductor core team to ask whether that's a real limitation or something that I can just ignore/patch, but if it's the former then while I'll merge bulk into master, probably 20% of bioconductor packages won't be buildable for a number of months.

nh3:
Thanks a lot for the information!

nh3:
Hi @dpryan79, I see that #15389 had merged bulk (including bioconductor-loomexperiment-1.2.0) into master 15 days ago, but somehow it's not reflected in https://bioconda.github.io/recipes/bioconductor-loomexperiment/README.html#package-bioconductor-loomexperiment and not visable in `conda search`. Is it still being built?

dpryan79:
Most of bioconductor 3.9 depends on R 3.6, which won't be available in conda-forge until after R 3.6.1 is released. Consequently, the remainder of the bioconductor packages will have to wait until then to be made available.

nh3:
Thanks @dpryan79. As I checked, there are discrepancies between the version dependencies listed at bioconductor.org and those specified in bioconda recipes. For example, on bioconductor.org, S4Vector-0.22.0 (part of release 3.9) requires BiocGenerics >=0.23.3, whereas in its bioconda recipe, it requires BiocGenerics >=0.30.0, which (perhaps unnecessarily) makes R 3.6 a requirement. Is this expected?

dpryan79:
The version listed on the bioconductor website isn’t actually correct, the bioconductor core team has requested that packages be pinned to each other as we do within releases, as they’re then guaranteed to be compatible.

nh3:
Thanks for your patient explanation. When you say the version listed on bioconductor website isn't actually correct, do you mean the lack of upper version cap or the minimum version requirement? There's no doubt that without a upper version cap (unlike conda), there's no guarantee the package will be compatible when dependency upgrades, but the question is, is the minimum version requirement listed on bioconductor.org correct or not? Can the bioconductor core team tell for each package? I had approached the LoomExperiment's author who agreed that no changes that make into release 3.9 wouldn't work with R-3.5 and the older 3.8 dependencies, and he set the dependency to R-3.5 as requested. Is the current narrow-range pinning between bioconductor recipes really a necessity for compatibility?

dpryan79:
I mean both the upper and lower version caps. Bioconductor packages are only intended to be compatible within releases, regardless of what their meta information might say or the opinion of package authors. I was also surprised by this, but this information comes directly from the bioconductor core team.

nh3:
Thanks @dpryan79! This settles it.

nsoranzo:
@nh3 See #16313 , we now have LoomExperiment 1.0.4 for R 3.5.

nh3:
Thanks @nsoranzo, I was after version >=1.1.5. As r-3.6.1 is already in conda-forge, I guess respective bioconductor 3.9 packages are getting bumped in due course. @dpryan79

nsoranzo:
@nh3 LoomExperiment 1.0.4 already contains at least the fix to correctly export rowData, not sure about the other one. So maybe you don't have to wait for BioConductor 3.9.

bgruening:
> in due course

Not sure about this :) But we are working on it. But it will probably take a month ...? 14 days rebuilding conda-forge and 14 days for Bioconda?

dpryan79:
I was hoping to do conda-forge and bioconda in parallel. There are so many noarch packages at this point that we (bioconda) shouldn't have to wait for conda-forge for most things.

nh3:
> @nh3 LoomExperiment 1.0.4 already contains at least the fix to correctly export rowData, not sure about the other one. So maybe you don't have to wait for BioConductor 3.9.

Just checked that 1.0.4 contains the fix for setting 0-based graph index, but not the fix for exporting rowData or the fix for exporting sparse matrix.

PR #13643, opened 2019-02-14
Added ProbCons package

edgano:
:information_source:

----------------


recipes/probcons/meta.yaml
```diff
@@ -0,0 +1,29 @@
+{% set version = "1.12" %}
+package:
+  name: probcons
+  version: {{ version }}
+
+build:
+  number: 1
```
bgruening: this should start with 0

edgano:
I'm not able to configure circleCI on local env. 
How can I test/fix the ```test-linux``` ?
With ```conda build``` I dont get any issue on a Linux OS

edgano:
Could you help me?
I am able to compile in local with ```conda build``` or using bioconda-utils without Docker ```bioconda-utils build recipes config.yml --git-range master```

But it fails on this PR or using Docker in bioconda-utils I get this error: ```BIOCONDA ERROR COMMAND FAILED: docker build --network host -t tmp-bioconda-builder /tmp/tmpbx8n9db_```

Thanks in advance

npavlovikj:
@edgano , if you look at the log file from CircleCI, you will see errors like:
_In file included from Main.cc:7:0:
SafeVector.h:26:22: error: expected ‘)’ before ‘size’_

So, the errors you are getting have nothing to do with Docker itself (the setup you have looks ok), but they originate when the code is compiling. Using `bioconda-utils --docker --mulled-test` is more stringent way of testing and catching errors compared to local `conda build`. Since that is what the Bioconda build system uses, I would suggest you to always use `bioconda-utils --docker --mulled-test` and fix the errors you are getting there.

Back to your error - I think the issue is in the Makefile. There, `CXX` is set to `g++` which overwrites the environment variables set within the build (you can check the values of these variables at the beginning of the log file). Commenting out the line "CXX = g++" and replacing `$(CXX)` with `$(GXX)` should fix the error you are currently getting. `GXX` is the variable that points to `g++` in the build system (it is actually `x86_64-conda_cos6-linux-gnu-g++`). When you were testing the recipe locally, you probably have `g++` already installed and set in your path, and that's why the build worked for you (it was using the local `g++`).

In addition to this, there are few other issues with the recipe.
First, change the build number to 0 as Bjorn suggested.
Second, instead of using `sed` in `build.sh`, I would suggest you to use patch, and add the changes from the Makefile there too. Using patches provides better robustness in case of future updates and changes in the code.
Thirdly, looks like `probcons` produces 4 executables (compare, makegnuplot, probcons, project), so you will need to test for all 4 of them in `meta.yaml`.

I hope you find these suggestions useful!

edgano:
On the test-macos I get this error ```O3 -W -Wall -pedantic -DNDEBUG -DNumInsertStates=2 -DVERSION="1.12" -funroll-loops -lm -o probcons Main.cc 
/bin/sh: O3: command not found```
Seems like it doesn't like the ```$(GXX) $(CXXFLAGS) -lm -o probcons Main.cc ``` solution.... 
Any hint? I have never work on MacOS

edgano:
Is there a way to skip macos test and compile it just for linux? I asume it's for a problem with the GXX compiler on mac. 
I will continue working on it.

Thanks

npavlovikj:
@edgano , I am not a Mac user, so I can not give you a clear direction from the top of my head. Maybe you can check other similar recipes in the bioconda repo and see how to set the variables correctly?
The proper way is to have both "linux" and "osx" recipes for "probcons". You can skip the "osx" build by adding the line `skip: True  # [osx]` in the `build` section. However, I would suggest you to try and fix the "osx" build first.

npavlovikj:
`CC` replaces "gcc" and `CXX` replaces "g++" in the system, so can you please fix them accordingly in the Makefile? To work on both systems you should use `CC` for `GCC` and `CXX` instead of `GXX` (what I mentioned earlier works only on Linux...).

bgruening:
ping @martin-raden

martin-raden:
@bgruening looks fine to me. ;)

PR #13660, opened 2019-02-15
Update for Gromacs 2019.1

stain:
:information_source:

----------------

### DIFF
--- a/recipes/gromacs/2019/meta.yaml
+++ b/recipes/gromacs/2019/meta.yaml
@@ -1,3 +1,15 @@
+# Update from dwnload page, e.g. http://manual.gromacs.org/2019.1/download.html
+# for latest release see
+# http://manual.gromacs.org/documentation/
+{% set name = "gromacs" %}
+{% set version = "2019.1" %}
+{% set doi_release = "10.5281/zenodo.2564764" %}
+{% set doi_doc = "10.5281/zenodo.2564761" %}
+{% set md5 = "dccfadda6a7d4ab80fcb4059606a4ef3" %}
+# Calculate manually
+{% set sha256 = "b2c37ed2fcd0e64c4efcabdc8ee581143986527192e6e647a197c76d9c4583ec" %}
+
+
 about:
   home: http://www.gromacs.org/
   license: GNU Lesser General Public License (LGPL)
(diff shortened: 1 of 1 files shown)


stain:
I am getting a bit worried about using libhwloc 2.x, as they seem to change their ABI on a whim even in patch-versions. (See https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/190 )

This could mean that any other package that needs libhwloc 2 would need to be in lock in the exact same x.y.z version (unless we do lots of tricks to allow dual-version installations).

However in BioConda only Gromacs depend on libhwloc, and in [conda-forge](https://github.com/search?q=org%3Aconda-forge+libhwloc&type=Code) only [pocl](https://github.com/conda-forge/pocl-feedstock) which in bioconda is used by [cas-offinder](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/cas-offinder/meta.yaml#L18). Nothing else in conda-forge use pocl.

I had a look "hoping" for a conflict, but it works well as both pocl and it's binary uses libhwloc 2.0.2 statically at the moment.

```
stain@biggie:~$ conda create -n gromacs-cas gromacs cas-offinder
  ... 
  cas-offinder       bioconda/linux-64::cas-offinder-2.4-0
  gromacs            bioconda/linux-64::gromacs-2019-h6bb024c_0
  ocl-icd            conda-forge/linux-64::ocl-icd-2.2.12-h14c3975_1003
  pocl               conda-forge/linux-64::pocl-1.2-h6bb024c_1003
  ...
stain@biggie:~$ conda activate gromacs-cas

(gromacs-cas) stain@biggie:~$ ln -s /etc/OpenCL/vendors /home/stain/miniconda3/envs/gromacs-cas/etc/OpenCL/vendors

(gromacs-cas) stain@biggie:~$ gmx -version | egrep -i '(Hw|GPU)'
GPU support:        OpenCL
Hwloc support:      hwloc-2.0.2

(gromacs-cas) stain@biggie:~$ cas-offinder -h | tail -n3
Available device list:
Type: GPU, ID: 0, <NVS 5200M> on <NVIDIA CUDA>

(gromacs-cas) stain@biggie:~$ poclcc -l
LIST OF DEVICES:
0:
  Vendor:   NVIDIA Corporation
    Name:   NVS 5200M
 Version:   OpenCL 1.1 CUDA
```

.. so perhaps nothing to worry about for now..?

recipes/gromacs/2019/meta.yaml
```diff
@@ -44,7 +53,7 @@ test:
 extra:
   identifiers:
     - biotools:gromacs
-    - doi:10.5281/zenodo.2424363
-    - doi:10.5281/zenodo.2424486
+    - doi:{{ doi_release }}
```
bgruening: @stain I'm not sure if this is overdoing it.

It complicates downstream tools that just want to read metadata and with this needs to render the file.

stain: Agree, I'll change it back. 

A related question here is which DOI - there is one for each release (Zenodo archive), one for documentation, and one for "cite me". Hope all are OK for now, [Citation File Format](https://citation-file-format.github.io/) could be an alternative that is (too) detailed.

PR #13663, opened 2019-02-15
Barcode splitter multi

hepcat72:
:information_source:

----------------


hepcat72:
Huh.  What do these mean: `You're pointing it at a single subdir.  Please update your code to point it at the channel root, rather than a subdir.`?

And I think this may be referring to the `run_test.sh` script: `Passing program arguments as a string may be a security hazard if combined with untrusted input`?  But I'm not entirely sure.

If either of these are the case, I'm curious as to why my local `circleci build` passed...

```
The update_index function has changed to index all subdirs at once.  You're pointing it at a single subdir.  Please update your code to point it at the channel root, rather than a subdir.
WARNING:conda_build.index:The update_index function has changed to index all subdirs at once.  You're pointing it at a single subdir.  Please update your code to point it at the channel root, rather than a subdir.
The update_index function has changed to index all subdirs at once.  You're pointing it at a single subdir.  Please update your code to point it at the channel root, rather than a subdir.
WARNING:conda_build.index:The update_index function has changed to index all subdirs at once.  You're pointing it at a single subdir.  Please update your code to point it at the channel root, rather than a subdir.
..16:08:33 BIOCONDA ERROR COMMAND FAILED: mulled-build build-and-test barcode_splitter=0.18.3--py_0 -n biocontainers --test run_test.sh --channels conda-forge,file:///home/circleci/project/miniconda/conda-bld,bioconda,defaults --involucro-path /home/circleci/project/miniconda/bin/involucro
16:08:33 BIOCONDA ERROR STDOUT+STDERR:
/home/circleci/project/miniconda/bin/involucro -v=2 -f /home/circleci/project/miniconda/lib/python3.6/site-packages/galaxy/tools/deps/mulled/invfile.lua -set CHANNELS='conda-forge,file:///home/circleci/project/miniconda/conda-bld,bioconda,defaults' -set TARGETS='barcode_splitter=0.18.3=py_0' -set REPO='quay.io/biocontainers/barcode_splitter:0.18.3--py_0' -set BINDS='build/dist:/usr/local/,/home/circleci/project/miniconda/conda-bld:/home/circleci/project/miniconda/conda-bld' -set CONDA_IMAGE='continuumio/miniconda3:4.3.27' -set TEST=run_test.sh build-and-test
.Passing program arguments as a string may be a security hazard if combined with untrusted input
```

hepcat72:
OK. I re-consulted the contrib doc.  Previously, I'd thought the bootstrap test was only useful if the local circleci build fails.  Now I see it's necessary to test mulled-builds.  Going through that now.

hepcat72:
It appears that when I bootstrap to run the mulled build, the test fails because `set -e` appears to be the system default.  I was ablt to add `set +e` at the top of the script to get the mulled build to continue the script (because I expect a non-zero exit code), but for some reason, the test still fails - but only in the mulled build...

hepcat72:
OK, I tried changing to a new release with a fix for the exit code thing so I didn't have to use a run_test.sh script, but the mulled build still fails.  I can't seem to figure this out.

hepcat72:
I pushed and am just crossing my fingers that the remote circleci will work with the mulled build.  I can't think of anything else to do.

hepcat72:
OMG, it worked!

.circleci/common.sh
```diff
@@ -0,0 +1,2 @@
+MINICONDA_VER=4.5.4
```
bgruening: @hepcat72 can you please remove those changes.

recipes/barcode_splitter/run_test.sh
```diff
@@ -1,11 +0,0 @@
-#!/bin/bash
-
```
bgruening: I'm sorry. The reason is that our current Docker tests do not support additional files. These files, run_tests.sh in your case, are not included in the Docker container and therefor can not be executed. Sorry for the confusion.

.circleci/common.sh
hepcat72: Whoops.  Sorry about that. Unintended commit.

PR #13680, opened 2019-02-17
MAINT: Update parasail-python to 1.1.16

jvkersch:
@bioconda/core Could I ask for a review? This is not a new recipe, but it is my first PR to this repo and I wanted to make sure everything is OK. Thanks!

----------------

### DIFF
--- a/recipes/parasail-python/meta.yaml
+++ b/recipes/parasail-python/meta.yaml
@@ -18,9 +18,9 @@ build:
 requirements:
   host:
     - python
+    - setuptools
     - zlib
   build:
-    - setuptools
     - {{ compiler('c') }}
     - {{ compiler('cxx') }}
     - m4
--- a/recipes/parasail-python/meta.yaml
+++ b/recipes/parasail-python/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "parasail-python" %}
-{% set version = "1.1.12" %}
+{% set version = "1.1.16" %}
 
 package:
   name: "{{ name|lower }}"
@@ -7,10 +7,10 @@ package:
 
 source:
   url: https://github.com/jeffdaily/parasail-python/archive/v{{ version }}.tar.gz
-  md5: 109fc497924030eebda60bde33f0976b
+  md5: aacd7de464f24ec64daaae2c12ca8c6c
 
 build:
-  number: 2
+  number: 1
 
 requirements:
   host:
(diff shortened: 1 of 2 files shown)


recipes/parasail-python/meta.yaml
```diff
@@ -1,16 +1,16 @@
...
+{% set version = "1.1.16" %}
 
 package:
   name: "{{ name|lower }}"
   version: "{{ version }}"
 
 source:
   url: https://github.com/jeffdaily/parasail-python/archive/v{{ version }}.tar.gz
-  md5: 109fc497924030eebda60bde33f0976b
+  md5: aacd7de464f24ec64daaae2c12ca8c6c
```
druvus: Could you swap to sha256?

recipes/parasail-python/meta.yaml
```diff
@@ -1,16 +1,16 @@
...
 
 build:
-  number: 2
+  number: 1
```
druvus: Should start at 0 for new version

jvkersch:
@druvus Thank you for the swift reply! I made the changes you suggested, but there is a build failure due to a linker incompatibility. I think this is due to Parasail prepending `/usr/bin` to the PATH in `setup.py`. I will need some time to properly test for this, and to make a patch.

jvkersch:
I got the Linux build to work (the recipe was lacking a dependency on the C++ compiler, which caused the linking issues) but there's something wrong with the Mac OS build. On Mac OS, parasail seems to insist on downloading and installing the libtool toolchain, I am not sure that that is necessary in a controlled environment like the Bioconda build environment.

jvkersch:
In the end I did have to patch `setup.py` on Mac OS to account for a missing tool name (libtool). My assumption is that the parasail-python author was relying on the Homebrew-installed version of these tools, all of which start with a 'g' prefix to distinguish them from the system versions.

@druvus Would you have time for another round of reviews?

jvkersch:
@druvus Would you mind giving this another look? Having an updated version of this package would help with some other projects I'm working on. Thanks!

recipes/parasail-python/meta.yaml
```diff
@@ -21,6 +24,7 @@ requirements:
     - pip
     - setuptools
     - {{ compiler('c') }}
```
bgruening: In build should be only the compilers autoconf, automake, libtool and m4 I think.

jvkersch: @bgruening That makes sense, thanks for pointing that out. I've updated the recipe to leave only the build tools under build, and moved the rest to host. I've also removed numpy as a build-time requirement.

jvkersch:
@bgruening Ready for another round of reviews at your convenience. Thanks!

bgruening:
Merci @jvkersch!

PR #13702, opened 2019-02-19
Updated portcullis to the most recent version (1.1.2)

lucventurini:
:information_source:

----------------


maplesond:
Hello.  We are getting errors in the linux build:

```
/opt/rh/devtoolset-2/root/usr/libexec/gcc/x86_64-redhat-linux/4.8.2/ld: /opt/conda/conda-bld/portcullis_1550581146757/work/deps/htslib-1.3/.libs/libphts.a(libphts_la-rANS_static.o): unrecognized relocation (0x29) in section `.text'

/opt/rh/devtoolset-2/root/usr/libexec/gcc/x86_64-redhat-linux/4.8.2/ld: final link failed: Bad value

```

I suspect that this is to do with conflicts between bioconda's boost and libhts dynamic linking dependencies interfering with those built by portcullis, which are statically linked.  Is there some way of excluding boost and libhts from the build environment?

marcelm:
Sorry, this is too much to go through for me. You should probably explain a little bit more what you are trying to do and what doesn’t work in order to inspire someone to help you. My thoughts so far:
* You should rebase on top of master (it’s now building a couple of unblacklisted recipes)
* Please set the build number to 0.
* You should very likely not create a subdirectory for the new version, see https://bioconda.github.io/guidelines.html#versions (that is, place the `meta.yaml` directly under the `recipes/portcullis/` directory instead). This would also make it easier to see what has changed between recipe versions. (I would now need to run a manual diff to see what changed between 1.1.1 and 1.1.2.)
* Not sure if that is feasible, but could you un-vendor the included boost and htslib and instead link against the conda versions? This would reduce build time and installation size. And it would also help in pinpointing the problem because when using the conda versions, it would probably be safe to assume that they were built with the proper flags.

recipes/portcullis/1.1.2/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - samtools >=1.9
+    - perl
+    - setuptools
+    - tabulate
+    - numpy >=1.15
+    - pandas
+    - samtools >=1.9
+    - python    
+  host:
+    - autoconf
```
marcelm: I’m myself not 100% sure, yet, about what should go into `build:` and `run:` requirements, but [according to the documentation](https://docs.conda.io/projects/conda-build/en/latest/source/resources/define-metadata.html#requirements-section), build tools such as `autoconf`, `automake` and compilers should be listed under `build:`. Anything specific to the target system such as libraries needs to be under `host:`.

lucventurini:
Dear @marcelm,
                             thank you for taking a look at the issue. The problem with using Conda Boost and HTSLib is (from what I understand) that Portcullis requires them to be compiled statically rather than dynamically. This creates problems with the default versions installed by Conda. @maplesond can elaborate on the point, and on the feasibility of using the Conda libraries, much better than me, as I am personally not a C/C++ coder.

marcelm:
It depends on the program somewhat, but it should usually be possible to switch from static to dynamic linking. Possibly Portcullis used static linking to work around the problem of letting the binary find the dynamic libraries. But since Conda would take care of that problem, it could just work. But then I’m just speculating, and I also don’t know whether that would actually solve your problem.

One other thought: You should probably not set `LD_LIBRARY_PATH` in the build script. Using that variable is in my opinion a hack that is mostly useful for debugging or as a workaround in some special cases. There are also [some warnings about using it in the conda documentation](https://docs.conda.io/projects/conda-build/en/latest/source/resources/use-shared-libraries.html).

lucventurini:
Hi @marcelm ,
                         I double checked, and the Makefile *explicitly* uses the libraries installed in deps/ inside the software directory - ie, the bundled Boost, HTSlib and Ranger. This implies that to solve the issue as you recommend, @maplesond would have to heavily amend the Makefile.

maplesond:
Hi @marcelm, @lucventurini is correct.  Portcullis is designed to use its internal versions of boost and htslib.  They are all statically linked into the executable so will not pollute the wider environment.  We have a functionality [docker file](https://github.com/maplesond/portcullis/blob/develop/Dockerfile) to see the steps required to build portcullis from scratch in case that's useful for you.

marcelm:
Ok, then you should not list boost under any requirements sections in the recipe (unless something else needs it).

marcelm:
My (also failing) attempt is in the `portcullis-marcelm` branch. I haven’t opened a PR, but https://github.com/bioconda/bioconda-recipes/compare/portcullis-marcelm will allow you to see the CircleCI logs. When I run it locally, it fails at `make check` with
```
.../_build_env/bin/../lib/gcc/x86_64-conda_cos6-linux-gnu/7.3.0/../../../../x86_64-conda_cos6-linux-gnu/bin/ld: ../deps/boost/build/lib/libboost_timer.so: undefined reference to `memcpy@GLIBC_2.14'
```

maplesond:
Hi @marcelm, maybe I'm missing something but as far as I can tell we were not trying to include the conda boost into portcullis.  This seems to be provided in conda whether you want it or not.  If there's some way of turning it off that I'm missing we'd be grateful for the suggestion.
Also, I think we weren't requesting that `make check` should run (at least in the original reciepe).  I'm happy for this step to be skipped as there are various linking issues that may legitimately crop up, providing that the main executables start: `portcullis` and `junctools`.

marcelm:
The most recent version of the `meta.yaml` in this PR includes `boost` as a build requirement, see the 'Files changed' tab above. There were lots of commits, so perhaps it was not included from the beginning. Same for `make check`: It is in the current version of the `build.sh` script, but perhaps it was added by @lucventurini during one of the commits for debugging purposes.

Anyway, I now get the same failures as @lucventurini got now on CircleCI. Apparently something on my local system was interfering that made the build go through.

lucventurini:
Yes, I confirm I did try to include explicitly boost at a certain point, for debugging purposes; same goes for the make check.

@marcelm thank you for testing this. For some of the errors, I got around them updating the CFLAGS etc. (e.g. adding the -fPIC switch).

marcelm:
Apparently boost is compiled by the system compiler, which is gcc 4.8.2, while the rest uses gcc 7 (installed by conda). I think this is the main problem. It seems the standard mechanism of setting the `$CXX` etc. variables doesn’t work. This is a line from one of the log outputs:
```
gcc.compile.c++ bin.v2/libs/system/build/gcc-gnu-4.8.2/release/threadapi-pthread/threading-multi/error_code.o
```
The [Boost recipe in conda-forge](https://github.com/conda-forge/boost-feedstock/blob/master/recipe/build.sh) has some code in its build script to deal with this.

Please do consider modifying portcullis in a future version such that it can use an external boost instead. In the long run, my impression is that it would save trouble, at least with the bioconda recipe.

marcelm:
So here is my final suggestion. I have just copied over and adjusted a bit the conda-forge/boost-feedstock build script. The branch is here: https://github.com/bioconda/bioconda-recipes/tree/portcullis-marcelm/recipes/portcullis/ . It fails "only" at `make check` now.
There’s still some debugging stuff, so it is not usable as-is, but feel free to copy the parts you want over to your PR. This nearly exhausts the time I can spend on this, so I won’t do any further work in that branch.

lucventurini:
Dear @marcelm , thank you so much for your time and help. We will try out the revised recipe as soon as possible.

lucventurini:
The last recipe failed because there was a mistake in the recipe for uscs tools which we do not use. It might be an issue with the Conda repository for bioconda; I will relaunch it in a little bit.

marcelm:
You need to rebase on top of master or merge master to get rid of the problem with ucsc tools.

lucventurini:
@marcelm , now the recipe functions, thanks a lot!
@maplesond, if you think that the recipe is now OK, we can squash and merge everything into the main branch.

maplesond:
Fantastic!  @lucventurini if you're happy you can access portcullis and junctools I'm happy for this to go through.  Thanks both for all the work done here!

lucventurini:
@maplesond, not a problem! @marcelm deserves most of the merit here :-)

bgruening:
LGTM. In general it would be preferable to use the conda boost and other libs if possible as @marcelm already pointed out. Maybe this could be a long term plan?

marcelm:
Glad I could help!

PR #13706, opened 2019-02-19
updated scprep v0.11.0

scottgigante:
:information_source:

----------------

### DIFF
--- a/recipes/scprep/meta.yaml
+++ b/recipes/scprep/meta.yaml
@@ -1,13 +1,13 @@
 {% set name = "scprep" %}
-{% set version = "0.8.1" %}
+{% set version = "0.10.2" %}
 
 package:
   name: "{{ name|lower }}"
   version: "{{ version }}"
 
 source:
-  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


recipes/scprep/meta.yaml
```diff
@@ -1,13 +1,13 @@
...
+{% set version = "0.10.2" %}
 
 package:
   name: "{{ name|lower }}"
   version: "{{ version }}"
 
 source:
-  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
-  sha256: 4da05d30f5841c977a091a66e5367d85ad3c64f7b8b51ff84a3847fce322fb49
+  url: https://github.com/KrishnaswamyLab/{{ name|lower }}/archive/v{{ version }}.tar.gz
```
stain: Any reason why the old pypi.io link is not good enough? https://pypi.io/packages/source/s/scprep/scprep-0.10.2.tar.gz works on the checksum adf104ddc05faf5893a363e6c79f2ce55915552e8c08c8c13657d2bfd26bcf5e can be verified on https://pypi.org/project/scprep/#files

scottgigante: I just changed it to keep it consistent with the recipes I have on conda-forge, and avoid an extra step in the chain (github -> pypi -> conda). If you feel strongly that pypi is better I'm happy to revert.

PR #13732, opened 2019-02-21
Update CPAT to 1.2.4

rhpvorderman:

Cpat is installable from Pypi now. But this does not install the r-dependency. So long live conda! Also this will make a biocontainer for easy usage. 

Build the recipe using conda skeleton and updated the existing one accordingly.


rhpvorderman:
@bioconda/core 
Technically this is not a new package, neither is it my first package. But this package was made by somebody else, I completely overhauled it _and_ it is my first python package on conda.
I would feel a lot better if this was reviewed before the merge button is pressed. Thanks!

dpryan79:
Can you try enabling building on python 3? I think the most recent version of bx-python supports it finally.

rhpvorderman:
> Can you try enabling building on python 3? I think the most recent version of bx-python supports it finally.

No. Because CPAT is written entirely in 2.7. (Yes it sucks, but luckily we can run it in a container).

recipes/cpat/meta.yaml
```diff
@@ -1,30 +1,36 @@
...
 package:
-  name: cpat
-  version: 1.2.3
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
 
 build:
-  number: 1
-  skip: True # [not py27] # because of bx-python
+  noarch: true
```
dpryan79: You presumably mean `noarch: python`

rhpvorderman: Thanks!

rhpvorderman: @dpryan79 it is done. Thanks again.

dpryan79:
Nice job @rhpvorderman !

PR #13736, opened 2019-02-21
Add fastq-pair v1.0 recipe

peterk87:


recipes/fastq-pair/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/linsalrob/{{ name }}/archive/v{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 1
```
bgruening: please start with 0

recipes/fastq-pair/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+requirements:
+  build:
+    - cmake >=3.6
+    - {{ compiler('c') }}
+
+test:
+  files:
```
bgruening: We usually do not include test data in the repo. The command test should be enough.

If you like you can include a `make test` is it exists.

peterk87: Okay, thanks for the advice. I removed the test data and test script. Unfortunately, there's no `make test` for this tool.

recipes/fastq-pair/meta.yaml
peterk87: Changed to 0

druvus:
thanks @peterk87! fastq-pair was added in #13997

PR #13748, opened 2019-02-22
initial commit of Q (peak caller)

visze:
:information_source:

----------------


recipes/q/meta.yaml
```diff
@@ -0,0 +1,35 @@
+{% set version = "1.3.1" %}
+
+package:
+  name: samtools
```
bgruening: that looks wrong ... can we name it maybe `q_peakcaller` or something like that?

visze: thanks börn. copy and paste error

recipes/q/meta.yaml
```diff
@@ -1,7 +1,7 @@
 {% set version = "1.3.1" %}
 
 package:
-  name: samtools
+  name: Q
```
bgruening: @visze this should be small letters. But it would also be great if the name could be longer. We only have a namespace of 26 for single-letter tools ;)

visze: it is not my program. I just want to use it.. I can give him any name but no idea if the developer will like it

dpryan79:
@bioconda-bot please update

dpryan79:
This vendors an old version of seqan that's not compatible with more recent C++ standards, so I'll close it.

PR #13770, opened 2019-02-24
Add pomoxis recipe

cjw85:
:information_source:

----------------

### DIFF
--- a/recipes/pomoxis/meta.yaml
+++ b/recipes/pomoxis/meta.yaml
@@ -10,8 +10,8 @@ source:
   sha256: 4001024e7c2b200e4924c2924f3c423e3f7464b51501619b3c310f51b5120df9
 
 build:
-  number: 1
-  skip: True # [not py35 or not py36]
+  number: 0
+  noarch: python
   entry_points:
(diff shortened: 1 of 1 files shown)


recipes/pomoxis/meta.yaml
```diff
@@ -0,0 +1,100 @@
...
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 4001024e7c2b200e4924c2924f3c423e3f7464b51501619b3c310f51b5120df9
+
+build:
+  number: 1
```
bgruening: that should be 0

recipes/pomoxis/meta.yaml
```diff
@@ -0,0 +1,100 @@
...
+  skip: True # [not py35 or not py36]
```
bgruening: You can remove this as you pinned the version in the host and run section.
Instead you could try to add `noarch: python`

PR #13777, opened 2019-02-24
Tra genepop 4.7

tiagoantao:
This upgrades genepop to 4.7
----------------


stain:
Do we need both 4.6.x and 4.7.x? 

If so then better have two folders for that, I think it's confusing to have one `4.6.9/meta.yml` and one `meta.yml`

tiagoantao:
@stain Sorry for the delay in answering.
I have changed the directory structure.
Version 4.6 is still widely used and 4.6/4.7 and not totally compatible. E.g. Not having 4.6 would break existing versions of Biopython's PopGen module

bgruening:
> If so then better have two folders for that, I think it's confusing to have one 4.6.9/meta.yml and one meta.yml

But this is in general easier for automation. We essentially now without any heuristic where the latest version is stored. Important for any bio.tools integration or any bots.

recipes/genepop/4.7.0/meta.yaml
```diff
@@ -0,0 +1,26 @@
+package:
+  name: genepop
+  version: "4.7.0"
+
+build:
+  number: 1
```
bgruening: This needs to start with 0.

recipes/genepop/4.7.0/meta.yaml
```diff
@@ -0,0 +1,26 @@
+package:
+  name: genepop
+  version: "4.7.0"
```
bgruening: The version seems to be wrong, according to the link. Can you please use {{ verssion }} instead.

bgruening: Or the link is wrong :)

tiagoantao: @bgruening : First my apologies for taking so long.
In this case its neither: after talking with the author, we decided that for existing genepop versions we would take the out from the exinsting R packages. So, the R package is 1.0.5 but Genepop is 4.7.0. People do not really know the R package version.
In the future (hopefully) a standalone version of genepop will be provided (actually it is provided and always was, but the author did not kept old versions, just replaced the source code zip).
All this kerfufle is just to make sure we are able to maintain exact versions (reproducibility and all).
What probably makes sense is add this comment to the yaml, what do you think?

bgruening: Oh yes, please add this, otherwise it is very confusing :)

tiagoantao: Added the comment to the yaml file.

PR #13797, opened 2019-02-25
Add spectrum_utils recipe

bittremieux:

---

New recipe for the [`spectrum_utils`](https://github.com/bittremieux/spectrum_utils) Python package containing utility functions for mass spectrometry proteomics.

Simple recipe based on the corresponding PyPI package.

This is my first recipe, so I'd appreciate a review by @bioconda/core.

### DIFF
--- a/recipes/spectrum_utils/meta.yaml
+++ b/recipes/spectrum_utils/meta.yaml
@@ -16,18 +16,14 @@ build:
 
 requirements:
   host:
-    - matplotlib
-    - numba
-    - numpy
+    - python >=3
     - pip
-    - pyteomics
-    - python >=3.*
   run:
     - matplotlib
     - numba
     - numpy
     - pyteomics
-    - python >=3.*
+    - python >=3
 
 test:
   imports:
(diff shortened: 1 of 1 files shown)


stain:
Thanks for contributing!

See https://bioconda.github.io/cb3.html?highlight=host%20build#the-new-host-section - I think all your dependencies can be under `host` and you should not need `pip`.

dpryan79:
Sorry, I missed this ping. As @stain mentioned, dependencies should normally be in the `host:` section. I've changed that and also removed the `extra:` section, which isn't used by bioconda.

bittremieux:
Thanks @stain and @dpryan79.

Possibly most of the current `host` dependencies, except Python and pip, are actually also only needed under `run` I think?

And as this is a pure Python package should I use `noarch` instead or are the mentioned [technical compatibility issues](https://bioconda.github.io/guidelines.html?highlight=noarch#id3) still present?

dpryan79:
@bittremieux Good catch, I've taken the liberty of adding the `noarch: python` line and also fixed the pinning to python to work with it. I assume I'm correct that this was python 3 only, otherwise please fix that before merging.

PR #13816, opened 2019-02-26
Longshot recipe

pjedge:
:information_source:

----------------

* [ X] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).
* [] This PR updates an existing recipe.


pjedge:
If someone could review this recipe for me, that'd be great. @bioconda/core
It's written in Rust, and the recipe is based on the rust-bio-tools recipe with minimal modification.

dpryan79:
@johanneskoester I think you're the rust expert among us, can you have a look?

pjedge:
Hi, can someone please review this recipe for me?

recipes/longshot/build.sh
```diff
@@ -0,0 +1,18 @@
...
+# this build script is taken from the rust-bio-tools recipe
+# https://github.com/bioconda/bioconda-recipes/blob/master/recipes/rust-bio-tools/build.sh
+
+# taken from yacrd recipe, see: https://github.com/bioconda/bioconda-recipes/blob/2b02c3db6400499d910bc5f297d23cb20c9db4f8/recipes/yacrd/build.sh
+if [ "$(uname)" == "Darwin" ]; then
+
+    # apparently the HOME variable isn't set correctly, and circle ci output indicates the following as the home directory
+    export HOME="/Users/distiller"
+
+    # according to https://github.com/rust-lang/cargo/issues/2422#issuecomment-198458960 removing circle ci default configuration solves cargo trouble downloading crates
```
dpryan79: I'll show my ignorance of rust here (I really should start a project using it), but what dependencies or other crates does this recipe require? Ideally we'd make those conda packages too (assuming that makes any sense at all for rust).

recipes/longshot/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - make
+    - {{ compiler('c') }}  # [not osx]
+    - {{ compiler('cxx') }}  # [not osx]
+    - clang_osx-64 # [osx]
+    - clangxx_osx-64 # [osx]
+    - pkg-config  # [osx]
+  host:
+    - zlib
+    - xz
+    - clangdev
```
dpryan79: Do you really need this? If so, even under Linux?

recipes/longshot/build.sh
pjedge: The recipe requires a whole host of crates, but they are managed automatically by cargo. I don't think it's feasible to have them be managed by conda instead. I based this recipe off Johannes' rust-bio-tools recipe (very little is changed, it basically specifies some non-rust dependencies and calls ```cargo install```).

recipes/longshot/meta.yaml
pjedge: Yeah they are definitely all required. If I remember correctly, all three are rust-htslib dependencies.

dpryan79:
Thanks, looks fine to me then (not that that's worth terribly much, given my obvious lack of familiarity with rust).

pjedge:
Thanks. I think I may publish a second release in the coming days (recipe will be the same except version number bump), and if I do that can I merge it myself?

dpryan79:
Unless someone else from the core team objects before then I would take that at tacit approval.

recipes/longshot/build.sh
johanneskoester: Rust statically compiles pure-rust dependencies stuff into the binary. This is all fine. Those deps usually very small.

recipes/longshot/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  sha256: ec8ebfb6007d42c6952bac6843e903f899f33d6c0fadbd976cc272d8d7f4b279
+
+requirements:
+  build:
+    - rust >=1.32
+    - make
+    - {{ compiler('c') }}  # [not osx]
+    - {{ compiler('cxx') }}  # [not osx]
+    - clang_osx-64 # [osx]
+    - clangxx_osx-64 # [osx]
```
johanneskoester: Can't we use the compiler function here as well? I think the default compiler of macos is clang anyway.

pjedge: I removed those and left just the compiler function -- passed tests. thanks!

bgruening:
Thanks @pjedge!

pjedge:
Thanks for all your help! I tested installation on a centos machine and it worked great! However, I asked a friend to try installation on their macbook and it failed with the following error:
```
PackageNotFoundError: Packages missing in current channels:
 - longshot -> libcxx >=4.0.1
```
Any idea what I can do about this?

dpryan79:
@pjedge Make sure they still have `defaults` in their list of channels (that's where that package should be coming from).

PR #13834, opened 2019-02-27
Add sts-0.1

dfornika:
:information_source:

----------------


recipes/sts/meta.yaml
```diff
@@ -0,0 +1,53 @@
+{% set name = "sts" %}
+{% set version = "0.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+build:
+  skip: True  # [osx or win]
```
bgruening: you do not need to skip win on Bioconda

recipes/sts/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+git submodule init
```
bgruening: I'm not sure this is good? This is not reproducible right?

recipes/sts/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - git
+  host:
+    - cmake
+    - pkg-config
+    - boost
+    - gsl 1.16
```
bgruening: gsl will be pinned by conda automatically

recipes/sts/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  git_url: https://github.com/OnlinePhylo/sts.git
+  git_rev: v{{ version }}
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - git
+  host:
+    - cmake
+    - pkg-config
```
bgruening: pkg-config and cmake should be in the build section ... git (if you still needed it) should be in the host section

recipes/sts/build.sh
dfornika: I'm not certain about reproducibility. I think that the submodules are associated with a specific commit, so if we're pulling a specific tag of the parent repo (`sts`) then it seems to me that it should be reproducible:

https://github.com/OnlinePhylo/sts/tree/v0.1/lib

I've packaged the two submodules (`lcfit` and `smctc`) as separate bioconda packages too, but using those would involve making changes to the CMake files so it seems like a more complicated solution. I had tried doing that some time ago and could try again if the submodules aren't acceptable. But I've been chipping away at this recipe for a long time so I'm trying to find the simplest solution that will complete the build at this point.

epruesse: Since we "don't like" git_url anyway - perhaps you can just prepare a tar ball with the sources and upload it to `cargo-port`?

epruesse: The problem here is this: `sts` is under the GPL, which mandates that we must not distribute binaries without both sources and instructions on how to build them. For that reason, we have `cargo-port` which pulls and archives the sources after a package is built. I'm not sure it works with `git_url`, but it definitely won't work with `git submodule init` in the `build.sh`.

epruesse: You might have luck with `CPack` to build a source archive. I'm not very familiar with CMAKE though. Ideally, upstream would use a CI to create complete source archives and add them to their release tags.

epruesse:
@biocondabot lint

dpryan79:
@bioconda-bot please update

dpryan79:
This requires an ancient version of GSL that conflicts with global bioconda and conda-forge pinning and thus cannot be merged in in the foreseeable future. Consequently I'll close this.

PR #13839, opened 2019-02-28
Multiphate

carolzhou:
:information_source:

----------------


multiphate/meta.yaml
```diff
@@ -0,0 +1,29 @@
+package:
+  name: multiphate 
+  version: 0.5 
+
+source:
+  git_url: https://github.com/carolzhou/multiPhATEb.git
+  sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+
+build:
+  noarch: python
```
bgruening: you could also add `script: "{{ PYTHON }} -m pip install . --no-deps -vv"` and remove the build.sh script ...

multiphate/meta.yaml
```diff
@@ -0,0 +1,29 @@
+package:
+  name: multiphate 
+  version: 0.5 
+
+source:
+  git_url: https://github.com/carolzhou/multiPhATEb.git
```
bgruening: Can you create a release and use a release tarball instead?

Thanks!

carolzhou: Yes, I have created a release tarball and resubmitted the recipe for building.

carolzhou: Actually, I though the build failed because of the git_url. ???

multiphate/meta.yaml
carolzhou: Do you mean, replace the "build:  noarch: python" with what you have written there? Should I do this and resubmit the PR?

multiphate/meta.yaml
carolzhou: And I thought another PR failed due to an emboss problem.  ??? I have to say this is a confusing process.  (Sorry to complain.)

multiphate/meta.yaml
bgruening: Have a look at this one for example: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/hicexplorer/meta.yaml

You can just get rid of the additional build.sh file when your installation command is really simple.

multiphate/meta.yaml
bgruening: ? I don't see any changes here also no emboss failure.

carolzhou: This is the error message I copied out, and why I re-submitted a PR. Maybe I should let the most recent version go through again, since I changed the meta.yaml (with a copy in the multiphate.tar.gz), which changes the sha256.  Sorry for the confusion.  My first time with Conda/bioconda packaging.  07:01:40 BIOCONDA ERROR BUILD SUMMARY: SKIPPED recipe multiphate due to failed dependencies ['recipes/perl-file-path', 'recipes/perl-file-temp/0.2304', 'recipes/perl-test-simple', ‘recipes/emboss']

carolzhou: That was for #13872

multiphate/meta.yaml
carolzhou: Thank you. That's a very informative example.

bgruening: But why haven't you used it?

multiphate/meta.yaml
bgruening: Why haven't you used this URL instead? https://github.com/carolzhou/multiPhATEb/archive/a706cac23339588fe9fa4a34d607fcdcabccfe5a.zip

bgruening:
It's really not ok to merge your own PRs if there are not addressed comments :(

bgruening:
Please see my PR a follow-up pull request here: https://github.com/bioconda/bioconda-recipes/pull/13900

PR #13893, opened 2019-03-04
update savage to version 0.4.1

jbaaijens:
:information_source:

----------------

### DIFF
--- a/recipes/savage/meta.yaml
+++ b/recipes/savage/meta.yaml
@@ -1,16 +1,16 @@
-{% set version = "0.4.0" %}
+{% set version = "0.4.1" %}
 
 package:
   name: savage
   version: {{ version }}
 
 build:
-  number: 2
+  number: 0
   skip: True  # [py3k or osx]
 
 source:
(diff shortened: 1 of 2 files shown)


recipes/savage/meta.yaml
```diff
@@ -1,16 +1,16 @@
...
 
 build:
-  number: 2
+  number: 0
   skip: True  # [py3k or osx]
 
 source:
-  url: https://bitbucket.org/jbaaijens/savage/get/v{{ version }}.tar.gz
-  md5: eb1622cdd8a22c4d6c7fd34dc064b3cd
+  url: https://github.com/HaploConduct/HaploConduct/archive/v{{ version }}.tar.gz
```
johanneskoester: A name change? Shouldn't we have a new package called haploconduct then?

jbaaijens: Yes, I will make a package for haploconduct soon, but it's not ready yet. The latest version of savage is in the haploconduct repo, however, so I had to change the url for this update. Thanks for your help!

PR #13901, opened 2019-03-04
multiphate

bgruening:
:information_source:

----------------


epruesse:
Need to try something:

@biocondabot lint

epruesse:
Need to try something:

@biocondabot lint

carolzhou:
Just to let you know where I am with this, conda-build is building the multiPhATE package, and all tests pass except the last one, when invoking my own code (e.g, something like, "multiphate usage"). I have not been able to figure out why this test fails, why Conda cannot recognize the package. I have been trying to encode an "entry point", but so far this is not working. I am consulting with others on this. I will re-submit to Bioconda when I get this worked out. If you have suggestions, I am listening. Thank you for your patience.

epruesse:
This was the error:

```
/opt/conda/conda-bld/multiphate_1551774985838/test_tmp/run_test.sh: line 9: multiPhate.py: command not found
```

Which is probably because the file is missing. Depending on what you put into your `setup.py`, it's possible the command is called `multiPhate` (w/o `.py`).

epruesse:
You should be able to just test this locally. If you run `pip install -e .`, the `setup.py` will be invoked to install things. After that, you should find `command -V multiPhate.py` if it's working. Looking at your `setup.py`, you could e.g. just add a `scripts` parameter as described here: https://docs.python.org/3/distutils/setupscript.html#installing-scripts

carolzhou:
Thanks, I'll give that a try.  :-)

raivivek:
@bioconda-bot please update

druvus:
@bioconda-bot please update

Issue #12100, opened 2018-11-15
samtools 1.9 dependency pulls in wrong version of openssl


jamie-arcc:
`samtools` fails to run successfully due to a mismatch in library versions.  The samtools executable is expecting `libcrypto.so.1.0.0` but openssl 1.1.1 is getting installed, which has `libcrypto.so.1.1.1`.

```
[jschnaitter@emgnt1 ~]$ conda create -n test-samtools samtools
Solving environment: done

## Package Plan ##

  environment location: /apps/anaconda/anaconda3-5.3.0/envs/test-samtools

  added / updated specs:
    - samtools


The following NEW packages will be INSTALLED:

    bzip2:           1.0.6-h14c3975_5
    ca-certificates: 2018.03.07-0
    curl:            7.62.0-hbc83047_0
    libcurl:         7.62.0-h20c2e04_0
    libdeflate:      1.0-h470a237_0    bioconda
    libgcc-ng:       8.2.0-hdf63c60_1
    libssh2:         1.8.0-h1ba5d50_4
    libstdcxx-ng:    8.2.0-hdf63c60_1
    ncurses:         6.1-hf484d3e_0
    openssl:         1.1.1-h7b6447c_0
    samtools:        1.9-h8ee4bcc_1    bioconda
    xz:              5.2.4-h14c3975_4
    zlib:            1.2.11-ha838bed_2

Proceed ([y]/n)?

Preparing transaction: done
Verifying transaction: done
Executing transaction: done
#
# To activate this environment, use
#
#     $ conda activate test-samtools
#
# To deactivate an active environment, use
#
#     $ conda deactivate

[jschnaitter@emgnt1 ~]$ source activate test-samtools
(test-samtools) [jschnaitter@emgnt1 ~]$ samtools --version
samtools: error while loading shared libraries: libcrypto.so.1.0.0: cannot open shared object file: No such file or directory
(test-samtools) [jschnaitter@emgnt1 ~]$ ldd $(which samtools)
	linux-vdso.so.1 =>  (0x00007ffc61dcb000)
	libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fdc81b91000)
	libz.so.1 => /apps/anaconda/anaconda3-5.3.0/envs/test-samtools/bin/../lib/libz.so.1 (0x00007fdc81979000)
	libm.so.6 => /lib64/libm.so.6 (0x00007fdc81677000)
[...]

PR #13903, opened 2019-03-04
Update to version 2.1.1

AlessioMilanese:

This update does two things:

1.  Update the recepe with a new version
2.  Try to solve the problem with samtools mentioned in https://github.com/bioconda/bioconda-recipes/issues/12100. In the samtools recipe ([link](https://github.com/bioconda/bioconda-recipes/blob/e53d407b6ce688974659b88bdc501f48c260cad8/recipes/samtools/meta.yaml)) it's required `openssl <1.1` for `host`, but not for `run`. As a result, during run `openssl 1.1` is used, and samtools is not working. To solve this we:
- add `openssl=1.0` to our recipe in the `run` section
- change the test section, where we run a script that test if samtools is properly installed.

### DIFF
--- a/recipes/motus/meta.yaml
+++ b/recipes/motus/meta.yaml
@@ -1,5 +1,5 @@
-{% set version = "2.1.0" %}
-{% set sha256 = "df07ced0cde59a2223d6caa7fd718838809e304ce778ad403d87a6312853d67c" %}
+{% set version = "2.1.1" %}
+{% set sha256 = "acefbfff1ee9b59e5668a3b95ada87ce0fecd2db8961f8d2c0e0ea79e6c21787" %}
 
 package:
   name: motus
@@ -10,7 +10,7 @@ source:
   sha256: '{{ sha256 }}'
 
 build:
-  number: 2
+  number: 1
   # motus requires python3
   skip: True # [not py3k]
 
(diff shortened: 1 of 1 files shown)


unode:
Hi @AlessioMilanese motus doesn't depend on openssl directly so you shouldn't need to pin an openssl version here.

samtools should be fixed instead. I've created #13909 to address that issue.

Once that's fixed you can remove the `openssl=1.0` line.
Depending on `samtools` alone should be sufficient to get a consistent environment.

unode:
@AlessioMilanese can you check if the openssl issue is fixed?
It should no longer try to install `openssl-1.1`.

unode:
Ok since the openssl issue is going to take a few days to fix and pinning openssl here seems to work I'm merging this.

Note that this might still break in the future once most packages migrate to openssl-1.1.
Breaking here means that due to version conflicts conda might decide that an older motus (that has a flexible openssl dependency) is a better solution than the latest motus with an incompatible `openssl` dependency.

AlessioMilanese:
Ok,
I'm not even sure that adding `openssl=1.0` to our recipe in the run section will solve the problem.
In any case, let's remove it once the samtools issue is solved.

Issue #232, opened 2015-11-09
How to handle multiple versions for a package?


acaprez:
I was just wondering what the procedure was for adding recipes for multiple versions of a particular package.  I didn't see any existing examples offhand in the repo nor do the docs mention it.  In my personal repo I just use a two-level `<package name>/<package version> ` directory structure, but I'm not sure how to handle it here.

johanneskoester:
This is an interesting question. So far, I thought it would be sufficient to keep only the newest version. I would be interested to hear the opinion of all devs here.

Can you provide a use case for having to store multiple versions?

acaprez:
The main use case I'm thinking of at the moment is QIIME.  It requires very specific (and sometime quite dated) versions of its [dependencies] (http://qiime.org/install/install.html#installing-qiime-natively-with-a-full-install).  So it would be nice to be able to keep those alongside the current version.  

Another use case might be for a package that takes advantage of the Features functionality.  For example, if you want to have a MKL and non-MKL version of the same package release, it might be useful to do something like `<package name>/<package version>-mkl`.

I do admit it's also somewhat of a personal preference, but for our environment we often have multiple package versions installed alongside each other on our HPC systems.  It's advantageous for me to keep each version's recipe as a distinct subdirectory.  That way if I need to rebuild and reinstall the older version for some reason it's simpler - I don't need to go back in the commit history to find the previous version.

johanneskoester:
Especially the second use case makes sense. The proposed structure would indeed give us some additional flexibility, and won't really hurt. We could convert our existing try with a simple script that parses the meta.yaml files.

johanneskoester:
@organization/bioconda: any objections, opinions?

kyleabeauchamp:
No comment from me, but I will point out two examples: https://github.com/conda/conda-recipes/tree/master/numpy-openblas and https://github.com/conda/conda-recipes/tree/master/numpy-mkl

johanneskoester:
So, here, openblas and mkl has become part of the name as opposed to the version/build string. That means that packages need to specifically depend on e.g. numpy-openblas?

kyleabeauchamp:
Not quite.  In the meta.yaml, the names are both numpy.  It's just the name of the recipe directory that changed.  Also, I don't think that repository is doing full continuous integration / building, so YMMV.

https://github.com/conda/conda-recipes/blob/master/numpy-openblas/meta.yaml#L2

[...]

PR #13915, opened 2019-03-05
Keep support for Unicycler 0.4.4

PertuyF:
:information_source:

----------------

Keep supporting Unicycler version 0.4.4 to prevent issue https://github.com/rrwick/Unicycler/issues/153

### DIFF
--- a/recipes/unicycler/0.4.4/conda_build_config.yaml
+++ b/recipes/unicycler/0.4.4/conda_build_config.yaml
@@ -1,0 +1,6 @@
+c_compiler:
+  - gcc      # [linux]
+  - clang    # [osx]
+cxx_compiler:
+  - gxx      # [linux]
+  - clangxx  # [osx]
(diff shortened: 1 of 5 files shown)


PertuyF:
Hi @bgruening, I have projects that require to stick with Unicycler `0.4.4`, but require rebuild for python 3.6 or 3.7. Is it still the way to do (from #232) to duplicate the recipe in a `<package name>/<package version>` directory?

recipes/unicycler/0.4.4/conda_build_config.yaml
```diff
@@ -0,0 +1,6 @@
+c_compiler:
```
bgruening: I don't think this is needed anymore.

PertuyF: You're right, removed that.

PR #13946, opened 2019-03-07
Fixing bioconductor datapackages: subdag 1 of 10

johanneskoester:
:information_source:

----------------

### DIFF
--- a/recipes/bioconductor-ag.db/meta.yaml
+++ b/recipes/bioconductor-ag.db/meta.yaml
@@ -12,7 +12,7 @@ source:
     - 'https://depot.galaxyproject.org/software/bioconductor-{{ name }}/bioconductor-{{ name }}_{{ version }}_src_all.tar.gz'
   md5: e5913da38fe4487202306cacd885840d
 build:
-  number: 0
+  number: 1
   rpaths:
     - lib/R/lib/
     - lib/
@@ -27,7 +27,7 @@ requirements:
     - 'bioconductor-annotationdbi >=1.44.0,<1.45.0'
     - 'bioconductor-org.at.tair.db >=3.7.0,<3.8.0'
     - r-base
-    - wget
+    - curl
 test:
   commands:
     - '$R -e "library(''{{ name }}'')"'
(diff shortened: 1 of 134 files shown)


bgruening:
@johanneskoester can you explain why we need to rebuild those recipes? It will create a lot of containers and will waste a lot of space down the road ...

Can it be that the openssl rebuild will fix the data packages?

johanneskoester:
Wget is coming from the defaults channel, so I don't think this will be fixed with the openssl migration. My rationale for migrating to curl was that it is in conda-forge, so it is easier to fix if something breaks in the future.  I thought that it is not a good idea to have about 1000 data packages dependent on a wget at runtime that we cannot quickly fix. And many R packages depend on these data packages.

johanneskoester:
But indeed, the containers are not affected by the current issue, as they already contain the downloaded datapackage, right?

johanneskoester:
Maybe we could find a way to temporarily disable the container upload for these?

johanneskoester:
A container-upload blacklist?

bgruening:
Mh, so the issue is wget/curl for the data packages. I'm all in for using curl for the next rebuild but now? Is there no way to get wget from conda-forge? Have you tried rebuilding wget on conda-forge?

johanneskoester:
No, haven't tried that. That would indeed be another reasonable solution.

johanneskoester:
Good idea :-)!

bgruening:
I just figured that there is no wget feedstock in conda-forge ... and I found that wget from anaconda is build against `openssl >=1.1.1,<1.1.2.0a0`. So this might be the underlying problem ... Maybe a conda-forge wget for old and then for new openssl will fix it?

bgruening:
Which version do you get when a data packages crashes?

johanneskoester:
Working on that now.

johanneskoester:
The latest wget causes the problem, earlier versions worked fine. openssl should be pinned, so I am not sure if that is the issue though.

dpryan79:
Can we change the pinnings post hoc after moving to the CDN?

bgruening:
I think we are working on fixing the defaults wget package or/and creating a conda-forge wget package.

johanneskoester:
Exactly. Thanks for pointing me to this much faster solution (hopefully at least ;-)).

johanneskoester:
At least this forced me to work on the subdag-commit solution, which might be handy in other cases.

PR #13962, opened 2019-03-08
Remove openssl=1.0 pinning in motus

unode:
This is unnecessary now that samtools was updated to build against openssl-1.1

:information_source:

----------------

### DIFF
--- a/recipes/motus/meta.yaml
+++ b/recipes/motus/meta.yaml
@@ -10,7 +10,7 @@ source:
   sha256: '{{ sha256 }}'
 
 build:
-  number: 1
+  number: 2
   # motus requires python3
   skip: True # [not py3k]
 
@@ -22,8 +22,6 @@ requirements:
     - samtools
     - bwa
     - python
-    # samtools requires openssl=1.0 to run
-    - openssl=1.0
 
 test:
   commands:
--- a/recipes/motus/post-link.sh
+++ b/recipes/motus/post-link.sh
@@ -1,12 +1,11 @@
 #!/usr/bin/env bash
 
-echo "Downloading mOTUs-v2 database"
+echo ">>> post-link: Downloading mOTUs-v2 database"
 
 # In build.sh we keep the tool files in a custom location
 pkgdir=$PREFIX/share/motus-${PKG_VERSION}/
 
+echo ">>> post-link: Running motus setup.py"
+
 # mOTUs-v2 requires a ~1GB download which is performed by the setup.py script
 python "$pkgdir/setup.py"
-
-# After install check that everything is available and working as expected
-python "$pkgdir/test.py"


unode:
I'm not sure what the build failure is but seems unrelated.
Force-pushing to trigger a rebuild and seeing if retrying solves it.

unode:
@bioconda/build-system I tried building this recipe locally and I get the [same error as online](https://circleci.com/gh/bioconda/bioconda-recipes/46926).
Something happens when executing [post-link.sh](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/motus/post-link.sh) that causes a failure, however the output of the post-link script isn't visible in the console so I can't tell what is causing it. I tried adding `--verbose` and `--logging=debug` but the output is still not shown and the extra info wasn't useful for debugging it.

Error logs also show `running your command again with -v will provide additional information` but I can't figure out how to pass `-v` to that step. Passing `--verbose` to either `circleci` or `bioconda-utils` didn't affect this step and the message mentioning `-v` is still shown.

The same error is seen in both mulled and regular build. The error is also consistent between both `circleci build` and `bioconda-utils build` execution modes.

Finally I also tried passing `--keep-image` and inspecting the image but all temporary folders and build scripts mentioned in the logs seem to have been removed so I don't really know where to look for additional info.

Is there any way to obtain more debugging information?
Does the output end up on some logfile inside the container?

unode:
Ok, by using some forced redirect in `post-link.sh` I managed to gain access to the output and confirm my suspicion:

```
 ------------------------------------------------------------------------------
|                               TEST MOTUS TOOL                                |
 ------------------------------------------------------------------------------

1-- ran setup.py: done

2-- Tools and versions:
- python:    correct
- bwa:       correct
- samtools:  correct
- metaSNV:   WARNING. metaSNV is not in the path


3-- Taxonomy profiling test:
- Run motus (-v 1, only error messages):
- end motus call

Check resulting profile: correct
```
`metaSNV` is not found which causes `test.py` to exit with non-zero returncode triggering the failure.
I don't fully understand why this fails. `metasnv` is correctly specified as a dependency.

Does `post-link.sh` run after all dependencies are installed or could it be running before `metasnv`'s installation is complete?

unode:
Adding `metasnv` to `host:` doesn't help. It fails all the same.

I propose to remove https://github.com/bioconda/bioconda-recipes/blob/37c77464ff0b86cdfbab2093814456e21e5ab653/recipes/motus/post-link.sh#L12 from the `post-link.sh` script. I see that there's now https://github.com/bioconda/bioconda-recipes/blob/37c77464ff0b86cdfbab2093814456e21e5ab653/recipes/motus/meta.yaml#L30 which seems to do the same in a more appropriate location.
I'm updating the pull request after the above to confirm that this fixes the issue.

@AlessioMilanese any objections or thoughts?

One thing to keep in mind is that in order to run the tests (`motus profile --test`) the `post-link.sh` must have finished successfully. This means downloading and unpacking ~1GB of data before testing.

@bgruening Would you be so kind to review this again and in particular report if the 1GB requirement for testing is going to affect the final package to upload to anaconda?
For sure the quay.io images are affected but this was already the case before.

AlessioMilanese:
@unode & @bgruening 
Can we merge?

unode:
Guess we can try-and-see. Worst case scenario we get a failure on upload. Hopefully not.

unode:
Merged but linting tests seem to be hanging since [2 days ago](https://github.com/bioconda/bioconda-recipes/commits/master) from which point the CI tests report also looks different. I guess someone is working on improving that.

I don't see the new motus release on [anaconda](https://anaconda.org/bioconda/motus/files) though so we'll have to wait for the other issues to be fixed. Troubled times... :rescue_worker_helmet:

PR #13977, opened 2019-03-10
March 2019 Release

PauAndrio:
:information_source:

----------------

### DIFF
--- a/recipes/biobb_wf_mutations/build.sh
+++ b/recipes/biobb_wf_mutations/build.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+
+python3 setup.py install --single-version-externally-managed --record=record.txt
+
+mkdir -p ~/biobb_notebooks
+cp $SP_DIR/biobb_wf_mutations/notebooks/mutations.ipynb ~/biobb_notebooks/
--- a/recipes/biobb_wf_mutations/meta.yaml
+++ b/recipes/biobb_wf_mutations/meta.yaml
@@ -1,0 +1,59 @@
+{% set name = "biobb_wf_mutations" %}
+{% set version = "0.0.5" %}
+{% set file_ext = "tar.gz" %}
+{% set hash_type = "sha256" %}
+{% set hash_value = "831de3942bc240b40f893faa81faf647fdb18b94ce613e20a49312080de5c5bc" %}
+
+package:
+  name: '{{ name|lower }}'
+  version: '{{ version }}'
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.{{ file_ext }}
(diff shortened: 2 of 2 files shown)


recipes/biobb_wf_mutations/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+python3 setup.py install --single-version-externally-managed --record=record.txt
+
+mkdir -p ~/biobb_notebooks
+cp $SP_DIR/biobb_wf_mutations/notebooks/mutations.ipynb ~/biobb_notebooks/
```
npavlovikj: These 2 lines are redundant - "mutations.ipynb" is already in "$PREFIX/lib/python*/site-packages/biobb_wf_mutations/notebooks/mutations.ipynb" after `python setup.py install`.

On the other hand, if you want to have `biobb_notebooks` directory as part of the recipe, you should use `mkdir -p ${PREFIX}/share/${PKG_NAME}-${PKG_VERSION}/biobb_notebooks` instead - the command you have is not creating `biobb_notebooks` within the environment.

PauAndrio: Hello Natasha,

Thank you for the comment. I am (unsuccessfully) trying to create the directory `biobb_notebooks` inside the home directory of the user that executes `conda install biobb_wf_mutations` and copy there the `mutations.ipynb` file. 

I imagine copying files outside the environment is not a good practice (for security reasons), but is it possible?

If it is not, how can I make available non executable files like `mutations.ipynb` to the user?

npavlovikj: @PauAndrio , creating a folder in home directory is not a good idea. Coming from an HPC center where we install our software system-wide in a specific root location, this recipe will not work for us as you have intended. The conda practice is to have everything within the environment.

What you can do with your recipe, is to copy the notebooks in `${PREFIX}/share/${PKG_NAME}-${PKG_VERSION}/biobb_notebooks` as I earlier suggested, and then create a `post-link.sh` script where you give instructions to the person that installs the package for manual setup they need to do (which will be `cp -r ${PREFIX}/share/${PKG_NAME}-${PKG_VERSION}/biobb_notebooks ~/biobb_notebooks` in your case). To read more about `post-link.sh`, please check https://bioconda.github.io/guidelines.html. This way, everything is within the environment, and depending on the user, they can do manual setup and copy the notebooks to their home directory if needed.
Non-executable flies, such as configuration files and databases, usually go in `${PREFIX}/share/${PKG_NAME}-${PKG_VERSION}/` (you can check the Bioconda guidelines for more details on this).

PauAndrio: Thank you @npavlovikj I have created the new PR with the changes that you proposed.

PR #13988, opened 2019-03-11
Crispr studio

KyleL1998:
:information_source:

----------------

Recipe for CRISPR Studio


KyleL1998:
Miss read README**

KyleL1998:
@bgruening I fixed the conflict errors that occurred after you made some changes to my CRISPR Studio the recipe, and was hoping you could merge it in or give me some feedback on what else you think needs to be changed. As I have the wrapper already written using a local copy so would like to be able to make a pull request for that after the recipe is done.

recipes/CRISPRStudio/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+ - url: http://faculty.virginia.edu/wrpearson/fasta/CURRENT/fasta-36.3.8g-linux64.tar.gz
+   sha256: 8aafc7cfff7286282439bafba327aa432ae4bba1917f2c96b0dc3e4433af4699
+   folder: fasta-36.3.8g
+ 
+build:
+ number: 0
+
+requirements:
+ build:
+  - {{ compiler('c') }}
```
dpryan79: You're not compiling anything, you can omit this

recipes/CRISPRStudio/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+   folder: fasta-36.3.8g
+ 
+build:
+ number: 0
+
+requirements:
+ build:
+  - {{ compiler('c') }} 
+ host:
+  - pip
```
dpryan79: You're not using `pip`, you can omit this. Actually, you can omit the entire `host:` section.

recipes/CRISPRStudio/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  - python >=3.6,<3.7
+ run:
+  - numpy <=1.16.2
+  - pandas >=0.24.1
+  - scipy <=1.2.1
+  - scikit-bio >=0.4.2
+  - python >=3.6,<3.7
```
dpryan79: Do you really need to pin to python 3.6 and prevent any future version of numpy?

recipes/CRISPRStudio/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+mkdir -p $PREFIX/bin/tools/
+cp CRISPR_Studio_1.0.py $PREFIX/bin
+cp -r fasta-36.3.8g $PREFIX/bin/tools
+ln -s  $PREFIX/bin/CRISPR_Studio_1.0.py $PREFIX/bin/CRISPR_Studio
```
dpryan79: Why not just rename the script?

recipes/CRISPRStudio/meta.yaml
KyleL1998: Are you sure? As I'm downloading fasta36 and if I remember correctly I added this line because the files in fasta36 weren't getting linked

dpryan79: That's actually another issue, that fasta36 is already a package, so just list it as a run time dependency.

KyleL1998: the CRISPR studio script has the path to look for fasta36 hardcoded to a certain location. Which is why I did it this way.

dpryan79: Can you patch that?

recipes/CRISPRStudio/meta.yaml
KyleL1998: Yes. There is incompatibility issues with scikit-bio and more recent versions of python3.

recipes/CRISPRStudio/build.sh
KyleL1998: I thought it needed to be soft linked I guess not. Changed.

recipes/CRISPRStudio/meta.yaml
KyleL1998: Made a change that seems to work. It just introduced a `ClobberWarning` that I can't seem to get rid of.

dpryan79:
@KyleL1998 Let me know if you're not part of the bioconda group, I'll add you if not.

KyleL1998:
Don't believe I'm part of it

dpryan79:
@KyleL1998 I've added you now, just accept the invite and you should be able to merge this then.

PR #14003, opened 2019-03-12
stacks2.3d

bernt-matthias:
:information_source:

----------------


recipes/stacks/2.3d/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+  run:
+    - perl
+    - perl-bioperl-core
+    - perl-file-spec
+    - perl-file-temp
+    - perl-posix
+    - python >=3
+    - velvet
+    - samtools
+    - zlib
```
bgruening: FYI: zlib does not need to be added anymore, it has a run_dependency set.

PR #14009, opened 2019-03-12
Shapeshifter

frytime32:
:information_source:

----------------


recipes/shapeshifter/meta.yaml
```diff
@@ -1,28 +1,42 @@
 {% set name = "shapeshifter" %}
-{% set version = "0.0.3" %}
+{% set version = "1.1.1" %}
+{% set file_ext = "tar.gz" %}
+{% set hash_type = "sha256" %}
```
epruesse: You are overdoing it here. The Jinja2 variables don't actually help with machine processing the recipe. I'd suggest dropping all but the version number. The name won't change, nor will file_ext or hash_type. The hash itself only occurs in one location in the recipe, so having a variable won't improve readability.

recipes/shapeshifter/meta.yaml
```diff
@@ -34,9 +48,11 @@ test:
 about:
   home: https://github.com/srp33/ShapeShifter
   license: MIT
-  summary: A tool for managing and transforming large data sets
+  license_family: MIT
+  summary: A tool for managing large datasets
 
 extra:
-  recipe-maintainers:
+  recipe-maintainers:
```
epruesse: needles space at end of line

recipes/shapeshifter/meta.yaml
```diff
@@ -1,28 +1,42 @@
...
 
 requirements:
   host:
-    - pip
     - python >=3
-  run:
+    - setuptools
     - pandas
     - pyarrow
+    - numpy ==1.15.4
```
epruesse: I think that should be `>=1.15.4` so that it will allow newer versions as well. If the package does really special stuff, `>=1.15.4,<1.16a0`  would make it allow only that range.

epruesse: Also, if the package can work with older numpy, please remove this. The current default pinngs are for numpy 1.9 - that's what all binary packages are built against. See link below.

https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/9b8e67fcac4a4199dc7af7f120564aceab1cc7b8/recipe/conda_build_config.yaml#L516-L517

recipes/shapeshifter/meta.yaml
frytime32: These were automatically generated using the "conda skeleton pypi" command as suggested in the "Guidelines for bioconda recipes" page. This is how the file was created.

recipes/shapeshifter/meta.yaml
frytime32: Previously, numpy was not working correctly with hdf files in versions other than 1.15, but has been patched to fix this.

Version has now been changed to any

recipes/shapeshifter/meta.yaml
```diff
@@ -1,28 +1,42 @@
...
+    - numpy 
+    - sqlalchemy
+    - xlsxwriter
+    - xlrd
+    - nbformat
+  run:
     - python
+    - pandas
+    - pyarrow
+    - numpy ==1.15.4
```
epruesse: Forgot this one

frytime32: Whoops, must have forgotten to save the changes. Thanks, it's fixed now.

epruesse:
- removed the empty line you added at the bottom
- pinned runtime to py>3 as well
- removed some of the jinja (extra indirection makes it harder for human and machine to parse and edit)

epruesse:
I'll merge this when it's done.

PR #14022, opened 2019-03-13
pbwt support

tiagoantao:
This supports pbwt
----------------


recipes/pbwt/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+  skip: true # [osx]
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - htslib
+    - zlib
+  run:
+   - htslib
```
bgruening: I think those can be removed now, they export themself with run-exports nowadays.

recipes/pbwt/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+   - zlib
+
+test:
+  commands:
+    - pbwt > /dev/null
```
bgruening: /dev/null should not be needed

recipes/pbwt/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+
+about:
+  home: https://github.com/richarddurbin/pbwt
+  license: GPL
```
bgruening: can you please add license_file and license_family ... if it is GPL we need to ship a license_file ... strictly ... or just ignore me ;)

tiagoantao: I actually had the license wrong :(

bgruening:
Merci.

PR #14023, opened 2019-03-13
Fix version support for Python2.7

prihoda:

Fix for #14007

@dpryan79 Can you please take a look at this fix for the DeepBGC module? It turns out the original python version specification was correct, specifying `python >=2.7, >3` will only accept version `>3`. 

Is it possible to fix this for version 0.1.3 or do we have to do a new version? 

Thanks!

### DIFF
--- a/recipes/deepbgc/meta.yaml
+++ b/recipes/deepbgc/meta.yaml
@@ -19,7 +19,7 @@ build:
 
 requirements:
   host:
-    - python >=3
+    - python>=2.7, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*
     - setuptools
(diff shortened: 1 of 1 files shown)


dpryan79:
It was `>=3`, I assume you just wanted python 3. We don't build for 3.1 through 3.5, so there's no need to specify that.

dpryan79:
BTW, if you need to allow python2 as well, just don't specify any versions and increment the build number by one.

prihoda:
AFAIK Before I specified "python>=2.7,>=3". Could it have been automatically changed to ">=3"? Since ">=2.7,>=3" apparently applies both conditions which just translates to ">=3".

dpryan79:
That would have been my guess, yes.

PR #14025, opened 2019-03-13
Bump blast to 2.8.1; switch recipe to use precompiled binaries

winni2k:
:information_source:

----------------


winni2k:
Hmm, this recipe works on my OSX computer. Does anyone have an idea why I am getting linking errors for libraries that are in the `host` section? For example:

```
[Mar 13 20:35:48] SERR blastn: error while loading shared libraries: libbz2.so.1: cannot open shared object file: No such file or directory
```

epruesse:
Why do you want to switch back to the pre-compiled binaries?

I spent a lot of time getting blast to work from source so that the binaries would be dynamically linked against things we already have and the package not be hundreds of megabytes large.

epruesse:
@biocondabot lint

winni2k:
I did not mean to offend anybody by switching back to binaries.  I can see that the conda recipe for blast is complex. I appreciate all the work that @epruesse has done on this so far.  However, here are a few points:

1. Version 2.8.1 of blast+ fixes a high-profile bug (see the [release notes](https://www.ncbi.nlm.nih.gov/books/NBK131777/) and the [original bug](https://www.ncbi.nlm.nih.gov/pubmed/30247621))
2. Version 2.8.1 was released three months ago
3. Despite the high profile nature of this bug, the bioconda version of blast+ remains at 2.7.1
4. Getting an old C tool to compile from source can be a lot of work, and as @epruesse states, [for blast+ this was in fact the case](https://github.com/bioconda/bioconda-recipes/pull/14025#issuecomment-473579234).
5. I cannot get blast+ 2.8.1 to compile from source using its bioconda recipe.

I conclude that if compiling blast+ 2.8.1 for conda was easy for the average bioinformatician, then it would have been done already. 

I really do need to use blast+ 2.8.1 right now, but I do not have the time right now to spend what will likely amount to days on getting blast+ to compile, especially if there are pre-compiled binaries available. 

However, I do understand if bioconda is not the right place for these binaries. Luckily, anaconda channels are cheap, and I will see if I can find another channel for the blast binaries.

epruesse:
@winni2k I apologize for putting you off with my comment.

Let's have a look at building 2.8.1 from source together.

winni2k:
@epruesse fair enough. I would be happy to help.

PR #14030, opened 2019-03-14
Add recipe for halla

npavlovikj:
:information_source:

----------------


npavlovikj:
@bioconda/core , I would like your feedback on this recipe if possible. 
I have hard time understanding why the OSX build is sometimes successful, and sometimes it is not. As you can see, the second commit built ok, I tried few other things, and reverted to the successful one, which then failed... I was encountering the same for the Linux builds in my previous PR, but that is working ok now ¯\_(ツ)_/¯ To me, it looks like the "_2to3_" changes are something accepted, sometimes not, and I don't know what causes that. 
If you have any idea, or notice something in my recipe is not compatible, I would appreciate you letting me know.

dpryan79:
It looks like `2to3` isn't running all of the time for some reason. Do you mind if I play around with this PR to try and diagnose things?

dpryan79:
At the moment this is only building for python 3.6 (presumably due to a dependency being absent for python 2), is that what you really want?

npavlovikj:
@dpryan79 , many thanks for looking into this! 
Do you know what dependencies for Python 2.7 are missing (if that is the reason for not being built with Python 2.7)? I checked them all, and installed them locally in a Python 2.7 environment, and they all were available.
Since this is Python 2/3 package I would like to try to have both builds available, unless you recommend otherwise.

dpryan79:
I'm not really sure, I'll play around with the recipe a bit and see if I can figure out what's going on with python 2.7. Is there a reason you're not building for python 3.7?

npavlovikj:
Thanks @dpryan79 , please go ahead and play with the recipe. For Pyhon 3.7 some of the dependencies were missing. In the meantime, I can check and see is it feasible to have those dependencies for Python 3.7 available.

dpryan79:
I'm already bumping minepy :) https://github.com/bioconda/bioconda-recipes/pull/14043

dpryan79:
The `2to3` changes turn out to not be needed, it's already python3 (though it's pretty hacky how they did it). Now that minepy is built for python 3.7 we should be able to make this `noarch: python`.

dpryan79:
I spoke too soon, there were a couple things I missed and `2to3` is still needed. I don't want to patch the source, so I'll revert to the previous commit that was working on all python builds.

npavlovikj:
Thanks for all the updates @dpryan79  - you are few time zones ahead of me ;)
Fingers crossed the build passes 🤞

npavlovikj:
Many thanks for all the help @dpryan79 ! 
So, allowing Python 3.7 fixed the issue? And what was the reason for it?

dpryan79:
I don't know why it wasn't building for python 2.7 before, but it is now so I'm happy :)

npavlovikj:
That makes two of us :)

PR #14044, opened 2019-03-14
Berokka -- Copy db under PREFIX/db

Kurt-Hanel:
:information_source:

----------------

In order to run, berokka requires a db with controls.fna and dnaA.fna. This pull request will make sure that that database is set when berokka is installed.


bgruening:
@Kurt-Hanel how big are they? Usually, we use post-link script to download DB's after installation ...

bgruening:
For example here: https://github.com/bioconda/bioconda-recipes/blob/ccbaa13b33826851ab5012b6c72b7388eb6e4090/recipes/bioconductor-lrbase.rno.eg.db/post-link.sh

Kurt-Hanel:
controls.fna is 2 KB and dnaA.fna is 5 MB

bgruening:
@Kurt-Hanel I think we should us this post-link and post-unlink tricks here. 5MB seems to be too large imho.

Kurt-Hanel:
@bgruening Since the db is already included in berokka's [latest release](https://github.com/tseemann/berokka/releases), is the post-link script necessary? Wouldn't using a post-link script mean downloading the tar release twice?

Kurt-Hanel:
Hi @bgruening, please let me know what your thoughts are on my last question.

bgruening:
The post-link script only makes sense when the DB is not copied into the PREFIX.
You need to increase the build number to get a new package.

Kurt-Hanel:
Thanks @bgruening !

PR #14066, opened 2019-03-17
Update blast to 2.8.1

epruesse:
Update `blast`: **2.7.1** &rarr; **2.8.1**

***

[Package Info](https://bioconda.github.io/recipes/blast/README.html) | [Recipe Folder](https://github.com//bioconda/bioconda-recipes/tree/bump/blast/recipesblast) | [Upstream Home](http://blast.ncbi.nlm.nih.gov/Blast.cgi?PAGE_TYPE=BlastDocs) | [Upstream Releases](https://ftp.ncbi.nlm.nih.gov/blast/executables/blast+/)

***


epruesse:
ping @winni2k let's see how this goes

winni2k:
Okidoki.

epruesse:
If it doesn't I'd consider adding another package that just wraps the upstream packages. Something like `blast-static` perhaps. That way we can have both the smaller package we build ourselves and the quick-to-update package that just re-packages the upstream binaries.

The reason I spent so much time building blast from source was that the blast packages are huge, which makes them quite annoying when you e.g. use them in a CI environment.

winni2k:
I think the idea of a `blast-static` package might be a good one if the compilation route does not succeed.

epruesse:
Ok, now we're down to linking
```
warning: libgnutls.so.30, needed by .../c++/ReleaseMT/lib/libxconnect.so, not found (try using -rpath or -rpath-link)

x86_64-conda_cos6-linux-gnu/bin/ld: work/c++/ReleaseMT/lib/libxconnect.so: undefined reference to `gnutls_transport_get_ptr@GNUTLS_3_4'
```

winni2k:
This is an odd error message:
```
conda.exceptions.CondaValueError: Malformed version string '3.3|3.3.*': invalid character(s).
```
That meta.yaml file line has not been touched 😕

winni2k:
Hmm, it looks like there is no conda version of gnutls 3.3 (https://anaconda.org/conda-forge/gnutls/files). I am surprised that this version of blast would need a downgrade from gnutls version >3.5 to 3.3?

epruesse:
I don't think it was a problem with the wrong gnutls version installed. We didn't get to the point where it would be testing. Something with the build system I think.

epruesse:
ping @klymenko - Would any of you guys be able to help here?

klymenko:
> ping @klymenko - Would any of you guys be able to help here?

How can I help you?

winni2k:
Hey @klymenko, we are trying to bump the version of blast to 2.8.1, but we need some help getting the linking to work. Would that be something you could help us with?  

I have tried a few things, but I'm not sure they were helpful. The commit you should probably start from/look at is not the most recent, but instead https://github.com/bioconda/bioconda-recipes/pull/14066/commits/c0babc9dcd3eef2e9a048ee6d2014ea8ccff2ee5.

klymenko:
> Hey @klymenko, we are trying to bump the version of blast to 2.8.1, but we need some help getting the linking to work. Would that be something you could help us with?
But why are you contacting me?

epruesse:
Hi @klymenko,  I pinged you as the currently most active member of github.com/ncbi - hope that was ok.

Mainly, I wanted to make sure you guys are aware of our build and ok with the build options we chose. We got some 60k downloads for BLAST 2.7.1 in the past year, so mistakes we make trickle down into a lot of science. The BLAST build system is quite complex and unusual, so I'm not 100% sure all I do is "correct". 

The point of doing a build of our own, rather than repackaging the NCBI binaries, is that within the context of conda we can use DSOs so we can keep the package size a lot smaller than the official statically linked binaries.

One "patch" was already needed - something changed in the build system between 2.7.1 and 2.8.1 that made it use `ar` without arguments on Linux. I.e. `$AR somelib.a obj1.o obj2.o` instead of `$AR rv somelib.a obj1.o obj2.o`. (I just stuck the arguments into $AR, but it seems to be an error in the build system - not sure why it just pops up for us, but then there are so many options it's impossible to test all).

W.r.t. the current hold-up - somehow when linking the unit tests (`gapinfo_unit_test`), `ld` can't find `gnutls` required by `xconnect.so`. We'll figure that one out though. It's there and it was used to build `xconnect.so` in the first place and the compiler args seem to be pointing at the right place.

klymenko:
Here is the response:
> I looks like they’re having problems with the GnuTLS library, but BLAST+ moved away from this library over a year ago. We now use mbedTLS (courtesy of the SRA toolkit IIRC). As for build issues in the core C++ toolkit, I’d refer you to cpp-core@ncbi.nlm.nih.gov.

epruesse:
mbedTLS is not yet available to us. See https://github.com/conda-forge/staged-recipes/issues/3949

epruesse:
closing in favor of #14620

PR #14096, opened 2019-03-19
Update selene-sdk to 0.3.0

kathyxchen:
:information_source:

----------------


epruesse:
@bioconda-bot lint

epruesse:
@biocondabot lint

kathyxchen:
The build continually fails on `libtorch.so.1: undefined symbol: nvrtcGetProgramLogSize`. I haven't been able to determine how I should fix this.

npavlovikj:
I believe the reason for your error are the cuda libraries. 
The "pytorch" version used in your recipe is built with "cuda92py37h65efead_0" (pytorch 1.0.1 cuda92py37h65efead_0). There is no "cuda" in the build environment, and therefore the error you get. 
I assume you have cuda on you computer, that's why "conda build ." worked for you locally. I don't have cuda on my laptop, and even with "conda build .", I got the same error as here. 
The lower "pytorch" versions are not built with cuda, so if you have `pytorch>=0.4.1,<1` in your recipe, I believe the build should pass.

kathyxchen:
Thank you @npavlovikj for looking into my issue! I really appreciate it. That makes sense. It is a bit unfortunate, because I'd like users to have the option of using anything >= 0.4.1 in PyTorch (since the package does work across diff versions of PyTorch when built locally), but I guess that won't be an option for the conda-installable package.

PR #14103, opened 2019-03-20
Booster

Matt-BF:
:information_source:

----------------

### DIFF
--- a/recipes/booster/build.sh
+++ b/recipes/booster/build.sh
@@ -1,5 +1,24 @@
 #!/bin/bash
 
+CXXFLAGS="$CXXFLAGS";
+LDFLAGS="$LDFLAGS -Wl,-rpath ${PREFIX}/lib";
+CXX=g++;
+CC=gcc;
+
+if [ `uname` == Darwin ] ; then
+    CXXFLAGS="$CXXFLAGS -stdlib=libc++"
+    LDFLAGS="$LDFLAGS -stdlib=libc++"
+    CXX=clang++;
+    CC=clang;
+else ## linux
+    CXXFLAGS="$CXXFLAGS -fopenmp"
+fi
+
+export CC=${CC}
+export CXX=${CXX}
+export CXXFLAGS=${CXXFLAGS}
+export LDFLAGS=${LDFLAGS}
+            
 cd $SRC_DIR/src
 make
 
--- a/recipes/booster/build.sh
+++ b/recipes/booster/build.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+cd $SRC_DIR/src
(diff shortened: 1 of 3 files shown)


Matt-BF:
@bioconda/core @bioconda/osx I would like some help with building this recipe for osx. I’ve tried a few workarounds with clang but so far haven’t gained much ground. Any suggestions that may solve this problem? Thanks in advance!

recipes/booster/build.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+CXXFLAGS="$CXXFLAGS";
+CFLAGS="$CFLAGS";
+LDFLAGS="$LDFLAGS -Wl,-rpath ${PREFIX}/lib";
+CXX=g++;
+CC=gcc;
```
dpryan79: Please don't change CXX or CC.

recipes/booster/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  url: https://github.com/evolbioinfo/booster/archive/v{{ version }}.tar.gz
+  sha256: eee92b91c98d3fccb2b94201dfdc775d1c381d22c5953240562b8de20e763e18
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - make
+    - llvmdev  # [osx]
+    - libcxx 4.0.*  # [osx]
```
dpryan79: try adding `llvm-openmp` to the `build:` and `run:` sections. You can then remove `openmp` (actually, you should be able to remove the entirety of what's now in your `host:` and `run:` sections).

recipes/booster/build.sh
Matt-BF: > If you can get `make` to respect the compilers specified by `CC` and `CXX` then the builds should work.

Could you elaborate more on that? I've no experience with C programs. Also, I added your suggestions to the `meta.yaml` file and rolled back the `build.sh` file to a simple `make` call.

dpryan79: It's mostly a matter of looking into the Makefile and seeing how it's defining which compiler to use. In many cases, it's defining a variable like `${CC}`, which makes life easy. In other cases the compiler may be hard-coded (e.g., `gcc -o foo foo.c`), which means you'll have to heavily modify the Makefile so it uses the correct compilers. I'll try to take a look at the make file used by this project and see if I can provide any pointers.

dpryan79:
Looks like the changes I made should suffice.

Matt-BF:
@dpryan79 wow! Thank you very much, I'm gonna read carefully what you did and see if I might learn something for next time!

PR #14117, opened 2019-03-20
Add recipe for unitig-counter

johnlees:
:information_source:

----------------

### DIFF
--- a/recipes/unitig-counter/build.sh
+++ b/recipes/unitig-counter/build.sh
@@ -1,74 +1,7 @@
 #!/bin/bash
-
-# set up directory for building boost compiled dependencies
-# see
-# https://github.com/conda-forge/boost-cpp-feedstock/blob/master/recipe/build.sh
-# https://github.com/boostorg/program_options/blob/develop/.travis.yml
-git clone --depth 1 --single-branch --branch boost-1.69.0 https://github.com/boostorg/boost.git boost
-pushd boost
-rmdir libs/program_options libs/filesystem libs/regex libs/system
-git clone --depth 50 https://github.com/boostorg/program_options.git libs/program_options
-git clone --depth 50 https://github.com/boostorg/filesystem.git libs/filesystem
-git clone --depth 50 https://github.com/boostorg/system.git libs/system
-git clone --depth 50 https://github.com/boostorg/regex.git libs/regex
-git submodule update -q --init libs/algorithm
-git submodule update -q --init libs/graph
-git submodule update -q --init tools/boostdep
-git submodule update -q --init tools/build
-git submodule update -q --init tools/inspect
-python2 tools/boostdep/depinst/depinst.py program_options --include example
-python2 tools/boostdep/depinst/depinst.py filesystem --include example
-python2 tools/boostdep/depinst/depinst.py system --include example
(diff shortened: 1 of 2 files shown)


johnlees:
Could I have a review please, @bioconda/core 

[This adds a program to count unitigs in bacterial populations, based off the DBGWAS software.]

I had to turn off OS X as the Clang is v4.0, and >4.5 is required to compile. I think it would work with GCC on OS X though. Is that possible, or is skipping OS X the best thing to do in this case?

recipes/unitig-counter/build.sh
```diff
@@ -0,0 +1,75 @@
+#!/bin/bash
+
+# set up directory for building boost compiled dependencies
+# see
+# https://github.com/conda-forge/boost-cpp-feedstock/blob/master/recipe/build.sh
+# https://github.com/boostorg/program_options/blob/develop/.travis.yml
+git clone --depth 1 --single-branch --branch boost-1.69.0 https://github.com/boostorg/boost.git boost
```
dpryan79: Can't you use boost from conda-forge?

recipes/unitig-counter/build.sh
```diff
@@ -0,0 +1,75 @@
...
+    cxxflags="${CXXFLAGS}" \
+    linkflags="${LINKFLAGS}" \
+    --layout=system \
+    -j"${CPU_COUNT}" \
+    install
+popd
+
+# get submodules
+rmdir gatb-core
+git clone https://github.com/GATB/gatb-core.git
```
dpryan79: It's often easier to use a `git_url` and version so submodules are handled for you. You'll have to add a skip_lint for that, but it's easier.

recipes/unitig-counter/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+build:
+  number: 0
+  skip: True #  [osx]
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - python <3
```
dpryan79: Please put python in the host section.

recipes/unitig-counter/build.sh
johnlees: Yes, this does work with this one! Changed to do that

recipes/unitig-counter/meta.yaml
johnlees: Not needed, now boost is being pulled from conda-forge

recipes/unitig-counter/build.sh
johnlees: Done

johnlees:
@dpryan79 This build timed out, and so didn't give any diagnostics in the log. This didn't happen for me when I ran the new commit build locally. Any suggestions? Thanks!

dpryan79:
@johnlees Did you run it locally in docker?

johnlees:
No, just with conda build

johnlees:
Also odd because very little changed between this and the previously working commit.

dpryan79:
Try testing with `bioconda-utils build --docker` instead, it's what the real build is using.

johnlees:
Problem was ssh git submodules, leading to hanging waiting for a password. Changed to https now

PR #14118, opened 2019-03-20
Irida uploader

Kurt-Hanel:
:information_source:

----------------


Kurt-Hanel:
The linter says to specify the package as `should_be_noarch`, but since it has a command line interface, I'm not sure if that's alright.

recipes/irida-uploader/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - upload_run.patch
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - python =3.6
+    - pip
+    - make
```
bgruening: why is make and pip needed, you just copying files isn't it?

recipes/irida-uploader/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  sha256: 8080b411c0fcbc40d7792c65f1c9736186b0f5f436dc7115e0979cb80de1ea3e
+  patches: 
+    - upload_run.patch
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - python =3.6
```
dpryan79: Please rename `build:` to `host:`. Also, do you really need to pin to python 3.6 exclusively?

recipes/irida-uploader/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  url: https://github.com/phac-nml/{{ name }}/archive/{{ version }}.tar.gz
+  sha256: 8080b411c0fcbc40d7792c65f1c9736186b0f5f436dc7115e0979cb80de1ea3e
+  patches: 
+    - upload_run.patch
+
+build:
+  number: 0
+
+requirements:
+  build:
```
bgruening: this should be `host` and I think nothing in a host section is needed as you only copying files

recipes/irida-uploader/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  version: '{{ version }}'
+
+source:
+  url: https://github.com/phac-nml/{{ name }}/archive/{{ version }}.tar.gz
+  sha256: 8080b411c0fcbc40d7792c65f1c9736186b0f5f436dc7115e0979cb80de1ea3e
+  patches: 
+    - upload_run.patch
+
+build:
+  number: 0
```
bgruening: noarch: python is correct

recipes/irida-uploader/meta.yaml
dpryan79: `make` is the only dependency that should be in the `build:` section. This also suggests that you need some compilers (e.g., `{{ compiler("c") }}`)

recipes/irida-uploader/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - requests==2.21.0
+    - appdirs==1.4.3
+    - pytest
+    - mkdocs==1.0.0
+    - cerberus
+    - argparse
+    - selenium 
+
+  run:
+    - python =3.6
```
bgruening: is it really just working for python 3.6?

recipes/irida-uploader/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - rauth==0.7.3
```
bgruening: we usually do not pin the requirements that strict, can you relax this or remove it if not needed?

recipes/irida-uploader/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - setuptools
+    - rauth==0.7.3
+    - autopep8==1.4
+    - pycodestyle 
+    - requests==2.21.0
+    - appdirs==1.4.3
+    - pytest
+    - mkdocs==1.0.0
+    - cerberus
+    - argparse
```
dpryan79: Argparse has been built in since python 2.7. pytest is probably only needed for testing (likewise with things like pycodestyle and autopep8).

recipes/irida-uploader/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - selenium 
+
+  run:
+    - python =3.6
+    - rauth==0.7.3
+    - autopep8==1.4
+    - pycodestyle
```
dpryan79: Do you actually need all of this for normal running (in particular, pycodestyle, pytest, and autopep8)? I suspect not.

recipes/irida-uploader/meta.yaml
dpryan79: This looks like a pure python package, so please add `noarch: generic`.

recipes/irida-uploader/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+mkdir -p $PREFIX/bin/irida_uploader
+cp -a * ${PREFIX}/bin/irida_uploader
+ln -s $PREFIX/bin/irida_uploader/upload_run.py $PREFIX/bin/irida-uploader
```
dpryan79: Why not just name it that and omit the symlink?

recipes/irida-uploader/meta.yaml
Kurt-Hanel: The current release is just 3.6. But, for the next release, python 3.5, 3.6 and 3.7 will all be valid.

recipes/irida-uploader/meta.yaml
Kurt-Hanel: The build passes without the build/host section so I removed it. And it is just python 3.6 for this release

recipes/irida-uploader/meta.yaml
Kurt-Hanel: Stating noarch: python caused the build to fail for me now, so I switched it to noarch:generic as per Devon's suggestion.

recipes/irida-uploader/meta.yaml
Kurt-Hanel: I've removed make for now. Future versions of the recipe may require it, though.

Kurt-Hanel:
Thank you @bgruening and @dpryan79 for the reviews!

dpryan79:
It's annoying that they didn't structure their package as a normal python package.

bgruening:
@Kurt-Hanel was the last commit necessary. Just keep in mind this will make it very hard to mix with other tools in one environment, as you for the entire environment to be that minor python version.

Kurt-Hanel:
@bgruening that's a good point. I added the .7 to get it to pass CI testing, but using .* should work just fine. Unfortunately the current release is still limited to python 3,6

bgruening:
Thanks!

PR #14134, opened 2019-03-21
Add idba_subasm, since the current PR is behaving oddly.

dpryan79:
@abishara Maybe a fresh PR will solve things.

:information_source:

----------------

### DIFF
--- a/recipes/idba_subasm/build.sh
+++ b/recipes/idba_subasm/build.sh
@@ -8,8 +8,6 @@ aclocal
 autoconf
 automake --add-missing
 
-export CC=${GCC}
-export CXX=${GXX}
 export INCLUDE_PATH="${PREFIX}/include"
 export LIBRARY_PATH="${PREFIX}/lib"
 export DYLD_FALLBACK_LIBRARY_PATH="${PREFIX}/lib"
--- a/recipes/idba_subasm/build.sh
+++ b/recipes/idba_subasm/build.sh
@@ -1,0 +1,24 @@
+# fix automake
+
+# fix other python and perl scripts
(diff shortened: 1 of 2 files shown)


abishara:
Hmm, it looks like test-linux actually passed, but test-macos fails.   Not quite sure what should be revised from looking at the log.

dpryan79:
There appear to be a couple issues, I'll see if I can fix them real quick.

dpryan79:
It got much further this time, but I think it's generally not compatible with libc on OSX. Unless you really need that I suggest just skipping the OSX build.

abishara:
I'm totally fine with that, how do I specify that in `meta.yaml`?  I tried referencing the existing recipe for idba and I didn't see such a directive specified.

dpryan79:
See my last commit. Feel free to merge this once the CI testing is green.

abishara:
@dpryan79 Thanks so much for all the help getting this to work.  I don't think I would have been able to figure this out without your assistance.  

I have one more python project, of which `idba_subasm` is a dependency, but I think that one should be straightforward to figure out.

abishara:
@dpryan79 Actually bioconda-nightly-upload-linux and bioconda-nightly-upload-macos both failed and as a result this was not accepted.  Any suggestions for how to figure out what is wrong?  From the log it doesn't appear the failures are at least directly related to `idba_subasm`.

dpryan79:
That's unfortunate, I'll take a look.

dpryan79:
It was uploaded correctly, so you can install it. You can ignore the "nightly" jobs, they run once a night and try to find packages that didn't get uploaded. They use whatever commit happens to be the most recent, which was yours in this case. Since some of our packages (a few perl-related ones) have circular dependencies and can't be built the nightly jobs will always finish in a "fail" state.

PR #14145, opened 2019-03-22
Update ngless to v0.11.1

luispedro:
This is redistributing the binary, instead of compiling from scratch.

This is far from the ideal solution, but given that there is no
currently-working method to obtain a working recent-enough GHC on the
bioconda build systems, I do not see how else to package NGLess. I note
that stack is also packaged in this fashion.

----------------

### DIFF
--- a/recipes/ngless/build.sh
+++ b/recipes/ngless/build.sh
@@ -1,20 +1,7 @@
-#!/bin/bash
-
-export LIBRARY_PATH="${PREFIX}/lib:/usr/lib:/usr/lib64"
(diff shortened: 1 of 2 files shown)


notestaff:
This PR builds recent GHC, then builds git-annex with it:
https://github.com/conda-forge/git-annex-feedstock/pull/44

notestaff:
also, git-annex-feedstock uses stack with ghc 8.4.2 bindist and lts-13.6 to build git-annex

luispedro:
I am going to merge this now so that users can have the more up-to-date version of ngless, but I will try to adapt your methods to ngless too.

PR #14158, opened 2019-03-23
athena_meta

abishara:
:information_source:

----------------

### DIFF
--- a/recipes/athena_meta/meta.yaml
+++ b/recipes/athena_meta/meta.yaml
@@ -1,0 +1,41 @@
+{% set version = "1.2" %}
+
+package:
+    name: athena_meta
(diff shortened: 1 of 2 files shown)


recipes/athena_meta/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+        - python 2.7
+    run:
+        - python 2.7
+        - bwa 0.7.*
+        - flye 2.4
+        #- samtools 1.3
+        #- htslib 1.3
+        #- pysam 0.9
+        - samtools 1.9
+        - htslib 1.9
```
dpryan79: FYI, we already pin to htslib 1.9 in all recipes.

PR #14164, opened 2019-03-24
Copy bam.h in include/bam

npavlovikj:
:information_source:

----------------

### DIFF
--- a/recipes/samtools/build.sh
+++ b/recipes/samtools/build.sh
@@ -25,3 +25,7 @@ fi
 ./configure CPPFLAGS="$CPPFLAGS" --prefix=$PREFIX --enable-libcurl LDFLAGS="$LDFLAGS"
 make all LIBS+=-lcrypto LIBS+=-lcurl
 make install
+
+# Copy bam.h and other files in bam directory
+mkdir -p $PREFIX/include/bam
+mv ./* $PREFIX/include/bam/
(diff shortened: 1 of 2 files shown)


dpryan79:
Is there a reason for doing this?

npavlovikj:
@dpryan79 , I was planning of picking your brain on this once the build passes...
I am working on a recipe for "braker2" (tested with "samtools/1.7"), that requires newer version of "augustus" than the one available in "bioconda". This version of "augustus" (3.3.2) has a dependency, https://github.com/Gaius-Augustus/Augustus/tree/master/auxprogs/bam2wig, that needs "bam.h" and "libbam.a" to compile.
As you know, "bam.h" and "libbam.a" are only available in "samtools/0.1.19", and not the newer recipes, so that is why I opened this PR. Is there a reason why the newer "samtools" packages don't contain these files?

dpryan79:
I don't think copying the header files from newer samtools releases will work. The API from pre-htslib versions of samtools was somewhat different than what's available with current releases. The good news is that `bam.a` would be statically compiled into `bam2wig`, so you don't need to try and shoe-horn old file names into more modern samtools/htslib releases.

npavlovikj:
I see... Does this mean I should use "samtools/0.1.19" for "augustus"?

dpryan79:
Yes, that should work

npavlovikj:
Many thanks Devon, I will try that next.

PR #14172, opened 2019-03-25
Bump perl-net-ssleay

dpryan79:
:information_source:

----------------

### DIFF
--- a/recipes/perl-net-ssleay/build.sh
+++ b/recipes/perl-net-ssleay/build.sh
@@ -12,24 +12,17 @@ else
     export LD_LIBRARY_PATH="${PREFIX}/lib"
 fi
 
-perl -V
-wrong_prefix="$(perl -V | sed -rn "s|\s+cc='([^']+)/bin/.*|\1|p")"
-if test -n "$wrong_prefix"; then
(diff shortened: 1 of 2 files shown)


dpryan79:
We don't have the luxury of assuming that a suitable perl installation is available, in part because (A) many users are using docker/singularity containers lacking those and (B) many clusters don't provide that. Further, some packages seem to only be available for specific perl versions, so we need to ensure that something compatible is installed. It says something that we have to ship zlib to ensure that everything will work properly on all systems.

Anyway, I have this largely corrected locally using `sed`. I'll get it compiled on Linux today and probably working on OSX tonight.

dpryan79:
I'll check into what's wrong with `make test` (probably not handling running in a docker environment properly) and ensure that the .bak files aren't making it into the built tarball before merging.

dpryan79:
The only test that's causing issues is `t/local/07_sslecho.t`. I suspect the openSSL API has changed for that particular function call.

dpryan79:
OK, this was all fixed in perl-net-ssleay 1.86_06. I'll update this to use the most recent development release from github, since releases are rather slow and openSSL 1.1.1 has been out for almost a year now.

Issue #13784, opened 2019-02-25
post-link script not working as expected/intended


mbhall88:
The [post-link script for the mykrobe recipe](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/mykrobe/post-link.sh) is intended to download some data which is used by the program as a panel for drug resistance prediction at runtime.  
However, when running the bioconda recipe version of mykrobe it seems unable to see this data
```
INFO:mykrobe.cmds.amr:Running AMR prediction with panels data/panels/tb-species-170421.fasta.gz, data/panels/tb-hunt-probe-set-jan-03-2019.fasta.gz
Traceback (most recent call last):
  File "/usr/local/bin/mykrobe", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/site-packages/mykrobe/cli.py", line 8, in main
    args.func(parser, args)
  File "/usr/local/lib/python3.7/site-packages/mykrobe/parser.py", line 40, in run_subtool
    run(parser, args)
  File "/usr/local/lib/python3.7/site-packages/mykrobe/cmds/amr.py", line 256, in run
    cp.run()
  File "/usr/local/lib/python3.7/site-packages/mykrobe/typing/typer/genotyper.py", line 73, in run
    self._run_cortex()
  File "/usr/local/lib/python3.7/site-packages/mykrobe/typing/typer/genotyper.py", line 90, in _run_cortex
    self.mc_cortex_runner.run()
  File "/usr/local/lib/python3.7/site-packages/mykrobe/cortex/mccortex.py", line 183, in run
    self._check_panels()
  File "/usr/local/lib/python3.7/site-packages/mykrobe/cortex/mccortex.py", line 195, in _check_panels
    panel.filepath)
ValueError: Could not find a panel at /usr/local/lib/python3.7/site-packages/mykrobe/data/panels/tb-species-170421.fasta.gz.
```
Have a misinterpreted what the post-link script does or made an error somewhere?

nsoranzo:
xref. https://github.com/bioconda/bioconda-recipes/issues/13846

mbhall88:
@nsoranzo I'm not sure it's the same problem? The post-link script runs fine, it's just that the files don't seem to be where I expect them to be at runtime.

nsoranzo:
@mbhall88 I still don't understand what's causing these issues, so I can't tell if they have the same origin. But they both involve the post-link scripts, so I've added the xref for your information.

[...]

PR #14175, opened 2019-03-25
Data placed in correct location. No longer need to manually move afte…

KyleL1998:
…r download

:information_source:

----------------

Data needed to be manually moved to the correct folder on install, this fix makes sure it gets places there on default.


npavlovikj:
If you check the build logs, you will see the message "BIOCONDA INFO Nothing to be done for recipe recipes/mykrobe". To trigger a new build, you will need to increment the build number in meta.yaml, https://bioconda.github.io/guidelines.html#buildnum.

Next, how big is the data you want to copy? If it small, is there a reason why you are copying it to "$PREFIX/lib/python3.6/site-packages/mykrobe/data"? Preferably, all additional configuration files and data should go to "${PREFIX}/share/${PKG_NAME}-${PKG_VERSION}/" instead.

KyleL1998:
Ah knew I was forgetting something will change.

It has to be moved to that location as mykrobe uses this data to run. And in the newest version that just came out that is where the data is expected to be.

recipes/mykrobe/post-link.sh
```diff
@@ -28,5 +28,5 @@ fi
 
 tar -vxzf "$TARBALL"
 rm -fr "$PREFIX"/src/mykrobe/data
-mv mykrobe-data "$PREFIX"/src/mykrobe/data
+cp -r mykrobe-data/* "$PREFIX"/lib/python3.6/site-packages/mykrobe/data
```
dpryan79: This should be put under `${PREFIX}/share` or something like that, since python 3.6 may or may not be present (in particular, python 3.7 is the default currently).

KyleL1998: Yes, didn't consider different python versions as when I installed it, it always defaulted to this one. Switching it to share though does not fix the problem and I'm right back to where I started trying to fix this issue: https://github.com/bioconda/bioconda-recipes/issues/13784.

KyleL1998: @dpryan79 I've implemented a change. It's no longer locked into a specific python version, and the data is stored in share.

dpryan79: Great!

PR #14209, opened 2019-03-28
update adVNTR to 1.3.0

mehrdadbakhtiari:
:information_source:

----------------


mehrdadbakhtiari:
@bioconda/core 
I'd like to exclude python3.7 from the build since some of the required packages are not compatible with it and original noarch is not useful anymore. Could you please help me to do it? I've tried to write `python <3.7|<=2.9` in the host section of requirements but it causes to only build and test for python3.6, so it also excludes 3.5 and 2.7.
I've also tried to add `skip: True # [py37]` in the build section but that also didn't help.
Thanks

PR #14217, opened 2019-03-28
Adds MALVA recipe

ldenti:
:information_source:

----------------

@bioconda/core Hi, this is my first recipe, can you please review it?

Note:
* I had some problems linking against the version of sdsl-lite available in conda-forge. It seems that it has been compiled with an older version of gcc (maybe 4.9, I'm not sure): it complained about something related to `std::__cxx11::basic_string`. I solved as suggested [here](https://stackoverflow.com/a/37478234) by adding the compiler option `-D_GLIBCXX_USE_CXX11_ABI=0`. Maybe there is a smarter way to solve this problem


dpryan79:
The compiler issue will go away once this is fixed: https://github.com/conda-forge/sdsl-lite-feedstock/pull/4

dpryan79:
I'll see if sdsl-lite can finally get rebuilt for the new compilers.

ldenti:
Ok, I see.

Another thing I forgot to say: I had to include the `kmc_api.tar.gz` since (as far as I saw) they are not installed when installing kmc via bioconda.

dpryan79:
A rebuilt sdsl-lite will be available within ~30 minutes. The compiler hacks shouldn't be needed then.

dpryan79:
I wonder if it makes sense to add the API to the KMC recipe, that'd seem like a more appropriate place. I'll look into that.

ldenti:
The recipe fails to compile on mac since it's not possible to compile kmc (and therefore kmc_api) using clang (see [this issue](https://github.com/refresh-bio/KMC/issues/21)). Maybe this could be a problem if you want to add kmc_api to the kmc recipe. I really have no idea how to solve this problem...

dpryan79:
We'll see how this goes https://github.com/bioconda/bioconda-recipes/pull/14224

dpryan79:
Can you make malva use a KMC shared library? My update to KMC ships API headers under `$PREFIX/include` and a shared library as `$PREFIX/lib/libkmc.so`.

dpryan79:
It's annoying that KMC refers to itself as a library and then only produces programs.

dpryan79:
I've merged in my KMC update, hope that helps.

ldenti:
Thank you very much!

I tried to link against the new lib but `ld` returns some `undefined reference`:
```
main.cpp:(.text.startup+0xd1): undefined reference to `CKMCFile::CKMCFile()'
main.cpp:(.text.startup+0xe0): undefined reference to `CKMCFile::OpenForListing(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)'
main.cpp:(.text.startup+0xd82): undefined reference to `CKMCFile::~CKMCFile()'
main.cpp:(.text.startup+0xe56): undefined reference to `CKMCFile::~CKMCFile()'
main.cpp:(.text.startup+0x14a1): undefined reference to `CKMCFile::Info(unsigned int&, unsigned int&, unsigned int&, unsigned int&, unsigned int&, unsigned int&, unsigned long long&, unsigned long lo
ng&)'
main.cpp:(.text.startup+0x1599): undefined reference to `CKMCFile::ReadNextKmer(CKmerAPI&, unsigned int&)'
main.cpp:(.text.startup+0x15df): undefined reference to `CKmerAPI::char_codes'
```
 I'm pretty sure that the problem is in the new patch you uploaded, [shared_library.patch](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/kmc/shared_library.patch). Line [23](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/kmc/shared_library.patch#L23) should be replaced with:
```
+	$(CC) -shared $(LDFLAGS) -o $@ $^
```
(note the ending ^ and not <).

In such a way all the `.pico` objects will be compiled into the library and not only the first one. Can you update the file or do I have to open a new pull request?

dpryan79:
Oops, good catch! Can you make a PR real quick?

ldenti:
Opened (https://github.com/bioconda/bioconda-recipes/pull/14261). Hope it'll pass the tests

ldenti:
I'm having problems in testing the recipe on osx. Malva fails to compile since it cannot find the kmc headers:
```
main.cpp:37:10: fatal error: 'kmc_file.h' file not found
#include "kmc_file.h"
         ^~~~~~~~~~~~
```

I see [here](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/kmc/meta.yaml#L17) that kmc is not built on osx. Since I need the .h files and the lib to compile malva, do I have to skip the build step on osx too? Do I have to provide a precompiled binary?

dpryan79:
Skip OSX, kmc isn't compatible with clang.

ldenti:
Hi, do you have any idea why the linting is succesfull but the bioconda-test fails on the lint step? Locally, test doesn't produce any error.

dpryan79:
That was just a time out.

dpryan79:
@ldenti Go ahead and merge this whenever you're ready.

ldenti:
Done, thank you!

PR #14218, opened 2019-03-28
Nanopolish bioconda update

paultsw:
:information_source:

----------------

### DIFF
--- a/recipes/nanopolish/meta.yaml
+++ b/recipes/nanopolish/meta.yaml
@@ -6,7 +6,7 @@ package:
   version: {{ version }}
 
 build:
-  number: 1
+  number: 0
 
 source:
   url: https://github.com/jts/nanopolish/archive/v{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


paultsw:
@bgruening to be honest I'm actually not sure w/r/t the gcc exception for OS X. I'm updating the recipe primarily because we've just tagged a new release of nanopolish (v0.11.1) and wanted to be sure the bioconda release was as up-to-date as possible. As I only have access to a linux machine, I'm in favour of otherwise keeping the recipe as-is, aside from the sha256 hash and the bumped version. What's your opinion?

bgruening:
Can we just try and if it does not work revert?

bgruening:
In general, I'm fine with that. We should just remember to fix that at some point. Feel free to merge.

paultsw:
Okay, I'll make a note to fix this for another day.

PertuyF:
The exception is because `gcc` is [not supported for OSX](https://github.com/ContinuumIO/anaconda-issues/issues/10081) using the new compilers, and `nanopolish` does not build with `clang`. See #14014 and #14016 for details.

Using a selector on `libgcc` in run dependencies should be fine I guess.

@paultsw merging for you as I pass by

PertuyF:
@bgruening looks like new packages are available in the channel, but CircleCi noticed that upload failed for Linux.
Is it normal?

PR #14233, opened 2019-03-29
Sepp

sjanssen2:
:information_source:

----------------


sjanssen2:
I could need some advice for the lint'ing issue. SEPP is written in python, comes with a CLI and ships pre-compiled binaries for Linux and Darwin. Thus, I think it should be platform dependent. How can I state that correctly in the recipe?

sjanssen2:
btw: SEPP is the wrapped program of https://anaconda.org/qiime2/q2-fragment-insertion a core part of Qiime2, used by a large community to analyse microbial datasets

sjanssen2:
I have the feeling that the cause of the test failure is not due to my recipe. Can anybody have a look?

npavlovikj:
Looks like anaconda.org is down, and therefore those errors. Maybe try and re-run the workflow again later.

npavlovikj:
Now you have different errors to fix :)
FYI, you "patch" syntax is wrong, please see https://github.com/bioconda/bioconda-recipes/blob/0d379f853882f9a971e801575d2200f1a8d42790/recipes/malva/meta.yaml as an example for how to do it.
Also, please replace "CONDA_PREFIX" with "PREFIX" in the test section.

sjanssen2:
Hi @npavlovikj thanks for your help. I am changing CONDA_PREFIX to PREFIX now.
Regarding the patches. If I understood correctly, patches in the meta.yaml operate on the sources before the build.sh script is executed. At least some of my patches must be executed after some of the build steps have been done. Can this be specified in meta.yaml?

npavlovikj:
The patches are valid during the build process, so please remove all the patches from "build.sh" and add them in "meta.yaml" as suggested earlier. When the build starts, the current directory you are in is the unzipped, root directory of "sepp-0.5.0". Then, the patches are applied, and the rest of the build steps are performed using the modified files.
You can browse the documentation for more examples, https://bioconda.github.io/guidelines.html#patching.

sjanssen2:
Moving patches from build.sh to meta.yaml will work for 2 of the three. But the third patches a file that first needs to be changed by one of the steps in the build script: `python setup.py install -c`. Otherwise, I would have to differentiate myself if I am on Darwin or Linux to copy the correct pre-combiled binaries (shipped with sources) into the bin dir.

npavlovikj:
Can't you use two patches in that case, one for Linux and one for OSX? You can distinguish them when adding the patches in "meta.yaml" with "  # [linux]" and "  # [osx]".

sjanssen2:
it is also differentiating between 32 and 64 bit if linux. Is there also an appropriate tag [] ?

npavlovikj:
You can check the selectors here, https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#preprocessing-selectors, I believe using "[linux64]" and "[linux32]" should work here too.

sjanssen2:
@npavlovikj  I think I now found a proper solution without the need for patches in build.sh.

npavlovikj:
That is great @sjanssen2 , please find few more comments  below:

These 2 lines are redundant:
    - hmmer ==3.1b2 [linux]
    - hmmer ==3.1b2 [osx]
You should just have:
    - hmmer ==3.1b2
instead.

Also, I believe you should rename the "build" section under "requirements" to "host" (please see https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#requirements).

Are "relocate.run-sepp.sh.patch" and "relocate.sepp.config.patch" used during install time, or run time?
If the latter, I think you should replace "/opt/anaconda1anaconda2anaconda3/" with "${CONDA_PREFIX}" instead. "${CONDA_PREFIX}" will be replaced with the location of the installed environment, while "/opt/anaconda1anaconda2anaconda3/" is set during build time. However, you can install and test the package locally, and see what paths you have in the two scripts. 

Please add "license_family" and "license_file" (if available). If not, since you are the developer of the package, you can add it upstream and update the release.

sjanssen2:
Hi @npavlovikj , if I remove the switches (see test) linting falsely states that the package should be noarch, which is NOT true, since compiling IS platform dependent.

npavlovikj:
ah, I see. In that case you can use the `should_be_noarch` linting, https://bioconda.github.io/linting.html#should-be-noarch, and add the following in "meta.yaml":
extra:
  skip-lints:
    - should_be_noarch

sjanssen2:
> If not, since you are the developer of the package, you can add it upstream and update the release.

I am not, but I asked the right person to add those files

sjanssen2:
Hi @npavlovikj,
I worked in all of your helpful comments.

>These 2 lines are redundant:
> - hmmer ==3.1b2 [linux]
> - hmmer ==3.1b2 [osx]

done.

> Also, I believe you should rename the "build" section under "requirements" to "host" (please see https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#requirements).

done.

> Please add "license_family" and "license_file" (if available). If not, since you are the developer of the package, you can add it upstream and update the release.

done.

Regarding requirements: I had some issues for OSX only with dendropy and python (see the many failing tests). Somehow the no-arch version of dendropy on OSX requires specific python versions which are not always compatible when solving the conda environment. Not sure if I am doing it wrong - I fear not. Can you please double check?

sjanssen2:
Hi  @bioconda/core
I figure this PR is almost good to go, see above dendropy OSX issue.

dpryan79:
Let me see if I can fix `dendropy` on OSX to make things a bit easier (I think I see the issue with it).

sjanssen2:
that help would be very welcome! I tried several things (see CircleCI logs)

npavlovikj:
@sjanssen2 , thank you for implementing all these changes. I don't have Mac handy to test the `dendropy` issues, so if Devon can look into that, that would be great.

dpryan79:
I fixed the dendropy issue, which I confirmed on a local mac. I'm trying to stream-line the recipe a bit at the moment to bring it more in line with others in Bioconda.

dpryan79:
Hmm, the dendropy issue is back on OSX, I don't understand what that's happening since it doesn't locally.

dpryan79:
Wild, the dendropy tarball made in my PR works, but the one the master branch uploaded doesn't.

dpryan79:
Gah! Now the previously functional afa8167 commit isn't functional anymore. @sjanssen2 do you really need OSX?

sjanssen2:
I am a Linux guy, but most users use OSX - those biologists ;-)

sjanssen2:
reverted back to a working version. @dpryan79 and @npavlovikj can you please review and merge? I know it is not perfect regarding OSX, but neither @dpryan79 nor I were able to fix it within this package. Seems to be some issue with the dendropy package.

recipes/sepp/relocate.run-sepp.sh.patch
```diff
@@ -0,0 +1,36 @@
+--- /home/sjanssen/Desktop/sepp/sepp-package/run-sepp.sh	2019-03-19 16:33:17.267182143 +0100
++++ sepp-package/run-sepp.sh	2019-03-19 20:23:08.591954892 +0100
+@@ -111,7 +111,7 @@
+ done
+ 
+ 
+-DIR=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
++DIR=/opt/anaconda1anaconda2anaconda3/share/sepp/
```
dlaehnemann: Are you sure, this change is intended here? To me, this folder looks like something that will only be present during the building of the package, but not in the package itself and thus not on the system where it is eventually installed.

From looking at the original file `run-sepp.sh`, I would think that it will not find the default files that you copied to `$PREFIX/share/sepp/ref/` in `build.sh`, i.e. this would break the following lines once the respective options are called:

https://github.com/smirarab/sepp/blob/060779a0eece74db3367509c6a25d7d65f3e43f1/sepp-package/run-sepp.sh#L114-L137

I think the patch should instead leave the `DIR=` definition as is, as this just finds the location of `run-sepp.sh` itself, and change the resource definitions relative to it, so e.g. change:

https://github.com/smirarab/sepp/blob/060779a0eece74db3367509c6a25d7d65f3e43f1/sepp-package/run-sepp.sh#L119

To:
```
	t="$DIR/../share/sepp/ref/reference-gg-raxml-bl-rooted-relabelled.tre"
```

dpryan79: The weird path (`/opt/anaconda1...`) should be found by conda and changed automatically post-install, so that's probably OK.

dlaehnemann: Ah, OK. If conda is that smart, just ignore my comment and feel free to merge if there are no further points.

dpryan79:
@sjanssen2 Shall I add you to bioconda so you can merge this or would you like one of us to merge it?

npavlovikj:
@sjanssen2 , thank you so much for working on this recipe!

The recipe looks good to me - if I am being nitpicky though, I would suggest you to add `license_family` and `license_file` in `meta.yaml` since you already have that information in the source tarball :)
Also, do you expect the "osx" build to use Python 2.7? I could only see the Python 3.6 tarball generated...

@dpryan79 , is `python [linux]` == `python  # [linux]` now?

dpryan79:
@npavlovikj Good catch

sjanssen2:
Thanks @dlaehnemann for looking into this recipe!

@dpryan79 I'd be delighted to become a member. Merging my own PR doesn't feel right. Can someone else please hit the button once all issues are solved?

@npavlovikj I love nitpicky people, that's how I can improve my coding! I don't know how to trigger py27 and py36 builds for OSX. Honestly, I don't care so much for py27, but I didn't want to be too restrictive in case people want to install other packages in the same conda env.

dpryan79:
Merged, thanks @sjanssen2!

sjanssen2:
amazing! thanks @dlaehnemann @npavlovikj and @dpryan79 for your great help.

PR #14250, opened 2019-03-31
Add recipe for eval

npavlovikj:
:information_source:

----------------

### DIFF
--- a/recipes/eval/paths.patch
+++ b/recipes/eval/paths.patch
@@ -2,7 +2,7 @@
 +++ combine_eval_reports.pl.new	2019-04-02 03:48:07.807126910 -0500
 @@ -1,4 +1,4 @@
 -#!/usr/bin/perl
-+#!/usr/bin/env perl
++#!/usr/bin/env perl 
  
  use strict;
  use Eval;
--- a/recipes/eval/paths.patch
+++ b/recipes/eval/paths.patch
@@ -1,0 +1,112 @@
+--- combine_eval_reports.pl	2003-05-20 11:52:45.000000000 -0500
++++ combine_eval_reports.pl.new	2019-03-31 01:58:52.812255998 -0500
+@@ -1,4 +1,4 @@
+-#!/usr/bin/perl
++#!/opt/anaconda1anaconda2anaconda3/bin/perl
+ 
+ use strict;
+ use Eval;
+--- eval.pl	2004-05-03 12:10:39.000000000 -0500
++++ eval.pl.new	2019-03-31 01:59:12.675255916 -0500
+@@ -1,4 +1,4 @@
+-#!/usr/bin/perl -w
++#!/opt/anaconda1anaconda2anaconda3/bin/perl
+ use strict;
+ use lib '.';
+ use Tk;
+--- evaluate_gtf.pl	2006-06-02 15:20:47.000000000 -0500
++++ evaluate_gtf.pl.new	2019-03-31 01:59:36.350255817 -0500
+@@ -1,4 +1,4 @@
+-#!/usr/bin/perl
++#!/opt/anaconda1anaconda2anaconda3/bin/perl
+ use strict;
+ use lib '.';
+ use Eval;
+--- filter_badlist.pl	2006-06-02 15:20:47.000000000 -0500
++++ filter_badlist.pl.new	2019-03-31 01:59:56.506255733 -0500
+@@ -1,4 +1,4 @@
+-#!/usr/bin/perl -w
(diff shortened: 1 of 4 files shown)


npavlovikj:
Hmm, I don't understand why the build takes so long... `conda build` works fine, but even locally `bioconda-utils build` times-out ¯\_(ツ)_/¯

npavlovikj:
@bioconda/core , I would appreciate any feedback on this recipe. 
It is a simple recipe, but for some reason using `bioconda-utils build` times out after 5 hours (both here and locally), while `conda build` works fine. I don't see any issues with the debug option - just lots of dots for this step:
_20:46:04 BIOCONDA DEBUG DOCKER: cmd: ['docker', 'run', '-t', '--net', 'host', '--rm', '-v', '/tmp/tmp8hzw4zmi/build_script.bash:/opt/build_script.bash', '-v', '/tmp/min
iconda/miniconda/conda-bld:/opt/host-conda-bld', '-v', '/home/fedora/bioconda-recipes/recipes/eval:/opt/recipe', '-e', 'CONDA_SHLVL=1', '-e', 'CONDA_EXE=/tmp/miniconda/
miniconda/bin/conda', '-e', 'LANG=en_US.UTF-8', '-e', 'CONDA_PYTHON_EXE=/tmp/miniconda/miniconda/bin/python', '-e', 'CONDA_PROMPT_MODIFIER=(base) ', '-e', 'CONDA_DEFAUL
T_ENV=base', '-e', 'CONDA_ROOT=/tmp/miniconda/miniconda', '-e', 'CONDA_ALLOW_SOFTLINKS=false', '-e', 'HOST_USER_ID=1000', 'tmp-bioconda-builder', '/bin/bash', '/opt/bui
ld_script.bash']_

I don't know why the build never finishes, so if you notice something wrong and odd in the recipe, please let me know. I tried simplifying the recipe, as well as removing `noarch`, but the issue persists...

recipes/eval/paths.patch
```diff
@@ -0,0 +1,112 @@
+--- combine_eval_reports.pl	2003-05-20 11:52:45.000000000 -0500
++++ combine_eval_reports.pl.new	2019-03-31 01:58:52.812255998 -0500
+@@ -1,4 +1,4 @@
+-#!/usr/bin/perl
++#!/opt/anaconda1anaconda2anaconda3/bin/perl
```
dpryan79: You'll want to change these to `#!/usr/bin/env perl`.

npavlovikj:
@dpryan79, you helped me again - I really appreciate your prompt replies and valuable assistance! 
So, I should avoid using `/opt/anaconda1anaconda2anaconda3/bin/` in future recipes?

dpryan79:
Yes, that path won't actually exist outside of the testing environment.

recipes/eval/paths.patch
epruesse: @dpryan79 We may have to revisit this. Technically, it should be the absolute path to the current `perl`, so that conda can, same as for python, pick up that path and rewrite it to point to the perl from the env into which the package is installed. Relevant because the perl include path isn't set AFAIK, so it needs to be the right perl instance for the deps to be found.

dpryan79: Doesn't perl change the path in the shebang once it actually installs modules, as python does?

acaprez:
@dpryan79 - that's not correct, unless I'm misunderstanding what you're saying.  Conda will change `/opt/anaconda1anaconda2anaconda3` to whatever the env path is at environment create/install time, irregardless of whether that's a testing environment or a final install.  Using `/opt/anaconda1anaconda2anaconda3` guarantees that the correct interpreter is always used, whereas `/usr/bin/env perl` may point at a different `perl` depending on what `$PATH` contains.

npavlovikj:
So, I did look at my earlier patch file, and noticed that I had one of the files mixed up.
I had
**-#!/opt/anaconda1anaconda2anaconda3/bin/perl**
**+#!/usr/bin/perl**
instead of
**+#!/opt/anaconda1anaconda2anaconda3/bin/perl**
**-#!/usr/bin/perl**
Since this is wrong, probably the interactive prompt showed and the build got stuck there.

I didn't pay attention to this when I used "_conda build_" and although I checked the recipe few times, I didn't notice this until now.

I have used "_/opt/anaconda1anaconda2anaconda3/bin/_" in other recipes in the past few weeks (because of the reasons Adam has mentioned) which built and worked fine. 

I believe the wrong patch was the reason for my build being stuck.
I appreciate everyone's input, and I am sorry for not seeing the mistake sooner and bothering you all in the meantime.

PR #14252, opened 2019-03-31
Change fgbio receipe to avoid over-allocation of virtual memory

tfenne:
@nh13 I'd appreciate a second set of eyes on this if you would please?

### DIFF
--- a/recipes/fgbio/fgbio.py
+++ b/recipes/fgbio/fgbio.py
@@ -2,9 +2,7 @@
 #
 # Wrapper script for invoking the jar.
 #
-# This script is written for use with the Conda package manager and is ported
-# from a bash script that does the same thing, adapting the style in
-# the womtool wrapper
+# This script is written for use with the Conda package manager.
 
 import subprocess
 import sys
@@ -74,6 +72,14 @@ def main():
         jar_arg = '-cp'
     else:
         jar_arg = '-jar'
+        
+    # If not already set to some value, set MALLOC_ARENA_MAX to constrain the number of memory pools (arenas)
+    # used by glibc to a reasonable number.  The default behaviour is to scale with the number of CPUs, which
+    # can cause VIRTUAL memory usage to be ~0.5GB per cpu core in the system, e.g. 32GB of a 64-core machine
+    # even when the heap and resident memory are only 1-4GB!  See the following link for more discussion:
+    #  https://www.ibm.com/developerworks/community/blogs/kevgrig/entry/linux_glibc_2_10_rhel_6_malloc_may_show_excessive_virtual_memory_usage?lang=en
+    if not os.environ.get("MALLOC_ARENA_MAX"):
+        os.environ["MALLOC_ARENA_MAX"] = "4"
 
     jar_path = path.join(jar_dir, JAR_NAME)
     java_args = [java] + mem_opts + prop_opts + [jar_arg] + [jar_path] + pass_args
--- a/recipes/fgbio/meta.yaml
+++ b/recipes/fgbio/meta.yaml
@@ -11,7 +11,7 @@ package:
 
 build:
   noarch: python
-  number: 0
+  number: 1
 
 source:
   url: https://github.com/fulcrumgenomics/fgbio/releases/download/{{ version }}/fgbio-{{ version }}.jar


epruesse:
@bioconda-bot lint

epruesse:
Also, please set a title for the PR that makes sense to everyone. Something that includes the recipe name and very briefly the problem you are trying to address.

epruesse:
Not sure what the scenario is here either. Is this specific to the recipe in question, or affect all of our recipes? It it specific to Java or everything? Is it something that will happen on all systems, just Linux, just Redhat or just RHEL6?

tfenne:
@epruesse - sorry for the rather cryptic description.  It's been a long running problem for many users of various java/jvm-based bioinformatics toolkits that on modern linuxes (not just RHEL, but any with a recent glibc version) that the applications reserve way more virtual memory than they allocate physical memory.  This is particularly problematic when running tools under Grid Engine or another HPC system which places hard limits on virtual memory as opposed to resident memory.

Having run into this many times previously I was recently motivated to find a real solution, and came across the article linked in the comment in the recipe.  What this does is ensure that on systems with many cores (common in HPC clusters) that the glibc, on behalf of the JVM, won't reserve too much virtual memory.

epruesse:
Ping @bioconda/core

This might need/justify change of all wrappers and the skeleton, as well as further testing, mention in the docs and review of java tools bringing their own wrappers (bbmap, fastqc, ...).

chapmanb:
Tim -- thanks for this fix and the description of the problem. I hadn't realized there were memory issues not handled by Xmx and Xms settings so this definitely seems useful, although the impact of setting it is not totally clear to me. The linked article hedges on what a reasonable default setting is and I don't really understand the impact of limiting virtual memory. Would it make sense to tie this to the machine cores somehow so that we maintain a consistent default independent of total machine cores?

More generally, I'm +1 for merging this for fgbio for fixing a practical problem but am a bit more nervous about applying it globally across all wrappers without knowing how it will impact across lots of different systems and cases.

tfenne:
I agree with @chapmanb - it might be good to give this a little runtime in the fgbio recipe before updating more broadly.

@chapmanb: this doesn't place any hard limits on virtual memory.  What it does, indirectly, is place a limit on how much virtual memory `glibc/malloc` will allocate in order to manage the process of allocating memory.  So a given process can still memory map or otherwise use virtual memory as much as it likes, but it won't have a high incidental virtual memory usage caused by the internals of `malloc`.

This is generally only a problem on systems that care about virtual memory usage.  The one that always bites me is Grid Engine, as it is very frequently setup to manage memory as a consumable hard-limited resource but measures _virtual_ memory instead of resident or physical memory.  For any system without similar limits this isn't generally a problem, as on 64-bit systems we can allocate more virtual memory than we could ever imagine.

This is where my understanding gets a little less clear - it would appear the arena-based strategy in more recent `glibc` (and recent here is, I think circa 2010) is intended to improve `malloc` throughput when there are lots of threads allocating memory, by allocating pools of address space per thread.  However, I suspect that for the JVM, where most heap allocation is managed (i.e. he JVM `malloc`s large blocks of memory and then handles memory allocation & GC internally) the benefit of this new strategy is probably very minor.

nh13:
I support this PR.

epruesse:
Ok, I think I understand what's going on.

### TL'DR: 

IMHO this is a site specific problem the cluster admins need to address. I can see no fault on the side of `glibc` or the JVM.

The value of `SIZE` (`VM`) gives the size of the address space requested by a process. *Not* the memory allocation of the process. It is therefore of little use to control resource usage on a cluster. It's particularly pointless for Java processes as those request a fixed, usually large amount of address space on launch.

 `RSS` does not include the portion of a process that has been "swapped" (really paged) to disk, so while much better than `SIZE`, it's not ideal either. The size of a process as a sum of its swapped and allocated packages is not available via `ps` or `top`, but tools like `smem` can be used to measure it. 

### VM vs RSS (simplified, glossing over details):

Applications usually request memory from the kernel via either `mmap` & friends or `sbrk` & friends. The second increases the point at which an application will get a "stack overflow". The first asks the kernel for a piece of address space that may or may not be backed by a file. Neither ask for memory directly, just for address space (=`VM`). That is,  unless the application specifically asks to have the memory "pinned", it leaves the kernel free to back the request with physical memory on an as needed basis. If the address space is not associated with a file, it'll be backed by memory only when it is written to. If it's a file, the file will be read on demand. If physical memory runs out, the kernel will "unback" some memory, writing it to disk until it is needed again ("paging").

That's how we get "VM" as a large chunk that has nothing to do with the reality of a process, and "RSS" which shows the actual usage. "SHR" indicates memory that can be used by multiple processes (e.g. a BLAST database `mmap`ed read-only can be used by multiple processes at once, or the memory into which `glibc` or `libopenssl` etc are loaded - this is part of the point of shared libs: saving memory).

Java makes heavy use of this "on demand" allocation. When you tell it that the JVM may have up to 32GB of RAM for it's heap via `-Xmx32G`, it will `mmap` that amount and it will immediately show up as `VM` size. But it won't actually be used. VM size is meaningless for Java processes. You might as well say that a non-Java process "uses" all of the available memory, because that is what it can in theory allocate at a later date.

### Allocation "Arenas"

IIRC `jemalloc` was the first to introduce the concept. There is a paper on it somewhere. 

Basically, if you have a lot of threads concurrently requesting and releasing memory but only one data structure to store free/allocated memory regions, the threads have to wait on one another when they need to acquire or free memory. The "arena" approach is to simply have a number of independent areas to allocate memory from. That way the threads don't collide as much. Typically you'd scale the number of arenas with the number of threads and have each thread pick one in the manner of a hash map selecting bins. With a bit of tuning, you can get rid of most of the contention and things perform and scale much better.

An additional benefit is that if each thread tends to get memory from a region mostly dedicated to it is that "locality" improves. The caches in the CPUs and cores work better if there are no other threads interfering. The worst scenario in a mixed layout is that a cache-line is shared between two threads running on different CPUs. Each time that data is accessed, the CPUs have to check whether the data was modified by the other thread, and that costs enormous amounts of time (relative to e.g. doing math).

The need to "mmap" a few large chunks of memory to start with for each thread again doesn't matter because the kernel will not back it with physical memory until it is actually used. So while it does increase "VM" size, it doesn't impact "RSS".

tfenne:
@epruesse That's a very nice summary, thank you.  I'd agree with you that neither `glibc` or the JDK are to blame here.  But I think the confluence of the two does produce some (to most) unexpected results, with the virtual memory usage exploding on machines with many cores.  I suspect the default of `arenas = cores * 8` for 64-bit systems was introduced long before systems with 32-64 became common-place.

If anyone is to blame, I'd say it's the systems that use SIZE or VIRT for measuring a processes memory usage.  While I don't disagree that in an ideal case this would be a job for cluster admins, I can also state with a straight face that a) this was an issue for me when running the sequencing pipelines at the Broad and b) is a problem for me today at a large clinical sequencing vendor.  At both locations, nobody knew about this or how to resolve it!  Over the years I've also seen pleny of conversation on the GATK mailing lists and forums with users experiencing similar problems, without ever being offered a solution.

That's why I'd like to fix it in the fgbio wrapper (I'm one of the authors of fgbio, @nh13 is the other), to ensure nobody has these issues with fgbio-via-bioconda going forward.  I've also submitted a PR to the gatk repo to adapt their wrapper script.

PR #14255, opened 2019-04-01
Store built packages on CI's artifact system after test

epruesse:
:information_source:

----------------

Changes (hopefully) the CircleCI config so that the built-packages get uploaded as artifacts after the test. This should make debugging a little easier. Potentially might even allow installing the package locally after successful test, to double check things that cannot be easily tested with the package tests facility. (X11 stuff e.g.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,7 @@ variables:
       command: |
         mkdir -p /tmp/artifacts/packages
         cd miniconda/conda-bld || exit 0
-        for n in rss.xml index.html channeldata.json linux-64 osx-64 noarch; do
+        for n in rss.xml index.html channeldata.json linux-64 osx-64 noarch broken; do
           cp -rv $n /tmp/artifacts/packages || true
         done
   store_artifacts: &store_artifacts
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,6 +45,19 @@ variables:
             echo "Please use a branch other than 'master' or 'bulk' on your fork"
             exit 1
         )
+  store_built_packages: &store_built_packages
+    run:
+      name: Copying packages
+      when: always
+      command: |
+        mkdir -p /tmp/artifacts/packages
+        cd miniconda/conda-bld || exit 0
+        for n in rss.xml index.html channeldata.json linux-64 osx-64 noarch; do
+          cp -rv $n /tmp/artifacts/packages || true
+        done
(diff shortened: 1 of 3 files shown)


epruesse:
ping @bioconda/core - What do you think?

Basically, this PR makes it so that the `conda-bld` directory gets uploaded to CircleCI artifacts. The Bot listens to `check_run/completed` events, finds the matching build on CircleCI, scans the list of artifacts and generates the above posting.

- Only one post will be created - at least unless someone edits the first line (that's what I use for matching). If the build is updated, the post is also updated. I could alternatively delete the old post and create a new one, so that it "wanders" downwards for long PRs, but I definitely didn't want tons of those longish comments.
- There is a bit of a disconnect in the checks API. It's based on the commit SHA, which is not always connected to a PR. Sometimes the webhooks contain the PRs, sometimes they don't. Seems to be pretty much random. Not sure how to fix it because there is no real API to get a PR from a commit.
- The next disconnect is on the way to CircleCI. The build number is needed to get the artifacts, but that is only available via either the branch name (for PRs within a repo) or the "pseudo branch" created by Github (refs/pulls/<#>). So I need to go from SHA to PR to BRANCH to get the build number. Sigh.

epruesse:
On the note of why - this allows a bunch of things to be done locally:
 - testing of X11 and similar stuff manually
 - debugging of packages that fail tests by downloading the built binary
 - testing of conda package resolution by adding temporary channel 

It's not the `bioconda-canary` channel that I was thinking about at some point, but it serves a similar purpose.

It also serves to illustrate some other things we might be able to do with the bot. Once the CircleCI has been identified, it should be possible to do post-mortem analysis of the logs:
 - restart job if there is an "Anaconda Gateway Error (TM)"
 - parse and comment on typical problems like "import error"
 - verify that the whole thing didn't end in "no packages built"
 - ...

johanneskoester:
Great work! We could also just upload the artifacts of a PR once it is merged to master couldn't we? This way we would avoid the double build. Regarding the API issues and indirections: would it be possible to just create the notification message from CircleCI, without the bot? From the CircleCI job, everything necessary should be known.

dpryan79:
This is great, it will certainly facilitate debugging things!

epruesse:
Hmm. Interesting idea to just use those builds when merging. We’d have to still build the container though.

The reason none of this can be done from Circle itself are the secrets. Everything we make available to the builds can be accessed by anyone.

Issue #13964, opened 2019-03-08
packages that should be migrated to conda-forge


bgruening:
A few packages in bioconda, should actually be migrated to conda-forge. This can be a very simple task and is ideally suited to learn both bioconda and conda-forge.

We will keep a list here of packages that we think should be migrated:

- [ ] grep

npavlovikj:
In case this helps, here are few other general packages as well:
- [ ] shapeshifter
- [ ] shapeshifter-cli
- [ ] sharedmem
- [ ] r-ldrtools
- [ ] r-syntactic
- [ ] r-msm
- [ ] r-swamp
- [ ] r-perfmeas
- [ ] r-mess 
- [ ] smhasher
- [ ] fitter
- [ ] gromacs
- [ ] logbook
- [ ] xopen
- [ ] libdeflate
- [ ] optbuild
- [ ] fit_nbinom
- [ ] pytest-workflow
- [ ] cromwell
- [ ] scikits-datasmooth

simonvh:
I'm moving bucketcache and its deps (logbook and represent) to conda-forge. What is the procedure once they're up and running at conda-forge? Delete them from bioconda-recipes via PR?

pb-cdunn:
Somewhat related question:

There is a package in Anaconda on a non-standard channel that I would like to use at build-time for Bioconda packages, but it's not in `conda-forge`. How can I use it in a Bioconda recipe? (At build-time only.)

Would it be easier if that package were actually in conda-forge?

dpryan79:
Regarding what to do once packages are migrated, just put them in the `build-fail-blacklist` file (put the reason why above as a comment). At some point we might due a "grand purge" of such recipes.

dpryan79:
@pb-cdunn Yeah, try to get it into conda-forge. You can probably download the package from anaconda.org and largely copy what's under `info/`.

rhpvorderman:
>  cromwell

I would say that WDL is more aimed at the life sciences than anything else. But I can understand if this needs to be in conda-forge.

If this is the case then I will migrate `pytest-workflow` also to conda-forge as soon as cromwell is moved.

dpryan79:
N.B., recipes can be deleted once migrated to conda-forge.

marcelotrevisani:
Please, remove ``pytest-workflow`` from ``bioconda``

jenzopr:
I reviewed all `r-` packages from above and will comment only *problematic* packages here:

[...]

PR #14280, opened 2019-04-02
Updated build-fail-blacklist for packages moved to conda-forge

simonvh:

See #13964, these recipes have been migrated to conda-forge.


dpryan79:
Good catch, thanks @simonvh!

epruesse:
Should the ones for which the migration is complete be listed in `build-fail-blacklist`? Or should they just be removed? Including the `recipes/package-name` folder?

epruesse:
Since we have facility for multiple blacklist, it could be a `migrated-to-conda-forge` blacklist, actually.

druvus:
I believe that also the actual `recipes/package-name` folders should be deleted for migrated packages.

dpryan79:
I'm also fully supportive of deleting the folders and would also suggest removing deprecated versions of bioconductor packages and things like that were the source tarballs vanish with time.

dpryan79:
I'll merge this for now and add this topic to the core meeting agenda.

epruesse:
+1 - just what I was going to suggest. We need a policy on that and consensus on it. But I see no reason to keep the recipes. We might still keep a list of those packages somewhere, perhaps with the hash of the last commit at which they were present, so that they can be resurrected should there be a need at some point. I don't really see a situation for that though.

PR #14285, opened 2019-04-02
Adding PyBDA: trial 2

dirmeier:
:information_source:

----------------

### DIFF
--- a/recipes/pybda/meta.yaml
+++ b/recipes/pybda/meta.yaml
@@ -23,7 +23,7 @@ requirements:
     - python ==3.6
     - click >=6.7
     - joypy >=0.1.10
-    - matplotlib >=2.2.3
+    - matplotlib >=3.0.2
     - numpy >=1.15.0
     - pandas >=0.23.3
     - pyspark ==2.4.0    
--- a/recipes/pybda/meta.yaml
+++ b/recipes/pybda/meta.yaml
@@ -11,11 +11,11 @@ source:
 
 build:
   number: 0
-  skip: true # [not py35 and not py37]
+  noarch: python
 
 requirements:
   host:    
-    - python
+    - python ==3.6
     - pip
     - numpy
     - setuptools
@@ -46,6 +46,5 @@ about:
   doc_url: https://pybda.readthedocs.io/en/latest/
   dev_url: https://github.com/cbg-ethz/pybda
 
-extra:
-  recipe-maintainers:
-    - dirmeier
+recipe-maintainers:
+  - dirmeier
--- a/recipes/pybda/meta.yaml
+++ b/recipes/pybda/meta.yaml
@@ -16,8 +16,8 @@ build:
 requirements:
   host:    
     - python
-    - pip   
-    - numpy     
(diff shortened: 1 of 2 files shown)


dirmeier:
Ok, this seems to work now. Could someone from @bioconda/core please have a look at this? PyBDA is a package for big data analysis of biological data sets using Spark and Snakemake. Would be great if this would get merged. 

Thanks,
Simon

recipes/pybda/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 541908e3874550de3fdceb118279e7530180e1fe9caf7319fe0e29408227f788
+
+build:
+  number: 0
+  skip: true # [not py35 and not py37]
```
dpryan79: Aren't you going for `skip: True  # [not py36]` ? You can probably just remove this, make it `noarch: python` and then specify `python 3.6` in the host and run sections.

recipes/pybda/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+requirements:
+  host:    
+    - python
```
dpryan79: Given the run section, I presume you need `- python 3.6` here.

recipes/pybda/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  home: https://github.com/cbg-ethz/pybda
+  license: GNU General Public v3 (GPLv3)
+  license_family: GPL3
+  summary: Big biological data analytics powered by Apache Spark
+  doc_url: https://pybda.readthedocs.io/en/latest/
+  dev_url: https://github.com/cbg-ethz/pybda
+
+extra:
+  recipe-maintainers:
+    - dirmeier
```
dpryan79: You can remove the `extra:` section, we don't end up using it in bioconda (not that it's hurting anything).

recipes/pybda/meta.yaml
dpryan79: This is actually causing it to not build anything, which is why all of the testing passes (there's nothing to fail).

dirmeier: Oh man, thanks!

dirmeier:
Right...I had the same kind of error with another PR and couldn't fix it. Any ideas @dpryan79? Thanks for your help!

dpryan79:
@dirmeier It passes CI testing and builds a package now. If you post a link to the other PR I'll have a look.

dirmeier:
Wow, thanks so much, @dpryan79 . The other PR was the same package. I closed it, since I assumed it failed due to some old the pinning or so. So no need to look at it. Thanks again!

dpryan79:
Great, go ahead and merge this whenever you're ready then.

dirmeier:
Cool, thanks for the help again!

PR #14317, opened 2019-04-03
Increment build number of Genomedata for GCC migration and Python 3.7

EricR86:
:information_source:

----------------


epruesse:
@biocondabot lint

epruesse:
@BiocondaBot lint

epruesse:
@BiocondaBot lint

epruesse:
@BiocondaBot lint

epruesse:
@EricR86 Apologies for hijacking your PR to work on the bot. Does the Linter message help you at all? What do you think it should say to make it easier to find the problem?

EricR86:
@epruesse nope it's fine. The original linting messages were not helpful (about a deprecated package I no nothing about) but the new one is just fine and has been addressed.

EricR86:
@epruesse 
Specifically I'm again running into the same linting error message from `bioconda-utils`:
```shell
15:01:48 BIOCONDA DEBUG lint recipes/logbook
15:01:56 BIOCONDA DEBUG lint recipes/represent
	recipe	check	info	exists_in_channels	fix	output
0	recipes/logbook	in_other_channels	{'exists_in_channels': {'conda-forge'}, 'fix': 'consider deprecating', 'output': 'logbook'}	{'conda-forge'}	consider deprecating	logbook
```
I'm not certain why these recipes are being linted in this PR (this recipe does not depend on them afaik and they have not been modified). But the linter stops because it says they should be considered deprecated? I believe in an earlier build it says that they were newly unblacklisted.

Not sure what is going on here.

EricR86:
@BiocondaBot recheck

epruesse:
> I'm not certain why these recipes are being linted in this PR (this recipe does not depend on them afaik and they have not been modified). But the linter stops because it says they should be considered deprecated? I believe in an earlier build it says that they were newly unblacklisted.
> 
> Not sure what is going on here.

Sorry for the late reply. It probably got lost with the way Github clears the notice as soon as you open the page once.

Basically, there is a latent bug in `bioconda-utils`, a sort of regression when we moved to CircleCI from Travis. Travis builds the "merge commit", so the way the repo looks after a merge. Circle builds the "merging branch", so the way that your branch looks. We need the former though, to make sure that the tests run the same as they would upon merge. Somehow, this leads to problems when you build from a branch you made in your fork with a master that has gotten old, or a master you updated later.

I think I have a fix, but it's so long that it's likely to introduce a bug statistically, so its trading a known problem for an unknown one...

build-fail-blacklist
```diff
@@ -36,7 +36,8 @@ recipes/emboss
 recipes/embassy-phylip
 recipes/triform2
 recipes/haploclique/1.3.1
-recipes/genomedata
+recipes/segtools
+recipes/transrate-tools
```
epruesse: Adding those two became necessary to get it to not build those?

EricR86: Must have been a mistake I think. Merge conflict said that master had those two lines (segtools and transrate) in it while my version had genomedata removed. I just wanted master without the genomedata line so I kept the lines from master and it seemed to be marked as additions?

Transrate is oddly enough already in the blacklist but segtools, which previously was, is not anymore on master. I will adjust accordingly so this only accidentally readds segtools. I will fix.

epruesse:
looking good! Thanks!

PR #14319, opened 2019-04-03
PepGenome Recipe

ypriverol:
:information_source:

----------------


recipes/pepgenome/meta.yaml
```diff
@@ -0,0 +1,39 @@
+# Do not forget to update the version string in the peptide-shaker.py file
+{% set version = "v1.0.beta" %}
```
bgruening: This should probably be `1.0+beta` according to semantic versioning.
Please also not the skipped `v` but the `v` as part of your URL.

ypriverol:
@bgruening I can't see what is the error

PR #14330, opened 2019-04-04
ganon recipe

pirovc:

### DIFF
--- a/recipes/ganon/meta.yaml
+++ b/recipes/ganon/meta.yaml
@@ -6,10 +6,8 @@ package:
   version: "{{ version }}"
 
 source:
-  url: https://github.com/pirovc/{{ name }}/archive/{{ name }}_v_{{ version }}-test.tar.gz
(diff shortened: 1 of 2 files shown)


pirovc:
skipping osx build due to version of clang which is currently on version 4 (ganon requires clang>=6)

recipes/ganon/meta.yaml
```diff
@@ -0,0 +1,45 @@
+{% set name = "ganon" %}
+{% set version = "0.1.2" %}
+
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  url: https://github.com/pirovc/{{ name }}/archive/v{{ version }}.tar.gz
+  sha256: 84814caf51616911166840819fd14b9436bcf2cdfd261040ac1064a9acd4d16a
```
bgruening: You can use here multiple URLs and checksums to download multiple repos. This would remove the git dependency and we can backup the tarball.

pirovc: thanks, it makes the recipe much better indeed

PR #14373, opened 2019-04-08
Updated Trimmomatic 0.39 (same version) requires new build and SHA256

rspreafico-vir:
Trimmomatic 0.39 was updated due to bug. Same version.

:information_source:

----------------

### DIFF
--- a/recipes/trimmomatic/meta.yaml
+++ b/recipes/trimmomatic/meta.yaml
@@ -5,11 +5,11 @@ package:
   version: '{{ version }}'
 
 build:
-  number: 0
+  number: 1
   noarch: generic
 
 source:
-  sha256: 33c882201cc81cf51e533f2129ccdfdb027f0c799886258cbf5c4815fa0593f7
+  sha256: 2f97e3a237378d55c221abfc38e4b11ea232c8a41d511b8b4871f00c0476abca
   url: http://www.usadellab.org/cms/uploads/supplementary/Trimmomatic/Trimmomatic-{{ version }}.zip
 
 requirements:


epruesse:
Curious - do you know why the SHA changed?

rspreafico-vir:
Yes, the tool was updated due to a null pointer issue but the version remained 0.39 I think because this version was only released only a few days ago. Effectively this would be version 0.39a, so to speak, but it's still tagged as 0.39 in the original tool.

epruesse:
0.39a would come before 0.39 in conda and python version ordering (alpha release), so don't do that. It would be great if you could coordinate with upstream to call it say 0.39.1. It's a bit of an academic thing, but we are in academia here after all. :)

epruesse:
If it takes too long, add a note to the package/recipe explaining the difference between build 0 and build 1. It'll show up on the docs page, so people googling for 'bioconda trimmomatic' will find it. 

If you had added a patch it would be a minor code change too, so that should be ok. Even though since it was from upstream, a proper new number would be preferred.

rspreafico-vir:
I agree, but the creator released the update as 0.39 and the old binaries are gone even from the website. By all means, the most recent version is 0.39, meaning that the current bioconda recipe will pull down the update tool but with the older SHA, so won't work - hence why it would be urgent to fix the recipe.

rspreafico-vir:
Ok let me add a note to the recipe

epruesse:
Thanks

epruesse:
Feel free to merge once that's in. Thanks for taking care of this.

rspreafico-vir:
Not a prob. I think you will need to merge though. No permissions here :)

epruesse:
(BTW: we do have a backup of the old download.)

epruesse:
Ah, I see you are not a member yet. Just post on #1 - you would be very welcome!

rspreafico-vir:
Awesome! How about this?

```
Build 1 of Trimmomatic 0.39 pulls a different tools than build 0. Both tools were given version 0.39, but the initial codebase suffered from a null pointer error. That error was fixed in a later release of version 0.39. This new version, and its SHA256, are now referenced by build 1 of the bioconda recipe.
```

Incidentally, is this something you can add on your end without re-triggering CI/CD?

epruesse:
It would have been better to have the note in there before going ahead an merging.

rspreafico-vir:
I did add the note directly to the merge commit

epruesse:
Ah, sorry. I meant add it there **and** put something in extra. We had the core meeting, so I couldn't follow up here. Well. It's done.

epruesse:
We might want to have a changelog for the recipes down the line I think.

PR #14376, opened 2019-04-08
Bump wg-blimp version to v0.9.2

MarWoes:
:information_source:

----------------


npavlovikj:
For future reference, when the version is updated, the build number should be reset to 0.

MarWoes:
Thanks you for the hint, I wasn't aware of that. I will reset it to 0 with the next version bump!

PR #14382, opened 2019-04-09
Hyphy version 2.3.14 and mac

rdvelazquez:
:information_source:

----------------

2.3.14 doesn't use HYPHYMPI (HYPHYMPI is really only used for GARD and 2.3.14 doesn't have a stable version of GARD); because no HYPHYMPI, 2.3.14 doesn't need openmpi


recipes/hyphy/conda_build_config.yaml
```diff
@@ -0,0 +1,6 @@
+c_compiler:
```
bgruening: Any reason this file and the old compilers are needed?

rdvelazquez: Thanks for looking at this @bgruening

Not sure if this file is still needed, just following what was done in the past. 

The old compilers shouldn't be needed.

@davebx any thoughts?

davebx: I honestly don't remember adding that. I say remove it and see what happens.

rdvelazquez:
Do I need to do anything else to get `bioconda-test` to run?

rdvelazquez:
Having Circle-CI enabled on my fork may have been causing the CI to hang. I've deactivated CI on my fork and pushed https://github.com/bioconda/bioconda-recipes/pull/14382/commits/2230773237bfb96cd7b67ec03dbe070b2a619a0b to re-trigger a re-build.

bgruening:
Thanks @rdvelazquez

Issue #14372, opened 2019-04-08
Lumpy biocontainer: missing samtools/sambamba in the config


cagaser:
Hello, 

I tried to run the lumpyexpress from the  [quay.io/biocontainers/lumpy-sv:0.3.0--h0b85cd1_0](url) 
However I get the following error message:
`Error: neither samtools nor sambamba were found. Please set path of one of these in /usr/local/bin/lumpyexpress.config file`

The config file looks like this:

```
LUMPY_HOME=/opt/conda/conda-bld/lumpy-sv_1554621096029/work/

LUMPY=/opt/conda/conda-bld/lumpy-sv_1554621096029/work//bin/lumpy
HEXDUMP=/usr/bin/hexdump
SAMBLASTER=
SAMBAMBA=
SAMTOOLS=/exports/sascstudent/cedrick/conda/envs/vcf_parser/bin/samtools
PYTHON=/opt/conda/bin/python

PAIREND_DISTRO=/opt/conda/conda-bld/lumpy-sv_1554621096029/work//scripts/pairend_distro.py
BAMGROUPREADS=/opt/conda/conda-bld/lumpy-sv_1554621096029/work//scripts/bamkit/bamgroupreads.py
BAMFILTERRG=/opt/conda/conda-bld/lumpy-sv_1554621096029/work//scripts/bamkit/bamfilterrg.py
BAMLIBS=/opt/conda/conda-bld/lumpy-sv_1554621096029/work//scripts/bamkit/bamlibs.py

```

PR #14396, opened 2019-04-09
Fix lumpyexpress config.

rhpvorderman:
Apparently the lumpy express config file contains fixed values from the build environment
See #14372 

This fix makes sure these values are variable after installation. This should work in any environment, including the container.

----------------


arton2008:
Hello, Did you find an answer to this question? I'm having a similar problem when using a conda environment for lumpy. I understand you have made some updates to the config file to correct this problem but I can't find a file to download on this page.

PR #14423, opened 2019-04-10
Draft welcome message

dpryan79:
:information_source:

----------------

@epruesse Want to have a look?

### DIFF
--- a/.github/welcome_new_contributor.md
+++ b/.github/welcome_new_contributor.md
@@ -1,0 +1,12 @@
+Thanks for opening your first pull request! As you're new to bioconda, please read through our [documentation about contributing recipes](https://bioconda.github.io/contributing.html). If you need help, please ping one of the following teams:
+
+ - `@bioconda/haskel` for Haskell packages
+ - `@bioconda/lib` for C/C++ packages
+ - `@bioconda/osx` for help with builds erroring on OSX
+ - `@bioconda/perl` for perl packages
+ - `@bioconda/python` for python packages
+ - `@bioconda/r` for R packages
+ - `@bioconda/rust` for Rust packages
+ - `@bioconda/core` in other cases
+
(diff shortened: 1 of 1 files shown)


epruesse:
Can non-members ping? I don't recall. Might be that you can only ping public teams.

dpryan79:
I’m honestly not sure, I can make a new account tonight and see. If not maybe the bot can ping us?

epruesse:
Yes - that's what I thought. If in doubt, the bot is a member and can issue pings.

epruesse:
In theory, we could also add API for gitter.im to have it present in the channel(s). I had briefly looked into that. Gitter.im seems to use a simple HTTP stream to send messages to listening clients. Messages are separated by `\r` not '\n' though, because '\n' may occur in the messages themselves. Would need some extra work. Not sure if I can get around to that.

dpryan79:
I can confirm that non-members cannot ping us (see https://github.com/bioconda/bioconda-recipes/pull/14427 where I made a new account). So we'll have to have the bot do that, which is quite annoying.

epruesse:
Did you check whether it works when you make the team public? IIRC its about visibility.

dpryan79:
It's already visible. I just looked and the settings for that do mention that only organization members can ping it, even if it is visible.

epruesse:
:-(

epruesse:
You know ... the bot actually looks at each and every comment anyway. It could grep for the pings and just repost them if the user isn't a member.

epruesse:
Just need to get it to be stable enough to actually run all the time...

dpryan79:
Just doing a grep for a ping seems reasonable, we can restrict that it has to be at the beginning of the comment to limit false positives.

epruesse:
The hurdle is mainly determining who's a member and who is not. Not sure if I should load the list at startup and try to keep it maintained, or whether I should make calls every time and hope that the API caching (on the bot side) keeps the requests down to a reasonable limit. Need that anyway for the welcome message though.

dpryan79:
I'd suggest just loading the list once a day. That's often enough.

epruesse:
I'm merging this. Hope I get around to using it soon.

PR #14426, opened 2019-04-10
Crispr studio patch

KyleL1998:
:information_source:

----------------

The aspect that I removed does not work when its implemented into a wrapper. Leaving it in causes the flag `-f` to no longer work. Patching out the problem area brings partial functionality back to `-f` so removing the problem code until it can be changed


recipes/CRISPRStudio/change_f_flag.patch
```diff
@@ -0,0 +1,20 @@
+--- CRISPR_Studio_1.0.py.orig	2019-03-12 13:43:36.644438096 -0500
++++ CRISPR_Studio_1.0.py	2019-04-10 13:12:03.262862265 -0500
+@@ -771,15 +771,8 @@
+         writeFasta(spacerDict, outFasta)
+         checkSpacer(spacerDict, averageLength, outFasta)
+     elif args.checkFasta:
+-        print('\n\tTHE FLAG -f (--checkFasta) WAS USED, NO VERIFICATION OF THE FASTA FILE WILL BE RUN')
+-        print('\tIf you want CRISPR_Studio to extract the spacers from the gff file')
+-        print('\twithout runing the verification step, just hit enter (leave the field empty)')
```
pvanheus: typo: runing should be running

KyleL1998: That's their code which I'm removing @pvanheus no point for me to fix that. I will go add the typo to the issue I raised for them though.

PR #14435, opened 2019-04-11
Adding mark-nonconverted-reads

mattsoup:
:information_source:

----------------

Thanks for your help!


mattsoup:
@bioconda/core I think this is ready to go.  It would also be great if I could be added as a member here. Thanks!

recipes/mark-nonconverted-reads/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+requirements:
+  host:
+    - python >=3
+
+  run:
+    - python >=3
+    - pysam
+
+build:
+  noarch: python
```
dpryan79: Can you add a build number (0)? That makes upgrades easier.

dpryan79:
I've added you to the team, go ahead and merge after adding a build number! Thanks

recipes/mark-nonconverted-reads/meta.yaml
mattsoup: Done!

PR #14495, opened 2019-04-16
Add r-stitch 1.5.6

rwdavies:
:information_source:

----------------

### DIFF
--- a/recipes/r-stitch/meta.yaml
+++ b/recipes/r-stitch/meta.yaml
@@ -11,7 +11,7 @@ source:
 
 build:
   number: 0
-  script: 'ls && mkdir -p temp && cd temp && mv ../STITCH_1.5.6.tar.gz ./ && tar -zxvf STITCH_1.5.6.tar.gz && sed ''s/configure/configure CXX="$(CXX)"/g'' STITCH/src/Makevars > STITCH/src/Makevars2 && mv STITCH/src/Makevars2 STITCH/src/Makevars && rm STITCH_1.5.6.tar.gz && tar -czvf ../STITCH_1.5.6.tar.gz * && cd .. && echo hello! && $R CMD INSTALL --build --install-tests .'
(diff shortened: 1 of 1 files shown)


recipes/r-stitch/meta.yaml
```diff
@@ -11,7 +11,7 @@ source:
 
 build:
   number: 0
-  script: '$R CMD INSTALL --build --install-tests .'
+  script: 'ls && echo now showing original Makevars && cat src/Makevars && sed ''s/configure/configure CXX="$(CXX)"/g'' src/Makevars > src/Makevars2 && mv src/Makevars2 src/Makevars && echo now showing modified Makevars && cat src/Makevars && $R CMD INSTALL --build --install-tests .'
```
dpryan79: Can you put this in `build.sh`? It'll be easier to modify in a regular shell script.

dpryan79:
You can probably copy `build.sh` from some of the bioconductor recipes. That will properly inject compilers so it stops trying to use `gcc`.

recipes/r-stitch/meta.yaml
rwdavies: Hey, sorry about this. I'm trying to understand how the compiler information is being passed through from the environment to the compilation, and since I can't replicate this clean environment on my own machine, I've taken to the above horrible line of code (or variants thereof). Once I properly confirm the problem and can correct it, I intend to patch my own package and release a new version that should use the original recipe with it's very simple installation

dpryan79: Ah, gotcha :)

For R packages I've found this to work well: 

```
#!/bin/bash
mkdir -p ~/.R
echo -e "CC=$CC
FC=$FC
CXX=$CXX
CXX98=$CXX
CXX11=$CXX
CXX14=$CXX" > ~/.R/Makevars
$R CMD INSTALL --build .
```

rwdavies: It's similar to that, but I need to pass through the information to the stuff that is external and also gets compiled (htslib and seqlib)
I've got to run to the vet, but the thing I'm struggling with is this
https://github.com/rwdavies/STITCH/blob/master/STITCH/src/Makevars
the original error message on the original commit above leaves me to believe I'm not passing the R CPP information properly through to the configure / subsequent compilation below, so I need something like CPP=$(shell "R CMD config CPP"), but this is giving me non-sensical error messages on my machine so I know there must be a typo somewhere... argh...

dpryan79: I wish R made this sort of thing more convenient.

rwdavies:
Closing this now as I fixed the problem upstream in the package itself and have pushed a new version of the package which I'm now adding to bioconda

PR #14555, opened 2019-04-22
Add recipe for HMMER3 3.2.1

davebx:

This PR adds a recipe for HMMER3 version 3.2.1, following the pattern of previous recipes by appending '3' to the executables' filenames.


epruesse:
This looks recundant with `hmmer`

epruesse:
The `hmmer2` package renames the binaries. Do we need another copy? I'm also not sure why one would install multiple versions of hmmer into a single env. You can have multiple envs easily, so what drives the need here?

davebx:
Apparently not. Serves me right for assuming the absence of a three meant a one.

PR #14603, opened 2019-04-22
Remove build-kite

epruesse:
@johanneskoester Do we still use / need this somewhere?


bgruening:
No, and if we fallback to it we need to update it either way. My guess is the next things we could try is Azure-Pipelines.

PR #14604, opened 2019-04-22
Move multiphate to recipes folder

epruesse:
@bgruening @carolzhou I somehow doubt this was meant to go into toplevel folder.


epruesse:
Are you guys going to fix the recipe, or should I just remove it?

dpryan79:
I moved and blacklisted this during the last bioconductor update. We should probably just delete it though since it never worked.

PR #14615, opened 2019-04-23
Mlpy

jfallmann:
:information_source:

----------------

### DIFF
--- a/recipes/mlpy/meta.yaml
+++ b/recipes/mlpy/meta.yaml
@@ -30,8 +30,8 @@ requirements:
     - gsl
 
 test:
-  commands:
-    - mlpy -h
+  importss:
+    - mlpy
 
 about:
   home: http://mlpy.sourceforge.net/index.html
--- a/recipes/mlpy/meta.yaml
+++ b/recipes/mlpy/meta.yaml
@@ -12,7 +12,8 @@ source:
 
 build:
   number: 0
-
+  skip: true  # [osx]
+  
 requirements:
   build:
     - {{ compiler('cxx') }}
--- a/recipes/mlpy/meta.yaml
+++ b/recipes/mlpy/meta.yaml
@@ -18,13 +18,13 @@ requirements:
   build:
     - {{ compiler('cxx') }}
   host:
-    - python >=2.6
+    - python >=2.6,<3.7
     - setuptools
     - numpy >=1.3.0 #,<1.6
     - scipy >=0.7.0 #,<0.12
     - gsl >=1.11 #,<1.17
   run:
-    - python >=2.6
+    - python >=2.6,<3.7
     - numpy >=1.3.0 #,<1.6
     - scipy >=0.7.0 #,<0.12
     - gsl >=1.11 #,<1.17
--- a/recipes/mlpy/meta.yaml
+++ b/recipes/mlpy/meta.yaml
@@ -24,7 +24,7 @@ requirements:
     - scipy >=0.7.0,<=0.11
     - gsl >=1.11,<=1.16
   run:
-    - python >=2.6
+    - python
     - numpy >=1.3.0,<=1.5
     - scipy >=0.7.0,<=0.11
     - gsl >=1.11,<=1.16
--- a/recipes/mlpy/meta.yaml
+++ b/recipes/mlpy/meta.yaml
@@ -12,7 +12,7 @@ source:
 build:
   number: 0
   skip: true  # [osx]
-  script: python setup.py install
+  script: "{{ PYTHON }} setup.py install"
   
 requirements:
   build:
--- a/recipes/mlpy/meta.yaml
+++ b/recipes/mlpy/meta.yaml
@@ -35,7 +35,7 @@ test:
 about:
   home: http://mlpy.sourceforge.net/index.html
   license: GPL3
-  license_file: mlpy-{{ version }}/LICENSE.txt
+  license_file: LICENSE.txt
   summary: mlpy is a Python module for Machine Learning built on top of NumPy/SciPy and the GNU Scientific Libraries.
   dev_url: http://hg.code.sf.net/p/mlpy/code/
   doc_url: http://mlpy.sourceforge.net/docs
--- a/recipes/mlpy/meta.yaml
+++ b/recipes/mlpy/meta.yaml
@@ -19,16 +19,15 @@ requirements:
     - {{ compiler('cxx') }}
   host:
     - python >=2.6 #,<3.7
-    - cython
     - setuptools
     - numpy >=1.3.0 #,<1.6
     - scipy >=0.7.0 #,<0.12
-    - gsl >=1.11 #,<1.17
+    - gsl >=1.11,<1.17
   run:
     - python >=2.6 #,<3.7
     - numpy >=1.3.0 #,<1.6
     - scipy >=0.7.0 #,<0.12
-    - gsl >=1.11 #,<1.17
+    - gsl >=1.11,<1.17
 
 test:
   imports:
--- a/recipes/mlpy/meta.yaml
+++ b/recipes/mlpy/meta.yaml
@@ -20,14 +20,14 @@ requirements:
   host:
     - python >=2.6
     - setuptools
-    - numpy >=1.3.0,<=1.5
-    - scipy >=0.7.0,<=0.11
-    - gsl >=1.11,<=1.16
+    - numpy >=1.3.0,<1.6
+    - scipy >=0.7.0,<0.12
+    - gsl >=1.11,<1.17
   run:
     - python >=2.6
-    - numpy >=1.3.0,<=1.5
-    - scipy >=0.7.0,<=0.11
-    - gsl >=1.11,<=1.16
+    - numpy >=1.3.0,<1.6
+    - scipy >=0.7.0,<0.12
+    - gsl >=1.11,<1.17
 
 test:
   imports:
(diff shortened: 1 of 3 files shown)


jfallmann:
According to the source  (http://mlpy.sourceforge.net/docs/3.5/install.html) it is necessary, gcc is also used during the build.

recipes/mlpy/meta.yaml
```diff
@@ -0,0 +1,39 @@
+{% set version = "3.5.0" %}
+{% set sha256 = "344fa75fbf9f76af72f6a346d5309613defc4d244bac13c218e509a51d68bf6a" %}
+#source file is named download, has to be unpacked manually, update build sh accordingly when updating version
+
+package:
+  name: mlpy
+  version: {{ version }}
+
+source:
+  url: https://sourceforge.net/projects/mlpy/files/mlpy%203.5.0/mlpy-{{ version }}.tar.gz/download
```
epruesse: Try this without the `/download` part. `wget` won't work, but `conda` should be able to download. It should then get unpacked correctly allow you to get away without the build.sh. Look at other recipes or the python skeleton for the right line to call `setup.py`.

recipes/mlpy/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+#  skip: True # [py27]
```
epruesse: If this is no longer correct, remove it.

recipes/mlpy/meta.yaml
```diff
@@ -0,0 +1,39 @@
+{% set version = "3.5.0" %}
+{% set sha256 = "344fa75fbf9f76af72f6a346d5309613defc4d244bac13c218e509a51d68bf6a" %}
+#source file is named download, has to be unpacked manually, update build sh accordingly when updating version
```
epruesse: See below

recipes/mlpy/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+build:
+  number: 0
+#  skip: True # [py27]
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+  host:
+    - python >=2.6
```
epruesse: If you can make this `noarch: python`, you need to use the `skip` line instead of the constraint here. (Don't ask, it's how conda likes it)

recipes/mlpy/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - setuptools
+    - numpy =1.5
+    - scipy =0.11
+    - gsl =1.16
```
epruesse: Do these need to be pinned precisely? Can you use `>=` or at least `1.5.*`? It's easier to get tools to cooperate in an env if there is some leeway on what they require.

recipes/mlpy/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+  run:
+    - python
+    - numpy
+    - scipy
+    - gsl
```
epruesse: If you research results in that the packages have to be pinned, they should be pinned here as well.

recipes/mlpy/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+test:
+  importss:
```
epruesse: One s too many

recipes/mlpy/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - mlpy
+
+about:
+  home: http://mlpy.sourceforge.net/index.html
+  license: GPL3
+  summary: mlpy is a Python module for Machine Learning built on top of NumPy/SciPy and the GNU Scientific Libraries.
```
epruesse: Would you like to add yourself as maintainer? We appreciate if people "adopt" packages, so that if there is an issue or an update to the package, there is someone we can ping who's already familiar with the software.

epruesse:
> According to the source (http://mlpy.sourceforge.net/docs/3.5/install.html) it is necessary, gcc is also used during the build.

From that:
```
Python >= 2.6 or 3.X
NumPy >= 1.3.0 (with header files)
SciPy >= 0.7.0
GSL >= 1.11 (with header files)
```

epruesse:
Also, `license_file` is mandatory for GPL - we have to ship a copy of the license with the binaries. The package seems to vendor a couple of libraries. It would be good if you could check into that. I know, it's a little tedious, but it's necessary.

recipes/mlpy/meta.yaml
jfallmann: Sorry, gcc is required, noarch not possible

recipes/mlpy/meta.yaml
jfallmann: This are the newest versions that worked for me, unless I can pin them >=x<=y I do not see another way

recipes/mlpy/meta.yaml
jfallmann: makes sense, will add

recipes/mlpy/meta.yaml
jfallmann: thx for the catch

recipes/mlpy/meta.yaml
jfallmann: I just built this as requirement for squigglekit, never used it myself, but if nobody else is willing I can add myself

recipes/mlpy/meta.yaml
jfallmann: Does not work for me:
An HTTP error occurred when trying to retrieve this URL.
HTTP errors are often intermittent, and a simple retry will get you on your way.
Could not download https://sourceforge.net/projects/mlpy/files/mlpy%203.5.0/mlpy-3.5.0.tar.gz

recipes/mlpy/meta.yaml
jfallmann: Done, thx

recipes/mlpy/meta.yaml
jfallmann: solved

recipes/mlpy/meta.yaml
epruesse: Yeah, that's the way it usually goes. You end up being maintainer for most of the dependency tree of the things you want on here. I've even got X11 stuff over on CF on my list for that reason.

recipes/mlpy/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+
+test:
+  imports:
+    - mlpy
+
+about:
+  home: http://mlpy.sourceforge.net/index.html
+  license: GPL3
+  license_file: LICENSE.txt
```
epruesse: Not the right file. Or not the only one required. :-/ It says
```
mlpy: See gpl-3.0.txt
libsvm: See mlpy/libsvm/libsvm/COPYRIGHT
liblinear: See mlpy/liblinear/liblinear/COPYRIGHT
fastcluster: See gpl-3.0.txt
```

That's a bit of a mess. The ideal option would be to unbundle those and use the packages. At least fastcluster and libsvm have packages on conda-forge. I'm not sure about liblinear.

Alternatively, you'd probably have to compose a license file manually and add it to the recipe.

epruesse:
Sorry that this is a complicated one. I've spent a few months before I could get SINA into Bioconda, first building a bunch of libs including X11 stuff on CF, then adding ARB (massive package), and then finally getting to add SINA.

recipes/mlpy/meta.yaml
jfallmann: Well, one for team it is then ;)

recipes/mlpy/meta.yaml
jfallmann: Thx for the tip, wasn't aware of that, quite useful

recipes/mlpy/meta.yaml
jfallmann: License is fixed I hope, compiled a collection, they do not ship the COPYRIGHT files as they claim, but this is an upstream problem IMHO, also version ranges seem to work, thanks again for that neat trick

jfallmann:
@epruesse  please wait before merging, trying to fix a dependency clash with the a package depending on this one first, sorry for the premature review request

epruesse:
Ok. Ping me again when you are ready.

jfallmann:
@epruesse Ready to merge now, sorry for the delay!

jfallmann:
ping @epruesse

epruesse:
Guys ... 405 files changed and you call it ready to merge? ;-)

jfallmann:
Oh what have I done, rebased to master, shouldn't that have done the trick?

recipes/mlpy/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  build:
+    - {{ compiler('cxx') }}
+  host:
+    - python >=2.6,<3.7
+    - setuptools
+    - numpy >=1.3.0 
+    - scipy >=0.7.0 
+    - gsl >=1.11 
+  run:
+    - python >=2.6
```
epruesse: also <3.7 I presume

epruesse:
There. I repaired it. Take your local branch and rename it (by creating another branch from it and then deleting it), then go and pull this one. Sorry. Rebase can be dangerous, and it's pretty hard to fix from somewhere else. This was the best I could do.

epruesse:
@biocondabot recheck

epruesse:
@biocondabot recheck

epruesse:
@biocondabot recheck

epruesse:
@BiocondaBot recheck

epruesse:
@biocondabot recheck

epruesse:
@biocondabot lint

PR #14620, opened 2019-04-23
Updated blast to version 2.9.0

eggzilla:
:information_source:

----------------


epruesse:
I don't think we need to keep a copy of 2.7.1

epruesse:
I think we can also skip on 2.8 (https://github.com/bioconda/bioconda-recipes/pull/14066) and go right to 2.9

epruesse:
Relevant quote from NCBI:

> I looks like they’re having problems with the GnuTLS library, but BLAST+ moved away from this library over a year ago. We now use mbedTLS (courtesy of the SRA toolkit IIRC). As for build issues in the core C++ toolkit, I’d refer you to cpp-core@ncbi.nlm.nih.gov.

eggzilla:
Hi @epruesse , thanks for your feedback! I will remove 2.7.1 and skip 2.8.1. I had a problem with the PATH variable set in the get_species_taxids.sh script which I would patch. Furthermore the script requires entrez-direct, should I add that to the recipe dependencies? Thanks :-)

epruesse:
How how much does the size of the env increase by adding entrez-direct and its dependencies? BLAST is a very frequent dependency, so it adds to a lot of environments. If it's small add it, if not, add something to the package description that says that to use that script you need a another package. Technically you could use `outputs` to have `blast` and `blast-extras` or something like that, so that people can have both the "all dependencies" package and the "I just want to `blastn`" package.

epruesse:
(We can look into those options once the build completes)

eggzilla:
@dpryan79 
Hi, thanks a lot for updating the recipe! Sorry for the delay I tested the remote option and it is not working for me. I used following testsequence:
`>testprotein
GGGGGGGGGGGGGGGGGGGGGAAAGGA
`
I get this error message:
Error: NCBI C++ Exception:
    T0 "/opt/conda/conda-bld/blast_1558698596330/work/c++/src/objects/seq/../seqalign/Score_.cpp", line 90: Error: CScore::C_Value::GetInt(): Invalid choice selection: NCBI-Seqalign::Score.value.real

dpryan79:
That looks like a blast bug at least.

eggzilla:
I will test offline mode in a moment

pmenzel:
Offline blastn works fine for me and produces the same output as blast 2.7.1 on my example cases.

dpryan79:
Thanks @pmenzel, that's reassuring!

PR #14681, opened 2019-04-25
Add recipe for gbmunge

npavlovikj:
:information_source:

----------------

### DIFF
--- a/recipes/gbmunge/makefile.patch
+++ b/recipes/gbmunge/makefile.patch
@@ -1,0 +1,30 @@
+--- Makefile.old	2019-04-24 17:17:58.618580369 -0500
++++ Makefile	2019-04-24 17:23:40.137613637 -0500
+@@ -1,5 +1,13 @@
+-all:
++PREFIX :=
++EXECUTABLE := gbmunge 
++
++all: $(EXECUTABLE)
++
++$(EXECUTABLE):
+ 	$(MAKE) -C src
+ 
+ clean:
+ 	$(MAKE) -C src clean
++
++install:
++	cp src/$(EXECUTABLE) $(PREFIX)/bin/
+--- Make.inc.old	2019-04-24 17:15:43.499565380 -0500
++++ Make.inc	2019-04-24 17:16:10.072568329 -0500
+@@ -10,10 +10,8 @@
+ 
+ ifeq ($(USECLANG),1)
+ USEGCC = 0
+-CC = clang
+ endif
+ 
+ ifeq ($(USEGCC),1)
+-CC = gcc
+ CFLAGS-add += -D__USE_XOPEN -D_GNU_SOURCE
+ endif
--- a/recipes/gbmunge/meta.yaml
+++ b/recipes/gbmunge/meta.yaml
@@ -1,0 +1,32 @@
+{% set name = "gbmunge" %}
+{% set version = "2018.07.06" %}
+{% set sha256 = "1297e73be293a7b98bc080faeb191737b0ad4cc5ca3a9ec2564c9b82e95cc014" %}
+{% set git_rev = "4117e04da03a0dad41417397c623dc5e995794df" %}
+
+package:
+  name: {{ name|lower }}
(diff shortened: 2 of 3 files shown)


recipes/gbmunge/makefile.patch
```diff
@@ -0,0 +1,30 @@
...
++++ Make.inc	2019-04-24 17:16:10.072568329 -0500
+@@ -10,10 +10,8 @@
+ 
+ ifeq ($(USECLANG),1)
+ USEGCC = 0
+-CC = clang
+ endif
+ 
+ ifeq ($(USEGCC),1)
+-CC = gcc
```
epruesse: You can achieve the same by calling

```
make CC=$CC
```

Variables set on the `make` commandline override variables set inside of the `makefile` (even if there is `CFLAGS += -bla` going on!).

recipes/gbmunge/makefile.patch
```diff
@@ -0,0 +1,30 @@
...
++all: $(EXECUTABLE)
++
++$(EXECUTABLE):
+ 	$(MAKE) -C src
+ 
+ clean:
+ 	$(MAKE) -C src clean
++
++install:
++	cp src/$(EXECUTABLE) $(PREFIX)/bin/
```
epruesse: Can't you just use `cp src/gbmunge $PREFIX/bin` instead of `make install` in the `build.sh`?

recipes/gbmunge/meta.yaml
```diff
@@ -0,0 +1,32 @@
+{% set name = "gbmunge" %}
+{% set version = "2018.07.06" %}
+{% set sha256 = "1297e73be293a7b98bc080faeb191737b0ad4cc5ca3a9ec2564c9b82e95cc014" %}
+{% set git_rev = "4117e04da03a0dad41417397c623dc5e995794df" %}
```
epruesse: Add 
```
# Requested tag from upstream in https://github.com/sdwfrost/gbmunge/issues/3
```
```

File: recipes/gbmunge/meta.yaml [hunk 4]
```diff
@@ -0,0 +1,32 @@
+{% set name = "gbmunge" %}
+{% set version = "2018.07.06" %}
```
epruesse: If you created this version, use something like `0.0.2018.07.06`. That way it's sure to be less than whatever versioning scheme upstream might decide in the future. If they release `1.0`, well, `1.0 < 2018.07`, so it won't get installed. (At least not without setting epoch, which I've never tried on conda).

recipes/gbmunge/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/sdwfrost/gbmunge/archive/{{ git_rev }}.zip
+  sha256: {{ sha256 }}
+  patches:
+    - makefile.patch
```
epruesse: Just a suggestion, but if you do the `cp` in the `build.sh` and use `make CC=$CC` you can probably get away without the patch file.

recipes/gbmunge/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+test:
+  commands:
+    - gbmunge -h
+
+about:
+  home: https://github.com/sdwfrost/gbmunge
+  license: MIT
+  license_family: MIT
+  license_file: LICENSE
+  summary: "Munge GenBank files into FASTA and tab-separated metadata."
```
epruesse: Would you like to add yourself as a maintainer here?

npavlovikj:
Many thanks for the feedback @epruesse , I really appreciate it!

The updated recipe is here, https://github.com/bioconda/bioconda-recipes/pull/14702.
The only thing I didn't change is the version - if I put "0.0.2018.07.06" now, "2018.07.06" will still be picked up.
If the developer pushes tagged release, I will change the version and we can remove the current package then.

PR #14712, opened 2019-04-25
Scaden 0.9.0

KevinMenden:
:information_source:

----------------

This is my first recipe on Bioconda, so I could need a review.

Local tests were successful.

### DIFF
--- a/recipes/scaden/meta.yaml
+++ b/recipes/scaden/meta.yaml
@@ -21,16 +21,21 @@ requirements:
     - twine
   run:
     - python
-    - pandas ==0.21
+    - pandas
     - numpy
     - scikit-learn
     - scipy
     - seaborn
-    - tensorflow ==1.10.0
+    - tensorflow <=1.13.0
     - matplotlib
     - scanpy
     - tqdm
     - click
+
+# tests fail because of tesnorflow, commenting out
+#test:
+#  commands:
+#    - scaden --help
 
 about:
   home: https://github.com/KevinMenden/scaden
--- a/recipes/scaden/meta.yaml
+++ b/recipes/scaden/meta.yaml
@@ -1,16 +1,13 @@
 {% set name = "scaden" %}
 {% set version = "0.9.0" %}
-{% set file_ext = "tar.gz" %}
-{% set hash_type = "sha256" %}
-{% set hash_value = "c5c3c58aa9a9c36151dcbd746445ed038fbc33e7906529a2bbff1d84dcfb7239" %}
 
 package:
   name: '{{ name|lower }}'
   version: '{{ version }}'
 
 source:
-  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.{{ file_ext }}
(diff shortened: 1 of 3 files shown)


KevinMenden:
Trying to ping @bioconda/core :)

recipes/scaden/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - click
+
+about:
+  home: https://github.com/KevinMenden/scaden
+  license: MIT
+  summary: Cell type deconvolution using single cell data
+
+extra:
+  skip-lints:
+    - missing_tests
```
epruesse: - Add a comment explaining that tests are disabled because of tensor flow problem. 
- Also add some tests that work for you locally (commented out), so that we have something to enable once the tensorflow libc problem has been fixed.
- Would appreciate if you could add yourself as maintainer.

recipes/scaden/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  noarch: python
+
+requirements:
+  host:
+    - python
+    - pip
+    - twine
+  run:
+    - python
+    - pandas ==0.21
```
epruesse: Must be this tight? No range?

recipes/scaden/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - numpy
+    - scikit-learn
+    - scipy
+    - seaborn
+    - tensorflow ==1.10.0
```
epruesse: Same as above - can this be a range?

epruesse:
@bioconda/core - tensorflow from conda-forge needs a newer libc than what we have in the docker. Might be necessary to have another "special" testing container. Or update the libc in the one we have so that these things can run tests.

kyleabeauchamp:
Long term we'd like to build tensorflow from source in conda-forge, with better glibc compatibility.  Short term: it's not likely to happen soon.

epruesse:
Yep. We have the "special" container for click & co. Wonder if that would help.

recipes/scaden/meta.yaml
KevinMenden: I got some issues with pip with different pandas version and fixed it, but I'm sure it can also be a range. I'll try it out again and see what works.

recipes/scaden/meta.yaml
KevinMenden: Yes, it just needs to be below 1.13.0 actually, so I'll change that too.

KevinMenden: Yes it actually can be a range, just has to be below tensorflow 2. I'll change it accordingly.

recipes/scaden/meta.yaml
KevinMenden: Added a comment about that in new commit.
Also added one test that at least locally worked (scaden --help).

epruesse: Personally, I think the `--help` type test is completely insufficient. But then we have a lot of precedent for those. ;)

Issue #14466, opened 2019-04-13
Trinity: need to remove `@` from paths in perl script


bernt-matthias:
Happens within Galaxy which uses the @ character in the environment names.

Error is: 

```
Possible unintended interpolation of @2 in string at /gpfs1/data/galaxy_server/galaxy-dev/database/dependencies/_conda/envs/__trinity@2.8.4/bin/Trinity line 46.
```

Had the same problem in the stacks2 wrapper. For variables setting path to binaries that are on the PATH anyway the solution is to set them to an empty string. 

For the others (like paths to jar files), a fix might involve using some `pwd` magic.

epruesse:
Is this a Bioconda issue (packaging) or an upstream problem at Trinity (software)?

ping @bgruening - if this is galaxy specific, maybe you know who could help.

bernt-matthias:
Kind of both: The cause of the problem is that galaxy uses an `@` character in the environment names (`__name@version`). 

If perl packages/programs use hard coded paths (e.g. via an install script that uses BIN_DIR as parameter) for dependencies, then the perl interpreter has problems in this case since the @ should be quoted. 

One solution is to remove the hard coded paths (which I did for stacks) since the binaries are on the PATH anyway. Alternatively the @ could be quoted. Don't know if conda could do this in general for BIN_DIR (or all its path).

epruesse:
So basically, you are saying that `Trinity`'s perl code breaks if Trinity is installed into a directory that has an `@` in it? That sounds like something that needs to be fixed in Trinity. Galaxy is doing something unusual, but nothing wrong. Neither is Conda when it rewrites the paths. How does the proper Perl Quoting look like? It might be something to add to `conda` - rewrite the strings with correct quoting if it's a perl file.

bernt-matthias:
I'm no perl expert, but `\@` should do the trick.

epruesse:
@msarahan Can you pitch in with suggestions here?

From `bin/Trinity`:
```
#!/usr/bin/env perl
[...]
my $TRIMMOMATIC = "/opt/anaconda1anaconda2anaconda3/share/trimmomatic/trimmomatic.jar";
```

So installing into an env with `$` or `@` in it breaks the script during path rewriting.

Do you have code handling this for bash? Is there a chance to add code to `conda` to escape the `@` for perl files?

msarahan:
no, I don't think we have anything for that currently.  A PR to add it would be welcome.

[...]

PR #14741, opened 2019-04-26
trinity: fix for installation paths containing @

bernt-matthias:
https://github.com/bioconda/bioconda-recipes/issues/14466

:information_source:

----------------

### DIFF
--- a/recipes/trinity/build.sh
+++ b/recipes/trinity/build.sh
@@ -50,7 +50,7 @@ find $TRINITY_HOME -type f -print0 | xargs -0 sed -i.bak 's/FindBin::Bin/FindBin
 
 # Replace hard coded path to deps
 find $TRINITY_HOME -type f -print0 | xargs -0 sed -i.bak 's/$JELLYFISH_DIR\/bin\/jellyfish/jellyfish/g'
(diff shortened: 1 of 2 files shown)


epruesse:
Just saw that Trinity doesn't even add hard coded paths. It's installed into `/opt` by the package script and the way the paths are fixed isn't good. This should work. $ROOTDIR points to the real location of the `Trinity` perl script. Since that's installed into `$PREFIX/opt/trinity`, `$ROOTDIR/../../` is the env root. From there, Trimmomatic is in `share/trimmomatic`. This will work with any env name, including "Wh@t's thi$s me()[]s?".

epruesse:
It won't compile because it's resetting the compiler internally:
```
Using gnu compiler for Inchworm and Chrysalis
```

That probably needs to be overridden. And then tested. Can you check into it @bernt-matthias?

dpryan79:
I think this was recently done in another PR. Can you confirm whether this is still needed?

bernt-matthias:
Seems that I forgot about my own PR https://github.com/bioconda/bioconda-recipes/pull/16852 :)

dpryan79:
You know you're involved in a lot of things when... :)

Issue #8869, opened 2018-05-09
Preferred solution for dealing with openmp?


aaronk:
I'm putting together a recipe for a C tool from which OpenMP would be difficult to extricate.  Linux builds are fine, but OSX is a problem.

I can add llvm-openmp to meta.yml, but am not sure of the proper way to include the path for omp.h.  In my test conda setup, it's under lib/clang/4.0.1/include and I can certainly add that to my CFLAGS, but it seems like a brittle solution.

epruesse:
@bioconda/core - Anyone have opinions on this? 

I've been simply using `gcc` instead of `llvm` to get openmp on macos.

daler:
@bioconda/lib might have some ideas?

acaprez:
From my *very brief* testing with an [OpenMP hello world](https://computing.llnl.gov/tutorials/openMP/samples/C/omp_hello.c) on OS X, using the new [Anaconda OS X compilers](https://conda.io/docs/user-guide/tasks/build-packages/compiler-tools.html) along with the `llvm-openmp` package creates a working OMP binary.  The compile line that worked was: `clang -Wl,-rpath /anaconda/envs/clang/lib  -L/anaconda/envs/clang/lib -fopenmp omp_hello.c`.

So you'll likely need to add something like `-Wl,-rpath ${PREFIX}/lib -L${PREFIX}/lib -fopenmp` to your build flags.

dpryan79:
For a number of recipes we've added `llvm-openmp` to both the `build:` and `run:` sections, which seems to suffice for this. We set a few thinks like `C_INCLUDE_PATH`, which are likely needed. That's probably the closest we have to a best-practice, which we should really document.

PR #14959, opened 2019-04-28
Add recipe for defiant

### LINKED ISSUE #13624: Defiant: Differential methylation: Easy, Fast, Identification and ANnoTation
Hi together,

it would be great if you could please add the tool "Defiant" to bioconda. It is a caller for differential methylation and outperformed every other DMR caller in our benchmark studies. 

https://github.com/hhg7/defiant

All the best,
Claudius

npavlovikj:
:information_source:

----------------

fixes #13624


epruesse:
`defiant.c:8:10: fatal error: 'omp.h' file not found`

epruesse:
https://github.com/bioconda/bioconda-recipes/issues/8869

epruesse:
@npavlovikj Can you update your master branch and then rebase the defiant branch? It's failing because there was a change to the master's CI configuration (well, because I changed it...).

npavlovikj:
@epruesse , I think I fixed the osx build to work with `-fopenmp`, so I would appreciate if you can review the recipe when you have a chance.

recipes/defiant/build.sh
```diff
@@ -0,0 +1,23 @@
...
+export LIBRARY_PATH="${PREFIX}/lib"
+export LD_LIBRARY_PATH="${PREFIX}/lib"
+
+mkdir -p ${PREFIX}/bin
+
+unzip defiant.zip
+
+# Fix defiant.c for OSX
+if [ `uname` == Darwin ]; then
+    sed -i.bak 's/restrict argv\[\]/argv\[\]/g' defiant.c
```
epruesse: This looks weird. I’ll have to look at the code for context.

recipes/defiant/build.sh
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+# Compile binaries
+${CC} -O4 -o defiant defiant.c -Wall -pedantic -std=gnu11 -lm -fopenmp -I${PREFIX}/include
+mv regions_of_interest regions_of_interest.c
+${CC} -o roi regions_of_interest.c -lm -Wall -std=gnu11 -Wextra -pedantic -Wconversion
+
```
epruesse: Use CFLAGS and LDFLAGS as well if it works. They some optimization flags that can be useful.

epruesse:
I’ll look more closely tomorrow

recipes/defiant/build.sh
npavlovikj: If I run the `gcc` command as it is, I am getting the following error:
**$ ${CC} -O4 -o defiant defiant.c -Wall -pedantic -std=gnu11 -lm -fopenmp -I${PREFIX}/include**
clang-4.0: warning: -O4 is equivalent to -O3 [-Wdeprecated]
In file included from defiant.c:12:
./read_bisulfite_data.h:99:25: warning: equality comparison with extraneous parentheses [-Wparentheses-equality]
        if ((number_of_columns == 1)) {// && (strstr(temporary_char_string,"chr") != NULL)) {
             ~~~~~~~~~~~~~~~~~~^~~~
./read_bisulfite_data.h:99:25: note: remove extraneous parentheses around the comparison to silence this warning
        if ((number_of_columns == 1)) {// && (strstr(temporary_char_string,"chr") != NULL)) {
            ~                  ^   ~
./read_bisulfite_data.h:99:25: note: use '=' to turn this equality comparison into an assignment
        if ((number_of_columns == 1)) {// && (strstr(temporary_char_string,"chr") != NULL)) {
                               ^~
                               =
defiant.c:36:5: error: second parameter of 'main' (argument array) must be of type 'char **'
int main (int argc, char *restrict argv[]) {
    ^
1 warning and 1 error generated.

I tried few things, but removing `restrict` from the definition of `main` worked. Please let me know if there are better ways of fixing this error.

npavlovikj:
Tomorrow sounds good Elmar - please go get some sleep now :)

recipes/defiant/build.sh
epruesse: No, that's correct.

Background:
`restrict` tells the compiler that the object a pointer references will not be referenced by any other pointer. That's relevant e.g. for something that copies data. E.g., if you do:
```
void copy(OBJ** source, OBJ** dest, int n) {
for (int i = 0; i <n; ++i) {
  dest[i] = source[i];
}
```
then the compiler can't know whether the regions overlap. You could be copying data within the same array, so the loop has to be executed in order. If you add the `restrict` keyword, it means that `source` and `dest` can't overlap, and any order of running the loop, vectorization etc. are ok ("pointer aliasing" is a keyword here). So in summary, `restrict` can make things faster, sometimes a tiny bit, sometimes a lot, depends.

For `main()` it is actually wrong. That's the "API" the OS calls to start a program and give it the command line parameters. It's declared as `int main(int, char**)`, and while some compilers allow you to change it, it won't have an effect on what the OS does. Plus - I don't see a point in it here. Even if it had a performance impact, how many command line arguments can there be?

So you are fine to patch this. Might want to use a patch instead of sed, to have context for people reading the recipes and not accidentally remove `restrict` where it might be needed correct. In this case, either is fine I guess.

You don't have to patch only on OSX. It's a fix for both - future GCC versions are not unlikely to complain (rightfully) just as CLANG is already doing.

epruesse:
`mv: cannot stat `regions_of_interest': No such file or directory`

recipes/defiant/build.sh
npavlovikj: Many thanks for the detailed explanation @epruesse - this totally makes sense why the fix worked!

I would prefer patch as a cleaner solution as well - just the first step of the build is to "unzip" "defiant.zip" that contains "defiant.c". Since the patch is applied before "unzip defiant.zip", the build fails because "defiant.c" is not found, so `sed` was a workaround...

Currently `regions_of_interest` is not part of "defiant.zip", but if the developer adds `regions_of_interest` in "defiant-1.1.4.zip" as requested in his repo's issue, then using patch will be feasible.

epruesse: Ah - didn't think of that. True.
Let's see if we get an updated release archive. If not, feel free to merge.

npavlovikj: Thanks for looking at this recipe @epruesse . The developer updated the tarball, so I went ahead and merged the recipe with the proper changes.

PR #14965, opened 2019-04-29
Store images

epruesse:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,6 +50,7 @@ variables:
       name: Copying packages
       when: always
       command: |
+        set +e  # Just to be safe. Don't want workflows to fail because of this.
         mkdir -p /tmp/artifacts/packages
         cd miniconda/conda-bld || exit 0
         find -name .cache | xargs rm -rf || true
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -50,6 +50,7 @@ variables:
       name: Copying packages
       when: always
       command: |
+        set +e  # Just to be safe. Don't want workflows to fail because of this.
```
mbargull: You'd rather still want `set -e` here, I guess, so that subsequent parts inside this `command: |` block don't get executed when previous ones failed. Just enclose everything in a subshell like so:
```yaml
command: |
  (
    mkdir -p /tmp/artifacts/packages
    ...
    fi
  ) || true
```

epruesse: Yeah, they do. Feel free to fix it. It requires every open PR to have a merge though if we mess with `.circleci/config.yml`. If you do, could you add a `echo -n "Compressing docker image $n..."; ...; echo "done"` to the for loop? I had meant to do that earlier. It takes >10 minutes for large (5GB) images to copy them, so knowing what it's doing would be good.

PR #14991, opened 2019-04-30
Cromwell to use the released jars.

rhpvorderman:
The original recipe did a lot of compiling and the resulting jar would not match the Jar as released by broadinstitute. 
This downloads those jars directly. It saves a lot of compile time and other hassle.
@micknudsen do you approve of this? Since you last updated cromwell.
@bioconda/core is this an okay way of packaging java packages?

----------------

### DIFF
--- a/recipes/cromwell/build.sh
+++ b/recipes/cromwell/build.sh
@@ -6,11 +6,10 @@ mkdir -p $outdir
 mkdir -p $PREFIX/bin
 
 cd $SRC_DIR
-sbt assembly
 
 # cromwell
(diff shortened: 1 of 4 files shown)


tomkinsc:
We aim to compile packages where it results in binaries with maximal compatibility, but for Java packages like Cromwell copying in the pre-built jar files makes sense. It's what bioconda does for GATK4.

micknudsen:
@rhpvorderman I think this is a great suggestion. For the sake of reproducibility, I think using the officially supported jar files from Broad is the way to go.

rhpvorderman:
All right then! Thanks for your input @tomkinsc and @micknudsen!

PR #14993, opened 2019-04-30
Updated pycoqc version

jfallmann:
:information_source:

----------------

### DIFF
--- a/recipes/pycoqc/meta.yaml
+++ b/recipes/pycoqc/meta.yaml
@@ -12,15 +12,14 @@ source:
 build:
   noarch: python
   number: 0
-  skip: true # [py2k]
   script: '{{ PYTHON }} -m pip install --no-deps --ignore-installed --no-cache-dir .'
 
 requirements:
   host:
-    - python
+    - python  >=3
     - pip
   run:
-    - python
+    - python  >=3
     - numpy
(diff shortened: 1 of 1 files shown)


recipes/pycoqc/meta.yaml
```diff
@@ -12,15 +12,14 @@ source:
 build:
   noarch: python
   number: 0
-  skip: true # [py2k]
   script: '{{ PYTHON }} -m pip install --no-deps --ignore-installed --no-cache-dir .'
 
 requirements:
   host:
-    - python
+    - python  >=3
```
micknudsen: @bgruening How does one decide which is the correct way? Compare e.g. with https://github.com/bioconda/bioconda-recipes/pull/14974/commits/2b21679fbbd33117b60eca69aa1ef8b40f40d468 where the change was made the other way.

jfallmann: I don't know but the original pycoQC states it's python3 only so I guess pinning to >=3 is correct?

druvus: @micknudsen noarch python doesn't work with selctors, see [cf documentation](https://conda-forge.readthedocs.io/en/latest/meta.html#building-noarch-packages)

PR #15007, opened 2019-05-01
bump vsearch and add gzip support

colinbrislawn:

This bumps vsearch to version 2.13.3 and (hopefully!) restores built-in gzip and bz2 support.


epruesse:
> This bumps vsearch to version 2.13.3 and (hopefully!) restores built-in gzip and bz2 support.

Probably doesn't. Adding to `build` is actually wrong. If you were e.g. using a Linux box to build for MacOS, then the compiler installed from `build` and the tools you need to build all would have to be Linux. But the things you link against, e.g. bzip, would have to be MacOS. Those go into `host`.

epruesse:
@colinbrislawn Can you come up with a tiny unit test that fails if compression is not supported? That way, this regression can be prevented from occurring again.

colinbrislawn:
What's the best way to include a compressed file for the unit test? I think my way might be... poor 😬

epruesse:
```yaml
requirements:
  build:
# these are the build tools. They "run" on the machine that build the package
    - '{{ compiler("c") }}'
    - '{{ compiler("cxx") }}'
    - automake
    - autoconf
  host:
# these are the host libraries. E.g. `libz.a` or `libz.so`. They are needed on the build 
# machine, but don't necessarily work there. They are from the target architecture.
    - zlib
    - bzip2
  run:
# Same as build, but this is the set needed at run time. So e.g. tools that you call that
# don't actually have to be present during build time.
    - zlib
    - bzip2
```

colinbrislawn:
Thank you for explaining build/host/run setup. You can tell I'm not a C developer, so this is very helpful.

This build is failing, but I'm not sure if the error is coming from my poor unit test, or a bad dependency setup.

Vsearch stores its test data in a separate repo. If I can't build the test data using `echo` and `gzip`, could I upload a test file from the bioconda recipe (insead of using a test file in the vsearch tarball)?

Thank you for all your help with this.

colinbrislawn:
>18:18:35 BIOCONDA INFO Recipes to lint:
recipes/augustus/3.2.3
recipes/vsearch

Can you help me figure out why `augustus` is also getting linted? Did I mess up my push?

epruesse:
I've been wondering whether it would be useful to have some shared test data for all of our recipes. Your initial approach, just echo something, looked fine to me, actually. Too much data in this repo isn't good because it makes every `clone` slower and slower. And there is some upper bound on Github repo size as well, IIRC.

epruesse:
It's because the blacklist on master has changed. Can you merge in master? That should fix it.

colinbrislawn:
>I've been wondering whether it would be useful to have some shared test data for all of our recipes.

This is a good set of data for nucleotides: https://github.com/torognes/vsearch-data 

I'll merge in master and try again.

colinbrislawn:
Current error:
>Fatal error: Unable to open file for reading (small.fna.gz)

Did I get this part wrong?
```yaml
test:
  files:
    - small.fna.gz
```

Does the test run in the root dir, or inside the tarball folder? If it runs from the folder, should I do this?
`vsearch --derep_fulllength ../small.fna.gz --output tmp.fna`

Sorry for all these questions. Clearly I'm unfamiliar with conda builds.

epruesse:
No that's fine. You got it right. Bioconda mulled tests just don't support everything. I even filed a bug for that ages ago: https://github.com/bioconda/bioconda-utils/issues/183

epruesse:
Your recipe is fine on the conda side. Bioconda-utils does one extra test: We take the recipe and install it into a clean docker container (via `mulled`, hence the term "mulled tests") and run the tests again. That ensures that our containers work, and that the package doesn't depend on things randomly present on the build host.

Catch is that the way those are run isn't 100% the same as done by conda-build. So it breaks. You ran into the `test: files:` not supported problem.

Options:
- revert to the `echo asd | gzip -c | test.fna.gz` approach (and hope the quoting doesn't kill the test)
- put the test into `run_tests.sh` - those don't get executed in the container, so they are fine

epruesse:
Err, `run_test.sh`

epruesse:
Also -- we all once were. Not so long ago for me. So don't worry. If you have time, have a look at our documentation and try to fit in your experiences for the next girl/guy.

colinbrislawn:
I feel like I should know this, but I'm stuck again:

Lint failed with this error:
```
Your bioconda-recipes CI configuration is out of date.
You can do this, e.g., by running:
  git fetch https://github.com/bioconda/bioconda-recipes.git master
  git merge FETCH_HEAD
``` 

So I did that, but I get two errors: 
```
error: Your local changes to the following files would be overwritten by merge:
...<lots of files>
error: The following untracked working tree files would be overwritten by merge:
...<even more files>
```

I appreciate your help with this. (If closing this and starting a new PR from master would be easier, we can try that too.)

epruesse:
@biocondabot recheck

epruesse:
There you go :)

colinbrislawn:
@epruesse Thank you for much for all your help! We did it!

>Dereplicating file small.fna.gz 100%
17:11:02 BIOCONDA INFO (OUT) 35 nt in 1 seqs, min 35, max 35, avg 35
Sorting 100%
17:11:02 BIOCONDA INFO (OUT) 1 unique sequences, avg cluster 1.0, median 1, max 1
Writing output file 100%
17:11:02 BIOCONDA INFO (OUT) Open a compressed .fna.gz file...PASS
17:11:02 BIOCONDA INFO (OUT) 
17:11:02 BIOCONDA INFO (OUT) ALL TESTS PASSED


>17:12:21 BIOCONDA INFO (ERR) [May 25 17:12:21] SOUT Compiled with support for gzip-compressed files, and the library is loaded.
17:12:21 BIOCONDA INFO (ERR) [May 25 17:12:21] SOUT Compiled with support for bzip2-compressed files, and the library is loaded.

PR #11815, opened 2018-11-07
CodingQuarry OSX Testing

kastman:
This PR is an independent test building on work started by @nextgenusfs in https://github.com/bioconda/bioconda-recipes/pull/10097 - simply my own testing that we'll pull into his original PR when the OSX version builds successfully.

PR #15012, opened 2019-05-02
fix ENV variable to QUARRY_PATH for codingquarry recipe, bump to build 1

nextgenusfs:
----------------

CodingQuarry tool requires an ENV variable to be set -- it is expecting $QUARRY_PATH, however looks like we set $QUARRYFILES in the first build.sh, bumped build number and set QUARRY_PATH.


nextgenusfs:
Looks like failed on macOS -- note sure we ever got it working due to compiler issues.  I've compiled on macOS using GCC manually.  Shall we list as linux only?

kastman:
Hi @nextgenusfs - We did end up getting OSX to run in #11815, and that was merged into master; what is currently in master built on OSX without problem when we pulled it in in Nov 2018, but it's showing a problem with openmp again now. 

I'm checking it out now; let's not toss it completely just yet.

kastman:
Looks like the main fix here was the `-I` include, right? Nice find! Good to merge?

nextgenusfs:
Yeah I think good to go. Adding llvm to the host seemed to be necessary as well.

kastman:
Yeah, I remember I added `llvm` but it wasn't sufficient; I think it needed both. Merged!

PR #15037, opened 2019-05-02
MicroHapDB 0.2 -> 0.3

standage:
Update MicroHapDB from 0.2 to 0.3

----------------

### DIFF
--- a/recipes/microhapdb/meta.yaml
+++ b/recipes/microhapdb/meta.yaml
@@ -1,5 +1,5 @@
-{% set version = "0.2" %}
-{% set sha256 = "450faabc3420fc7ed9107c0de3a0643b82352a592bde487c4afc6dc064bc8f7a" %}
+{% set version = "0.3" %}
+{% set sha256 = "8a34ff047640cc425ffc03dbfa9db1196003b3bc807825c1ab4e5316db9b470c" %}
 
 package:
   name: microhapdb


standage:
I would love to be added to @bioconda/core so I can manage this recipe.

standage:
@bioconda/core I forgot to increment the build number on this update, and thus MicroHapDB 0.3 isn't available from bioconda. Will opening a PR to fix this be sufficient to fix the problem, or will it require manual intervention?

dpryan79:
You shouldn't need to increment the build bumber, new versions should have build numbers starting at 0.

dpryan79:
Oh, well that's weird. It looks like this was uploaded, but for some reason it's not showing up. I'll have to dig into the nightly upload job to see why this isn't getting uploaded then.

standage:
> You shouldn't need to increment the build bumber, new versions should have build numbers starting at 0.

Sorry, n00b mistake. I misinterpreted the purpose of the build number, and assumed it was the problem since the package hadn't showed up.

epruesse:
> I would love to be added to @bioconda/core so I can manage his recipe.

/core privileges are really only for dealing with the infrastructure, you don't need that to become a package maintainer. You are welcome to add yourself to `extra: recipe-maintainers:` for MicroHapDB.

standage:
Thanks @epruesse. I misinterpreted the message at the end of https://github.com/bioconda/bioconda-recipes/blob/master/.github/welcome_new_contributor.md.

dpryan79:
I'll try to clarify that a bit https://github.com/bioconda/bioconda-recipes/pull/15167 :)

PR #15040, opened 2019-05-02
Microhapulator

standage:
New recipe for MicroHapulator 0.1.

----------------


standage:
Hi @bioconda/python, is there any way to mix bioconda dependencies and pip dependencies? I'm pretty sure that's the issue with my build failure: some of the packages this depends on are in bioconda, but some are only in PyPI.

standage:
InSilicoSeq is the only dependency not in conda/bioconda. I just opened up #15142 to address that.

epruesse:
@standage In theory, `conda` can install `pip` dependencies. In practice, we don't allow that. It can't be controlled well. Among other things, PyPI does not have dependency meta data. Each `setup.py` get's to decide on those on the fly based on code. So it's a feature from the early days that I'd consider deprecated for good reason.

standage:
Thanks @epruesse. I figured it would be better to have it in bioconda anyway. #15240

recipes/microhapulator/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  script: python -m pip install --no-deps --ignore-installed .
+  number: 0
+
+requirements:
+  host:
+    - python >=3
+    - pip
+  run:
+    - happer >=0.1
+    - insilicoseq >=1.3.5,!=1.3.5-py_0
```
standage: Yeah, I tried something like this earlier and also failed the linter.

epruesse: It's not the linter... it just breaks in `conda`.

epruesse: Sorry, I forgot about this one.

standage:
The problematic package just pushed a new bugfix release (in #15353). I'm going to give that several minutes to propagate through the system, and then push an update to this package's `meta.yaml` to depend on this new version.

That doesn't solve the problem that's been halting progress on this thread, but hopefully it should allow this package to finally build successfully.

standage:
AWESOME! Now I'm getting issues with the package tests. I can *do* something about that! 😀

standage:
Good grief, I'm glad the build is finally passing. Ready for review and merge!

epruesse:
Great! That sidesteps the issue!

Elmar

--
message sent from mobile device

recipes/microhapulator/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - python >=3
+    - pip
+  run:
+    - happer >=0.1
+    - insilicoseq >=1.3.6
+    - microhapdb >=0.3
+    - numpy >=1.15.4
+    - python >=3
+    - pyfaidx >=0.5.5.2
+    - pysam >=0.15.2
```
epruesse: Want to pin that one to `>=0.15.2,<0.16.0a0`? I think they changed API before on patch level bumps.

recipes/microhapulator/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+test:
+  imports:
+    - microhapulator
+  requires:
+    - pytest >=3.10
+    - pytest-cov >=2.6
+    - wget
```
epruesse: This was generated I think. Do you need the requires? 

Is there some other useful test you can run, like a command? Nothing big, just something to test whether it looks like it should in principle work.

recipes/microhapulator/run_test.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+wget -O hg38.fasta.gz http://hgdownload.soe.ucsc.edu/goldenPath/hg38/bigZips/hg38.fa.gz
+gunzip hg38.fasta.gz
+faidx hg38.fasta chr1:100-150 > /dev/null
+
+pytest -m 'not known_failing' --cov=microhapulator --pyargs microhapulator
```
epruesse: Ah, never mind, you are running the pytests. That's a lot more testing than a lot of our recipes see. We really need a  way to store the test data though. Hg38 will be pretty big. Let's hope the link never goes away.

epruesse:
@biocondabot merge

recipes/microhapulator/meta.yaml
standage: Good to know. I'll consider that for the next update.

recipes/microhapulator/run_test.sh
standage: These are the same tests that are run as part of MicroHapulator's CI build on Travis. For the next update I'm considering a less comprehensive smoke test just for bioconda purposes.

standage:
Thanks for your feedback @epruesse!

recipes/microhapulator/meta.yaml
standage: Addressed in https://github.com/bioconda/bioconda-recipes/pull/15522/commits/9f0879f8d8205d76eda2a903218182dceaee9a77.

PR #15047, opened 2019-05-02
Add mitos 2.0.1 (and archive 1.0.1)

bernt-matthias:
:information_source:

----------------

note that the mitos and mitos2 branches will probably never merge
hence I guess its best to have both versions independently

I guess maintaining the recipe of mitos(1) in a separate directory is reasonable (note that I needed to increase the build number of mitos1 because of the move to the subdir)

### DIFF
--- a/recipes/mitos/meta.yaml
+++ b/recipes/mitos/meta.yaml
@@ -20,15 +20,15 @@ requirements:
     - python <3
   run:
     - biopython ==1.73
-    - ete2 ==2.3.10
+    - ete2 ==2.3
     - python <3
     - reportlab
-    - blast-legacy ==2.2.26
-    - hmmer ==3.2.1 
-    - infernal ==1.1.2
+    - blast-legacy ==2.2
+    - hmmer ==3.2
+    - infernal ==1.1
     - viennarna <2
     - r-base  
-    - r-ggplot2 ==3.1.1
+    - r-ggplot2 ==3.1
     - openjdk
 
 test:
--- a/recipes/mitos/meta.yaml
+++ b/recipes/mitos/meta.yaml
@@ -20,15 +20,15 @@ requirements:
     - python <3
   run:
     - biopython ==1.73
-    - ete2 ==2.3
+    - ete2 ==2.3*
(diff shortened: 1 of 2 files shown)


recipes/mitos/meta.yaml
```diff
@@ -24,7 +24,8 @@ requirements:
     - python <3
     - reportlab
     - blast-legacy ==2.2.26
-    - infernal ==1.0.2
+    - hmmer ==3.2.1 
+    - infernal ==1.1.2
```
epruesse: A lot of these are probably overpinned.

recipes/mitos/mitos1/meta.yaml
```diff
@@ -0,0 +1,71 @@
+{% set name = "mitos" %}
```
epruesse: Do we need to archive this?
Basically, there are two reasons - a) we have things depending on `mitos <2` or b) there will be future releases of `mitos 1`. If those don't apply - just remove this.

bernt-matthias: I will provide bugfix releases for MITOS1 if necessary.

epruesse: Think that that's likely? You can always resurrect the old recipe from git. I'm just trying to keep the number of sub-recipes down. Most of us don't have time to support old branches, so it's unusual for backports to actually happen.

The real reason for the sub-recipes is that we have some tools that depend on older versions. So if something pins `blast=2.7.*` then we need to keep the blast 2.7 recipe around in case we need to update e.g. openssl and rebuild.

recipes/mitos/meta.yaml
bernt-matthias: I removed the minor versions.

bernt-matthias:
Linting seems to fail because also recipes/augustus/3.2.3 is tested. Don't know why.

epruesse:
Merge in master - then that goes away

bernt-matthias:
@epruesse thanks for the tip. it worked. ready from my side.

PR #15070, opened 2019-05-03
Python edlib 1.2.3 1

pb-cdunn:
:information_source:

----------------


pb-cdunn:
I get the same `execvp` error when I build locally, but it works fine with *my* `gcc-6.4.0`. Why is conda using `gcc-4.8.5`? I guess we need `cxx` if we want a newer `gcc`. Trying that now ...

PR #15073, opened 2019-05-03
Bump build, change nothing.

roryk:
:information_source:

----------------

Heya, don't merge this-- I just want to see if this build fails in the same way with the automated testing as it does with circleci. Here I bumped the build of GenomeInfoDbData and changed nothing else, and the package fails as it can't find GenomeInfoDbData. I ran into this trying to update a package that depends on it, which fails for the same reason. If I download the same GenomeInfoDbData tarball that bioconda is grabbing and install it locally with R CMD INSTALL it works fine, trying to figure out where the problem is coming from and just want to make sure its not something weird with my circleci setup.


roryk:
...

PR #15172, opened 2019-05-08
Busco update

pvanheus:
:information_source:

----------------


pvanheus:
with Augustus 3.3 I experienced core dumps from Augustus in BUSCO, so this recipe shifts the Augustus dependency to 3.3.2, which does not dump core.

druvus:
Does the BLAST version needs to be pinned to an old version? I can not find any info about it in the BUSCO repo. It seems they are running latest BLAST version in their docker.

druvus:
I also did some limited test of BUSCO using BLAST 2.2.31 vs. 2.7.1 and got identical results.  Would it be possible to relax the BLAST pinning?

pvanheus:
@druvus there was a problem previously with multithreaded BLAST. See [this](https://github.com/nextgenusfs/funannotate/issues/236) discussion where it is mentioned as existing up to at least BLAST+ 2.7.

druvus:
Thanks for the explanation!

beroe:
Is it clear that v 2.9 still has this error? 
https://www.ncbi.nlm.nih.gov/books/NBK131777/

PR #15188, opened 2019-05-08
Add C++ compiler as a necessary preprocessor runtime requirement for the GMTK recipe

EricR86:
:information_source:

GCC also needs to be added as a build requirement since the chosen compiler name is compiled in as a string into the resulting GMTK binary. Using jinja templates results in a temporary build compiler path inserted into the GMTK binary.

----------------


dpryan79:
Does this really need a C++ compiler after it's been built? Does it actually matter that the resulting compiler string is long?

EricR86:
@dpryan yes unfortunately GMTK needs the same compiler path for both building and for runtime.
This is the current error in the recipe:
```shell
$ gmtkTriangulate -strFile test-20190506.ZjXi8C/traindir/segway.str
sh: /opt/conda/conda-bld/gmtk_1557160854811/_build_env/bin/x86_64-conda_cos6-linux-gnu-cpp: No such file or directory
Unexpected EOF Error: expecting GM magic keyword at line 1
Exiting Program
```
The GMTK program uses the compiler preprocessor itself during runtime to process some of it's input files. If a jinja templated compiler is used for the build machine, the unique build machine compiler path is compiled into the recipe:
```shell
$ strings $(which gmtkTriangulate) | grep cpp
libhdf5_cpp.so.103
cppCommandOptions
Command line options to give to 'cpp'
/opt/conda/conda-bld/gmtk_1557160854811/_build_env/bin/x86_64-conda_cos6-linux-gnu-cpp -x assembler-with-cpp
WARNING: opening file '%s' via cpp but also using char '%c' as a comment, unexpected results may occur
ERROR: unable to open standard input via cpp
cppCommandOptions
```
Technically all that matters is equivalent strings for the compiler path are used for both build and runtime. Traditionally on systems it's just always been plain 'gcc' and it's the default for GMTK.

EricR86:
Any workarounds or perhaps a method of having the linter ignore a line would be greatly appreciated.

dpryan79:
Why not just add the directory with the compiler to the `$PATH` inside `build.sh` and then just not pass the full path to the compiler when building GMTK? That should work correctly on Linux and OSX.

EricR86:
@dpryan79 how do I get access to the directory of the compiler, or even the name of the compiler inside build.sh? I don't have access to the jinja templates inside build.sh correct?

It also seems like it wouldn't be the `BUILD_PREFIX` environment variable and the `BUILD_PREFIX/bin` should already on the `PATH` [according to the conda docs ](https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#env-vars)as well

dpryan79:
They compiler is the `$CC` or `$CXX` environment variable. So just parse that.

PR #15190, opened 2019-05-08
Minfi version updated 1.30.0

kpbioteam:


dpryan79:
I'm taking care of the bioconductor 3.9 release packages in the `bulk` branch. I don't know when minfi will get processed, but it shouldn't take more than a week or so.

kpbioteam:
Hi @dpryan79  we have been working on this per urgent editor request that we hoped to address by the end of this week. Is there a way we could have it sooner, we are happy to help ?

dpryan79:
No and unfortunately this is one of the packages that's likely to take longer. Anything depending on BiocGenerics will be delayed until the bioconductor core team gets back to me about whether that package will function correctly under R 3.5 (it states that it's only compatible with R >=3.6, which nothing in the conda universe will be built against until later in the summer).

PR #15207, opened 2019-05-09
Update seqCAT to version 1.4.1

fasterius:
This is my very first PR to Bioconda, so would appreciate feedback from @bioconda/core. I started to read up on how to add a recipe for my `seqCAT` Bioconductor package, and got to the point of creating a skeleton when I realised it had already been added in a bulk submission some months back. I thus wanted to update the package to the latest version, but it seems Bioconda does not yet use `R 3.6`? At least I thought I could update from `1.4.0` to `1.4.1`. I tried to build it locally, which ran successfully, but please let me know if I did something wrong or if I need to change something. 

----------------

### DIFF
--- a/recipes/bioconductor-seqcat/meta.yaml
+++ b/recipes/bioconductor-seqcat/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "1.4.0" %}
+{% set version = "1.4.1" %}
 {% set name = "seqCAT" %}
 {% set bioc = "3.8" %}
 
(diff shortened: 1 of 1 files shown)


dpryan79:
Thanks @fasterius ! I'm working on the Bioconductor 3.9 packages in a different branch and working with the bioconductor core team on some of the R 3.6 restrictions. So hopefully we'll have the most recent version of seqcat available before R 3.6.1 is released (the normal time at which packages are updated for a new R release).

fasterius:
Thanks for the info, @dpryan79! What do I need to do as a package author in order to help with keeping my package(s) updated for Conda? I'm an avid user of Conda, but this is my first foray into actually contributing something to do. Given that the previous version (`1.4.0`) was automatically/bulk added (by you, I believe), is this something that I as author need to keep track of? What is the general idea behind Bioconductor and Conda as far as updates go?

dpryan79:
I'll add you to bioconda so you can update it more easily in the future yourself. In general, we (the core team) handle new bioconductor releases, since that involves a huge number of packages that need to be pushed out in a coordinated manner (and it allows us to give feedback to the bioconductor team on things that should be changed in their packages). If you want to make additional updates after that you're certainly quite welcome to do so :)

fasterius:
Thanks! So, the core team would update the first time for a new bioconductor release (for which package authors do not need to do anything), and then package authors would keep subsequent version of their packages (*i.e.* bugfixes and cherry-picked updates) updated within the same version of bioconductor?

dpryan79:
Yup, exactly.

fasterius:
Great, good to know - thanks for all the help!

PR #15217, opened 2019-05-09
Update seqcluster version

lpantano:
This version is fully compatible with python3

:information_source:

----------------

### DIFF
--- a/recipes/seqcluster/meta.yaml
+++ b/recipes/seqcluster/meta.yaml
@@ -1,4 +1,4 @@
-{% set tag="fd0a6c6" %}
+{% set tag="v1.2.4" %}
 
 about:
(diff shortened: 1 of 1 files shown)


recipes/seqcluster/meta.yaml
```diff
@@ -30,6 +30,7 @@ requirements:
     - progressbar2
     - biopython
     - mirtop >=0.4.17a
+    - viennarna
```
micknudsen: Is this needed? It is not listed in the `requirements.txt` file.

lpantano: yes, it is. I need to update the requirements as well. I always use conda, so I forget about that one. Thanks for looking at it.

PR #15238, opened 2019-05-10
Initial commit.

lucventurini:
GenomeThreader aligner.
----------------


PertuyF:
@lucventurini possible trouble here as the license (ISC) is not standard and clearly states it should be packaged with the software. Any though @bioconda/core ?

dpryan79:
Can you add a copy of the license in? You can use `license_file:` for that under `about:` if memory serves.

lucventurini:
@PertuyF, @dpryan, yes I can find and copy a licence of the file in. I will correct the recipe next week.

PR #15240, opened 2019-05-10
Fix InSilicoSeq dependencies

standage:
The original InSilicoSeq recipe didn't reference dependencies correctly. This PR updates the config to make sure all dependencies are available when InSilicoSeq is imported as a dependency for another recipe.

----------------


standage:
@bioconda/core:

The [latest build for MicroHapulator](https://circleci.com/gh/bioconda/bioconda-recipes/55773?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) failed since InSilicoSeq didn't correctly install its prerequisites (trying to import BioPython results in `ModuleNotFoundError: No module named 'Bio'` error messages).

This PR correctly adds all the prerequisites to meta.yaml. I downloaded the insilicoseq tarball (CircleCI artifact linked above) and `conda install`ed it in a fresh environment. The package itself installed correctly, but it didn't install the prerequisites correctly. Is this expected behavior for a tarball, or do I still have something wrong with my configuration?

dpryan79:
:+1: Is there any test that you can think of to add to the recipe? That'd help future proof this.

standage:
> *Is there any test that you can think of to add to the recipe?*

Yes, I'll come up with a quick test.

standage:
Hi @HadrienG, any idea about the error that is causing the [latest build failure](https://circleci.com/gh/bioconda/bioconda-recipes/56294)?

```
ValueError: Object arrays cannot be loaded when allow_pickle=False
```

I think I can trace it down to [this line](https://github.com/HadrienG/InSilicoSeq/blob/acfe39845268cc9ad7d0c7b42eea0dc69f910fbf/iss/error_models/__init__.py#L39). Perhaps it needs `allow_pickle=True`? Odd if this never came up in your testing, though.

standage:
Boom. See https://github.com/numpy/numpy/releases/tag/v1.16.3. I'll open up a ticket on the InSilicoSeq repo. In the meantime, I'll pin a lower numpy version for this release.

standage:
@bioconda/core @dpryan79 It looks like the `shasum` command is not available on the CircleCI image used for the test build. It looks like it's also not available from conda-forge. Any way I can compute a checksum with `shasum` as part of the test build?

recipes/insilicoseq/run_test.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+curl -L https://ndownloader.figshare.com/files/15127220 > bogus-genome.fasta
+curl -L https://ndownloader.figshare.com/files/15127223 > checksums.sha
+
+iss generate --model MiSeq --seed 42 --cpus 1 --draft bogus-genome.fasta -n 100 -o bogus-reads
+shasum -c checksums.sha
```
epruesse: try `md5sum`, that might be more available

recipes/insilicoseq/run_test.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+set -eo pipefail
+
+curl -L https://ndownloader.figshare.com/files/15127220 > bogus-genome.fasta
+curl -L https://ndownloader.figshare.com/files/15127223 > checksums.sha
```
epruesse: How big are those? Can we have them in-repo?

@bioconda/core - I think we need a directory in the repo, or a second repo, with some stanard dest-data that tools can share. This repo is a beast already, adding more and more testdata will not help, but without many things cannot be verified properly.

recipes/insilicoseq/run_test.sh
standage: Will try.

recipes/insilicoseq/run_test.sh
standage: @epruesse That was my strategy with https://github.com/bioconda/bioconda-recipes/pull/15240/commits/3aad15ed63bcf5d7439a2becc795aa0f8d2e31eb. Alas, the file could not be found.

```
FileNotFoundError: [Errno 2] No such file or directory: 'bogus-genome.fasta'
```

See https://circleci.com/gh/bioconda/bioconda-recipes/56296.

recipes/insilicoseq/run_test.sh
standage: Hallelujah that did it!

standage:
Ready for merge.

recipes/insilicoseq/run_test.sh
epruesse: See https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#test-section

You can have the test files as part of the source, then they need to be listed as `source_files` under `test`, or as files in the recipe folder, then they need to be listed as `files` under `test`. That should work - we've got plenty recipes. Might want to zip them, they won't change so no need to keep them clear text. The smaller the better.

epruesse: Catch with the figshare hosting is that the files might go away.

epruesse: Opened an issue #536

standage: figshare is actually a fairly stable archival service, but yes I agree it's not ideal. I'll see if I can dig up an example with test files in the recipe.

standage: Ok, test files are bundled *in situ*.

epruesse:
LGTM, thanks @standage!

@biocondabot merge

epruesse:
Just a note, now that I see the test and what you needed shasum for - bio results are, even if you specify a seed, often "fragile" at byte level. The next version, even for the same seed, is pretty likely to produce a slightly different file. The test will fail then. It's usually better to test for some more stable property. Honestly, often just that the tool was able to produce some kind of output is sufficient. Whether the results of a new version are better or worse than that of an old version... leaving that to the next review roundup paper.

standage:
@epruesse I'll keep that in mind.

@bioconda/core It looks like the insilicoseq-1.3.5_py1 artifact didn't upload correctly. Can you confirm? I don't see it on the [bioconda package index](http://bioconda.github.io/recipes/insilicoseq/README.html#package-insilicoseq), and it's not grabbing the latest version with the latest [microhapulator build](https://github.com/bioconda/bioconda-recipes/pull/15040).

epruesse:
It's there - https://anaconda.org/bioconda/insilicoseq/files

We now have a CDN in front of the packages - we were getting too big for anaconda to host us directly, which is why there were all those 50x gateway error failures. The price is that it takes ~20-30 minutes for packages to appear everywhere.

epruesse:
The docs will rebuild at midnight UTC.

standage:
@epruesse The reason I ask is that #15040 is still failing. It pulls in insilicoseq-1.3.5-py_0 instead of py_1, even though it has been hours since py_1 was merged.

epruesse:
I see that off and on. Allegedly it's been fixed. Basically, conda in my experience whatever others say does like to pick an old build number if the requirements are easier to satisfy. :(

PR #15246, opened 2019-05-10
Update mirtop to last devel

lpantano:
----------------

### DIFF
--- a/recipes/mirtop/meta.yaml
+++ b/recipes/mirtop/meta.yaml
@@ -1,5 +1,5 @@
 {% set version="0.4.18a" %}
-{% set revision="a61d043" %}
+{% set revision="b0c325d" %}
 
 package:
   name: mirtop
@@ -7,11 +7,11 @@ package:
 
 source:
   url: https://github.com/miRTop/mirtop/archive/{{ revision }}.tar.gz
-  sha256: 2f4e91232034b519840da002bb1fedf04b6195e8a865720e606d8b47f974989b
+  sha256: 57018fcf240b721e4e551df42006f504b437b707f51de5550179f7dc22e4cbdc
 
 build:
   script: $PYTHON setup.py install --single-version-externally-managed --record=record.txt
-  number: 1
+  number: 2
   entry_points:
     - mirtop=mirtop.command_line:main
 


epruesse:
@lpantano - Unlike PyPI, Conda will not hide alpha releases from users, but will always install the most recent version if it can, often older versions if the requirements are best met that way. Having alphas and betas on Bioconda is therefore not the best of ideas.

Has this seen some testing? Can you tag a 0.4.18 and release that?

bgruening:
@lpantano there was an open question from a reviewer. Please do not merge PRs without addressing those questions.

Thanks for your understanding and contributions!

lpantano:
@bgruening, when I merged I didn't see that comment, I don't know exactly why. I think the web didn't refresh. @epruesse , sorry about that, it wasn't my intention.

The beta versions are as well tested in travis for this package.  I till tag the release of each version from now on. Thanks.

lpantano:
can I ask if now the idea is to have only stable versions? because I know many that are not doing that so I just want to know if this is something is changing now? Thanks!

epruesse:
I don't want to spawn another discussion right now - /core is already stretched thin. But, to answer your question:

- "Stable" is a matter of perspective. Some software, say databases or filesystems, takes years until it is considered stable. Bioinformatics software certainly ranges on the very other end of the scale. Requiring people to only publish "stable" software will just lead to them tagging something as stable, whether or not it actually is. So no real point there.

- As I mentioned,  `conda` has no concept of pre-releases. While `pip` will not install pre-releases unless you explicitly ask for them (`--pre`), `conda` will attempt to always install the most recent version, if that is possible as per requirements. So if you push an unstable release to Bioconda, that release will end up on people's desktops and they will be unaware that they got a release meant for testing, not for production. They can't even easily prevent that.

- Bioconda packages will never be removed from the channel again. As soon as there are people using the software, you will have `conda export`ed environments and those will break if we remove packages. We did that in the beginning and learned the hard way that not even re-labeling is a good idea. Assume that things uploaded to Bioconda are to remain for ever.

- We are in fact trying to get Bioconda to become more stable. When the project started, it was mostly about getting as much in as possible, to become comprehensive. Now, with 7k recipes and 1k contributors and 500+ members, the challenge is to be reliable and up-to-date. We are a global resource for tools, many people use our packages and containers in production. Throwing beta stuff in there breaks things.

- If someone tried to update a recipe and wasn't also the author of the software in question, I would reject it. The beta tag given by the author means that it's not prime-time ready yet, and pushing out something broken might harm the popularity of the tool and therefore the author. 

- For self publishing. Well. It's up to you to decide when it's ready for a broader audience anyway. As long as you aren't the author of `pysam` or something as frequently relied upon as that. Won't allow a beta of `pysam` to wreak havoc (if only because I'd be part of the ranting mob).

So in summary - use caution and judgement. I was merely pointing out that the beta will be pushed to everyone newly installing it or calling `conda update`, and that if you label it beta, that may not be what you want to do.

I myself would like to have something like a `bioconda-testing` channel for bleeding edge software. One where it's made clear that broken packages will be retracted and environments should not be expected to work forever. I'm always nervous when I push the "release" button on my own things, and I'd like to have a way to get actual by-the-user-testing before I publish new versions. We just don't have the manpower for that.

At the moment, the best we could do was the command the bot posted above. It allows installing without actually pushing to the Bioconda channel, so the release isn't "for posterity" yet and can be withdrawn.

Did this remove any remaining clarity? ;-)

lpantano:
OK, I got it. I will adopt the best as possible to this. Thanks for the extensive explanation!

epruesse:
It's just my $.02. Hope it makes sense.

lpantano:
Sure, I am quite open to adapt what makes sense to the majority and the ones who have more expertise. I will improve my workflow. I value your comments a lot.

PR #15247, opened 2019-05-10
[WIP] Rebuild antismash with bcbio-gff

npavlovikj:
:information_source:

----------------


npavlovikj:
This recipe builds fine, but the `Copying packages` phase throws the `Too long with no output (exceeded 10m0s)` error. When I tested it locally with `bioconda-utils`, the testing phase was successful, but it did take 20-30 minutes. @bioconda/core , do you have any suggstions how to fix this error?

I read somewhere that I should remove the pinnings, but all the pinned versions are available in the channels, and the test does pass - it just takes more than 10 minutes.... Any feedback and suggestions are appreciated!

recipes/antismash/meta.yaml
```diff
@@ -1,28 +1,30 @@
...
   url: https://dl.secondarymetabolites.org/releases/{{ version }}/antismash-{{ version }}.tar.gz
   sha256: {{ sha256 }}
   patches:
-    - smiles.patch [osx]
+    - smiles.patch  # [osx]
 
 requirements:
   host:
-    - python >=2.7,<3.0a0
+    - python >=2.7,<3
```
epruesse: Just FYI - there is "method to this madness". Basically, it's that `3.0a0 < 3.0`. So if you say `<3.0`, then `3.0a0` or `3.0beta1` would be ok, because they are not yet `3.0`. So for upper bound, using the annoyingly cryptic `3.0a0` is actually correct if it's supposed to say `2 point something`. 

In the case at hand, there won't be a Python 3.0a0, so it's just impeding legibility. But if you see it elsewhere, it may make sense, or could at least be argued for. While we hope not to have any one add versions like `2.0a0`, you know how that can go...

epruesse:
@npavlovikj This is my bad. I added that stage to copy the packages and images build during testing into the artifacts thingy, so that we can just upload and merge without wasting time building things again. Looks like this one builds a huge container. Perhaps 20 min suffices. Let's see...

recipes/antismash/meta.yaml
npavlovikj: Thanks for the explanation Elmar! I was indeed wondering why "3.0a0a" was used intead of "3", and your reason behind it makes total sense.

epruesse: In case there is ever a question:

```python
>>> from conda.exports import VersionOrder
>>> VersionOrder('3.0a') < VersionOrder('3.0') 
True
```

That's the function doing it. So it's what we've got to live by. Its mostly PEP compatible, but there are some idiosyncrasies to it. E.g. a special case is
```python
>>> VersionOrder('3.0post') < VersionOrder('3.0') 
False
```

epruesse:
Holy....

```
Uploading /tmp/artifacts/images/antismash:4.2.0--py27hb89731f_1.tar.gz (5.3 GB): DONE
```

Why is that thing this crazy huge? What is in there?

epruesse:
Looks like it became that big with  4.0.2--py27_3. Well. No way the bot can do that at the moment. Doesn't have enough disk. This needs to merge with normal merge button.

npavlovikj:
Wow, that is a HUGE image. The local ".tar.bz2" file I got was about 50MB, so I have not idea why the image is so big (the dbs are additionally downloaded as far as I can see)...

npavlovikj:
The increased time fixed the Linux build, but the OSX one is now failing with:
_conda.exceptions.PaddingError: Placeholder of length '80' too short in package /Users/distiller/project/miniconda/conda-bld/antismash_1557555451496/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/bin/alphtype.
The package must be rebuilt with conda-build > 2.0._

I think I have seen this error when "r" has been used as a dependeny. Since that is not the case here, I guess I need to check which "conda-build<=2.0" packages are being installed.

epruesse:
Hmm. Weird. This should only affect the currently build files. Is alphtype part of antismash?

npavlovikj:
I think `alphtype` comes from meme, http://meme-suite.org/doc/alphtype.html.

npavlovikj:
Looks like there was similar issue with the previous "antismash" build, and "meme==4.11.2" was re-built, https://github.com/bioconda/bioconda-recipes/pull/9789.

However, for the OSX build here, an older version of `meme==4.11.2` is used:
**meme:                          4.11.2-py27pl5.22.0_0         bioconda**
instead of the newest `meme=4.11.2=py27pl526hc595358_2` one.
I will try and pin the OSX "meme" version to the latest one, and see what happens...

epruesse:
@biocondabot recheck

epruesse:
@biocondabot merge

epruesse:
@BiocondaBot merge

epruesse:
@biocondabot merge

npavlovikj:
Thanks for merging this @epruesse  - I hope loosing the pinning won't cause some issues in running the program...
And did you magically reduce the image size, or the Bot is capable of merging big images now? :)

epruesse:
Should be ok. Few things bump the patch level for significant changes.

I gave the bot more disk space and changed some timeouts, so now it should be ok with 10-20GB. It's just slow. It has only 100mbit, so it takes 10 minutes to copy a 5GB image (have to download and then upload because the docker upload tool needs the image tar).

PR #15318, opened 2019-05-11
Update rsem to 1.3.2

epruesse:
Update [`rsem`](https://bioconda.github.io/recipes/rsem/README.html): **1.3.1** &rarr; **1.3.2**

[install with bioconda](http://bioconda.github.io/recipes/rsem/README.html) [Conda](https://anaconda.org/bioconda/rsem/files)

Info | Link
-----|-----
Recipe | [`recipes/rsem`](https://github.com//bioconda/bioconda-recipes/tree/bump/rsem/recipes/rsem) (click to view/edit other files)
Releases |[https://github.com/deweylab/RSEM/tags](https://github.com/deweylab/RSEM/tags)
Author | `@deweylab`
***

This pull request was automatically generated (see [docs](https://bioconda.github.io/updating.html)).

PR #15319, opened 2019-05-11
Bump RSEM 1.3.1

epruesse:

Just bump RSEM - hoping that it will fix the install problem. At the moment, `conda create -n rsem rsem` just times out. :(

### DIFF
--- a/recipes/rsem/build.sh
+++ b/recipes/rsem/build.sh
@@ -2,6 +2,8 @@
 
 set -x -e
 
+CXX=$CXX
+
 pushd $SRC_DIR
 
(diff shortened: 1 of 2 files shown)


bgruening:
@epruesse we should add a linter rule that checks, "if boost in host; assert cxx in build"

epruesse:
@bgruening I think I'll close this one. It's too complicated, and there are other things "wrong" with it. E.g. it vendors a whole bunch of packages, including samtools/htslib, which is where it will fail next.

I've rewritten it in #15318 for the 1.3.2.

PR #15337, opened 2019-05-13
Genomethreader - adding the license file

lucventurini:
Adding the license file in, and added myself as the recipe maintainer.

---------------


lucventurini:
@dpryan79 , @PertuyF , I have added the license file as a URL, as the @dpryan79 suggestion. Thank you for having double checked!

dpryan79:
Is there a reason you're using precompiled binaries? That's usually not great for portability.

lucventurini:
> Is there a reason you're using precompiled binaries? That's usually not great for portability.

This is how the software was distributed until a short time ago, it used to be accessible only through the distribution page I am pulling it from.

Moreover, compiling genomethreader also requires pulling and compiling genometools, which would add to the complexity of the recipe. I am trying myself to compile the binary on my machine, and unfortunately it is one of those cases where the recipe will require considerable massaging before functioning.

dpryan79:
Fair enough, at least genometools is in conda, so that's not a big deal. You'll need to download the file and copy it into the recipe directory.

lucventurini:
> Fair enough, at least genometools is in conda, so that's not a big deal. You'll need to download the file and copy it into the recipe directory.

Actually and unfortunately, no, GenomeTools is *not* in bioconda. I know it looks like it as there is a package so named, but it references (this Python package)[https://github.com/flo-compbio/genometools], not the more famous and widespread C/C++ suite. The "proper" GenomeTools is still off 
BioConda.

Back on GTH, any idea of why the MacOSX test is blocked? The linux test passed successfully, but the test is currently hanging due to MacOSX not starting.

dpryan79:
Just give it time, there's only one OSX job at a time.

lucventurini:
> Just give it time, there's only one OSX job at a time.

OK, thanks!

lucventurini:
>  Actually and unfortunately, no, GenomeTools is not in bioconda. I know it looks like it as there is a package so named, but it references this Python package, not the more famous and widespread C/C++ suite. The "proper" GenomeTools is still off BioConda.

Apologies on this, I just realised that "genometools-genometools" on BioConda is actually the package that I thought missing.

PR #15351, opened 2019-05-14
Rebuild desman with bcbio-gff

npavlovikj:
:information_source:

----------------


npavlovikj:
@druvus , can you please review this recipe when you have a chance?

I tried removing `build.sh` and adding `script` in `meta.yaml`. However, regardless of the command used (python -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv"; or python setup.py install --single-version-externally-managed --record=record.txt; or replacing "python" with Jinja2 variables) I kept getting the error `ModuleNotFoundError: No module named 'desman'`. If you have an idea why this is not working, please let me know.

Also, since both `python` and `perl` are in the `run` section, looks like the `test` import uses the Perl syntax. Since `desman` and `sampletau` are Python imports, and not Perl ones, "desman" was blacklisted earlier. The only way I could think of overwriting this default behavior was to add `run_test.sh` and `run_test.py`.

druvus:
@npavlovikj thanks for your help rebuilding all depending packages. I have been out of office so I didn't have a chance to look at this until now. 

I think it would be better to keep the command tests in meta.yaml as run_test.* will not be tested in mulled-test.

npavlovikj:
No worries at all @druvus - I know you will reply when you can, so I appreciate your feedback :)

So, I should add the `commands` part in `meta.yaml` and keed `run_test.py` for the Python imports? Will mixing those 2 approaches work?

npavlovikj:
Please ignore my previous comment @druvus , I fixed the recipe accordingly.

druvus:
@bioconda-bot recheck

PR #15363, opened 2019-05-14
ngs-bits 2019_05

marc-sturm:
:information_source:

----------------


marc-sturm:
@bioconda/core Any idea why qmake cannot find g++? It worked with the same setup  on 9th of April when I built the last version. Any hints are appreciated...

dpryan79:
There is no `g++`, but there's a `${CXX}`.

marc-sturm:
@dpryan79 Thanks for the hint.

I already tried to use the environment variable in the build.sh:
qmake ../src/tools.pro CONFIG-=debug CONFIG+=release DEFINES+=QT_NO_DEBUG_OUTPUT QMAKE_CXX=${CXX}
It seems to have no effect.

Now I added debug output, perhaps then it becomes clear why it fails...

marc-sturm:
@dpryan79 Hi Ryan, any idea what's wrong with the Qt version string? Thanks, Marc

dpryan79:
Maybe this has something to do with recipes being pinned to `qt` 5.9 already, so you should be able to just omit `>=5.9` entirely.

marc-sturm:
@dpryan79 Thanks a lot for your help! Everything works now. Is the recipe ok like this? If so, can you merge it?

dpryan79:
:+1: from me!

dpryan79:
Merged, thanks @marc-sturm!

PR #15402, opened 2019-05-15
Add unitig-caller

johnlees:
:information_source:

----------------

Most of the work is done in the `setup.py` provided with the package

### DIFF
--- a/recipes/unitig-caller/meta.yaml
+++ b/recipes/unitig-caller/meta.yaml
@@ -14,7 +14,7 @@ build:
   skip: True #  [osx]
   script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv"
   entry_points:
-    - unitig_caller = unitig_caller.__main__:main
+    - unitig-caller = unitig_caller.__main__:main
 
 requirements:
   build:
@@ -24,7 +24,6 @@ requirements:
     - python >=3
     - pip
     - cmake
-    - zlib
     - pybind11
   run:
     - python >=3
@@ -37,7 +36,7 @@ test:
     - unitig-caller --version
 
 about:
-   home: https://github.com/johnlees/unitig-counter
+   home: https://github.com/johnlees/unitig-caller
    license: Apache-2.0
    license_family: APACHE
    license_file: LICENSE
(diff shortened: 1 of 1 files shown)


johnlees:
Request for review please @bioconda/core (is this still the correct group to request review from?)

johnlees:
Q: This worked ok, but I had intended to include `pthread-stubs` in the host section. When trying to find it, it seems to come from having the compiler tools in the build section:
```
Needed DSO x86_64-conda_cos6-linux-gnu/sysroot/lib/libpthread.so.0 found in CDT/compiler package conda-forge::gcc_impl_linux-64-7.3.0-habb00fd_1
```
Is that ok, or should I add `pthread-stubs` to the host section too?

recipes/unitig-caller/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+  host:
+    - python >=3
+    - pip
+    - cmake
+    - pybind11
```
dpryan79: Can you try moving `cmake` to the `build:` section? `pybind11` only contains headers, correct?

recipes/unitig-caller/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  run:
+    - python >=3
+    - pthread-stubs
```
dpryan79: Is this actually linking against a library from Xorg?

recipes/unitig-caller/meta.yaml
johnlees: Yes, I believe `pybind11` is header only (so not in run). I will move `cmake`

recipes/unitig-caller/meta.yaml
johnlees: I'm not sure exactly what you mean here, but that wasn't my intention anyway

johnlees:
@dpryan79 Do these changes look like what you were after? I just got rid of the pthreads stuff
Did you want me to do anything with `pybind11`?

dpryan79:
:+1: from me. No need to do anything with pybind11, I was just double checking.

epruesse:
```
 - python >3
```
doesn't do what you want unless you also have `noarch: python` set.

Since you are linking to python, you needto build for all python versions (3.6 and 3.7), so we need to use skip and add python.

It's funky conda stuff. Should work now.

johnlees:
Great, thank you both very much 😀

PR #15414, opened 2019-05-16
Add XS read simulator

TMiguelT:


epruesse:
@TMiguelT Do you want to add yourself as maintainer?

TMiguelT:
Of the bioconda package? Sure.

recipes/xs/meta.yaml
```diff
@@ -0,0 +1,31 @@
+package:
+  name: xs
```
bgruening: xs is very generic, I'm afraid of namespace clashes, can we name it more verbose?

recipes/xs/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  build:
+    - {{ compiler('c') }}
+
+test:
+  commands:
+    - XS -n 1 /dev/null
+
+about:
+  home: https://github.com/pratas/xs
+  license: GPL-2.0
```
bgruening: We need a license file for GPL. Could you please add one to the repo I guess or if you have upstream power also to upstream :)

recipes/xs/meta.yaml
epruesse: Technically it should be ok. There are no conflicts at present, excepting perl-xs. Two letters is unusually short though.

recipes/xs/meta.yaml
TMiguelT: There isn't a license file in the repo, but there is this: https://github.com/pratas/xs/blob/master/README.md#license. Sadly I don't have upstream power

recipes/xs/meta.yaml
TMiguelT: It could also be capitalised to XS, since that's what the developer calls it? But that might be even worse

recipes/xs/meta.yaml
epruesse: Open an issue or a PR then and log it in the recipe.

TMiguelT: But the author links to a license already. Do I really need to PR upstream? Can't we just include the file they hyperlink to?

epruesse: The license the author links to requires that a copy of the license be distributed with each copy of the software.

TMiguelT: Okay, I've made an issue: https://github.com/pratas/xs/issues/3

pratas: It is now updated. Thanks TMiguelT !

epruesse: updated checksums and file

should this remain version `1.0.0`? The tool itself does not seem to have a version number.

pratas: I set only a simple tag as 2.0 in the release

epruesse: Technically, you called it `v2`, which would indeed be considered equivalent to `2.0.0.0.0`.

As long as this wasn't merged, we can still change this. As a rule, we don't take packages back though, because it breaks things on the user side, so now would be the time to decide on a versioning scheme (semver perhaps). 

One word of advice - if you expect users, put the version number into the code. There are tools that will help do that (bumpversion, rever, ...). It's just a matter of time until someone runs into problems, and knowing which version they have (`xs --version`) help a lot...

Commonly, things like missing files would be patch level changes. So `1.0.1`.

As per `semver`, the assumption is that `patch level` just fixes things, `minor level` is additions (you added a function), `major` is, a bit counter intuitive, when you remove things. 

The point is that I can say that the range from `[1.12,2.0)` is good for me, i.e. `1.12` brought the latest feature I need, and only `2.0` will remove anything I might need. It's similar (not quire the same) to library versions on linux (SO version - difference there is that the last digit actually gives the backwards compatibility range, a bit more complicated)

pratas: It not supports the --version
~/git/xs$ ./XS --version
Version: 2.1

epruesse: Ah - I didn't see that. Can you make tag and version match? That makes it easier for us, because the download file name and the version of the tool will be the same. Also means that future updates can be picket up by the bot.

TMiguelT:
Is anything else needed for this to get merged?

epruesse:
Just lingering issues that aren't ideal, the short package name and the tag/version mismatch. @bgruening - thoughts? It could be `xs-read-simulator` or `xs-sim` perhaps?

epruesse:
ping @bgruening

epruesse:
@Biocondabot recheck

pratas:
It does now. Commited!

TMiguelT:
@pratas would you mind fixing the tag/version mismatch, and maybe suggesting a longer name for the package than `xs`?

dpryan79:
Can we name this `fastq-xs` or `xs-fastq`? It'd also be much more apparent what it actually does then (I assumed that it did something with the XS auxiliary tag that bowtie2 produces...).

TMiguelT:
I've renamed the *package* to `xs-sim`, although the executable it installs is still called `XS`. Is that any better?

dpryan79:
Yup, that'll work

dpryan79:
@bioconda-bot merge

Issue #11583, opened 2018-10-25
Singularity contains built from bioconda have broken /etc/resolv.conf


pvanheus:
A test with a container built using `singularity build` from the Docker images on `quay.io` using Singularity 2.5.2.

```
$ singularity exec /tools/containers/sra-tools-2.9.1_1--h470a237_0.simg ping www.google.com
ping: bad address 'www.google.com'
$ singularity exec /tools/containers/sra-tools-2.9.1_1--h470a237_0.simg cat /etc/resolv.conf
cat: can't open '/etc/resolv.conf': No such file or directory
$ singularity exec /tools/containers/sra-tools-2.9.1_1--h470a237_0.simg ls -ld /etc/resolv.conf
lrwxrwxrwx    1 root     root            18 Feb 27  2014 /etc/resolv.conf -> ../tmp/resolv.conf
```

This is mentioned by @vsoch as a [Weirdo File](https://github.com/sylabs/singularity/issues/749#issuecomment-309934407). It is not clear where the link to `../tmp/resolv.conf` comes from.

vsoch:
`resolv.conf` should be mapped from the host to resolve dns names, see [here](http://man7.org/linux/man-pages/man5/resolv.conf.5.html). Is there either something different going on with binds, or permissions here? Also, I believe that Singularity 2.5* family has (runtime?) security issues, so I would consider upgrading and seeing if you reproduce the error there.

vsoch:
This still might be the busybox issue, e.g.,

> there is still the issue of the missing /etc/resolv.conf (noted in #749 (comment)) in biocontainers images pulled from quay.io due to their use of a version of busybox that is incompatible with how singularity exposes the host resolv.conf (unless a privileged user subsequently modifies the image with the workaround described in #749 (comment)).

pvanheus:
I think this is a `busybox` issue rather than a 2.5.* issue. I get the same result with 3.0:

```
$ /var/tmp/src/github.com/sylabs/singularity/builddir/singularity exec /var/tmp/diamond-0.9.21--1.simg ls -ld /etc/resolv.conf
lrwxrwxrwx    1 root     root            18 Oct 25 16:48 /etc/resolv.conf -> ../tmp/resolv.conf
$ /var/tmp/src/github.com/sylabs/singularity/builddir/singularity --version
singularity version v3.0.0
```

Apologies for the long-winded path to `singularity` - it is a local build referring to an imagine built from a bioconda container.

[...]

PR #15436, opened 2019-05-16
MAKER: add support for running Biocontainers-derived Singularity images using host MPI

nathanweeks:

This PR implements several minor changes to the maker recipe to allow running maker from a Biocontainers-derived Singularity container on multiple nodes of an HPC cluster using the host MPI.
@abretaud : could you please review?

1. Add `-mpi` option to maker script

The "maker" script tries to detect if its parent process is an MPI process manager, and if so enables MPI support. This autodetection can fail when MAKER is run from a Singularity container.  The patch adds a command-line "-mpi" option to the "maker" script to override MPI autodetection and force MPI on. I have communicated with MAKER maintainer Carson Holt; he may include something similar in the next MAKER release, in which case this patch could be removed. 

2. Use the extended base image

If the host (MPICH-derived) MPI process launcher is used to launch the current Biocontainers MAKER on multiple compute nodes, the MPI processes are unable to wireup in MPI_Init() due to being unable to resolve hostnames. This can be illustrated with a simple example:

```
$ singularity pull maker_2.31.10--pl526_13.sif docker://quay.io/biocontainers/maker:2.31.10--pl526_13
...
$ cat test.c
#include <mpi.h>
#include <stdio.h>

int main(void) {
   int rank, size, rank_sum;
   MPI_Init(NULL, NULL);
   MPI_Comm_size(MPI_COMM_WORLD, &size);
   MPI_Comm_rank(MPI_COMM_WORLD, &rank);
   MPI_Allreduce(&rank, &rank_sum, 1, MPI_INT, MPI_SUM, MPI_COMM_WORLD);
   printf("rank: %d size: %d rank_sum: %d\n", rank, size, rank_sum);
   MPI_Finalize();
}
$ singularity exec maker_2.31.10--pl526_13.sif mpicc test.c
$ srun -p test --pty --mem-per-cpu=4G  --nodes=4 --ntasks-per-node=2 -t 00:00:00 bash
[test-node]$ mpiexec -n 8 singularity exec maker_2.31.10--pl526_13.sif ./a.out
...
Fatal error in MPI_Init: Other MPI error, error stack:
MPIR_Init_thread(474)..............:
MPID_Init(190).....................: channel initialization failed
MPIDI_CH3_Init(89).................:
MPID_nem_init(320).................:
MPID_nem_tcp_init(173).............:
MPID_nem_tcp_get_business_card(420):
MPID_nem_tcp_init(379).............: gethostbyname failed, holy7c19302 (errno 2)
...
```

This is because when the Biocontainers BusyBox base image is used, the resulting Singularity containers have a broken /etc/resolv.conf (issue https://github.com/bioconda/bioconda-recipes/issues/11583). Until that issue is resolved, the extended image can be used as a workaround.

An example using a Singularity image created incorporating the maker recipe changes in this PR:

```
$ singularity exec maker_2.31.10--pl526_14-2019-05-16-4d08d97ea398.simg mpicc test.c
$ srun -p test --pty --mem-per-cpu=4G  --nodes=4 --ntasks-per-node=2 -t 00:00:00 bash
[test-node]$ mpiexec -n 8 singularity exec maker_2.31.10--pl526_14-2019-05-16-4d08d97ea398.simg ./a.out
rank: 0 size: 8 rank_sum: 28
rank: 4 size: 8 rank_sum: 28
rank: 1 size: 8 rank_sum: 28
rank: 2 size: 8 rank_sum: 28
rank: 5 size: 8 rank_sum: 28
rank: 3 size: 8 rank_sum: 28
rank: 6 size: 8 rank_sum: 28
rank: 7 size: 8 rank_sum: 28
```

3. Require perl-bioperl-core that includes PR https://github.com/bioconda/bioconda-recipes/pull/15001

That build removes Inline::C code that would otherwise cause a runtime error for a MAKER Singularity container when trying to compile the Inline::C code (in addition to resolving any oddities observed when run under MPI outside of a container that prompted the patch in the first place).

```
"/usr/local/bin/perl" -MExtUtils::Command::MM -e 'cp_nonempty' -- IndexedBase_14e0.bs blib/arch/auto/Bio/DB/IndexedBase_14e0/IndexedBase_14e0.bs 644
./..//bin/x86_64-conda_cos6-linux-gnu-gcc -c  -I"/usr/local/bin" -D_REENTRANT -D_GNU_SOURCE --sysroot=./..//x86_64-conda_cos6-linux-gnu/sysroot -fwrapv -fno-strict-aliasing -pipe -fstack-protector-strong -D_LARGEFILE_SOURCE -D_FILE_OFFSET_BITS=64 -D_FORTIFY_SOURCE=2 -O2   -DVERSION=\"0.00\" -DXS_VERSION=\"0.00\" -fPIC --sysroot=./..//x86_64-conda_cos6-linux-gnu/sysroot "-I/usr/local/lib/5.26.2/x86_64-linux-thread-multi/CORE"   IndexedBase_14e0.c
/bin/sh: ./..//bin/x86_64-conda_cos6-linux-gnu-gcc: not found
make: *** [Makefile:330: IndexedBase_14e0.o] Error 127
```

Also, MAKER apparently needs only Perl modules from perl-bioperl-core, rather than requiring the entire perl-bioperl package.

### DIFF
--- a/recipes/maker/meta.yaml
+++ b/recipes/maker/meta.yaml
@@ -10,7 +10,7 @@ source:
     - "mpi_init.patch"
 
 build:
-  number: 13
+  number: 14
 
 requirements:
   build:
@@ -23,7 +23,7 @@ requirements:
     - snap
     - snoscan
     - perl
-    - perl-bioperl >=1.7
+    - perl-bioperl-core >=1.007002=pl526_1
     - perl-bit-vector
     - perl-dbd-pg
     - perl-dbd-sqlite
@@ -49,7 +49,7 @@ requirements:
     - snap
     - snoscan
     - perl
-    - perl-bioperl >=1.7
+    - perl-bioperl-core >=1.007002=pl526_1
     - perl-bit-vector
     - perl-dbd-pg
     - perl-dbd-sqlite
--- a/recipes/maker/meta.yaml
+++ b/recipes/maker/meta.yaml
@@ -23,7 +23,7 @@ requirements:
     - snap
     - snoscan
     - perl
-    - perl-bioperl-core >=1.007002=pl526_1
+    - perl-bioperl-core >=1.007002
     - perl-bit-vector
     - perl-dbd-pg
     - perl-dbd-sqlite
@@ -49,7 +49,7 @@ requirements:
     - snap
     - snoscan
     - perl
-    - perl-bioperl-core >=1.007002=pl526_1
+    - perl-bioperl-core >=1.007002
     - perl-bit-vector
     - perl-dbd-pg
     - perl-dbd-sqlite
--- a/recipes/maker/meta.yaml
+++ b/recipes/maker/meta.yaml
@@ -77,5 +77,8 @@ test:
     - maker --version
 
 extra:
+  container:
+    # mpich needs /etc/resolv.conf : https://github.com/bioconda/bioconda-recipes/issues/11583
+    extended-base: true
   identifiers:
     - biotools:maker
--- a/recipes/maker/mpi_init.patch
+++ b/recipes/maker/mpi_init.patch
@@ -19,9 +19,42 @@
 +carp "Calling MPI_Init";
 +MPI_Init();
 +
---- src/lib/Parallel/Application/MPI.pm	2014-12-01 23:37:10.000000000 +0100
-+++ src/lib/Parallel/Application/MPI.pm	2018-09-24 15:36:25.167387409 +0200
-@@ -232,7 +232,9 @@
+--- src/bin/maker
++++ src/bin/maker
+@@ -198,6 +198,11 @@ Options:
+      -cpus|c  <integer>  Tells how many cpus to use for BLAST analysis.
+                          Note: this is for BLAST and not for MPI!
+ 
++     -mpi                Forces MAKER to use MPI (overriding MPI autodetection).
++                         Useful if MAKER doesn't correctly detect that it was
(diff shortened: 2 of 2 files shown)


abretaud:
Cool, I'll have a look shortly and make some tests with the Galaxy wrapper
Thanks!

abretaud:
@BiocondaBot recheck

recipes/maker/meta.yaml
```diff
@@ -23,7 +23,7 @@ requirements:
     - snap
     - snoscan
     - perl
-    - perl-bioperl >=1.7
+    - perl-bioperl-core >=1.007002=pl526_1
```
abretaud: Is it necessary to pin the build number too (pl526_1)? I would just write `>=1.007002` (here and on line 52)

nathanweeks:
@abretaud I assume the next biocontainer image would be built with that build, as it's the latest build for that version? If so, it's OK to exclude the build number.

abretaud:
Yep, it should take the latest build. You make the change?

bgruening:
Gracias! Image will be created shortly.

nathanweeks:
Thanks @bgruening and @abretaud ! The resulting Singularity image works great on our cluster. FWIW, if RepeatMasker is used, it requires [this workaround](https://github.com/rmhubley/RepeatMasker/issues/19#issuecomment-491344400) to account for the read-only RepeatMasker/Libraries directory.

PR #15470, opened 2019-05-20
Update mosdepth to 0.2.6.

cbrueffer:

### DIFF
--- a/recipes/mosdepth/meta.yaml
+++ b/recipes/mosdepth/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "mosdepth" %}
-{% set version = "0.2.5" %}
-{% set sha256hash = "3cce77297a190bc24684dac9f1d3bb1281ed79d120572d2d474c21f0942600c4" %}
+{% set version = "0.2.6" %}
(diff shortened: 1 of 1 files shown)


dpryan79:
Xref: https://github.com/bioconda/bioconda-recipes/pull/15495

pmenzel:
The error message from bioconda-test is `/bin/sh: gcc: command not found`.

I also noticed that the [new blast 2.9.0 recipe fails](https://github.com/bioconda/bioconda-recipes/pull/14620/checks?check_run_id=110867192) with `no acceptable C compiler found in $PATH`.

Maybe it's the same underlying issue..

dpryan79:
The only issue is the makefile or similar not respecting the `$CC` variable. We've patched around this in most of our recipes by now, but not all of them. For mosdepth, there are other issues (#15495 ).

pmenzel:
Ok, do you think you could have a look at the blastn 2.9.0 issue, it would be very nice to have it in bioconda, since it has some nice new features and a new database format compared with 2.7

dpryan79:
Sure, is there a PR I can look at?

pmenzel:
Great, it is this one: https://github.com/bioconda/bioconda-recipes/pull/14620

dpryan79:
I've updated this now.

PR #15517, opened 2019-05-21
SearchGUI 4.0.1_SNAPSHOT & PeptideShaker 2.0.1_SNAPSHOT

CarlosHorro:


CarlosHorro:
After so many fixes, this is ready to be merged :-)

recipes/searchgui/meta.yaml
```diff
@@ -1,29 +1,33 @@
...
 build:
-  number: 0
+  number: 1
 
 source:
-  url: http://genesis.ugent.be/maven2/eu/isas/searchgui/{{ name }}/{{ version }}/{{ name }}-{{ version }}-mac_and_linux.tar.gz
-  sha256: da610fa266a2d30cb3ebfd5794e87d0fe7e2105afcd5d88c6a6576ead6e7073a
+#  url: http://genesis.ugent.be/maven2/eu/isas/searchgui/{{ name }}/{{ version }}/{{ name }}-{{ version }}-mac_and_linux.tar.gz
+  url: https://github.com/CarlosHorro/PeptideShaker2-tests/raw/master/SearchGUI-4.0.1-SNAPSHOT-mac_and_linux.tar.gz
+  sha256: 76c0dc1470c6ff549d14f0b3f4ec19f82e02bbdd80c695873ca09074b0c5f88b
```
epruesse: Has this package moved?

recipes/peptide-shaker/meta.yaml
```diff
@@ -1,34 +1,37 @@
 {% set name = "PeptideShaker" %}
-{% set version = "1.16.36" %}
+{% set version = "2.0.1_SNAPSHOT" %}
```
epruesse: Please wait until there is a release. "Snapshot" sounds a lot like something that may change on a whim.

recipes/searchgui/meta.yaml
CarlosHorro: This release is a special release not hosted in the same place than usual. Next official releases will be hosted as usual.

recipes/peptide-shaker/meta.yaml
CarlosHorro: I will explain myself. 
These SearchGUI 4.0.1_SNAPSHOT and PeptideShaker 2.0.1_SNAPSHOT are the result of almost 2 years of work and multiple changes, improvements and some new functionalities in both packages. Almost everything is already done, but the official releases may still need some more months until it is released. But, I am currently available to make as much work as possible, but in some months I will have my paternity leave and I will probably not be available again until February 2019, so it is really important to us to move this work forward as soon as possible in order that, when the official releases are available, the work to be done then is minimal.

These releases are quite stable, and, of course, stable enough to be used for checking all these new characteristics in a real way, and are the key to check and make them also available in the Galaxy tool I am updating at the same time (dependent on this). 
Indeed, something will change in the future with these versions, but not immediately unless something really unexpected happens (which, on the other hand, is also the objective of checking them now).

I hope to have made clear that the release of these versions has not been something capricious but something really necessary to be able to finish the development of both versions (and have the corresponding publication ready in time).

Greetings,
Carlos

PR #15518, opened 2019-05-21
Biotradis: pin samtools to 1.9

sbastkowski:

### DIFF
--- a/recipes/biotradis/meta.yaml
+++ b/recipes/biotradis/meta.yaml
@@ -12,6 +12,7 @@ source:
 
 build:
   number: 4
+  noarch: True
 
 requirements:
   build:
--- a/recipes/biotradis/meta.yaml
+++ b/recipes/biotradis/meta.yaml
@@ -11,8 +11,8 @@ source:
   md5: unused
 
 build:
-  number: 3
-  skip: True # [osx]
+  number: 4
+  skip: True
 
 requirements:
   build:
--- a/recipes/biotradis/meta.yaml
+++ b/recipes/biotradis/meta.yaml
@@ -12,7 +12,6 @@ source:
 
 build:
   number: 4
-  skip: True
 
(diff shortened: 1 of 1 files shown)


epruesse:
@bioconda/core We need a fix for this.

@sbastkowski The sad thing is, while it should work, Conda will choose a lower build number of this package over a lower version of samtools. :-/

dpryan79:
What was the original problem, was installing this downgrading the samtools in an environment?

epruesse:
I understood the PR to be pinning samtools to 1.9 to keep it lower. Maybe I was wrong.

dpryan79:
1.9 is the most recent release of samtools

epruesse:
Ah. Then this should work I think.

epruesse:
@sbastkowski - Can you fix the missing `license_file`?

dpryan79:
> Ah. Then this should work I think.

I'm still not sure what was broken, we should wait for feedback from @sbastkowski

epruesse:
Also, if this is about "at least 1.9", then it should be `>=1.9`. If biotradis works with any samtools version, this PR should be closed. You can force 1.9 or higher into your env by just adding `'samtools>=1.9'` to your `conda install` or `env.yaml'.

sbastkowski:
Hi,
thanks for your help. It seems like for macOS installation the samtools version needs to be very specific (conda install -c bioconda samtools=1.9=h8ee4bcc_1). I think we leave it as it is for now as it seems to work. Potentially needs an additional run of specific samtools installation.

dpryan79:
Do the newer build not work for some reason?

sbastkowski:
As far as I understand, this is a samtools macOS issue. There seems to be some incompatibility.

dpryan79:
Hmm, OK, if you get any details about that then please let us know.

PR #15531, opened 2019-05-22
mzr: try to fix linking error by rebuilding

bernt-matthias:
the so file currently seems to link wrong:

```ldd CONDA_ENV/lib/R/library/mzR/libs/mzR.so
...
	libnetcdf.so.11 => not found
	libhdf5_hl.so.10 => not found
	libhdf5.so.10 => not found
```

### DIFF
--- a/recipes/bioconductor-mzr/meta.yaml
+++ b/recipes/bioconductor-mzr/meta.yaml
@@ -12,7 +12,7 @@ source:
     - 'https://depot.galaxyproject.org/software/bioconductor-{{ name|lower }}/bioconductor-{{ name|lower }}_{{ version }}_src_all.tar.gz'
   md5: 0d04f2321954557784e8080983a7d0e6
 build:
-  number: 0
+  number: 1
   rpaths:
     - lib/R/lib/
     - lib/
--- a/recipes/bioconductor-mzr/meta.yaml
+++ b/recipes/bioconductor-mzr/meta.yaml
@@ -1,6 +1,6 @@
 {% set version = "2.18.0" %}
 {% set name = "mzR" %}
-{% set bioc = "3.9" %}
+{% set bioc = "3.8" %}
 
 package:
   name: 'bioconductor-{{ name|lower }}'
--- a/recipes/bioconductor-mzr/meta.yaml
+++ b/recipes/bioconductor-mzr/meta.yaml
@@ -20,8 +20,8 @@ build:
 # SystemRequirements: C++11, GNU make
 requirements:
   host:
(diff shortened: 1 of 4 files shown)


dpryan79:
The version from bioconductor 3.9 won't be buildable for a few months due to R 3.6.1 having not been released yet. If you need to do something with the recipe, then make a subdirectory for the bioconductor 3.8 version.

recipes/bioconductor-mzr/meta.yaml
```diff
@@ -20,8 +20,8 @@ build:
 # SystemRequirements: C++11, GNU make
 requirements:
   host:
-    - 'bioconductor-biobase >=2.44.0,<2.45.0'
-    - 'bioconductor-biocgenerics >=0.30.0,<0.31.0'
+    - 'bioconductor-biobase'
+    - 'bioconductor-biocgenerics'
```
dpryan79: This is done to comply with how bioconductor pinning should be done (also at the request of the bioconductor core team), please do not undo it. I'll update your PR shortly to add the 3.8 version in.

dpryan79:
The bioconductor 3.9 version will fail, but that can be ignored and I'll force merge the PR if it fixes the problems you're seeing with the version for bioconductor 3.8.

dpryan79:
@bioconda-bot recheck

dpryan79:
@bernt-matthias Can you check the package that the bioconda-bot just posted? If it fixes things I'll force-merge this.

bernt-matthias:
@dpryan79 it works. 

- should we add zlib to the dependencies (seems that it is used during linking)
- should I create subdirectory for the bioconductor 3.8 version.

One technical question: Why is building not complaining about the md5? We switched one of the downloads but did not change the md5...

dpryan79:
There's currently no md5 mismatch for it to complain about. I already created the subdirectory for the 3.8 version (2.16.2). Yeah, it wouldn't hurt to add `zlib` in as a dependency, since you're right that it really is linking against it.

dpryan79:
@bioconda-bot recheck

dpryan79:
I'll merge this in now unless you see anything else that might be an issue.

bernt-matthias:
Failing test seems to be caused by the call to a program called `involucro`. Seems that some parameter needs quoting. 

I'm fine with merging, if a failing test is no problem for you...

dpryan79:
That test passes, the build fails due to missing dependencies in bioconductor 3.9. Anyway, it's merged now and will be available in 30 minutes or so.

PR #15539, opened 2019-05-22
Update to plantcv v3.3.0

nfahlgren:


npavlovikj:
@nfahlgren , your recipe is failing at the Liniting step, and not while building. The error is:
`NotImplementedError: overlap in comment [None, [CommentToken('# [linux]\n', line: 20)]] [None, [CommentToken('# [linux]\n', line: 21)]]`

I am not sure what the error exactly means, but I would try removing `noarch: python` - `noarch` packages are for architecture-independent packages, while yours need to use libGL on linux, and not osx, which is not independent. 

Regarding "libGL", I believe just using the extended-base should work, and you will see if that gives you any error during build afer the linting step is fixed.

npavlovikj:
Ah, please ignore what I said earlier @nfahlgren  - I didn't notice that you were not using any compilers...
In that case, please keep `noarch: python`, and remove `# [linux]` from the `build` section, so you have something like in https://github.com/bioconda/bioconda-recipes/blob/0c10b43ca0b7acdba91527f8573e872507bb4d5c/recipes/bioconductor-rdrtoolbox/meta.yaml.

nfahlgren:
Thanks @npavlovikj, the comments were definitely causing linting to fail. I updated the PR and the CircleCI log now shows the build error I was talking about.

nfahlgren:
Awesome, thanks @npavlovikj!

npavlovikj:
@BiocondaBot merge

PR #15569, opened 2019-05-24
OpenSSL build fix.

t-neumann:
* [ x ] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
* [ x ] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).

### DIFF
--- a/recipes/slamdunk/meta.yaml
+++ b/recipes/slamdunk/meta.yaml
@@ -14,7 +14,7 @@ build:
     - slamdunk=slamdunk.slamdunk:run
     - alleyoop=slamdunk.alleyoop:run
     - splash=slamdunk.splash:run
-  number: 0
+  number: 1
   skip: True # [py3k or osx]
 
 requirements:
@@ -42,6 +42,7 @@ requirements:
     - curl >7.59
     - xz
     - libdeflate
+    - openssl=1.0
   run:
     - python
     - joblib >=0.9.4
@@ -61,6 +62,7 @@ requirements:
     - curl >7.59
     - xz
     - libdeflate
+    - openssl=1.0
 
 test:
   imports:
--- a/recipes/slamdunk/meta.yaml
+++ b/recipes/slamdunk/meta.yaml
@@ -39,10 +39,10 @@ requirements:
     - ncurses
     - zlib
     - bzip2
-    - curl >7.59
+    - curl
     - xz
     - libdeflate
-    - openssl=1.0
+    - openssl
   run:
     - python
     - joblib >=0.9.4
@@ -59,10 +59,10 @@ requirements:
     - ncurses
     - zlib
     - bzip2
-    - curl >7.59
+    - curl
     - xz
     - libdeflate
-    - openssl=1.0
+    - openssl
 
 test:
   imports:


t-neumann:
@druvus Fixed the outdated branch now.

druvus:
@t-neumann is this PR based on this issue t-neumann/slamdunk#48 ?

I think you may remove your openssl pinning. The issue should be fixed by rebuilding the recipe as the current version were built before transition to the current openssl 1.1.1 version. 

@dpryan79 could you jump in here as you did the openssl migration.

recipes/slamdunk/meta.yaml
```diff
@@ -14,7 +14,7 @@ build:
     - slamdunk=slamdunk.slamdunk:run
     - alleyoop=slamdunk.alleyoop:run
     - splash=slamdunk.splash:run
-  number: 0
+  number: 1
   skip: True # [py3k or osx]
```
druvus: Any reason for not building for py3k or osx? It seems to build fine locally on my mac.

recipes/slamdunk/meta.yaml
```diff
@@ -42,6 +42,7 @@ requirements:
     - curl >7.59
     - xz
     - libdeflate
+    - openssl=1.0
```
druvus: This should probably be removed and just fixed by rebuilding as it is pinned centrally.

druvus:
@t-neumann I also observed that the generated packages are quite bit (200Mb). I am not familiar to the tool but it seems to mainly be due to compiling and packaging external tools. Could the setup.py be modified so you depend on conda version of samtools/varscan/etc. instead of building them.

recipes/slamdunk/meta.yaml
dpryan79: This needs to be removed. Openssl is pinned via the conda-forge-pinning repository.

dpryan79: Please remove the openssl pinning, as well as the version restriction for `curl`. Both of these should be coming from conda-forge-pinning.

t-neumann: @dpryan79 The `curl` pinning was originally introduced by @bgruening to fix the latest build. So you are telling me removing the `curl` and `openssl` pinning and creating a new build number should fix all issues? The current build is definitely not working for people due to the newer `openssl` version.

recipes/slamdunk/meta.yaml
t-neumann: For earlier versions, the osx and py3k builds never worked, so I decided to exclude them. Have never tried since then, but if you tell me they work for you, I can definitely include them again.

recipes/slamdunk/meta.yaml
dpryan79: Yes, remove the pinning and just bump the build number. That should resolve the issue.

t-neumann:
@dpryan79 Build number bumped and `openssl` and `curl` pinning removed.

dpryan79:
@bioconda-bot recheck

dpryan79:
Can you try installing the package that the bioconda-bot just posted and see if that fixes things?

t-neumann:
@dpryan79 I tried but it is forever stuck in the environment resolving stage:

```
conda install -c https://59042-42372094-gh.circle-artifacts.com/0/tmp/artifacts/packages slamdunk
Solving environment: -
```

dpryan79:
I suspect that either you're running that in an environment with a bunch of other things or that your channel order is wrong. On my laptop it's done in a few seconds.

dpryan79:
Try:

    conda create -n foo
    conda activate foo

And then install slamdunk. If that's still slow, then there's something wrong with your channels.

t-neumann:
@dpryan79 Thanks that did the trick - looks all good now.

PR #15605, opened 2019-05-27
Unicycler 0.4.7 - b1 - fixes

apeltzer:

This is an attempt to fix some of the weird errors seen in Unicycler 0.4.6 and 0.4.7 when using bioconda, e.g. https://github.com/rrwick/Unicycler/issues/140 

I tried tracking this down so far but wasn't successful until now - I think the change I removed here in the build recipe *could* potentially get the issue fixed as reverting to 0.4.4 is also not really an option for some people (e.g. when having/requiring multiple tools in the same container installed).


PertuyF:
@apeltzer please see https://github.com/rrwick/Unicycler/issues/140#issuecomment-496222163.
This change was made for a reason, and attempts were made to try fix this isue, be sure you get is all before trying this.

PertuyF:
Looks like `unicycler` now builds fine with the genuine `seqan 2.3` included in the original repo. Which would suggest the fix introduced in PR https://github.com/bioconda/bioconda-recipes/pull/10510 is no longer required.
Am I missing something from the build log or does it look right @corburn @bgruening ?

Could you test this package and confirm you do not experience the stalling issue before we consider merging @apeltzer ? any help appreciated in testing this.

apeltzer:
Nice one! Thanks a bunch, I'm rebuilding my docker container for nf-core/bacass and will then start a new run to see if this changes things :-) !

apeltzer:
Couple of issues: The recipe uploaded was built using python3.7:

https://anaconda.org/bioconda/unicycler/files

Although explicitly marked to be not built using this version? 

https://github.com/bioconda/bioconda-recipes/blob/master/recipes/unicycler/meta.yaml

PertuyF:
It could be due to python pinning at the Bioconda scale. Older versions are dropped progressively, so if python 3.6 is not supported anymore it may have forced 3.7.
@bioconda/core am I right? And is there a way to have a lint warning for this, like "Recipe pinning does not match bioconda's pinning for package xxx"?

Is 3.7 an issue for you @apeltzer?

apeltzer:
Unfortunately yes :-( 

I'm trying to build a pipeline and many of the tools required are build with older versions of Python3.X, meaning I can't really install these in the same environment anymore due to Python3.7 being used for the new build of unicycler 0.4.7 - the old (incorrect recipe) gets installed then automatically, which breaks my pipeline due to the stalling issue mentioned above.

apeltzer:
One could bump all other recipes to Python3.7 however, but not sure that all of these will be built correctly then :-(

PertuyF:
In most cases rebuilding recipes for a new python 3.x should be painless.
But if required there may be a way to force rebuild unicycler for 3.6. Although I'm not certain.

dpryan79:
For C/C++ packages, you need to put a `skip: True  # [something]` in the build section for actually skip building. Otherwise I think conda ignores it. Regarding rebuilding dependencies for python 3.7, that should generally work fine (at least if they already have a python 3.6 build). I should probably start that in the `bulk` branch so you don't have to deal with individual packages.

apeltzer:
Would be great - its just a bunch of recipes though:

https://github.com/nf-core/bacass/blob/092a082fb08076aed5affffe7d092b2d3d7b75a1/environment.yml

apeltzer:
* MultiQC
* Quast
* Nanopolish
* Racon
* Porechop
* Nanoplot
* Pycoqc 

(I think the rest is already in Python3.7 on conda-forge or bioconda)

dpryan79:
Are those the ones needing to be rebuilt? At least a few of them are `noarch: python` (e.g., multiqc), so they should be fine already.

apeltzer:
Fair point - I probably should definitely update the `quast` recipe, but the others could be fine. 

Yes, that's the complete list of Python packages/python dependent packages.

dpryan79:
Yeah, the only one of those that needs rebuilding is quast. Note that that had compilation errors the last time I tried to bump everything for new pinnings, so it'll need some significant TLC.

epruesse:
Quast also has a serious data package issue. Or had, last time I checked. It downloads SILVA somewhere in the middle, unpacks it, then renames the sequences and builds a BLAST index. So the package is incomplete, uses a *different* dataset than plain SILVA, modified in a trivial and technically uncessary manner, and breaks with clobbered data if you run it in parallel on a cluster. (end rant...)

apeltzer:
I am quite tempted to drop that tool just because of this. Are there alternatives that can do similar things-did a quick google search but couldn't find much unfortunately... ?

epruesse:
Not that I am aware of. Double check current versions. It's not like this would be hard to fix.

PR #15613, opened 2019-05-27
glimmerHMM bump build to 3; move required perl modules into bin/

nextgenusfs:

GlimmerHMM training script is a perl script that has modules missing in the current build. Updated build.sh to include those perl module files.

### DIFF
--- a/recipes/glimmerHMM/build.sh
+++ b/recipes/glimmerHMM/build.sh
@@ -14,10 +14,11 @@ cd sources && make
 cd ../train && make
 cd ..
 
-#copy the execulatbles	
+#copy the executables and perl modules needed for training
 cp bin/glimmhmm.pl $PREFIX/bin/.
 cp sources/glimmerhmm $PREFIX/bin/.
 cp train/trainGlimmerHMM $PREFIX/bin/.
+cp train/*.pm $PREFIX/bin/.
 
 # copy the training data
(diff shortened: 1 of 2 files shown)


nextgenusfs:
This failed -- but seems that more is required than just the perl modules.  There are several "binaries" in the train directory that are also required, so not sure what the best way is to move them, i.e. might make more sense to retain the /train/ directory somewhere and then just symlink the trainGlimmerHMM script.

zachcp:
hi @nextgenusfs , are you testing the build independently of the PR? Currently it looks like the build is failing a linting due to 

1. "glimmerHMM" in the recipe folder vs "glimmerhmm" in the meta.yaml.   For tha I would sugges lower-casing the folder name as is the convention on the other folder. 

2. g++ is not found in the machine. This probably hasn't been rebuilt since the bioconda overhaul so we may need to specify the compiler or patch the Makefile.

nextgenusfs:
I haven’t had time to look through all of the changes in the build docs with compilers, etc. I had hoped it would just work... coupled with build problems I’ve had in my Mac more recently I tried to just update it quickly and that obviously failed.

dpryan79:
Was this all taken care of in #15746?

nextgenusfs:
Yes I will close this one.

PR #15622, opened 2019-05-28
Add snptools recipe

winni2k:

### DIFF
--- a/recipes/snptools/build.sh
+++ b/recipes/snptools/build.sh
@@ -1,8 +1,23 @@
 #!/usr/bin/env bash
 
-sed -i.bak 's/INCLUDE=-Isamtools-0.1.16 -Itabix-0.2.5//; s/-Lsamtools-0.1.16 -Ltabix-0.2.5//' makefile
+cd tabix-0.2.5
+make lib
+cd -
 
-make all CC=$CXX
+cd samtools-0.1.16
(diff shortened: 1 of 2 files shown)


dpryan79:
Does this really need such an ancient samtools? I think the OSX error is due to a random object file not being cleaned up before building.

winni2k:
@dpryan79 Thanks for your input. The short answer is: I don't know. 

Do you happen to know if there is a newer samtools version that does not change behavior/API compared to 0.1.16?

Also, are you implying that the old version of samtools is the cause of the "random object" not being cleared up?

dpryan79:
Do you know if snptools is just running samtools or if it actually tries to link against parts of it? I think 0.1.19 had the same syntax, but I've only been in bioinformatics since about samtools 0.1.18, so can't say with certainty.

winni2k:
It's definitely linking to samtools and tabix: https://github.com/winni2k/snptools/blob/master/makefile#L4

I am very hesitant to bump any versions on this code because it does not come with any tests. However, I have manually tested this exact version extensively.

dpryan79:
yikes

winni2k:
Story of my PhD -,-

daler:

daler:

daler:

PR #15624, opened 2019-05-28
Add tabix library to tabix recipe

winni2k:


jmarshall:
What is the reason for wanting to use historical libraries that have not been maintained in over five years, rather than equivalents from HTSlib?

winni2k:
I am trying to conda package an old tool (`snptools`, last released in 2011) that links against this exact version of the tabix library. `snptools` is unmaintained and has no regression tests, so I am keen to link against exactly the same code that was used in the past.

PR #15630, opened 2019-05-28
bioconductor-mzr: new build for 2.10.0

bernt-matthias:
Follow up from https://github.com/bioconda/bioconda-recipes/pull/15531 . For the current Galaxy wrapper for profia the 2.10.0 version in needed. Currently mzr-2.10.0-boost1.64_0 is installed (the dependency chain is bioconductor-profia-1.4.0-r3.4.1_0 .. bioconductor-xcms-1.50.1-r3.4.1_0 and the bioconductor-mzr). 

Would be really good to have a working recipe for reproduibility. Currently it fails with 

```
Error: package or namespace load failed for ‘mzR’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/home/berntm/miniconda3/envs/mulled-v1-0aac71eabd627c852daf922fc2ada2ca25012162d88fa88e2fe10c1f0dc1234a/lib/R/library/mzR/libs/mzR.so':
  libnetcdf.so.11: cannot open shared object file: No such file or directory
```

For this recipe I manual merged version form d3ed749dc0c0b1a3e4e037773b8ce5ed18fe80f1
with the current state in the 2.16.2.

Don't know if we should pin the versions of all bioconductor dependencies as it is the case for the current recipes.

@dpryan79 sorry for the extra work .. somehow I mixed up versions. 

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -2265,7 +2265,6 @@ recipes/bioconductor-mwastools
 recipes/bioconductor-mwgcod.db
 recipes/bioconductor-mygene
 recipes/bioconductor-myvariant
-recipes/bioconductor-mzr
 recipes/bioconductor-nadfinder
 recipes/bioconductor-nanostringdiff
 recipes/bioconductor-nanostringqcpro
--- a/recipes/bioconductor-mzr/2.10.0/build.sh
+++ b/recipes/bioconductor-mzr/2.10.0/build.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+mv DESCRIPTION DESCRIPTION.old
+grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
+mkdir -p ~/.R
+echo -e "CC=$CC
+FC=$FC
+CXX=$CXX
+CXX98=$CXX
+CXX11=$CXX
+CXX14=$CXX" > ~/.R/Makevars
+$R CMD INSTALL --build .
--- a/recipes/bioconductor-mzr/2.10.0/meta.yaml
+++ b/recipes/bioconductor-mzr/2.10.0/meta.yaml
@@ -24,7 +24,7 @@ requirements:
     - bioconductor-protgenerics
     - bioconductor-zlibbioc
     - libnetcdf
-    - r-base
+    - 'r-base ==3.4.1'
     - r-ncdf4
     - 'r-rcpp >=0.10.1'
     - zlib
@@ -35,7 +35,7 @@ requirements:
     - bioconductor-protgenerics
     - bioconductor-zlibbioc
     - libnetcdf
-    - r-base
+    - 'r-base ==3.4.1'
     - r-ncdf4
     - 'r-rcpp >=0.10.1'
     - zlib
--- a/recipes/bioconductor-mzr/2.10.0/meta.yaml
+++ b/recipes/bioconductor-mzr/2.10.0/meta.yaml
@@ -1,0 +1,68 @@
+{% set version="2.10.0" %}
+{% set name="mzR" %}
+{% set bioc="3.5" %}
+
+package:
+  name: 'bioconductor-{{ name|lower }}'
+  version: '{{ version }}'
+source:
+  url:
+    - 'https://bioconductor.org/packages/{{ bioc }}/bioc/src/contrib/{{ name }}_{{ version }}.tar.gz'
+    - 'https://bioarchive.galaxyproject.org/{{ name }}_{{ version }}.tar.gz'
+    - 'https://depot.galaxyproject.org/software/{{ name }}/{{ name }}_{{ version }}_src_all.tar.gz'
+  sha256: a776226cac45b19ea622dd6a4765e3c3052128aa031b3ff3dac20231a02139fe
+build:
+  number: 1
+##  string: boost{{CONDA_BOOST}}_{{PKG_BUILDNUM}}
+  rpaths:
+    - lib/R/lib/
+    - lib/
+requirements:
+  host:
+    - bioconductor-biobase
+    - 'bioconductor-biocgenerics >=0.13.6'
+    - bioconductor-protgenerics
+    - bioconductor-zlibbioc
+    - libnetcdf
+    - r-base
+    - r-ncdf4
+    - 'r-rcpp >=0.10.1'
+    - zlib
+    - boost
+  run:
+    - bioconductor-biobase
+    - 'bioconductor-biocgenerics >=0.13.6'
+    - bioconductor-protgenerics
+    - bioconductor-zlibbioc
+    - libnetcdf
+    - r-base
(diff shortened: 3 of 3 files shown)


dpryan79:
Can't you update the galaxy wrapper to use a more recent profia? It'll be rather painful to build anything for R 3.4.1 anymore, since a lot of the stuff has been pulled from the main conda-forge channels.

bernt-matthias:
I understand your concerns and indeed the an upgrade has been discussed. But with Galaxies goal for reproducibility it seems important that a version that worked once keeps working.

dpryan79:
Fair enough and I share Galaxy's goal. I have a feeling this will require adding a `conda_build_config.yaml` and possibly building this locally with added channels (namely, `conda-forge/cf201901`) :(

bernt-matthias:
To my surprise the local tests just worked.

Currently the unpinned dependencies get resolved as follows:

- bioconductor-biobase #2.42.0
- 'bioconductor-biocgenerics >=0.13.6' #0.28.0
- bioconductor-protgenerics #1.16.0
- bioconductor-zlibbioc #1.30.0
- r-base #3.5.1
- r-ncdf4 #1.16.1
- 'r-rcpp >=0.10.1' #1.0.1

Do you see any reason to pin them now?

dpryan79:
For such an old version that's probably fine, since only people using the galaxy wrapper would ever actually use this. Just note that the packages may break in weird ways due to not being designed to work together.

bernt-matthias:
ping @sneumann, @lecorguille : we are trying to get this version of mzR to work again. It is a dependency of bioconductor-xcms-1.50.1-r3.4.1_0 which again is a dependency of bioconductor-profia-1.4.0-r3.4.1_0. The last one is used in the current profia galaxy wrapper. 

Do you have a suggestion if we should pin the versions of the other dependencies (in particular the bioconductor and R ones) and if so to which versions? Maybe one of you has somewhere a working bioconductor-profia-1.4.0 environment and could post the versions installed there...

@dpryan79 the failing test seems again be caused by `involucro`. Can or should we fix this?

bgruening:
@bernt-matthias isn't the error: `conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform linux-64: {"bioconductor-biocgenerics[version='>=0.30.0,<0.31.0']", "bioconductor-biobase[version='>=2.44.0,<2.45.0']"}`

bernt-matthias:
@bgruening this error comes from the mzr version in the main directory. Is it possible to blacklist the version in the main directory and still build the 2.10 and 2.16 versions.

dpryan79:
If we blacklist the main one we blacklist all of the subdirectories too. Only worry about the CI tests for the old bioconductor 3.5 version passing.

sneumann:
Hi, as with the previous issues with mzR in Bioconda, I am overwhelmed 
by what exactly to do to reproduce, locally I am still on an ancient miniconda :-(
So you're trying to build mzR-2.10.0 http://bioconductor.org/checkResults/3.5/bioc-LATEST/mzR/
which is this version https://github.com/sneumann/mzR/tree/815dd3606e8a66d2a7ca1b70cdf07b9439b4bc54
This 2.10.0 is linking against *system* netcdf libraries, *not* linking against r-ncdf4 or r-hfb5lib as later ones.

sneumann:
I booted an ancient laptop and it has ubuntu 12.04 with mzR-1.9.5 
and it has libnetcdf-dev version 4.1.1 Maybe that helps ?

bgruening:
@bernt-matthias for this specific problem, do you think you can add libnetcdf to the Galaxy wrapper. If you find the correct version is should work.

bernt-matthias:
Am I right that `TEST SUCCESS recipes/bioconductor-mzr/2.10.0` actually indicates success? 

And the only thing that we need to worry about is 

```
11:26:17 BIOCONDA INFO (OUT) WARNING (bioconductor-mzr,lib/R/library/mzR/libs/mzR.so): Needed DSO lib/libnetcdf.so.13 found in ['libnetcdf']
11:26:17 BIOCONDA INFO (OUT) WARNING (bioconductor-mzr,lib/R/library/mzR/libs/mzR.so): .. but ['libnetcdf'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)
```

I was hoping that `r-ncdf4` in the `host` section would be sufficient since it requites `libnetcdf`. Should I list all packages explicitly, even if the dependency is fulfilled indirectly?

dpryan79:
@bioconda-bot recheck

dpryan79:
@bernt-matthias Yeah, that version is actually passing tests. It's building against R 3.5, so I can't guarantee that nothing will randomly break.

dpryan79:
Regarding that warning about libnetcdf, yes please add it explicitly to the recipe.

bernt-matthias:
OK done. Added libnetcdf as host and run requirement. 

One question out of interest. How is conda ensuring that the same version of libnetcdf is installed that was used during compilation?

bgruening:
The state-of-the-art in conda package design is that libraries expose runtime dependencies on their own with run-exports. You can see it here: https://github.com/conda-forge/libnetcdf-feedstock/blob/master/recipe/meta.yaml#L52

So if you use libraries like libnetcdf or boost these should have run-exports defined and will export the appropriate packages on their own as run-dependencies.

dpryan79:
Once this passes testing, can you try injecting it into your mulled environment in Galaxy and see if it solves the problem? We can merge it in then.

bernt-matthias:
@bioconda-bot recheck

bernt-matthias:
Hmm. I don't understand why

```conda create -y --quiet --override-channels --channel iuc --channel conda-forge --channel https://59930-42372094-gh.circle-artifacts.com/0/tmp/artifacts/packages --channel bioconda --channel defaults --name mulled-v1-0aac71eabd627c852daf922fc2ada2ca25012162d88fa88e2fe10c1f0dc1234a r-batch=1.1_4 bioconductor-profia=1.4.0```

installs: bioconda::bioconductor-mzr-2.10.0-boost1.64_0

Any ideas?

**Comment by dpryan79** [comment]
That version of profia was built against R 3.4.1, you've built mzr here against R 3.5.1

**Comment by dpryan79** [comment]
Your choices are either to rebuild profia (and all of its dependencies) for more recent pinnings, or to rebuild mzr against the old pinnings but with the fixed netcdf4 pinning. I suspect the latter will be easier, though I suspect you'll find it easier to do locally rather than via PRs.

**Comment by bernt-matthias** [comment]
Hey, just read on gitter that R 3.6 is available. Does this simplify matters here?

**Comment by dpryan79** [comment]
At the moment this will get built against R 3.6, which is likely to break spectacularly. You'll need to add a custom conda_build_config.yaml with the following:

```
r_base:
  - 3.4.1
```

That's more likely to work.

PR #15635, opened 2019-05-29
fixing scipy and sklearn dependencies

wdecoster:

Adding missing dependencies scipy and sklearn to pauvre

### DIFF
--- a/recipes/pauvre/build.sh
+++ b/recipes/pauvre/build.sh
@@ -1,3 +1,0 @@
-#!/bin/bash
-
-$PYTHON setup.py install --single-version-externally-managed --record=record.txt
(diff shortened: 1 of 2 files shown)


wdecoster:
Hmm haven't came across `noarch` recipes before.

So I remove all dependencies, from `host` but not from `run`? 

And under build I add:  
`script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv"`? (based on changes I saw here: https://github.com/bioconda/bioconda-recipes/blob/master/recipes/nanocomp/meta.yaml)

Although that meta.yaml lists setuptools rather than pip, but perhaps that doesn't matter as much. 

And a `build.sh` is no longer necessary?

bgruening:
Nice! Thanks @wdecoster!

PR #15665, opened 2019-05-31
Build 0 for portcullis 1.2.0

lucventurini:

### DIFF
--- a/recipes/portcullis/meta.yaml
+++ b/recipes/portcullis/meta.yaml
@@ -39,6 +39,7 @@ requirements:
   run:
     - python
     #- zlib
+    - boost
     - tabulate
     - numpy
     - pandas
--- a/recipes/portcullis/meta.yaml
+++ b/recipes/portcullis/meta.yaml
@@ -28,8 +28,6 @@ requirements:
     - automake >=1.15.1
     - libtool
     - perl
-    - boost
-    - zlib
   host:
     - zlib
     - pandas
@@ -38,7 +36,6 @@ requirements:
     - boost
   run:
     - python
-    #- zlib
     - boost
     - tabulate
     - numpy
--- a/recipes/portcullis/meta.yaml
+++ b/recipes/portcullis/meta.yaml
@@ -33,10 +33,10 @@ requirements:
     - pandas
     - python
     - samtools >=1.9
-    - boost
+    - boost-cpp
   run:
     - python
-    - boost
+    - boost-cpp
     - tabulate
     - numpy
     - pandas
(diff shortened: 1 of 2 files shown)


npavlovikj:
Try setting the flags for `zlib` errors, https://bioconda.github.io/troubleshooting.html#zlib-errors.

lucventurini:
Thanks, I will try tomorrow!

maplesond:
I had a look at this, and reenabled the build environment flags as it appears these are required for zlib.  Still encountering the same boost issue though.  Are there any docs for using bioconda boost?  Build and runtime linking for boost works fine on most systems with a version of boost installed from a package manager.

lucventurini:
Dear @npavlovikj , any insight on this?

npavlovikj:
@lucventurini , can you please try adding these two lines in "build.sh":
```
export LIBS="-lm ${PREFIX}/lib/libz.a"
 ./configure --disable-silent-rules --disable-dependency-tracking --prefix=$PREFIX --with-boost=${PREFIX} --with-boost-libdir=${PREFIX}/lib
```
?

npavlovikj:
Also, `boost` needs to be added to the `run` requirements.

lucventurini:
Hi @npavlovikj, thank you for the suggestions, much appreciated! Test currently underway.

lucventurini:
Dear @npavlovikj , it seems to have worked! If you could please authorize the merging, we would ever be so grateful.

recipes/portcullis/meta.yaml
```diff
@@ -28,14 +28,18 @@ requirements:
     - automake >=1.15.1
     - libtool
     - perl
+    - boost
```
bgruening: I don't think boost and zlib should be in build. host should be enough

lucventurini: Updated according to the suggestion. Waiting for the build to complete.

lucventurini:
Hi @bgruening , I have updated the recipe according to your suggestion (ie removing boost and zlib from the "build" section). Everything seems to be working, if so, we would be grateful if you could authorise the merge! 
Kind regards

recipes/portcullis/meta.yaml
```diff
@@ -33,9 +33,10 @@ requirements:
     - pandas
     - python
     - samtools >=1.9
+    - boost
   run:
     - python
-    #- zlib
+    - boost
```
epruesse: I think you want `boost-cpp`, not `boost`.

lucventurini: Testing now with boost-CPP. If it passes, would it be possible to merge into master, please?

lucventurini:
@epruesse yes please re: being added as the maintainer. Thank you for taking the time to review this recipe!

epruesse:
@lucventurini To do that, just add the fragment I posted above to the meta.yaml next time you update the recipe.

PR #15698, opened 2019-06-03
RNA-SeQC v2.x.x

stekaz:

Code review requested, please @bioconda/core


recipes/rna-seqc/meta.yaml
```diff
@@ -1,27 +1,28 @@
...
+  url: '{{base}}/v{{version}}/rnaseqc.v{{version}}.linux.gz'   # [linux]
+  sha256: '{{sha256linux}}'                                    # [linux]
 
-build:
-  number: 2
-  noarch: generic
+  url: '{{base}}/v{{version}}/rnaseqc.v{{version}}.osx.gz'     # [osx]
+  sha256: '{{sha256osx}}'                                      # [osx]
 
-requirements:
```
bgruening: You need to define the requirements. If this is now a python only package you should have python as run-requirement and pip and python as host requirement.

stekaz: Thanks very much for the review @bgruening. These are just static executables/binaries so there are no host or run requirements. I am of course happy to add an empty requirements block if needed, but I think that is redundant.

bgruening: Is it possible to build it from source? We recommend building from source to stay compatible with various old architectures.

stekaz: Yes - there's a "full source" tarball that includes the SeqLib submodule/dependency and some test data, IIRC. I'm happy to give this a go tomorrow. Cheers.

stekaz: Sorry - I think I see the problem with the test. Will try and fix tonight/tomorrow and let you know when I'm done.

stekaz: Done; but a code review for this guy would be very much appreciated. There are also a couple of python scripts that come bundled alongside the source and these have quite a few dependencies. Should these also be included in the build?

dpryan79:
@bioconda-bot update

recipes/rna-seqc/meta.yaml
dpryan79: The question is mostly how vital those are. A lot of tools come with random R/python helper scripts that are never used. We can focus on getting the primary package working and fix helper scripts as the need arises.

dpryan79:
@bioconda-bot merge

PR #15701, opened 2019-06-03
Update Circos and Circos-Tools with current perl version

apeltzer:

Bumping both circos and circos-tools to get newer Perl (5.26+) to be used for rebuild.

### DIFF
--- a/recipes/circos/meta.yaml
+++ b/recipes/circos/meta.yaml
@@ -11,6 +11,7 @@ source:
 
 build:
   number: 5
+  noarch: generic
 
 requirements:
   run:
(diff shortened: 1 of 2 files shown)


recipes/circos/meta.yaml
```diff
@@ -10,7 +10,7 @@ source:
   sha256: 52d29bfd294992199f738a8d546a49754b0125319a1685a28daca71348291566
 
 build:
-  number: 4
+  number: 5
```
dpryan79: Can you make these both `noarch: generic`? That should future-proof things.

dpryan79:
I have a feeling that this won't end up solving the problem you're running into (though it'll help). The CI testing seems to be using an older version of perl now, which suggests that there may be a dependency of these that are the real problem. But I guess we can merge this in and see.

apeltzer:
One step closer maybe though ;-)

Issue #15332, opened 2019-05-13
Requiring pull request reviews


dpryan79:
@bioconda/all When the Bioconda project was originally created it needed to be extremely simple for anyone and everyone to add new recipes and modify old ones, since there was no base of recipes to build upon. Today, Bioconda has 500+ members, approximately 1000 contributors, almost 7000 packages and has become a primary part of the global bioinformatics infrastructure. Given this, we need to ensure  that new and modified recipes adhere to best practices such as those outlined at https://bioconda.github.io/guidelines.html. To do so, we would like to begin requiring pull request reviews for almost all bioconda members beginning June 1st. Any bioconda member (not just @bioconda/core!) can review and approve pull requests, so we expect the turn-around time on pull requests to continue to be low.

After June 1st, please use the “please review & merge” label for PRs once you’re satisfied with them and would like to request a review. It’s very easy to [peruse such labeled PRs](https://github.com/bioconda/bioconda-recipes/labels/please%20review%20%26%20merge), so we hope this will be adopted by the community. You’re additionally welcome to post a link to your PR on [Gitter](https://gitter.im/bioconda/Lobby) or ping one of the expert [teams](https://github.com/orgs/bioconda/teams) or persons, both for getting help and for getting a review. We hope that this will also increase interaction between community members, improve exchange of knowledge and foster best-practices.

To review a pull request with the "please review & merge" label, check to make sure it satisfies the [guidelines](https://bioconda.github.io/guidelines.html) and kindly point out any improvements that could be made (and don't forget to adhere to our [code of conduct](https://github.com/bioconda/bioconda-recipes/blob/master/CODE_OF_CONDUCT.md)). Feel free to pull in Bioconda's expert teams to take a look, or call in specific people you know could help. Once it looks good, simply [mark the pull request as approved](https://help.github.com/en/articles/approving-a-pull-request-with-required-reviews). After that, either you or the submitter can merge it (either directly, or by typing `@BiocondaBot merge`).

We welcome feedback on this proposal before June 1st, as we would like Bioconda to continue to be the vibrant and open community that it is today.

apeltzer:
Makes a lot of sense - also, to get feedback for the one submitting PRs!

holtgrewe:
Would it be possible to get an exception for trivial changes?

[...]

PR #15786, opened 2019-06-09
Guessmylt recipe

Juke34:


recipes/guessmylt/build1.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+ setup.py install --single-version-externally-managed --record=record.txt
```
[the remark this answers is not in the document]
Juke34: done

recipes/guessmylt/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  version: "{{ version }}"
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: b0b5da19e762a3e4a60b913c7f6ab398c2000251bb6d1f8b5ea4c2909a6183b2
+
+build:
+  entry_points:
+    - GUESSmyLT = GUESSmyLT.GUESSmyLT:main
+    - GUESSmyLT-example = example_run.example:main
```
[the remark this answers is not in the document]
Juke34: fixed

recipes/guessmylt/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  host:
+    - python>3
+    - bcbio-gff ==0.6.4
+    - biopython ==1.67
+    - pip
+    - pysam >=0.13.0
+    - snakemake ==5.4.0
+    - trinity ==2.8.4
+    - busco ==3.0.2
+    - bowtie2 ==2.3.4.3
```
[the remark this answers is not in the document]
Juke34: tool version relaxed for snakemake and bowtie2

recipes/guessmylt/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - GUESSmyLT
+    - example_run
+  commands:
+    - GUESSmyLT --help
+    - GUESSmyLT-example
+
+about:
+  home: https://github.com/NBISweden/GUESSmyLT
+  license: GPLv3
+  license_family: GPL
```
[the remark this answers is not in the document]
Juke34: LICENSE file included within the package dist by adding it in the MANIFEST.
So added also here in yams file. Hoping it will work as expected

recipes/guessmylt/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  summary: Software to guess the RNA-Seq library type of paired and single end read files using mapping and gene annotation.
```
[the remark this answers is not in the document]
Juke34: summary shortened

recipes/guessmylt/meta.yaml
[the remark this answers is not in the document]
HadrienG: you may have broken the import though:

```
10:57:29 BIOCONDA INFO (OUT)   File "/opt/conda/conda-bld/guessmylt_1560249820032/test_tmp/run_test.py", line 5, in <module>
10:57:29 BIOCONDA INFO (OUT)     import example_run
10:57:29 BIOCONDA INFO (OUT) ModuleNotFoundError: No module named 'example_run'
10:57:29 BIOCONDA INFO (OUT) import: 'GUESSmyLT'
10:57:29 BIOCONDA INFO (OUT) import: 'example_run'
10:57:31 BIOCONDA INFO (OUT) Tests failed for guessmylt-0.2.2-py_0.tar.bz2 - moving package to /opt/conda/conda-bld/broken
```

Remove l45 in test/imports too to fix:

```
test:
  imports:
    - GUESSmyLT
    - example_run
```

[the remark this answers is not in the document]
Juke34: done

Juke34:
The package builds successfully on Linux architecture now (I had to modify snakemake pipeline code at some point. Was hard to catch because only automated build was failing, while in the same container in parallel activating the same env built by bioconda-build I was not able to reproduce the fail...). 
BUSCO does not behave the same way on OSX, so I would like to avoid OSX build. But it doesn't work with the skip osx platform tag.
@bioconda/osx any idea how to deal with testing and building only on linux?

epruesse:
Just put a lint exclude with a comment behind explaining why.

Juke34:
I don't get how works the **lint exclude**. The **skip** mechanism does not work in conjunction with **noarch**. So you mean I can keep both in the yaml file and use **lint exclude** in a way to avoid the `Linting Recipe` to fail?

recipes/guessmylt/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  home: https://github.com/NBISweden/GUESSmyLT
+  license: GPLv3
+  license_family: GPL
+  license_file: LICENSE
+  summary: Software to guess the RNA-Seq library type.
+  dev_url: https://github.com/NBISweden/GUESSmyLT
+
+extra:
+  recipe-maintainers:
+    - juke34
```
HadrienG: Here in `extra` you should add your `skip-lints:` block

Something like:

```
skip-lints:
  - check_noarch
```

recipes/guessmylt/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - pysam >=0.13.0
+    - snakemake ==5.4.*
+    - trinity ==2.8.4
+    - busco ==3.0.2
+    - bowtie2 >=2
+  run:
+    - python>3
+    - git >=2.11.1
+    - bcbio-gff ==0.6.4
+    - biopython ==1.67
```
HadrienG: I'd still really like you to use `>=`.

As a general rule when your are publishing an application you should not pin exact versions, excepted when you know it breaks with the most recent.

Juke34: In another tool I had problem between bcbio-gff and biopython versions. I know this two versions work perfectly well together. I should do a test before accepting more recent versions.

recipes/guessmylt/meta.yaml
Juke34: Thx I will give a try

HadrienG: my bad, try with `should_not_be_noarch_skip`

Juke34: Ok I get how it is supposed to work then, I will give a try

Juke34: It works fine for the first linting, but it doesn't work for the linting in the bioconda-test... Is there any extra tip to be aware of?

Juke34:
BUSCO fails within the singularity test :( 
`The environment variable AUGUSTUS_CONFIG_PATH is not set`
I have to find a work around

recipes/guessmylt/meta.yaml
Juke34: adding `  skip-lints: #  The skip mechanism does not work in conjunction with noarch but I need to skip osx due to inconsistency in BUSCO
    - should_not_be_noarch_skip
    - should_not_be_noarch` works now. Thx

recipes/guessmylt/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  version: "{{ version }}"
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: ce97be361e0fbfa0409fbce85d2d31b56cc05b52b4fea23492b184ac92c9e9ca
+
+build:
+  number: 0
+  noarch: python
+  skip: True # [osx]
```
epruesse: noarch and skip are mutually exclusive

recipes/guessmylt/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  license: GPLv3
+  license_family: GPL
+  license_file: LICENSE
+  summary: Software to guess the RNA-Seq library type.
+  dev_url: https://github.com/NBISweden/GUESSmyLT
+
+extra:
+  recipe-maintainers:
+    - juke34
+  skip-lints: #  The skip mechanism does not work in conjunction with noarch but I need to skip osx due to inconsistency in BUSCO
```
epruesse: As mentioned above, you cannot have both noarch and skip. It just doesn't work. This isn't the linter, this is conda. If your package will not work on OSX, you need to make it `noarch:false` (or remove line) and add `- should_be_noarch  # will not work on OSX due to BUSCO ...`.

epruesse:
> I don't get how works the **lint exclude**. The **skip** mechanism does not work in conjunction with **noarch**. So you mean I can keep both in the yaml file and use **lint exclude** in a way to avoid the `Linting Recipe` to fail?

**Noarch: false**: Packages using Python are built 6 times: py27, py36 and py37 each for OSX and Linux. You can "skip" some of these using the `build: skip` line. You cannot pin Python with `python >3` (won't work).

**Noarch: python**:  Instead of having special packages for each OS/python combination, only one package is built. It can only contain python code, as everything will be installed into `python-X.Y/site-libraries` with X.Y matching the Python that was chosen by the user. You cannot use `skip` as there are no builds that can be skipped.

**Noarch: generic**: Also builds only one package, but does not handle `python-X.Y/site-libraries`, so will not work for Python modules. It's e.g. for Java tools.

epruesse:
In your case ... it might be easiest to just allow installation on OSX, but issue a warning from your tool that the pipeline will not work as expected, or that there are known issues with BUSCO on OSX or something of the like.

recipes/guessmylt/meta.yaml
Juke34: Thx for the explanation. I understood how it works. Should be fine now.

recipes/guessmylt/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - pysam >=0.13.0
+    - snakemake ==5.4.*
+    - trinity ==2.8.4
+    - busco ==3.0.2
+    - bowtie2 >=2
+  run:
+    - python>3
+    - git >=2.11.1
+    - bcbio-gff ==0.6.4
+    - biopython ==1.67
```
epruesse: As @HadrienG said - not too happy with the specific pin.

Here's the problem: If you install both guessmylt and packageB and packageB requires bioython ==1.66 then conda will (after churning for an hour) tell you it can't install.

That's why we prefer ranges. If you require a package that has a history of breaking changes, you can use things like `>=1.67,<1.68.0a0` to keep it to whatever upstream appears to use for bug fixes only.

Juke34: I did some test and it sounds to work fine. The problem I had in the past was with python 2.7 anyway. So, I  have relaxed the required versions.

Juke34:
I have addressed the requested changes, would it be fine for accepting the PR?

epruesse:
@biocondabot lint

epruesse:
@biocondabot merge

epruesse:
Looks like @HadrienG has to sign off the requested changes.

epruesse:
I'm ok merging it like this. Still don't think anything should be pinned to exact versions without very very special reasons. It's ok so long as this is a "terminal" end tool that does not need to be included in pipelines or such.

@biocondabot merge

PR #15795, opened 2019-06-10
Ariba v2.14.0 release

kpepper:
I have read the [guidelines for bioconda recipes](https://bioconda.github.io/guidelines.html).
AFAIK, this recipe is directly relevant to the biological sciences
This PR updates an existing recipe.

### DIFF
--- a/recipes/ariba/meta.yaml
+++ b/recipes/ariba/meta.yaml
@@ -1,5 +1,5 @@
-{% set version = "2.14.0" %}
-{% set sha256 = "361d58dc2bfa1c8722aa695633771676bbc86cf22ecd879d25106930726651d7" %}
+{% set version = "2.14.1" %}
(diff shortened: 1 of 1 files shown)


kpepper:
Hi **@druvus**. Thanks for the review. As can be seen from my commit comments, I spotted the license file property was missing from the meta.yml, added it in, and then got a Bioconda build failure as the actual file is missing from the PyPI distribution (it is in GitHub). I would need to do another Ariba release to fix it which I just don't have time for at the moment. I have already implemented/tested the change required (see https://github.com/sanger-pathogens/ariba/pull/272) and it will be in the next Ariba release so the issue is resolved, but can you allow it through for this one please?

epruesse:
@biocondabot merge

epruesse:
@BiocondaBot merge

epruesse:
Hmm. That's a bug in the bot. Sigh.

Issue #12466, opened 2018-12-03
Install of pysamstats results in ValueError: pysam.libchtslib.HTSFile has the wrong size


lparsons:
`conda create -n pysamstats pysamstats`

```
Traceback (most recent call last):
  File "/Genomics/grid/users/lparsons/miniconda3/envs/pysamstats/bin/pysamstats", line 10, in <module>
    import pysamstats
  File "/Genomics/grid/users/lparsons/miniconda3/envs/pysamstats/lib/python3.6/site-packages/pysamstats/__init__.py", line 8, in <module>
    from .pileup import *
  File "/Genomics/grid/users/lparsons/miniconda3/envs/pysamstats/lib/python3.6/site-packages/pysamstats/pileup.py", line 6, in <module>
    import pysamstats.opt as opt
  File "libchtslib.pxd", line 2590, in init pysamstats.opt
ValueError: pysam.libchtslib.HTSFile has the wrong size, try recompiling. Expected 88, got 80
```

```
# packages in environment at /Genomics/grid/users/lparsons/miniconda3/envs/pysamstats:
#
# Name                    Version                   Build  Channel
bcftools                  1.9                  h4da6232_0    bioconda
blas                      1.1                    openblas    conda-forge
blosc                     1.14.4               hfc679d8_0    conda-forge
bzip2                     1.0.6                h470a237_2    conda-forge
ca-certificates           2018.11.29           ha4d7672_0    conda-forge
certifi                   2018.11.29            py36_1000    conda-forge
curl                      7.62.0               h74213dd_0    conda-forge
hdf5                      1.10.3               hc401514_2    conda-forge
htslib                    1.9                  hc238db4_4    bioconda
krb5                      1.16.2               hbb41f41_0    conda-forge
libcurl                   7.62.0               hbdb9355_0    conda-forge
libdeflate                1.0                  h470a237_0    bioconda
libedit                   3.1.20170329         haf1bffa_1    conda-forge
libffi                    3.2.1                hfc679d8_5    conda-forge
libgcc-ng                 7.2.0                hdf63c60_3    conda-forge
libgfortran               3.0.0                         1    conda-forge
libssh2                   1.8.0                h5b517e9_3    conda-forge
libstdcxx-ng              7.2.0                hdf63c60_3    conda-forge
lzo                       2.10                 h470a237_0    conda-forge
ncurses                   6.1                  hfc679d8_1    conda-forge
[...]

PR #15827, opened 2019-06-12
pysamstats: pin pysam version

lparsons:

### DIFF
--- a/recipes/pysamstats/meta.yaml
+++ b/recipes/pysamstats/meta.yaml
@@ -7,7 +7,7 @@ source:
   md5: 050294fe9fe99aae3b07ea28f6cbe449
 
 build:
-  number: 1
+  number: 2
 
 requirements:
   build:
@@ -19,7 +19,7 @@ requirements:
     - zlib
   run:
     - python
-    - pysam >=0.15
+    - {{ pin_compatible('pysam') }}
     - zlib
     - numpy
     - pytables


lparsons:
@alimanfoo While I think the recent update (and hence rebuild) of pysamstats resolved the issue for now, this should prevent errors like https://github.com/alimanfoo/pysamstats/issues/94 in the future.

lparsons:
Should also fix #12466.

dpryan79:
@bioconda-bot merge

dpryan79:
Thanks @lparsons!

alimanfoo:
Thanks @lparsons, much appreciated :+1:

alimanfoo:
Just learning about [how pin_compatible works](https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#pinning-at-the-recipe-level) and wondering if this should be ``{{ pin_compatible('pysam', max_pin='x.x') }}``. Pysam does not follow semver and regularly changes it's cython/C API between minor versions.

alimanfoo:
Advice appreciated, I'm still a novice about how (bio)conda builds work.

dpryan79:
Looking at the repodata you're correct. Can you make a PR?

alimanfoo:
> Looking at the repodata you're correct. Can you make a PR?

Happy to. I'm still a bit murky on what exactly `pin_compatible` is doing, I'd be grateful if you could confirm whether it should be:

``pin_compatible('pysam', max_pin='x.x')``

...or whether it should be an explicit version number, e.g.:

``pin_compatible('pysam', max_pin='0.16')``

dpryan79:
The `x.x` version is what's needed, then when new pysam releases come out that doesn't need to be changed. For pysam this might actually need to be `x.x.x` to match the htslib pinning.

epruesse:
What do you mean by matching the htslib pinning?

dpryan79:
Just that we pin to exact versions of htslib (I think at least).

epruesse:
Hmm. I don't know. We need to somehow parse `pin_compatible` for update_pinnings...

PR #15861, opened 2019-06-13
plastid - fix using incompatible numpy

lparsons:

### DIFF
--- a/recipes/plastid/meta.yaml
+++ b/recipes/plastid/meta.yaml
@@ -36,7 +36,7 @@ requirements:
   run:
     - python
     - {{ pin_compatible('numpy', max_pin='x.x') }} # Later numpy causes issue: https://github.com/joshuagryphon/plastid/issues/18, remove max_pin when fix released
-    - {{ pin_compatible('pysam', max_pin='x.x') }}
+    - {{ pin_compatible('pysam', max_pin='x.x.x') }}
     - termcolor
     - scipy >=0.15.1
     - pandas >=0.17.0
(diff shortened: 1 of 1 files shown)


dpryan79:
The repodata indicates that the pinning should work properly this time, so :+1: and I'll merge this once the CI testing finishes.

dpryan79:
Is this linking against pysam too? In that case it needs the same max_pin change.

lparsons:
It is linking against pysam, but I thought the pin compatible would handle that? The reason for the bumpy issue is an API change in bumpy that broke things. Happy to update if it does indeed need that. I still don't think I understand what pin_compatible is doing...

--
Lance Parsons - Scientific Programmer
Carl C. Icahn Laboratory - Room 136
Lewis-Sigler Institute for Integrative Genomics
Princeton University

________________________________

dpryan79:
There's no pinning done for pysam normally, so as is this ends up pinning it to `pysam >=0.15.2,<1.0a0`, which won't work. So yes, please add `max_pin='x.x'` or maybe `max_pin='x.x.x'` (the latter may be overkill, but since I don't use cython with pysam I'm not sure).

lparsons:
Thanks @epruesse. I think this is ready to go, perhaps one quick check on the repo.json when the bioconda bot produces it.

lparsons:
Hmmm... Not sure I understand what happened here, but the max_pin on pysam is now gone...

epruesse:
you force pushed ... overwriting the part where I had added it.

lparsons:
OK, put the `max_pin` back.

dpryan79:
I made the pysam pinning a bit stricter, hopefully that will future-proof things.

PR #15915, opened 2019-06-15
Ribolands

s-will:


druvus:
@BiocondaBot please update

eggzilla:
Hi @s-will  I will have a look :-)

eggzilla:
Is it possible that the shebang of drtransformer.py is breaking the build?

> 14:07:55 BIOCONDA INFO (OUT) /opt/conda/conda-bld/ribolands_1612965590876/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/bin/DrTransformer.py: line 7: from: command not found

eggzilla:
https://github.com/bad-ants-fleet/ribolands/blob/master/scripts/DrTransformer.py
I think you need a insert a
`#! /usr/bin/env python`

s-will:
HI @eggzilla, I agree, this seems the current (hopefully last remaining) problem. Many thanks for finding this and sorry for inviting to review prematurely (I was too optimistic ;)). I also think that patching the shebang is the best solution.

recipes/ribolands/meta.yaml
```diff
@@ -27,7 +26,6 @@ requirements:
     - matplotlib-base
     - future
     - crnsimulator
-    - treekin ==0.5.1
   run:
     - python <3.9
```
eggzilla, line 30: Do you really need the upper bound?

recipes/ribolands/meta.yaml
```diff
@@ -36,7 +34,6 @@ requirements:
     - matplotlib-base
     - future
     - crnsimulator
-    - treekin ==0.5.1
 
 test:
```
eggzilla, line 38: If possible a more extensive test then --version would be awesome :-)

eggzilla: @s-will : Please comment concerning the upper version bound on python and optionally adding a test.

s-will:
Hi @eggzilla it passed the test, but I cannot produce a Mac treekin 0.5.1, because mlapack is not available.

eggzilla:
Do you want to temporarily add a skip for MacOs?

s-will:
I skip MacOs now. Maybe I would return later to check again, why mlapack failed.

s-will:
For the ribolands pacakge there is an interesting problem, which I was not sure how to best resolve - the package must be architecture independent, since it is pure python (otherwise lint complains). But this means that I can neither skip osx (lint complains) or define treekin 0.5.1 as dependency (test on osx is going to fail).

s-will:
I resolved it by dropping the treekin dependency in ribolands, which is not ideal, but at least makes everything usable, so I think we can justify it

s-will:
@eggzilla concerning your two remarks to the code: 1) the above also means that the pin to 0.5.1 is dropped. Still doctor transformer, one of the scripts in ribolands requires it, because previous versions didn't have mlapack support. 2) What do you have in mind? (Actually, I thought it's common practice to only run the tools with help or version in conda tests... I can of course call the other scripts in the package as well, or directly test import of the python modules)

eggzilla:
If you like it would be awesome if you could add calls with --version for the other scripts. 
Concerning the noarch & skip situation, I think you can use a lint-skip for no-arch python:
https://bioconda.github.io/contributor/linting.html#skipping-a-lint-check
https://bioconda.github.io/contributor/linting.html#id14
I guess like this:
extra:
  skip-lints:
    - should_be_noarch_python

Then you should be able to use skip with macos again :-)

s-will:
cool. thanks, I can try the skip-lint (later) and add tests

eggzilla (approved):
Looks good :-)

s-will:
:+1: (I forgot the to explain the strange viennarna PIN; it seems that this is temporarily required, since the viennarna does not provide Python bindings for 3.9 - I tried to fix this by rebuilding viennarna, but this didn't succeed and I cannot see why.)

eggzilla:
Oh, I hope I did not merge prematurely ?!
I guess you are referring to this issue: https://github.com/ViennaRNA/ViennaRNA/issues/110 ?

s-will:
yes, this issue is discussing the same problem with the current viennarna conda package. I don't think the merge is premature - IMO it's more important to finally have the treekin and ribolands conda package. The pin can still be removed in a new ribolands package build when the viennarna issue is resolved.

eggzilla:
ok :-) Ping me if I can help in future updates

PR #15975, opened 2019-06-17
Bump pisces

TMiguelT:


TMiguelT:
Notably the linter is always going to fail here because we can't add tests since `dotnet` isn't available as a conda package.

TMiguelT:
...unless I skip the tests entirely

npavlovikj:
Btw, the version of `VariantQualityRecalibration` in `build.sh` is 5.2.7.47. Shouldn't this be updated to `5.2.10.49` as well?

npavlovikj:
Also, as you can see from the linting information, for GPL licensed software, the license file needs to be packaged as well. If the license is not in the tarball, then please copy it to the recipe directory and add it accordingly.

recipes/pisces/meta.yaml
```diff
@@ -29,3 +30,9 @@ about:
   home: "https://github.com/Illumina/Pisces"
   license: GPLv3
   summary: "Somatic and germline variant caller for amplicon data. Recommended caller for tumor-only workflows."
+
+
+extra:
+  skip-lints:
+    # dotnet isn't in Conda, so we can't test pisces
+    - missing_tests
```
dpryan79: Can you try making mono a dependency and then using it to run tests?

TMiguelT: Pisces 5.2 targets the dotnet framework only, not mono: https://github.com/Illumina/Pisces#system-requirements

npavlovikj:
Since the License file is in the recipe directory now, please use:
`license_file: '{{ environ["RECIPE_DIR"] }}/LICENSE.txt'`
instead.

recipes/pisces/meta.yaml
dpryan79: Mono is an open source implementation of the dot net framework.

bgruening:
@npavlovikj  I think if its in the recipe dir we do not need to prefix it. LICENSE.txt should just work.

TMiguelT:
It is failing to find the file though. Any ideas why?

npavlovikj:
@TMiguelT , the error message you are getting is:
` 05:50:06 BIOCONDA INFO (OUT) ValueError: License file given in about/license_file (/opt/recipe/README.txt) does not exist in source root dir or in recipe root dir (with meta.yaml)`
The license file is named "LICENSE.txt" and not "README.txt", so please change that.

recipes/pisces/meta.yaml
TMiguelT: Yes but pisces targets dotnet core, not the dotnet framework, so mono won't help.

npavlovikj:
> @npavlovikj I think if its in the recipe dir we do not need to prefix it. LICENSE.txt should just work.

Thanks @bgruening - I use "environ" due to habit, but you are correct that "LICENSE.txt" is sufficient too.

dpryan79:
Thanks @TMiguelT!

PR #15979, opened 2019-06-17
Raxml ng

johnlees:

### DIFF
--- a/recipes/raxml-ng/build.sh
+++ b/recipes/raxml-ng/build.sh
@@ -7,7 +7,7 @@ pushd build_pthreads
    make
    install -d ${PREFIX}/bin
    install ../bin/raxml-ng ${PREFIX}/bin
-pushd
+popd
 
 # mpi
 mkdir build_mpi
@@ -16,4 +16,4 @@ pushd build_mpi
    make
    install -d ${PREFIX}/bin
    install -T ../bin/raxml-ng ${PREFIX}/bin/raxml-ng-MPI
-pushd
+popd
--- a/recipes/raxml-ng/build.sh
+++ b/recipes/raxml-ng/build.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+# pthreads
+mkdir build_pthreads
+pushd build_pthreads
+   cmake ..
+   make
+   install -d ${PREFIX}/bin
+   install ../bin/raxml-ng ${PREFIX}/bin
+pushd
+
+# mpi
+mkdir build_mpi
+pushd build_mpi
+   cmake -DUSE_MPI=ON ..
+   make
+   install -d ${PREFIX}/bin
+   install -T ../bin/raxml-ng ${PREFIX}/bin/raxml-ng-MPI
+pushd
--- a/recipes/raxml-ng/build.sh
+++ b/recipes/raxml-ng/build.sh
@@ -15,5 +15,5 @@ pushd build_mpi
    cmake -DUSE_MPI=ON ..
    make
    install -d ${PREFIX}/bin
-   install -T ../bin/raxml-ng ${PREFIX}/bin/raxml-ng-MPI
+   install ../bin/raxml-ng ${PREFIX}/bin/raxml-ng-MPI
 popd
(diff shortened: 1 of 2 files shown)


recipes/raxml-ng/meta.yaml
```diff
@@ -0,0 +1,44 @@
+{% set name = "raxml-ng" %}
+{% set version = "0.9.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  git_url: https://github.com/amkozlov/{{ name|lower }}.git
+  git_rev: {{ version }}
```
dpryan79: Do you need to use git rather than https://github.com/amkozlov/raxml-ng/archive/0.9.0.tar.gz ?

recipes/raxml-ng/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+  host:
+    - gmp
+    - bison
+    - flex
+    - cmake
```
dpryan79: cmake, bison, and I think flex can be moved to the `build:` section.

recipes/raxml-ng/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - pthread-stubs
```
dpryan79: Is this actually needed? pthreads is there regardless.

recipes/raxml-ng/build.sh
```diff
@@ -0,0 +1,19 @@
...
+   cmake ..
+   make
+   install -d ${PREFIX}/bin
+   install ../bin/raxml-ng ${PREFIX}/bin
+popd
+
+# mpi
+mkdir build_mpi
+pushd build_mpi
+   cmake -DUSE_MPI=ON ..
```
dpryan79: Is there an openmpi that's getting pulled in?

recipes/raxml-ng/meta.yaml
epruesse: yes, flex also

recipes/raxml-ng/meta.yaml
epruesse: don't think that's necessary any more

recipes/raxml-ng/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  run:
+    - pthread-stubs
```
epruesse: same here

recipes/raxml-ng/meta.yaml
johnlees: There are two submodules hidden in the libs/ subdirectory that this deals with nicely

recipes/raxml-ng/meta.yaml
johnlees: Thanks, have removed

recipes/raxml-ng/build.sh
johnlees: Have added openmpi to host section now, but maybe I should be putting this in run (will check tomorrow)

recipes/raxml-ng/meta.yaml
johnlees: Removed

recipes/raxml-ng/meta.yaml
johnlees: Thanks, I have moved these (but I don't understand the rules for which go in `build:` and which go in `host:`?)

dpryan79: The general idea is that anything linked against (zlib, libcurl, etc.) should go in the `host:` section, whereas things used purely for building the package should go in the `build:` section (compilers, preprocessing packages, make/cmake, etc.).

recipes/raxml-ng/build.sh
dpryan79: Yeah, it should be in `run:` then too.

dpryan79: I expect you'll need `gmp` in the run section as well.

johnlees: I checked the executables with `ldd`, looks like `openmpi` is needed at run but not `gmp`.

johnlees:
@dpryan79 @epruesse Thanks for the review comments! I will merge when the tests have passed

PR #16011, opened 2019-06-19
New recipe Magicblast

boratyng:

### DIFF
--- a/recipes/magicblast/build.sh
+++ b/recipes/magicblast/build.sh
@@ -3,5 +3,8 @@ set -euxo pipefail
 
 mkdir -p $PREFIX/bin
 
+# the binary expects libbz1.so.1, but the available one is libbz2.so.1.0
+patchelf --replace-needed libbz2.so.1 libbz2.so.1.0 bin/magicblast
+
 mv bin/magicblast $PREFIX/bin
 
--- a/recipes/magicblast/build.sh
+++ b/recipes/magicblast/build.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+set -euxo pipefail
+
+mkdir -p $PREFIX/bin
+
+mv bin/magicblast $PREFIX/bin
+
--- a/recipes/magicblast/build.sh
+++ b/recipes/magicblast/build.sh
@@ -3,7 +3,7 @@ set -euxo pipefail
 
 mkdir -p $PREFIX/bin
 
-# the binary expects libbz1.so.1, but the correct name is libbz2.so.1.0
+# the binary expects libbz2.so.1, but the correct name is libbz2.so.1.0
 if [[ $(uname) = Linux ]] ; then
     patchelf --replace-needed libbz2.so.1 libbz2.so.1.0 bin/magicblast
 fi
--- a/recipes/magicblast/build.sh
+++ b/recipes/magicblast/build.sh
@@ -3,8 +3,10 @@ set -euxo pipefail
 
 mkdir -p $PREFIX/bin
 
-# the binary expects libbz1.so.1, but the available one is libbz2.so.1.0
-patchelf --replace-needed libbz2.so.1 libbz2.so.1.0 bin/magicblast
+# the binary expects libbz1.so.1, but the correct name is libbz2.so.1.0
+if [[ $(uname) = Linux ]] ; then
+    patchelf --replace-needed libbz2.so.1 libbz2.so.1.0 bin/magicblast
+fi
 
 mv bin/magicblast $PREFIX/bin
 
--- a/recipes/magicblast/meta.yaml
+++ b/recipes/magicblast/meta.yaml
@@ -5,32 +5,33 @@ package:
   version: {{ version }}
 
 source:
-  url: ftp://ftp.ncbi.nlm.nih.gov/blast/executables/magicblast/LATEST/ncbi-magicblast-{{ version }}-x64-linux.tar.gz # [linux]
-  url: ftp://ftp.ncbi.nlm.nih.gov/blast/executables/magicblast/LATEST/ncbi-magicblast-{{ version }}-x64-macosx.tar.gz # [osx]
-  sha256: 9995524a2ce67850516debb3341f268d485113f1c28005e39b5fbf47fe27d250 # [linux]
-  sha256: f48daeb86065be4f256820b5c189f0141eafdf50e5a0b8e837b381e82a447b0e # [osx]
+  - url: ftp://ftp.ncbi.nlm.nih.gov/blast/executables/magicblast/LATEST/ncbi-magicblast-{{ version }}-x64-linux.tar.gz # [linux]
+    sha256: 9995524a2ce67850516debb3341f268d485113f1c28005e39b5fbf47fe27d250 # [linux]
+  - url: ftp://ftp.ncbi.nlm.nih.gov/blast/executables/magicblast/LATEST/ncbi-magicblast-{{ version }}-x64-macosx.tar.gz # [osx]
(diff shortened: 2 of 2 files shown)


epruesse:
@biocondabot rebuild

recipes/magicblast/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - zlib
+    - bzip2
+    - pcre
+    - libxml2
+  run:
+    - zlib
+    - bzip2
+    - pcre
+    - libxml2
+    - blast # needed for makeblastdb
```
epruesse: Do we know which version is required?

recipes/magicblast/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+package:
+  name: magicblast
+  version: {{ version }}
+
+source:
+  url: ftp://ftp.ncbi.nlm.nih.gov/blast/executables/magicblast/LATEST/ncbi-magicblast-{{ version }}-x64-linux.tar.gz # [linux]
+  url: ftp://ftp.ncbi.nlm.nih.gov/blast/executables/magicblast/LATEST/ncbi-magicblast-{{ version }}-x64-macosx.tar.gz # [macos]
+  sha256: 9995524a2ce67850516debb3341f268d485113f1c28005e39b5fbf47fe27d250 # [linux]
+  sha256: f48daeb86065be4f256820b5c189f0141eafdf50e5a0b8e837b381e82a447b0e # [macos]
```
epruesse: Any chance to build this from source?

recipes/magicblast/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - patchelf
```
epruesse: Hmm. These are not needed if you just copy the binaries. If you can at all, please do try to build from source though!

recipes/magicblast/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -euxo pipefail
+
+mkdir -p $PREFIX/bin
+
+# the binary expects libbz1.so.1, but the available one is libbz2.so.1.0
+patchelf --replace-needed libbz2.so.1 libbz2.so.1.0 bin/magicblast
```
epruesse: This is hairy. If you must stick to the binaries and must absolutely patch the ELF, please add a test case that reads/writes bz2 files to make sure this actually works.

This also will not work on OSX. Different tools there.

recipes/magicblast/meta.yaml
boratyng: For blast the latest available, but which specific version is not critical. It is only needed for makeblastdb binary. For bzip2, pcre, and libxml2 I can add vesrions. How do I specify  version?

recipes/magicblast/meta.yaml
boratyng: I am planning to update this recipe later to build from sources. There are a few dependencies that are not easy to deal with and I wanted to have this available sooner.

recipes/magicblast/meta.yaml
boratyng: I thought I removed them. Sorry. Removed now.

recipes/magicblast/meta.yaml
epruesse: promise? ;-)

recipes/magicblast/meta.yaml
epruesse: Check the docs - can get complicated. Briefly, ` >=1.2.3,<1.3.0a0` is a good way to do it. The `,` is an `AND`, so this says "at least version 1.2.3" and "less than 1.3.0a0". The latter is the smallest possible 1.3 version. The `0a0` is because `1.3.a1 < 1.3`. Should not be strictly necessary because we don't allow alphas, but who knows what other channels a user might add. Easier than writing `1.2.999999999999999....`. 

For your later from source build: you can use `pin_compatible('libxml2', max_pin="x.x")` in the `run` section to generate dependencies based on what was installed in the `host` section. The number of `x`s determines where it pins. Given `libxml 1.2.3`, a single `x` will pin to `>=1.2.3,<2`, `x.x` to `>=1.2.3,<1.3`, while `x.x.x` will pin to `>=1.2.3,<1.2.4` (plus the `0a0` of course, it does that).

recipes/magicblast/meta.yaml
marcelm: As you may know, it is *a lot* of work getting a source build to work for tools that use the BLAST build system. It took me many hours over the course of a week for IgBLAST. If under these circumstances a binary works with some minor ELF patching, then that should be good enough.

The recipe should be accepted without the condition that it be "fixed" later, especially when it comes from a first-time contributor.

recipes/magicblast/build.sh
marcelm: The comment is misleading: It’s not `libbz1.so.1`, but `libbz2.so.1` in both cases, except that the available has an additional `.0` in the end. The binary is linked against a more generic soname than it should be. And even then, libbz2 hasn’t changed in years, so there is really no need to add a test case. Also, with the patch, the binary runs. Without it, it doesn’t. So running the binary in the tests is enough. (I’m using the same patch in the IgBLAST recipe.)

marcelm:
I am working on a source-based IgBLAST recipe just now. It uses the same build system, so when you want to switch to building from source at a later time, just look at the BLAST recipe as a template (or the IgBLAST one when it’s ready). (Possibly you’ve even looked at the current IgBLAST recipe already?)

recipes/magicblast/build.sh
marcelm: If I remember correctly, you can just skip the `patchelf` part on macOS, so wrap it in a `if [[ $(uname) == Linux ]]; then ... fi` and add a `# [linux]` selector to the patchelf dependency in the recipe.

recipes/magicblast/meta.yaml
boratyng: I added library versions.

recipes/magicblast/meta.yaml
boratyng: @epruesse, I promise cross my heart ... :-)

recipes/magicblast/build.sh
boratyng: Thanks @marcelm I added changes that you suggested.

@epruesse, magicblast does not read/write any bz2 files directly. I added a test case that downloads data from NCBI. bz2 may be used there.

boratyng:
@marcelm, thanks for the suggestion. Yes, I knew about the blast and igblast recipes. I copied most of things in my current magicblast recipe from your igblast one. Magicblast build will be based on blast build, but there are a few additional libraries that come as dependencies and need to be worked into the build.

recipes/magicblast/meta.yaml
epruesse: @marcelm I am very well aware - it cost me a week to migrate the BLAST recipe to building from source. There are, however, many good reasons as to why we should build from source rather than just repackage binaries. That said, I also understand when people need a package ASAP and can't fight the compiler right now.

@boratyng Thank you! That would be greatly appreciated! :)

recipes/magicblast/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -euxo pipefail
+
+mkdir -p $PREFIX/bin
+
+# the binary expects libbz1.so.1, but the correct name is libbz2.so.1.0
```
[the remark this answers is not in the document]
boratyng: The typo is fixed. Thanks for catching that.

boratyng:
@BiocondaBot rebuild

PR #16078, opened 2019-06-25
genform: initial commit

bernt-matthias:
adds genform recipe

TODOs: 
- [x] ask developers if they prefer to make a proper release instead of having a svn revision
- [x] ask developers if they want to add a license file
- [x] otherwise change md5sum to md5 and skip the corresponding linting checks https://bioconda.github.io/linting.html

### DIFF
--- a/recipes/genform/build.sh
+++ b/recipes/genform/build.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-${CXX} ./src/*.cpp -o genform
+${CXX} ${CPPFLAGS} ${CXXFLAGS} ${LDFLAGS} ./src/*.cpp -o genform
 mkdir -p "$PREFIX"/bin/
 mv genform "$PREFIX"/bin/
--- a/recipes/genform/build.sh
+++ b/recipes/genform/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+${CXX} ./src/*.cpp -o genform
+mkdir -p "$PREFIX"/bin/
+mv genform "$PREFIX"/bin/
--- a/recipes/genform/meta.yaml
+++ b/recipes/genform/meta.yaml
@@ -1,0 +1,40 @@
+{% set version = "r8" %}
+{% set name = "genform" %}
+{% set sha256 = "d9199aa9c13c6aa68025b15e4d30a4e445a265ce828b3314103307dc7b6c13b5" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+build:
+  number: 0
+
+source:
+  # download is a stable link for https://sourceforge.net/code-snapshots/svn/g/ge/genform/code/genform-code-r8-trunk.zip
+  url: https://depot.galaxyproject.org/software/genform/genform_r8_src_all.zip
+  sha256: {{ sha256 }}
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+
+# genform without parameter or w help has exit code 1, so we need test data
+# since test data can't be taken from the source dir its created manually
+test:
+  commands:
+    - 'echo -e "132.07686\t147049184\n133.08012\t4004960.8" > CreatineMs.txt && genform ms=CreatineMs.txt'
+about:
+  home: https://sourceforge.net/projects/genform/
+  # note sources say GPL (wo version) but LGPL licence file is shipped since 
+  # - LGPL allows to relicense under the GPL and
+  # - the package contains just a single executable and not a library
+  # we say GPL-2.0-only
+  license: 'GPL-2.0-only'
+  license_file: src/lgpl-2.0.txt 
+  summary: Generation of molecular formulas by high-resolution MS and MS/MS data
+
+extra:
+  maintainers:
+    - bernt-matthias
+  identifiers:
+    - doi:10.1101/295071


recipes/genform/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+$GXX ./src/*.cpp -o genform
```
mbargull: The build log is quite scrambled, so it's hard to catch, but the error is at
```
/Users/distiller/project/miniconda/conda-bld/genform_1563451308638/work/conda_build.sh: line 6: ./src/main.cpp: Permission denied
```
Which is probably because `GXX` is unset, thus this tries to execute `./src/*.ccp` with `./src/main.cpp` as the first entry for that glob.

Try `${CXX}` instead.

bernt-matthias:
thanks for the tips. this worked.

recipes/genform/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - 'echo -e "132.07686\t147049184\n133.08012\t4004960.8" > CreatineMs.txt && genform ms=CreatineMs.txt'
+about:
+  home: https://sourceforge.net/projects/genform/
+  license: 'LGPL (>= 2.0)'
+  license_file: src/lgpl-2.0.txt 
+  summary: Generation of molecular formulas by high-resolution MS and MS/MS data
+
+extra:
+  maintainers:
+    - Matthias Bernt
```
mbargull: Can you change that to you GitHub handle?

recipes/genform/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - {{ compiler('cxx') }}
+
+# genform without parameter or w help has exit code 1, so we need test data
+# since test data can't be taken from the source dir its created manually
+test:
+  commands:
+    - 'echo -e "132.07686\t147049184\n133.08012\t4004960.8" > CreatineMs.txt && genform ms=CreatineMs.txt'
+about:
+  home: https://sourceforge.net/projects/genform/
+  license: 'LGPL (>= 2.0)'
```
mbargull: I checked if the `>= 2.0` is accurate. Acc. to https://sourceforge.net/projects/genform/ and the `licence_file`, this should rather be `LGPL-2.0-only`.

Acc. to the source files: Well... 
```
> grep -rhim1 license genform-code-r8-trunk/ | sort -u                                                                                                                                                                                                                           
                  GNU LIBRARY GENERAL PUBLIC LICENSE
 *  it under the terms of the GNU General Public License as published by
 *  it under the terms of the GNU General Public License as published by
```
... Y U NO GIVE CORRECT LICENSE? \*sigh\*

So... even though the SourceForge project page and `src/lgpl-2.0.txt` say "LGPL", it is actually "GPL" with an unknown version. Großartig!

recipes/genform/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+source:
+  # download is a stable link for https://sourceforge.net/code-snapshots/svn/g/ge/genform/code/genform-code-r8-trunk.zip
+  url: https://depot.galaxyproject.org/software/genform/genform_r8_src_all.zip
+  sha256: {{ sha256 }}
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+
+# genform without parameter or w help has exit code 1, so we need test data
```
mbargull: If you want to test whether the help text is printed correctly, you could do something like
```yaml
    - '{ genform || true ; } | grep -qF some-known-substring-of-the-help-text'
```
or
```yaml
    - '{ genform 2>&1 || true ; } | grep -qF some-known-substring-of-the-help-text'
```
in case the text is printed to the error stream.

recipes/genform/meta.yaml
bernt-matthias: I think its OK. The file actually says: 

```
                  GNU LIBRARY GENERAL PUBLIC LICENSE
                       Version 2, June 1991
```

which is LGPL2 according to https://en.wikipedia.org/wiki/GNU_Lesser_General_Public_License

recipes/genform/meta.yaml
bernt-matthias: done: -> bernt-matthias

recipes/genform/run_test.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
bernt-matthias: removed the file

recipes/genform/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+{% set name = "genform" %}
+{% set sha256 = "d9199aa9c13c6aa68025b15e4d30a4e445a265ce828b3314103307dc7b6c13b5" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+build:
+  number: 0
+#  skip: True  # [osx]
```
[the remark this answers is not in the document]
bernt-matthias: done

recipes/genform/meta.yaml
bernt-matthias: good idea. unfortunately the version number is not in the output, so I think that the current test is sufficient

recipes/genform/meta.yaml
mbargull: Yes, that's what the license file says. However, this is just the generic license text. It is more important how the thing is actually licensed, i.e., the file headers which all say something like
```
>  head -n12 genform-code-r8-trunk/src/main.cpp
/*
 *  main.cpp, part of GenForm by M. Meringer Copyright (C) 2015
 *
 *  This program is free software; you can redistribute it and/or modify
 *  it under the terms of the GNU General Public License as published by
 *  the Free Software Foundation.
 *
 *  This program is distributed in the hope that it will be useful,
 *  but WITHOUT ANY WARRANTY; without even the implied warranty of
 *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 *  GNU General Public License for more details.
*/
```

recipes/genform/meta.yaml
mbargull: Yes, you don't have to do the `grep` thing since you have the other test. I just wanted to address your `meta.yaml` comment ;).

recipes/genform/meta.yaml
bernt-matthias: Damn - you are right. I could contact the maintainer to change the text and/or the license file. But then the new release would need to be in cargo port before we can have it here. Which needs time -- but "gut ding will weile haben"

Do you see another option?

mbargull: > I could contact the maintainer to change the text and/or the license file.

That would be most helpful, yes, please do! Thanks!

> Do you see another option?

I'm not an expert on the \*GPLs as I never released something under such a license and thus avoided dissecting those (comparably) long license texts. But according to our "scientifically sound source" \*cough\* https://en.wikipedia.org/w/index.php?oldid=901931174#Compatibility :
> One feature of the LGPL is the permission to relicense under the GPL any piece of software which is received under the LGPL

Which would mean we are free to distribute it under the GPL even if it would be distributed upstream as LGPL.

And since this package is just a single executable and not a library anyway, the LGPL / GPL distinction doesn't matter much.

Ergo: I'm fine to tag it as `GPL`. And since `LGPL 2.0` has been "mentioned" upstream, it would probably be okay to say `GPL-2.0-only`, I guess.

bernt-matthias:
depot.galaxyproject.org seems down. narf

recipes/genform/build.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 
-${CXX} ./src/*.cpp -o genform
+${CXX} ${CPPFLAGS} ${CXXFLAGS} ${LDFLAGS} ./src/*.cpp -o genform
```
mbargull: I forgot about/missed the missing flags before. We generally want to use them to make sure the compiled binaries are most compatible, e.g., run on older hardware.
Hope you don't mind me adding them.

bgruening:
Merci!

PR #16093, opened 2019-06-26
add Platon recipe

oschwengers:


epruesse:
The package downloads a file "https://s3.computational.bio.uni-giessen.de/swift/v1/platon/db.tar.gz" which doesn't seem to be available any more.

oschwengers:
@epruesse : The URL is correct but we're just in the middle of a scheduled maintenance outage right now. This is a long planned update of our complete compute and cloud/S3 infrastructure which might last a couple of days. As soon as everything is back online, I'll trigger circleci again.

epruesse:
@oschwengers You must not  download ***anything*** in the `build.sh`. Consider that 15 years from now, this package may have to be rebuilt. We keep copies of the sources listed in `source/url`, but we have *no* way of parsing build scripts for this kind of thing.

***IFF** the database is versioned synchronously with the source code and small, you could add it as another URL. It would be preferable for it to be released as part of the source archive, however.

If the database is large, it should be hosted somewhere else and downloaded by the user.

oschwengers:
@BiocondaBot rebuild

oschwengers:
```
$ hash & extract packages for noarch: 0% 0/1 [00:00<?, ?it/s]Too long with no output (exceeded 10m0s)
```
Might this be related to the package size? Is there a certain size limit for a conda package?
For the sake of user convenience I included a necessary DB (2.8 Gb) into the package. Is it ok or do I need to exclude the database?

oschwengers:
@BiocondaBot rebuild

oschwengers:
@epruesse : Thanks for the hint. I'll exclude the db download and ask users to manually download and setup the db.

recipes/platon/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+set -e
+
+# install PLATON via pip
+$PYTHON -m pip install --no-deps --ignore-installed --no-cache-dir .
```
bgruening: This could be moved directly into the meta.yml as build/script if you like.

recipes/platon/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  noarch: python
+  number: 0
+  entry_points:
+    - platon = platon.platon:main
+
+requirements:
+  host:
+    - python >=3
+
+  build:
```
bgruening: build should be moved up and host should be the second section.
Also pip should be in host not in build.

tar is not needed anymore?

recipes/platon/build.sh
oschwengers: Ok. Good point, now that the necessary database download is passed to the user's responsibility.

recipes/platon/meta.yaml
oschwengers: Ok, I wasn't aware of this. 

Tar is still needed as it's the source of the local pip installation. This way, it merely depends on the released GitHub version but not on a published PyPi.org-release.

oschwengers:
Somehow the `macos` test is blocked. `Lint` + `linux` have successfully passed.
Any ideas?

oschwengers:
@BiocondaBot rebuild

bgruening:
If you are using wget in your post-link step you also need to add it to your run environment. You can test this locally with `conda build ./recipes/platon/`

PR #16204, opened 2019-07-04
Fixed merge MetaPhlAn tables

fbeghini:


recipes/metaphlan2/2.9.0/meta.yaml
```diff
@@ -1,6 +1,6 @@
 {% set name = "metaphlan2" %}
 {% set version = "2.9.13" %}
-{% set md5 = "cf0c953d53713d07cd5cbd68302497eb" %}
+{% set md5 = "1a2653e243c9fceac36b6ce943d597ad" %}
```
dpryan79: Do you know why the md5sum of the source changed but the version didn't? That's VERY odd.

fbeghini: Just forgot to change the version number, resolved with a new commit

PR #16242, opened 2019-07-08
Adding surpyvor

wdecoster:
This recipe adds my Python tool surpyvor, which is a wrapper around SURVIVOR with additional convenience functions and plotting subcommands.

_Bioconda requires reviews prior to merging pull-requests (PRs). To facilitate this, once your PR is passing tests and ready to be merged, please add the `please review & merge` label so other members of the bioconda community can have a look at your PR and either make suggestions or merge it. Note that if you are not already a member of the bioconda project (meaning that you can't add this label), please ping `@bioconda/core` so that your PR can be reviewed and merged (please note if you'd like to be added to the bioconda project). Please see https://github.com/bioconda/bioconda-recipes/issues/15332 for more details._

### DIFF
--- a/recipes/surpyvor/meta.yaml
+++ b/recipes/surpyvor/meta.yaml
@@ -13,7 +13,7 @@ build:
     - surpyvor=surpyvor.surpyvor:main
   noarch: generic
   number: 0
-
+  script: "{{ PYTHON }} -m pip install . -vv"
 
 requirements:
   host:
--- a/recipes/surpyvor/meta.yaml
+++ b/recipes/surpyvor/meta.yaml
@@ -1,0 +1,53 @@
+{% set version = "0.5.0" %}
+
+package:
+  name: surpyvor
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/s/surpyvor/surpyvor-{{ version }}.tar.gz
+  sha256: 58b07eb9848e0b5d38a56f33062bc0e8b06366c58529923cef8aacfc3bd98fd6
+
+build:
+  entry_points:
+    - surpyvor=surpyvor.surpyvor:main
+  number: 0
+
+
+requirements:
+  host:
+    - python >=3
+    - setuptools
+    - bcftools
+    - tabix
+    - survivor
+    - numpy
+    - cyvcf2
+    - matplotlib
+    - matplotlib-venn
+    - upsetplot
+
+  run:
+    - python >=3
+    - bcftools
+    - bgzip
+    - tabix
+    - survivor
+    - numpy
+    - cyvcf2
+    - matplotlib
+    - matplotlib-venn
+    - upsetplot
+
+test:
+  imports:
+    - surpyvor
+
+  commands:
+    - surpyvor --help
+
+about:
+  home: https://github.com/wdecoster/surpyvor
+  license: MIT License
+  summary: 'Evaluating, merging and plotting SV vcf files'
+  license_family: MIT
--- a/recipes/surpyvor/meta.yaml
+++ b/recipes/surpyvor/meta.yaml
@@ -11,7 +11,7 @@ source:
 build:
   entry_points:
     - surpyvor=surpyvor.surpyvor:main
-  noarch: generic
+  noarch: python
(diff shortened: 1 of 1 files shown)


recipes/surpyvor/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/s/surpyvor/surpyvor-{{ version }}.tar.gz
+  sha256: 58b07eb9848e0b5d38a56f33062bc0e8b06366c58529923cef8aacfc3bd98fd6
+
+build:
+  entry_points:
+    - surpyvor=surpyvor.surpyvor:main
+  noarch: generic
```
npavlovikj: This should be `noarch: python` instead for pure python packages. `noarch: generic` is for java/perl or packages where you don't compile anything, but only copy the binaries.

recipes/surpyvor/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  number: 0
+  script: "{{ PYTHON }} -m pip install . --no-deps -vv"
+
+requirements:
+  host:
+    - python >=3
+    - pip
+
```
npavlovikj: Please remove space.

recipes/surpyvor/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - numpy
+    - cyvcf2
+    - matplotlib
+    - matplotlib-venn
+    - upsetplot
+
+test:
+  imports:
+    - surpyvor
+
```
npavlovikj: Please remove space.

recipes/surpyvor/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  commands:
+    - surpyvor --help
+
+about:
+  home: https://github.com/wdecoster/surpyvor
+  license: MIT License
+  summary: 'Evaluating, merging and plotting SV vcf files'
+  license_family: MIT
```
npavlovikj: If the license file is in the tarball, you can consider it adding it with `license_file`.

Also, you can add yourself as a maintainer if you want and add the `doi` of your publicatiton, e.g.:
```
extra:
  recipe-maintainers:
    - wdecoster
  identifiers:
    - doi:10.1101/gr.244939.118
```

recipes/surpyvor/meta.yaml
wdecoster: Alright, does this also apply if there are non-python dependencies?

recipes/surpyvor/meta.yaml
wdecoster: Thanks for the suggestion, will do! The license file is not included in the tarball though. I can add it next to the meta.yaml if that would be an added value?

wdecoster:
Thanks for the changes @druvus! Can you give me some intuition or point me to guidelines to know which dependencies go under `host`, and which only under `run`?

recipes/surpyvor/meta.yaml
npavlovikj: Yes, as long as the main package doesn't use architecture dependent packages to build, then it should be `noarch`, https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#architecture-independent-packages.

recipes/surpyvor/meta.yaml
npavlovikj: Yes, that will work too. Just copy the license file from your repo to the recipe directory and use `license_file: LICENSE`.

npavlovikj:
@wdecoster , you can read more about the `requirements` section here, https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#requirements-section. In general, the packages defined in `build` and `host` are used to build/compile/install the package, while the ones defined in `run` are needed when the tool is run, not necessary for building it. In your case, to install the package you only need `python` and `pip`, but you need other dependencies in order to run it.

wdecoster:
@npavlovikj Thanks for clarifying! I've pushed the meta.yaml with the suggested changes.

PR #16256, opened 2019-07-08
Add SeqAn 3.0.0

johnlees:

Add SeqAn 3.0.0

### DIFF
--- a/recipes/seqan3/build.sh
+++ b/recipes/seqan3/build.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+install -d $PREFIX/include
+install -d $PREFIX/include/seqan3/submodules/cereal/include/
+install -d $PREFIX/include/seqan3/submodules/range-v3/include/
+install -d $PREFIX/include/seqan3/submodules/sdsl/include/
+install -d $PREFIX/share/cmake
+cp -r $SRC_DIR/include/seqan3 $PREFIX/include
+cp -r $SRC_DIR/build_system/seqan3-config.cmake $PREFIX/share/cmake
+cp -r $SRC_DIR/submodules/cereal/include/cereal $PREFIX/include/seqan3/submodules/cereal/include/
+cp -r $SRC_DIR/submodules/range-v3/include/range $PREFIX/include/seqan3/submodules/range-v3/include/
+cp -r $SRC_DIR/submodules/range-v3/include/meta $PREFIX/include/seqan3/submodules/range-v3/include/
+cp -r $SRC_DIR/submodules/range-v3/include/concepts $PREFIX/include/seqan3/submodules/range-v3/include/
+cp -r $SRC_DIR/submodules/range-v3/include/std $PREFIX/include/seqan3/submodules/range-v3/include/
+cp -r $SRC_DIR/submodules/sdsl-lite/include/sdsl $PREFIX/include/seqan3/submodules/sdsl/include/
--- a/recipes/seqan3/meta.yaml
+++ b/recipes/seqan3/meta.yaml
@@ -1,0 +1,38 @@
+{% set name = "SeqAn3" %}
+{% set version = "3.0.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/seqan/{{ name|lower }}/releases/download/{{ version }}/seqan-{{ version }}-with-submodules.tar.gz
+  sha256: ba8f0fd5d19289f95823235b69fdc5a12aba3578292463eecf750722e8662bd3
+  patches:
+    - bioconda.patch
+
+build:
+  noarch: generic
+  number: 0
+
+requirements:
+  run:
+    - bzip2
+    - zlib
+
+test:
+  commands:
+    - test -f $PREFIX/share/cmake/seqan3-config.cmake
+    - test -f $PREFIX/include/seqan3/version.hpp
+
+about:
+  home: https://www.seqan.de
+  development: https://github.com/seqan/seqan3
+  license: BSD-3-Clause
+  license_family: BSD
+  license_file: LICENSE.md
+  summary: 'SeqAn3 is the new version of the popular SeqAn template library for the analysis of biological sequences.'
+
+extra:
+  maintainers:
+    - eseiler
--- a/recipes/seqan3/bioconda.patch
+++ b/recipes/seqan3/bioconda.patch
@@ -1,0 +1,35 @@
+From: Michael R. Crusoe <michael.crusoe@gmail.com>
+Subject: Support discover of vendored dependency headers
+--- seqan3.orig/build_system/seqan3-config.cmake
++++ seqan3/build_system/seqan3-config.cmake
+@@ -158,6 +158,7 @@
+             endif ()
+         endforeach ()
+     endif ()
++
+ endif ()
+
+ # ----------------------------------------------------------------------------
+@@ -168,6 +169,22 @@
+     find_path (SEQAN3_INCLUDE_DIR "seqan3/version.hpp" HINTS ${SEQAN3_DEPENDENCY_INCLUDE_DIRS})
+ endif ()
+
++# ----------------------------------------------------------------------------
++# Find any headers that are vendored below our include directory, for example in a Debian Package
++# ----------------------------------------------------------------------------
++
++if (NOT SEQAN3_CLONE_DIR)
++    if (EXISTS "${SEQAN3_INCLUDE_DIR}/seqan3/submodules")
++        file (GLOB submodules ${SEQAN3_INCLUDE_DIR}/seqan3/submodules/*/include)
++        foreach (submodule ${submodules})
++            if (IS_DIRECTORY ${submodule})
++                seqan3_config_print ("  …adding submodule include:  ${submodule}")
++                set (SEQAN3_DEPENDENCY_INCLUDE_DIRS ${submodule} ${SEQAN3_DEPENDENCY_INCLUDE_DIRS})
++            endif ()
++        endforeach ()
++    endif ()
++endif ()
++
+ mark_as_advanced (SEQAN3_INCLUDE_DIR)
+
+ # find include directory


marehr:
Thank you.

Since seqan3 is a complete new project, I would also want to add a new package called seqan3.

Similar to python2 and python3.

johnlees:
Ok, I have moved those files into a new directory and therefore recipe, and reverted the changes to v2.

Have the location of the header files changed? Could you take a look at `build.sh`?

marehr:
I'll do.

recipes/seqan3/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - cmake
+
+test:
+  commands:
+    - test -e ${PREFIX}/include/seqan/align.h
+    - test -e ${PREFIX}/share/cmake/seqan/seqan-config.cmake
+    - test -e ${PREFIX}/share/pkgconfig/seqan-2.pc
```
marehr: We don't have this anymore.

```suggestion

```

recipes/seqan3/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+source:
+  url: https://github.com/seqan/{{ name|lower }}/releases/download/{{ version }}/seqan-3.0.0-with-submodules.tar.gz
+  sha256: ba8f0fd5d19289f95823235b69fdc5a12aba3578292463eecf750722e8662bd3
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
```
marehr: We are c++ only so I think we don't need this.

recipes/seqan3/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - {{ compiler('cxx') }}
+    - cmake
+
+test:
+  commands:
+    - test -e ${PREFIX}/include/seqan/align.h
+    - test -e ${PREFIX}/share/cmake/seqan/seqan-config.cmake
```
marehr: This path changed. And I'm not to sure if the `seqan3-config.cmake` should be in the `/share/cmake/seqan/` folder.

```suggestion
    - test -e ${PREFIX}/share/cmake/seqan3/seqan3-config.cmake
```

recipes/seqan3/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - test -e ${PREFIX}/share/pkgconfig/seqan-2.pc
+
+about:
+  home: http://docs.seqan.de/seqan/3-master-user/
```
marehr: Our official website is https://www.seqan.de.

Can we provide a project url like https://github.com/seqan/seqan3 or a documentation link that would be http://docs.seqan.de/seqan/3-master-user/?

recipes/seqan3/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/seqan/{{ name|lower }}/releases/download/{{ version }}/seqan-3.0.0-with-submodules.tar.gz
+  sha256: ba8f0fd5d19289f95823235b69fdc5a12aba3578292463eecf750722e8662bd3
+
+build:
+  number: 0
```
marehr: I imagine this must be set to show that we are header only?

recipes/seqan3/meta.yaml
```diff
@@ -0,0 +1,32 @@
+{% set name = "SeqAn3" %}
+{% set version = "3.0.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/seqan/{{ name|lower }}/releases/download/{{ version }}/seqan-3.0.0-with-submodules.tar.gz
+  sha256: ba8f0fd5d19289f95823235b69fdc5a12aba3578292463eecf750722e8662bd3
```
marehr: I have seen other versions where sha256 is defined above. I imagine this is okay.

recipes/seqan3/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+install -d $PREFIX/include
+install -d $PREFIX/share/cmake
+install -d $PREFIX/share/doc
+install -d  $PREFIX/share/pkgconfig
```
marehr: We don't have this anymore.

```suggestion

```

recipes/seqan3/build.sh
```diff
@@ -0,0 +1,10 @@
...
+install include/seqan $PREFIX/include
+install share/cmake/seqan $PREFIX/share/cmake
+install share/doc/seqan $PREFIX/share/doc
+install share/pkgconfig/* $PREFIX/share/pkgconfig
```
marehr: We don't have this anymore.

```suggestion

```

recipes/seqan3/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+install -d $PREFIX/include
+install -d $PREFIX/share/cmake
+install -d $PREFIX/share/doc
```
marehr: The documentation must be build from source (uses `doxygen` that generates html files, not to sure how to add `doxygen` generated documentation to  conda packages)

documentation can be build like this:

```terminal
/my/path/to/seqan3

mkdir build_doc
cd build_doc
cmake /my/path/to/seqan3/test/documentation
make doc_usr
```

in `build_doc/doc_usr/html/` is the complete html documentation.

recipes/seqan3/build.sh
```diff
@@ -0,0 +1,10 @@
...
+install -d  $PREFIX/share/pkgconfig
+install include/seqan $PREFIX/include
```
marehr:
```suggestion
install include/seqan3 $PREFIX/include
```

recipes/seqan3/build.sh
```diff
@@ -0,0 +1,10 @@
...
+install share/cmake/seqan $PREFIX/share/cmake
```
marehr: `build_system/seqan3-config.cmake`

recipes/seqan3/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - cmake
+
+test:
+  commands:
+    - test -e ${PREFIX}/include/seqan/align.h
```
marehr:
```suggestion
    - test -e ${PREFIX}/include/seqan3/version.hpp
    - test -e ${PREFIX}/include/seqan3/core/platform.hpp
```

recipes/seqan3/meta.yaml
johnlees: I get confused as to when you still need both, but can try removing

recipes/seqan3/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
```
marehr: I think we could also do the package/install with cmake, but it will require more reading how to do it. For now let's do it like this.

https://cmake.org/cmake/help/v3.15/command/install.html

recipes/seqan3/meta.yaml
johnlees: I think whichever you'd prefer really, it's the link that comes up on the anaconda download page (e.g. https://anaconda.org/conda-forge/seqan-library), but command line install is the usual method so it's not usually seen

recipes/seqan3/meta.yaml
johnlees: build must be set, and starts from 0 for each new version. It is incremented when recompiling the same source with update libraries, for example

recipes/seqan3/meta.yaml
johnlees: I think this is preferred, as it's only used once

recipes/seqan3/meta.yaml
marehr: >  Home: https://www.seqan.de
Development: https://github.com/seqan/seqan 

I think we can use the same format, just updating the Development url.

recipes/seqan3/build.sh
johnlees: Can you access it with `man`? If not, might be best just to leave it out of this package as it's available online

marehr: We can also build `man` pages. Just tried it out.

For that we need to add `GENERATE_MAN = YES` to the generated Doxyfile `build_doc/doc_usr/Doxyfile`. We might need to add a `OPTION` in `cmake` to make this work.

recipes/seqan3/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - test -e ${PREFIX}/include/seqan3/version.hpp
+    - test -e ${PREFIX}/include/seqan3/core/platform.hpp
+    - test -e ${PREFIX}/share/cmake/seqan3/seqan3-config.cmake
+
+about:
+  home: https://www.seqan.de
+  development: https://github.com/seqan/seqan3
+  license: BSD-3-Clause
+  license_family: BSD
+  license_file: LICENSE.md
```
marehr: Is this `license_file` relative to the downloaded seqan source?

johnlees: I believe so

johnlees:
I think it might be best to include `cereal`, `range-v3` and `sdsl-lite` as dependencies in `meta.yaml` (and use the standard github release), otherwise these also need to be installed from the submodules directory.

Also recursing with `install` seems to be a pain, so it's probably best to revert to `mkdir` and `cp -r`

marehr:
I wanted to ask how you handle our external dependencies and if you could directly use a git checkout.

So you want to create / reuse a conda package for each of the external libraries?

Do you know how to read the ci-log of bioconda?

[Where can I see what went wrong?](https://circleci.com/gh/bioconda/bioconda-recipes/63930?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link)

marehr:
I think we need to update the PR description. We technically add a new receipt and do not update an existing one.

johnlees:
> I wanted to ask how you handle our external dependencies and if you could directly use a git checkout.
>
> So you want to create / reuse a conda package for each of the external libraries?

`sdsl-lite` has a package on conda-forge, but I think it is not the development version you are using. `cereal` and `range-v3` do not have packages so would need to be added to conda-forge which would be a bit of a pain.

So it would probably be best to leave them as submodules. The current package downloaded includes these, but an alternative is to have something like:
```
source:
  git_url: https://github.com/seqan/{{ name|lower }}.git
  git_rev: v{{ version }}

extra:
   skip-lints:
     - uses_git_url
```

The only issue is where they are installed to - it would be bad to overwrite an existing `sdsl-lite` install for example. Can these also go in the `seqan3` include directory?

> Do you know how to read the ci-log of bioconda?
> 
> Where can I see what went wrong?

In the build log, at the bottom:
```
19:52:55 BIOCONDA INFO (OUT) install: omitting directory `/opt/conda/conda-bld/seqan3_1562615495023/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeh/include'
19:52:57 BIOCONDA INFO (OUT) Traceback (most recent call last):
19:52:57 BIOCONDA INFO (OUT)   File "/opt/conda/bin/conda-build", line 11, in <module>
19:52:57 BIOCONDA INFO (OUT)     sys.exit(main())
19:52:57 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.6/site-packages/conda_build/cli/main_build.py", line 456, in main
19:52:57 BIOCONDA INFO (OUT)     execute(sys.argv[1:])
19:52:57 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.6/site-packages/conda_build/cli/main_build.py", line 447, in execute
19:52:57 BIOCONDA INFO (OUT)     verify=args.verify, variants=args.variants)
19:52:57 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.6/site-packages/conda_build/api.py", line 208, in build
19:52:57 BIOCONDA INFO (OUT)     notest=notest, need_source_download=need_source_download, variants=variants)
19:52:57 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.6/site-packages/conda_build/build.py", line 2314, in build_tree
19:52:57 BIOCONDA INFO (OUT)     notest=notest,
19:52:57 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.6/site-packages/conda_build/build.py", line 1477, in build
19:52:57 BIOCONDA INFO (OUT)     cwd=src_dir, stats=build_stats)
19:52:57 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.6/site-packages/conda_build/utils.py", line 374, in check_call_env
19:52:57 BIOCONDA INFO (OUT)     return _func_defaulting_env_to_os_environ('call', *popenargs, **kwargs)
19:52:57 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.6/site-packages/conda_build/utils.py", line 354, in _func_defaulting_env_to_os_environ
19:52:57 BIOCONDA INFO (OUT)     raise subprocess.CalledProcessError(proc.returncode, _args)
19:52:57 BIOCONDA INFO (OUT) subprocess.CalledProcessError: Command '['/bin/bash', '-e', '/opt/conda/conda-bld/seqan3_1562615495023/work/conda_build.sh']' returned non-zero exit status 1.
19:52:59 BIOCONDA ERROR COMMAND FAILED: docker run -t --net host --rm -v /tmp/tmpwumu_b3o/build_script.bash:/opt/build_script.bash -v /home/circleci/project/miniconda/conda-bld:/opt/host-conda-bld -v /home/circleci/project/recipes/seqan3:/opt/recipe -e LANG=en_US.UTF-8 -e HOST_USER_ID=1001 bioconda/bioconda-utils-build-env:latest /bin/bash /opt/build_script.bash

19:52:59 BIOCONDA ERROR BUILD FAILED recipes/seqan3
```
The `install` command in `build.sh` is wrong - it needs changing to `cp -r`

(you should be able to download the log as plaintext if your browser doesn't load the whole thing)

> I think we need to update the PR description. We technically add a new receipt and do not update an existing one.

I will update that now

recipes/seqan3/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+  url: https://github.com/seqan/{{ name|lower }}/releases/download/{{ version }}/seqan-3.0.0-with-submodules.tar.gz
+  sha256: ba8f0fd5d19289f95823235b69fdc5a12aba3578292463eecf750722e8662bd3
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - cmake
```
eseiler: Can you add `zlib` and `bzip2` ?

recipes/seqan3/meta.yaml
eseiler: Propbably have to add `skip: True  # [osx]` since it uses clang on osx. we currently do not support clang. If, however, it is possible to specify to use gcc on osx, we could build it.

recipes/seqan3/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - test -e ${PREFIX}/include/seqan3/core/platform.hpp
+    - test -e ${PREFIX}/share/cmake/seqan3/seqan3-config.cmake
+
+about:
+  home: https://www.seqan.de
+  development: https://github.com/seqan/seqan3
+  license: BSD-3-Clause
+  license_family: BSD
+  license_file: LICENSE.md
+  summary: 'SeqAn3 is the new version of the popular SeqAn template library for the analysis of biological sequences.'
```
eseiler: Is it common to add
```
extra:
  recipe-maintainers:
    - ...
```
?

marehr:
> `sdsl-lite` has a package on conda-forge, but I think it is not the development version you are using. `cereal` and `range-v3` do not have packages so would need to be added to conda-forge which would be a bit of a pain.

Our sdsl-lite version is also the new, unreleased 3.0 version, so that would need a new package.

> Can these also go in the `seqan3` include directory?

You mean?
```
+ seqan3/
    + sdsl-lite/
    + range-v3/
    + cereal/
```

This would only work if each of those folders would be in the include path.

We assume normally the following include folder structure.
```
+ seqan3/
+ sdsl-lite/
+ range-v3/
+ cereal/
```

But, you are right overwriting `sdsl-lite/` would be wrong. Do you know how conda handles packaging multi-versions of libraries?

recipes/seqan3/meta.yaml
johnlees: I believe this is unnecessary now (definitely for `zlib` but probably for `bzip2`), as they are always included

recipes/seqan3/meta.yaml
johnlees: I think that's only on conda-forge, as bioconda doesn't spawn off a repo for each recipe

johnlees:
> > `sdsl-lite` has a package on conda-forge, but I think it is not the development version you are using. `cereal` and `range-v3` do not have packages so would need to be added to conda-forge which would be a bit of a pain.
> 
> Our sdsl-lite version is also the new, unreleased 3.0 version, so that would need a new package.
> 
> > Can these also go in the `seqan3` include directory?
> 
> You mean?
> 
> ```
> + seqan3/
>     + sdsl-lite/
>     + range-v3/
>     + cereal/
> ```
> 
> This would only work if each of those folders would be in the include path.
> 
> We assume normally the following include folder structure.
> 
> ```
> + seqan3/
> + sdsl-lite/
> + range-v3/
> + cereal/
> ```
> 
> But, you are right overwriting `sdsl-lite/` would be wrong. Do you know how conda handles packaging multi-versions of libraries?

I'm not sure how to handle this, or what would be preferred. Perhaps somebody in @bioconda/core could advise?

recipes/seqan3/meta.yaml
eseiler: That's cool, I thought they were just automatically pinned now.

recipes/seqan3/meta.yaml
eseiler: I did a grep, and sometime `recipe-maintainers:` is added, but not very often (about 400 out of 6400).
But updating the recipe will be part of our release procedure anyways.

eseiler:
> But, you are right overwriting `sdsl-lite/` would be wrong. Do you know how conda handles packaging multi-versions of libraries?

Do you mean having multiple version of a library? This works and you can set a version you want to use. But, for example it doesn't make much sense, imo, to package `range-v3` since it's technically a reference implementation for C++20 - `sdsl-lite` is not yet released and `cereal` is also currently a fork (we need yet a upstream release with some fixes we need).

Could we put the `submodules` folder somewhere and hard code the includes?

marehr:
> Could we put the `submodules` folder somewhere and hard code the includes for seqan3?

This depends on how someone requires seqan3 as a conda package and how conda packages glues stuff together. You have to show me tomorrow how everything works together and then we can think of a way to make it work for local private versions.

recipes/seqan3/meta.yaml
bgruening: We should also add `run_exports` like here: https://github.com/conda-forge/zlib-feedstock/blob/master/recipe/meta.yaml#L13

mr-c:
For the Debian package we are vendoring the needed dependencies: making local copies and teaching the `seqan3-config.cmake` how to find them

https://salsa.debian.org/med-team/seqan3/commit/324a84086cae519ae1a4268f63471d88582a4691

eseiler:
I worked on it for a bit. Thanks to @mr-c patch it also kinda works now.
The only problem for now is that there is only gcc7.3, which doesn't have `std::aligned_alloc`. That means `container_of_container_test` and `aligned_allocator_test` won't compile.
I'll try excluding it from the test suite later.

mr-c:
@eseiler Cool!

In regards to ` recipes/seqan3/run_test.sh`, that is how we test `SeqAn3` post-install. During the build we do the equivalent of `make test` in the build directory; which is much simpler :-)

eseiler:
This seems to be it.
It's a bit easier than I thought.
But I can't build any tests to see if it works properly - the compiler is not available in a mulled build.
(Still, the non-mulled worked as expected.)
I'm just checking file existence now.

recipes/seqan3/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  url: https://github.com/seqan/{{ name|lower }}/releases/download/{{ version }}/seqan-{{ version }}-with-submodules.tar.gz
+  sha256: ba8f0fd5d19289f95823235b69fdc5a12aba3578292463eecf750722e8662bd3
+  patches:
+    - bioconda.patch
+
+build:
+  noarch: generic
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('seqan3', max_pin='x.x') }}
```
dpryan79: This is a header-only library now, yes? I don't think we need a run-export then. Will this require anything else during runtime though? That is, does using seqan require anything like zlib, or would that be from the program that uses seqan then?

@bgruening Shouldn't the run-export set the run-time pinning, which won't actually be relevant?

bgruening: Yes, run-exports will set run-time pinning. If this is a header-only package it is only used during build-time isn't it?

eseiler: This is a header only library, and we will use zlib/bzip2 if it is available (I'd actually prefer those packages to be installed together with this package, since stuff like BAM support relies on this).

dpryan79: Should we add zlib/bzip2 to the run time and then a run_exports for them then? I'm not entirely sure how all of that works... :(

bgruening: They will only end up in the run-time deps but we will need them in the host section. I don't think this helps. We should remove the run_exports here.

dpryan79: I guess if user-programs end up needing zlib/bzip2 they'll notice that anyway during testing, add that to the host/run and then the normal pinning work resolve the issues. So I guess we don't need to worry about somehow shoehorning them in here.

eseiler: Like this, or both host and run requirements?

dpryan79:
This looks good to me, if any user-applications start having issues we can update this as needed then. Great job everyone!

recipes/seqan3/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  patches:
+    - bioconda.patch
+
+build:
+  noarch: generic
+  number: 0
+
+requirements:
+  run:
+    - bzip2
```
bgruening: Mh, this does not make much sense imho. People (seqan users) still need to put zlib and co into their host-sections, so also have it during build-time - and this in turn puts them into the run-time section of the users-package. So its save to remove this imho.

eseiler: I can just remove the requirements section with the next update, if that is soon enough :)

bgruening: Sure :)

PR #16273, opened 2019-07-09
include unifrac versions and unit tests

gwarmstrong:
This PR includes version requirements to restrict the build to only support versions of dependencies that the (unifrac)[https://github.com/biocore/unifrac] developers explicitly support. Additionally, this PR runs the `unifrac` unittests and tests whether the `faithpd` CLI introduced in 0.10.0 is available to the user. The description has also been updated to reflect the more general nature of 0.10.0

### DIFF
--- a/recipes/unifrac/meta.yaml
+++ b/recipes/unifrac/meta.yaml
@@ -9,7 +9,6 @@ source:
   sha256: 97507caa5536bf445ca712ec495a39c4c95db3388d3259be3270e0e15ceb5ce5
 
 build:
-  preserve_egg_dir: True
   number: 0
   skip: True # [py27]
 
@@ -20,26 +19,30 @@ requirements:
   host:
     - python
     - pip
-    - hdf5
-    - cython
-    - numpy
+    - hdf5 >=1.8.17
+    - cython >=0.28.3
+    - numpy >=1.12.1
 
   run:
     - python
     - {{ pin_compatible('numpy') }}
-    - h5py
+    - h5py >=2.7.0
     - biom-format
-    - scikit-bio
+    - scikit-bio >=0.5.1
 
 test:
+  requires:
+    - nose
   imports:
     - unifrac
   commands:
     - ssu --help
+    - faithpd --help
+    - nosetests unifrac
 
 about:
   home: https://github.com/biocore/unifrac
   license: BSD-3-Clause
   license_family: BSD
-  summary: 'High performance UniFrac'
+  summary: 'Fast phylogenetic diversity calculations'
 
--- a/recipes/unifrac/meta.yaml
+++ b/recipes/unifrac/meta.yaml
@@ -31,14 +31,12 @@ requirements:
     - scikit-bio >=0.5.1
 
 test:
-  requires:
-    - nose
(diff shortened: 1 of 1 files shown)


gwarmstrong:
It seems that the test environment does not end up installing/importing `nosetests` which I had listed in the `tests/requires` section of the recipe. I believe this is occurring because the testing environment is initialized only by installing the `unifrac` package it built previously.
```
21:49:50 BIOCONDA INFO (ERR) [Jul  9 21:49:50] STEP Run image [quay.io/dpryan79/mulled_container:latest] with command [[/bin/sh -c conda install  -c conda-forge -c file:///home/circleci/project/miniconda/conda-bld -c bioconda -c defaults  unifrac=0.10.0=py37h6bb024c_1 -p /usr/local --copy --yes --quiet]]
```
It also seems that the testing is performed in a separate working directory from where the git repo is cloned, so tests cannot easily be run from local files.

I have tried to get around this for the moment by running unittest instead of nosetest, but this approach is brittle since I had to list the individual testing modules in `unifrac`.

@bioconda/core Is there anything I may have missed that would result in a more elegant solution?

dpryan79:
Hmm, this looks a bit like a conda error somewhere.

gwarmstrong:
> Hmm, this looks a bit like a conda error somewhere.

@dpryan79  Thanks! I think it is at least bioconda-specific in some way. The above paradigm worked fine on `conda-forge`. See https://github.com/conda-forge/staged-recipes/pull/8755 commits 10631d9 and 1c3e806

I guess it would have been more helpful if I included this overall error at the end from the `bioconda` build:
```
21:53:31 BIOCONDA ERROR COMMAND FAILED: mulled-build build-and-test unifrac=0.10.0--py37h6bb024c_1 -n biocontainers --test ssu --help && faithpd --help && nosetests unifrac --channels conda-forge,file:///home/circleci/project/miniconda/conda-bld,bioconda,defaults --involucro-path /home/circleci/project/miniconda/bin/involucro
```
To me this says that the `mulled-build build-and-test` command does not know that it needs to import `nose` even though it is specified in the test requirements. Probably depends either on how that command is set up by `bioconda`, which I think makes more sense since I see no mention of `nose` in that command, or an issue with galaxy-lib?

gwarmstrong:
The above aside, I think the outstanding question for the time being is if we want to run our module's unittests in our conda-build for `bioconda`, what is the canonical way to do so?
@bioconda/python

dpryan79:
Typically the testing in recipes is very minimal (for python packages usually just import and a couple `foo --help` sorts of tests). Having said that, we have the build capacity to support some quick unit tests at the moment.

gwarmstrong:
@ebolyen mind commenting on this PR?

ebolyen:
Nothing stands out to me here, LGTM. How quickly do these tests run?

gwarmstrong:
> Nothing stands out to me here, LGTM. How quickly do these tests run?

A few seconds, maybe up to 15s or so? Pretty minimal, I would like to include them because I have had issues in the past where UniFrac has imported fine but the C bindings are messed up.

PR #16344, opened 2019-07-12
Add ELECTOR: EvaLuator of Error Correction Tools for lOng Reads

natir:

### DIFF
--- a/recipes/elector/build.sh
+++ b/recipes/elector/build.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+set -x
+
+mkdir -p $PREFIX/bin/
+# build and install binary tools
+cd src/simulator
+make
+cp simulator $PREFIX/bin/
+
+cd ../poa-graph
+make poa
+cp poa $PREFIX/bin/
+
+cd ../utils
+make
+cp fq2fa $PREFIX/bin/
+
+cd ../split
+make
+cp masterSplitter $PREFIX/bin/
+cp Donatello $PREFIX/bin/
+cd ../..
+
+# setup and install python script
+mkdir -p $PREFIX/share/elector/
+cp src/poa-graph/blosum80.mat $PREFIX/share/elector/
+
+sed -i "s#os.path.dirname(os.path.realpath(__file__))+\"/../bin/\"#${PREFIX}/bin/#" elector/utils.py
+sed -i "s#os.path.dirname(os.path.realpath(__file__))+\"/../src/poa-graph/\"#${PREFIX}/share/elector/#" elector/utils.py
+
+$PYTHON setup.py install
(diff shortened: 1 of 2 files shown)


natir:
Hi, 

I didn't understand why my build failed I have `- {{ compiler('cxx') }}` in build requirements.

Anyone have an idea ?

Pierre

npavlovikj:
The error is `g++: Command not found`. This usually means that the Makefile has `g++` hardcoded that is being used instead of the conda compilers (https://github.com/kamimrcht/ELECTOR/blob/master/src/simulator/makefile). Using `make CC="${CC}" CXX="${CXX}"` for `simulator` should fix that (or patching the Makefile is another approach).

However, compiling dependencies as part of the main recipe is not a good practice, so I would strongly suggest you to create separate conda packages for all dependencies needed for `elector`.

natir:
Hi,

I understand the problem and how to solve it, but I didn't understand why call `g++` in [sparc build.sh](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/sparc/build.sh) work and it not work in my makefile ?

And all binary build in this package isn't dependency it's a part of the package.

Thanks for your help.

recipes/elector/build.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+set -x
+
+mkdir -p $PREFIX/bin/
+# build and install binary tools
+cd src/simulator
+sed -i "s#CC=g++#CC=${CXX}#" makefile
+make
```
mbargull: Please use @npavlovikj's suggestion `make CC="${CC}" CXX="${CXX}"` instead. It's more robust and portable than `sed` + `make`.

recipes/elector/build.sh
```diff
@@ -0,0 +1,39 @@
...
+make poa
+cp poa $PREFIX/bin/
+
+cd ../utils
+sed -i "s#CC=g++#CC=${CXX}#" makefile
+make
+cp fq2fa $PREFIX/bin/
+
+cd ../split
+sed -i "s#CC=g++#CC=${CXX}#" makefile
```
mbargull: Interesting, looks like you change is correct, meaning it's wrong in the source.
In this case, just use `make CC="${CXX}" CXX="${CXX}"`.

mbargull:
> I didn't understand why call `g++` in [sparc build.sh](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/sparc/build.sh) work and it not work in my makefile ?

It probably won't work with `sparc` now either. There has been a change in the build system (updated compilers) a couple of months back.

recipes/elector/build.sh
natir: You're right I think that since the variable has been defined in the makefile the change in the command line would not be taken into account.

We learn something new every day.

Thank

natir:
Thank you for helping both of you.

dpryan79:
@bioconda-bot merge

PR #16380, opened 2019-07-15
Rpmcmc

simonvh:


recipes/rpmcmc/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  build:
+    - {{ compiler('cxx') }}
+
+test:
+  commands:
+    - multi_motif_sampler 2>&1 | grep "usage"
+
+about:
+  home: http://daweb.ism.ac.jp/yoshidalab/motif/
+  license: Other
```
bgruening: Does the license allow redistribution of the software?

simonvh: There is no license...

simonvh: So @bgruening , is a license required for inclusion in bioconda?

bgruening: Yes, no license is legally very complicated. We need to be sure that we are allowed to distribute the software.

simonvh: I understand that. I'll close the PR for now. Would an OK from the author be sufficient? Or would we really need something in writing (ie. a license :))?

bgruening: Leave the PR open and point the author to this issue. This is probably just unintentional. A lot of Bioinformatic software has no license in the beginning, it's nothing that we are trained at.

The License file would be super, but if this is strictly needed depends on the license that will be chosen. GPL, for example, requires a license file.

Issue #3859, opened 2017-02-19
picard has occasional need for R


jefferys:
I'm not familiar with bioconda recipes and packaging, so I apologize in advance if I'm wrong about where the error I'm seeing occurs.

I'm trying to use a docker container with picard. The dockerfile builds the container using bioconda (biodckr/picard). It mostly works fine, but at least one picard tools command, "CollectInsertSizeMetrics", uses R *(Rscript). Is R a dependency that will be auto-installed when installing picard as described in the recipe? If not, should it be? It is a big dependency that is only used in a couple of places, I think just to plot graphics/make pdfs.

Any help is appreciated.

daler:
Using just conda, the fix would be to add an additional `conda install r` to have R alongside picard, but this doesn't work when it's used as a container. There's not really a good mechanism for auto-installing only when needed, so R should probably be added as a dependency. @bioconda/core, thoughts?

bgruening:
Optional dependencies are dependencies and I would declare them as such. For example in secure environments you don't even have network connection to install things after initial deployment.

rvalieris:
picard is a collection of different tools, and not every tool needs R.
maybe a compromise would be to have a `picard+r` recipe to generate a container with both packages.

bgruening:
@raphenya does the R dependency hurt? I'm not really thinking about the Docker stuff, but also plain conda environments for sensitive data. I guess you would expect that if you install picard you get the entire package and all tools will just work.

chapmanb:
I'm -1 on having R as a dependency of Picard. That's a pretty heavy weight requirement for a small subset of the tools. I agree with Renan's suggestion for a meta-package.

jefferys:
I suspect this doesn't come up except in the context of docker-type installs because many people  who use picard are doing data analysis and hence are likely to need R. They will never notice the missing dependency. That means the dependency, although heavy, is not likely to matter practically.

How optional tools are treated is a philosophical line-drawing problem. In this case, at least one component throws a messy java error if R does not exist, and R being a dependency is not clearly discussed in the picard documentation. My preference would be for the default picard recipe to include R, but maybe have an optional picard-lite package that leaves out R.

[...]

PR #16398, opened 2019-07-16
Add r-base as a dependency

rhpvorderman:
r-base is needed for picard CollectMultipleMetrics, CollectInsertSizeMetrics etc. (basically anything with metrics).


bgruening:
https://github.com/bioconda/bioconda-recipes/issues/3859 :(

rhpvorderman:
> #3859 :(

@bgruening. I posted my opinion on the issue.

rhpvorderman:
@bgruening , given that there was no great backlash on the issue, I think that nobody strongly cares about this. Shall we go ahead and merge this?

bgruening:
@dpryan79 any opinion?

dpryan79:
I’m a very weak “no” on this. Picard is just so bloated as is and I have to say this is the first time I’d heard that Picard did any plotting on this.

cbrueffer:
I tend to agree that things should work out of the box, although I have to admit I shared Devon's unawareness regarding the R dependency. Given the bloat having a `picard-lite` recipe and pointing to that in the `picard` install message would be good.

rhpvorderman:
@bgruening @dpryan79 @cbrueffer 
The previous iteration of picard has been renamed to picard-slim. 
While copying the package I found that a bash wrapper script is used to run picard instead of a python wrapper script. So I removed the python dependency in both picard and picard-slim.

Is this compromise to your liking?

dpryan79:
:+1: from me. @bgruening @cbrueffer thoughts?

cbrueffer:
Looks good; a post-install message in each of the recipes pointing to the other one would be nice ("A package without the R dependency is available as `picard-slim` or similar); see e.g. `recipes/braker2/post-link.sh`.

bgruening:
Fine with me ...

rhpvorderman:
> Looks good; a post-install message in each of the recipes pointing to the other one would be nice ("A package without the R dependency is available as picard-slim or similar); see e.g. recipes/braker2/post-link.sh.

@cbrueffer  I do agree that users need to be notified. While working on the post-link script however, I felt that I had to write a really, really good message that would be worth it for the user to see every  time they install picard. This is IMO not possible. It is a very  annoying way to let users know.

I have added information about picard and picard-slim  to the package descriptions, so that users can make an informed decision on which package to install based on the information on bioconda.github.io.

cbrueffer:
Thanks @rhpvorderman , I like the idea.

alanhoyle:
Can this be added to the gatk4 as a dependency as well?

PR #16403, opened 2019-07-16
Hulk

will-rowe:

### DIFF
--- a/recipes/hulk/meta.yaml
+++ b/recipes/hulk/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "0.1.2" %}
+{% set version = "1.0.0" %}
 
 package:
   name: hulk
@@ -6,7 +6,7 @@ package:
 
 source:
   url: https://github.com/will-rowe/hulk/archive/{{ version }}.tar.gz
-  sha256: d27a5e862be461861e10a8c36835749b208d2b1c997e0d81e297525b592f5e64
+  sha256: c3a9f80c3918551304a1c639f1e157771c97f2437cbbacf4837795a3456b6dbf
 
 build:
   number: 1
--- a/recipes/hulk/meta.yaml
+++ b/recipes/hulk/meta.yaml
@@ -10,11 +10,10 @@ source:
 
 build:
   number: 0
-  noarch: generic
 
 requirements:
   build:
-    - go
+    - {{ compiler('go') }}
     - mercurial
 
 test:
--- a/recipes/hulk/meta.yaml
+++ b/recipes/hulk/meta.yaml
@@ -9,7 +9,8 @@ source:
   sha256: c3a9f80c3918551304a1c639f1e157771c97f2437cbbacf4837795a3456b6dbf
 
 build:
-  number: 1
+  number: 0
+  skip: False
 
 requirements:
   build:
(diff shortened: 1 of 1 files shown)


recipes/hulk/meta.yaml
```diff
@@ -1,15 +1,16 @@
...
 
 source:
   url: https://github.com/will-rowe/hulk/archive/{{ version }}.tar.gz
-  sha256: d27a5e862be461861e10a8c36835749b208d2b1c997e0d81e297525b592f5e64
+  sha256: c3a9f80c3918551304a1c639f1e157771c97f2437cbbacf4837795a3456b6dbf
 
 build:
-  number: 1
+  number: 0
+  skip: False
```
bgruening: you can remove this skip line if you like

recipes/hulk/meta.yaml
```diff
@@ -24,4 +25,9 @@ test:
 about:
   home: https://github.com/will-rowe/hulk
   license: MIT
-  summary: Histosketching Using Little Kmers.
+  summary: Histosketching Using Little Kmers
+
+extra:
+  skip-lints:
+    - should_use_compilers
+    - should_be_noarch_generic
```
bgruening: Why have you chosen to not make it `noarch: generic`?

will-rowe: I got a bit confused there. Circle error first time was:
```
ERROR: recipes/hulk/meta.yaml:13: should_be_noarch_generic: The recipe should be build as `noarch`
```
I thought I had to change it to should_be_noarch_generic

Do I need to change it as you suggest?

bgruening: Try removing the skip-lints and adding `noarch: generic` instead to the build section. Read more here: https://bioconda.github.io/contributor/linting.html#should-be-noarch-generic

will-rowe: Tried this and now failing again. This recipe is using the Go compiler - as this was said to be better than shipping the binaries when I last asked. Should I be just giving binaries now or should I be sticking with compilation?

bgruening: The linter is still unhappy: `ERROR: recipes/hulk/meta.yaml:17: should_use_compilers: The recipe requires a compiler directly`

Are you using go to compile your recipe? Then it should go to the reuquirements/build section - the go compiler I mean.

will-rowe: Yes - the go compiler is being used. It's already in the meta as this is just an updated recipe, normally it passes CI fine when I just update the version number and sha256 in my recipes.

mbargull: AFAIK the `bioconda-utils` linter code has changed a bit which is why you may see different behavior than before.
What we need for this recipe is 
1. ```diff
   -  noarch: generic
   ```
    (can't be `noarch` since we have compiled code here)
2. ```diff
   -      - go
   +      - {{ compiler('go') }}
   ```
3. maybe
   ```diff
   -      - mercurial
   ```
   ^-- is `mercurial` really used during the build?

Please apply those changes and then we can see/act upon linter/compile errors we get.

will-rowe: thanks! Mercurial was used by one of the Go packages imported by my tool - will see if I still need this or if the package has updated it's import method

will-rowe:
Thank you both for your help!

PR #16434, opened 2019-07-18
Add recipe for PDF::Table perl module

astaric:

While this is a general perl package, it is intended to be used with PDF::API2, which is available on bioconda, not conda-forge, so I am directing this PR to you.

### DIFF
--- a/recipes/perl-pdf-table/build.sh
+++ b/recipes/perl-pdf-table/build.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+# If it has Build.PL use that, otherwise use Makefile.PL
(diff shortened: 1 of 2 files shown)


astaric:
@bioconda/perl, this recipe passes tests and is ready for review.

As I wrote in the description, this is not directly biology related (though we use it in our pipeline), but since other packages it works with are available on bioconda, I would prefer to install them all from the same channel.

astaric:
@bioconda/core, ready for review

While this is not a bioinformatics package (and could be contributed to conda-forge), the module is used as a "helper" for perl-pdf-api2, which is only available on bioconda, so I would prefer to have it on the same channel.

PR #16484, opened 2019-07-22
Blast legacy osx

wclose:
@bioconda/core this PR adds the missing osx package. Please add me to the bioconda project so I can maintain future packages. Thanks!

### DIFF
--- a/recipes/blast-legacy/2.2.26/build.sh
+++ b/recipes/blast-legacy/2.2.26/build.sh
@@ -1,1 +1,2 @@
-cp -R bin $PREFIX
+### Copying binaries over to env bin
+cp -R bin/ "${PREFIX}"/bin/
--- a/recipes/blast-legacy/2.2.26/meta.yaml
+++ b/recipes/blast-legacy/2.2.26/meta.yaml
@@ -3,13 +3,13 @@ package:
   version: 2.2.26
 
 source:
-  fn: blast-2.2.26-x64-linux.tar.gz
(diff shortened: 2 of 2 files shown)


bgruening:
@bioconda-bot merge

recipes/blast-legacy/2.2.26/build.sh
```diff
@@ -1 +1,2 @@
-cp -R bin $PREFIX
+### Copying binaries over to env bin
+cp -R bin/ "${PREFIX}"/bin/
```
mbargull: Please be careful with such changes. In this case, where we don't have anything in the `requirements/host` section, this works fine. In other cases `${PREFIX}/bin` may exist beforehand such that this `cp` call will result in `${PREFIX}/bin/bin` being created. For a more detailed description of `cp`'s semantics, see https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cp.html.
Easiest thing would be to do
```sh
mv ./bin "${PREFIX}/"
```
if `${PREFIX}/bin` does not exist, or more generally,
```sh
mkdir -p "${PREFIX}/bin"
mv ./bin/* "${PREFIX}/bin/"
```
if `${PREFIX}/bin` *might* exist (`mv` semantics are similarly at https://pubs.opengroup.org/onlinepubs/9699919799/utilities/mv.html.)

(I hope the above doesn't read too much like "get schooled boy!" :X -- just trying to be helpful and spread knowledge ;) )

wclose: No worries, this is great! I really appreciate the feedback and will definitely keep that in mind going forward. Thanks!

Issue #14975, opened 2019-04-29
minced recipe update breaks prokka?


aschuerch:
With prokka=1.13.3, I get the error
[14:42:16] Looking for 'minced' - found /home/travis/build/aschuerch/bactofidia/.snakemake/conda/27a7b7f2/bin/minced
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.UnsupportedClassVersionError: minced has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0

The bioconda prokka=1.13.3 package worked fine until recently. I think it might be due to an update of the minced recipe. Is this possible @epruesse ?

epruesse:
It would be great if you could have a look at this!

I have no experience with Prokka or Minced. Just reading the error message it sounds like java needs to be pinned for packages providing JNI binaries, and you have one there that didn’t pin.


--
message sent from mobile device

aschuerch:
I'll give it a try!

Jingmingxia:
> I'll give it a try!

Does it work? I've met the same problem when I use the prokka.

aschuerch:
@dfornika did pull request #15407 fix the prokka conda recipe for you? I am still getting the same error with a fresh environment

PR #16528, opened 2019-07-24
pinning java as in #15407

aschuerch:
In addition to #15407, adding the java dependency is needed to firx the prokka recipe (see   #14975). Not sure if it is necessary to pin version. 

Please  `@bioconda/core` review my PR and add me to the bioconda project. 


bgruening:
ping @tseemann

tseemann:
@aschuerch thanks for looking into this. Prokka itselfv doesn't depend on Java, but it uses `minced` which does. I think minced recently updated, and the author built the .jar file with a more modern version of Java, so old Java can't execute it.  This happened previously, and I asked them to rebuild the .jar with Java 1.8 for compatibility.

So I think this should be moved to `minced` recipe?
https://bioconda.github.io/recipes/minced/README.html

aschuerch:
thank you @tseemann, I agree that the problem is minced not prokka itself. In the minced recipe, java is already a dependency, and the version was recently pinned (see  #15407 and  #14975) but this did not fix the problem, while adding java to a fresh prokka env does. If there is a better way, I'll be happy to learn about it.

dpryan79:
@tseemann Is this maybe related to the JAVA_HOME issue in minced 0.4.0?

@aschuerch A new version of minced was just pushed out, can you see if that fixes things?

aschuerch:
@dpryan79 @tseemann yes, the new version of minced does the trick! This pull request can be closed then.

dpryan79:
Great, thanks @aschuerch!

tseemann:
@aschuerch yes i think i've fixed minced, prokka, snippy, etc now so things should all work with Java now!

PR #16552, opened 2019-07-25
Mothur v1.42.3

wclose:
@bioconda/core this recipe updates the mothur version and should fix issues with gzipped files as part of the build. Please add me to the bioconda team. Thanks!


recipes/mothur/meta.yaml
```diff
@@ -1,35 +1,48 @@
...
 package:
   name: mothur
   version: {{ version }}
 
 source:
-  url: https://github.com/mothur/mothur/archive/v{{ version }}.tar.gz
-  sha256: 6b61591dda289ac2d8361f9c1547ffbeeba3b9fbdff877dd286bad850bbd5539
+  - url: https://github.com/mothur/mothur/archive/v.{{ version }}.tar.gz
+    sha256: {{ sha256 }}
+  # - url: https://support.hdfgroup.org/ftp/HDF5/releases/hdf5-1.10/hdf5-1.10.5/src/hdf5-1.10.5.tar.gz
```
bgruening: can you clean up this recipe a little bit?

recipes/mothur/1.42.1/build.sh
```diff
@@ -0,0 +1,8 @@
+# the flags the new gxx activate script sets breaks the build, so unset them here
```
bgruening: Do you think we should remove those old versions?

recipes/mothur/meta.yaml
wclose: Already planned on it. I was having issues with linking/libraries so those were remnants from my backup plan.

recipes/mothur/1.42.1/build.sh
wclose: Are you sure removal would be the best course? Not sure if that would mess with people that already have them implemented. I can apply the same fixes as this version and they 'should' compile as desired.

bgruening: Removing the recipe does not mean removing the package. Lets formulate it the other way around, keeping the subdir means people want to maintain this particular version. So it gets rebuild against new python versoin etc. Is that really needed for this particual version?

wclose: Ah, I understand what you're saying now. In that case, yeah, let's remove them. Should I just delete all the old subdirs then?

bgruening: Yes :)

wclose: Ok, it should be all set to go now. Thanks for your help!

bgruening:
Thanks @wclose!

PR #16585, opened 2019-07-27
Add recipe for CARNAC-LR

npavlovikj:


npavlovikj:
@bioconda/osx , can someone please help me with the OSX build? I am getting:
```
12:47:19 BIOCONDA INFO (ERR) Undefined symbols for architecture x86_64:
12:47:19 BIOCONDA INFO (ERR)   "___kmpc_barrier", referenced from:
12:47:19 BIOCONDA INFO (ERR)       _.omp_outlined. in clustering_cliqueness.o
12:47:19 BIOCONDA INFO (ERR)       _.omp_outlined..20 in clustering_cliqueness.o
...
```
errors.

The `CARNAC-LR` authors say that the OSX version doesn't work with `-fopenmp` and multiple threads. I tried with and without `llvm-openmp`, and modifying `LDFLAGS`, `CFLAGS` and `DYLD_LIBRARY_PATH` according similar recipes locaclly, but the OSX errors persist.

I would appreciate any help and suggestions!

acaprez:
If you look at the build output from both linux and OS X, it's not actually honoring the `CFLAGS` variable from the environment.  It's just that the linux build works even with the one from the makefile.  You need to pass `CFLAGS` on the `make` line to have it use the value from the environment, i.e. do `make CC=$CC CXX=$CXX LDFLAGS="$LDFLAGS" CFLAGS="$CFLAGS"`

Note the anaconda compilers set `-std=c++14` in `CFLAGS`, which broke the compile, at least on OS X.  To get it to build, I had to do `export CFLAGS="$CFLAGS -std=c++11"` in the Darwin if block in the build script.  By putting `-std=c++11` _after_ all the other flags, it will use that instead of c++14.  I suspect you'll have to do the same thing on linux, but I didn't check that.

npavlovikj:
@acaprez , thank you so much - your suggestions fixed the issue!
Both builds passed now, so if you don't mind, can you please review the recipe at your convenience as well?

recipes/carnac-lr/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+  run:
+    - python >=3
+
+test:
+  files:
```
bgruening: is this file now needed?

npavlovikj: The test file or the makefile patch?

bgruening: the test file...

npavlovikj: The test file is needed for the test in run_test.sh ("CARNAC-LR -f test.input -o output.txt
"). Should I not include test at all? The file is 3.2KB, so I thought having the test is a good practice?

npavlovikj:
@bgruening , I am sorry for pinging you here, but can you please let me know if you have any additional comments I need to address before this PR is approved for merging?

npavlovikj:
Thank you so much for the merge @dpryan79  - I really appreciate it!

PR #16666, opened 2019-08-01
Bioconductor tximeta Bump

apeltzer:

Needs this for the nf-core/scrnaseq workflow :-) And therfore removed it from blacklist :-)


apeltzer:
Hey @bioconda/core - I initially only wanted to get `bioconductor-tximeta` to be R3.6 compatible but had to (piece by piece) add in more recipes to get it built properly. 

Not sure it will suceed, but at the moment the build looks fine and okay and the recipes have been on the blacklist marked as "maybe too old and nobody needs them" (which isn't the case - at least I need them for the pipeline linked above ;-)).

apeltzer:
These are a hell of other R packages as dependencies...

dpryan79:
This is why I just do them all at once in bulk :) Anyway, it's in the queue, everything will get rebuilt in a week or so I expect.

apeltzer:
So I can basically close this then @dpryan79 ?

dpryan79:
Yeah, it's not needed. For the November bioconductor release I need to modify bulk and bioconda-utils to run multiple jobs in parallel. Then this won't take so painfully long.

PR #16764, opened 2019-08-07
Added SvABA recipe

micknudsen:

### DIFF
--- a/recipes/svaba/build.sh
+++ b/recipes/svaba/build.sh
@@ -1,6 +1,9 @@
 #!/bin/bash
 set -eu -o pipefail
 
-./configure --prefix=$PREFIX
+./configure
 make
 make install
+
+mkdir -p ${PREFIX}/bin
+cp bin/svaba ${PREFIX}/bin/
--- a/recipes/svaba/build.sh
+++ b/recipes/svaba/build.sh
@@ -2,7 +2,7 @@
 set -eu -o pipefail
 
 ./configure
-make
+make CC=${CC}
 make install
 
 mkdir -p ${PREFIX}/bin
--- a/recipes/svaba/build.sh
+++ b/recipes/svaba/build.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -eu -o pipefail
 
-./configure CC=${CC} CFLAGS="${CFLAGS}"
+./configure
 make CC=${CC} CFLAGS="${CFLAGS}"
 make install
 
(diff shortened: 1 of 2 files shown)


micknudsen:
@bioconda-bot rebuild

micknudsen:
I am getting an `ld: cannot find -lz` error which I suspect is due to submodules not being able to see `zlib`. Does anybody know if it possible to somehow "propagate" `zlib` to submodules?

npavlovikj:
1.
The error can be fixed if `-lz` is replaced with `-lm ${PREFIX}/lib/libz.a`. Maybe try adding:
```
export LIBS="-lm ${PREFIX}/lib/libz.a"
```
?

2.
Also, it is recommended practice to create separate conda packages for the needed submodules, instead of recursively installing them. I would suggest you to look into that if possible.

3.
Please make sure the order of sections in `meta.yaml` is: Package, Source, Build, Requirements, Test, About, Extra.

micknudsen:
@npavlovikj Thanks for your suggestion. I would really prefer to have the submodules as separate packages, but I am not sure whether that is possible. I suppose that the reason for having them as submodules here is that the source code is needed and not just the builds. For example, simply adding `htslib` as a dependency is not enough. Maybe I am missing something here.

micknudsen:
@npavlovikj Didn't manage to solve the problem using your first suggesting. Now going with the second (https://github.com/bioconda/bioconda-recipes/pull/16785).

micknudsen:
@bioconda-bot rebuild

micknudsen:
As this PR clearly shows, I have no experience with C++ at all. However, I am still puzzled why it doesn't just work. When I do this locally (on a Linux machine), it works right away:

```
git clone --recursive git@github.com:walaj/svaba.git
cd svaba
git checkout 1.1.0
./configure
make
make install
```

How is bioconda + CI any different?

dpryan79:
Without looking, I would guess that it's because the compiler you're using is named "gcc" and everything is in the default location on your system. Compiling things in that way will result in packages that won't work on systems where any of those default things is either a different version or place/name. I'll have a look and see if I can get this to compile.

micknudsen:
@dpryan79 Thanks! I am really eager to get a recipe for SvABA. It is an excellent tool, and I would like to use it in a production pipeline. So far all software versions are managed with conda, and I would do (almost) everything to avoid having to have a custom installation.

dpryan79:
Should work now, go ahead and merge if you're happy with it.

micknudsen:
Thanks! That is really amazing!

micknudsen:
Unfortunately, there is something not quite right about this recipe (https://github.com/walaj/svaba/issues/73).

PR #16844, opened 2019-08-12
Update watchdog-wms to 2.0.0

klugem:


recipes/watchdog-wms/build.sh
```diff
@@ -19,12 +19,16 @@ mv "${OUT}/workflowDesigner.sh" "${OUT}/watchdog-gui.sh"
 ln -s "${OUT}/watchdog-cmd.sh" "${PREFIX}/bin/watchdog-cmd"
 chmod 0755 "${PREFIX}/bin/watchdog-cmd"
 
+# get the modules
+/bin/bash "${OUT}/modules/downloadCommunityModules.sh" --auto
```
dpryan79: What's this downloading?

recipes/watchdog-wms/build.sh
```diff
@@ -19,12 +19,16 @@ mv "${OUT}/workflowDesigner.sh" "${OUT}/watchdog-gui.sh"
...
+
 # test, if javafx is working
 cd "${RECIPE_DIR}"
 RET=$(java "TestJavaFx")
 if [ "$RET" == "INSTALLED" ]; then
 	ln -s "${OUT}/watchdog-gui.sh" "${PREFIX}/bin/watchdog-gui"
 	chmod 0755 "${PREFIX}/bin/watchdog-gui"
+	echo "Found Javafx."
```
dpryan79: I won't pretend to be a java expert, but this is only telling you what was in the build environment. If openjdk is going to support Javafx going forward then that's fine, otherwise I think this would be better as a post-install script.

recipes/watchdog-wms/build.sh
klugem: Modules for the workflow management system that were previously shipped together with Watchdog but are now located at https://github.com/watchdog-wms/watchdog-wms-modules.

dpryan79: We really try not to vendor so many packages, why not use conda to install most of those?

recipes/watchdog-wms/build.sh
klugem: You're right. I will change that.

recipes/watchdog-wms/build.sh
klugem: In almost all cases the modules only contain specifications on how a binary could be used (XSD files). The dependend binaries must be installed on the system and are not part of that repository.
But probably it might be better to leave that out because the modules are then up-to date after installation.

recipes/watchdog-wms/build.sh
klugem: I removed that test completely becuase I have a javafx detection in watchdog-gui.sh anyway.

PR #16855, opened 2019-08-12
Add biopython-convert

innovate-invent:


innovate-invent:
@dpryan79 Maybe you can help. I am struggling with a error when running conda build myself:
```
Setuptools downloading is disabled in conda build. Be sure to add all dependencies in the meta.yaml
```

I can't find a clear solution for this, there is no way that I actually have to copy requirements.txt into meta.yml.

dpryan79:
@innovate-invent I'll have a look. Usually this happens when there are `install_requires` or such in `setup.py`. The normal methods of `python -m pip --no-deps` doesn't skip those then. I sometimes just cut those out of setup.py with `sed` or a patch and then everything works.

innovate-invent:
@dpryan79 Thanks. I use pbr and requirements.txt. I don't know how that affects things.

dpryan79:
@innovate-invent This should be ready to go.

innovate-invent:
Ah, so the answer is you do actually have to copy the requirements.txt into meta.yaml.
Is there a more robust way of doing this? I was hoping the requirements could be pulled from the pip package

dpryan79:
The conda skeleton will pull that from pypi, but it has to all be specified in the standard way.

PR #16868, opened 2019-08-13
New Recipe for pypgatk tool: ProteoGenomics tool. 

ypriverol:


ypriverol:
Ping @bgruening. The recipe I think is ready. Can you review it?

dpryan79:
Do all of the dependencies really have such exact versions?

ypriverol:
I use all the versions because I like to restrict to the versions that I 100% know it works. But if is not recommended I can remove the versions @dpryan79.

dpryan79:
You're best off supplying only minimal versions and assuming that everything going forward will always work. The main issue with exact version pinning is that one of those will get updated and what you're pinning against will never get rebuilt for new versions of compilers/R/python/etc., which will preclude your own package from being updated.

recipes/pypgatk/meta.yaml
```diff
@@ -0,0 +1,115 @@
...
+  sha256: "c249729dcd9c344f88469826a7d023cd4ada3236172832fa90a950208b8576da"
+
+build:
+  number: 0
+  noarch: python
+  script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed -vv "
+
+requirements:
+  host:
+    - argcomplete ==1.9.5
```
bgruening: You also most likely do only need python and pip here in the host section.

recipes/pypgatk/meta.yaml
```diff
@@ -0,0 +1,115 @@
...
+  commands:
+    - which pypgatk_cli.py
+    - pypgatk_cli.py -h
+
+
+about:
+  home: "http://github.com/bigbio/py-pgatk"
+  license: "Apache 2"
+  license_family: "APACHE"
+  license_file:
```
bgruening: add `LICENSE` here.

ypriverol: I added the link to the license but one of you tests complain about it because it was not in the same folder than meta.yaml. I will check for an example and add it.

recipes/pypgatk/meta.yaml
ypriverol: I generate this with the skeleton option of conda and did some refinements. I will remove other dependencies. Is that the same for requirements run?

dpryan79: Keep everything in the "run:" section since they're needed for the package to actually be used.

ypriverol: @dpryan79 changes added!!!! Thanks a lot for the feedback.

PR #16938, opened 2019-08-16
remove openslide

jcupitt:

I'm trying to add libvips to conda-forge and I'll like to depend on openslide. I can't do that if openslide is in bioconda.

I've added openslide to conda-forge here:

https://github.com/conda-forge/staged-recipes/pull/9169

This PR removes openslide from bioconda. Other bioconda recipes should be unaffected since they'll pick up the conda-forge version.

There have been a few small changes:

- add `licence_file`
- add a maintainer (me, but of course any bioconda people can add themselves)
- reorder sections
- reset build to 0
- load a tarball, not a zip from github, so we don't need `autotools`
- sort deps alphabetically
- remove perl as a host dep (since we don't need `autotools`)
- revise `build.sh`, it was very complex
- remove xorg and perl deps, we no longer need them
- mark as skip for win, since there's no `gdk-pixbuf` on Windows
- use sha356 rather than md5

Ping @bioconda/core -- please review and merge.


dpryan79:
Just wait until it's been merged on conda-forge before you merge here :)

jcupitt:
OK, I'll ping here when it's on conda-forge. Thanks!

jcupitt:
OK, openslide is now merged into conda-forge, and can be removed from here. 

Please wait 1 hour for all the updates to propagate and index.

Thank you for this!

dpryan79:
No need to wait, removing the recipe doesn't remove the package (it'll always be there).

bgruening:
Thanks @jcupitt!

PR #17048, opened 2019-08-23
Phast

RomainFeron:
Small update to Phast recipe: cleanup meta.yaml, updated build.sh to include all binaries from PHAST.


RomainFeron:
It was exactly that, I did not know this replace syntax. Probably not worth changing for this recipe at the moment but I will use it in the future, thanks !

PR #17086, opened 2019-08-25
Updated python/argparse requirements for mofapy

khughitt:
Updated mofapy meta.yaml to indicate that python >= 2.7 is required (see project [setup.py](https://github.com/bioFAM/MOFA/blob/52c430356ba42a1abc8ab1338efbcc931c9d83a6/setup.py#L23)) and that argparse is only needed for python >2.7 and <3.2. Currently, the unconditional argparse requirement results in python <= 3.6 being requested, leading to many conflicts and unneccessary downgrades.

### DIFF
--- a/recipes/mofapy/meta.yaml
+++ b/recipes/mofapy/meta.yaml
@@ -15,15 +15,15 @@ build:
 
 requirements:
   host:
-    - python
+    - python >=2.7
     - pip
 
   run:
-    - argparse
+    - argparse # [py>27 and py<32]
     - h5py
     - numpy
     - pandas
-    - python
+    - python >=2.7
     - scipy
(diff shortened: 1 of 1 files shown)


recipes/mofapy/meta.yaml
```diff
@@ -10,20 +10,20 @@ source:
...
 
 requirements:
   host:
-    - python
+    - python >=2.7
     - pip
 
   run:
-    - argparse
+    - argparse # [py>27 and py<32]
```
dpryan79: argparse has been part of python since 2.7. Since the odds of getting a python2 <2.7 or python3 <3.6 these days is near 0 I suspect this change isn't needed.

PR #17140, opened 2019-08-28
Smudgeplot

RomainFeron:
This pull request adds a recipe for [smudgeplot](https://github.com/KamilSJaron/smudgeplot), a tool for inference of ploidy and heterozygosity structure using whole genome sequencing data.


recipes/smudgeplot/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    sha256: 2445564d12a852dcd5ade521b522cabde0df23512687b4ab5b46f73f9b315608
+
+requirements:
+    build:
+        - {{ compiler('c') }}
+        - make
+        - r-devtools
+        - r-base
+        - r-viridis
+        - r-argparse
```
dpryan79: Please remove the R packages from the `build:` section.

recipes/smudgeplot/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    host:
+        - python >=3.6
+        - r-base
+        - r-viridis
+        - r-argparse
+        - numpy
+        - scipy
```
dpryan79: Are the python packages actually needed for installation or just during run time?

recipes/smudgeplot/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+test:
+    commands:
+        - smudgeplot.py --version 2>&1 | grep 'Running'
+
+about:
+    home: https://github.com/KamilSJaron/smudgeplot
+    license: Apache-2.0
```
dpryan79: I think the license is part of the tarball, so remove the license you manually added and add a `license_file: LICENSE` line.

recipes/smudgeplot/meta.yaml
dpryan79: OK, I guess I was surprised to see an R script used for installation rather than `R CMD install`.

RomainFeron: Indeed. I will transfer this comment to the author, I think they want to improve the install process anyway

PR #17166, opened 2019-08-29
Add VirSorter package

wclose:
This pull request is to add the VirSorter program to bioconda.


recipes/virsorter/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  sha256: {{ virsorterSha256 }}
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - perl-bioperl
```
bgruening: you probably want to have perl here as well

recipes/virsorter/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  run:
+    - blast
```
bgruening: please add perl as well

recipes/virsorter/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - diamond =0.9.14
```
bgruening: do you need the strict pinning?

wclose:
@bgruening I looked into it a bit more and it turns out the incompatible packages are all being installed due to using `perl-bioperl` so I don't think they are supposed to be mutually exclusive. They work fine in a normal environment but not in the build environment. Any ideas on how to fix it/avoid it?

recipes/virsorter/meta.yaml
wclose: done

recipes/virsorter/meta.yaml
wclose: done

recipes/virsorter/meta.yaml
wclose: The software developer has them pinned at these versions when using the software on other servers so I want to maintain the settings for comparability.

dpryan79:
@bioconda-bot merge

PR #17273, opened 2019-09-04
Update samtools & bcftools to depend on htslib package

jmarshall:
This PR changes samtools and bcftools to depend on the Bioconda htslib package and its shared libhts.so/.dylib instead of monolithically static-linking their own inferiorly-built copies of htslib.

This has a number of advantages:

* Samtools and bcftools use the htslib package's libhts, which is configured to use plugins, rather than their own libhts.a which they previously did not configure with plugins. This means that any additional HTSlib file access plugins the user has installed will now work with samtools and bcftools as well, not just with bgzip/htsfile/tabix.

* The only object in the htslib/samtools/bcftools universe that actually needs to be linked with libcrypto is htslib's _hfile_s3.so/.bundle_ plugin. In particular, with this PR _samtools_ is no longer itself linked with libcrypto and the samtools package does not directly depend on openssl. Hence the `samtools` command will work even if your conda installation has the previously somewhat common openssl version mismatch problem. (The only thing that would not work in that case would be accessing `s3://bucket/` URLs.)

* The samtools and bcftools packages no longer directly depend on the libdeflate package, so that is two fewer packages to sort out during the imminent libdeflate conda-forge migration and libdeflate build pinning migration.

as well as just plain being the right thing to do given that Bioconda produces a htslib package.

Also the samtools recipe has been updated to explicitly pull in ncurses's libtinfo as well as libncurses. This means that it will work fine with ncurses packages from either conda-forge or the default channels, eliminating another historical source of user frustration and bug reports.

---

### DIFF
--- a/recipes/bcftools/build.sh
+++ b/recipes/bcftools/build.sh
@@ -3,11 +3,7 @@
 # Overwrite plot-vcfstats with updated version
 curl https://raw.githubusercontent.com/samtools/bcftools/44c660490411cbc0611e405a8ebf3faa0d65c544/misc/plot-vcfstats -o misc/plot-vcfstats
 
-export CPPFLAGS="-DHAVE_LIBDEFLATE -I$PREFIX/include"
-export CFLAGS="-DHAVE_LIBDEFLATE -I$PREFIX/include"
-export LDFLAGS="-L$PREFIX/lib"
-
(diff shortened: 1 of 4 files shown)


jmarshall:
@bioconda/core: This is passing tests and appears to be ready to be merged.
(I do not wish to join Bioconda.)

dpryan79:
Thanks @jmarshall for the improvements!

PR #17384, opened 2019-09-11
reference package for SEPP

sjanssen2:
I hope this PR suits bioconda's policies. This recipe ships the most important reference file set for the program SEPP (for which bioconda already hosts a recipe). Since I foresee many updates of SEPP, but basically none for this reference set, I'd like to spilt those into two packages - without depending on each other, although this package shall place its files into the directory of SEPP (i.e. $PREFIX/share/sepp/ref/).

There are several user of SEPP asking for an alternative reference set for Silva, which I intend to build as well.


sjanssen2:
Many thanks to @dpryan79

Issue #17212, opened 2019-09-02
List of packages with broken pinnings that are currently causing issues


dpryan79:
Please feel free to add to this, as we're trying to assess how wide-spread the issues are to better determine if we need to start patching metadata.

 - [ ] htslib <1.9 lacked pinning to libdeflate
 - [ ] pysam < 0.15 lacked pinning to libdeflate

guillaumecharbonnier:
Pysam < 0.15 is looking for libcrypto.so.1.0.0 from openssl <1.1 whereas it currently gets libcrypto.so.1.1.1 without explicit openssl requirement.

dpryan79:
Another one where pysam pinnings are causing issues: https://github.com/bioconda/bioconda-recipes/pull/17260

dpryan79:
Another: https://github.com/bioconda/bioconda-recipes/pull/17247

dpryan79:
Here's another one: https://github.com/bioconda/bioconda-recipes/pull/11698

dpryan79:
ariba is getting broken because of this now too :(

rpetit3:
This should be fixed with Ariba and its dependents, now.

After adding the pysam pinnings to Bactopia and Seroba, the latest build of Ariba (bioconda/linux-64::ariba-2.14.3-py36he1b5a44_1) is selected and working.

bernt-matthias:
Maybe another one: https://github.com/bioconda/bioconda-recipes/issues/24264? Was there a solution for the libcrypto problem? I guess in my case its samtools itself and not pysam?

rpetit3:
Looks like checkm-genome is affected by pysam pinning. After bumping pysam version the issue is gone.

```
(base) rpetit@staphopia:~$ conda create -n checkm -c conda-forge -c bioconda checkm-genome
(base) rpetit@staphopia:~$ conda activate checkm
(checkm) rpetit@staphopia:~$ checkm data setRoot /local/home/rpetit/miniconda3/envs/checkm/checkm_data
Traceback (most recent call last):
  File "/local/home/rpetit/miniconda3/envs/checkm/bin/checkm", line 36, in <module>
    from checkm import main
  File "/local/home/rpetit/miniconda3/envs/checkm/lib/python3.6/site-packages/checkm/main.py", line 28, in <module>
    from checkm.resultsParser import ResultsParser
[...]

PR #17406, opened 2019-09-11
Updating dependency to htslib from libdeflate

jpuritz:
This should help with issues identified in #17212

### DIFF
--- a/recipes/bedtools/meta.yaml
+++ b/recipes/bedtools/meta.yaml
@@ -16,18 +16,15 @@ requirements:
   build:
     - {{ compiler('c') }}
     - {{ compiler('cxx') }}
+    - libdeflate
   host:
     - zlib
     - bzip2
-    - curl
     - xz
-    - htslib
   run:
     - zlib
     - bzip2
-    - curl
     - xz
-    - htslib
 
 test:
   commands:
--- a/recipes/mauve/2.4.0.r4736/meta.yaml
+++ b/recipes/mauve/2.4.0.r4736/meta.yaml
@@ -1,0 +1,41 @@
+{% set version = "2.4.0.r4736" %}
+{% set sha256 = "98aa1c7bfeef0dac68f7cedbee81dae424192c5cd7a22960d63efe6e943645aa" %}
+
+package:
+  name: mauve
+  version: {{ version }}
+
+build:
+  noarch: generic
+  number: 1
+
+source:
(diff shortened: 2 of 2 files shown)


dpryan79:
This was already pinned properly, so I don't think this will have any effect. The issue is with older builds sometimes being installed if someone uses `conda install` in an existing environment.

jpuritz:
OK.  I wasn't 100% sure this would fix my original issue, but this also made sense to me because (as I read it) `bedtools` depends on `htslib` and not `libdeflate` directly.

My original issue is that when I am creating a new OS X environment for `ddocent` (https://github.com/bioconda/bioconda-recipes/tree/master/recipes/ddocent), the pinning is not working properly and I get:

```
bioconda/osx-64::bedtools-2.26.0-0
bioconda/osx-64::htslib-1.8-h57f8fdf_2
bioconda/osx-64::samtools-1.4.1-0
```

Within the environment `conda update bedtools` gives me this:

```
The following packages will be UPDATED:

  bedtools                                         2.26.0-0 --> 2.29.0-hbf07bdc_0
  htslib                                     1.8-h57f8fdf_2 --> 1.9-h3a161e8_7
  samtools                                          1.4.1-0 --> 1.9-h8aa4d43_12

The following packages will be SUPERSEDED by a higher-priority channel:

  libdeflate         conda-forge::libdeflate-1.3-h01d97ff_0 --> bioconda::libdeflate-1.0-h1de35cc_1
```

I was following the recent updates to `samtools` that fixed a similar issue.  Though, it's entirely possible that this has nothing to do with these issues.

dpryan79:
Ah, yeah, this would be fixed by adding libdeflate in as a runtime dependency of ddocent. It's obviously not really a runtime dependency, but it'd get around the conda solver doing the wrong thing. Running `conda update` seems to be fixing the env, for what little that's worth :(

The original libdeflate dependency here is due to bedtools vendoring htslib, so thereby needing libdeflate to compile it. It actually doesn't need either of them during run time, though maybe that could be changed (it'd certainly make the binary smaller).

dpryan79:
curl could be removed too it seems. Mind changing your PR to remove those? That'd certainly clean the bedtools recipe up, which would be nice :) Also, libdeflate could be moved to the build section, since if it's used it's statically compiled in.

jpuritz:
I think that will work.  Thanks for your help.

dpryan79:
Great, that certainly cleans the recipe up :)

dpryan79:
@bioconda-bot merge

PR #17420, opened 2019-09-12
SEPP: Silva 12.8 reference package

sjanssen2:
Similar to #17384, this PR provides the second most popular reference Silva 12.8 for SEPP.


recipes/sepp-refsilva128/meta.yaml
```diff
@@ -0,0 +1,30 @@
+{% set version = "v4.3.6" %}
```
dpryan79: Can you change the version to 4.3.6 and just add the "v" in the URL? The v at the beginning can much of the symantic versioning.

sjanssen2: good point.

PR #17444, opened 2019-09-13
PacBio: Remove `isoseq3` dependencies

armintoepfer:

### DIFF
--- a/recipes/isoseq3/meta.yaml
+++ b/recipes/isoseq3/meta.yaml
@@ -10,6 +10,14 @@ source:
   url: https://github.com/PacificBiosciences/IsoSeq3/releases/download/v{{ version }}/isoseq3
   sha256: {{ sha256 }}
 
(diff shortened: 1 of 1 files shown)


dpryan79:
@armintoepfer Was the previous build not working on OSX?

armintoepfer:
Right. We only build static binaries for linux, so far. As mentioned earlier on another PR, at some point recipes were changes such that only skipping linux did not work and we were told to use `noarch: generic` to get them merged. I'm slowly moving recipes back to the state where we only allow installation from linux.

dpryan79:
Ah, gotcha, makes sense then.

dpryan79:
@bioconda-bot merge

armintoepfer:
@dpryan79 Any idea why a fresh `conda install isoseq3` still tries to install all those dependencies that I deleted in this PR?

PR #17457, opened 2019-09-13
Build kyototycoon with fewer side effects

nathanweeks:
Removing "-ansi" from compiler options also works around the noted compiler error involving stdlib.h, (and the previous workaround of copying the contents of a glibc's include directory to ${PREFIX}/include would have unintended side effects).

Setting / unsetting LD_LIBRARY_PATH in conda activate/deactivate scripts has unintended side effects (and is unexpected; no other bioconda recipe appears to do this); remove for safety.

* [ x] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/contributor/guidelines.html).

### DIFF
--- a/recipes/kyototycoon/build.sh
+++ b/recipes/kyototycoon/build.sh
@@ -1,43 +1,14 @@
 #!/bin/bash
 
 
-# using default conda libraries for gcc causes the following error:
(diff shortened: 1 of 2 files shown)


dpryan79:
@bioconda-bot merge

PR #17563, opened 2019-09-20
DFAST recipe updated

nigyta:

### DIFF
--- a/recipes/dfast/build.sh
+++ b/recipes/dfast/build.sh
@@ -9,7 +9,6 @@ rm bin/*/blastdbcmd
 rm bin/*/blastp
 rm bin/*/ghost*
 rm bin/*/hmm*
-rm -r bin/Darwin/lib
 rm bin/*/makeblastdb
 rm bin/*/mga
 rm bin/*/rpsblast
@@ -17,6 +16,7 @@ if [ "$(uname)" == "Darwin" ]; then
     rm -rf bin/Linux
 else
     rm -rf bin/Darwin
+    LD_LIBRARY_PATH=APPROOT/bin/Darwin/lib
 fi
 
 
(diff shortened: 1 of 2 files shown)


dpryan79:
@nigyta Is it possible to get the source code for the modified last binaries? This recipe would be more portable if we could just compile them.

nigyta:
Dear @dpryan79,

I really appreciate all of your help.
I am sorry, but I am out of my lab for several days, it may take time until I will prepare the files.
I will ask for your help when the files are ready.

I would like to include 2 custom binaries (last and rpsbproc).
rpsbproc is a utility program to parse the result of RPSBlast, but it is not included in the standard Blast package.

I am thinking about two more options.
One is to download required binaries using the "post-link.sh" script after building the software package. Is that possible?
The other option is to give up MacOS package. I confirmed that Linux testing was successful although testing for MacOS failed.

dpryan79:
I don't think we could permit downloading the binaries afterward, since we couldn't ensure they'd work on all systems. Can you make the source code available? We could just compile them then. Otherwise yes, we should just skip OSX.

nigyta:
@dpryan79 Thank you for the quick response.
I see. I'm going to prepare the source codes.

One more question.
Is it possible to download souce codes and modify them to create customized binaries during the build step?

dpryan79:
Yes, generally we just ship patch files for the source code.

nigyta:
@dpryan79 

Could you review and merge this?
After setting binary_relocation option to False, build testing for MacOS succeeded.

* As you recommended, I tried to compile the binary from the source code, but I couldn't do that due to some unexpected errors during compilation.

Thank you again for all your help.

dpryan79:
Thanks  @nigyta

PR #17570, opened 2019-09-20
Recompile pysam using a newer libcrypto.so

rhpvorderman:
Current build uses libcrypto.so.1.0.0 which is part of a deprecated openssl version.
This would give the error:
    import pysam
     ../../miniconda3/envs/cromwell/lib/python3.6/site-packages/pysam/__init__.py:5: in <module>
         from pysam.libchtslib import *
    E   ImportError: libcrypto.so.1.0.0: cannot open shared object file: No such file or directory

Current build uses libcrypto.so.1.0.0 which is part of a deprecated openssh version.
This would give the error:
```
import pysam
../../miniconda3/envs/cromwell/lib/python3.6/site-packages/pysam/__init__.py:5: in <module>
    from pysam.libchtslib import *
E   ImportError: libcrypto.so.1.0.0: cannot open shared object file: No such file or directory
```

### DIFF
--- a/recipes/pysam/meta.yaml
+++ b/recipes/pysam/meta.yaml
@@ -9,7 +9,7 @@ source:
   sha256: f553d48d0345404b6b103d0b82bad09c8d78420e1cc6bef33040553fc579e284
 
 build:
-  number: 1
+  number: 2
   binary_relocation: False # [linux]
 
 requirements:


dpryan79:
This is built against the same openssl as the previous build (1.1.1c).

rhpvorderman:
@dpryan79 Whoops sorry, my bad. It turns out that I could not resolve an environment with both python 3.7 and pysam due to strict channel priority being active by default. Having disabled that it works perfectly.

Sorry for the inconvenience.

PR #17722, opened 2019-09-29
Adding dependency drmaa to digestiflow-demux recipe.

holtgrewe:

### DIFF
--- a/recipes/digestiflow-demux/meta.yaml
+++ b/recipes/digestiflow-demux/meta.yaml
@@ -6,7 +6,7 @@ package:
   version: "{{ version }}"
 
 build:
-  number: 0
+  number: 1
   noarch: python
   script: python -m pip install --no-deps --ignore-installed .
   entry_points:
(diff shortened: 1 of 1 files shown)


bgruening:
@holtgrewe drmaa is hardly a dependency of this tool, but rather from snakemake isn't it?

holtgrewe:
@bgruening well, Snakemake uses it in a `try import` fashion and does not depend on this. @johanneskoester is this lack of dependency an oversight or intentional?

johanneskoester:
Good question. We could relatively savely add it to the snakemake package, since it is only the python frontend, right? On the other hand, I so far had the policy to not add execution backends as dependencies, but rather leave that to the user. Otherwise, we would also need to add kubernetes, tibanna, ... which would introduce quite a lot of other dependencies that are not needed for 90% of the users. I think whether those things, or drmaa is needed, depends on the place where the user wants to execute the workflow. Consequently, wouldn't it make sense to leave it out of this recipe as well?

Issue #6107, opened 2017-09-22
name collision: jellyfish -> conda-forge/jellyfish-feedstock


croth1:
bioconda's jellyfish recipe has a name collision with conda-forge/jellyfish-feedstock.

bgruening:
:( but expectable ... can we rename the bioconda one and update all dependent packages?

bgruening:
ping @croth1

croth1:
Hmh, back then it was just a random observation - I'd need help to solve this:
- a suggestion of a new name
- help on how renaming works - is it just submitting a new recipe + one of the admins moving the old to the deprecated channel?

epruesse:
I had to rename `arb` to `arb-bio` so that it would not conflict with `libarb`. Annoying because `arb` does get called `arb` name in Debian/Ubuntu...

bgruening:
Ok, I put up a rename PR. Please comment on this.

PR #17748, opened 2019-09-30
Jellyfish renaming - name collision with conda-forge

bgruening:
This PR will rename our jellyfish recipe to kmer-jellyfish to avoid a name collision.

Those name collisions are way more unlikely now as we lint on the conda-forge site and on the bioconda side to avoid this. But we need to deal with this now. So here is my proposal to fix https://github.com/bioconda/bioconda-recipes/issues/6107


dpryan79:
We should actually rename jellyfish first, since otherwise it won't usually be around for dependency resolution.

dpryan79:
This can be run again in about 30 minutes.

bgruening:
Oh cool! ... thanks  ... sorry last week was a mess here.

PR #17799, opened 2019-10-02
HMF tools: add LINX

vladsaveliev:

### DIFF
--- a/recipes/hmftools-linx/build.sh
+++ b/recipes/hmftools-linx/build.sh
@@ -7,6 +7,6 @@ TGT="$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM"
 cd "${SRC_DIR}"
 mv sv-linx*.jar $TGT/sv-linx.jar
 
-cp $RECIPE_DIR/LINX.sh $TGT/LINX
-ln -s $TGT/LINX $PREFIX/bin
-chmod 0755 "${PREFIX}/bin/LINX"
+cp $RECIPE_DIR/linx.sh $TGT/linx
+ln -s $TGT/linx $PREFIX/bin
+chmod 0755 "${PREFIX}/bin/linx"
--- a/recipes/hmftools-linx/build.sh
+++ b/recipes/hmftools-linx/build.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+TGT="$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM"
+[ -d "$TGT" ] || mkdir -p "$TGT"
+[ -d "${PREFIX}/bin" ] || mkdir -p "${PREFIX}/bin"
+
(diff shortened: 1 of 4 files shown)


recipes/hmftools-linx/build.sh
```diff
@@ -0,0 +1,12 @@
...
+TGT="$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM"
+[ -d "$TGT" ] || mkdir -p "$TGT"
+[ -d "${PREFIX}/bin" ] || mkdir -p "${PREFIX}/bin"
+
+cd "${SRC_DIR}"
+mv sv-linx*.jar $TGT/sv-linx.jar
+
+cp $RECIPE_DIR/LINX.sh $TGT/LINX
+ln -s $TGT/LINX $PREFIX/bin
+chmod 0755 "${PREFIX}/bin/LINX"
```
bgruening: can we call it `linx` is capital writing needed?

recipes/hmftools-linx/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - xorg-libxtst
+    - r-base ==3.6.*
+    - r-tidyr
+    - r-dplyr
+    - r-ggplot2
+    - r-cowplot
+    - r-magick
+    - bioconductor-gviz
+    - circos>=0.69.6                   # for circos
+    - libgd                            # for circos
```
bgruening: cirocos should bring this in on there own, you don't need to specify it here I think

recipes/hmftools-linx/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  url: https://github.com/hartwigmedical/hmftools/releases/download/sv-linx-v{{ version }}/sv-linx_{{ version }}.jar
+  sha256: '{{ sha256 }}'
+
+build:
+  noarch: generic
+  number: 0
+
+requirements:
+  run:
+    - zlib
```
bgruening: Can we remove zlib here? Why is this needed?

recipes/hmftools-linx/build.sh
vladsaveliev: Not really, just following the style of other Hartwig tools ([purple](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/hmftools-purple/meta.yaml), amber, cobalt). @chapmanb, wondering if there was a specific reason to capitalize? Happy to help changing it for all tools if it makes sense.

vladsaveliev:
Thanks so much @bgruening for the review! Will address asap.

dpryan79:
Please remove the jar file, it will be downloaded when the package is built and doesn't need to be in the recipe itself.

vladsaveliev:
Sorry, added the jar accidentally; downloaded it into cwd when checking the sha256 sum and forgot to clean up. Should be good now.

PR #17923, opened 2019-10-05
Add `fastobo` recipe

althonos:
This PR adds a bioconda recipe for the [`fastobo`](https://pypi.org/project/fastobo) Python package, a library to parse Open Biomedical Ontologies.

### DIFF
--- a/recipes/fastobo/build.sh
+++ b/recipes/fastobo/build.sh
@@ -1,0 +1,26 @@
+#!/bin/bash -e
+# build script taken from `rust-bio-tools/build.sh`
+
+# taken from yacrd recipe, see: https://github.com/bioconda/bioconda-recipes/blob/2b02c3db6400499d910bc5f297d23cb20c9db4f8/recipes/yacrd/build.sh
+if [ "$(uname)" == "Darwin" ]; then
+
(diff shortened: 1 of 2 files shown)


althonos:
@bioconda/python: Hi! I am trying to compile a Python extension written in Rust but the Rust nightly compiler found on Anaconda is outdated. However, I already provide pre-built wheels on PyPI, is it possible to build the conda package from them instead of the source distribution?

althonos:
Other question to @bioconda/rust : is it possible to use the latest nightly compiler to build a package?

althonos:
Thanks for the heads-up @dpryan79

dpryan79:
Would it help if the rust package were updated to 1.40.0? The issue with nightly versions of rust or other packages is that they tend to randomly break things.

althonos:
@dpryan79 : I mean I'm fine downloading the latest nightly to build against it, the issue now seems to be with `libiconv` in MacOSX. Since the stack is almost stable (`pyo3` uses only one nightly feature) I don't think a lot of things could break between the moment I push a release and the moment it is build for `conda`.

dpryan79:
Ah, gotcha. For the moment I'd suggest just skipping OSX.

dpryan79:
Thanks @althonos!

PR #17994, opened 2019-10-09
anvio-minimal v6

ozcan:

### DIFF
--- a/recipes/anvio-minimal/meta.yaml
+++ b/recipes/anvio-minimal/meta.yaml
@@ -17,10 +17,10 @@ build:
 
 requirements:
   host:
-    - python 3.6
+    - python 3.7
     - pip
   run:
-    - python 3.6
+    - python 3.7
     - numpy
     - bottle ==0.12.13
     - pysam ==0.15.2
--- a/recipes/anvio-minimal/meta.yaml
+++ b/recipes/anvio-minimal/meta.yaml
@@ -17,10 +17,11 @@ build:
 
 requirements:
   host:
-    - python 3.7
+    #This pins to an old snakeMake that predated 3.7
+    - python 3.6
     - pip
   run:
-    - python 3.7
+    - python 3.6
     - numpy
     - bottle ==0.12.13
     - pysam ==0.15.2
--- a/recipes/anvio-minimal/meta.yaml
+++ b/recipes/anvio-minimal/meta.yaml
@@ -12,19 +12,16 @@ source:
 
 build:
   number: 0
-  skip: True  # [not py36]
-  script: python -m pip install --no-deps --ignore-installed .
+  noarch: python
+  script: python -m pip install --no-deps --ignore-installed -vv .
 
 requirements:
   host:
-    - python
-    - numpy
-    - setuptools ==41.2.0
-    - wheel ==0.33.6
-    - pip ==19.2.3
+    - python 3.6
+    - pip
   run:
-    - python
-    - {{ pin_compatible('numpy') }}
+    - python 3.6
+    - numpy
     - bottle ==0.12.13
     - pysam ==0.15.2
     - ete3 ==3.1.1
@@ -57,7 +54,3 @@ about:
   license_file: LICENSE.txt
   summary: "An interactive analysis and visualization platform for omics data"
   dev_url:  https://github.com/merenlab/anvio
-
-extra:
-  skip-lints:
-    - should_be_noarch_python
--- a/recipes/anvio-minimal/meta.yaml
+++ b/recipes/anvio-minimal/meta.yaml
@@ -56,3 +56,7 @@ about:
   license_file: LICENSE.txt
   summary: "An interactive analysis and visualization platform for omics data"
   dev_url:  https://github.com/merenlab/anvio
+
+extra:
+  skip-lints:
+    - should_be_noarch_python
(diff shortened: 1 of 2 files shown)


recipes/anvio-minimal/meta.yaml
```diff
@@ -15,40 +15,34 @@ build:
...
 requirements:
-  build:
-    - {{ compiler('c') }}
   host:
-    - python >=3,<3.7
-    - pip
-    - numpy
-    - cython
-    - gsl
+    - python ==3.6
```
bgruening: @ozcan all these strict pins are not recommended. Can you loosen them? Python 3.7 is probably also supported isn't it?

recipes/anvio-minimal/meta.yaml
```diff
@@ -15,40 +15,34 @@ build:
...
+    - numpy ==1.17.1
```
bgruening: I would also loosen the numpy version.

recipes/anvio-minimal/meta.yaml
```diff
@@ -15,40 +15,34 @@ build:
...
+    - setuptools ==41.2.0
+    - wheel ==0.33.6
```
bgruening: why do you need wheel and why in this exact version?

recipes/anvio-minimal/meta.yaml
```diff
@@ -59,5 +53,10 @@ about:
   home: http://merenlab.org/software/anvio/index.html
   license: GPL-3.0-or-later
   license_family: GPL3
+  license_file: LICENSE.txt
   summary: "An interactive analysis and visualization platform for omics data"
   dev_url:  https://github.com/merenlab/anvio
+
+extra:
+  skip-lints:
+    - should_be_noarch_python
```
bgruening: If we don't have a compiler, it seems you have a noarch version. Is that not the case?

recipes/anvio-minimal/meta.yaml
```diff
@@ -12,43 +12,38 @@ source:
...
+    - setuptools ==41.2.0
+    - wheel ==0.33.6
+    - pip ==19.2.3
   run:
-    - python >=3,<3.7
-    - numpy
+    - python
+    - {{ pin_compatible('numpy') }}
     - bottle ==0.12.13
     - pysam ==0.15.2
```
dpryan79: Do the version pinnings really have to be so exact?

recipes/anvio-minimal/meta.yaml
```diff
@@ -12,43 +12,38 @@ source:
...
+  skip: True  # [not py36]
   script: python -m pip install --no-deps --ignore-installed .
 
 requirements:
-  build:
-    - {{ compiler('c') }}
   host:
-    - python >=3,<3.7
-    - pip
+    - python
```
dpryan79: Does this still not work with python 3.7?

recipes/anvio-minimal/meta.yaml
ozcan: Many of our dependencies have a similar subset of dependencies like `scipy`, `numpy`, `pandas`, which causes some long conflict chains, especially on `Python 3.7`. We found suggesting users create a new environment with frozen dependencies on `Python 3.6` works best under these circumstances, for now. We are planning to loosen dependencies for the next release and hoping to target `Python 3.7` too. We still have limited experience in distributing on Conda and any suggestions and tips are highly appreciated.

recipes/anvio-minimal/meta.yaml
ozcan: Hi thank you for your review, I tried to answer version pinning here:
https://github.com/bioconda/bioconda-recipes/pull/17994#discussion_r333356134

recipes/anvio-minimal/meta.yaml
ozcan: The version number does not have to be exact but we pinned it as well with others. I don't know exactly what depends on it but I remember seeing `wheel` many times in `conda-build` conflict logs.

recipes/anvio-minimal/meta.yaml
ozcan: Unfortunately not, I tried with the newest versions of our dependencies, some does not support Python 3.7. Expectedly some older versions also do not work on 3.7. So we had to manually change versions to find a working combination. We pinned the first combination we got it to work.

bgruening: Do you recall which of you dependencies do not support 3.7. I guess the package is just missing for 3.7 and we can fix this package instead.

dpryan79: I'll tweak the recipe and have it test for the python 3.7 issue.

dpryan79:
Oh, it's the strict snakeMake pinning that's a problem. That old version predated python 3.7.

ozcan:
Thank you for looking into this @dpryan79. It seems we directly import and use `snakemake` here: https://github.com/merenlab/anvio/blob/master/anvio/workflows/__init__.py#L164
I am kindly pinging @ShaiberAlon here to get his feedback on this subject.

ShaiberAlon:
Is the issue here the fact that we require an older version of snakemake?

We indeed use the snakemake API, as shown in the code line @ozcan cited.

We could test our workflows with the latest version of snakemake and if all is well then we can require the latest version.

Please let me know if I understood the issue correctly.

dpryan79:
@ShaiberAlon Yes, that's correct.

meren:
Thank you very much for your time and help with this, @dpryan79!

dpryan79:
No problem, let me know if you end up needing help loosening up version restrictions for the next release.

meren:
Next time we will not leave everything to the last minute and be very happy to have your input. Thank you very much.

PR #18090, opened 2019-10-15
RSEM Bump for R3.6.1

apeltzer:

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -39,7 +39,6 @@ recipes/blast/2.5.0
 recipes/quast
 recipes/tadarida-c
 recipes/w4mclstrpeakpics
-recipes/rsem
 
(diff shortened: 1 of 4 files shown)


dpryan79:
I don't think this actually needs to be pinned against any R version, since it appears to be a purely run-time dependency.

apeltzer:
Yeah I just tried installing it alongside some R packages and got an incompatibility notice so thought to get this off the blacklist and make it work with R3.6.1 would be good :+1:

apeltzer:
`ValueError: License file given in about/license_file (/opt/recipe/LICENSE) does not exist in source root dir or in recipe root dir (with meta.yaml)`

dpryan79:
We should make sure that the R scripts are picked up as having a prefix in them before merging this.

dpryan79:
They're listed in the `has_prefix` file, so this should work and never again need to be rebuilt for R :)

PR #18108, opened 2019-10-16
Put stable versions for all dependent packages.

lucventurini:

Currently, installation of maker through bioconda fails due to conflicts in the environment resolution.
To solve this, I have modified the recipe so to pin all versions of packages to those I got by creating the environment with all dependent packages manually.

This should hopefully fix the installation issue.

### DIFF
--- a/recipes/maker/meta.yaml
+++ b/recipes/maker/meta.yaml
@@ -13,7 +13,7 @@ source:
 
 build:
   noarch: generic
-  number: 15
+  number: 16
 
 requirements:
   host:
--- a/recipes/maker/meta.yaml
+++ b/recipes/maker/meta.yaml
@@ -17,58 +17,58 @@ build:
 
 requirements:
   host:
-    - augustus >=3.2.3
+    - augustus ==3.2.3=pl526hcfae127_5
     - blast {{ blast_version }}
-    - exonerate
-    - infernal
-    - mir-prefer
-    - repeatmasker
-    - snap
-    - snoscan
-    - perl
-    - perl-bioperl-core >=1.007002
-    - perl-bit-vector
-    - perl-dbd-pg
-    - perl-dbd-sqlite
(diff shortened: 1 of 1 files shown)


abretaud:
What exactly is your problem when trying to install maker?
Pinning everything with exact versions and hashes like this would bring other problems (when installing maker with other packages for example)

lucventurini:
The problem I am encountering is one of unsatisfiable dependencies. I attach the log from a failed installation.

The problem disappears when creating the environment manually (using the packages contained within the `meta.yaml`), activate it, then install maker.

dpryan79:
`conda create -n maker -d maker` doesn't report any issues, so it's likely that whatever you have in that environment isn't directly compatible according to conda. This is one of the regions we suggest people make new environments rather than installing packages into old ones.

lucventurini:
> `conda create -n maker -d maker` doesn't report any issues, so it's likely that whatever you have in that environment isn't directly compatible according to conda. This is one of the regions we suggest people make new environments rather than installing packages into old ones.

Mmm, that is strange. The log from before comes from the following command:

`conda create -p /software/stable/maker/2.31.10/x86_64/ -c bioconda -- maker`

I have updated Conda to the latest version, and I also had deactivated all environments. **The errors from above still persist**.

dpryan79:
Please use `-c conda-forge -c bioconda`.

lucventurini:
Hi @dpryan79 , indeed this functions. Many thanks for pointing to the correct solution.

At this point, I agree with you that probably pinning the package versions is not the right solution.
However, the current state of affairs is not good either - users would not necessarily know to add conda-forge to channels (or, like me, would expect the conda package to bring in those dependencies automatically).

Is there anything that can be done in this sense?

dpryan79:
Conda-forge is pretty much always required by bioconda packages, there will never be a different solution for this.

PR #18137, opened 2019-10-17
GenMap: fixed recipe

cpockrandt:
Recipe was previous built with `-march=native` which lets genmap crash when executed on older processors.


dpryan79:
Good catch!

PR #18158, opened 2019-10-18
Topic/cutadapt pigz

bernt-matthias:
cutadapt: remove pigz run requirement

currently cutadapt calls pigz (via xopen) without limiting the number of used CPUs (https://github.com/marcelm/cutadapt/issues/290). hence pigz uses the default = all CPUs which is not acceptable on shared compute ressources.

if pigz is not installed single threaded python libraries are used.

alternatively (would be better) a new xopen version should be used once it is available https://github.com/marcelm/xopen/pull/20

### DIFF
--- a/recipes/cutadapt/meta.yaml
+++ b/recipes/cutadapt/meta.yaml
@@ -9,7 +9,7 @@ source:
   sha256: ced79e49b93e922e579d0bb9d21298dcb2d7b7b1ea721feed484277e08b1660b
 
 build:
-  number: 0
+  number: 1
   script: "{{ PYTHON }} -m pip install . --no-deps -vv"
   skip: True  # [py27]
 
--- a/recipes/cutadapt/meta.yaml
+++ b/recipes/cutadapt/meta.yaml
@@ -27,7 +27,8 @@ requirements:
     - python
     - xopen >=0.5.0
     - dnaio >=0.3
-    - pigz
+    # remove pigz until https://github.com/marcelm/xopen/pull/20 has found its way into cutadapt
+    #- pigz
 
 test:
   imports:


dpryan79:
Wow, yeah, that's definitely a problem. Are you hitting this on Galaxy or something where you can't easily just remove pigz from the conda env? If so, I'm happy to just merge this to get things working properly until xopen is tweaked.

bernt-matthias:
Jep. Galaxy. Will remove pigz from all conda envs I guess.

Seems that new xopen and cutadapt releases are in sight. So let's wait with merging

dpryan79:
OK, sounds like a game plan then. I'll reject my PR approval so this doesn't accidentally get merged then. Just ping me if it looks like the changes elsewhere are taking to long and we can just merge this in.

dpryan79:
It looks like there's a new xopen out, I'll update the cutadapt recipe to require it.

rhpvorderman:
This is really a problem of cutadapt and not of pigz.
`pigz` is by far a better implementation of the deflate compression algorithm than `gzip`. It compresses and decompresses faster even when limited to only one thread. 

Delegating the decompression and compression throught a separate pigz process creates more room for the python interpreter to do its actual job (cutadapt) instead of using the python gzip library to compress and decompress threads. This causes a tremendous speedup.

As far as the amount of threads go, that should be solved in cutadapt. But removing pigz from the cutadapt recipe is in my opinion  shooting ourselves in the foot. This will mean that pigz is removed from the biocontainer and that the fastest deflate implementation will simply not be available for users of that container.

dpryan79:
@rhpvorderman I just noticed that you commented on the xopen issue. Am I correct in understanding that updating xopen will not in fact resolve the issue that @bernt-matthias is running into?

rhpvorderman:
Nope. I am working on a fix in cutadapt as we speak.

bernt-matthias:
@dpryan79 @rhpvorderman you are 100% right. its an xopen + cutadapt issue. 

the xopen issue is now already solved: https://github.com/marcelm/xopen/pull/20 and a new release available

unfortunately cutadapt does not pass threads argument to xopen. @rhpvorderman my PR is nearly done ..

bernt-matthias:
https://github.com/marcelm/cutadapt/pull/411

dpryan79:
Thanks to both of you, I'll close #18171 then.

rhpvorderman:
I feel strongly about this. Consider the following benchmark: unzipping a fastq file that has been compressed with gzip. Compression level = 1. Resulting file size: 2.3 gb. 

We are going to decompress this file. We do 3 warmup runs to make sure the file is cached (for fair comparison) and then 10 runs to take the average.

command | wall clock time (average) | total cpu time (average) | comment
---|---|---|---
`pigz -cd very_big.fastq.gz > /dev/null` | 24.4 seconds | 32.7 seconds | Uses more seconds total cpu time because threads cause overhead.
`pigz -cd -p 1 very_big.fastq.gz > /dev/null` | 27.4 seconds | 27.4 seconds | Forces pigz to use ony one thread. As you can see, the cpu overhead for more threads is quite substantial.
`gzip -cd very_big.fastq.gz > /dev/null` | 44.2 seconds| 44.2 seconds | gzip sucks!

dpryan79:
> I feel strongly about this.

I agree. Since changes to cutadapt seem to be moving along quickly I think we can wait for them.

bernt-matthias:
Sure. Lets wait.

My temporary solution for out cutadapt conda envs will be to `conda remove --force pigz`.

Still load might be > 1 since independent threads/processes are used for reading, processing and writing. I guess for the Galaxy wrapper we will need some kind of formula to compute the number of additional threads, i.e. the value of `--cores ` given the number of available cores. I guess for using gzip 

I guess --cores should be `no_available_cpus - no_input_files + no_output_files` (which also means that one needs to request at least 5 cores for processing paired end data).

I think this should also work for the default values suggested in https://github.com/marcelm/cutadapt/pull/412

dpryan79:
Galaxy has supported compressed fastq files natively for a few years now :)

rhpvorderman:
Sorry, my bad! In that case the `-Z` flag will force cutadapt to use a compression level of 1. It will use much less CPU power. Since these files are intermediate anyway and not stored for long periods of time there is no utility for compressing them to very small sizes.

marcelm:
I’m closing this stale PR now. [Since version 2.7, Cutadapt no longer uses pigz for writing gzip files when it’s running single-threaded](https://cutadapt.readthedocs.io/en/latest/changes.html#v2-7-2019-11-22).

Apart from that, the `pigz` dependency should actually be removed from the cutadapt recipe because the `xopen` library already depends on it. I’ll do that when releasing the next version. Please report further problems at marcelm/cutadapt#290 .

PR #18345, opened 2019-10-29
bioconda-utils: test requirements satisfaction

mbargull:

@dpryan79: Haven't tested this, but maybe this works as a regression test.
PS: This will create a merge conflict on `bulk`. I cherry-picked parts of your commit from `bulk`.

### DIFF
--- a/recipes/bioconda-utils/meta.yaml
+++ b/recipes/bioconda-utils/meta.yaml
@@ -78,7 +78,7 @@ test:
       --file="$( python -c
       'from os.path import dirname, join; import bioconda_utils; print(join(dirname(bioconda_utils.__file__), "bioconda_utils-requirements.txt"))'
       )"
-      --json | tee /dev/stderr | grep -qF 'All requested packages already installed.'
+      2>&1 | tee /dev/stderr | grep -qF 'All requested packages already installed.'
 
 about:
   home: http://bioconda.github.io/build-system.html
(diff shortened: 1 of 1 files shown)


dpryan79:
This is definitely worth a try

mbargull:
The test failed! Which is good -- means it might work ;). Let's give the CI another go.

recipes/bioconda-utils/meta.yaml
```diff
@@ -18,10 +18,10 @@ build:
 
 requirements:
   host:
-    - python >=3.6
+    - python >=3.7
```
dpryan79: You can ignore this requirement in practice, I think it's a hold-over from some earlier testing.

recipes/bioconda-utils/meta.yaml
```diff
@@ -78,7 +78,7 @@ test:
       --file="$( python -c
       'from os.path import dirname, join; import bioconda_utils; print(join(dirname(bioconda_utils.__file__), "bioconda_utils-requirements.txt"))'
       )"
-      --json | grep -qF 'All requested packages already installed.'
+      --json | tee /dev/stderr | grep -qF 'All requested packages already installed.'
```
mbargull: IDK whether `tee /dev/stderr` works on macOS or not. We'll see what the CI says..

recipes/bioconda-utils/meta.yaml
mbargull: Possibly, but it doesn't hurt to go for 3.7.
For the test we definitely need to have the requirements in `meta.yaml` and `bioconda_utils-requirements.txt` in agreement, though.

PR #18350, opened 2019-10-29
Initial HapCUT2 recipe

marcelm:


recipes/hapcut2/meta.yaml
```diff
@@ -0,0 +1,42 @@
+package:
+  name: hapcut2
+  version: '1.1'
```
jenzopr: Maybe add package name and version as variable in the reipe:
```
{% set name = "hapcut" %}
{% set version = "1.1" %}

package:
  name: {{ name|lower }}
  version: {{ version }}
```

recipes/hapcut2/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+source:
+  - url: https://github.com/vibansal/HapCUT2/archive/v1.1.tar.gz
```
jenzopr: Also include the version variable here

recipes/hapcut2/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    sha256: f9fb1838b4c3f7fa4dc7bd6de72341f950caa3b2cf98b40a73927734dc75e26e
+    patches:
+    - shebang.patch
+    - flags.patch
+    - segfault.patch  # Patch from upstream master that’s not in the 1.1 release
+  - url: https://github.com/samtools/htslib/releases/download/1.2.1/htslib-1.2.1.tar.bz2
```
jenzopr: Does it really require `htslib` and `samtools` as source in the `submodules`? Or would it be enough to add them in requirements? Both are in bioconda....

jenzopr: I see.. it needs the includes of those two. Maybe its still enough to have them as dependency? What do you think @bioconda/review ?

marcelm: The Makefile wants to statically link against `libbam.a`, which is part of samtools. Version 1.2 of the samtools Bioconda package does not ship a `libbam.a`. I can find a `libbam.a` only in the ancient 0.1.19 version, but HapCUT2 wants version 1.2. None of the samtools recipes ship shared libraries, so using those is also not possible.

I don’t even know whether the shared or static libraries, such as `libbam.a`, that samtools creates when you build it are meant to be used from other projects (I thought the libraries were now in htslib), so I didn’t try to update the samtools recipe.

HapCUT2 also links against htslib (statically). I may be able to change it such that the Bioconda htslib is used instead, but then the recipe isn’t as "pure" as we want it anyway, so I decided not to do it. I can try if you prefer.

marcelm: And I forgot to mention: The samtools packages also don’t include `.h` files.

recipes/hapcut2/meta.yaml
marcelm: I’ll change it to be more consistent with other recipe. (Personally, I think this just adds clutter. The only advantage is that I don’t have to change the URL when bumping the version manually, but then I’ll have the file open anyway because I need to copy and paste the new checksum.)

marcelm:
I’ve updated the recipe according to review requests, but also added a more traditional all-lowercase `hapcut2` alias for the `HAPCUT2` command. (I saw this being done in the TIDDIT recipe.)

recipes/hapcut2/meta.yaml
jmarshall: Your sensible options are

1. Continue to bundle old samtools & htslib as per upstream and this recipe

2. Hope that HapCUT2 applies vibansal/HapCUT2#87 so it needs only htslib, and makes a new release, which you will be able to build against bioconda's htslib

3. Extract that PR as a patch to be applied by your recipe, and build against bioconda's htslib

marcelm: Cool, thanks so much! I think your PR has a good chance of getting merged considering what @pjedge wrote in vibansal/HapCUT2#86 . I’ll wait a bit and hope that we can go for option 2. I think any remaining cleanup can then wait for the next release.

dpryan79:
Can you confirm that headers and libraries from samtools/htslib aren't getting accidentally bundled? As long as that's all statically linked then the changes are fine.

recipes/hapcut2/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  build:
+    - {{ compiler('c') }}
+  host:
+    - zlib
+  run:
+    - zlib
+    - python
+    - pysam
+
+build:
```
bgruening: `build` should be above requirements, usually ... :)

marcelm: Fixed now :)

marcelm:
@dpryan79 The build script doesn’t do `make install` for any of the involved packages, so it should be fine. I also double-checked the created `.tar.bz2` now and I can confirm that no headers or libraries are bundled.

dpryan79:
Great, let's merge this then :)

dpryan79:
Oh, I see you have it labeled as WIP still, so just merge when you're ready (I've already approved the PR).

jenzopr:
Done :-) Thanks for your help!

marcelm:
Yes, I relabeled it because I was going to wait for a new HapCUT2 release, perhaps with @jmarshall’s patch that removes the samtools dependency.

However, I think I’ll just merge this now after all. It’s working fine and I don’t want to put any pressure on @pjedge. We can then package the next version when it’s released, hopefully removing some of the workarounds.

marcelm:
... or I’m not going to merge it (myself) after all ...

pjedge:
Hey everybody, sorry i was not involved in this discussion. I tested things out and merged the PR. Just cleaning things up a bit and i'll have a new release ready

pjedge:
[New HapCUT2 release] (https://github.com/vibansal/HapCUT2/releases/tag/v1.2) that removes submodules and uses the updated htslib API

PR #18437, opened 2019-11-04
Bump mfqe to 0.5.0.

wwood:

### DIFF
--- a/recipes/mfqe/build.sh
+++ b/recipes/mfqe/build.sh
@@ -6,4 +6,5 @@ LIBRARY_PATH=$PREFIX/lib \
 cargo build --release
 
 # Install the binary
+mkdir -p ${PREFIX}/bin
 cp target/release/mfqe $PREFIX/bin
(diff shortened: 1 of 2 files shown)


wwood:
Thanks @tseemann - OK, will try to remember. I would rather just run `cargo test` though to test it properly - is that supposed to be simple? I vaguely remember trying that once and it not working out for some reason. Is there value in running your suggested test alongside that?

tseemann:
I can't see `cargo test` in `build.sh` anywhere, only `cargo build` ?

Those would be build time tests. The `test:` block in `meta.yaml` is a run-time test, in the environment of the user. It's just a proper check that the thing actually runs on their system.

A _very_ common bug in software is that the `--version` they report is not the version they named their release; they often forget to update their variables. It also ensures the test is running the _correct_ binary, and not another one contaminating the environment somehow.

It's not compulsory, and most recipes aren't doing it, but I think it is a simple test that catches something that other tests don't.  But maybe the `cargo test` suite covers that.

Happy for this PR to be accepted as-is however!

wwood:
I see, ok. Thanks for the explanation.

PR #18449, opened 2019-11-04
Add GeneRax

BenoitMorel:
This pull request adds a new recipe for our gene tree inference tool GeneRax 
(https://www.biorxiv.org/content/10.1101/779066v1)

* [ok] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/contributor/guidelines.html).
* [ok] This PR adds a new recipe.
* [ok] AFAIK, this recipe **is directly relevant to the biological sciences** (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).

### DIFF
--- a/recipes/generax/build.sh
+++ b/recipes/generax/build.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+# pthreads
+mkdir build
+pushd build
+  cmake ..
+  make
+  install -d ${PREFIX}/bin
+  install ../build/bin/generax ${PREFIX}/bin
+  install ../build/bin/JointSearch ${PREFIX}/bin
+popd
+
--- a/recipes/generax/meta.yaml
+++ b/recipes/generax/meta.yaml
@@ -1,0 +1,45 @@
+{% set name = "generax" %}
+{% set version = "1.0.0" %}
+
+package:
+  name: {{ name|lower }}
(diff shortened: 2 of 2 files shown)


dpryan79:
Could you tag a release on that repository? We can merge in without that, but it'd make things rather more official to use a tagged release :)

BenoitMorel:
Thanks for the review.
I am not sure that I understand what you mean with tagging a release. There is a recent release (https://github.com/BenoitMorel/GeneRax/releases/tag/1.0.0)  that I tried to reference through the  `git_rev` field. Is that incorrect?

dpryan79:
Ah, I hadn't seen that there was an actual release. I've modified the recipe to use it.

BenoitMorel:
Ok now I get it, you want to directly link the archive. Unfortunately this will not work because I am using some submodules repository, and github does not include them in its archives (https://github.community/t5/How-to-use-Git-and-GitHub/Create-Release-that-contains-submodule/td-p/12073)
I could manually generate a tar.gz file including the missing sources and add it to the release. Would it help?

dpryan79:
Ah, we should just mention that in the meta.yaml and then revert my changes.

BenoitMorel:
Ok, great. Thank you so much for your help :-)

PR #18463, opened 2019-11-04
update keras/tf dependency in medaka

cjw85:

### DIFF
--- a/recipes/medaka/meta.yaml
+++ b/recipes/medaka/meta.yaml
@@ -12,7 +12,7 @@ source:
 build:
   # keras is 2.7 or 3.6 only. We need py3: 3.6 is the only possibility
   skip: True # [not py36]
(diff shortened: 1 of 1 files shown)


cjw85:
@druvus Can you explain the libdeflate issue? I think I recall seeing something about this on an bioconda issue?

dpryan79:
The version of pysam you specified hasn't been rebuilt for the current version of libdeflate we use. If you really need to directly link against libdeflate then we'll need to rebuild that pysam version (otherwise, just remove libdeflate from the dependencies). Alternatively, just use the most recent version of pysam.

cjw85:
Thanks, I'll unpin the pysam (that's simply the version we pin to for pypi builds because the later versions don't have wheels, which is not an issue here); I suspect later version is fine given change log.

PR #18567, opened 2019-11-09
New recipe for GoGetData (GGD)

mikecormier:

### DIFF
--- a/recipes/ggd/build.sh
+++ b/recipes/ggd/build.sh
@@ -1,5 +1,0 @@
-#!/bin/bash
-
-rm -rf tests
-
-$PYTHON setup.py install
--- a/recipes/ggd/build.sh
+++ b/recipes/ggd/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+rm -rf tests
+
+$PYTHON setup.py install
--- a/recipes/ggd/meta.yaml
+++ b/recipes/ggd/meta.yaml
@@ -1,0 +1,71 @@
+{% set version = "0.1.0" %}
+
+package:
+  name: ggd
+  version: "{{ version }}"
+
+source:
+  url: https://github.com/gogetdata/ggd-cli/archive/v{{ version }}.tar.gz
+  sha256: 540c67b4beb9dee34fa07c55e86e6f58c17dbc345a4700207cbc87615040a7ff
+
+build:
+    number: 0
+    binary_relocation: False
+    noarch: python
+
+requirements:
+  build:
+    - python
+    - requests>=2.20.0
+    - pyyaml
+    - gitpython
+    - git
+    - conda=4.7.12
+    - conda-build=3.18.11
+    - ripgrep
+    - fuzzywuzzy
+    - python-levenshtein
+    - htslib
+    - samtools
+    - zlib
+    - check-sort-order
+    - krb5
+    - gsort>=0.1.2
+    - wget
+    - pytest
+    - future
+    - oyaml
+    - pytest-socket
+  run:
+    - python
+    - requests>=2.20.0
+    - pyyaml
+    - gitpython
+    - git
+    - conda=4.7.12
+    - conda-build=3.18.11
+    - ripgrep
+    - fuzzywuzzy
+    - python-levenshtein
+    - htslib
+    - samtools
+    - zlib
+    - check-sort-order
+    - krb5
+    - gsort>=0.1.2
+    - wget
+    - pytest
+    - future
+    - oyaml
+    - pytest-socket
+
+test:
+  commands:
+    - ggd --version
+
+about:
+  home: https://github.com/gogetdata/ggd-cli
+  docs: https://gogetdata.github.io/
+  license: MIT
+  summary: "GoGetData (GGD) is a genomic data managment system. It provide simple and reproducible access to a repository of genomic data. Simply put, it is 'Conda' for genomic data"
+
--- a/recipes/ggd/meta.yaml
+++ b/recipes/ggd/meta.yaml
@@ -2,7 +2,7 @@
 
 package:
   name: ggd
-  version: "{{ version }}"
+  version: {{ version }}
 
 source:
   url: https://github.com/gogetdata/ggd-cli/archive/v{{ version }}.tar.gz
@@ -10,34 +10,17 @@ source:
 
 build:
     number: 0
-    binary_relocation: False
     noarch: python
+    entry_points:
+      - ggd = ggd.__main__:main
+    script: "python -m pip install . --no-deps --ignore-installed -vv" 
 
 requirements:
-  build:
-    - python
-    - requests>=2.20.0
+  host:
+    - pip
     - pyyaml
-    - gitpython
-    - git
-    - conda=4.7.12
-    - conda-build=3.18.11
-    - ripgrep
-    - fuzzywuzzy
-    - python-levenshtein
-    - htslib
-    - samtools
-    - zlib
-    - check-sort-order
-    - krb5
-    - gsort>=0.1.2
-    - wget
-    - pytest
-    - future
-    - oyaml
-    - pytest-socket
   run:
-    - python
+    - pip
     - requests>=2.20.0
     - pyyaml
     - gitpython
--- a/recipes/ggd/meta.yaml
+++ b/recipes/ggd/meta.yaml
@@ -18,9 +18,9 @@ build:
 requirements:
   host:
     - pip
-    - pyyaml
+    - python
   run:
-    - pip
+    - python
     - requests>=2.20.0
     - pyyaml
     - gitpython
--- a/recipes/ggd/post-link.sh
+++ b/recipes/ggd/post-link.sh
@@ -1,14 +1,8 @@
 #!/bin/bash
 set -eo pipefail -o nounset
 
-## Organize required conda channels
-conda config --add channels defaults
-conda config --add channels ggd-genomics
-conda config --add channels bioconda
-conda config --add channels conda-forge
-
 ## initialize local metadata
-cat << EOF > initialize_metadata.py
+"${PREFIX}/bin/python" -c "$( cat <<'EOF'
 """
 Set up initial ggd local metdata
 """
@@ -32,7 +26,9 @@ for x in channels:
 
 ## Initialize pkg metadata tracking
 utils.update_installed_pkg_metadata()
-
 EOF
-python initialize_metadata.py 
-rm initialize_metadata.py
+)" >> "${PREFIX}/.messages.txt" 2>&1
+
+echo -e "\n\nNOTE: Please run the following command to add the required conda channels for GGD: " >> "${PREFIX}/.messages.txt" 2>&1  
+echo -e "\tconda config --add channels defaults\n\tconda config --add channels ggd-genomics\n\tconda config --add channels biconda\n\tconda config --add channels conda-forge" >> "${PREFIX}/.messages.txt" 2>&1
+
--- a/recipes/ggd/post-link.sh
+++ b/recipes/ggd/post-link.sh
@@ -1,0 +1,38 @@
+#!/bin/bash
+set -eo pipefail -o nounset
+
+## Organize required conda channels
+conda config --add channels defaults
+conda config --add channels ggd-genomics
+conda config --add channels bioconda
+conda config --add channels conda-forge
+
+## initialize local metadata
+cat << EOF > initialize_metadata.py
+"""
+Set up initial ggd local metdata
+"""
+from ggd import utils
+
+## Update local genomic metadata files
+utils.update_genome_metadata_files()
+
+##Update local channeldata.json metadata file
+utils.update_channel_data_files("genomics")
+
+## Get species
+utils.get_species(update_files=True)
+
+## get channels
+channels = utils.get_ggd_channels()
+
+## get channel data
+for x in channels:
+    utils.get_channel_data(x)
+
+## Initialize pkg metadata tracking
+utils.update_installed_pkg_metadata()
+
+EOF
+python initialize_metadata.py 
+rm initialize_metadata.py


mikecormier:
ping @johanneskoester

recipes/ggd/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+  name: ggd
+  version: "{{ version }}"
+
+source:
+  url: https://github.com/gogetdata/ggd-cli/archive/v{{ version }}.tar.gz
+  sha256: 540c67b4beb9dee34fa07c55e86e6f58c17dbc345a4700207cbc87615040a7ff
+
+build:
+    number: 0
+    binary_relocation: False
```
dpryan79: Is there a reason this is set to false?

recipes/ggd/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+    noarch: python
+
+requirements:
+  build:
```
dpryan79: Please change `build:` to `host:`

recipes/ggd/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+    - python
+    - requests>=2.20.0
+    - pyyaml
+    - gitpython
+    - git
+    - conda=4.7.12
+    - conda-build=3.18.11
```
dpryan79: Does this package really require conda and conda-build internally?

recipes/ggd/post-link.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+set -eo pipefail -o nounset
+
+## Organize required conda channels
+conda config --add channels defaults
+conda config --add channels ggd-genomics
```
dpryan79: If we're going to change people's channel order on them we had better advertise that we're doing so.

bgruening: We should not mess around with the channel order in post-link ... 
If you want to do these things print a message to your user with instructions but don't mess around with settings that are outside of the package dir please.

dpryan79: We just discussed this internally within the core team and we're pretty uncomfortable with using post-link actions to modify people's channels. If adding `ggd-genomics` to the channels is really required then I suggest adding this package there instead.

recipes/ggd/meta.yaml
mbargull: Apparently, yes.
It even deliberately uses non-public functions from `conda` which is, let's say "adventurous" xD. But they pin down the version requirements exactly, so if they know what they are downing, that *could be* fine.

recipes/ggd/post-link.sh
mbargull: This gets a big fat **not approved** from my part.
It should *at least* be
```sh
\. "${PREFIX}"/etc/profile.d/conda.sh
conda activate "${PREFIX}"
conda config --env ...
```
to restrict it to the install environment.

But even then:
General advice: If anyhow possible, avoid post-link scripts. It's non-standard code that needs non-standard maintenance and may fail in non-standard ways. Overall, it's just costly.

recipes/ggd/post-link.sh
```diff
@@ -0,0 +1,38 @@
...
+conda config --add channels bioconda
+conda config --add channels conda-forge
+
+## initialize local metadata
+cat << EOF > initialize_metadata.py
```
mbargull: `conda` *does* execute the post-link scripts inside the installation prefix. But I'm not sure if we guarantee that.
To not rely on the current working directory, write to files inside the prefix, e.g., `>"${PREFIX}/unique-file-name`.
But in this case, you can avoid that temporary file altogether:
```sh
"${PREFIX}/bin/python" -c "$( cat <<'EOF'
# python code
EOF
)" >"${PREFIX}/.messages.txt" 2>&1
```
Also note the `"${PREFIX}/bin/python"` and `>"${PREFIX}/.messages.txt" 2>&1` parts: make sure to use the correct `python` binary and output any messages to `.messages.txt` or else `conda` won't display them to the user.

Then again, as noted above:
General advice: If anyhow possible, avoid post-link scripts.

recipes/ggd/post-link.sh
```diff
@@ -0,0 +1,38 @@
...
+"""
+Set up initial ggd local metdata
```
mbargull: *IF* you write files into or generally alter the environment via post-link, please also add an accompanying pre-unlink script to revert those actions.

Then again, as noted above:
General advice: If anyhow possible, avoid post-link scripts.

recipes/ggd/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+rm -rf tests
+
+$PYTHON setup.py install
```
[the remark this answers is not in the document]
mbargull: (NB: I'm no fan of the "Python special handling" via `${PYTHON} ...` `"{{ PYTHON }} ..."`, but it's common convention, I think. At least on conda-forge it's recommended, can't remember if we do that on Bioconda, too..)

recipes/ggd/meta.yaml
mikecormier: Yes, this package does use conda and conda-build and non-public functions of conda. Conda acts as the primary infrastructure for our package.

recipes/ggd/post-link.sh
mikecormier: I note your concerns, however, the post-link script was deliberately set up this way to provide usability with "GGD" channels. The `gg-genomics` channel is added here as a requirement of GGD. However, we didn't want to force the "GGD" channels to take precedence over other channels such as conda-forge or bioconda. Additionally, it is recommended that conda-forge have the highest priority and bioconda the second highest. I am happy to put a message that informs the user that we are changing the channel priority, or even only adding GGD channels, but this is a vital part of the recipe. 

Additionally, the `ggd-genomics` and any other GGD channel is primarily for data packages and does not support software.

recipes/ggd/build.sh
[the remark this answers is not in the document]
mikecormier: @mbargull, thanks for the suggestions, however, this convention is pretty standard and as such I am going to keep it this way.

recipes/ggd/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+  version: "{{ version }}"
+
+source:
+  url: https://github.com/gogetdata/ggd-cli/archive/v{{ version }}.tar.gz
+  sha256: 540c67b4beb9dee34fa07c55e86e6f58c17dbc345a4700207cbc87615040a7ff
+
+build:
+    number: 0
+    binary_relocation: False
+    noarch: python
```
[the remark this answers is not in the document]
mikecormier: The entry point is established in the setup.py, thus this change is not necessary

recipes/ggd/build.sh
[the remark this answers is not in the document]
bgruening: @mikecormier we do have reasons for those recommendations :)
Please stick to them if you want your recipe be part of Bioconda.

recipes/ggd/post-link.sh
bgruening: I think it is fine if you give instructions and a nice readme which is displayed when people install this package, but please do not change "system" setting on behalf of the user.

mikecormier: @bgruening, one of the advantages of using "conda" and the ideas of recipes is to provide dependency handling for the user so they do not have to do it themselves. This is the intent for this step of the process.

dpryan79: We're aware of that, but this change isn't one we can permit in Bioconda recipes. This is a deal breaker for accepting the package.

recipes/ggd/build.sh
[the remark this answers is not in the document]
mikecormier: @bgruening, I understand the reasons for recommendations, however, the recommended is unwarranted as the current install is a convention held in conda and bioconda, and is the traditional way to use a setup command.

recipes/ggd/post-link.sh
mikecormier: Is it allowable if add only the `ggd-genomics` channel?

dpryan79: No, we can't allow changing people's channel orders on the system level. What you probably can do is change your internal usage of `conda` to inject it there (e.g., making system calls to `conda install -c ggd-genomics ...` or figuring out where in the conda code it reads in the channels and modify that). I imagine asking users to instead modify their channels is easier though.

recipes/ggd/build.sh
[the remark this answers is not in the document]
mbargull: > as the current install is a convention held in conda and bioconda

No:
```sh
grep -rh 'setup.py install\|pip install' bioconda-recipes/recipes | sed 's/^\s*-\?\s*//' | sort | uniq -c | sort -n | tail
     20 script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv"
     20 script: "{{ PYTHON }} -m pip install . --no-deps -vv"
     21 $PYTHON -m pip install . --ignore-installed --no-deps -vv
     21 script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed -vvv "
     23 $PYTHON setup.py install
     28 script: python setup.py install --single-version-externally-managed --record=record.txt
     83 script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed -vv "
    111 script: python -m pip install --no-deps --ignore-installed .
    181 $PYTHON setup.py install --single-version-externally-managed --record=record.txt
    225 script: {{ PYTHON }} -m pip install . --ignore-installed --no-deps -vv
```
We can see that there are recipes with only `setup.py install`, but only few and they should probably be changed.
The most prevalent conventions are `pip install ...` or `setup.py install --single-version-externally-managed --record=record.txt` and as @bgruening mentioned, there are reasons for this.
Some reasons:
1. Additional parameters to `setup.py` are to prevent `setuptools` to download additional dependencies (which would erroneously be added to the package we are trying to build).
   (In practice you'll likely get a build failure because `conda-build` sets things up to error out when such downloads would happen -- but this should only be regarded as a safety net, IMO.)
2. `pip install` you'll get nicer entry point scripts which do not enforce a direct run dependency on `setuptools`.
   (`setuptools` entry points use `pkg_resources` which is part of `setuptools` itself.)
   (`conda-build` sets up some environment variables for `pip`, too. If you think that might be a little non-transparent (which I might agree with), you can use `pip install . --no-deps --ignore-installed -vv`, too)

recipes/ggd/meta.yaml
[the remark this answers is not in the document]
mbargull: Yes and no. If you change the build to use `pip install`, I would be fine with not adding the `entry_points` entry here.
(For Windows users it is still advantageous to let `conda` create those scripts/exe wrappers -- Bioconda does not "do Windows" so this is not that important. However, there are `noarch` packages on `bioconda` (with entry points) that are totally usable on Windows too.)

recipes/ggd/post-link.sh
mbargull: If we reach consensus on this PR to use a post-link script which `conda config` invocations, the points from my previous comment still stand:
1. should use "its own `conda`", i.e., the one in "${PREFIX}"
2. has to add `--env` since otherwise we are messing with users' configuration outside of the environment.

mbargull: BTW, there are several ways to "inject" additional channels for `conda` to consider. I don't know what the workflow with `ggd` is, but if the channel `ggd-genomic` is only queried thorough `ggd` and the user is not expected to necessarily interact with `conda` directly, it might be worthwhile to inject `ggd-genomic` during runtime. Feel free to ask/ping me on https://github.com/gogetdata/ggd-cli if you need pointers (I know one or two things about `conda`'s innards ;).

mikecormier: This has been removed from the post-link script

recipes/ggd/build.sh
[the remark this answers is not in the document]
mikecormier: Thanks for the suggestions and info @mbargull. I have updated the recipe to use `script: python -m pip install . --no-deps --ignore-installed -vv` in the meta.yaml file and have removed the build.sh

recipes/ggd/meta.yaml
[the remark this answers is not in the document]
mikecormier: This has been changed in the recipe. Thanks for the suggestions.

recipes/ggd/post-link.sh
mikecormier: I have changed it per your suggestion. Rather than writing a new python script, executing it, and removing it, it now uses the -c command to do it from a string.

recipes/ggd/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    number: 0
+    noarch: python
+    entry_points:
+      - ggd = ggd.__main__:main
+    script: "python -m pip install . --no-deps --ignore-installed -vv" 
+
+requirements:
+  host:
+    - pip
+    - pyyaml
```
mbargull:
```suggestion
    - python
```
We should have both `pip` and `python` here since we run `python -m pip`, i.e., the build directly depends on both (better be explicit in this case).
(Plus, I don't know whether `conda-build` does some funky `python`-specific things that expect `requirements/host/python` to be present.)
We can remove `pyyaml` since with `pip install --no-deps` the build won't touch `pyyaml` at all now, which makes things easier.

recipes/ggd/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  run:
+    - pip
```
mbargull:
```suggestion
    - python
```
Unless `ggd` uses `pip` directly (a GitHub search on your repo didn't yield anything), we can remove this direct dependency. (If `conda` uses `pip`, that's fine since it's a `conda` dependency.)
We should have `python` here because we have a `python` package ;).

recipes/ggd/post-link.sh
```diff
@@ -0,0 +1,34 @@
...
+channels = utils.get_ggd_channels()
+
+## get channel data
+for x in channels:
+    utils.get_channel_data(x)
+
+## Initialize pkg metadata tracking
+utils.update_installed_pkg_metadata()
+EOF
+)" >> "${PREFIX}/.messages.txt" 2>&1
```
mbargull: Right, I forgot to write `>>` in my previous suggestion -- but this is exactly the right thing to do. Good catch!

recipes/ggd/post-link.sh
mbargull: Ideally, we'd have a pre-unlink script here. If not, a `conda remove ggd` won't remove files created here since `conda` does not keep track of post-link created files.

I'd like to encourage you to consider/look into how to revert the post-link actions in a pre-unlink script. However, if this is not easily doable right now, I won't block your PR on this.

recipes/ggd/meta.yaml
mikecormier: When python is added to the meta.yaml, as suggested above, it installs python2 in a python3 environment. GGD works for both python2 and 3 so I can't designate which python version to install. I'm not sure why this is going on, but I would rather not change the base python version when installing GGD. Any suggestions?

recipes/ggd/meta.yaml
mikecormier: It does not use pip directly

recipes/ggd/meta.yaml
mbargull: Hm, sound like a coincidence, maybe? With the `build/noarch: python` specified, we should get a package that just depends on `python` with no additional version constraints, so I'm a bit surprised about what you described.
Can you change the recipe in this PR as suggested with `python` both in `host` and `run`? then I can inspect the build log and created package to check what's happening.

recipes/ggd/meta.yaml
mbargull: Good, then we can omit that dependency here :)

recipes/ggd/meta.yaml
mikecormier: meta.yaml is changed.

mikecormier: The build log is consistent with what I see on my local machine, that is, python 2 is being installed when `- python` is set in `build/host` even when you are using a python 3 environment.

dpryan79: I'm not sure where you see that in the logs, conda just randomly chooses different python versions for different independent steps in cases like this, which is perfectly fine.

mikecormier: bioconda-utils depends on python3, which I assume is installed because bioconda-utils is running. From the log:
```
16:00:43 BIOCONDA INFO (OUT)     _libgcc_mutex:   0.1-main             defaults
16:00:43 BIOCONDA INFO (OUT)     ca-certificates: 2019.9.11-hecc5488_0 conda-forge
16:00:43 BIOCONDA INFO (OUT)     certifi:         2019.9.11-py27_0     conda-forge
16:00:43 BIOCONDA INFO (OUT)     libffi:          3.2.1-he1b5a44_1006  conda-forge
16:00:43 BIOCONDA INFO (OUT)     libgcc-ng:       9.1.0-hdf63c60_0     defaults
16:00:43 BIOCONDA INFO (OUT)     libstdcxx-ng:    9.1.0-hdf63c60_0     defaults
16:00:43 BIOCONDA INFO (OUT)     ncurses:         6.1-hf484d3e_1002    conda-forge
16:00:43 BIOCONDA INFO (OUT)     openssl:         1.1.1d-h516909a_0    conda-forge
16:00:43 BIOCONDA INFO (OUT)     pip:             19.3.1-py27_0        conda-forge
16:00:43 BIOCONDA INFO (OUT)     python:          2.7.15-h5a48372_1009 conda-forge
16:00:43 BIOCONDA INFO (OUT)     readline:        8.0-hf8c457e_0       conda-forge
16:00:43 BIOCONDA INFO (OUT)     setuptools:      41.6.0-py27_1        conda-forge
16:00:43 BIOCONDA INFO (OUT)     sqlite:          3.30.1-hcee41ef_0    conda-forge
16:00:43 BIOCONDA INFO (OUT)     tk:              8.6.9-hed695b0_1003  conda-forge
16:00:43 BIOCONDA INFO (OUT)     wheel:           0.33.6-py27_0        conda-forge
16:00:43 BIOCONDA INFO (OUT)     zlib:            1.2.11-h516909a_1006 conda-forge
```

Note that python 2 is designated. 

A warning from pip, suggesting it is the version being used: 
```
16:01:07 BIOCONDA INFO (OUT) DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
```

mbargull: You're right, but everything is fine ;).
- `bioconda-utils`+`python=3` is used as the base for the CI.
- The build via `conda-build` happens in a container that also has `bioconda-utils`+`python=3`
- `conda-build` chooses whatever it likes (within the given constraints, of course) during the build. Very often it chooses `python=2` (possibly due to the order in which possible `python` versions are given in our build setup). This is a little strange, admittedly, but does not matter since `conda-build`+`pip` usually build the same `noarch: python` package, no matter which `python` version is used during the build.
- The built package is tested in an additional container at the end of the CI. You can see that it chooses `python=3.7` there, i.e., we've built with `python=2` but the package tests fine in a `python=3` environment.

mbargull:
@dpryan79 if you agree with my approval and @mikecormier is fine with the final state of the recipe, feel free to merge.

mikecormier:
@mbargull  and @dpryan79  I'm good with how it is. Thanks for working with me to get this out

PR #18694, opened 2019-11-18
	new file:   kart/build.sh

hsinnan75:
new file:   kart/meta.yaml


hsinnan75:
please review & merge

recipes/kart/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+if [ ! -d $PREFIX/bin ] ; then
+  mkdir $PREFIX/bin
+fi
+
+make Compiler=$CXX CXX=$CXX CC=$CC LDFLAGS="-L${PREFIX}/lib" CFLAGS="-I${PREFIX}/include -L${PREFIX}/lib" CXXFLAGS="-I${PREFIX}/include -L${PREFIX}/lib"
```
mbargull: The compiler activation scripts also set `{LD,C,CXX}FLAGS` which also set the `-L` and `-I` (or rather, as of recently ,`-isystem`) for the `include`/`lib` paths.
You should be able to use
```suggestion
make Compiler=$CXX CXX=$CXX CC=$CC LDFLAGS="${LDFLAGS}" CFLAGS="${CFLAGS}" CXXFLAGS="${CXXFLAGS}"
```
If the `-L` is in fact somehow expected to be part of the `{C,CXX}FLAGS`, then
```suggestion
make Compiler=$CXX CXX=$CXX CC=$CC LDFLAGS="${LDFLAGS}" CFLAGS="${CFLAGS} ${LDFLAGS}" CXXFLAGS="${CXXFLAGS} ${LDFLAGS}"
```
or (just in case the above fails)
```suggestion
make Compiler=$CXX CXX=$CXX CC=$CC LDFLAGS="${LDFLAGS}" CFLAGS="${CFLAGS} -L${PREFIX}/lib" CXXFLAGS="${CXXFLAGS} -L${PREFIX}/lib"
```
might work.

recipes/kart/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  sha256: '99f28beb55ebd118c7e7e27507805d9725f0e490557107cf7058351dbed05981'
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('c') }}
+  host:
```
mbargull: I saw that this builds `htslib` in the `makefile`.
If you want to, you could reuse our pre-built `htslib` for this (also includes `htslib.a`).
Since you statically link, I guess it does not make much difference, but I'm not expert on this.

tseemann: @mbargull i think it may be a custommodified version of `htslib` but i"ll wait for @hsinnan75 to confirm

recipes/kart/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - boost-cpp
+  run:
+    - zlib
+    - bzip2
+    - xz
+    - boost-cpp
+
+test:
+  commands:
+    - kart
```
tseemann: Maybe a better test might be `kart -version | grep '{{ version }}'`

recipes/kart/meta.yaml
hsinnan75: @mbargull The htslib is a modified version. Thanks to @tseemann for the reminder.

mbargull: Good to know, thanks @tseemann!

recipes/kart/meta.yaml
mbargull: Sounds good.
```suggestion
kart --version | grep -qF '{{ version }}'
```
with `-F` we can compare the `.`s in the version string correctly.

recipes/kart/build.sh
mbargull: Argh, looks like https://github.com/hsinnan75/Kart/blob/2.5.2/src/BWT_Index/makefile#L14 is no fan of using `LDFLAGS`. We have to use one of the other suggestions then. (Or patch the makefile, or do a separate `make` of `bwt_index` with the altered flags. But I'd assume injecting it into `{C,CXX}FLAGS` is most convenient for you.)

mbargull: Please reuse the flags the compiler activation scripts set. They ensure compatibility between libraries as well as with the minimum system requirements for the users.
Try
```
make Compiler=$CXX CXX=$CXX CC=$CC LDFLAGS="${LDFLAGS}" CFLAGS="${CFLAGS} ${LDFLAGS}" CXXFLAGS="${CXXFLAGS} ${LDFLAGS}"
```
and if that does not work
```
make Compiler=$CXX CXX=$CXX CC=$CC LDFLAGS="${LDFLAGS}" CFLAGS="${CFLAGS} -L${PREFIX}/lib" CXXFLAGS="${CXXFLAGS} -L${PREFIX}/lib"
```

hsinnan75: I've fixed the problem and updated build.sh using LDFLAGS="${LDFLAGS}" CFLAGS="${CFLAGS}" CXXFLAGS="${CXXFLAGS}"

recipes/kart/meta.yaml
hsinnan75: Thanks for the suggestion. However, it produces nothing with the commands. I used "kart -h" instead.

mbargull: Do you mean `kart` or the test command?
For `kart`, I assumed it did because of https://github.com/hsinnan75/Kart/blob/2.5.2/src/main.cpp#L155.
As for the test command, yes, no text output. The output is just exitcode 0 if it succeeded and 1 if not (i.e., tests will fail)
If you want to see the output as well as test it in a single command, you can use
```
kart --version | tee /dev/stderr | grep -qF '{{ version }}'
```
but seeing the output is not needed here.
You can also do
```suggestion
    - kart -h
    - kart --version
    - kart --version | grep -qF '{{ version }}'
```
if you want to test the output and see everything in the build log.

hsinnan75: Thank you! I got it. I'll use those commands for testing new packages.

mbargull:
Oh, I only read `requested review from mbargull` and overlooked the `and tseemann` part.
@tseemann, hope you're okay with it!?

mbargull:
@hsinnan75, do you want to join Bioconda? If yes, I can send you an invitation.

hsinnan75:
@mbargull Thank you for the invitation. I'd like to join Bioconda. It's my great honor.

PR #18743, opened 2019-11-19
Building odgi

apeltzer:
This is the first attempt at bringing odgi to bioconda-recipes. 


apeltzer:
@subwaystation x-ref

dpryan79:
@bioconda-bot please update

subwaystation:
@dpryan79 Thanks for helping out. Keep going, if you have any ideas!

As I discovered, the current release does not compile out of the box. One would have to check out the latest commit of that release involving git commands.

@ekg recommended to me to take a look at https://bioconda.github.io/recipes/freebayes/README.html. The project's structure should be similar to odgi.
@apeltzer Maybe this will help us further.

dpryan79:
It looks like the build instructions for odgi are simply imcomplete.

ekg:
I guess I need to specify that you have to use git to clone the repository, and that git is a required dependency of this software.

To build a given release (in this case, v0.1)

```
git clone https://github.com/vgteam/odgi.git
git checkout v0.1
```

Now follow the build instructions in the README:

```
cmake -H. -Bbuild && cmake --build build -- -j 4
```

If you want to build the python bindings, you'll need pybind11 installed on your system. I was unable to get it to build within the tree without using a bleeding edge version of cmake to run the configuration.

There may be a minimal set of system dependencies (including a new version of gcc/g++). I have not sorted out exactly which they are.

ekg:
I'm not sure if this is the problem. I've tested builds within a directory (build) but not within the source tree. Please use the same cmake command. If needed, maybe a `git submodule update --init --recursive` should be issued after the git checkout. I don't really understand what the failure is here. It look like the makefile generated by the cmake call is never run.

dpryan79:
We're using a quite recent version of cmake (3.15.5). At the moment I'm not sure that anything is actually getting compiled or, if it is, whether it's actually getting installed anywhere.

ekg:
Is the call equivalent to this:

```
cmake -H. -Bbuild && cmake --build build -- -j 4
```

?

dpryan79:
Without the `-- -j 4`, yes, since that seems to confuse cmake for some odd reason.

dpryan79:
If we change those then nothing could be installed into the correct place.

ekg:
> If we change those then nothing could be installed into the correct place.

I agree, but let's see if the build progresses without them. Then we can fix it upstream.

dpryan79:
Hmm, it does in fact actually start to compile then.

ekg:
Should I fix upstream to abide by these set parameters?

dpryan79:
Yes, that'd make life easier.

ekg:
I have a recipe to build odgi with guix. https://github.com/ekg/guix-genomics/blob/master/odgi.scm 

Would this perhaps help?

I also moved to using git submodules, as this allows the tree to be checked out and validated against a known hash before compilation begins.

dpryan79:
Maybe? We can try switching to the master branch just to test things out.

dpryan79:
@bioconda/bot please update

dpryan79:
@bioconda-bot please update

subwaystation:
While building `sonlib` the linker has problems with the `-lz` argument.
Any ideas?

dpryan79:
Cmake needs to pass on the appropriate flags to sonlib, or at the very least the sonlib cmake file needs to respect environment variables, since `LIBRARY_PATH` and `LD_LIBRARY_PATH` are set and specify the correct library location.

subwaystation:
@dpryan79 Would a `guix` build as described in https://github.com/vgteam/odgi/issues/55#issuecomment-580208901 help you further?

dpryan79:
I don't think so, no.

ekg:
It does explain exactly what dependencies this has.

How are the builds done here? What's a good tutorial?

dpryan79:
The issue doesn't seem to be having the right dependencies, but rather convincing cmake to propogate compiler flags along that say exactly where those dependencies are.

ekg:
In the guix build, I have to set CC=gcc. I think this was to make sonLib's build happy. But, the error I got was different.

I'm slightly inclined to axe sonLib. It's included for a feature that's not even used, and which isn't even usable for inputs of any significant dimension.

subwaystation:
@dpryan79 We have a branch https://github.com/vgteam/odgi/tree/conda where `sonLib` is completely removed. Can you try to pull in the branch?
Or do we have to merge into master and make a new release?

dpryan79:
We can give that a go to see if it works at least.

dpryan79:
@bioconda-bot please update

dpryan79:
@subwaystation Looks like that'll move this along significantly. What's the preferred way of getting this installed in a given location? I'm modifying the build script to just move stuff manually, but I suspect you have a better way to do this using cmake.

subwaystation:
I think the `odgi` binary should be available globally.
And the python bindings should be available in the `PATH`.
@ekg What do you think would be the best way to go here?
@jervenbolleman Would that suffice for `Spodgi`?

subwaystation:
Thanks for the hard tries so far @dpryan79 . Happy that we moved forward :)

dpryan79:

dpryan79:
Currently this should work on linux, I'll try to fix the OSX build now. The python bindings will need to get moved to python-specific library locations I expect. Which python versions are supported by odgi? Namely, is 2.7 still supported?

subwaystation:
I just installed the package on Linux, and it worked :)

As https://github.com/pybind/pybind11 is integrated as a submodule, I would follow their supported versioning:

- 2.7
- 3.x

2.7 is outdated and not supported anymore. I vote to leave that one out @ekg ?

ekg:
@subwaystation let's not support 2.7. I think any 3-series should be fine. I just don't have a mechanism to figure out what the minimum requirements would be, and I don't want to set something too strict.

dpryan79:
I’ll have it make python 3 only builds later tonight then. We’ll support the same versions as conda-forge for now.

apeltzer:
Sorry - as job changed, couldn't have a look earlier than today. Thanks for moving this forward @dpryan79 @ekg @subwaystation & everyone else 👍

dpryan79:

dpryan79:
How do I specify the path to python for cmake here? It seems to be using whichever version it finds that's higher, rather than just "whichever python comes first in the path".

subwaystation:
I don't have experience with cmake. @ekg Do you have an idea?

JervenBolleman:
Probably need to set PYTHON_PREFIX to the path of the right library installation, See the tools directory of deps/pybind11/ where some cmake magic is hidden.

dpryan79:
This is now compiling properly with all python 3 versions. If someone could tag a release so we have a stable source code target then we can finally merge this :)

ekg:
Are releases required for inclusion in bioconda? Happy to make one, but I'm
just curious about the practice and policy. AFAICT, they are just
associating a name with a git commit.

dpryan79:
They're not explicitly required, but it's tough to have a stable version if we just use a random branch that will change over time.

subwaystation:
@dpryan79 We now have a release with the conda branch merged in: https://github.com/vgteam/odgi/releases/tag/0.2.

Could you please update the building process?

How is the procedure, if there is a new release? We just open an issue here?

dpryan79:
For new releases a PR here will be automatically created within about an hour. If nothing in the build process needs to change than anyone who has joined the bioconda project can merge it.

subwaystation:
I see. Could you please explain how this repository watches releases of e.g. `odgi`? Just wondering how that is implemented and an automated PR is generated.. Thanks!

dpryan79:
There's an hourly cron job that runs `bioconda-utils autobump`. That checks github repositories listed in the `source:` section for new tags and changes recipes to use them if they look like a higher version than what the associated recipe already has.

dpryan79:
@bioconda-bot please merge

Thanks everyone!

subwaystation:
Thank you all for your patience!

PR #18771, opened 2019-11-20
Add r-vision recipe

tdido:

### DIFF
--- a/recipes/r-vision/meta.yaml
+++ b/recipes/r-vision/meta.yaml
@@ -28,7 +28,6 @@ requirements:
     - r-jsonlite
     - r-loe
     - r-logging
-    - parallel
     - r-pbmcapply
     - r-matrix
     - r-matrixstats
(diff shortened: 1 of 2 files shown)


recipes/r-vision/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+    - r-vegan
+    - r-wordspace
+
+test:
+  commands:
+    - $R -e "library('VISION')"
+
+about:
+  home: https://github.com/YosefLab/VISION
+  license: MIT
```
dpryan79: Can you add a `license_file: LICENSE` line so a copy of the license is included?

recipes/r-vision/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+    - r-wordspace
+  run:
+    - r-base
+    - r-fastica
+    - r-igraph
+    - r-irlba
+    - r-jsonlite
+    - r-loe
+    - r-logging
+    - parallel
```
dpryan79: Do you really mean parallel and not r-parallel or something like that?

tdido: Uff, good catch.

PR #18897, opened 2019-11-25
add squeezemeta

fpusan:
Hello, @bioconda/core, first timer here. If I got it right, I have to create the PR before in order for the tests to happen, so here I go. Will add the `please review & label` when tests are passed.


recipes/squeezemeta-dev/build.sh
```diff
@@ -0,0 +1,12 @@
...
+mkdir $SQM_DIR
+cp -r $SRC_DIR/* $SQM_DIR
+cd $PREFIX/bin
+ln -s $SQM_DIR/scripts/SqueezeMeta.pl .
+ln -s $SQM_DIR/scripts/restart.pl .
+ln -s $SQM_DIR/scripts/preparing_databases/make_databases.pl .
+ln -s $SQM_DIR/scripts/preparing_databases/download_databases.pl .
+ln -s $SQM_DIR/scripts/preparing_databases/configure_nodb.pl .
+ln -s $SQM_DIR/utils/* .
+R CMD INSTALL $SQM_DIR/bin/DAS_Tool/package/DASTool_*.tar.gz
```
dpryan79: das_tools is available as a dependency, do you need to vendor it?

fpusan: If possible, I'd prefer to. This is a complex pipeline that is meant to be easily installed without resorting to conda, thus we redistribute several pieces of software. Or course, conda makes things much easier (that's why we're trying to figure it out!) but we'd like to preserve compatibility for non-conda users too.

dpryan79: Right, but that doesn't mean you explicitly need to vendor it here then. This is also vendoring an insane number of things like bwa and canu, that should be removed from the resulting conda package and moved to regular dependencies.

fpusan: For most of the packages this should be easy to do. Others have minor tweaks in the code (e.g. to accept custom database paths) and can not be replaced by their original versions.

I'll try to thin it down as much as possible when I find some extra time, and also correct the meta.yaml file so that it's bioconda-compliant. I'll make a new PR then.

PR #18999, opened 2019-11-29
More blacklist updates

dpryan79:

### DIFF
--- a/recipes/groopm/meta.yaml
+++ b/recipes/groopm/meta.yaml
@@ -7,6 +7,8 @@ package:
 source:
   url: https://pypi.io/packages/source/G/GroopM/GroopM-{{ version }}.tar.gz
   sha256: 48bdaeb9010e7d81785697a55edc394ef36c2646b310d625a24bec740f1b6080
+  patches:
+    - patch
 
 build:
   noarch: python
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -40,6 +40,9 @@ recipes/detonate
 recipes/pbbam
 recipes/bmtagger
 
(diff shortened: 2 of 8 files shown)


dpryan79:

dpryan79:

recipes/groopm/meta.yaml
```diff
@@ -7,6 +7,8 @@ package:
 source:
   url: https://pypi.io/packages/source/G/GroopM/GroopM-{{ version }}.tar.gz
   sha256: 48bdaeb9010e7d81785697a55edc394ef36c2646b310d625a24bec740f1b6080
+  patches:
+    - patch
```
mbargull: I think we should better use `MPLBACKEND=Agg` in the tests than patching this in, right?

Something like this:
```yaml
test:
  commands:
    - MPLBACKEND=Agg python -c 'import groopm'
    - MPLBACKEND=Agg groopm
```

dpryan79: No, since we need it to run everywhere and it'll fail on most clusters and in containers at the moment.

mbargull: I understand the reasoning, but I don't agree. It's up to the user to set it. Or, in rarer cases, up to the upstream developers to decide. We should only patch it if it's not maintained upstream anymore and there is a *clear intend* to only want the `Agg` backend.

No matter the decision on the above, it most likely makes sense to change the dependency from `matplotlib` to `matplotlib-base` in all cases.

bgruening: A compromise would be to scan the env and if nothing is in there fallback to Agg? This would provide people a way to overwrite our defaults.

Issue #9988, opened 2018-07-19
Repeatmodeler package not working


helltish:
Dear developer Team,

we have compared the bioconda Repeatmodeler package with a local installation of Repeatmodeler (with all dependencies).

The local installation of Repeatmodeler produced an output, which is reasonable for our input file. The conda package does not produce any output. Please find our logs to compare. The input data was the same in both cases.

Thanks a lot in advance!

helltish:
[RepeatModeler_conda.log](https://github.com/bioconda/bioconda-recipes/files/2210248/RepeatModeler_conda.log)
[RepeatModeler_local_installation.log](https://github.com/bioconda/bioconda-recipes/files/2210249/RepeatModeler_local_installation.log)

corburn:
>NOTE: RepeatScout did not return any models.

mbnmbn00:
Exactly same issue here. Manually installed RepeatModeler worked well on the same assembly.
I also noticed that `nseg` which is needed for RepeatScout was not installed.
Manually compiling and locating at `bin` directory didn't help.

`<prefix>.fa.rscons.filtered` file which is one of RepeatScout outputs was empty, too. So I assume that RepeatScout has failed at some point.

astulaaa:
coda Repeatmodeler package still not working properly

jebrosen:
I believe the underlying issue is here:
https://github.com/bioconda/bioconda-recipes/blob/master/recipes/repeatmodeler/build.sh#L23

The RepeatScout package in bioconda does not include `filter-stage-1.prl` in the `bin` directory, but that program is used by RepeatModeler.

Juke34:
I'm working on it with this PR #19088.  I guess it could work like it is in this PR (without nseg) but to be perfect it would be nice to include nseg. But there is currently no recipe for it. nseg is available here ftp://ftp.ncbi.nih.gov/pub/seg/nseg.

Juke34:
In the build3 of RepeatScout nseg and filter-stage-1.prl are now included, but the result seems still the same. How can we check manually which step is failing? Maybe one of this tool do not behaves like they should

jebrosen:
[...]

PR #19137, opened 2019-12-06
Fix repeatmodeler recipe

Juke34:


Juke34:
This change should fix the problem with this recipe.

@BiocondaBot please merge

Juke34:
@BiocondaBot please add label

recipes/repeatmodeler/RepModelConfig.pm
```diff
@@ -104,14 +104,14 @@ BEGIN
...
+  $TRF_DIR = $ENV{'TRF_DIR'};
+  $TRF_PRGM = "$TRF_DIR/trf";
 ##
 ## NSEG Location
 ## =============
 ## Location of the NCBI nseg program
-  $NSEG_PRGM = $ENV{'NSEG_DIR'};
-
+  $NSEG_DIR = $ENV{'NSEG_DIR'};
+  $NSEG_PRGM = "$NSEG_DIR/nseg";
```
marcelm: If you depend on the `nseg` Bioconda package that you created yesterday, then `nseg` will be just installed into `$PATH` and it should not be necessary for this to include `$NSEG_DIR` (or actually to define a `$NSEG_DIR` at all).

recipes/repeatmodeler/meta.yaml
```diff
@@ -7,7 +7,7 @@ source:
   md5: c148b4b56bb3fbd4e9ba88ac8736baa3
 
 build:
-  number: 1
+  number: 2
   skip: True  #[osx]
 
 requirements:
```
marcelm: Did you miss adding the `nseg` dependency? Or did you want to add that in another PR?

marcelm:
Which problem does this PR address? Why do the current tests not catch that problem? Can you add a test that would fail with the current version of the recipe, but would pass with the change proposed in this PR?

recipes/repeatmodeler/meta.yaml
Juke34: I'm not sure it has to be in this recipe because nseg is part of RepeatScout and is run only if RepeatScout is available.

marcelm:
If those tests are hard to write then you should at least verify manually that the patch actually fixes the problem. You can use the `fetch artifacts` command, for example, and then install the revised package locally.

recipes/repeatmodeler/RepModelConfig.pm
Juke34: It's just to stay consistent with how the other paths are defined by the author in this file. And he define them in the `build.sh`. I don't know if it use elsewhere in the code...

marcelm: That said, I don’t know how this setting is used actually.

recipes/repeatmodeler/meta.yaml
marcelm: Ok, I agree it’s not needed here if that is the case. However, then why does the `NSEG_PRGM` configuration thing exist here?

Juke34: I would prefer to keep it like that for clarity

Juke34:
> Which problem does this PR address?

It has been a while this recipe is not working, it is like an empty shell. The tool seems to run but do not perform anything useful. ie (#9988, https://github.com/Dfam-consortium/RepeatModeler/issues/21, https://www.biostars.org/p/368150/, etc...). 

The current test just check actually the main tool compile. But RepeatModeler is a pack of several tools coming with RepatModeler plus several tools from external resource. If one of them fail the pipeline doesn't crash it just skip steps. Depending the crash the output could be empty or not, but we get a file (consensi.fa). The thing is to test the whole pipeline we need to include a data set and run RepeatModeler on it, that could run for hours...

Juke34:

Juke34:

Juke34:
I have added a test now. The test file is 300kb but I cannot make it smaller. The tool needed a fait amount of data to produce a useful output we can check. A least it will avoid to have the problem we got until now where people were trying to use a recipe that was not really working.

recipes/repeatmodeler/run_test.sh
```diff
@@ -0,0 +1,15 @@
...
+set -x -e
+
+# build the db with BuildDatabase
+BuildDatabase -engine ncbi -name db test.fa 2> /dev/null
+
+# run repeat modeler
+RepeatModeler -engine ncbi -database db 2> /dev/null
+
+# check the result
+if [ -s RM_*/consensi.fa ] ; then
```
marcelm: You can replace the entire `if` with `test -s RM_*/consensi.fa` (the exit code of the script is the exit code of the last command; also you have `set -e`, so it’ll exit anyway if `test` fails).

dpryan79:
@bioconda-bot please merge

I'm not overly enthused about a 300kb test file, but it looks like this is the only way currently to prevent regression in this recipe.

oushujun:
Thanks for fixing this. Has this updated to RepeatModeler2 yet?

Juke34:
Not yet... at least not by me.

jebrosen:
I have a bit of a backlog, but updating the RepeatMasker and RepeatModeler recipes is still on my radar.

oushujun:
@Juke34 OK that's good to know. I saw your other [post](https://github.com/bioconda/bioconda-recipes/pull/19805#issuecomment-575790244) (#19805), `RECON` and `RepeatScout` is still not available to OSX yet. @jebrosen Are these also on your radar? Thanks guys!

Juke34:
Done

PR #19219, opened 2019-12-11
MaRaCluster: new recipe

bernt-matthias:

### DIFF
--- a/recipes/maracluster/build.sh
+++ b/recipes/maracluster/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+mkdir -p build/maracluster
+cd build/maracluster
+
+cmake -DTARGET_ARCH=amd64 -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="${PREFIX}" -DCMAKE_PREFIX_PATH="${SRC_DIR}/build/tools" "${SRC_DIR}";
+make -j 4;
+make -j 4 package;
+make install
--- a/recipes/maracluster/meta.yaml
+++ b/recipes/maracluster/meta.yaml
@@ -1,0 +1,34 @@
+{% set name = "maracluster" %}
+{% set version = "1.00.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/statisticalbiotechnology/maracluster/archive/rel-1-00-1.tar.gz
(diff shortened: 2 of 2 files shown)


bernt-matthias:
ping @glormph

bgruening:
Looks like you need to set PWIZ_DATA_MSDATA_LIBRARY ?

jpfeuffer:
And looks like the pwiz recipe only packages the binaries not its includes and libraries, could that be?
https://github.com/bioconda/bioconda-recipes/blob/master/recipes/proteowizard/meta.yaml#L13

bernt-matthias:
> And looks like the pwiz recipe only packages the binaries not its includes and libraries, could that be?

Indeed. That's what I remember. There was an issue suggesting that the way to go would be extending the pwiz docker container. But I was never sure about this .. main problem for packaging in bioconda would be the vendor libraries I guess.

bernt-matthias:
But maybe having maracluster supporting the main formats (e.g. mzML) is sufficient .. given that there are tools for converting to them.

jpfeuffer:
Yes I think that would be enough, especially with the adapter functionality in OpenMS.
Did you try the corresponding CMake variables?
https://github.com/statisticalbiotechnology/maracluster/blob/master/CMakeLists.txt#L40

bernt-matthias:
Found the discussion https://github.com/bioconda/bioconda-recipes/pull/2432 (https://github.com/bioconda/bioconda-recipes/pull/19635)

jpfeuffer:
I agree with building it without vendor support. If you are using bioconda, you probably should have converted your files to mzML already on your windows machine anyway. Or use ThermoRawFileParser if it is thermo raw.

bernt-matthias:
@jpfeuffer do you know which version is supported by OpenMS? Latest seems to be [1.01.1](https://github.com/statisticalbiotechnology/maracluster/releases/tag/rel-1-01-1)

but the `THIRDPARTY` dir contains 0.05.0

jpfeuffer:
Hmm I don't know exactly. If the command line did not change much and the output format stayed it should be fine. If you have time, you can run the OpenMS test of the adapter with the new version and check.

bernt-matthias:
Doesn't work without pwiz headers, etc (which are not provided by the current pwiz recipe) since MaRaCluster also uses pwiz if vendor support is disabled.

ref https://github.com/ProteoWizard/pwiz/issues/2031

PR #19220, opened 2019-12-11
Specify gsl version

RomainFeron:
PhyloCSF requires libgsl.so.23, which is not available with the last version of gsl. 
I specified gsl version 2.5 to make sure this lib is available.
For some reason it doesn't build on osx anymore and the error is too complex for me to handle at the moment, so I disabled building for osx.

### DIFF
--- a/recipes/phylocsf/meta.yaml
+++ b/recipes/phylocsf/meta.yaml
@@ -6,7 +6,6 @@ package:
 
 build:
   number: 1
-  skip: True  # [osx]
 
 source:
   url: https://github.com/RomainFeron/PhyloCSF/archive/{{ version }}-conda.tar.gz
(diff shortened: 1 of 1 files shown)


dpryan79:
I suspect the issue was that when this was built originally it wasn't pinned to a gsl version. We currently pin to gsl 2.5, so moving gsl to the host section should take care of the issue.

RomainFeron:
Okay, I forgot about build vs host, it's more clear now. Thanks !
The osx-specific bug is different; one component of opam (ocaml package manager) does not build anymore, the error is:

> Fatal error: exception ("Fdecl.get: not set", {})
> 01:00:40 BIOCONDA INFO (ERR) # Raised at file "src/stdune/code_error.ml", line 9, characters 30-62

It used to work 2-3 months ago when I first built the recipe, and I couldn't find anything to help with this error after a quick search. I'm not familiar with ocaml at all, I'm not sure how to fix this at the moment.

dpryan79:
Ah, that explains it, I know nothing of opam, so let's just skip that :)

dpryan79:
@bioconda-bot please fetch

dpryan79:

dpryan79:
@bioconda-bot please merge

Issue #19200, opened 2019-12-10
pyopenms/openms restructuring


jpfeuffer:
Hi everyone!

I followed the recent integration of pyOpenMS and the changes to OpenMS a little.
I started to clean up the recipes locally a bit and I was wondering if it wouldn't be
possible to just split a general OpenMS recipe into multiple subpackages
via the "output" section. Since all the parts should have the same versions of dependencies
and sources anyway I think that would make sense (if it is supported in bioconda). Also it
is much less trouble to build them together. Or was it very important to reduce the build times
here?

We could have multiple subpackages:
- libopenms
- openms-tools
- (openms-gui) # if we can make it work
- pyopenms

One question that I would have then is how to specify the "outputs" for pyopenms.
We build it with distutils based on our CMake output but what kind of files should go where
for a python package. In the current recipe it seems like it just "pip install"s the wheel.

bgruening:
All I can say is that one recipe with multiple outputs is the preferred way. I raised this in the initial reivew and people agreed that this is the long-term strategy.

jpfeuffer:
Ok great. Then I am going to follow this path. My first attempts were promising.
pyOpenMS is a bit tricky since the wheel of course comes with all libs etc. bundled. 
I assume the conda way would be to reference libs installed by the future libopenms package right?
Does anyone know if the activate scripts act transitively, i.e. if I set the OPENMS_DATA_PATH in the libopenms package and pyOpenMS requires it, will it be loaded as well during activation?

bernt-matthias:
This seems to be done now, or?

PR #19340, opened 2019-12-17
[FEATURE] Restructure Openms related recipes

jpfeuffer:
This splits openms related packages into subpackages including pyopenms which would then be superseded by this metarecipe.

While fixing this recipe I added the fixes to OpenMS' develop whenever possible, so after the next release it should be much easier to build OpenMS with conda. We still might need one or two patches for pyopenms, such that e.g. the sysroot is picked up (if I find time, I will implement something to pass on the additional CMake variables)

Things to do:
- I don't know why conda still complains about missing specification of dependencies. It seems like the subpackages do not really pick up the run_exports from libopenms
- the pyopenms pip wheel still packages all dependencies, since this is what we want to have in the wheel usually. Of course this is not optimal for conda, since it occupies additional space and it is hard to check which dependency is actually used by the python C++ extension. The same holds for the share and example data (which is duplicated).
-------------------


jpfeuffer:
Possible TODOs here or in future PRs:
- Activate multiple python versions
- Check if pyopenms uses the correct libraries and installs correctly
- Add GUI tools (we will probably need to work around [this](https://github.com/ContinuumIO/anaconda-issues/issues/8779) )
- Split lib to lib and lib-devel (headers)
- Add docu subpackage

jpfeuffer:
supersedes #18776
related to #18701 
fixes #19200

jpfeuffer:
What does conda do for so long after the build??

jpfeuffer:
@BiocondaBot please add label

dpryan79:
The explicit `run_exports` probably aren't needed, but they also won't hurt anything. I removed some excess `{{ python }}` (python is always pinned) and will merge once the CI is green.

dpryan79:
@bioconda-bot please merge

@jpfeuffer Thanks for the huge amount of work that this must have taken!

jpfeuffer:
Quick follow-up on the python version:
So bioconda always uses one specific version of python only?
I added the variable to be able to "quickly" generate multiple versions of pyopenms though the
recipe based conda_build_config.yaml

dpryan79:
At the moment bioconda is building against 3 different python versions (2.7, 3.6 and 3.7). I hadn't noticed the conda_build_config, we can get rid of that and specify only building against python 3.6 in case pyopenms doesn't work with 3.7.

jpfeuffer:
Okay, no I just wanted to leave the option open. It _should_ work with any. But it is also pretty heavy on the build time, that is why I did only one version for now.

dpryan79:
Yeah, the build time is a very good thing to keep in mind. We're limited to ~5 hours at the moment, so we can probably still swing 2 python builds. I'll try to remember to do this once our current OSX CI issues are fixed.

PR #19518, opened 2020-01-03
Use newer macOS image on CircleCI

marcomorain:
* [ ] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/contributor/guidelines.html).

We are going to be removing support for Xcode 8.3.3 soon. This project is currently the biggest user of that image. Upgrading to the latest Xcode image will ensure that your build is stable for as long as possible.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ variables:
       command: .circleci/setup.sh
   macos: &macos
     macos:
-      xcode: 11.3.0
+      xcode: 9.4.1
     environment:
       LC_ALL: en_US.UTF-8
       LANG: en_US.UTF-8
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,7 +22,7 @@ variables:
       command: .circleci/setup.sh
   macos: &macos
     macos:
-      xcode: "8.3.3"
+      xcode: 11.3.0
     environment:
       LC_ALL: en_US.UTF-8
       LANG: en_US.UTF-8


dpryan79:
@bioconda/core Does anyone know what image conda-forge is using? I presume we should aim for that for compatibility.

mbargull:
I'm not sure if this is still the current way it's done and whether there are other setup steps, but we have https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/a1026adb523b6562c16329170e7e304a25ed4033/recipe/run_conda_forge_build_setup_osx#L10-L22, i.e., by default `https://github.com/phracker/MacOSX-SDKs/releases/download/10.13/MacOSX10.9.sdk.tar.xz` would be used (deployment target is set at https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/3ddd7cee1b34dd47aaec8a0890917f9c30ed288e/recipe/conda_build_config.yaml#L89).

dpryan79:
@mbargull Can we just directly use conda-forge-ci-setup on circleci or do you think that would muck other things up (I don't know what all it sets).

mbargull:
On macOS it should be fine to
```sh
# Pinned to 2.5.3 to make sure we don't get unexpected changes.
conda install conda-forge-ci-setup=2.5.3
# "CONFIG=" to avoid
# https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/a1026adb523b6562c16329170e7e304a25ed4033/recipe/run_conda_forge_build_setup_osx#L37-L50
# which we don't need/use.
CONFIG= OSX_FORCE_SDK_DOWNLOAD=1 run_conda_forge_build_setup
# Reactivate environment so that conda-forge-ci-setup-activate.sh is sourced.
conda activate
```
as far as I can tell.

On Linux maybe not, unless we would remove the activation script line from https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/a1026adb523b6562c16329170e7e304a25ed4033/recipe/run_conda_forge_build_setup_linux#L35 afterwards since that line sets `CONDA_BLD_PATH` to some semi-fixed value from https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/a1026adb523b6562c16329170e7e304a25ed4033/recipe/run_conda_forge_build_setup_linux#L9.

mbargull:
Having taken another looks at cf's macOS setup:
https://github.com/conda-forge/conda-smithy/blob/88b82bd151b4c7ea0fa4d4661fc0a70d8db27d70/conda_smithy/templates/azure-pipelines-osx.yml.tmpl#L8
https://github.com/conda-forge/conda-smithy/blob/88b82bd151b4c7ea0fa4d4661fc0a70d8db27d70/conda_smithy/templates/circle.yml.tmpl#L25

=> looks like `macOS 10.13`/`Xcode 9.*` is used as the base in any case, not matter the explicit SDK download.
=> maybe we should change to that one, too. If there are no other significant changes, `9.4.1` instead of `9.0` might make sense.

@marcomorain: Do you know if there are any plans for when CircleCI will drop the Xcode 9 series? And in any case: Thanks for bringing attention about the CircleCI macOS image changes to us via this PR, very much appreciated!

PR #19790, opened 2020-01-17
Updating expired tbl2asn

almahmoud:
Updated hash for new tbl2asn.

----

### DIFF
--- a/recipes/tbl2asn/meta.yaml
+++ b/recipes/tbl2asn/meta.yaml
@@ -1,12 +1,12 @@
 package:
   name: tbl2asn
-  version: "25.7"
+  version: "25.8"
 
 source:
   - url: ftp://ftp.ncbi.nih.gov/toolbox/ncbi_tools/converters/by_program/tbl2asn/mac.tbl2asn.gz  # [osx]
-    sha256: 8dab2f4ff804ad0678de1eebbdd4a3515b0740f5f2116bf91d55cc808f520cbc  # [osx]
+    sha256: d326f751784fa36b2eb46bbdb749fb17fe4ca0a5d3c859ac57397ba5b724a35a  # [osx]
   - url: ftp://ftp.ncbi.nih.gov/toolbox/ncbi_tools/converters/by_program/tbl2asn/linux.tbl2asn.gz  # [linux]
-    sha256: 1f41acb9eacd6e0fa78657b17727b8f94d97490617563686d28b3f6c24d39117  # [linux]
+    sha256: 05b4cec586fab9f07926413e61bed718ab401242859ebd5e8b9b11fbd6e6c203  # [linux]
 
 build:
   number: 0


bgruening:
`07:19:18 BIOCONDA INFO (OUT) tbl2asn: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by tbl2asn)` we are getting now really serious problems. Either be build if from source now or we need to patch this update-constrained out :(

bgruening:
xref: https://github.com/tseemann/prokka/issues/444

bgruening:
But the really bad part is that they __also__ stop supporting CentOS 7.

almahmoud:
I see that error here, but building it locally worked for me, although it’s OS X and then I converted from that

bgruening:
Locally means on your host not in the CentOS Docker container isn't it?

peterjc:
Oh boy, this problem is back - and worse in that the latest NCBI provided binaries are non-portable.

If we knew the right people to pressure within the NCBI, getting them to either fix the binary compatibility or better drop the auto-expiring code would be good. I'll try emailing a couple of people who might know.

If we have simple access to the source code, I think patching out the check and building our own binaries is best (see linked thread on prokka).

If we don't have easy access to the latest source code, I would personally be happy to patch the provided binary - but that's a bit outside my skill set, and doesn't solve the binary compatibility problem with the latest releases.

Even a non-expiring old release would work for many people - the exact opposite of what the authors probably intended by added this auto-expiring code 🙄

almahmoud:
@bgruening yes, not in the container. Sorry was replying on my phone and had only seen the first comment

tseemann:
See the `patchelf` and string replacements done in Brew:
https://github.com/brewsci/homebrew-bio/blob/e36583930b5ccd0a4a6ed64e6f3bf654b4977705/Formula/tbl2asn.rb

standage:
Commenting to follow this thread. This just made Prokka unusable on my systems.

tseemann:
@bgruening i use RHEL 7.x and the `tbl2asn 25.8` binary works on my system:

```
./tbl2asn  - | grep argument
tbl2asn 25.8   arguments:
```

Here is what it links against:
```
ldd ./tbl2asn
        linux-vdso.so.1 =>  (0x00007ffcbc985000)
        libidn.so.11 => /lib64/libidn.so.11 (0x00007f1062a48000)
        libz.so.1 => /lib64/libz.so.1 (0x00007f1062832000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f1062530000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f1062162000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f1062c7b000)
```

rpetit3:
@EricDeveaud dug up the source for v25.3 (available at ftp://ftp.ncbi.nih.gov/toolbox/ncbi_tools/old/20170106/ncbi.tar.gz) and proposed a patch to remove the expiration. See: https://github.com/tseemann/prokka/issues/444#issuecomment-561274989

Maybe this could be a special version with a separate build.sh and meta.yml, since it would involve building from source.

peterjc:
I think we can close this old PR with the #20073 `tbl2asn-forever` workaround?

Issue #19844, opened 2020-01-20
Microassembler missing in latest Scalpel recipe


guillaumecharbonnier:
Using latest conda recipe for `scalpel`, 
`scalpel-discovery  --single` fails because `Microassembler` is not available where it is looking for:
```
$PREFIX/share/scalpel-0.5.4-1/Microassembler/Microassembler: No such file or directory
```

Mentionning recipe contributors:
@Luobiny @chapmanb @epruesse @dpryan79

dpryan79:
It looks like the Microassembler source code was installed rather than being compiled first. I really don't know why microassembler isn't getting compiled, it's explicitly mentioned in the Makefile.

chapmanb:
Thanks for the report. I took a look at this and it appears to be a compilation problem due to changes in the environmental variables for conda builds. It no longer finds the bamtools include files and fails. I have a fix and am testing right now. Will try to upload first thing tomorrow, thanks again for identifying this.

chapmanb:
Thanks again for reporting this problem. We've fixed it in the latest build (0.5.4_2) and hopefully all will work cleanly now. Please let us know if you run into more issues.

eeleclair:
Hello, I am having the same error and found this page (thank you, Google).  But where can I get the **0.5.4_2** version of Scalpel?    Sourceforge has only **0.5.4** dated 1/11/2018.

YuanfengZhang:
> Hello, I am having the same error and found this page (thank you, Google). But where can I get the **0.5.4_2** version of Scalpel? Sourceforge has only **0.5.4** dated 1/11/2018.

Watch the resource they offer. This is a conda package, and requires samtools, bamtools and sereval other pkgs. So the best practice I propose is to create a new conda env. You can copy following cmdlns to run in shell on Linux system if you are a new hand:
`conda create -n somatic`

`conda install -c bioconda scalpel -y`

When you want to run scalpel, 
`conda activate somatic`
and scalpel-discovery will be avaliable.

Simply type `scalpel-discovery` and you shall see message like below:

> Local date and time: Tue Oct 12 18:46:09 2021
> 
> Program: scalpel-discovery (micro-assembly variant detection)
> Version: 0.5.4 (beta), January 11 2018
> Contact: Giuseppe Narzisi <gnarzisi@nygenome.org>
> 
> usage: scalpel-discovery <COMMAND> [OPTIONS]
> 
> COMMAND:
> 
>     --help    : this (help) message
[...]

PR #19868, opened 2020-01-21
scalpel: fix build for Microassembler

chapmanb:
Microassembler was not correctly building due to not finding bamtools
include files and libaries. Adjust to BUILD_PREFIX reference to enable.
Fixes bioconda/bioconda-recipes#19844

### DIFF
--- a/recipes/scalpel/build.sh
+++ b/recipes/scalpel/build.sh
@@ -22,11 +22,13 @@ sed -i "s/bamtools samtools bcftools//g" Makefile
 chmod 0755 FindVariants.pl
 chmod 0755 FindSomatic.pl
 
-make INCLUDES="-I$PREFIX/include/bamtools -L $PREFIX/lib" Microassembler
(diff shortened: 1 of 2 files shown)


dpryan79:

dpryan79:
Microassembler isn't getting compiled (or at least I don't see it in the tarball) and the source code from darn near everything is getting dumped under `share`.

chapmanb:

chapmanb:
Devon;
Thanks for checking this and apologies about it still being missing. The fetch artifacts trick is so useful, thank you for pointing that out. It was building locally for me but not on CI due to hardcoded g++'s in the Makefile. I'm not sure why the build passed with this error, but it's fixed now and they're in there. We do purposely copy over the source directory since it has a bunch of necessary perl code and is the recommended way to run it. Since it's not too big, the build script just copies it all over. It does help a lot for debugging in cases like this, so is pretty useful to have. Let me know if you spot any other problems.

dpryan79:
@bioconda-bot please merge

PR #19969, opened 2020-01-25
snp-pileup

dariober:
Dummy update of recipe to test libcrypto.so.1.0.0 

Try to rebuild the package to see if we can reproduce/fix the issue with
libcrypto.so.1.0.0 described at
dariober/cnv_facets#14 and also appearing at https://circleci.com/gh/bioconda/bioconda-recipes/92111?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

----

>
>
>
>

### DIFF
--- a/recipes/cnv_facets/meta.yaml
+++ b/recipes/cnv_facets/meta.yaml
@@ -16,6 +16,8 @@ build:
 requirements:
   build:
       - {{ compiler('c') }}
+      - {{ compiler('cxx') }}
+      - {{ compiler('fortran') }}
   host:
       - r-base
       - r-devtools
--- a/recipes/cnv_facets/meta.yaml
+++ b/recipes/cnv_facets/meta.yaml
@@ -23,7 +23,6 @@ requirements:
       - r-ggplot2
       - r-gridextra
       - r-jsonlite
-      - openssl
       - bioconductor-rsamtools
       - htslib
       - samtools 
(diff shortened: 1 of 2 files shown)


dpryan79:
I think the solver is just doing the wrong thing here.

dpryan79:
Adding htslib as a dependency prevents the solver from screwing things up, though it's really annoying that that's needed.

dpryan79:
I suspect python is just a run-time dependency, so it can just be added there (that will also cut the build time by 67%).

dpryan79:

Jumping the gun, but this will allow me to check if this needs to have python-specific builds

dpryan79:

dpryan79:
@dariober This should fix the issues people have reported.

dariober:
Thanks Devon! I see you have done a lot of stuff here - much appreciated all you help on bioconda and a million other things! I see the solution was to add htslib on the dependent recipe. If you get a chance, could you comment on why/what it has happened?

dpryan79:
The solution is basically a hack around the conda solver doing the wrong thing. Basically the older build of snp-pipelup wasn't pinning the version of openssl it was compiled against because it was made before that was done on any package. For some terrible reason, the solver really likes to minimize dependency restrictions rather than maximize version and build numbers. Within bioconda all packages are pinned to specific versions of htslib, which then pinned to a specific version of openssl. By specifying htslib explicitly we ensure that the conda solver cannot prefer older builds with looser restrictions.

Long-term we need to update the meta information on old builds like this so they're properly pinned to openssl, but that will be a huge amount of work for which I don't think anyone has time.

PR #19767, opened 2020-01-16
Update tobias to 0.10.1

BiocondaBot:
Update [`tobias`](https://bioconda.github.io/recipes/tobias/README.html): **0.9.0** &rarr; **0.10.1**

[install with bioconda](http://bioconda.github.io/recipes/tobias/README.html) [Conda](https://anaconda.org/bioconda/tobias/files)

Info | Link
-----|-----
Recipe | [`recipes/tobias`](https://github.com//bioconda/bioconda-recipes/tree/bump/tobias/recipes/tobias) (click to view/edit other files)
Releases |[https://pypi.org/pypi/tobias/json](https://pypi.org/pypi/tobias/json)
Recipe Maintainer(s) | @msbentsen, @jenzopr

***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

Issue #19996, opened 2020-01-28
Replace matplotlib dependency with matplotlib-base in most recipes


marcelm:
The `matplotlib` package has been split into a `matplotlib` and `matplotlib-base` package a while ago, see https://github.com/conda-forge/matplotlib-feedstock/pull/178 . The only difference is that `matplotlib-base` does not depend on `pyqt`, reducing installation size significantly. Unless Matplotlib’s PyQt backend is explicitly needed by a package, depending on `matplotlib-base` is sufficient.

To be clear: Interactive plotting is still possible even with only `matplotlib-base` because the TkAgg backend is still available. It actually looks very similar to the PyQt backend.

Also, Matplotlib 3+ has a backend fallback mechanism. For interactive plotting, it prefers PyQt, but falls back to TkAgg. Nothing needs to be configured. If the user wants to use PyQt (or PySide or similar), they simply need to install that package, and they’ll get PyQt instead of TkAgg automatically.

Around 230 packages are affected.

```bash
$ egrep -l " *- matplotlib(\s|$)" recipes/*/meta.yaml | wc -l
230
```

cf #19985

bgruening:
@marcelm Thanks a lot for pushing this! Maybe a good first step is to enhance our linter to warn about the matplotlib usage and recommend the -base version?

marcelm:
We’re currently at around 100 packages that still depend on `matplotlib`. I’m closing this now because the linter check is in place and because I don’t think anyone wants to manually go through the remaining recipes and update them. They have obviously not been updated in the last 1.5 years, so most of these will be for software that is not used so much anyway.

PR #20014, opened 2020-01-29
Fix to tobias 0.10.1 update

msbentsen:
Trying to resolve the xgboost dependency issues of #19767.

----

>
>
>
>

### DIFF
--- a/recipes/tobias/meta.yaml
+++ b/recipes/tobias/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "tobias" %}
-{% set version = "0.9.0" %}
+{% set version = "0.10.1" %}
 
 package:
   name: {{ name|lower }}
@@ -7,11 +7,12 @@ package:
 
 source:
   url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
-  sha256: addd4c981fdad316be594d3a5814c265fb0f95aa8b829098546738cc7774c0e9
+  sha256: 03dbd2e31715bb402039222dbf71c3b83405a3fd095c0f8368942c1e39a07536
 
 build:
   number: 0
-  skip: True # [py27]
+  # gimmemotif is currently on built for python 3.6
+  skip: True  # [py != 36]
 
 requirements:
   build:
@@ -36,6 +37,9 @@ requirements:
     - pybedtools
     - adjusttext
     - svist4get
+    - gimmemotifs >=0.14
+    - imagemagick
+    - py-xgboost
 
 test:
   commands:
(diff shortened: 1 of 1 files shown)


msbentsen:
This is odd... xgboost is pulled as a dependency (0.90-py37he1b5a44_4, conda-forge), but afterwards it is not known to python, which then states `'xgboost>=0.71' distribution was not found`.

I see a similar issue raised by #19299, but it was never solved (or was solved by removing xgboost as dependency). I can't tell at which end the issue is. Help is much appreciated! @jenzopr?

marcelm:
Hi, while you’re updating this recipe anyway, could you please change the `matplotlib` dependency to `matplotlib-base`? See #19996 .

I also looked at your build problem, but I don’t have a clue at the moment.

marcelm:
One idea for debugging: You could temporarily add `python -c "import xdgboost"` to the test commands and see whether that works.

marcelm:
First, I think now that the correct dependency to use is `py-xdgboost`. If you specify `xdgboost`, that will pull in `py-xdgboost` as well and it’ll work, but since you only need the Python module, `py-xdgboost` should be enough. (Both pull in `libxdgboost`.)

I can reproduce the problem locally: Creating a new environment with `conda create -n xdgboost py-xdgboost` and then running `pip freeze` in that environment will not list `xdgboost`. The problem appears to be that the `py-xdgboost` package does not contain the `.dist-info` directory that contains the meta information that pip uses to find out about installed packages. This should probably be reported at https://github.com/conda-forge/xgboost-feedstock .

msbentsen:
Thank you so much, @marcelm! I hope we can get it solved through an update of the xgboost conda-forge package as you suggested. I am still wondering how the 'gimmemotif' bioconda package managed to build without errors though (since it has the same dependency on xgboost). Maybe the this is a special case since it is a dependency of a dependency? Anyway, thanks :-)

marcelm:
> I am still wondering how the 'gimmemotif' bioconda package managed to build without errors

I think the difference is the way the command-line scripts are installed.

gimmemotifs [uses the `scripts` argument in its `setup.py` file](https://github.com/vanheeringen-lab/gimmemotifs/blob/3f76ea95220c32f8cb126553ef2ccd16e17cc7dd/setup.py#L102). This will just copy over the file during installation almost unmodified (except for adjusting the shebang).

TOBIAS [uses the setuptoos `entry_points` feature in its `setup.py`](https://github.com/loosolab/TOBIAS/blob/61ad7418483b2ba1c67ca4f49490da93081fd712/setup.py#L61). With this, `setuptools` generates the entire script automatically, and this generated script contains a version check.

msbentsen:
Note: Temporarily "fixed" by removing 'gimmemotifs' as a dependency for TOBIAS with #20643. Ideally though, I would like to bring it back in once the dependency issues are solved.

PR #20024, opened 2020-01-30
Add Tadbit

david-castillo:
Add recipe for TADbit, a python library to analyze, model and explore 3C-based data.

----

>
>
>
>

### DIFF
--- a/recipes/tadbit/build.sh
+++ b/recipes/tadbit/build.sh
@@ -1,0 +1,1 @@
+$PYTHON setup.py install
--- a/recipes/tadbit/build.sh
+++ b/recipes/tadbit/build.sh
@@ -1,1 +1,0 @@
-$PYTHON setup.py install --force
--- a/recipes/tadbit/build.sh
+++ b/recipes/tadbit/build.sh
@@ -1,1 +1,1 @@
-$PYTHON setup.py install
+$PYTHON setup.py install --force
--- a/recipes/tadbit/meta.yaml
+++ b/recipes/tadbit/meta.yaml
@@ -6,12 +6,12 @@ package:
 
 source:
   url: https://github.com/3DGenomes/TADbit/archive/v{{ version }}.tar.gz
-  sha256: c8f911eff5b7bf73cdc13c3720375baed6128be786576db77aaf627c6749d79c
+  sha256: 11f2b0d6bdab2f85c5d87fe813008fa19b079e11a12c7ed86ce4699329dc1db5
   
 build:
   number: 0
   skip: True [py3k]
-  script: {{ PYTHON }} -m pip install . --ignore-installed --no-deps -vv
+  script: {{ PYTHON }} -m pip install . --install-option="--bypasscheck" --ignore-installed --no-deps -vv
 
 requirements:
   build:
--- a/recipes/tadbit/meta.yaml
+++ b/recipes/tadbit/meta.yaml
@@ -21,25 +21,15 @@ requirements:
     - python <3
     - scipy
     - numpy
-    - matplotlib
+    - matplotlib-base
     - h5py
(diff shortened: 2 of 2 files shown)


david-castillo:
@BiocondaBot please add label

recipes/tadbit/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - tadbit
+about:
+  home: http://sgt.cnag.cat/3dg/tadbit/
+  license: GPL-3.0
+  license_file: LICENSE.txt
+  summary: "TADbit is a complete Python library to deal with all steps to analyze, model and explore 3C-based data."
+
+extra:
+  identifiers:
+    - doi:10.1371/journal.pcbi.1005665
```
marcelm: This is missing a newline at the end of the file

recipes/tadbit/build.sh
```diff
@@ -0,0 +1 @@
+$PYTHON setup.py install --force
```
marcelm: Since it’s just a single line, you can move this into the `meta.yaml` (as `script:` under the `build:` section). Also, won’t the standard `"{{ PYTHON }} -m pip install . --no-deps --ignore-installed -vv"` work here?

marcelm: I see now that you this is the first recipe you’re contributing, so I’m trying to find where this is written in the Bioconda  documentation. It seems the docs now recommend using `conda skeleton pypi`, which producese a `meta.yaml` without the `--no-deps` etc. part. Previously, the docs would the line I wrote above, and many recipes use i, so I don’t know what the curently recommend approach is. I’ll have to defer to someone who knows more.

recipes/tadbit/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+  host:
+    - python <3
+  run:
+    - python <3
+    - scipy
+    - numpy
+    - matplotlib-base
```
marcelm: And thanks for using `matplotlib-base` here :-)

recipes/tadbit/build.sh
david-castillo: We didn't go to pip, just directly to conda recipe. Is that a problem?

david-castillo: Maybe I should go for pip first ...

marcelm: Ignore the stuff about conda skeleton. It doesn’t matter how you created the recipe. Just remove the `build.sh` file and add this to your build section:
```
  script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed -vv"
```

dpryan79: I think we need to update the docs to include `{{ PYTHON }} -m pip install . --no-deps -vv` at least. I'll add that to the core agenda to add that as a point of discussion.

david-castillo: Then I'll need to change my setup.py. We originally had a raw input for the user to confirm installation even if a dependency is not present. That blocks the bioconda install so I include a --force argument to bypass that.
But with pip I'll have to change the way the argument --force is passed to the setup.py.

marcelm: Just add `mcl` to your `host:` requirements. Then it’ll be found at `pip install` time.

(But I would also recommend reworking your `setup.py`, it looks quite complicated.)

david-castillo: Main problem is for IMP in the salilab conda channel. That cannot be added to the host.
I'll work on my setup.py ... tonight

dpryan79:

dpryan79:
@bioconda-bot please merge

PR #20077, opened 2020-01-31
Update antiSMASH to 5.1.1

apcamargo:


apcamargo:
@druvus is the last atempt to update antiSMASH to v5 you mentioned that MEME version 4.11.2 patch 2 introduced Python 3 support. Do you remember where you read that? I just checked the code and it seems like it is Python 2.

apcamargo:
An option would be to create a new MEME recipe with just the binaries, without the scripts. However, I don't know if this is an appropriate solution.

bgruening:
> An option would be to create a new MEME recipe with just the binaries, without the scripts. However, I don't know if this is an appropriate solution.

Can we run 2to3 on meme?

apcamargo:
Do you think it is a good idea to distribute a modified version of a program? Maybe if we create a separate recipe for that.

apcamargo:
The recipe is finally building. @kpalin, do you have any suggestion regarding this PR?

apcamargo:
The recipe is finally building. @kblin, do you have any suggestions regarding this PR?

dpryan79:
I've approved this so you can merge it in when you're happy with it. Thanks!

recipes/antismash/meta.yaml
```diff
@@ -1,64 +1,44 @@
...
   version: {{ version }}
 
 build:
+  number: 0
   noarch: python
-  number: 2
 
 source:
-  url: https://dl.secondarymetabolites.org/releases/{{ version }}/antismash-{{ version }}.tar.gz
+  url: https://github.com/antismash/antismash/archive/{{version.replace(".", "-")}}.tar.gz
```
kblin: I'd prefer if you would continue to get the releases from the download server. That's the tarball we use for all of our other installs, and while they of course should be in sync, I'd like to avoid the "surprise mechanics" in support requests if they should ever be out of sync.

kblin:
Awesome, thank you!

dpryan79:
@bioconda-bot please merge

dpryan79:

dpryan79:
@bioconda-bot please merge

apcamargo:
Thanks for the review @kblin! I had changed the URL to GitHub so that new versions could be automatically detected by BiocondaBot. But I agree that keeping consistency should be a priority.

PR #20087, opened 2020-02-02
Use 2to3 to make MEME 4.11.2 compatible with Python 3

apcamargo:

I used 2to3 to make MEME 4.11.2 compatible with Python 3. This is necessary to use it with antiSMASH 5 (https://github.com/bioconda/bioconda-recipes/pull/20077).

### DIFF
--- a/recipes/meme/4.11.2/2to3.patch
+++ b/recipes/meme/4.11.2/2to3.patch
@@ -1,0 +1,437 @@
+--- ./doc/examples/sample_opal_scripts/FimoClient.py	(original)
++++ ./doc/examples/sample_opal_scripts/FimoClient.py	(refactored)
+@@ -13,7 +13,7 @@
+ from os import mkdir
+ from subprocess import call
+ from time import sleep
+-from urlparse import urlparse
++from urllib.parse import urlparse
+ from ZSI.TC import String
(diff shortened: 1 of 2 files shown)


apcamargo:
@bgruening Here's the 2to3 conversion you suggested. What do you think?

bgruening:

bgruening:
That looks good to me, can you test the artifacts, please?

dpryan79:
You'll want to enable builds for python 3.

marcelm:
Shouldn’t running `2to3` be done as a last resort only? This way, it essentially becomes a fork that needs to be maintained. Isn’t there a way to get upstream to update their package?

marcelm:
At least run `2to3` during the build (so that the patch file isn’t needed).

apcamargo:
@dpryan79 How do I do that? I thought that skipping py27 was enough.

@marcelm I agree with you that this solution is not ideal, but as this is an old MEME version, I don't think the developers will update it to Python 3.
Do you think that running `2to3` during the build is better? I chose to apply the patch to make the changes more explicit.

dpryan79:

dpryan79:
@bioconda-bot please merge

PR #20092, opened 2020-02-02
NGLess: 1.0.1 -> 1.1.0 (WIP)

luispedro:
Attempt a new way to build, using the extended containers

This worked locally, but want to see how it performs on CI.

### DIFF
--- a/recipes/ngless/build.sh
+++ b/recipes/ngless/build.sh
@@ -1,7 +1,18 @@
 #!/usr/bin/env bash
 
 
-mkdir -p ${PREFIX}/bin
-mv ngless-1.0.1-static-Linux64 ngless
-chmod +x ngless
-mv ngless ${PREFIX}/bin
+export LDFLAGS="-L${PREFIX}/lib"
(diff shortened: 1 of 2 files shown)


npavlovikj:
The error is:
```
13:20:40 BIOCONDA INFO (OUT) No setup information found for ghc-8.6.5 on your platform.
13:20:40 BIOCONDA INFO (OUT) This probably means a GHC bindist has not yet been added for OS key 'linux64-gmp4'.
13:20:40 BIOCONDA INFO (OUT) Supported versions: ghc-7.8.4, ghc-7.10.2, ghc-7.10.3, ghc-8.0.1, ghc-8.0.2, ghc-8.2.1, ghc-8.2.2, ghc-8.4.2
```

npavlovikj:
I wonder if `ghc` should be added as `host` dependency.

luispedro:
The problem is that that version of `ghc` is too old and the newer versions are not packaged for CentOS

Closing here in favour of https://github.com/bioconda/bioconda-recipes/pull/20486 (which is unsatisfactory in many ways, but working)

PR #20339, opened 2020-02-17
paleomix recipe

gtrichard:
Paleomix pipelines recipe: https://github.com/MikkelSchubert/paleomix

----

>
>
>
>

### DIFF
--- a/recipes/paleomix/LICENSE
+++ b/recipes/paleomix/LICENSE
@@ -1,0 +1,18 @@
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/recipes/paleomix/build.sh
+++ b/recipes/paleomix/build.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-cp paleomix/main.py $PREFIX/bin/.
+cp paleomix/main.py $PREFIX/bin/main.py
 cp -r paleomix $PREFIX/lib/python2.7/site-packages/paleomix
--- a/recipes/paleomix/build.sh
+++ b/recipes/paleomix/build.sh
@@ -1,5 +1,0 @@
-#!/bin/bash
-mkdir -p ${PREFIX}/bin
-mkdir -p ${PREFIX}/lib/python2.7/site-packages/paleomix
-cp paleomix/main.py $PREFIX/bin/main.py
-cp -r paleomix $PREFIX/lib/python2.7/site-packages/
--- a/recipes/paleomix/build.sh
+++ b/recipes/paleomix/build.sh
@@ -1,0 +1,2 @@
+cp ${RECIPE_DIR}/paleomix/main.py $PREFIX/bin/.
+cp -r ${RECIPE_DIR}/paleomix $PREFIX/lib/python2.7/site-packages/paleomix
--- a/recipes/paleomix/build.sh
+++ b/recipes/paleomix/build.sh
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-cp paleomix/main.py $PREFIX/bin/
+cp paleomix/main.py $PREFIX/bin/.
 cp -r paleomix $PREFIX/lib/python2.7/site-packages/paleomix
--- a/recipes/paleomix/build.sh
+++ b/recipes/paleomix/build.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
-
+mkdir -p ${PREFIX}/bin
+mkdir -p ${PREFIX}/lib/python2.7/site-packages/paleomix
 cp paleomix/main.py $PREFIX/bin/main.py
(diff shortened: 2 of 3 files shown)


recipes/paleomix/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
marcelm: Is this build script needed? Can’t you use the normal way to install Python packages instead? That is, use this in the `build` section in the recipe:
```
script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed -vv"
```

recipes/paleomix/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - gatk
+    - coloredlogs
+    - pysam
+    - ruamel.yaml
+    - biopython
+    - setproctitle
+
+test:
+  commands:
+    - main.py --help
```
marcelm: This should really not be `main.py`. Using the proper `pip install` incantation would install a script named `paleomix` instead.

recipes/paleomix/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+source:
+  url: https://github.com/MikkelSchubert/{{ name }}/archive/v{{ version }}.tar.gz
+  sha256: 7c46e8652f1c560aca81d865aec5cf7dc0309f9b3ef52c23f694f582963e33d0
+
+build:
+  number: 0
+  noarch: generic
+
+requirements:
+  run:
```
marcelm: You need `python` and `pip` host requirements.

recipes/paleomix/LICENSE
```diff
@@ -0,0 +1,18 @@
+
```
marcelm: You should not need to add this. (It will be retrieved from the downloaded source tarball.) Also, something is wrong: This is not the GPLv3.

recipes/paleomix/build.sh
gtrichard: pip is kind of broken for paleomix... I never successfully installed it using pip.

recipes/paleomix/meta.yaml
gtrichard: This would be the case if the tool would have a bin folder with an executable called paleomix I think? Or perhaps pip should do the trick, but I never successfully used pip to install paleomix...

recipes/paleomix/LICENSE
gtrichard: circleci was complaining about missing LICENSE file... On other recipes I did not have the issue though (the LICENSE file was pulled directly from the github repo).

recipes/paleomix/meta.yaml
marcelm: Yes, `pip` takes care of this. The `setup.py` file defines an entry_point, which tells `pip` (well, setuptools) to create a wrapper script at install time.

recipes/paleomix/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - biopython
+    - setproctitle
+
+test:
+  commands:
+    - main.py --help
+
+about:
+  home: https://github.com/MikkelSchubert/paleomix
+  license: GPL3
```
marcelm: The license isn’t GPL3

recipes/paleomix/meta.yaml
```diff
@@ -0,0 +1,50 @@
+{% set name = "paleomix" %}
+{% set version = "1.2.14" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/MikkelSchubert/{{ name }}/archive/v{{ version }}.tar.gz
```
marcelm: Please use the version from PyPI. URL could be `https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz`

gtrichard:
I think the author should fix the pip installation for a working conda recipe (issue comes from coverage>=4.0). I will post an issue on the paleomix github.

recipes/paleomix/meta.yaml
gtrichard: Sorry about that. Corrected to MIT.

marcelm:
Hm, I can actually install the program locally with `pip`. This sequence of commands works for me:
```
$ virtualenv -p python2.7 tmp
$ tmp/bin/pip install paleomix
$ tmp/bin/paleomix 
PALEOMIX - pipelines and tools for NGS data analyses.
Version: 1.2.14
```

recipes/paleomix/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - bowtie2
+    - bwa
+    - mapdamage2
+    - gatk
+    - coloredlogs
+    - pysam
+    - ruamel.yaml
+    - biopython
+    - setproctitle
+    - coverage
```
marcelm: The `setup.py` script specifies `coverage` as a `setup_requires` dependency, so it must be available at installation time. Translated into Conda terms, this means that you need to add it as a `host` dependency.

The full list of dependencies is at https://github.com/MikkelSchubert/paleomix/blob/v1.2.14/setup.py#L78 : 
`nose>=1.3.0`, `flexmock>=0.9.7`, `coverage>=4.0.0`

It’s nota good idea to have these development-only dependencies within `setup_requires`, and this has actually been fixed in the master version of paleomix. But to package 1.2.14, you’ll need to add the above to the host requirements. Or perhaps you can convince the author to release a new version shortly.

recipes/paleomix/LICENSE
marcelm: Yes, you‘re right. Downloading and unpacking the paleomix-1.2.14.tar.gz file manually, one can see that there are actually two license files `licenses/gpl.txt` and `licenses/mit.txt`. I don’t know if `license_file` supports multiple license files, but if it does, then you could refer to those files.

gtrichard:
flexmock seems to be an issue. It is available on conda but in specific repositories only that I'm not sure bioconda checks.

https://anaconda.org/search?q=flexmock

gtrichard:
Hum I also successfully installed via pip, several users told me it was not working via pip.... Well I got some errors related to deeptools (?) but this must be due to my setup, paleomix installed well using the following:

```
conda create -n paleomix python=2.7 pip
pip install paleomix
```

marcelm:
If you don’t want to wait for upstream, you could patch the `setup.py` file yourself, for example like this:
```
sed -i.bak 's/setup_requires/ignore_me/' setup.py
```
Perhaps add a comment that this needs to be removed on the next version.

gtrichard:
I used a bit paleomix and I think that the recipe will not take care of all the dependencies anyway... Paleomix is set to look for gatk (unused in the next version) and picard tools in the following folder: `~/install/jar_root/picard.jar` ... 

So even if the conda recipe works, paleomix will still be malfunctioning and would require end-user manipulation... basically the following:

```
mkdir -p ~/install/jar_root/
cd  ~/install/jar_root/
wget https://github.com/broadinstitute/picard/releases/download/2.21.9/picard.jar
```

The author said a new version of paleomix is coming soon... So I guess I will eventually wait for the new version to make a conda recipe, hoping for a more integrated picard call.

Meanwhile I'm closing this PR.

marcelm:
Oh, good observation. Then waiting for the next version makes a lot of sense!

PR #4649, opened 2017-05-09
Shapeit

hardingnj:
* [x] I have read the guidelines above.
* [x] This PR adds a new recipe.
* [ ] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

Issue #1763, opened 2016-06-16
SHAPEIT license


marcelm:
I would like to add a recipe for SHAPEIT, which is free for academic use. The web site states, however:
> A specific license must be obtained for any commercial or for-profit organization or for any web-diffusion purpose.

I do not know what a "web-diffusion purpose" is, but to be on the safe side, I assume it includes making a bioconda package available.

I have therefore asked the license holders Prof. Zagury and Prof. Marchini whether they would agree to include SHAPEIT into bioconda. I will record their decision in this issue, in order to prevent that others bother them with the same request.

bgruening:
Hi @marcelm,

please have a look at this section: https://github.com/bioconda/bioconda-recipes/blob/master/GUIDELINES.md#other-examples
You could do something similar for the gatk package.

Thanks to @tomkinsc for updating this document!

marcelm:
Thanks, that is a possibility if the answer is negative - but it’s worth asking.

bgruening:
Definitely! Thanks for doing so, hopefully it will be positive!

marcelm:
Jonathan Marchini has thanked for the offer, but has declined it. 

I will myself not create a GATK-like recipe.

bgruening:
This is so sad :(

PR #20455, opened 2020-02-21
Remove SHAPEIT

marcelm:
A couple of years ago, I wanted to package SHAPEIT, but checked the license and it didn’t allow re-distribution. I asked the authors for their permission, and they did not give it, see #1763.

A SHAPEIT recipe has been created in the meantime, which I noticed today.

The [SHAPEIT web site](https://mathgen.stats.ox.ac.uk/genetics_software/shapeit/shapeit.html) still includes this statement:
> A specific license must be obtained for any commercial or for-profit organization or for any web-diffusion purpose.

If anyone wants to contact the SHAPEIT authors again and ask them whether they would reconsider, feel free to do so, but in the meantime, Bioconda cannot redistribute SHAPEIT, and all shapeit packages need to be removed from anaconda.org.

There is also an `extractpirs` recipe, which may be affected as well as it packages another tool from the SHAPEIT website.

Another question is how to avoid this in the future. There should be a way to block recipes from being added.

Ping @bioconda/core, @hardingnj

cf #4649


mbargull:
Thanks for having a look out for those licensing issues!

> > A specific license must be obtained for any commercial or for-profit organization or for any web-diffusion purpose.

Since we are not commercial or for-profit, this would generally not prevent us from distributing it (apart from the usual spiel that people might argue that our hoster, Anaconda Inc, is commercial -.-' ). The software *users* will have to check the licenses in any case.

However, the included `LICENSE` file states
> The SHAPEIT software remains at your Institution and is not published, distributed, or otherwise transferred or made available to other than Institution employees and students involved in research under your supervision.

which quite clearly states that we cannot distribute it.

As for `extractpirs`, they don't provide any explicit license whatsoever (unless I missed it) at https://mathgen.stats.ox.ac.uk/genetics_software/shapeit/shapeit.html#readaware .
So, the best guess is that they intended to use the same license as for SHAPEIT. (Either way, if no license is provided, we are not granted distribution rights anyway.)
Can you also remove the `extractpirs` recipe in this PR? (And thanks for catching that recipe as well!)

mbargull:
> Another question is how to avoid this in the future. There should be a way to block recipes from being added.

Yeah, we don't have any good way to handle this currently.
Can you add something like
```
# License prevents distribution on Bioconda
recipes/extractpirs
recipes/shapeit

```
to https://github.com/bioconda/bioconda-recipes/blob/master/build-fail-blacklist ?
At least this would make the builds fail in case someone would try to re-add those.

hardingnj:
Thanks for catching this- my mistake. 

A new shapeit version has been published which is available under MIT: https://github.com/odelaneau/shapeit4

mbargull:
@hardingnj, that's good to hear!
... and apparently we already have a recipe for this: https://github.com/bioconda/bioconda-recipes/tree/master/recipes/shapeit4 :tada:

hardingnj:
good news! So it's just a case of deprecating the original `shapeit` recipe then

marcelm:
@mbargull
> > A specific license must be obtained for any commercial or for-profit organization or for any web-diffusion purpose.

> Since we are not commercial or for-profit, this would generally not prevent us from distributing it 

Note how it says "... *or* for any web-diffusion purposes", so even that sentence would prevent distribution, even non-commercial.

> Can you also remove the extractpirs recipe in this PR?

Done, and both are now added to the blacklist.

@hardingnj Cool that there’s an MIT-licensed version! This makes me feel a lot less bad about noticing this at all.

@mbargull Can you also remove the packages from anaconda.org?

mbargull:
> Note how it says "... *or* for any web-diffusion purposes", so even that sentence would prevent distribution, even non-commercial.

TBH, I just ignored that phrase because I find it too unspecific/confusing so it's unclear to me if they mean software distribution or, e.g., rather usage in a webservice by this (I was more interpreting it in the latter sense and even searched for the term, but https://fr.wikipedia.org/wiki/Webdiffusion was the only meaningful thing under the first few search results.. so no idea.).
But thankfully we don't have have to ponder about what they meant with it because of the `LICENSE` text :D
.

> Done, and both are now added to the blacklist.

Thanks!

> @mbargull Can you also remove the packages from anaconda.org?

We'll remove it, yes.

dpryan79:
These are now both gone from the conda cloud.

PR #20645, opened 2020-03-01
Add gargammel recipe

jfy133:
This adds a new bioconda recipe for [gargammel](https://github.com/grenaud/gargammel), a tool for simulating ancient DNA datasets
----

>
>
>
>

### DIFF
--- a/recipes/gargammel/build.sh
+++ b/recipes/gargammel/build.sh
@@ -3,6 +3,8 @@
 export CFLAGS="$CFLAGS -I$PREFIX/include"
 export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
 
+export CPATH=${PREFIX}/include
+
 sed -i.bak  's/make -C libgab/make -C libgab CXX=$(CC)/' Makefile
 
 ## Following https://bioconda.github.io/contributor/troubleshooting.html#g-or-gcc-not-found
--- a/recipes/gargammel/build.sh
+++ b/recipes/gargammel/build.sh
@@ -2,6 +2,7 @@
 
 export CPATH=${PREFIX}/include
 
+# Will be inbuilt to github repo in future
 sed -i.bak 's/make -C libgab/make -C libgab CXX=$(CC)/' Makefile
 sed -i.bak 's/make -C src/make -C src CXX=$(CC)/' Makefile
 sed -i.bak 's/CXXFLAGS =/CXXFLAGS +=/' src/Makefile
@@ -16,6 +17,8 @@ make CC=$CXX ${binaries}
 
 mkdir -p $PREFIX/bin
 cp ${binaries} ${PREFIX}/bin
+cp Makefile ms2chromosomes.py msprime_chromosomes.py ${PREFIX}/bin
+
 # Fix hard-coded paths to external tools
 sed \
   -e 's|$pathdir."/src/|"|' \
--- a/recipes/gargammel/build.sh
+++ b/recipes/gargammel/build.sh
@@ -2,10 +2,10 @@
 
 export CFLAGS="$CFLAGS -I$PREFIX/include"
 export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
-
 export CPATH=${PREFIX}/include
 
-sed -i.bak  's/make -C libgab/make -C libgab CXX=$(CC)/' Makefile
+sed -i.bak 's/make -C libgab/make -C libgab CXX=$(CC)/' Makefile
+sed -i.bak 's/make -C src/make -C src CXX=$(CC)/' Makefile
 
 ## Following https://bioconda.github.io/contributor/troubleshooting.html#g-or-gcc-not-found
 make CC=$CXX
--- a/recipes/gargammel/build.sh
+++ b/recipes/gargammel/build.sh
@@ -7,7 +7,7 @@ export CPATH=${PREFIX}/include
 sed -i.bak 's/make -C libgab/make -C libgab CXX=$(CC)/' Makefile
 sed -i.bak 's/make -C src/make -C src CXX=$(CC)/' Makefile
 sed -i.bak 's/CXXFLAGS =/CXXFLAGS +=/' src/Makefile
-sed -i.bak 's/LDFLAGS =/LDFLAGS +=/' src/Makefile
+sed -i.bak 's/LDFLAGS  =/LDFLAGS +=/' src/Makefile
 
 ## Following https://bioconda.github.io/contributor/troubleshooting.html#g-or-gcc-not-found
 make CC=$CXX
--- a/recipes/gargammel/build.sh
+++ b/recipes/gargammel/build.sh
@@ -6,6 +6,8 @@ export CPATH=${PREFIX}/include
 
 sed -i.bak 's/make -C libgab/make -C libgab CXX=$(CC)/' Makefile
 sed -i.bak 's/make -C src/make -C src CXX=$(CC)/' Makefile
+sed -i.bak 's/CXXFLAGS =/CXXFLAGS +=/' src/Makefile
+sed -i.bak 's/LDFLAGS =/LDFLAGS +=/' src/Makefile
 
 ## Following https://bioconda.github.io/contributor/troubleshooting.html#g-or-gcc-not-found
 make CC=$CXX
--- a/recipes/gargammel/build.sh
+++ b/recipes/gargammel/build.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+export CFLAGS="$CFLAGS -I$PREFIX/include"
+export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
+
 sed -i.bak  's/make -C libgab/make -C libgab CXX=$(CC)/' Makefile
 
 ## Following https://bioconda.github.io/contributor/troubleshooting.html#g-or-gcc-not-found
--- a/recipes/gargammel/build.sh
+++ b/recipes/gargammel/build.sh
@@ -1,7 +1,5 @@
 #!/bin/bash
 
-export CFLAGS="$CFLAGS -I$PREFIX/include"
-export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
 export CPATH=${PREFIX}/include
 
 sed -i.bak 's/make -C libgab/make -C libgab CXX=$(CC)/' Makefile
@@ -10,11 +8,18 @@ sed -i.bak 's/CXXFLAGS =/CXXFLAGS +=/' src/Makefile
 sed -i.bak 's/LDFLAGS  =/LDFLAGS +=/' src/Makefile
 
 ## Following https://bioconda.github.io/contributor/troubleshooting.html#g-or-gcc-not-found
-make CC=$CXX
+binaries="src/fragSim src/deamSim src/adptSim src/fasta2fastas"
+# We list the targets explicitly to prevent the Makefile from downloading and
+# building the external art_illumina.o dependency (added as a runtime
+# requirement instead)
+make CC=$CXX ${binaries}
 
-outdir=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
-mkdir -p $outdir
 mkdir -p $PREFIX/bin
-cp -R * $outdir/
-
-ln -s $outdir/gargammel.pl $PREFIX/bin/gargammel
+cp ${binaries} ${PREFIX}/bin
+# Fix hard-coded paths to external tools
+sed \
+  -e 's|$pathdir."/src/|"|' \
+  -e 's|$pathdir."/art_src_MountRainier/|"|' \
+  -e 's|^fileExists(.*);||' \
+  -e 's|#!/usr/bin/perl|#!/usr/bin/env perl|' \
+  gargammel.pl > ${PREFIX}/bin/gargammel
--- a/recipes/gargammel/build.sh
+++ b/recipes/gargammel/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-sed -i  's/make -C libgab/make -C libgab CXX=$(CC)/' Makefile
+sed -i.bak  's/make -C libgab/make -C libgab CXX=$(CC)/' Makefile
 
 ## Following https://bioconda.github.io/contributor/troubleshooting.html#g-or-gcc-not-found
 make CC=$CXX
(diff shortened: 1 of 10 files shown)


recipes/gargammel/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+sed -i  's/make -C libgab/make -C libgab CXX=$(CC)/' Makefile
```
marcelm: On macOS, the `-i` option to `sed` works a bit differently and you need to write `sed -i.bak` instead (this also works on Linux).

recipes/gargammel/build.sh
```diff
@@ -0,0 +1,15 @@
...
+make CC=$CXX
+
+outdir=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+mkdir -p $outdir
+mkdir -p $PREFIX/bin
+cp -R * $outdir/
+
+ln -s $outdir/gargammel.pl $PREFIX/bin/gargammel
+
+
```
marcelm: The extra newlines at the end of the file should be removed.

recipes/gargammel/build.sh
jfy133: Hi Marcel, thanks for the tip and the review (s). I will fix once I get this to actually compile properly on Linux....

I'm having issues with the fact there is 'compilation with compilation' (ugly, but not my decision) and the C++ variables for paths aren't inherited into the internal make files.

marcelm: I can understand your pain. For these types of tools that download their dependencies at build time, packaging is a bit annoying. The cleanest way to handle this would be to patch the tool so it can work with already-installed dependencies, and list those in the recipe. But this is only a suggestion since it sometimes a bit hard to achieve and many other recipes aren’t perfect in that regard either.

jfy133:
I've done enough internal patches and added enough depencies for the package to compile, but there is a conda error now (I think). Does anyone have any idea?

Note the patching currently done in `build.sh` will be done in the actual github archive on the next release (Dev waiting for this to compile first before making a new release).

```
19:50:54 BIOCONDA INFO (OUT) WARNING :: `patchelf --print-rpath` failed for /opt/conda/conda-bld/gargammel_1588621571732/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/share/gargammel-1.1.2-0/bamtools/build/src/api/CMakeFiles/BamTools.dir/internal/io/TcpSocketEngine_p.cpp.o, will proceed with LIEF (was None)
 
19:50:54 BIOCONDA INFO (OUT) WARNING :: get_rpaths_raw()=['/opt/conda/conda-bld/gargammel_1588621571732/_build_env/lib:/opt/conda/conda-bld/gargammel_1588621571732/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib'] and patchelf=[['/opt/conda/conda-bld/gargammel_1588621571732/_build_env/lib', '/opt/conda/conda-bld/gargammel_1588621571732/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/lib']] disagree for /opt/conda/conda-bld/gargammel_1588621571732/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/share/gargammel-1.1.2-0/libgab/testRecons ::
 
19:50:54 BIOCONDA INFO (OUT) Warning: rpath /opt/conda/conda-bld/gargammel_1588621571732/_build_env/lib is outside prefix /opt/conda/conda-bld/gargammel_1588621571732/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla (removing it)
 
19:50:54 BIOCONDA INFO (OUT) patchelf: wrong ELF type
 
19:50:54 BIOCONDA INFO (OUT) WARNING :: `patchelf --print-rpath` failed for /opt/conda/conda-bld/gargammel_1588621571732/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/share/gargammel-1.1.2-0/src/misincorporation2prof.o, will proceed with LIEF (was None)
 
19:50:54 BIOCONDA INFO (OUT) patchelf: cannot find section '.dynamic'. The input file is most likely statically linked
 
19:50:54 BIOCONDA INFO (OUT) WARNING :: `patchelf --print-rpath` failed for /opt/conda/conda-bld/gargammel_1588621571732/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/share/gargammel-1.1.2-0/bamtools/tools/Linux/clang-format, will proceed with LIEF (was None)
 
19:50:54 BIOCONDA INFO (OUT) Traceback (most recent call last):
 
19:50:54 BIOCONDA INFO (OUT)   File "/opt/conda/bin/conda-build", line 11, in <module>
 
19:50:54 BIOCONDA INFO (OUT)     sys.exit(main())
 
19:50:54 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.7/site-packages/conda_build/cli/main_build.py", line 474, in main
 
19:50:54 BIOCONDA INFO (OUT)     execute(sys.argv[1:])
 
19:50:54 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.7/site-packages/conda_build/cli/main_build.py", line 465, in execute
 
19:50:54 BIOCONDA INFO (OUT)     verify=args.verify, variants=args.variants)
 
19:50:54 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.7/site-packages/conda_build/api.py", line 209, in build
 
19:50:54 BIOCONDA INFO (OUT)     notest=notest, need_source_download=need_source_download, variants=variants)
 
19:50:54 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.7/site-packages/conda_build/build.py", line 2863, in build_tree
 
19:50:54 BIOCONDA INFO (OUT)     notest=notest,
 
19:50:54 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.7/site-packages/conda_build/build.py", line 2150, in build
 
19:50:54 BIOCONDA INFO (OUT)     newly_built_packages = bundlers[pkg_type](output_d, m, env, stats)
 
19:50:54 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.7/site-packages/conda_build/build.py", line 1504, in bundle_conda
 
19:50:54 BIOCONDA INFO (OUT)     files = post_process_files(metadata, initial_files)
 
19:50:54 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.7/site-packages/conda_build/build.py", line 1363, in post_process_files
 
19:50:54 BIOCONDA INFO (OUT)     post_build(m, new_files, build_python=python)
 
19:50:54 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.7/site-packages/conda_build/post.py", line 1174, in post_build
 
19:50:54 BIOCONDA INFO (OUT)     post_process_shared_lib(m, f, prefix_files, host_prefix)
 
19:50:54 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.7/site-packages/conda_build/post.py", line 1122, in post_process_shared_lib
 
19:50:54 BIOCONDA INFO (OUT)     mk_relative_osx(path, host_prefix, m.config.build_prefix, files=files, rpaths=rpaths)
 
19:50:54 BIOCONDA INFO (OUT)   File "/opt/conda/lib/python3.7/site-packages/conda_build/post.py", line 399, in mk_relative_osx
 
19:50:54 BIOCONDA INFO (OUT)     assert sys.platform == 'darwin'
 
19:50:54 BIOCONDA INFO (OUT) AssertionError
 ```
Source: https://app.circleci.com/pipelines/github/bioconda/bioconda-recipes/25400/workflows/fc53711f-9471-40ef-af28-47395d67d589/jobs/106970/steps

File: recipes/gargammel/build.sh [hunk 3]
```diff
@@ -0,0 +1,20 @@
...
+sed -i.bak 's/CXXFLAGS =/CXXFLAGS +=/' src/Makefile
+sed -i.bak 's/LDFLAGS  =/LDFLAGS +=/' src/Makefile
+
+## Following https://bioconda.github.io/contributor/troubleshooting.html#g-or-gcc-not-found
+make CC=$CXX
+
+outdir=$PREFIX/share/$PKG_NAME-$PKG_VERSION-$PKG_BUILDNUM
+mkdir -p $outdir
+mkdir -p $PREFIX/bin
+cp -R * $outdir/
```
marcelm, line 18: It appears that you are copying over all contents of the build directory to the $PREFIX (including Makefiles and object files etc.), but the $PREFIX should only contain the stuff that should be part of the package. This also explains the `patchelf` error messages (they refer to `.o` files).

recipes/gargammel/build.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+export CFLAGS="$CFLAGS -I$PREFIX/include"
```
marcelm, line 3: This line is not necessary as the CFLAGS as defined by Conda already add $PREFIX/include.

recipes/gargammel/build.sh
```diff
@@ -0,0 +1,20 @@
...
+export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
```
marcelm, line 4: Same here, LDFLAGS already contain `$PREFIX/lib`.

recipes/gargammel/build.sh
marcelm: It appears that the binaries that the main `gargammel.pl` script needs are listed in line 236ff:
```
$adptsim = $pathdir."/src/adptSim";
my $deamsim = $pathdir."/src/deamSim";
my $fragsim = $pathdir."/src/fragSim";
my $artprog = $pathdir."/art_src_MountRainier/art_illumina";
```
So it may be sufficient to just copy those from `src/` to `$PREFIX/bin`.

marcelm:
I’ve applied my suggested changes and also made it so that ART is no longer compiled, but instead an external dependency.

recipes/gargammel/build.sh
jfy133: Ah, thank you for the explanation of cause of the `patchelf` messages. That makes sense, I modified the script after a more simpler recipe, where it didn't require so much complication.

jfy133:
@marcelm Thank you very much, that helps a lot. I'm just testing a locally built version using your fixes, and found a few more internal issues that I need to check with the developer, which once fixed I will request review.

jfy133:
The python scripts aren't apparently requirements despite the github repo https://github.com/grenaud/gargammel/issues/7#issuecomment-625830982

Once the recipe passes tests, we are good to go for review and merge @marcelm . Thanks again!

recipes/gargammel/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+about:
+  home: https://github.com/grenaud/gargammel
+  license: GPL
+  license_file: LICENSE
+  summary: Tool for simulating ancient DNA datasets
+
+test:
+  commands:
+    - gargammel --help
```
marcelm, line 39: Could you also test running the binaries that the Perl script calls? The original script tested whether it could find those binaries, but the patched version doesn’t anymore, so while I trust that it works at the moment since you did some testing locally, it would be easy to miss problems in a future version of the recipe.

jfy133: Yes, will do. `art` definitely worked because I saw log output, but will check the others.

That said, wierdly, now after cleaning up my conda environments, I've found there is a unlisted depency of `samtools`.

But when I try locally re-building the package with the updated `meta.yaml`, and try `conda install . --use-local`, it doesn't pick `samtools` up...

marcelm: `art` is the one you don’t actually need to test because that is now pulled in as an external dependency.

Good you caught the samtools dependency!

jfy133:
> Could you also test running the binaries that the Perl script calls? The original script tested whether it could find those binaries, but the patched version doesn’t anymore, so while I trust that it works at the moment since you did some testing locally, it would be easy to miss problems in a future version of the recipe.

Done, I've added tests for those in the recipe as well.

> One thing that is still not so nice is that the Makefile downloads and builds its own version of `libgab` and `bamtools`. Since `libgab` is by the same author, it’s kind of ok, but for bamtools, it would be much cleaner to depend on the existing Bioconda package.
> 
> However, I agree that it might be a pain to do this de-vendoring, so it’s ok if this is left for a future version of the recipe.

Agreed. Will let the author know. 

I've gone through the gargammel README on the github repo and tested every command, and they all worked successfully, except for the methyl one, but that's only because I don't understand the input methylation matrix there. But as far as I can see from the code that spcific doesn't need any extra dependencies. 

So if you're happy with that and tests merge, hopefully we can merge?

jfy133:
@marcelm when running the binaries' help messages, they `exit 1`, meaning the CI fails.

Given I've run all the commands in the github repository (which use all of the binaries), so they do actually work, would this still be necessary, do you think? Or should I try finding mini data to download within the build, and test with that?

marcelm:
This is a common problem. The workaround used in many other recipes is to pipe into `grep` and search for some expected output. For example: `adptSim --help | grep -q adptSim`. (Perhaps you also need a `2>&1` before the `|`.)

marcelm:
And I forgot to actually answer your question: The tests need to pass, yes, but the workaround I desctribed is fine. And you don’t need to find a minimal dataset.

jfy133:
> This is a common problem. The workaround used in many other recipes is to pipe into `grep` and search for some expected output. For example: `adptSim --help | grep -q adptSim`. (Perhaps you also need a `2>&1` before the `|`.)

Done as requetsted @marcelm

marcelm:

marcelm:

marcelm:
@BiocondaBot please merge

marcelm:
Looks good now!

jfy133:
@marcelm Great! I've learnt a lot - thanks for your all your help!

marcelm:
No problem :-)

PR #20723, opened 2020-03-05
Build NGLess directly by first building GHC

luispedro:
This is a round-about way of doing it, but it works. Inspired by (copy & pasted from):
https://github.com/conda-forge/git-annex-feedstock/ from @notestaff 

### DIFF
--- a/recipes/ngless/meta.yaml
+++ b/recipes/ngless/meta.yaml
@@ -48,14 +48,8 @@ requirements:
     - zlib
     - bzip2
     - backports.lzma
-    - ncurses
-    - libmagic
-    - sqlite
-    - dbus
-    - libffi
     - libiconv
     - openssh
-    - libxml2
     - wget
   run:
     - python
--- a/recipes/ngless/meta.yaml
+++ b/recipes/ngless/meta.yaml
@@ -68,8 +68,6 @@ extra:
   container:
     # For GHC
     extended-base: true
-  skip-lints:
-    - should_be_noarch_generic
   recipe-maintainers:
     - luispedro
   identifiers:
--- a/recipes/ngless/meta.yaml
+++ b/recipes/ngless/meta.yaml
@@ -1,33 +1,82 @@
 {% set name = "ngless" %}
 {% set version = "1.1.0" %}
-{% set sha256 = "9706b5538f44f1962f6cd891672b4ff90a5b30523e91613df169e06ef61a8278" %}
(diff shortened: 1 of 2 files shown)


recipes/ngless/meta.yaml
```diff
@@ -1,33 +1,82 @@
...
+    - libtool
+    - stack >=1.9.3
+  host:
+    - gmp
+    - xz
+    - zlib
+    - bzip2
+    - backports.lzma
+    - ncurses
+    - libmagic
```
notestaff: Some of these dependencies (like libmagic) were for git-annex, prob can be dropped

luispedro: Thanks for the feedback. I'm running some of this locally now. Unfortunately, the testing is very slow on the machine I am using, so it may take a while if I have to do a few iterations. I'll only push a working version, though.

notestaff:
@luispedro Wow, you made it work on MacOS?  We've been trying to do that with git-annex-feedstock for a very long time, with no luck -- will study closely how it worked here, thanks a lot!

luispedro:
I think compilation on `Mac OS X`  was more of a happy accident than by design, probably because I have fewer dependencies now

luispedro:

notestaff:
there are some conda warnings at the end -- some can probably be ignored but best to check with the experts here ( @bgruening  @epruesse )

e.g.

13:58:27 BIOCONDA INFO (OUT) WARNING (ngless,bin/ngless): Needed DSO lib/libffi.so.6 found in ['libffi']
13:58:27 BIOCONDA INFO (OUT) WARNING (ngless,bin/ngless): .. but ['libffi'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)

in git-annex-feedstock, ghc is built with --with-system-libffi and libffi is a build and host dependency (the latter also exports it as a runtime dependency).

Also libgomp:

13:58:28 BIOCONDA INFO (OUT) WARNING (ngless,share/ngless/bin/ngless-1.1.0-megahit/megahit_sdbg_build): Needed DSO lib/libgomp.so.1 found in ['_openmp_mutex']
13:58:28 BIOCONDA INFO (OUT) WARNING (ngless,share/ngless/bin/ngless-1.1.0-megahit/megahit_sdbg_build): .. but ['_openmp_mutex'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)

Possibly relevant info here: https://conda-forge.org/docs/maintainer/knowledge_base.html#openmp

notestaff:
(turns out the macos compilation is actually skipped)

PR #20798, opened 2020-03-10
Psass

RomainFeron:
Adding recipe for PSASS (software to compare pooled-sequencing data for two populations)

----


recipes/psass/build.sh
```diff
@@ -0,0 +1,3 @@
+mkdir -p $PREFIX/bin/
```
raivivek: Perhaps add a shebang line? `/bin/bash`

RomainFeron: Thanks, I added it!

PR #20859, opened 2020-03-12
Update iqtree

andrewsanchez:
We encountered an inscrutable `UnsatisfiableError` from conda after trying to resolve a complicated environment on osx.  [See the integration-darwin step here](https://busywork.qiime2.org/teams/main/pipelines/master-distribution/jobs/integration/builds/163).

After noticing this unexpected behavior in our CI system, we started examining our conda environments to try and understand the source of the issue.  

To summarize our findings (thanks, @ebolyen and @thermokarst 🕵! ):
- scikit-learn had been updated to a new version
- scikit-learn started being pulled from defaults, instead of conda-forge
- the defaults channel version of sklearn leaves `llvm-openmp` unconstrained. Our release build chose version 8 from conda-forge, even though version 9 had been available for months
- the conda-forge version of sklearn sets that same library to >=9, which apparently we couldn't find a resolution for, or the unconstrained openmp would have been 9 in the release (as it was available)
- it appeared that `llvm-openmp` is versioned in lockstep with `clangdev`, `llvmdev` , and `openmp`
- 9.0.1 drops the dependency on libcxx

This led us to suspect that packages depending on `llvm-openmp` could be the culprit.

We discovered that, by removing a specific package which depended on `iqtree`, conda could resolve the environment.  `iqtree` was a culprit due using `llvm-openmp@8.0.1` and `openmp@8.0.1`.

This PR is an attempt to avoid unsolvable environments in packages depending on `iqtree`.  Unfortunately, we had very little success running tests locally and are pushing up the first iteration of changes in order to have a look at the circleci results.  Notably, `circleci build` failed *after only incrementing the recipe build number,* without any other changes to `meta.yaml`.

 We had a look at the [spades recipe](https://github.com/bioconda/bioconda-recipes/blob/05e6af43cea7ed46a4f05d63d1d16fc49f7643ac/recipes/spades/meta.yaml) for guidance on configuring this recipe.

----

>
>
>
>


recipes/iqtree/meta.yaml
```diff
@@ -16,14 +16,15 @@ requirements:
   build:
     - {{ compiler('cxx') }}
     - cmake
+    - clangdev  # [osx]
```
thermokarst: Hey @andrewsanchez - maybe `host` still needs `clangdev`? Not sure why, but you could try putting this line back into the `host` section, below.

The [CI run](https://circleci.com/gh/bioconda/bioconda-recipes/98781?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) had this little tidbit of a failure: `ld: library not found for -lcrt1.10.6.o`, which is why I am wondering if we need `clangdev` in `host`.

andrewsanchez: Good idea.  and right below that it says  `clang-4.0: error: linker command failed with exit code 1` which seems like it could be related to the last bit of info right before it exits:  `Linking CXX executable iqtree`

recipes/iqtree/build.sh
```diff
@@ -1,11 +1,18 @@
...
-export CFLAGS="-I$PREFIX/include"
-export LDFLAGS="-L$PREFIX/lib"
+export CFLAGS="$CFLAGS -I$PREFIX/include"
+export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
 export CPATH=${PREFIX}/include
 
+if [ "$(uname)" == Darwin ]; then
+    export MACOSX_DEPLOYMENT_TARGET=10.7
+    export CMAKE_C_COMPILER="clang"
+    export CMAKE_CXX_COMPILER="clang++"
```
thermokarst: Can you toss this env var in here, too?

```bash
export CMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY
```

This changes the way cmake performs its pre-run checks.

recipes/iqtree/meta.yaml
```diff
@@ -6,11 +6,14 @@ package:
   version: {{ version }}
 
 build:
-  number: 0
+  number: 1
+  skip: true # [linux]
```
thermokarst: TODO: revert this platform pin (currently on to prevent spending time testing on linux, since this PR is attempting to address build issues specific to darwin).

recipes/iqtree/build.sh
```diff
@@ -1,11 +1,19 @@
 #!/bin/bash
-export CFLAGS="-I$PREFIX/include"
-export LDFLAGS="-L$PREFIX/lib"
+export CFLAGS="$CFLAGS -I$PREFIX/include"
+export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
 export CPATH=${PREFIX}/include
 
+if [ "$(uname)" == Darwin ]; then
+    export MACOSX_DEPLOYMENT_TARGET=10.7
+    export CMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY
```
thermokarst: Hmm, the latest circleci run implies that this change didn't get picked up - it appears to still be building a test executable...

recipes/iqtree/build.sh
```diff
@@ -1,11 +1,19 @@
 #!/bin/bash
-export CFLAGS="-I$PREFIX/include"
-export LDFLAGS="-L$PREFIX/lib"
+export CFLAGS="$CFLAGS -I$PREFIX/include"
+export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
 export CPATH=${PREFIX}/include
 
+if [ "$(uname)" == Darwin ]; then
+    export MACOSX_DEPLOYMENT_TARGET=10.7
```
thermokarst: Please update this to 10.9 so that it matches the patchfile.

recipes/iqtree/build.sh
thermokarst: You know, it might be the case that this flag can only be set with the `-D CMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY` syntax, rather than as an env var.

andrewsanchez: Let's see what happens in the next build.  The `patch` directive was not indented properly underneath `source` 🤦‍♂

bootstrap.py
```diff
@@ -52,7 +52,7 @@
 
 # TODO: this mimics the override in the "common" job in .circleci/config.yaml
 with open('.circleci/common.sh', 'w') as fout:
-    fout.write("MINICONDA_VER=4.5.4\nBIOCONDA_UTILS_TAG=master\n")
+    fout.write("MINICONDA_VER=4.6.14\nBIOCONDA_UTILS_TAG=v0.16.11\n")
```
thermokarst: You might want to revert this, since it isn't part of the recipe.

andrewsanchez: Oops!  Thank you.

recipes/iqtree/build.sh
andrewsanchez: It was originally set to 10.9.  You asked to try 10.7 after that build failed.  Do you remember why you wanted it set to 10.7 on Friday?

recipes/iqtree/build.sh
thermokarst: Looks like something is still off with the patch, can you double-check that?

recipes/iqtree/build.sh
thermokarst: >  Do you remember why you wanted it set to 10.7 on Friday?

Yep! We temporarily tested against 10.7 because we noticed that half the build was using 10.7 while the other was using 10.9 - you changed this env var to 10.7 to see if we could get everything building against the same version of macos. That succeeded in getting everything building against 10.7, but didn't do anything to address the underlying problem. So then the next step was to try to build everything against 10.9. We started that by creating a patch to the CMake config to force that end to use 10.9, next we need to fix up this env var to use 10.9, too.

recipes/iqtree/macos.patch
```diff
@@ -0,0 +1,21 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
```
thermokarst: Don't forget to drop this patchfile.

recipes/iqtree/build.sh
```diff
@@ -1,11 +1,19 @@
 #!/bin/bash
-export CFLAGS="-I$PREFIX/include"
-export LDFLAGS="-L$PREFIX/lib"
+export CFLAGS="$CFLAGS -I$PREFIX/include"
+export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
 export CPATH=${PREFIX}/include
 
+if [ "$(uname)" == Darwin ]; then
+    export MACOSX_DEPLOYMENT_TARGET=10.9
```
thermokarst: Okay, as a quick sanity check, can you temporarily comment out this line, and the line in the recipe that applies the patch? I just want to make sure that forcing a 10.9 target is something necessary for this changeset.

recipes/iqtree/build.sh
andrewsanchez: Do you still want to run any tests re: `CMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY` ?

recipes/iqtree/build.sh
```diff
@@ -1,11 +1,19 @@
 #!/bin/bash
-export CFLAGS="-I$PREFIX/include"
-export LDFLAGS="-L$PREFIX/lib"
+export CFLAGS="$CFLAGS -I$PREFIX/include"
+export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
 export CPATH=${PREFIX}/include
 
+if [ "$(uname)" == Darwin ]; then
+    # export MACOSX_DEPLOYMENT_TARGET=10.9
```
thermokarst: Thanks for commenting out, but it looks like you missed the patch, [below](https://github.com/bioconda/bioconda-recipes/pull/20859/files#diff-ad768d0ca78f5f0c5a55d9858164f237R16).

recipes/iqtree/build.sh
thermokarst: Not just yet, let's focus on resolving one thing at a time right now - let's finish our 10.9 checks.

recipes/iqtree/build.sh
andrewsanchez: Ok, I will skip applying the patch as well.  Sorry, I missed that in your previous comment.

thermokarst: No worries, thanks!

recipes/iqtree/build.sh
andrewsanchez: Builds failed. Adding the patch and deployment target back in

recipes/iqtree/build.sh
thermokarst: Okay, now that 10.9 is sorted out, can you check and see if this flag is necessary, too?

thermokarst: PS - this "parametric" development cycle is painful - sorry about that!

andrewsanchez: To clarify, you just want to test with `export  CMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY`? commented out?

andrewsanchez: All checks have passed without this flag.

thermokarst:

andrewsanchez:
@BiocondaBot please add label

davebx:
Thanks for all the work, @thermokarst, @andrewsanchez 

@BiocondaBot please merge

davebx:
@BiocondaBot please merge

PR #21032, opened 2020-03-23
Metaphlan3

fbeghini:

>
>
>
>


recipes/metaphlan/meta.yaml
```diff
@@ -7,23 +7,25 @@ package:
...
-  sha256: 64040ce57892736cccf8a325bbfbf4e6c641d8cb295779f33bd12be0b25c573c
+  sha256: 30b1afdf5d66565192471381d340096d8da49422643ee5f2d50b3e5a2920cf71
+
 
 build:
   noarch: python
-  number: 1
-  script: {{ PYTHON }} -m pip install . -vv
+  number: 3
+  script: {{ PYTHON }} setup.py biocondainstall
```
marcelm: This is very unusual. Why did you find it necessary to introduce a bioconda-specific installation command?

fbeghini: I need to copy some dependencies by downloading a repo. I cannot trigger the cmdclass using pip

marcelm: Why can’t you use the normal mechanisms for installing dependencies? (`install_requires` in `setup.py` and `requirements` in the recipe)

fbeghini: It is not available in pypi nor in conda, it still has to be packaged. I agree that is not a clean way to install it

marcelm: Ok. It appears that [CMSeq](https://github.com/SegataLab/cmseq/) is the package in question. In your `setup.py`, you may want to try using [dependency_links](https://python-packaging.readthedocs.io/en/latest/dependencies.html#packages-not-on-pypi), which allows you to specify dependencies that are not on PyPI (I haven’t used it myself, so I don’t know whether this will actually work).

In any case, you should create a Bioconda package for CMSeq and then add it as a dependency (this can be done even if it is not on PyPI).

Don’t special-case Bioconda in your package installation routine. Either get it right so that it works for everyone, not only when building the Conda package. Or, if some Conda-specific treatment is still necessary, that needs to be part of the recipe.

fbeghini: @marcelm thank you for the advide, setuptools ignores the `dependency_links` field but with pip, but using a different format, it works.
I've opened a pull request for a new conda package for CMSeq

PR #21049, opened 2020-03-24
Add ExpressBetaDiversity 1.0.10

aaronmussig:
Adds the ExpressBetaDiversity package from https://github.com/dparks1134/ExpressBetaDiversity/

----

>
>
>
>


recipes/expressbetadiversity/build.sh
```diff
@@ -0,0 +1,37 @@
...
+mkdir -p ${SHARE_DIR}/bin/
+cd ../bin
+chmod +x ExpressBetaDiversity
+cp ExpressBetaDiversity ${SHARE_DIR}/bin/
+
+# Create the binary
+mkdir -p ${PREFIX}/bin/
+cat <<EOF >${PREFIX}/bin/ExpressBetaDiversity
+#!/bin/bash
+cd ${SHARE_DIR}/bin/
```
marcelm: Can you add a comment explaining why you need the `cd`?

recipes/expressbetadiversity/build.sh
```diff
@@ -0,0 +1,37 @@
...
+./ExpressBetaDiversity "\$@"
```
marcelm: You should prefix this with `exec`:
```
exec ./ExpressBetaDiversity "\$@"
```
This avoids the extra `bash` process hanging from around unnecessarily.

aaronmussig: Cheers, added.

recipes/expressbetadiversity/build.sh
aaronmussig: Added some comment, but this is mainly due to the unit tests being called relative to the binary location (../unit-tests). Alternatively, a patch file could be created before building it but I'm not entirely sure if that is the cleaner solution? I'm open to give it a go if that is what bioconda prefers.

dpryan79:

recipes/expressbetadiversity/build.sh
marcelm: Sorry I haven’t looked too closely now, but what I think I don’t understand is why the installed version of the package needs to be modified for unit tests, while I would expect unit tests to be run only at build time.

aaronmussig: As would I, but there is an argument that can be passed (-u) when running the program which runs all unit tests. So this is being kept purely for that functionality.

dpryan79: Are there otherwise ever file paths given to this? My primary concern is that the `cd` will cause issues if relative paths are provided (I make copious use of them when analyzing data). If that's not the case, or there's something that I don't understand that makes that not break then I'm fine with this. Otherwise I would suggest that breaking unit tests is a fine compromise.

aaronmussig:

dpryan79:
@bioconda-bot please merge

PR #21067, opened 2020-03-24
Pgt3.3 dep

lldelisle:
**Updating the dependencies of pygenometracks version 3.3: tqdm is needed, pybigwig and hicmatrix are increased and matplotlib is fixed.**

----

* [ ] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/contributor/guidelines.html).

>
>
>
>

### DIFF
--- a/recipes/pygenometracks/meta.yaml
+++ b/recipes/pygenometracks/meta.yaml
@@ -10,7 +10,7 @@ source:
   sha256: 32562129041755beb32be3d3ce37763c85ee5cfa8da6d0a9d5bc95df6489cd46
 
 build:
-  number: 0
+  number: 1
   script: "{{ PYTHON }} -m pip install . --no-deps -vv"
   noarch: python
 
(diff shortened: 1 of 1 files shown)


recipes/pygenometracks/meta.yaml
```diff
@@ -21,13 +21,14 @@ requirements:
   run:
     - python >=3.6
     - numpy >=1.16
-    - matplotlib >=3.0
+    - matplotlib-base =3.1.1
```
marcelm: Requiring this exact version is quite strict. For example, if one would try to install this at the same time as another tool that requires `>=3.1.2`, it wouldn’t work. Is that actually necessary? Or would perhaps something like `>=3.1.1,<3.2` work?

joachimwolff: We made bad experience concerning the stability of API of matplotlib even for minor changes like 3.0.1 to 3.0.2. Moreover, matplotlib likes to change the size of the images, default colors, default font and font sizes from (minor) update to (minor) update. If this leads to crashes in pyGenomeTracks or that the visualization does not look like it was intended by us. However, the users will blame us for all this. We are simply truly annoyed by this behaviour and fix therefore the version number. We recommend to use pyGenomeTracks in its own environment.

marcelm: Thanks for the explanation!

PR #21093, opened 2020-03-25
Update pp-sketchlib to v1.2.0

johnlees:
Requires CUDA to compile on host, and armadillo to run. Using openblas and lapack rather than intel libraries on linux, as they seem less likely to cause version conflicts.

----

>
>
>
>

### DIFF
--- a/recipes/pp-sketchlib/meta.yaml
+++ b/recipes/pp-sketchlib/meta.yaml
@@ -13,7 +13,15 @@ source:
 build:
   number: 0
   skip: True  # [py2k or win]
-  script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv"
+  script:
+    # CUDA 10.1 moves some headers to `/usr/include`
+    # So add that location last in the header search paths
+    - export CFLAGS="${CFLAGS} -I/usr/include"  # [linux64]
+    - export CPPFLAGS="${CPPFLAGS} -I/usr/include"  # [linux64]
+    - export CXXFLAGS="${CXXFLAGS} -I/usr/include"  # [linux64]
+    - "{{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv"
+  missing_dso_whitelist:
+    - "*/libcuda.*"
   entry_points:
     - poppunk_sketch = pp_sketch.__main__:main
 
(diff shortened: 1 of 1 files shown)


johnlees:
@bioconda/core I am having trouble to get the CUDA part of this package to compile. There is one .cu file which uses `nvcc` to compile:
- Using the `nvcc_linux-64` and `cudatoolkit` packages gave me an error about `CUDA_HOME` not being set, and the compiler not being found. Not sure is a post link script setting this variable might help?
- Using the `cudatoolkit-dev` package fails when running [this post install script](https://github.com/conda-forge/cudatoolkit-dev-feedstock/blob/master/recipe/post-link.sh) on the bioconda CI. I can't replicate this locally

Do you know of any other bioconda packages that have successfully compiled CUDA code, and if so how they achieved this?

dpryan79:
I don't think we have any GPUs available on circleci.

Last I heard, the tensorflow package from conda-forge was being compiled and uploaded outside the normal CI workflow, do you know if that's still the case @bgruening?

johnlees:
I think it should be possible to compile on (CI) systems without a GPU as long as the nvcc tools are installed?

dpryan79:
I would also have presumed so, but I have no experience with nvcc :(

johnlees:
Do you think it's worth trying to put this on conda-forge, as they have a package with `nvcc` (I think)? Or I can drop the dependency and it will compile CPU code only, and I can leave it up to users with CUDA installs to compile the GPU code themselves

dpryan79:
Before trying it on conda-forge let's just try explicitly adding nvcc.

dpryan79:
Yeah, maybe try this in conda-forge and see if you can get it to work there. I have a feeling we'd need to update the pinnings to get this working easily.

mbargull:
conda-forge uses a different container image that provides the CUDA toolchain. This is not something we currently support.
We'll probably support it at some point, but AFAICT it's not high priority. So, I would guess it'll take at least months (late this year, maybe?) before we do. Though, I know @johanneskoester expressed interest in this.

johnlees:
Hi both,
Thanks for the input, that's good to know. For now I'll see if I can get it to work on conda-forge, and if I fail will revive this as CPU code only. 
I would also be interested in this functionality in bioconda, but appreciate that it's not high priority at the moment. Perhaps in a year or two there will be more demand, now that some CUDA genomics applications have been released

johnlees:
This is now on conda-forge instead - I'll close this, but presumably it makes sense to leave the older versions on bioconda?

PR #21118, opened 2020-03-26
Warn GTDB-Tk users that they are running an old version

aaronmussig:
This comes about as there have been reports that conda will refuse to install any GTDB-Tk version above 0.3.2. Users may be running conda install gtdbtk and are unware that they are using an old version of GTDB-Tk. Old versions should not be used as significant improvements have been made since then, no users should be downloading old versions especially at the rate you can see below:

image

I'm using the recipe for that version and only updating the build number, and creating a patch file which warns the user it's not the most recent version. If this works, I plan to do this to all previous versions.

As CI can't seem to handle multiple versions of the package being updated I'll do this in multiple PRs.

**[not ready for merge]**

----

>
>
>
>


aaronmussig:

marcelm:
This is quite an intrusive, unprecedented change, and therefore would need to be justified well.

Note that the graph is hard to decipher because the colors are so similar and the y axis is unlabelled.

In my opinion, it would be more important to find out the underlying reason for why old versions are being used and solve that problem instead. If you retroactively change old versions, are you certain that people will update those old versions? Will they even see the message and if so, will they act on it? If they act on it, will it actually work for them to update to the most recent one?

If other tools in their environment require a particular (old) version of GTDB-Tk, then Conda won’t let them update.

But I’m just guessing. In short: You should motivate this change a bit more, and perhaps there’s even a different way to solve the problem that doesn’t require to keep that many old recipe versions around.

dpryan79:
You won't always get the most recent version of a package if installing it would require dependencies incompatible with other things in your current environment. For that reason, users are ALWAYS recommended to create new environments rather than trying to continuously install new packages into old environments. `conda create -d -n foo gtdbtk` will install 1.0.2.

aaronmussig:
I should have included the [full graph](https://github.com/bioconda/bioconda-recipes/files/4399954/counts.zip), not just a snippet but it is essentially showing linear growth for the first release of the software which is well over a year old (cumulative downloads on the Y axis). 

While we do recommend people to create it into a new environment, it seems that either due to a malformed base environment or some other reason creating a new environment as `conda create -n foo gtdbtk=1.0.2` would result in them installing version 0.3.2. Unfortunately, this can't be replicated so it's difficult to debug.

In both of your experiences, do you have any suggestions as to why downloads of old versions are still so common? In terms of GTDB-Tk use cases, installing an old version would only make sense if replicating results.

dpryan79:
If a user gets anything other than 1.0.2 with `conda create -n foo gtdbtk=1.0.2` then they have somehow managed to mess conda up beyond repair and should reinstall it from scratch. Otherwise, the cause of this is always using `conda install` in an environment with old packages.

PR #21154, opened 2020-03-27
update recipe for sequana 0.8.2

cokelaer:
**Major update of sequana recipes from 0.7 to 0.8.2**
----

>
>
>
>


recipes/sequana/meta.yaml
```diff
@@ -1,76 +1,88 @@
 {% set name = "sequana" %}
-{% set version = "0.7.1" %}
-{% set sha = "891e9a27fd992a9713de2e3afac9b4c8042d492703be65e1ae2b0ab098e275d6" %}
+{% set version = "0.8.2" %}
+{% set file_ext = "tar.gz" %}
```
marcelm: Please don’t introduce the `file_ext` and `hash_type` Jinja2 variables as they just complicate the recipe. Jinja2 variables should only be added if they are used at least twice.

cokelaer:
@marcelm  I'm going to close this pull request. I do not manage to build this recipes for now. Too many issues right. One issue is with bioservices, a software of mine, that I will update first. Another issue is with cython that is in my requirements but not really required. So I need to cleanup sequana first. The final issue is with multiqc, probably a new version that breaks this recipe.

marcelm:
I added `cython` and  `xmldict` to the run requirements and the test within `conda build` actually pass now. However, it seems that building the container in `mulled-build` seems to be stuck at `Collecting package metadata`. Either there’s some intermediate problem or the resolver is overwhelmed because the recipe has so many dependencies. You may want to try adding some version constraints to some of the dependencies so the resolver doesn’t have to work as much. This will also benefit every user who installs the package :-).

marcelm:
Tests are not running; I’ll try to trigger them.

cokelaer:
@marcelm thanks for the different fixes. locally, it seems to resolve the deps in a reasonnable amount of time. The restrictions you added seem good to me. 

Another point but I do not think this is related: I know that pyqt may cause trouble currently with some version of python (https://github.com/sequana/sequana/issues/574, https://github.com/mantidproject/mantidimaging/issues/339) in particular, I believe that I should build this recipe with a restriction on python<=3.7.3 so I will need to change that in the recipe.

marcelm:
> I know that pyqt may cause trouble currently with some version of python

The `matplotlib-base` dependency doesn’t pull in `pyqt` (see #19996), so perhaps you don’t have that problem. If you actually need PyQt, I recommend that you do not change the dependency back to regular `matplotlib` (because that will trigger the linter), but to add `pyqt` as an explicit dependency.

cokelaer:
@marcelm the linux build is successful but not the osx. Do you have any idea about the error message ? I do not see any obvious reason for the failed build. What would you suggest ? thanks in advance

marcelm:
I think the error was: `...placehold_placehold_p/bin/python: No module named pip` (but the log output is hard to read)

cokelaer:
I see it now, okay tough one. Let us see now. thanks again

marcelm:

cokelaer:
@marcelm thanks so much for your help. This recipe was not easy indeed. I have now merged it.

marcelm:
Yes, this was quite a bit of work I can imagine, glad it worked out in the end :-).

dpryan79:
Nice work everyone!

PR #21163, opened 2020-03-27
Add strling

christopher-schroeder:
Adds strling version 0.1.0, a tools to detect shot tandem repeats

----

>
>
>
>


recipes/strling/0.1.0/meta.yaml
```diff
@@ -0,0 +1,27 @@
+{% set name = "strling" %}
```
marcelm: The recipe should be moved one directory up (it should be at `recipes/strling/meta.yaml`). Subdirectories are generally not used for recipes unless there are good reasons for keeping multiple versions around.

recipes/strling/0.1.0/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
marcelm: Can you try to compile the tool at install? This is generally preferred in Bioconda over packaging a pre-compiled binary. That would also allow you to get a macOS version (the macOS build is currently failing because the binary is for Linux only).

recipes/strling/0.1.0/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+{% set version = "0.1.0" %}
+{% set sha256 = "9da4963574c44ccee529834fd594818955c367abb9083dc0fc972286d6b3f728" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/quinlan-lab/STRling/releases/download/v0.1.0/strling
```
marcelm: You should put the `{{ version }}` into the URL:
```
https://github.com/quinlan-lab/STRling/releases/download/v{{ version }}/strling
```

recipes/strling/0.1.0/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
```
marcelm: If it turns out that you need to package the upstream already compiled binary, then you should remove this dependency.

recipes/strling/0.1.0/meta.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+test:
+  commands:
+    - strling --help
+
+about:
+  home: https://github.com/quinlan-lab/STRling
+  license: MIT
+  summary: STRling (pronounced like “sterling”) is a method to detect large STR expansions from short-read sequencing data.
```
marcelm: A newline is missing at the end of the file.

marcelm:
Hi Christopher :-)

recipes/strling/0.1.0/build.sh
christopher-schroeder: I tried hard. There suggested way of using nimble does not work because of type int64 vs. int mismatches. I could no figure out how to solve this problem.

marcelm: Is the binary statically linked? Then I think it’s ok for now.

marcelm: And add 'skip: True  # [osx]` so the macOS package isn’t built.

christopher-schroeder:
Hey Marcel :)
So I finally managed to get this thing compiled, but doing a static linking of hts-lib ist stupidly complicated in nimble. You have to use the hts_nim_static_builder tool, which itself is a linux binary file and requires a specific docker container for the linking.
The executable compiled with this recipe uses a dynamically linked htslib 10, which requires the LD_LIBRARY_PATH to be set. I wrote a script, that sets the LD_LIBRARY_PATH to the current conda environment before executing the tool.
I don't know if this works for mac-os. I've read that the LIBRARY_PATH has been removed several years ago? But maybe I have outdated information.

marcelm:
So if I understand correctly, to compile a Nim tool, you need to install choosenim, which you download as part of the build script. But then you get this error:
```
/lib64/libc.so.6: version `GLIBC_2.14' not found (required by /tmp/choosenim-0.6.0_linux_amd64)
```
This tells you that the downloaded binary links against a too recent Glibc version. To avoid this, you would need to compile choosenim from source (that is, add a recipe for it to conda-forge). Or perhaps some other package for the Nim compiler, no idea how that ecosystem works. Apparently, no one has done so, yet.

I commented above: "Is the binary statically linked? Then I think it’s ok for now." I’ve checked this now by running `ldd` on the pre-compiled binary from your initial recipe:
```
$ ldd strling 
        not a dynamic executable
```
I think you should just package this static binary, and it’ll be ok. The main problem with precompiled binaries is when they are dynamically linked, in particular against too recent glibc versions (as choosenim), but since it’s entirely static, there is no risk for that. Then when someone packages a nim compiler, this recipe can be switched.

marcelm:
Hm, i was wrong, [nim is in conda-forge](https://github.com/conda-forge/nim-feedstock), so you can just add this as a host requirement and don’t need to compile it yourself.

christopher-schroeder:
Damn you are right. I should have checked that before -.-
Anyway, now it is compiling and running for mac os and linux, so I think this is superior compared to just copying the file. Unfortunately I cannot use the 0.1.0 archive tarball for that, because nim cannot compile it due to errors. Instead I'm pointing at the current git commit, which fixes this error.
So If you don't have any complaints, I think its done now.

recipes/strling/init.sh
```diff
@@ -0,0 +1,181 @@
+#!/bin/sh
+# Copyright 2017 Dominik Picheta and Nim developers.
```
marcelm, line 2: This can now be removed, right?

recipes/strling/build.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+# Compile nim
```
marcelm, line 2: This comment isn’t accurate anymore :-)

recipes/strling/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+{% set sha256 = "6d3aff31de690b47f19dec31617185e7d3547b05d46ccb5a4cf01b3fb5b8c0e0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  - url: https://github.com/quinlan-lab/STRling/archive/614a48877a32f5b2fd628df7a56ce0dfe54bc8df.zip
+    sha256: {{ sha256 }}
+  - url: https://nim-lang.org/download/nim-1.0.6.tar.xz
```
marcelm, line 12: This isn’t necessary anymore, correct?

recipes/strling/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+    sha256: 134e5751cf62fa189f76616a698ce0df4b4a0c1419ec0b582924087e585f66b3
+    folder: nimsrc
+
+build:
+  number: 0
+  skip: True  # [osx]
```
marcelm, line 18: You’re still skipping the macOS build. It is a bit confusing because the test on macOS is marked as passing, but that is only because it successfully detects that it has nothing to do.

recipes/strling/strling.sh
```diff
@@ -0,0 +1,6 @@
+#bin/bash
```
marcelm, line 1: I think this wrapper is not necessary. Conda will usually take care of setting the correct paths.

recipes/strling/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+    - htslib=1.10
+
+test:
+  commands:
+    - strling -h
+
+about:
+  home: https://github.com/quinlan-lab/STRling
+  license: MIT
+  summary: STRling (pronounced like “sterling”) is a method to detect large STR expansions from short-read sequencing data.
```
marcelm, line 36: Please add a final newline.

recipes/strling/build.sh
```diff
@@ -0,0 +1,14 @@
...
+
+cp "${CC}" "${BUILD_PREFIX}/bin/gcc"
+
+mkdir -p $PREFIX/share
+mkdir -p $PREFIX/bin
+cp "$RECIPE_DIR/strling.sh" "$PREFIX/share/strling.sh"
```
marcelm, line 8: See my comment below about the wrapper likely being unnecessary.

christopher-schroeder:
Hmm, I get an
Unsatisfiable dependencies for platform osx-64: {'nim'}
error. What is the best practice in this case? Should I go back to compiling nim or just ignore macos?
Best
Christo

bootstrap.py
```diff
@@ -52,7 +52,7 @@
 
 # TODO: this mimics the override in the "common" job in .circleci/config.yaml
 with open('.circleci/common.sh', 'w') as fout:
-    fout.write("MINICONDA_VER=4.5.4\nBIOCONDA_UTILS_TAG=master\n")
+    fout.write("MINICONDA_VER=4.6.14\nBIOCONDA_UTILS_TAG=master\n")
```
marcelm, line 55: This change cannot be part of this PR.

marcelm:
So you’ll definitely need to stop building nim yourself as part of this PR. Can you push a version where you use `nim` as build dependency?

bootstrap.py
christopher-schroeder: Meh, this should not be in here

christopher-schroeder:
So nim is a dependency and macos is skipped

recipes/strling/meta.yaml
christopher-schroeder: i removed it

recipes/strling/meta.yaml
christopher-schroeder: since nim is not available for macos and via conda and it cannot be compiled, we keep skipping

recipes/strling/strling.sh
christopher-schroeder: you are right

recipes/strling/meta.yaml
christopher-schroeder: added

recipes/strling/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+cp "${CC}" "${BUILD_PREFIX}/bin/gcc"
```
marcelm, line 2: Is this line still necessary?

recipes/strling/build.sh
```diff
@@ -0,0 +1,5 @@
...
+nimble install -y --verbose
+chmod a+x strling
```
marcelm, line 4: This line can also be removed (the created binary is already executable)

recipes/strling/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+cp "${CC}" "${BUILD_PREFIX}/bin/gcc"
+nimble install -y --verbose
```
marcelm, line 3: I tried to use `nimble` locally just now, and I think you should run `nimble build -y --verbose` instead here (with `install`, it will also copy a symlink to `~/.nimble/bin`, which isn’t really needed).

recipes/strling/build.sh
christopher-schroeder: Apparently it is not.

recipes/strling/build.sh
christopher-schroeder: right

recipes/strling/build.sh
christopher-schroeder: also right

marcelm:
Now I need to mention one more thing: It’s Bioconda policy not to package pre-releases, and technically, you’re no longer packaging version 0.1.0 of strling. However, as I see it, this is essentially 0.1.0 plus a single patch to make it work with htslib 1.10. This is so close that I think it’s fine, considering that Bioconda recipes also add patches sometimes and still leave the version number unchanged. So if the test passes, I think this is good to go.

christopher-schroeder:
Thank you a lot!

marcelm:
@BiocondaBot please merge

PR #20874, opened 2020-03-13
Updated blast+ to 2.10.1 please merge soon

ialbert:
updated blast version to 2.10.0 

the change is urgent as blast 2.9.0 remote access is broken. See:

https://www.biostars.org/p/400841/

> *Date: Thu, 10 Oct 2019 07:40:48 ... Thank you for the report. There is a bug in 2.9.0 when using the '-remote' option. It should work in 2.8.1, and it has been fixed for the upcoming 2.10.0 release, which should be out in 2 or 3 weeks."*

No blast search can be run remotely on the current blast version.

**Replace this sentence with a detailed description of your pull request**

----

> Bioconda requires reviews prior to merging pull-requests (PRs). To facilitate this, once your PR is passing tests and ready to be merged, please add the `please review & merge` label so other members of the bioconda community can have a look at your PR and either make suggestions or merge it.

* [x] I have read the [guidelines for bioconda recipes](https://bioconda.github.io/contributor/guidelines.html).
* [ ] This PR adds a new recipe.
* [ ] AFAIK, this recipe **is directly relevant to the biological sciences**
      (otherwise, please submit to the more general purpose [conda-forge channel](https://conda-forge.org/docs/)).
* [x] This PR updates an existing recipe.
* [ ] This PR does something else (explain below).

> Everyone has access to the following BiocondaBot commands, which can be given in a comment:
>
>  * `@BiocondaBot please update` will cause the BiocondaBot to merge the master branch into a PR
>  * `@BiocondaBot please add label` will add the `please review & merge` label.
>  * `@BiocondaBot please fetch artifacts` will post links to packages and docker containers built by the CI system. You can use this to test packages locally before merging.
>
> For members of the Bioconda project, the following command is also available:
>
>  * `@BiocondaBot please merge` will cause packages/containers to be uploaded and a PR merged. Someone must approve a PR first! This has the benefit of not wasting CI build time required by manually merging PRs.
>
> If you have questions, please post them in gitter or ping `@bioconda/core` in a comment (if you are not able to directly ping `@bioconda/core` then the bot will repost your comment and enable pinging).

PR #21180, opened 2020-03-29
CircleCI: macOS: use 10.13 SDK

mbargull:
cc @bioconda/osx
Update the macOS SDK version from `10.9` to ~~`10.10`~~`10.11`.
In gh-19610 we added the SDK download for `10.9` (same version as used by conda-forge) to ensure compatibility for older Mac systems. Before gh-19610 (and gh-19518), we used the SDK provided by the CI image, which was `10.12` acc. to https://circleci.com/gh/bioconda/bioconda-recipes/75813 :
> 23:29:01 BIOCONDA INFO (OUT) +CONDA_BUILD_SYSROOT=/Applications/Xcode-8.3.3.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.12.sdk

I use `10.10` \[edit: now `10.11`, see https://github.com/bioconda/bioconda-recipes/pull/21180#issuecomment-606598822 \] because that is the lowest version that seems to be needed to build `blast` in https://github.com/bioconda/bioconda-recipes/pull/20874 .
At some point we may want to support using other SDK versions on a per-recipe basis, but https://gitter.im/bioconda/Lobby?at=5e4e980ac2c73b70a4491683 :
> >@milot-mirdita I don't think so, we hard-coded that to 10.9 to match conda-forge. Maybe that's something @mbargull can think about implementing (time permitting).
> 
> @dpryan79, @milot-mirdita
> That's certainly possible to do, but this would be a rather low priority item. We may have to change a few bits here and there for this to work: it might require changing settings outside of the Conda build setup between builds of different recipes, IIUC. (On conda-forge we always build each recipe in a separate CI job so setting the SDK version only has to happen once when setting up the build.)
> If we happen to have high demand for this at some point, we could prioritize this more -- but right now I can't tell when we'll be able to get this implemented.

Note: Even after this PR we still use a deployment target of `10.9`. Due to that "deployment target", we could in theory use even newer SDKs. However, we may want to be conservative with what we use due to https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/67#issuecomment-596270454 :
> Issue is that most autotools based packages would check whether functions exist and using a newer SDK would make them available even if they are weak symbols. (For eg: clock_gettime is in 10.12, but autotools detects it even if target is set to 10.9 if the SDK is 10.12 or higher)

----

>
>
>
>


mbargull:
> Note: Even after this PR we still use a deployment target of `10.9`.

Oh, hold on, this might not be true. I'll add another commit to address this.

.circleci/setup.sh
```diff
@@ -79,8 +79,8 @@ if ! type bioconda-utils 2> /dev/null || [[ $BOOTSTRAP == "true" ]]; then
     # step 3: install bioconda-utils
     additional_packages='git pip'
     if [[ $OSTYPE == darwin* ]]; then
-        # Pinned to 2.5.3 to make sure we don't get unexpected changes.
-        additional_packages="${additional_packages} conda-forge-ci-setup=2.5.3"
+        # Pinned conda-forge-ci-setup to a specific version to make sure we don't get unexpected changes.
+        additional_packages="${additional_packages} conda-forge-ci-setup=2.6.0"
```
mbargull: `2.5.3` and `2.6.0` should work the same apart from that https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/69 is included, which may come in handy if/when we support per-recipe SDK choices and have packages that need newer versions.

dpryan79:
Approving so you can merge when you're happy. I foresee us needing to implement per-recipe SDKs in the not too distant future, or at least enabling a single newer alternative (10.12 or 10.13 are reasonably new).

mbargull:
> Approving so you can merge when you're happy.

Thanks Devon!
> at least enabling a single newer alternative (10.12 or 10.13 are reasonably new).

Yeah, AFAIU, Apple does not have "extended" support releases and with only gives support for \~3 years with their latest 3 versions, i.e., `10.15`/`10.14`/`10.13` currently, right?
Hence, I'm not sure what the benefit is for us to support much older versions. Maybe for some institutes that have old-and-never-updated computers hooked up to some machines (akin to those still running WinXP...)? IDK. As a non-Apple person, it's hard for me to assess what makes sense for us to do here. Maybe someone with more insights can give us pointers in the future...

dpryan79:
Maybe we can put this out to the community together with conda-forge. If the community doesn't really need these older versions supported then we could migrate both projects to something newer. Perhaps some joint tweets and a survey?

mbargull:
Yep, one of my first thoughts was to bring this up to conda-forge and if they decided to update their default SDK version to a more recent one, it'd be much easier to reason this like "we're following along".

mbargull:
I have no idea how accurate this is, but searching for "mac supported macos version" gave me this as the first result: https://eshop.macsales.com/guides/Mac_OS_X_Compatibility
Acc. to that, `10.11` might be a reasonable choice if we want to support Macs from 2008/2009.

dpryan79:
I wouldn't even bother going that far back, macs tend to be updated more frequently. So I'd say maybe just the last 5 years.

mbargull:
> macs tend to be updated more frequently.

o_O ... can I get back to the "instead invest in Bioconda & co." campaign again, then?! XD

----
Okay, let's at least go for `10.11` in this PR now then, since that one still seems like a very conservative choice but probably is not more \*cough\* "disruptive" \*cough\* than going for `10.10`.
We can change it to `10.12`/`10.13` in a subsequent PR when we get more information from macOS folks or conda-forge.

Juke34:
Any decision made about integrating the new SDK version? It sounds important for several recipes. I'm thinking in particular to blast+

mbargull:
@bioconda/core, @bioconda/osx, anyone opposed to this change?

This now uses `10.13` for the SDK because:
- I only have heard (multiple times) that it's unlikely that people run old Mac hardware (at least not from before 2010), so it's probably fine to go past `10.11`, thus potentially dropping support for older Macs.
- Given the aforementioned "Mac OS Compatibility Guide", it doesn't seem to matter if we go with `10.12` or `10.13`. Since `10.12` is EOL already, `10.13` makes more sense.
- We still set the deployment target to `10.9` for compatibility (though, acc. to https://github.com/conda-forge/conda-forge-ci-setup-feedstock/issues/67#issuecomment-596270454 this might not be honored in all cases).

dpryan79:
I'm a definite :+1: on this, it'll enable things like updating blast :)

mbargull:
Let's do it then \\(^o^)/

PR #21202, opened 2020-03-30
PhyloPhlAn

fasnicar:
**Replace this sentence with a detailed description of your pull request**

----

>
>
>
>


marcelm:
For testing, you can temporarily skip the linter check:
```
skip-lints:
  - uses_vcs_url
  - missing_hash
```
But the final recipe should use a proper release.

And an observation on your `setup.py`: The `check_call` looks very fragile. If you get weird failures from `pip install`, this could be the reason.

fasnicar:
Thank you very much Marcel for your suggestions.
I'm gonna change the check_call() to call(). The script run there is needed to generate standard configuration files tailored for the machine where the package will be installed, but it shouldn't be blocking if none of the config files can be generated.

marcelm:
Sorry I should have been more explicit: The problem is running something unconditionally on every `setup.py` invocation since it should only run when specific subcommands are run. For example, the `call()` will now also be done when running `python3 setup.py --version`. The proper way would be to customize one of the subcommands (perhaps `sdist`), but I guess you’re fine at the moment since it doesn’t hurt to regenerate the files. But as I said, just something to keep in mind in case you see weird failures.

fasnicar:
Ah, I see. Thanks for the heads up, I'll check it out and update on the next release.

PR #21203, opened 2020-03-30
add plasmidid recipe

pedroscampoy:
This pull request add a new recipe for PlasmidID, a computational pipeline that identifies annotate and reconstruct plasmids from WGS data

----

>
>
>
>

### DIFF
--- a/recipes/plasmidid/build.sh
+++ b/recipes/plasmidid/build.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+mkdir -p $PREFIX/bin/config_files
+mkdir -p $PREFIX/config_files
+
+cp bin/* $PREFIX/bin
+cp plasmidID $PREFIX/bin
+
+cp config_files/* $PREFIX/bin/config_files
+
+echo "INSTALING MASH 2"
+echo "#############################################"
+
+git clone https://github.com/marbl/Mash.git
+
+cd Mash
+
+./bootstrap.sh
+./configure --with-capnp=$PREFIX --with-gsl=$PREFIX --prefix=$PREFIX
+make
+make install
+
+echo "INSTALING CIRCOS WITH CUSTOM SETTINGS"
+echo "#############################################"
+
+wget http://circos.ca/distribution/circos-0.69-8.tgz
+tar zxvf circos-0.69-8.tgz
+cd circos-0.69-8
+sed -i "s/max_points_per_track =.*/max_points_per_track = 80000000/g" etc/housekeeping.conf
+cp -r . $PREFIX
--- a/recipes/plasmidid/build.sh
+++ b/recipes/plasmidid/build.sh
@@ -7,19 +7,3 @@ cp bin/* $PREFIX/bin
 cp plasmidID $PREFIX/bin
 
 cp config_files/* $PREFIX/bin/config_files
-
-echo "INSTALING MASH 2"
-echo "#############################################"
-
-wget https://github.com/marbl/Mash/releases/download/v2.2/mash-Linux64-v2.2.tar
-tar -xvf mash-Linux64-v2.2.tar
-cp mash-Linux64-v2.2/* $PREFIX/bin
-
-echo "INSTALING CIRCOS WITH CUSTOM SETTINGS"
-echo "#############################################"
-
-wget http://circos.ca/distribution/circos-0.69-8.tgz
-tar -zxvf circos-0.69-8.tgz
-cd circos-0.69-8
-sed -i "s/max_points_per_track =.*/max_points_per_track = 80000000/g" etc/housekeeping.conf
-cp -r . $PREFIX
--- a/recipes/plasmidid/build.sh
+++ b/recipes/plasmidid/build.sh
@@ -11,20 +11,15 @@ cp config_files/* $PREFIX/bin/config_files
 echo "INSTALING MASH 2"
 echo "#############################################"
 
-git clone https://github.com/marbl/Mash.git
-
-cd Mash
-
-./bootstrap.sh
-./configure --with-capnp=$PREFIX --with-gsl=$PREFIX --prefix=$PREFIX
-make
-make install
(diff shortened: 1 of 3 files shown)


recipes/plasmidid/build.sh
```diff
@@ -0,0 +1,30 @@
...
+
+cp bin/* $PREFIX/bin
+cp plasmidID $PREFIX/bin
+
+cp config_files/* $PREFIX/bin/config_files
+
+echo "INSTALING MASH 2"
+echo "#############################################"
+
+git clone https://github.com/marbl/Mash.git
```
marcelm: This isn’t reproducible. You need to at least make sure you use a certain revision.

What was the problem with specifying mash as a normal dependency?

pedroscampoy: Thank you for this comment. Now mash is installed from a reproducible source.
Indeed, mash is very problematic. By default conda installs v1.1 and version 2.2 can't be forced, it appears in conda search list but won't install.
If I add mash as a normal dependency basically all packages return a version conflict. This problem has to be addressed, but since the recipe seems fine, I don't know who should fix it.

druvus: What happens if you specify `mash >= 2.2`?

pedroscampoy: In the recipe `conda-build` thows an error like that:

```
Package curl conflicts for:
prokka -> blast[version='>=2.7.1'] -> curl[version='>=7.64.1,<8.0a0']
samtools -> curl[version='>=7.59.0,<8.0a0|>=7.64.0,<8.0a0|>=7.64.1,<8.0a0']
blast -> curl[version='>=7.64.1,<8.0a0']
bedtools -> curl[version='>=7.64.1,<8.0a0']
```

For every package and dependency, either with mash and mash >= 2

With` conda install `also fails to install, I have to specify the build (mash=2.2.2=h3d38be6_1) and add -c conda-forge to install it properly, but within the recipe I get the previous error for every package.

marcelm: Since you mention that you have to specify `-c conda-forge` when you ran `conda install`: Did you perhaps leave out `conda-forge` when you ran `conda build`? Bioconda requires conda-forge, so if this fails, it’s not a suprise. I suggest you [add both conda-forge and bioconda](https://bioconda.github.io/user/install.html#set-up-channels) in the correct order to your `.condarc`.

marcelm: I’ve just tested the recipe locally and when I add `mash` to the dependencies and remove the corresponding part from `build.sh`, it works fine.

I also ran `conda create -n tmp mash` just now and it works fine (with `conda-forge` and `bioconda` in the channel list). Note that when you install packages in Conda, it’s always recommended that you create a new environment to avoid potential conflicts with the packages already installed. These conflicts could explain the trouble you had installing it.

recipes/plasmidid/build.sh
```diff
@@ -0,0 +1,25 @@
...
+echo "#############################################"
+
+wget https://github.com/marbl/Mash/releases/download/v2.2/mash-Linux64-v2.2.tar
+tar -xvf mash-Linux64-v2.2.tar
+cp mash-Linux64-v2.2/* $PREFIX/bin
+
+echo "INSTALING CIRCOS WITH CUSTOM SETTINGS"
+echo "#############################################"
+
+wget http://circos.ca/distribution/circos-0.69-8.tgz
```
marcelm: This is the same problem as with Mash: Have you tried just adding a `circos` dependency? There exists a package for it. If the custom settings are the only reason for doing it like this, there must be a different way.

If you include other packages in your recipe like this, you get all sorts of problems. For example, if someone installs both plasmidid and circos into the same environment, the files conflict with each other. Perhaps it could even happen that someone explicitly installs a circos version that is not 0.69, but then gets the version from plasmidid.

recipes/plasmidid/build.sh
```diff
@@ -0,0 +1,25 @@
...
+tar -zxvf circos-0.69-8.tgz
+cd circos-0.69-8
+sed -i "s/max_points_per_track =.*/max_points_per_track = 80000000/g" etc/housekeeping.conf
```
marcelm: I haven’t used Circos myself, only briefly looked at the tutorial right now. From that, I wonder whether you actually should modify the `etc/housekeeping.conf` file, no matter in which way you install plasmidID. Wouldn’t it be possible tojust add
```
max_points_per_track = 80000000
```
after each
```
<<include etc/housekeeping.conf>>
```
in the Circos configurations that are part of plasmidID?

pedroscampoy: Thank you for the guidance.
-Mash now works by specifying conda-forge channel (I swear I tried already).
-Circos is now installed properly and the specific line in housekeeping.conf is modifyed in post-link.sh file. This line doesn't afect circos overal performance.

marcelm: Thanks for the changes. I’m not entirely sold, yet, on using `post-link.sh` to modify another package’s configuration. Perhaps a second opinion is necessary. @dpryan79, do you have time to give us your thoughts on this?

dpryan79:

recipes/plasmidid/build.sh
dpryan79: Am I correct in understanding that the post-link script is modify a circos conf file? I'm really uncomfortable doing that, packages should really not modify each other.

marcelm: Yes, that’s what the script does. @pedroscampoy What about the suggestion to change plasmidID itself so that it sets this configuration value correctly? Or would that not work?

pedroscampoy: I have adapted the code so I no longer have to modify any circos config file. Thank to your suggestions, they have been very helpful.
Please review the recipe whenever is convenient for you.

dpryan79:
@bioconda-bot please merge

PR #21226, opened 2020-04-01
Add MSAProbs recipe

warownia1:


foreveremain:
Hi @warownia1 - I think you need openmp as a dependency (at least for building) MSAprobs.

warownia1:
@BiocondaBot please add label

recipes/msaprobs/build.sh
```diff
@@ -0,0 +1,5 @@
+#! /bin/bash
+
+mkdir -p $PREFIX/bin
+(cd $SRC_DIR/MSAProbs && make all)
```
marcelm, line 4: Since the `build.sh` script starts in the unpacke source dir, you can simplify this and the next line slightly to
```
cd MSAProbs
make all
cp msaprobs $PREFIX/bin/
```
Note also that the build script is called with the Bash option `-o errexit`, which means that you don’t need to connect commands with `&&`.

recipes/msaprobs/meta.yaml
```diff
@@ -0,0 +1,40 @@
+package:
+  name: msaprobs
+  version: 0.9.7
```
marcelm, line 3: You should introduce a Jinja2 variable for the version. At the top of the file, add
```
{% set version = "0.9.7" %}
```
Then you can write here:
```
version: "{{ version }}"
```

recipes/msaprobs/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+source:
+  url: https://sourceforge.net/projects/msaprobs/files/MSAProbs-0.9.7.tar.gz
```
marcelm, line 6: ... and then here write this as
```
url: https://sourceforge.net/projects/msaprobs/files/MSAProbs-{{ version }}.tar.gz
```
In that way, the version can be changed in a single position when a new release is mode.

recipes/msaprobs/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+  host:
+    - llvm-openmp  # [osx]
+    - openmp  # [osx]
+  run:
+    - llvm-openmp  # [osx]
+    - openmp  # [osx]
+
+test:
+  commands:
+    - msaprobs -version 2>&1 | grep "version 0.9.7"
```
marcelm, line 29: Use also `grep "version {{ version }}"` here

recipes/msaprobs/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+about:
+  home: http://msaprobs.sourceforge.net/homepage.htm
+  license: "General Public License v3.0"
+  author: Yongchao Liu
```
marcelm, line 34: The GPL 3 requires that a copy of the license is distributed along with the Conda package. Does the `.tar.gz` include a copy of the GPL? Then use `license_file: LICENSE.txt` (or whatever the filename) here.

dpryan79:
@bioconda-bot please merge

PR #21215, opened 2020-03-31
Add TreeSAPP recipe

cmorganl:
I am adding a bioconda recipe for TreeSAPP, a tool for annotating proteins derived from genomes and metagenomes with phylogenetic placement methods.

----

Please read the [guidelines for Bioconda recipes](https://bioconda.github.io/contributor/guidelines.html) before opening a pull request (PR).

* If this PR adds or updates a recipe, use "Add" or "Update" appropriately as the first word in its title
* New recipes not directly relevant to the biological sciences need to be submitted to [conda-forge channel](https://conda-forge.org/docs/) instead of Bioconda
* PRs require reviews prior to being merged. Once your PR is passing tests and ready to be merged, please add the `please review & merge` label.
* Please post questions [on Gitter](https://gitter.im/bioconda/Lobby) or ping `@bioconda/core` in a comment.

<details>
  <summary>BiocondaBot commands</summary>
      
Everyone has access to the following BiocondaBot commands, which can be given in a comment:

  * `@BiocondaBot please update` will cause the BiocondaBot to merge the master branch into a PR
  * `@BiocondaBot please add label` will add the `please review & merge` label.
  * `@BiocondaBot please fetch artifacts` will post links to packages and docker containers built by the CI system. You can use this to test packages locally before merging.

For members of the Bioconda project, the following command is also available:

 * `@BiocondaBot please merge` will cause packages/containers to be uploaded and a PR merged. Someone must approve a PR first! This has the benefit of not wasting CI build time required by manually merging PRs.
</details>

PR #21241, opened 2020-04-02
Update anvio-minimal to v6.2

meren:
Updates the anvio-minimal recipe to match the latest release of anvi'o, a software platform directly relevant to the biological sciences.

### DIFF
--- a/recipes/anvio-minimal/meta.yaml
+++ b/recipes/anvio-minimal/meta.yaml
@@ -40,7 +40,7 @@ requirements:
     - statsmodels
     - snakemake-minimal ==5.2.4
     - colored
-    - illumina-utils
+    - illumina-utils ==2.6
     - tabulate
     - numba
 test:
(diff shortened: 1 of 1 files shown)


druvus:
@dpryan79 Do you why we are running into the pypy dependency issue when python is specified to 3.6? Similar issue as was also seen in #21215

dpryan79:
I think the combination of the skip with a package that should have been noarch confused conda-build. The pypy stuff should be mostly fixed now, but if conda-build gets confused I think that still pops up. In general, the "skip" syntax should only be used in the case that there are compilers getting used.

meren:
Thank you both very much for your help with this, @druvus and @dpryan79. I will now try to relax the dependency requirement for one package (`illumina-utils`, which has been failing due to `pypy`) and see if the recipe continues to pass the checks.

I really appreciate the help. Thank you.

meren:

dpryan79:
@bioconda-bot please merge

meren:
Hi @dpryan79, 

Thank you for merging this. Although, I was still testing the CircleCI built image to see if there are any libraries that require specific versions :)

I just realized that it is indeed the case for `cherrypy` and `snakemake`, and they must be set to the following two versions:

```
cherrypy ==8.0.0
snakemake ==5.10.0
```

I guess I need to send another PR with correct versions after increasing the build number, is it correct?

Thank you for your help,

dpryan79:
Yes, please just open a new PR and we can update that.

meren:
Thank you, @dpryan79. The updated PR is at #21281.

PR #21626, opened 2020-04-20
Update radsex to 1.0.0

RomainFeron:
Updated meta for 1.0.0.
Cleaned up build script after changes to radsex build system.

### DIFF
--- a/recipes/radsex/build.sh
+++ b/recipes/radsex/build.sh
@@ -1,7 +1,10 @@
 #!/bin/bash
-pushd include/bwa
-make CC=$CC
-popd
-make CC=$CXX CFLAGS="$CXXFLAGS" LDFLAGS="$LDFLAGS -pthread -static-libstdc++ -lz"
+
+# Create bin directory
 mkdir -p $PREFIX/bin/
-cp bin/radsex $PREFIX/bin/
+
+# Build radsex specifying CPP compiler and C compiler
+make CXX=$CXX CC=$CC
+
+# Copy all generated binaries to bin
+cp bin/radsex $PREFIX/bin/radsex
--- a/recipes/radsex/meta.yaml
+++ b/recipes/radsex/meta.yaml
@@ -7,6 +7,7 @@ package:
 
 build:
   number: 0
+  skip: True  # [osx]
 
 source:
   url: https://github.com/SexGenomicsToolkit/radsex/archive/{{ version }}.tar.gz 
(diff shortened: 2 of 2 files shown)


RomainFeron:
I've had to disable building on osx because of clang not recognizing 'uint', 'ulong' (works with clang v6 but not the v4 which is used to build on bioconda). If there is a solution to bypass this issue, it'd be nice, otherwise I can modify radsex's code to use 'unsigned int' everywhere in the future.

HenningTimm:
I have little experience with osx, but I have seen recipes (e.g. [pretextmap](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/pretextmap/meta.yaml)) that pin a clang version in the build requirements. So maybe something like this

    requirements:
      build:
        - {{ compiler('c') }}
        - {{ compiler('cxx') }}
        - clang>=6.0.0  # [osx]

could help.

RomainFeron:
Thanks for the suggestion, I've tried to do that and also only load {{ compiler('cxx') }} for linux64 like in the pretextmap recipe, but I'm still getting build failure on osx:

06:41:50 BIOCONDA INFO (ERR) conda.exceptions.UnsatisfiableError: The following specifications were found to be in conflict:
06:41:50 BIOCONDA INFO (ERR)   - clang[version='>=6.0.0']
06:41:50 BIOCONDA INFO (ERR)   - clang_osx-64=4 -> clang=4.0.1

Do you have an idea why this is happening? I don't understand why it seems to work for pretextmap but not in this recipe.

HenningTimm:
Sorry, no idea :(

RomainFeron:
Then I will disable building on osx for now, so 1.0.0 can be available on linux at least

RomainFeron:
Would it be ok to merge now and I'll see if I can figure a solution for osx in the future?

RomainFeron:
You're correct; I chose to go with this versioning scheme because the changes were specifically for building in bioconda and I wanted to keep the release number to 1.0.0 for the official release. I used github tags to get around that, but I agree it could be confusing and I could also just delete the tags and create a new 1.0.0 release (as it hasn't been announced yet, it'd be fine)

dpryan79:
@bioconda-bot please merge

EdHarry:
@RomainFeron I may be very late to this, but you'll need to add `# [linux64]` to `{{ compiler('c') }}` as well as `{{ compiler('cxx') }}`, since the default C compiler for OSX is also clang-4

RomainFeron:
Hi @EdHarry ,
Sorry for the late response; I'm not sure I get what you're suggesting, at the moment the recipe works on linux and is disabled on OSX. What would be the result of adding this instruction?

PR #21806, opened 2020-04-28
Add cnvnator

christopher-schroeder:
Add cnvnator. First commit is CI test, recipe not clean yet.


recipes/cnvnator/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  - url: https://github.com/abyzovlab/CNVnator/releases/download/v0.4.1/CNVnator_v{{ version }}.zip
+    sha256: a2b3d85909ac5533ed079440000fdbf15aede7a56ec1ca555ae185a4162e8675
+  - url: https://github.com/samtools/samtools/releases/download/1.10/samtools-1.10.tar.bz2
+    sha256: 7b9ec5f05d61ec17bd9a82927e45d8ef37f813f79eb03fe06c88377f1bd03585
+    folder: samtools
```
johanneskoester, line 13: Why is this needed? You seen to use htslib.a correctly from the htslib dependency.

christopher-schroeder: Well, it took me 6 days to get this compiled and maybe there is a way to make it better.
I could modify compiling process to make use of htslib.a from the conda installation. But it also requires a compiled version of samtools as a softlink. It makes use of libbam.a (and probably more files), which is not included in the conda installation of samtools. So the only option I could see is following the official installation guide, by compiling samtools by myself.

johanneskoester: I guess then the solution would be to modify the samtools and htslib packages to provide the required libs?

christopher-schroeder: Yes, that would probably be a solution, but I don't want to modify such essential packages, only because one random tool is (in my opinion) misconfigured and requires some intermediate files, that are generated during samtools compile.

dpryan79: We will never provide libbam.a from samtools because no one is supposed to use it. Please suggest to the cnvnator package that the samtools authors really hate it when people do that.

christopher-schroeder: I hate that too, but they won't listen. They are currently busy programming cnvpytor. Also I cannot wait for weeks just to hope for a cleaner installation implementation. If you, dpryan, decline Johannes request, then I don't see any other way of compiling samtools, just to get the libbam.a file.

dpryan79: There's nothing more you can do about that then. Let's not let the perfect be the enemy of the good :)

dpryan79:

dpryan79:
@bioconda-bot please merge

PR #21817, opened 2020-04-28
Update GeMoMa recipe

JensKeilwagen:
reduced the requirements of the python version to 3.0

@BiocondaBot please update
@BiocondaBot please add label


dpryan79:
This doesn't actually do anything other than bump the build number. The python version restriction is already present in the current package.

JensKeilwagen:
Hi Devon,

You're right. Although we only need python version 3, the test was not successful. Hence, I changed back to 3.7. However, if you have an idea, I would be happy to hear it.

Stay healty and best regards, Jens

PR #21968, opened 2020-05-06
gitlab changed the url (archive -> releases)

pmjklemm:
Simple change of the url in meta.yaml


pmjklemm:
wops, my bad, archive still works

Issue #21982, opened 2020-05-06
How do I update a recipe for an old version of tool?


nh13:
Currently, only versions `1.3.3` and `1.3.4` (d) are available for [stringtie](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/stringtie/).  I would like to use bioconda to instal `1.3.3b`.  The motivation is to have a reproducible environment for a clinical pipeline.  Upgrading to `1.3.4` or `1.3.3` will not reproduce the current clinical environment (non-starter).  

I think it would be pretty easy for me to submit a PR that takes the commit prior to the `1.3.4` version bump and modify the `meta.yaml` to version `1.3.3b`.  Is this possible without making the latest recipe `1.3.3b`?  Not sure how to do this, but I could see doing this a non-trivial amount of times in the future for other software as I try to reproduce environments using `bioconda` and discovering missing recipes for old versions.

nh13:
Thanks @johanneskoester for pointing me to: https://bioconda.github.io/contributor/guidelines.html#versions

PR #22010, opened 2020-05-07
Add stringtie version 1.3.3b

nh13:
Adding a previous version of stringtie (1.3.3b) that I need to use to reproduce a previous analysis.  Related to https://github.com/bioconda/bioconda-recipes/issues/21982

I copied the `1.3.3` directory, modified the URL and SHA256 checksum.  That's it!

### DIFF
--- a/recipes/stringtie/1.3.3b/build.sh
+++ b/recipes/stringtie/1.3.3b/build.sh
@@ -5,7 +5,7 @@ export LIBRARY_PATH=${PREFIX}/lib
 
 wget http://ccb.jhu.edu/software/stringtie/dl/prepDE.py
 
-make release
+make release CC=$CXX LINKER=$CXX
 mkdir -p $PREFIX/bin
 mv stringtie $PREFIX/bin
 
--- a/recipes/stringtie/1.3.3b/build.sh
+++ b/recipes/stringtie/1.3.3b/build.sh
@@ -2,6 +2,8 @@
 export C_INCLUDE_PATH=${PREFIX}/include
 export CPLUS_INCLUDE_PATH=${PREFIX}/include
 export LIBRARY_PATH=${PREFIX}/lib
+export CFLAGS="$CFLAGS -I${C_INCLUDE_PATH}"
+export LDFLAGS="$LDFLAGS -L$LIBRARY_PATH"
 
 wget http://ccb.jhu.edu/software/stringtie/dl/prepDE.py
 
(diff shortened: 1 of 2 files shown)


nh13:
@BiocondaBot please add label

dpryan79:
Do you really need this old version for some reason? I'd prefer to get rid of 1.3.3 then at the same time unless it's needed for something.

nh13:
@dpryan79 yes I really do need this version for a clinical pipeline that I am moving to a reproducible build (wanting to use bioconda).  Thanks!

dpryan79:

nh13:
@dpryan79 any reason why Python 3 isn’t supported, if I interpret Python<3 correctly? The 1.3.3 recipe doesn’t require that I don’t think.

nh13:
And thank you for pushing this!

dpryan79:
@nh13 Your original PR had it as python 2 only, I just changed it so it would build. If I recall, the python wrappers may not have been updated for python 3 (someone would have to look).

dpryan79:
Looks like you'll have to try putting $CXX into a few make files

nh13:
I can try to take a look tomorrow. I wonder if the 1.34d build could help us out in terms of what’s needed.

nh13:
@dpryan79 I got a little farther, but I am having trouble [finding exactly where to inject CXX](https://bioconda.github.io/contributor/troubleshooting.html#g-or-gcc-not-found) where I haven't already.  Do you have time to take a look?

nh13:
I think the latest commit will work because the stringtie Makefile uses `CC` variable for the `CXX` compiler.  Bioinformatics...

nh13:
@dpryan79 Look at the 1.3.3 sub-directory, which my original commit 5226f27 modified in this directory.  I just tested this on my local machine and both worked:
```
conda create -n stringtie -c bioconda stringtie=1.3.3 python=3
conda create -n stringtie -c bioconda stringtie=1.3.3 python=2
```
So it looks like that recipe built both python 2 & 3.  So I am not sure how 1.3.3 is working with python 3, but then [you said](https://github.com/bioconda/bioconda-recipes/pull/22010#issuecomment-626345276) this one doesn't?

nh13:
I stand by that this PR should go just with 5226f27 as it is the minimal modification of what's there in the 1.3.3 build.  The makefiles in 1.3.3 and 1.3.3b are the same FYI.

dpryan79:

dpryan79:
@bioconda-bot please merge

PR #22030, opened 2020-05-08
Add Virchip

mehrankr:
Added the Virchip recipe containing build.sh and meta.yaml

### DIFF
--- a/recipes/virchip/build.sh
+++ b/recipes/virchip/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON -m pip install -vv --disable-pip-version-check --no-deps --no-cache-dir --ignore-installed . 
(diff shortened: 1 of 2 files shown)


mehrankr:
@BiocondaBot please add label

dpryan79:
If you're in the Hoffman lab, can you please consider testing this on a version of python that's still supported? In the not too distant future we will disable python 2 by default, because it should not be used by anyone any more.

mehrankr:
> If you're in the Hoffman lab, can you please consider testing this on a version of python that's still supported? In the not too distant future we will disable python 2 by default, because it should not be used by anyone any more.

Hey Devon. Thanks for reviewing the request. The models trained with Virchip in 2016 used an old verison of scikit-learn and the models are not compatible with newer versions of scikit-learn and python3 unfortunately.
I will try to modify this in the future but would it be possible to have this repository available at the moment?

Thanks a lot
Mehran

dpryan79:

dpryan79:
@bioconda-bot please merge

PR #22194, opened 2020-05-18
Add pbipa-0.0.1

pb-cdunn:
New package: pbipa (PacBio Improved Phased Assembler)

PR #22199, opened 2020-05-19
Add pbipa

pb-cdunn:
Same as https://github.com/bioconda/bioconda-recipes/pull/22194

This PR should allow reviewers to update, in case that helps.

But I think we are having a problem with *snakemake* as a runtime dependency.


pb-cdunn:
Closing in preference to #22194

PR #22600, opened 2020-06-09
Add ANARCI

sdvillal:
This is an attempt at packaging ANARCI (Antibody Numbering and Receptor ClassIfication).

Note that at the moment they do not provide release versions, so I have gone with the date of the commit.

### DIFF
--- a/recipes/anarci/meta.yaml
+++ b/recipes/anarci/meta.yaml
@@ -28,9 +28,10 @@ build:
     - rm -f ${PREFIX}/bin/muscle ${PREFIX}/bin/muscle_macOS
 
 requirements:
+  build:
+    - hmmer >=3.1
   host:
     - python >=3.6
-    - hmmer >=3.1
   run:
     - hmmer >=3.1
     - biopython
--- a/recipes/anarci/meta.yaml
+++ b/recipes/anarci/meta.yaml
@@ -13,7 +13,7 @@ source:
 
 build:
   number: 0
-  noarch: python
+  # noarch: python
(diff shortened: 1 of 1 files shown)


sdvillal:

sdvillal:

sdvillal:
@BiocondaBot please add label

recipes/anarci/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+source:
+  url: https://github.com/oxpig/ANARCI/archive/{{ commit }}.zip
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+  noarch: python
+  script:
+    - "{{ PYTHON }} setup.py install"
```
dpryan79, line 18: Can you change this to `{{ PYTHON }} -m pip install . -vv --no-deps` and then remove `hmmer` from the host section?

sdvillal: I actually tried first with the "pip install" command, but that makes the building process, that generates the HMMs, not work well. I think it probably is an issue upstream.

I will anyway try to see if I can honor some of your suggestions.

dpryan79: Ah, good to know. If that fails then please just revert. I'll approve the PR so it can get merged either way.

sdvillal:
@dpryan79 I have commented why we need the non-standard install command and moved hmmer from host to build dependency, that I guess conveys better the fact that hmmer is needed both at build and runtime.

Green tick! :-)

Thanks for reviewing.

dpryan79:

dpryan79:
@bioconda-bot please merge

Thanks @sdvillal !

PR #22694, opened 2020-06-15
Tidy up lofreq recipe

jmarshall:
PR #22634 changed _build.sh_'s `--with-htslib` option to `--with-htslib=${PREFIX}/lib` without any explanation. This setting is incorrect as giving the _lib_ subdirectory does not supply the header files, so the build succeeded only because it implicitly fell back to use the build-env's system HTSlib from $CPPFLAGS/$LDFLAGS.

As requested on PR #22693, this reverts the recipe to use `--with-htslib=system` to explicitly use only the HTSlib already installed in the build-env:

> --with-htslib=system
> Ignores any nearby HTSlib source trees, and builds [the program] using an existing HTSlib installation in a system directory (i.e., a directory already being searched by $CPPFLAGS/$LDFLAGS).

### DIFF
--- a/recipes/lofreq/build.sh
+++ b/recipes/lofreq/build.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 set -eu -o pipefail
 
-./configure --with-htslib=${PREFIX}/lib  --prefix=${PREFIX}
-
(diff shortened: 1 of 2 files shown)


bgruening:
@bioconda-bot please merge

Issue #22806, opened 2020-06-22
bioconductor-org.hs.eg.db  download fail


Lix1993:
cannot download 3.11.1 by all 3 url,

3.11.4 can be downloaded by
https://bioconductor.org/packages/3.11/data/annotation/src/contrib/org.Hs.eg.db_3.11.4.tar.gz

tinavisnovska:
Hi there, 

same goes for mouse, with the active download links being 

https://bioconductor.org/packages/3.11/data/annotation/src/contrib/org.Mm.eg.db_3.11.4.tar.gz
https://www.bioconductor.org/packages/release/data/annotation/src/contrib/org.Mm.eg.db_3.11.4.tar.gz

by the way, is there some sort of regular automated check to catch broken links like these?

Lix1993:
thx

PR #22919, opened 2020-06-29
Update of bioconductor-org.Hs.eg.db to v3.11.4 from v3.11.1

tinavisnovska:
Update of bioconductor-org.Hs.eg.db to v3.11.4 to fix broken links of the recipe for the package version 3.11.1 which is currently out there. The update is related to issue #22806 and resolves one of the two mentioned broken packages.

### DIFF
--- a/recipes/bioconductor-org.hs.eg.db/meta.yaml
+++ b/recipes/bioconductor-org.hs.eg.db/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "3.11.1" %}
+{% set version = "3.11.4" %}
 {% set name = "org.Hs.eg.db" %}
 {% set bioc = "3.11" %}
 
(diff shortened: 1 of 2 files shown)


tinavisnovska:
@BiocondaBot please add label

dpryan79:
@bioconda-bot please merge

@bgruening Are we hosting these data packages on the Galaxy cargo-port? Bioconductor's habit of deleting these all the time is really getting annoying.

tinavisnovska:
Hi, 

I was wondering how many recipes have broken url links to their source and it looks like there is more then 500 of them (as of saturday night): over 200 bioconductor recipes, over 250 ucsc utilities and around 100 others (including mothur, breseq, spades). 
Plus there is around 400 recipes which I did not manage to check, because not all meta.yaml files are easy to parse, at least not easy for me ;)

Anyway, happy to share more info if it make sense to you, just  wanted to ask in what form this would be useful. I can imagine listing them in an issue, but that can be rather overwhelming and difficult to manage... Any ideas/preferences? 
 
> @bgruening Are we hosting these data packages on the Galaxy cargo-port? Bioconductor's habit of deleting these all the time is really getting annoying.

PR #22923, opened 2020-06-29
Update sra-tools (add missing curl dependency, rebuild hdf5)

zeehio:
Fixes gh-22929.

This is my first contribution to bioconda.

I just increased the build number of the sra-tools recipe to trigger a rebuild of sra-tools, to use an updated hdf5 1.10.6 dependency

Otherwise:

- `conda create -n myenv sra-tools hdf5=1.10.6` installs `sra-tools 2.8.0` (old version that does not depend on hdf5)
- `conda install sra-tools` on an environment with `hdf5=1.10.6` also installs `sra-tools 2.8.0` to prevent the downgrade.
- `conda install sra-tools=2.10.7` will downgrade hdf5 from 1.10.6 to 1.10.5.

So the main reason for this rebuild is to facilitate the installation of the *latest* version of sra-tools.

Feel free to close this PR, if you believe it does not make sense to trigger a sra-tools rebuild for this reason.

### DIFF
--- a/recipes/sra-tools/meta.yaml
+++ b/recipes/sra-tools/meta.yaml
@@ -19,7 +19,7 @@ source:
     folder: ngs
 
 build:
-  number: 1
+  number: 2
   skip: True  # [osx]
 
 requirements:
@@ -39,6 +39,7 @@ requirements:
     - hdf5
     - perl-xml-libxml
     - perl-uri
+    - curl
 
 test:
   commands:


nsheff:
I noticed there's no `curl` in these containers either:

https://github.com/BioContainers/containers/issues/376#issuecomment-651347098

can you add that as a dependency into the sra-tools recipe as well?

zeehio:
@BiocondaBot please add label

thanhleviet:
@BIocondaBot please merge

zeehio:
Is there anything else I should be doing to help BiocondaBot?

mbargull:
@BiocondaBot, please fetch artifacts

mbargull:
It looks like the previous `please merge` comment was registered as `@BIocondaBot...` (capital `I`) for some reason: https://github.com/bioconda/bioconda-recipes/runs/821946833?check_suite_focus=true#step:3:441

If you inspect https://114186-42372094-gh.circle-artifacts.com/0/tmp/artifacts/packages/linux-64/repodata.json you can see that this will still use `hdf5=1.10.5`.

Apparently `hdf5` is pinned to `1.10.5` in https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/1afc9df489efdf493f23556bd24333ec67ddcb85/recipe/conda_build_config.yaml#LL419-L420 and does not have an accompanying `max_pin`, which means it will restrict the upper version to be `<10.1.6.0a0`.
I don't know if that strict pinning is necessary. You'll have to check with the `hdf5` feedstock and/or whomever added that strict pinning to `conda-forge-pinning`.

zeehio:
Thank you very very much @mbargull for your detailed explanations. I did not know if/how conda-forge does the package pinning, and that bioconda is also using it. It makes a lot of sense.

This PR fixes a missing dependency on sra-tools, closing https://github.com/bioconda/bioconda-recipes/issues/22929, so I guess it would be valuable to merge this anyway. I will squash the commits to remove the hdf5-1.10.6 reference in the commit message.

The hdf5 version pinning, as you say, should be discussed elsewhere.

Besides all that, there is the underlying issue of how conda chooses a version to install, on a fresh miniconda3 docker image in a linux system:

- `conda create -n myenv -c bioconda sra-tools` installs sra-tools 2.8.0.
- `conda create -n myenv -c conda-forge -c bioconda sra-tools` installs sra-tools 2.10.7.

But that should also be discussed elsewhere, I guess.

mbargull:
@bIoCONdaBOT, heLLO

mbargull:
@bIOconDAboT, HeLLo

mbargull:
> This PR fixes a missing dependency on sra-tools, closing #22929, so I guess it would be valuable to merge this anyway. I will squash the commits to remove the hdf5-1.10.6 reference in the commit message.

:thumbsup:

> The hdf5 version pinning, as you say, should be discussed elsewhere.

I had a quick look. Given the record at https://abi-laboratory.pro/?view=timeline&l=hdf5 , it probably makes sense that it is tightly pinned. But yes, when/how the pinning gets updated to `1.10.6` (or `1.12.0` once that one is built on conda-forge?) shouldn't be decided here ;).

> Besides all that, there is the underlying issue of how conda chooses a version to install, on a fresh miniconda3 docker image in a linux system:
> 
>     * `conda create -n myenv -c bioconda sra-tools` installs sra-tools 2.8.0.
> 
>     * `conda create -n myenv -c conda-forge -c bioconda sra-tools` installs sra-tools 2.10.7.
> 
> 
> But that should also be discussed elsewhere, I guess.

Bioconda depends on conda-forge packages. As such we require `-c conda-forge -c bioconda` (see http://bioconda.github.io/user/install.html#set-up-channels). If you leave out `-c conda-forge`, you'll get packages from the `defaults` channels that are not guaranteed to be compatible (they are in most cases, though) and may not be available in the versions we expect them to be.

mbargull:
@bioconda-bot, please fetch artifacts

mbargull:
@bioconda-bot, please merge

mbargull:
> This is my first contribution to bioconda.

Thank you! :tada:

PR #23021, opened 2020-07-06
Add MACSE 1.2

elfyly:
Add MACSE 1.2

MACSE aligns coding NT sequences with respect to their AA translation while allowing NT sequences to contain multiple frameshifts and/or stop codons. MACSE is hence the first automatic solution to align protein-coding gene datasets containing non-functional sequences (pseudogenes) without disrupting the underlying codon structure. It has also proved useful in detecting undocumented frameshifts in public database sequences and in aligning next-generation sequencing reads/contigs against a reference coding sequence.


npavlovikj:
Looks like your recipe failed on the linting, https://app.circleci.com/pipelines/github/bioconda/bioconda-recipes/28760/workflows/c716273f-44d2-4a87-b6c7-b25dd0e6d75f/jobs/114779/steps, with the message:
```
ERROR: recipes/macse/meta.yaml:15: should_be_noarch_generic: The recipe should be build as `noarch`
WARNING: recipes/macse/meta.yaml:29: gpl_requires_license_distributed: The recipe packages GPL software but is missing copy of license.
ERROR: recipes/macse/meta.yaml:41: extra_identifiers_missing_colon: Each item in the extra/identifiers section must be of form `type:value`
```

npavlovikj:
Briefly, you need to make the recipe `noarch: generic`, and add new line at the end of the recipe. Is there a reason why you are modifying `.gitignore` though? That shouldn't be necessary afaik.

elfyly:
Hi @npavlovikj,

Thanks for your help. Much appreciated!

What about the warning: gpl_requires_license_distributed?
As far as I know, the information I could found was:
MACSE v2 is Java software freely available under the CECILL license (GPL variant) at https://bioweb.supagro.inra.fr/macse/, last accessed August 22, 2018.

Also, what do they mean by: extra_identifiers_missing_colon. Each item in the extra/identifiers section must be of form `type:value`?

Thanks for pointing .gitignore out. I wasn't aware of that.
Cheers,
Elfy

recipes/macse/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    MACSE aligns coding NT sequences with respect to their AA translation while allowing NT sequences to contain
+    multiple frameshifts and/or stop codons. MACSE is hence the first automatic solution to align protein-coding
+    gene datasets containing non-functional sequences (pseudogenes) without disrupting the underlying codon
+    structure. It has also proved useful in detecting undocumented frameshifts in public database sequences and
+    in aligning next-generation sequencing reads/contigs against a reference coding sequence
+
+extra:
+  identifiers:
+    - biotools:macse
+    - doi.org/10.1093/molbev/msy159
```
npavlovikj, line 42: This should be:
```
- doi:10.1093/molbev/msy159
```

recipes/macse/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - python
+
+test:
+  commands:
+    - macse -h
+    - macse --version
+
+about:
+  home: 'https://bioweb.supagro.inra.fr/macse/'
+  license: GPL
```
npavlovikj, line 29: GPL License requires the license file to be part of the distribution, thus the warning message you got. omicX shows the license for MACSE as `CeCILL version 2.1`, so I would suggest you to remove GPL and use `license: CeCILL 2.1` instead.

elfyly:
@npavlovikj 
I'm new to this. Glad you helped me out. Thanks a lot! 
I think I committed the changes, is that correct? However, I don't see bioconda-test testing. I did some changes in the files, but whenever I push it it says it's pushes successfully and whenever I make a new pull request it says there is already a pull request active.

npavlovikj:
@elfyly , if you check the log files, https://app.circleci.com/pipelines/github/bioconda/bioconda-recipes/28820/workflows/ef28a947-0f59-4482-a2a4-ebdd7c86fbfd/jobs/114931/steps, you will see that the recipe actually failed on the testing this time. In order for the test to pass, the exit code of the testing command should be 0. I wonder if running `macse -h` doesn't return 0 code? Can you please check this? If that is really the case, then you can may redirect the output so it returns 0 actually? (something like `macse -h > /dev/null`).

elfyly:
@npavlovikj  If I run this command: `$ java -jar macse_v1.2.jar -h` I get the 'help' of macse:

> This is the release 1.02 of MACSE
> ===================================
> 
> MACSE includes a series of tools dedicated to multiple sequence alignment of (pseudo-)coding DNA sequences.
> ...

is that what you mean?

dpryan79:
@bioconda-bot please merge

PR #23042, opened 2020-07-07
Update cellprofiler recipe with additional dependency backports.functools-lru-cache

sunyi000:
if build cellprofiler locally using the recipe with 'conda build' command, an error occurs complaining about missing dependency backports.functools-lru-cache.

this seems only happen when trying to locally build and install.

I'm not 100% certain why this is happening now but not before.


sunyi000:
@bgruening please have a look and review .

Thanks very much

bgruening:
@sunyi000 this recipe is skipping the build. Do you remember that it crashed because of OOM? So we needed to build it manually.

sunyi000:
> @sunyi000 this recipe is skipping the build. Do you remember that it crashed because of OOM? So we needed to build it manually.

yes I remember that, the build uses lots of RAM so it's disabled during merge.

on the other hand, I checked all my git and command history, even the oldest not-so-good-quality but working version does not have backports.functools-lru-cache specified,but it works before.

I also exported my old working env (before it gets into bioconda), the environment.yml does not have it.

mencian:
Closing in favor of #50744.

PR #23460, opened 2020-07-29
initial attempt to add bitsets

smirarab:
Bitsets is used by many bio packages. It is not available on conda-forge. Would be good to have it here.


smirarab:
@BiocondaBot please add label

PR #23689, opened 2020-07-31
Update jalview: enable maxMemory setting

martin-raden:

----

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

<table>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/recipes/jalview/jalview.sh
+++ b/recipes/jalview/jalview.sh
@@ -32,7 +32,7 @@ if [[ "$JALVIEW_JRE" != "j11" && "$JALVIEW_JRE" != "j1.8" ]]; then
 fi
 
 # check if memory maximum is set and if so forward to java-based jalview call
-if [ -z $JALVIEW_MAXMEM ]; then
+if [ -z "$JALVIEW_MAXMEM" ]; then
   VMMAXMEM=""
 else
   VMMAXMEM="-Xmx${JALVIEW_MAXMEM}"
--- a/recipes/jalview/jalview.sh
+++ b/recipes/jalview/jalview.sh
@@ -9,6 +9,9 @@
 # By default, this wrapper executes java -version to determine the JRE version
 # Set JALVIEW_JRE=j1.8 or JALVIEW_JRE=j11 to skip the version check 
 #
+# By default, this wrapper does NOT restrict the memory consumption of Jalview.
+# Set eg. JALVIEW_MAXMEM=1g to set the maximal memory of Jalview's VM
+#
 ###############################
 
 # Find original directory of bash script, resolving symlinks
@@ -28,4 +31,11 @@ if [[ "$JALVIEW_JRE" != "j11" && "$JALVIEW_JRE" != "j1.8" ]]; then
   if [[ $( java -version 2>&1 | grep '"1.8' ) != "" ]]; then JALVIEW_JRE=j1.8; fi
 fi
 
-java -jar $DIR/jalview-all-${JALVIEW_JRE}.jar ${@};
+# check if memory maximum is set and if so forward to java-based jalview call
+if [ -z ${JALVIEW_MAXMEM+x} ]; then
+  VMMAXMEM=""
+else
+  VMMAXMEM="-Xmx${JALVIEW_MAXMEM}"
+fi
+
+java $VMMAXMEM -jar $DIR/jalview-all-${JALVIEW_JRE}.jar ${@};
--- a/recipes/jalview/jalview.sh
+++ b/recipes/jalview/jalview.sh
@@ -32,7 +32,7 @@ if [[ "$JALVIEW_JRE" != "j11" && "$JALVIEW_JRE" != "j1.8" ]]; then
 fi
 
 # check if memory maximum is set and if so forward to java-based jalview call
-if [ -z ${JALVIEW_MAXMEM+x} ]; then
+if [ -z $JALVIEW_MAXMEM ]; then
   VMMAXMEM=""
 else
   VMMAXMEM="-Xmx${JALVIEW_MAXMEM}"
--- a/recipes/jalview/meta.yaml
+++ b/recipes/jalview/meta.yaml
@@ -3,7 +3,7 @@ package:
   version: 2.11.1.0
 
 build:
-  number: 0 
+  number: 1
   noarch: generic
 
 source:
--- a/recipes/jalview/meta.yaml
+++ b/recipes/jalview/meta.yaml
@@ -43,3 +43,4 @@ extra:
   notes: |
     This wrapper and installation is primarily for commandline-only use.
     Set JALVIEW_JRE=j1.8 or JALVIEW_JRE=j11 to specify the java runtime if you need jalview to start up as quickly as possible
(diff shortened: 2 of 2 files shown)


martin-raden:
the problem with the current jalview wrapper is its incapability to restrict jalviews memory consumption. the [standard way](https://www.jalview.org/help/html/memory.html) (using a .jvl file) is disabled since the wrapper directly. The alternative is to restrict the callings Java VM memory, which is now implemented via the (optional) `JALVIEW_MAXMEM` bash variable.

martin-raden:
ping @bgruening @foreveremain any suggestions? (please review my changes)

martin-raden:
@BiocondaBot please add label

foreveremain:
https://www.jalview.org/help/html/memory.html describes this behaviour. The jalview jar uses the same relaunch trick .. @benzoid  : can you provide a definitive fix here ?

recipes/jalview/jalview.sh
```diff
@@ -28,4 +31,11 @@ if [[ "$JALVIEW_JRE" != "j11" && "$JALVIEW_JRE" != "j1.8" ]]; then
   if [[ $( java -version 2>&1 | grep '"1.8' ) != "" ]]; then JALVIEW_JRE=j1.8; fi
 fi
 
-java -jar $DIR/jalview-all-${JALVIEW_JRE}.jar ${@};
+# check if memory maximum is set and if so forward to java-based jalview call
+if [ -z ${JALVIEW_MAXMEM+x} ]; then
```
epruesse, line 35: I'd use `-z "${JALVIEW_MAXMEM}"` (no `+x`), unless you really want `JALVIEW_MAXMEM="" jalview.sh` to call `java -Xmx -jar ...`?

epruesse:
> When launched as an Application, Jalview will automatically configure the amount of memory allocated to the program to be 90% of physical memory. You may wish to change this behaviour - particularly if you are working on a machine that runs other memory intensive processes. 

This seems doubtful. I'd assume it will *allow* the JVM to allocate 90% as java heap, as with `-Xmx`. I'm not even sure you can make the JVM lock memory. Not sure whether the JVM GC gets wasteful if this is too high. Is there an observable problem in RES mem size?

martin-raden:
We indeed observe the behavior on a memory limited queuing system. The job has only 24 of 32 g ram but jalview tries to runs the vm with xmx30g (terminal output), which causes jalview to fail since it cannot reserve the heapsize. Will dig out some logdata if needed.

recipes/jalview/jalview.sh
martin-raden: True enough, check for emptiness should do, was checking whether the var is set at all. Will change.

martin-raden:
here the log output for the problem (job with 24g mem restriction on a machine with 32g physical memory; jalview tries to use 90% of the physical memory)
```
Running jalview.bin.Jalview with -Xmx30422276096
Error occurred during initialization of VM
Could not reserve enough space for 29712384KB object heap
```
using a `.jvl`, as suggested in the jalview manual linked above, does not work, since the file is ignored when calling jalview via the jar, as done in the current wrapper. Thus, the only option I see (also suggested in the manual) is to explicitely restrict the maximal memory of the calling Java VM.

epruesse:
Odd. Is this a 32bit JVM? I'm not much of a Java expert, but I thought all it needed was address space, not physical or free memory.

martin-raden:
> Anyway, approving. Good to have option to alter this setting. Want to put it into the package description to document?

Sure.. will alter it.

epruesse:
@BiocondaBot please merge

martin-raden:
@epruesse thanks!

benzoid:
Hi, sorry for coming in late here.
Running the jalview-all-2.11.1.0-j1.8.jar (or later version, or the j11 jar) with `java -jar jalview-all-....jar` first runs a minimal wrapper found in the jar, `jalview.bin.Launcher`, which in turn  uses com.sun.management.OperatingSystemMXBean.getTotalPhysicalMemorySize() to try and determine total physical memory, and by default will set the -Xmx setting for jalview.bin.Jalview to be the smaller of 90% of physical memory and 32GB.
These two values (90 and 32g) can be adjusted in various ways, but in this case the simplest will be using application command line args -jvmmempc and -jvmmemmax, e.g.

java -jar jalview-all-2.11.1.0-j1.8.jar -jvmmempc=50 -jvmmemmax=24g

Apologies that this has not been updated in the documentation on the website yet, my bad!
Since application arguments already appear to be passed through, this can already be done by adding a `-jvmmemmax=24g`
Hope this helps.

-

epruesse:
@benzoid So what's the deal with the error? Why does Java crash if you set the max heap size to say 27GB on a 32GB machine with 16GB free? AFAIK the `-Xmx` max heap size is just an allocation of virtual memory?

martin-raden:
Hi @benzoid @epruesse 
I will check whether the problem is solved with the (hidden) commandline arguments. In case, I will undo the proposed changes to the conda wrapper.
Best,
Martin

benzoid:
I've updated our documentation to include those command line arguments in our develop branch and hopefully can squeeze it into our next patch release documentation.
@epruesse I'm not entirely sure why this causes Jalview to die straight away, although as you suggest a 32 bit JVM might fail (I believe a 32 bit JVM needs the memory address space allocation to be contiguous so overall free space might not be enough).  I would be quite surprised if a 32 bit JVM was in use though.
@martin-raden when you say "job with 24g mem restriction on a machine with 32g physical memory" is that 24g restriction due to a resource manager or job queuing system like on e.g. a HPC cluster?  In this case it might just be that the JVM detection of total available memory is inappropriate (e.g. it detects the physical memory, but doesn't realise it's being managed/limited, so thinks it can allocate more than the resource manager is limiting it to... and it can't). I would hope that -jvmmemmax=20g (or set to a little under whatever the job memory limit is) should make things work okay.  The heap stack is not the only JVM memory usage, so there should still be a small amount of memory allowed for the rest of the JVM.
Again, apologies for the inconvenience not having updated the documentation caused!

martin-raden:
@benzoid yes, hpc setup with managed resources. Thus, the detection fails as you said. Fortunately, the 'new' option solves the issue. Thus i put a new pr to undo my proposed changes..

benzoid:
@martin-raden, @epruesse
After discussion with @foreveremain we think the original PR should remain and be implemented.  It doesn't break the -jvmmempc -jvmmemmax arguments, and provides a recognised java way to set an absolute value for the maximum heap size.
jvmmemmax can be overridden by a smaller jvmmempc and some use cases might want that to just fail, rather than run with a smaller than expected maximum heap size.
Documentation for -jvmmem* should be added for the next patch release :)

martin-raden:
Fine for me, so i will close the 'undo' PR

benzoid:
@martin-Thanks, and thanks for raising this issue.  I'd tested the jvmmemmax default 32g limit worked on HPC but hadn't tested in a resource-limited way.  We'll also add an FAQ for such a use case.

epruesse:
Agreed. No need to create another build of the package. It doesn't hurt to have the env variable. The other java wrappers we have typically do that too anyway.

PR #23760, opened 2020-08-04
MOB-Suite: minor fix due to pandas version update causing EmptyDataFrame Exception error

kbessonov1984:
Added pandas restriction to an existing recipe to avoid pandas `EmptyDataFrame` exception. All `pandas` versions prior to `1.1.1` do work. Please merge with small fix to the existing recipe. In the next tool release will address this issue in the source code making it pandas version agnostic.

This is an urgent update as previous recipe does not put any restrictions on pandas version upper range

How I can add labels in this PR? The option is not available for me if I click on it.


kbessonov1984:
ping @BiocondaBot

dpryan79:
@bioconda-bot please merge

PR #24268, opened 2020-09-07
TIDDIT: fix TIDDIT executable.

cbrueffer:
Executing TIDDIT resulted in the following error:

TIDDIT: error while loading shared libraries: libbamtools.so.2.3.0: cannot open shared object file: No such file or directory

Caused by copying the bamutils source directory instead of the built
library files.  Added a test that executes the TIDDIT binary itself,
to catch issues such as these.


bgruening (approved):

bgruening:
@bioconda-bot please merge

PR #24295, opened 2020-09-08
Update rmblast to version 2.10.0

nathanweeks:
The latest version of RMBlast is 2.10.0: http://www.repeatmasker.org/RMBlast.html

### DIFF
--- a/recipes/rmblast/meta.yaml
+++ b/recipes/rmblast/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "rmblast" %}
-{% set version = "2.9.0" %}
+{% set version = "2.10.0" %}
 
 package:
   name: {{ name }}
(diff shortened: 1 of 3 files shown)


recipes/rmblast/meta.yaml
```diff
@@ -1,16 +1,16 @@
...
 
 package:
   name: {{ name }}
   version: {{ version }}
 
 source:
   url: https://ftp.ncbi.nlm.nih.gov/blast/executables/blast+/{{ version }}/ncbi-blast-{{ version }}+-src.tar.gz
-  sha256: a390cc2d7a09422759fc178db84de9def822cbe485916bbb2ec0d215dacdc257
+  sha256: 3acdd9cec01c4f43e56aeaf89049cb8f8013d60b9c1705eced10166967f1d926
   #patches:
```
bgruening, line 11: patches are not used anymore, isn't it? So we can remove them?

nathanweeks: I thought the "Needs the -p1 option, applied in build.sh" comment was helpful in providing the rationale for why the patch is applied in build.sh, but it's fine to remove the (commented-out) patches field entirely from meta.yaml if that's preferred.

Juke34:
@bgruening I tried without the patch and it does not work. see here: https://github.com/bioconda/bioconda-recipes/commit/e5b8fa936d5de6a2e1b9a0db614c31d006668e34

So we can let as it is here and merge

nathanweeks:
@BiocondaBot please add label

dpryan79 (approved):

dpryan79:
@bioconda-bot please merge

dpryan79:

dpryan79:
Oh, the last build was too long ago for there to still be artifacts. Not a problem

PR #24310, opened 2020-09-09
Update snoscan to version 1.0

nathanweeks:
snoscan 1.0 is now available. Several additional executables are included (snowscanA, snoscanH, and snoscanY == snoscan).

I've patched the sort-snos Perl script to use Getopt::Std (as was used in snoscan 0.9.1) instead of the obsolete (and removed in Perl 5.16) getopts.pl, and notified the author.


nathanweeks:
snoscan author Prof Todd Lowe fixed the sort-snos script (replacing the existing snoscan-1.0.tar.gz file), so the patch is no longer necessary.

nathanweeks:
lint timed out---could someone with appropriate permissions rerun?

nathanweeks:
I updated the command that moves the *.pl files into perl-build to guard against a (perhaps unlikely) race condition where `find` descends into perl-build and finds .pl files that have already been moved (which would result in an error like `mv: 'perl-build/foo.pl' and 'perl-build/foo.pl' are the same file`). But this was mainly just to do something to tickle CircleCI so it would rerun the tests.

cbrueffer (approved):

cbrueffer:
@bioconda-bot please merge

PR #24563, opened 2020-09-27
Update nim-falcon

pb-cdunn:
(was pbipa, but we re-used `master` oops.)

This is an attempt to fix several bugs at once, so we can stop touching this reciple forever after.

* https://github.com/bio-nim/nim-falcon/releases/tag/3.0.2

### DIFF
--- a/recipes/nim-falcon/meta.yaml
+++ b/recipes/nim-falcon/meta.yaml
@@ -12,7 +12,7 @@ extra:
     - should_be_noarch_generic # We use "skip", so this is a flawed lint-check.
 
 source:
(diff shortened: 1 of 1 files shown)


cdunn2001 (approved):

pb-cdunn:
```
conda create -n pba pb-assembly nim-falcon==3.0.2
```
If you do not specify `nim-falcon==3.0.2`, you may end up with the wrong nim-falcon/htslib versions because of intricacies of how conda works. We're trying not to update **pb-assembly** and **nim-falcon** at all anymore, now that there is a working combination.

PR #24630, opened 2020-10-01
Update - Bugfix/missing snprelate dep

sstadick:
On osX SNPRelate fails to load unless `libopenblas` is installed. I have not checked the same bug on linux to see if `libblas` alone works there. 

@dpryan79 I think you were the last one to modify this bit of the recipe. Do you have any insight here?


dpryan79:
That's quite odd, libblas should have resolved as pulling in libopenblas. I'll play around with this PR and have a look.

dpryan79:
@sstadick It doesn't seem that blas is actually linked to. I added that in for all compiled packages, since they don't list if they actually need that or not.

sstadick:
Interesting, but why is there a `BLAS_LIB` var in the makefile? https://github.com/zhengxwen/SNPRelate/blob/87a9f253b73cb4f86f5a480d6a225a50b9c3326d/src/Makevars#L9. I'll do some digging and see if I can reproduce it again as well.

dpryan79:
Looking through the source code a bit it should be linking against whatever blas R is linking to, since it's linking to the lapack that R is compiled against.

PR #25187, opened 2020-11-05
Update HTSlib recipe: patches to fix https/S3/GCS/etc plugins

jmarshall:
Changes between HTSlib 1.10 and 1.11 have led to samtools/htslib#1176, preventing programs using bioconda's htslib from accessing files remotely over HTTPS, S3, etc. This problem will be fixed in upstream HTSlib in due course; this PR applies patches to fix the bioconda package in the meantime:

* Apply samtools/htslib#1150 to fix some externally-supplied plugins such as iRODS access;

* Revert part of samtools/htslib#1072 to fix the main HTTPS/S3/etc issue. This simply reverts the macOS part of this as it was unnecessary on macOS anyway (Linux was unaffected by the bug).

Patches applicable to HTSlib 1.11; will not be needed for future upstream versions.

### DIFF
--- a/recipes/htslib/meta.yaml
+++ b/recipes/htslib/meta.yaml
@@ -5,13 +5,15 @@ package:
   version: {{ version }}
 
 build:
-  number: 0
+  number: 1
   run_exports:
     - {{ pin_subpackage('htslib', max_pin='x.x') }}
 
 source:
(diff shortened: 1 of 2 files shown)


jmarshall:

jmarshall:
Confirmed that the osx-64 package resolves the HTTPS, S3, etc issue.

jmarshall:
@BiocondaBot please add label

dpryan79 (approved):

PR #25292, opened 2020-11-11
Fix bioconductor-cytolib on macOS and rebuild dependent packages

jmarshall:
At least with Bioconductor 3.12, R library shared libraries on macOS do have the usual _.dylib_ extension. Patch bioconductor-cytolib to adjust `cytolibLibPath()` accordingly.

~Then enable building and bump the build numbers for all the bioconductor-* packages that depend on this library.~

PR #25299, opened 2020-11-11
Rebuild packages that depend on bioconductor-cytolib

jmarshall:
Followup to PR #25292. This rebuilds recipes/bioconductor-cydar, -cyto*, -flow*, -immunoclust, -ncdfflow, and -opencyto — it appears (from the previous build logs) that this needs to be done in two stages to ensure that the dependent packages are built using the freshly-built corrected bioconductor-cytolib from the previous PR.

For CytoML and CytoTree, it seems that while these have been listed in _build-fail-blacklist_ the upstream versions present in Bioconductor have changed. So this PR updates them to the upstream version actually present in Bioconductor 3.12 rather than bumping the build numbers as for the other dozen packages. (Just bumping the build number for these two wasn't enough: trying to download the non-bioc-3.12 source tarballs just resulted in 404.)

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1319,22 +1319,8 @@ recipes/r-stitch
 recipes/r-zerone
 
 # Fails on OSX
-recipes/bioconductor-cydar
-recipes/bioconductor-cytofast
-recipes/bioconductor-cytoml
-recipes/bioconductor-cytotree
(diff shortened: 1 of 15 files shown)


dpryan79 (approved):

dpryan79:
@bioconda-bot please merge

Thanks again @jmarshall !

PR #25327, opened 2020-11-12
Fix bioconductor-rbowtie build on macOS

jmarshall:
This was failing due to no `-headerpad` option being used when building the executables down in src/SpliceMap. Patch the Makefile to use R's LDFLAGS (alongside the existing R's CXXFLAGS) when building these executables, as (at least with conda's R) that provides `‑Wl,‑headerpad_max_install_names`.

@mbstadler: I'm happy to make a PR against **fmicompbio/Rbowtie** once this has demonstrated that it builds successfully on both Linux and macOS. Or if you want to apply _ldflags.patch_ to the repository yourself in due course, that's probably even easier.

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1309,7 +1309,6 @@ recipes/bioconductor-gmapr
 recipes/bioconductor-macpet
 recipes/bioconductor-qckitfastq
 recipes/bioconductor-quasr
-recipes/bioconductor-rbowtie
 recipes/r-misha
 
 # Missing tarball
--- a/recipes/bioconductor-rbowtie/build.sh
+++ b/recipes/bioconductor-rbowtie/build.sh
@@ -4,7 +4,7 @@ grep -v '^Priority: ' DESCRIPTION.old > DESCRIPTION
 mkdir -p ~/.R
 echo -e "CC=$CC
 FC=$FC
-CXX=$CXX -headerpad
+CXX=$CXX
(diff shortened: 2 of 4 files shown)


jmarshall:
@BiocondaBot please add label

mbargull (approved):

dpryan79:
@bioconda-bot please merge

mbstadler:
@jmarshall Thanks a lot - great that you could find a solution. I hope that means that `bioconductor-quasr`, which depends on `bioconductor-rbowtie`, will build now, too.

I will take care of introducing your fix into `fmicompbio/Rbowtie`, no need for a PR.

This will however not affect the current `Rbowtie` (1.30.0), as I would apply that fix to the current devel branch of `Rbowtie` (1.31.x), which will become the new release version (1.32.x) when the next Bioconductor (3.13) will be released in ~April 2021.
Let me know if you would prefer the release version to be patched, as well (creating an `Rbowtie` 1.30.1).

jmarshall:
Yes, bioconductor-quasr builds fine now too — there were no additional problems there.

As far as bioconda is concerned, we've patched this locally now so don't need a 1.30.1 in bioc 3.12. It's up to you whether this would be a benefit to other users of 1.30.0 — IMHO it would be pretty minimal as there are unlikely to be too many other critical LDFLAGS that people have been missing, so probably not worth doing.

PR #25365, opened 2020-11-14
Fix bioconductor-gmapr build on macOS

jmarshall:
Build the shared library as `gmapR$(SHLIB_EXT)` rather than hard-coding the filename `gmapR.so`.

Ensure that `src/samtools/*.[oa]` gets rebuilt — in gmapR_1.32.0.tar.gz at least, the source tarball incorrectly already contains built-for-Linux `*.o` files, which breaks building on other platforms.

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1301,7 +1301,6 @@ recipes/r-stitch
 recipes/r-zerone
 
 # Fails on OSX
-recipes/bioconductor-gmapr
 recipes/bioconductor-qckitfastq
 recipes/r-misha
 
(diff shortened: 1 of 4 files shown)


dpryan79 (approved):

dpryan79:
Nice find regarding the build artifacts in the release tarball!

jmarshall:
Bit of a head-scratcher when I couldn't reproduce it from a lawremi/gmapR git checkout :smile:

But quite quick once I figured out how to get `bioconda-utils build --keep-image --keep-old-work` + `R CMD INSTALL --no-clean-on-error` (possibly only a subset of that was actually needed) to preserve the intermediate files…

dpryan79:
@bioconda-bot please merge

PR #25372, opened 2020-11-15
Fix bioconductor-qckitfastq build on macOS

jmarshall:
Rather than futz around with `MACOSX_DEPLOYMENT_TARGET`, simply patch to avoid the problematic header — which is not needed in any way by the code anyway.

Avoid #including an unneeded seqan header that uses a std datatype that's not available in macOS's libc++ prior to macOS 10.12. This patch can be removed when `MACOSX_DEPLOYMENT_TARGET` is raised past 10.12.

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1301,7 +1301,6 @@ recipes/r-stitch
 recipes/r-zerone
 
 # Fails on OSX
-recipes/bioconductor-qckitfastq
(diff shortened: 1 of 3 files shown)


dpryan79 (approved):

dpryan79:
@bioconda-bot please merge

Great job @jmarshall!

PR #25379, opened 2020-11-16
Fix r-misha build on macOS

jmarshall:
A bonus post-hackathon PR to eliminate this “Fails on OSX” section of _build-fail-blacklist_…

This package contains code that uses a Linux-specific `sendfile()` system call. Fortunately the code in question (`FileUtils::copy_file()`/`move_file()`) is unused anyway, so we just patch to remove the problematic code.

Implementing an additional alternative instead so the code compiles everywhere has been proposed upstream in PR tanaylab/misha#2.

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -1297,9 +1297,6 @@ recipes/r-qorts
 recipes/r-stitch
 recipes/r-zerone
 
-# Fails on OSX
-recipes/r-misha
-
 # Missing tarball
 recipes/r-prestor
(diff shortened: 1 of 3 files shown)


jmarshall:
@BiocondaBot please add label

dpryan79 (approved):

dpryan79:

dpryan79:
@bioconda-bot please merge

dpryan79:
Thanks again @jmarshall!

PR #25396, opened 2020-11-17
Update r-misha to 4.0.11

jmarshall:
The 4.0.11 update removes src/FileUtils.cpp entirely (see tanaylab/misha#2 conversation), so the patch should now be removed.


jmarshall:
@BiocondaBot please add label

pvanheus:
Looks good!

pvanheus (approved):

PR #25742, opened 2020-12-09
update strling: htslib req

brentp:
expands the htslib version in strling

### DIFF
--- a/recipes/strling/meta.yaml
+++ b/recipes/strling/meta.yaml
@@ -19,9 +19,9 @@ requirements:
     - {{ compiler('c') }}
     - nim
   host:
-    - htslib >=1.10
+    - htslib
   run:
-    - htslib >=1.10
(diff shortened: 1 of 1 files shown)


dpryan79:

dpryan79 (approved):

dpryan79:
htslib is a "pinned dependency", meaning packages end up getting built against specific versions of it. At the moment we're building everything against htslib 1.10. A quirk of conda's meta.yaml is that if you specify >=1.10 for htslib, you end up pinning it to 1.11 at the moment, which will then conflict with other packages. I've adjusted the current PR to make it more broadly compatible. When we update the htslib pinning next then there will be a new build of strling for htslib 1.11.

dpryan79:
@bioconda-bot please merge

brentp:
Thanks @dpryan79

PR #25862, opened 2020-12-18
Add CHAP recipe

jamesabbott:
Adds recipe for CHAP, which carries out functional annotation of ion channel structures

There are some unfortunate version pinnings here: Gromacs versions >2018 are not compatible with the software, while using libcblas/liblapacke versions > 3.8.0 fail to compile. The netlib versions of these libraries are also required instead of the openblas versions which don't include the necessary headers for the compilation to succeed.

The software is also described as being only tested on ubuntu, and my attempts to build on macos have failed.

### DIFF
--- a/recipes/chap/LICENSE.TXT
+++ b/recipes/chap/LICENSE.TXT
@@ -1,0 +1,10 @@
+CHAP - The Channel Annotation Package
+
+Copyright (c) 2016 - 2018 Gianni Klesse, Shanlin Rao, Mark S. P. Sansom, and Stephen J. Tucker
+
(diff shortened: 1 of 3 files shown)


jamesabbott:
@BiocondaBot please add label

dpryan79 (approved):

dpryan79:
@bioconda-bot please merge

Wow, looks like this was a bit of a beast to get working. Nicely done @jamesabbott. I guess we'll hope that the authors update it to handle newer gromacs versions at some point, since this might be a bit painful to install in the same environment as other packages otherwise.

dpryan79:
Alternatively, we can consider rebuilding a compatible gromacs version if you end up being too constrained by the current dependency version restrictions.

jamesabbott:
It was a bit evil to get to build…there are build warnings about compiler compatibility but my tester has reported it runs ok. Otherwise I was thinking it would need a gromacs rebuild with the current compilers.  It doesn’t look like the author has done much with it in the last couple of years, so I’m not overly optimistic about it being brought up to date.  As long as no-one complains about it being broken we are probably safe leaving gromacs as it is.

PR #26014, opened 2021-01-03
Update anvio-minimal to v7

meren:
Updates the anvio-minimal recipe to match the latest release of [anvi'o](http://merenlab.org/software/anvio), a software platform directly relevant to the biological sciences.


meren:

meren:
The artifact installs nicely in my local, and the command that freaks out the build system works beautifully :(

meren:
@BiocondaBot please add label

dpryan79 (approved):

dpryan79:
This is the second package I've seen where importing openmp in python causes this error (I think the other was staden). I think there's something wrong with a conda-forge dependency.

meren:
Thank you @dpryan79 and @druvus for looking into this. Is it still OK to merge this since the package is working as expected outside of the build environment?

Best wishes,

druvus:
Seems downgrading scikit-learn solves the issue but I am not sure it is a long-term solution.

meren:
> Seems downgrading scikit-learn solves the issue but I am not sure it is a long-term solution.

This is interesting and I have no idea why it solves the problem. I'm curious how did you even come up with this idea. It would have never occurred to me.

druvus:
Started to read about the issue that seems to be quite common.  When I found an old closed issue at your github repo with the same error message (https://github.com/merenlab/anvio/issues/1489) I decided to give it a try.

druvus (approved):

meren:
Ah, I see! I hadn't realized that using the extended image  solved the ASCII encoder issue but then there was another error!

Thank you very much for going after that one too, @druvus. You are the best.

druvus:

dpryan79:
@bioconda-bot please merge

PR #26060, opened 2021-01-05
Fix MetaBat2 libdeflate.so/libdeflate.dylib issue

epruesse:
Hoping that rebuild will suffice


mencian:
Closing in favor of #56447 (and also since its been so long)

PR #26741, opened 2021-02-12
Update raven-assembler meta.yaml

rvaser:
I have removed git submodules from the Raven repository, so the auto-generated release archives are sufficient for builds.


rvaser:
@BiocondaBot please add label

pvanheus (approved):

PR #26932, opened 2021-02-21
Add SimBac

simojoe:
Adding the SimBac software.

The original software has not been versioned. I have kept the default recipe version value, and used the latest (4 years ago) commit archive. The software does not seem to be in development anymore.


simojoe:
@BiocondaBot please add label

recipes/simbac/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  run:
+    - gsl
+
+test:
+  commands:
+    - SimBac -h | grep "SimBac \[OPTIONS\]"
+
+about:
+  home: https://github.com/tbrown91/SimBac
+  license: GPL3
```
bgruening, line 30: can you use an SPDX identifier here: https://spdx.org/licenses/GPL-3.0.html

recipes/simbac/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  url: https://github.com/tbrown91/SimBac/archive/50158978df536e10e4fcb2f9cd3871deb6da99ca.zip
+  sha256: 82d42a4aef46c22c3d6472c1f3bf4915f6461bc7ec6eabca2bf1a3e9e3c2a914
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
```
bgruening, line 18: A cpp compile is enough isn't it?

simojoe: I tested it and it is. I have applied the modification.

simojoe:
Thank you for the review. I have modified accordingly.

recipes/simbac/meta.yaml
```diff
@@ -0,0 +1,34 @@
+{% set name = "simbac" %}
+{% set version = "0.1a" %}
```
dpryan79, line 2: Is this an alpha release? Those aren't meant for end-users.

simojoe: The software seems to be without an active maintainer, and has never been officially versioned. What would be a suitable value for the version field? I think we can consider it to be out of alpha since it was published.

dpryan79: It's fine as is then. I mostly wanted to check because sometimes people submit their own tools without giving them a proper release number first. But since this isn't your package there's nothing that can be done about it.

dpryan79 (approved):

dpryan79:
@bioconda-bot please merge

PR #27112, opened 2021-03-02
Update deepnog: require PyTorch

VarIr:
`deepnog` requires PyTorch, which was previously not available from conda standard channels. Thus, the current deepnog recipe is omitting it, requiring the user to manually install the pytorch dependency from the pytorch channel.

Now it seems, conda-forge has pytorch, and we, therefore, include it in the requirements section. Also, we can now actually do tests in this recipe. Finally, I think we had some superfluous host requirements.

### DIFF
--- a/recipes/deepnog/meta.yaml
+++ b/recipes/deepnog/meta.yaml
@@ -12,7 +12,7 @@ source:
 build:
   # We should be platform-independent (no compiled code etc.)
   noarch: python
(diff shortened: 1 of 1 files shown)


VarIr:

dpryan79 (approved):

dpryan79:
@bioconda-bot please merge

PR #27140, opened 2021-03-03
adds longestrunsubsequence

maxried:
This is an implementation of the longest run subsequence problem, introduced in (Schrinner et al., WABI 2020) https://drops.dagstuhl.de/opus/volltexte/2020/12795/. It describes a string problem, which looks for the longest subsequence of a string such that this sequence contains at most one consecutive run for each character in the underlying alphabet. The code contains two different algorithms (based on Integer Linear Programming and on Dynamic Programming) as described in the publication. The problem appears in the context of homology-based scaffolding of two contig sets, originating from two individuals of the same species.

### DIFF
--- a/recipes/longestrunsubsequence/meta.yaml
+++ b/recipes/longestrunsubsequence/meta.yaml
@@ -1,0 +1,40 @@
+{% set version = "1.0.1" %}
+
+package:
+  name: "longestrunsubsequence"
+  version: "{{ version }}"
+
+source:
+  url: https://github.com/AlBi-HHU/longest-run-subsequence/archive/v{{ version }}.tar.gz
+  sha256: 75440b9097ee018762febffd2495be3bd24ccbf3642643e1ed7e9b1cb157d76c
+
+build:
+  number: 0
+  noarch: python
+  script: "{{ PYTHON }} -m pip install --no-deps -vv ."
+
+requirements:
+  host:
+    - pip
+    - python >=3.5
+  run:
+    - python >=3.5
+  run_constrained:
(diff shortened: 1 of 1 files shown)


dpryan79:

recipes/longestrunsubsequence/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+  noarch: python
+  script: "{{ PYTHON }} -m pip install --no-deps -vv ."
+
+requirements:
+  host:
+    - pip
+    - python >=3.5
+  run:
+    - python >=3.5
+  run_constrained:
```
dpryan79, line 22: Please remove the `run_constrained:` entry, it isn't needed. Having `pulp >=2` is sufficient.

dpryan79 (changes requested):
See comment, just one small change and this is good to go.

recipes/longestrunsubsequence/meta.yaml
maxried: Thanks! My understanding of run_constrained is that it's meant to specify optional dependencies. longestrunsubsequence uses pulp if it detects its presence, but has no hard dependency on it. What's the proper way to specify this?

dpryan79: OK, that should presumably work then. We usually just add optional requirements that we expect users will want as normal run-time dependencies.

dpryan79 (approved):

dpryan79:
@bioconda-bot please merge

PR #27898, opened 2021-04-08
Update libsbml to contain extensions

jotech:
- added popular extensions to libsbml compilation to enable proper
  usage in constraint-based modelling
- these extensions are needed for example to use sbml in R (sybiSBML)
- linux distributions often contain these extensions by default, for
  example debian: https://salsa.debian.org/med-team/libsbml

### DIFF
--- a/recipes/libsbml/build.sh
+++ b/recipes/libsbml/build.sh
@@ -8,6 +8,14 @@ cmake -DCMAKE_INSTALL_PREFIX=${PREFIX} \
       -DWITH_SWIG=OFF \
       -DLIBSBML_DEPENDENCY_DIR="${PREFIX}" \
       -DLIBXML_INCLUDE_DIR=${PREFIX}/include/libxml2 \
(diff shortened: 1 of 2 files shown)


jotech:
@BiocondaBot please add label

cdiener:
libSBML has moved to conda-forge a while ago to provide windows builds: https://github.com/conda-forge/python-libsbml-feedstock, there is also the experimental package that includes more extension by default: https://github.com/conda-forge/python-libsbml-experimental-feedstock. With the default bioconda packages those conda-forge packages will have precedence over the bioconda ones, so you might want to send in PRs to those as well.

jotech:
but the ones on conda-forge are only the python libraries right?

cdiener:
Oh yeah good point, sorry.

dpryan79 (approved):

dpryan79:
@bioconda-bot please merge

PR #28073, opened 2021-04-17
Add PeptideDB

ZimmerA:
PeptideDB


recipes/peptidedb/meta.yaml
```diff
@@ -0,0 +1,32 @@
+{% set version = "0.0.1" %}
+{% set version_string = "0.0.1-beta" %}
```
bgruening, line 2: if the version of the package is 0.0.1 but the version_string is 0.0.1-beta ... what happens if you are out of beta? Will this be 0.0.2? We usually recommend to only package stable software. Is that possible here?

recipes/peptidedb/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+  sha256: '{{ sha256 }}'
+
+build:
+  noarch: generic
+  number: 0
+  script:
+    - dotnet tool install --add-source $SRC_DIR --tool-path $PREFIX/bin --version {{ version_string }} PeptideDB
+
+requirements:
+  build:
```
bgruening, line 20: can this also be `host` and not specify anything under build?

recipes/peptidedb/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - dotnet-sdk
+  run:
+    - dotnet-sdk
+
+test:
+  commands:
+    - dotnet tool list --tool-path $PREFIX/bin
```
bgruening, line 27: is it also possible to call something like `PeptideDB --help` here?

recipes/peptidedb/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+about:
+  home: https://csb.bio.uni-kl.de/
+  license: MIT
+  summary: PeptideDB
```
bgruening, line 32: A somehow more meaningful help would be nice. And a description with some more text.
There is also something like `dev_url:` where you can point to the github repo if you like.

bgruening (commented):
Thanks @ZimmerA! Some comments inline.

ZimmerA:
> Thanks @ZimmerA! Some comments inline.

Thank you for the tips @bgruening!
I addressed them in my latest commits and we should be good to merge now.

ZimmerA:
@bioconda/core this should be ready to merge

bgruening (approved):

bgruening:
@ZimmerA cool beans!

bgruening:
@bioconda-bot please merge

PR #28084, opened 2021-04-19
Update: metaSNV attempt to unblacklist to address dependency resolution issues of motus-2.6.0

AlessioMilanese:
We are trying to re-build the metaSNV in conda to resolve:
https://git.embl.de/costea/metaSNV/-/issues/4

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -474,7 +474,6 @@ recipes/roprofile
 recipes/chanjo
 recipes/xxmotif
 recipes/phylip
-recipes/metasnv
 recipes/gqt
 recipes/diamond_add_taxonomy
 recipes/shapemapper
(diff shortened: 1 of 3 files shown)


AlessioMilanese:
@unode @hjruscheweyh

unode:
As this is not a new version I believe `build` should be incremented instead of set to 0.

unode:
metasnv [seems to have been blacklisted](https://github.com/bioconda/bioconda-recipes/blob/8ac7a6bfa47052815f0cf2e0ca26648695375389/build-fail-blacklist#L477) due to bulk update failures and is preventing the build. This is probably what caused the issue with https://git.embl.de/costea/metaSNV/-/issues/4

Can you remove metasnv from the blacklist file to see if we can fix it here?

unode (approved):
We are now back to green. Looks ready to be merged for me.

unode:
For context to anyone reviewing:
* One of the internal Makefile recipes used `$(CC)` to compile C++ code and hardcoded `CC=g++`

The above commits change this using `sed` expressions by:
1. Commenting out the hardcoded `CC=g++` line
2. Replacing `$(CC)` with `$(CXX)`

These changes will be pushed upstream so they will be removed in future releases of metaSNV.

PR #28132, opened 2021-04-21
Update gromacs_mpi to include nompi support

longr:
Testing to see if nompi support and mpi support can exist together in one build.

----

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

<table>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/recipes/gromacs_mpi/2021/build.sh
+++ b/recipes/gromacs_mpi/2021/build.sh
@@ -25,8 +25,12 @@ for ARCH in SSE2 AVX_256 AVX2_256 AVX_512; do
     -DGMX_SIMD="${ARCH}"
     -DCMAKE_INSTALL_BINDIR="bin.${ARCH}"
     -DCMAKE_INSTALL_LIBDIR="lib.${ARCH}"
-    -DGMX_MPI=ON
   )
+  if [[ "${mpi}" == "nompi" ]]; then
+      cmake_args+=(-DGMX_MPI=OFF)
+  else
+      cmake_args+=(-DGMX_MPI=ON)
+  fi
   cmake .. "${cmake_args[@]}"
   make -j "${CPU_COUNT}"
   make install
@@ -49,7 +53,12 @@ ${SRC_DIR}/src/gromacs/hardware/identifyavx512fmaunits.cpp
 # Create wrapper and activation scripts
 # Variable declaration from MPI script fewer changes if left in.
 
-gmx='gmx_mpi'
+
+if [ "${mpi}" = 'nompi' ] ; then
+    gmx='gmx'
+else
+    gmx='gmx_mpi'
+fi
 
 mkdir -p "${PREFIX}/etc/conda/activate.d"
 touch "${PREFIX}/bin/${gmx}"
--- a/recipes/gromacs_mpi/2021/conda_build_config.yaml
+++ b/recipes/gromacs_mpi/2021/conda_build_config.yaml
@@ -1,3 +1,4 @@
 mpi:
   - mpich
   - openmpi
+  - nompi
--- a/recipes/gromacs_mpi/2021/meta.yaml
+++ b/recipes/gromacs_mpi/2021/meta.yaml
@@ -4,6 +4,7 @@
 {% set version = "2021.1" %}
 {% set md5sum = "3c13b7dc3d8627904dc401dc893934d6" %}
 {% set build = 0 %}
+{% set build = build + 100 %}  # [mpi == 'nompi']
 
 package:
   name: gromacs_mpi
@@ -11,7 +12,10 @@ package:
 
 build:
   number: {{ build }}
-
+  string: >-
+    nompi_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}  # [mpi == 'nompi']
+    mpi_{{ mpi }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}  # [mpi != 'nompi']
+    
 source:
   url: http://ftp.gromacs.org/pub/gromacs/gromacs-{{ version }}.tar.gz
   md5: {{ md5sum }}
@@ -32,18 +36,19 @@ requirements:
     - khronos-opencl-icd-loader  # [osx]
     - libhwloc 2.*
     - fftw
-    - {{ mpi }}
+    - {{ mpi }} # [mpi != 'nompi']
   run:
     - ocl-icd # [linux]
     - khronos-opencl-icd-loader  # [osx]
     - ocl_icd_wrapper_apple # [osx]
     - fftw
     - libhwloc 2.*
-    - {{ mpi }}
+    - {{ mpi }} # [mpi != 'nompi']
 
 test:
   commands:
-    - gmx_mpi -version
+    - gmx -version  # [mpi == 'nompi']
+    - gmx_mpi -version  # [mpi != 'nompi']
 
 about:
   home: http://www.gromacs.org/


longr:
Hello @mabraham @mbargull @mba @stain and @douglowe:

After a comment Doug made the other day, I came back to look at this, and realised that the problems we ran into last time I tried to have both an MPI and non-MPI version of GROMACS were not related to MPI, and just a failure to understand the build process, and some other configuration options.   After a quick look, I decided to see if I could enable 'nompi' on the MPI version of GROMACS.  I *think* this has been a success. I wanted to ask you some question:
- Is the build time too long? Could the build start fluctuating and hitting the time out occasionally?
- Would this be a better setup than 2 packages?
- Does the config/build look fine?
- Any other comments?

Thanks,
Robin

mabraham:
Looks fine to me. Can't comment on the build type, but there's no GROMACS-related reason that it would fluctuate.

longr:
Thanks @mabraham were you able to check the build.sh to make sure I have not enabled/disabled some parts by mistake, or used them incorrectly?  Also, would a single package be preferable for users over 2 packages?  Or would a non-mpi and mpi one be better for them?

mbargull:
> Is the build time too long? Could the build start fluctuating and hitting the time out occasionally?

Yes and yes.

GROMACS isn't something that builds particularly fast, i.e., the actual build (`make -j "${CPU_COUNT}"`) takes around 20 minutes. Having three variants, i.e., `nompi`, `openmpi`, `mpich`, takes this to around an hour, which is still reasonable.
However, since a single variant not only builds GROMACS once but 4 times completely from the ground up, this creates a very long build job (which, on a busy day with many other package updates can cause delays since we still use CircleCI with its limited resources).

> Would this be a better setup than 2 packages?

Yes, I don't think this should be 2 separate packages. But, as written above, it not ideal to create such long build jobs.
Meaning: In my opinion GROMACS' build should be adjusted to allow builds of the different SIMD variants without having to build from the ground up for each of them, i.e., so we don't have `4 * 20 minutes` but rather something like `20 + 4 * (20 * 0.2) minutes` of build time per OpenMP variant.

mbargull:
> In my opinion GROMACS' build should be adjusted to allow builds of the different SIMD variants without having to build from the ground up for each of them

What I meant with this was something along the lines of
`set_source_files_properties(${files_that_benefit_from_simd} PROPERTIES COMPILE_FLAGS ${simd_flags_for_gromacs})`
such that only parts of the code have to be recompiled for each SIMD variant instead of setting those flags globally causing a whole rebuild.

> Would this be a better setup than 2 packages?

@longr, IMO, it makes sense to apply these changes to the `gromacs` recipe and remove the `gromacs_mpi` one. If it is necessary (e.g., people already depend on `gromacs_mpi` explicitly), then you can create an additional output for the `gromacs` recipe with name `gromacs_mpi` that only depends on `requirements: { run: [ {{ pin_subpackage("gromacs", exact=True) }} ] }` and has a `skip: True  # [mpi != 'nompi']` selector.

longr:
> What I meant with this was something along the lines of
> `set_source_files_properties(${files_that_benefit_from_simd} PROPERTIES COMPILE_FLAGS ${simd_flags_for_gromacs})`
> such that only parts of the code have to be recompiled for each SIMD variant instead of setting those flags globally causing a whole rebuild.
> 
I don't know if this is possible, but I will have a look and ask @mabraham and the gromacs devs and see what we can do.

> @longr, IMO, it makes sense to apply these changes to the `gromacs` recipe and remove the `gromacs_mpi` one. If it is necessary (e.g., people already depend on `gromacs_mpi` explicitly), then you can create an additional output for the `gromacs` recipe with name `gromacs_mpi` that only depends on `requirements: { run: [ {{ pin_subpackage("gromacs", exact=True) }} ] }` and has a `skip: True # [mpi != 'nompi']` selector.

The MPI package is very new, and perhaps not used widely yet, but this could be used to with draw support for those who have used it.

mbargull:
> The MPI package is very new, and perhaps not used widely yet, but this could be used to with draw support for those who have used it.

Adding the `gromacs_mpi` output to the `gromacs` recipe as outlined above would not withdraw support since `gromacs_mpi` would still be built, just without the redundancy we currently have.

mabraham:
> > In my opinion GROMACS' build should be adjusted to allow builds of the different SIMD variants without having to build from the ground up for each of them
> 
> What I meant with this was something along the lines of
> `set_source_files_properties(${files_that_benefit_from_simd} PROPERTIES COMPILE_FLAGS ${simd_flags_for_gromacs})`
> such that only parts of the code have to be recompiled for each SIMD variant instead of setting those flags globally causing a whole rebuild.

That could certainly work, but is unlikely to be adopted by the broader GROMACS team. Even if we identified the compute-sensitive parts, that would still be well over half of the code base. Of the rest, even things like string handling in our pre-processing benefits from AVX. So the likely conclusion is that adding complexity to the build system in order to slow down some parts of the code so that multi-configuration CI builds take less time isn't worth it.

Dropping e.g. SSE4.1 gives a 25% speedup while only inconveniencing a small fraction of users.

mabraham:
> Thanks @mabraham were you able to check the build.sh to make sure I have not enabled/disabled some parts by mistake, or used them incorrectly? 

I didn't try anything, but things look fine

> Also, would a single package be preferable for users over 2 packages?

It's certainly easier, but very few people who'd use a bioconda package from GROMACS in production genuinely want the MPI support. It is used, e.g. for delivering GROMACS tutorials for the algorithms that genuinely want MPI support for teaching with toy problems.

> Or would a non-mpi and mpi one be better for them?

I think an separate MPI-enabled GROMACS package is easy enough for the minority of bioconda users who would want an MPI-enabled GROMACS.

There's no need to have the MPI build able to do both of openmpi and mpich, so there's a CI time reduction available there.

longr:
> Dropping e.g. SSE4.1 gives a 25% speedup while only inconveniencing a small fraction of users.

Thanks, @mabraham. This is what I was originally going to email about.  I was also going to ask about both MPI varients. My understanding is that GROMACS recommends MPICH, but from what I can see OpenMPI is more widely adopted.  Do you think we should drop one of these? and if that would be worth it?

mabraham:
> > Dropping e.g. SSE4.1 gives a 25% speedup while only inconveniencing a small fraction of users.
> 
> Thanks, @mabraham. This is what I was originally going to email about. I was also going to ask about both MPI varients. My understanding is that GROMACS recommends MPICH, but from what I can see OpenMPI is more widely adopted. Do you think we should drop one of these? and if that would be worth it?

The GROMACS install guide suggests either flavour. My personal suggestion is to drop MPICH if a 1/3 reduction in build time makes something more palatable.

I withdraw my earlier comment about SSE4.1 - we've already dropped support for that.

If we take the above yardstick of 20 minutes per build, and do only [nompi|openmpi] and [sse2|avx|avx2|avx512] then it's only 8 configurations, rather than 12. If it was still too long, I'd drop avx512. Most users of bioconda won't use it on server-grade current hardware that actually gets the benefit, and it'll be a year or two until laptop-grade chips with AVX512 will get benefit from it (the models people have today all don't).

longr:
Thanks.   The only reason I would see to drop avx512,  is if we think we can drop another option later when we reenable avx512.

mabraham:
> Thanks. The only reason I would see to drop avx512, is if we think we can drop another option later when we reenable avx512.

If we dropped avx512 now, then when we add it back in a year or two, if CI time is still a factor, we'd drop plain avx

mabraham:
2h 28 seems like a reasonable runtime for the 2 MPI and 3 SIMD flavours remaining. How does that feel, @mbargull ?

mbargull:
Yes, 2.5 hours is well a away from the "might hit CI timeout if CI server experiences slowdowns"-limit.

I had also run the build locally last weekend to get a feeling for the extend of
> Even if we identified the compute-sensitive parts, that would still be well over half of the code base.

and I'd agree that applying the SIMD-related flags selectively probably doesn't have extreme influence on the overall build time.
(Overall the single translation units seem to have a compilation time of about a second each on average -- if one would want to speed up the build time generally, resorting to profiling (e.g., Clang's `-ftime-trace) or (I'd guess more likely) grouping .cpp files to single translation units would offer more opportunity for improvement.)

From my PoV, the only thing left for this PR is to decide on the package's name. Having a single `gromacs` package with different MPI builds would be the common way as is done at conda-forge and the one I'd prefer. For legacy/compatibility purposes, a `gromacs_mpi` output can be provided then (see https://github.com/bioconda/bioconda-recipes/pull/28132#issuecomment-826595670 ff.)
(As is now, we'd have `gromacs` = `gromacs_mpi nompi_*`, i.e., not only duplicate recipes but also duplicate packages.)

PR #28260, opened 2021-04-28
Add recipe for alfa

fgypas:
Add recipe for ALFA (https://pypi.org/project/alfa/)

### DIFF
--- a/recipes/alfa/meta.yaml
+++ b/recipes/alfa/meta.yaml
@@ -34,7 +34,7 @@ about:
   home: "https://github.com/biocompibens/ALFA"
   license: MIT
   license_family: MIT
-  license_file: "https://github.com/biocompibens/ALFA/blob/master/LICENSE.txt" 
+  license_file: "LICENSE.txt" 
   summary: "A simple software to get a quick overview of features composing NGS dataset(s)."
 
 extra:
--- a/recipes/alfa/meta.yaml
+++ b/recipes/alfa/meta.yaml
@@ -16,7 +16,7 @@ build:
 
 requirements:
   host:
-    - matplotlib >=3.0,<3.1
+    - matplotlib-base >=3.0,<3.1
     - numpy >=1.15,<1.16
     - pip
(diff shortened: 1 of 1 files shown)


fgypas:
I don't get why this fails. The test was running for 5 and 1/2 hours...

dpryan79:
@fgypas I fixed the resolve conflict but (A) the license file is missing (not a problem, we can ignore that) and (B) I of course don't know how strictly we need to follow the version restrictions you had originally specified.

fgypas:
Hi @dpryan79 Thanks a lot. 
The version restrictions are based on this: https://github.com/biocompibens/ALFA/blob/master/setup.py I am also not sure, but I could test locally.
For the LICENCE I think that they are using MIT, so that should be fine as well I think. Please let me know, of you think we should process.

dpryan79:
@fgypas Can you contact the package author and ask for advise? That'd be the ideal path forward.

fgypas:
Hi @dpryan79 Done. 
Please see here: https://github.com/biocompibens/ALFA/issues/8

fgypas:
Hi @dpryan79 
The conclusion is that we can be more lenient with the versions, but we do not have a specific recommendation (you can see the comments here: https://github.com/biocompibens/ALFA/issues/8
How do you think we should proceed?

dpryan79:
@fgypas I've made an edit that I suspect will pass testing and match common package restrictions. Numpy and pybedtools should be pretty stable and I rather doubt a progress bar library is going to have frequent API-breaking changes. If you want to be a bit more conservative, perhaps set restrictions such that only a single major release version of dependencies is allowed (e.g., `numpy >=0115,<2`). That'd also be a perfectly acceptable way to proceed.

dpryan79 (approved):

dpryan79:
Anyway, I've gone ahead and approved this PR so you can have it merged whenever you'd like :)

fgypas:
@BiocondaBot please merge

PR #28506, opened 2021-05-14
Add recipe for tin-score-calculation

fgypas:
Add recipe for tin-score-calculation


fgypas:
The missing dependency is added in conda-forge, but now I get a weird pip error. The dependency is CPython only so my guess is that it's related to this?

dpryan79:
The error was due to the missing requirements.txt, which pip was looking for.

fgypas:
@BiocondaBot please add label

dpryan79 (approved):

dpryan79:
@bioconda-bot please merge

PR #28580, opened 2021-05-19
Update ragtag recipe for v2.0.0

malonge:
update for RagTag v2.0.0

### DIFF
--- a/recipes/ragtag/meta.yaml
+++ b/recipes/ragtag/meta.yaml
@@ -50,4 +50,4 @@ extra:
   recipe-maintainers:
     - malonge
   identifiers:
-    - doi:10.5281/zenodo.4773870
+    - doi:10.5281/zenodo.3887139
--- a/recipes/ragtag/meta.yaml
+++ b/recipes/ragtag/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "RagTag" %}
-{% set version = "1.1.1" %}
+{% set version = "2.0.0" %}
 
 package:
   name: "{{ name|lower }}"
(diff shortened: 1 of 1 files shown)


dlaehnemann (approved):
Can't see any dependency changes, so this looks good to go:
https://github.com/malonge/RagTag/compare/v1.1.1...v2.0.0

@malonge If you want to avoid having to update the Zenodo DOI in the future, there is a generic one that will always resolve to the newest one. That way, you can usually let the bioconda bot do the updates and you just have to review them (on Zenodo, you can find this generic DOI below the list of version DOIs on the right hand side):
https://doi.org/10.5281/zenodo.3887139

malonge:
Thanks @dlaehnemann ! That would be very useful! If I wanted to update to the generic DOI, should I update the recipe and then open a new PR?

recipes/ragtag/meta.yaml
```diff
@@ -50,4 +50,4 @@ extra:
   recipe-maintainers:
     - malonge
   identifiers:
-    - doi:10.5281/zenodo.4563117
+    - doi:10.5281/zenodo.4773870
```
dlaehnemann, line 53:
```suggestion
    - doi:10.5281/zenodo.3887139
```

malonge: thank you!

dlaehnemann:
I quickly updated it via the ```suggestion mechanism in `Files changed` tab. Is this PR good to go, then?

malonge:
Yes I believe so. Thank you!

dlaehnemann:
@BiocondaBot please merge

PR #28591, opened 2021-05-20
Update CellRank numba version requirement

WeilerP:
Require `numba<0.53.0` due to theislab/scanpy#1756.

----

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

<table>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/recipes/cellrank/meta.yaml
+++ b/recipes/cellrank/meta.yaml
@@ -28,7 +28,7 @@ requirements:
     - joblib >=0.13.1
     - matplotlib-base >=3.3.0
     - networkx >=2.2
-    - numba >=0.51.0
+    - numba >=0.51.0,<0.53.0
     - numpy >=1.2.0
     - pandas >=1.2.0
     - pygam >=0.8.0
--- a/recipes/cellrank/meta.yaml
+++ b/recipes/cellrank/meta.yaml
@@ -11,7 +11,7 @@ source:
 
 build:
   noarch: python
-  number: 0
+  number: 1
   script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed -vv"
 
 requirements:


WeilerP:
@BiocondaBot please add label

dlaehnemann:
Thanks for tracking this down. From the issue you reference (and the issues that references), it seems like this is a problem with `scanpy` and `scvelo`. So I would think it makes more sense to limit the version numbers there, because otherwise their respective installation won't work either when they pull in newer numba, right? And once they have this fixed, it should in turn be unnecessary, here. Or am I missing something important?

WeilerP:
> Thanks for tracking this down. From the issue you reference (and the issues that references), it seems like this is a problem with `scanpy` and `scvelo`. So I would think it makes more sense to limit the version numbers there, because otherwise their respective installation won't work either when they pull in newer numba, right? And once they have this fixed, it should in turn be unnecessary, here. Or am I missing something important?

Yes, you are right, it should be fixed in `scanpy` as well, I guess. Will open a PR there as well. Thanks for pointing it out.
The reason I opened this PR is that the `numba` version is pinned [here](https://github.com/theislab/cellrank/blob/f4a08e62f9edb4d3e97f82c6129ad083d43cfd04/requirements.txt#L9) to the correct one in the CellRank repo as well. Using `pip install cellrank`, the correct version of `numba` is installed, `conda install -c conda-forge -c bioconda cellrank` installs `numba>=0.53.0`.
Any thoughts/preferences on this @michalk8?

dlaehnemann:
Ah, I see. But I guess the same logic applies there. If it is properly pinned in the respective packages where the problem actually arises, a further pinning in cellrank should not be necessary?

Actually, from a quick look at scvelo, it seems that it also only indirectly uses numba through its scanpy dependency. So pinning it in scanpy in the following place and here on bioconda should do the trick, I think:
https://github.com/theislab/scanpy/blob/bcfa925f6941ab30395e56ef6d2aedef8e16a563/pyproject.toml#L61

michalk8:
@WeilerP
In general, this is `umap-learn` issue. I've played around the versions, seems to have been fixed in `umap==0.5.1`:
```python
import numba as nb
import umap
print(nb.__version__)  # 0.53.{0,1}
print(umap.__version__)  # 0.5.1
```
I'm fine with pinning`numba` in `scanpy` only. While we're at the topic of `umap-learn`, I'd also consider unpinning the upper bound on `umap-learn` in `scVelo`, see [here](https://github.com/theislab/scvelo/commit/b5d8063dda94192616d19236c38dfa8acfb40b24#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552) (should no longer be needed).

WeilerP:
Sounds good to me. The reason I opened this PR was that `pip install cellrank` and `conda install -c conda-forge -c bioconda cellrank` behave the same way.

I'll close this then and discuss it further in the scanpy issue. Thanks for pointing out the `umap-learn` version used in `scvelo`! I actually already removed the upper bound - it's only on `develop` for now and hasn't been merged into `master` yet.

PR #28660, opened 2021-05-25
Rebuild mosdepth

matthdsm:
Rebuild [`mosdepth`](https://bioconda.github.io/recipes/mosdepth/README.html) to update pinnings

[install with bioconda](http://bioconda.github.io/recipes/mosdepth/README.html) [Conda](https://anaconda.org/bioconda/mosdepth/files)

Info | Link
-----|-----
Recipe | [`recipes/mosdepth`](https://github.com//bioconda/bioconda-recipes/tree/bump/mosdepth/recipes/mosdepth) (click to view/edit other files)
Releases |[https://github.com/brentp/mosdepth/tags](https://github.com/brentp/mosdepth/tags)
Author | @brentp
***

(Unable to detect reason for pinning rebuild. Predicted hash has changed, however.)

***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).


matthdsm:
@bioconda/build-system @bioconda/osx, any idea what went wrong here? The CI log isn't very clear (to me).

Thanks
M

recipes/mosdepth/meta.yaml
```diff
@@ -1,36 +1,42 @@
...
 requirements:
   build:
     - {{ compiler('c') }}
+    - git
+    - nim
   host:
-    - pcre
     - htslib
-  run:
+    - htslib >=1.10
```
mbargull, line 22: Good catch, makes sense.
I've re-added the constraint-less so that we get the version from pinning itself too (`conda-build` will only touch those entries without constraints when it applies the pinning; if we don't have those, the `htslib` dep might get out of sync with the pinning, e.g., use a future `htslib=1.13` when other packages haven't yet been rebuilt for that version; which would in turn mean you wouldn't be able to install `mosdepth` into the same environment with the other packages until those are rebuilt.)

mbargull (approved):

mbargull:
@BiocondaBot, please merge
Thanks @matthdsm!

matthdsm:
No! Thank you for the help!

PR #28890, opened 2021-06-08
PR CodAn and MITGARD

pedronachtigall:
Trying to make it work with an updated version of CodAn that will work in macOS.


pedronachtigall:

PR #28923, opened 2021-06-09
Codan recipe

pedronachtigall:
CodAn recipe.

----

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

<table>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/recipes/codan/build.sh
+++ b/recipes/codan/build.sh
@@ -5,8 +5,8 @@ chmod +x bin/*
 cp bin/* $PREFIX/bin
 
 if [[ "$OSTYPE" == "darwin"* ]]; then
-  chmod +x ../work/for_MacOS_users/*
-  cp ../work/for_MacOS_users/* $PREFIX/bin
+  chmod +x for_MacOS_users/*
+  cp for_MacOS_users/* $PREFIX/bin
 fi
 
 # This will allow them to be run on environment activation.
--- a/recipes/codan/build.sh
+++ b/recipes/codan/build.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+
+mkdir -p $PREFIX/bin
+chmod +x bin/*
+cp bin/* $PREFIX/bin
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  chmod +x for_MacOS_users/*
+  cp for_MacOS_users/* $PREFIX/bin
+fi
+
+# This will allow them to be run on environment activation.
+for CHANGE in "activate" "deactivate";
+do
+  mkdir -p "${PREFIX}/etc/conda/${CHANGE}.d"
+  #cp "${RECIPE_DIR}/${CHANGE}.sh" "${PREFIX}/etc/conda/${CHANGE}.d/${PKG_NAME}_${CHANGE}.sh"
+done
+echo "#!/bin/sh" > "${PREFIX}/etc/conda/activate.d/${PKG_NAME}_activate.sh"
+echo "export PERL5LIB=$PREFIX/lib/perl5/site_perl/5.22.0/" >> "${PREFIX}/etc/conda/activate.d/${PKG_NAME}_activate.sh"
+echo "#!/bin/sh" > "${PREFIX}/etc/conda/deactivate.d/${PKG_NAME}_deactivate.sh"
+echo "unset PERL5LIB" >> "${PREFIX}/etc/conda/deactivate.d/${PKG_NAME}_deactivate.sh"
--- a/recipes/codan/build.sh
+++ b/recipes/codan/build.sh
@@ -5,8 +5,8 @@ chmod +x bin/*
 cp bin/* $PREFIX/bin
 
 if [[ "$OSTYPE" == "darwin"* ]]; then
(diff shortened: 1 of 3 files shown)


pedronachtigall:
@BiocondaBot please add label

recipes/codan/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+  run:
+    - python >=3.6
+    - biopython 1.69
```
dpryan79, line 20: Do you need this exact version of biopython, or will newer versions also presumably work? The difficulty with requesting an exact version is that in 2 years biopython will have newer versions out and this package will no longer be compatible with them, even if it could reasonably work.

pedronachtigall: First, I noticed that you changed the build.sh by commenting the lines 6,7 and 8.
```
#mkdir -p $PREFIX/for_MacOS_users
#chmod +x for_MacOS_users/*
#cp for_MacOS_users/* $PREFIX/for_MacOS_users
```
We need to keep the lines uncommented. They are necessary to make CodAn work properly in macOS systems (see post-link.sh script).
Please, keep it as it was before.

About the biopython version, I will make a quick test.

recipes/codan/meta.yaml
```diff
@@ -0,0 +1,39 @@
+{% set version = "1.2" %}
+
+package:
+  name: codan
+  version: {{ version }}
+
+source:
+  url: https://github.com/pedronachtigall/CodAn/archive/refs/tags/v{{ version }}.tar.gz
```
dpryan79, line 8: Is the source code for the single compiled binary I see in there not available anywhere? We normally just compile those as part of the recipe.

recipes/codan/meta.yaml
pedronachtigall: About the biopython, it can be the latest release of biopython.
It will work fine.

recipes/codan/meta.yaml
pedronachtigall: As I mentioned in the last comment.
We keep a version compiled to Linux dist, in the bin folder, and a version compiled to macOS in the folder "for_MacOS_users".
That the reason we need lines 6,7 and 8 to be uncommented.
During the installation tests, the build did not copy it, we need to have these commands in the post-link.sh script.

pedronachtigall: We don't need the build part in requirements added in the meta.yaml file.
```
  build:
    - {{ compiler('c') }}
    - {{ compiler('cxx') }}
```

recipes/codan/build.sh
```diff
@@ -0,0 +1,25 @@
...
+cp bin/* $PREFIX/bin
+
+mkdir -p $PREFIX/for_MacOS_users
+chmod +x for_MacOS_users/*
+cp for_MacOS_users/* $PREFIX/for_MacOS_users
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  chmod +x for_MacOS_users/*
+  cp for_MacOS_users/* $PREFIX/bin
+fi
```
dpryan79, line 14: Since this is compiled, it might make since instead to keep this as a non-generic package and `mv` the mac binary into bin/ in the build script. This then means only the appropriate binaries are shipped with each of the OSX and Linux packages. You can then shorten post-link.sh

pedronachtigall: Unfortunately (and I don't know the reason), the installation process does not allow these lines (from 11 to 14) to properly work in the build.sh. That's why we had to make a directory in the PREFIX to keep the "for_MacOS_users" in the installation path and, then, use the post-link.sh to cp the binary to the final bin folder (we did several tests to finally reach into this working build). I decided to keep these lines also in the build.sh, in the case conda get any release that allows it to work in the build.sh. I think it is better to have it copied twice and CodAn works in MacOS than not copying and CodAn fails in MacOS.
Let me know your thoughts about it... Should we remove these lines from the build.sh?

dpryan79: It was labeled as a generic package, so the script was never run on OSX. Generic packages are built on Linux for all architectures.

pedronachtigall: It may be the reason behind the failure when keeping these commands in the build.sh only. Is there a better way to set this? Should we try to change the noarch option (the Conda documentation is not good enough to help design it in a better way).
In our tests, using these commands in the "post-link.sh" makes it properly work in OSX, despite being build in a Linux architecture.

dpryan79: Yeah, the best way is to remove the noarch stuff. The linter will complain, so you can either put the compilers in the build section or surpress the linting error https://bioconda.github.io/contributor/linting.html . Then there will be a separate package for OSX and the `cp` stuff in the post-link script can be removed.

dpryan79 (approved):

dpryan79:
@bioconda-bot please merge

PR #29347, opened 2021-06-30
Update rgccacmd to 3.0.1

ecamenen:
Update to fix https://github.com/galaxyproject/tools-iuc/pull/3764


recipes/rgccacmd/meta.yaml
```diff
@@ -1,23 +1,24 @@
...
   noarch: generic
-  number: 1
+  number: 0
   rpaths:
     - lib/
     - lib/R/lib/
 
 requirements:
   host:
+    - wget
```
bgruening, line 21: you need that in the run section, do you?

ecamenen: It seems to be the case with my new version. Without it, the last commit failed.

ecamenen: This was to download the command line interface that was migrated to another directory in the latest version. Once installed during the build, it should be fine, but I can add it in the run section to be sure. Strangely the similar version of conda in my personal directory doesn't need "wget" (https://anaconda.org/icm-iconics/r-rgcca).

bgruening (approved):

bgruening (approved):

bgruening:
@BiocondaBot please merge

PR #29447, opened 2021-07-07
MaxQuant 1.6.17.0 use dotnet instead of mono

npinter:
In some instances newer MaxQuant versions complaining about not using .NET. It depends on the useDotNetCore parameter in the mqpar file, which by default is set to `True` in version >=1.6.17.0. In older versions the paramater is set to `False` (f.e. in 1.6.10.43). To avoid any further incompatibilities I would suggest to switch to `dotnet`.


bgruening:
Can you create a versioned subdir for that version?

PR #29511, opened 2021-07-12
Update Dependencies for scanpy-scripts

LustigePerson:
The dependencies in the receipt were not up-to-date with the dependencies in the Git repo ([setup.py](https://github.com/ebi-gene-expression-group/scanpy-scripts/blob/v1.0.0/setup.py)).

## Changed:
  - `scanpy>=1.6.0`              --> `scanpy>=1.8.0`
  - `bbknn>=1.3.12,<1.5.0`  -->  `bbknn>=1.5.0`

## Added:
  - `scrublet`

## Removed:
  - `fa2`
  - `numpy<1.19.5`

### DIFF
--- a/recipes/scanpy-scripts/meta.yaml
+++ b/recipes/scanpy-scripts/meta.yaml
@@ -35,23 +35,23 @@ requirements:
     - python>=3.6
   run:
     - anndata
+    - bbknn>=1.5.0
     - click<8
+    - h5py<3.0.0
+    - harmonypy>=0.0.5
     - leidenalg
     - loompy
     - louvain
-    - multicore-tsne
     - matplotlib-base
-    - pandas
-    - h5py<3.0.0
+    - mnnpy>=0.1.9.5
+    - multicore-tsne
     - packaging
+    - pandas
     - python>=3.6
     - scanpy>=1.8.0
     - scipy 
     - scrublet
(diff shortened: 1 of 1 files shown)


LustigePerson:
@BiocondaBot please add label

dpryan79 (approved):

dpryan79:
@bioconda-bot please merge

pcm32:
Hi, this merge is a bit problematic for us as it will change the underlying package version without changing scanpy-script version, which has a good potential to wreck things in a number of galaxy tools and others that we have. We were going to make a new release where we would be adding some fixes to 1.8.0 so that data tests on scanpy-scripts, not run here, would pass. Would it possible to unmerge this and delete the new container @dpryan79 ?

pcm32:
We were planning with @pinin4fjords to make a release 0.5.0 of this conda package very soon, and we could incorporate those changes there, but I'm troubled by the build bump only on the containers without the scanpy-script version change, as it means that Galaxy modules requesting for the 0.4.1 or so version that this had will now be getting a container (or conda package) with a scanpy version for which they haven't been tested.

pcm32:
Ok, my bad, the move to 1.8.0 had already happened, I thought that we hadn't done it yet. So all good here, don't mind me @dpryan79 .

PR #29593, opened 2021-07-16
metaplatanus

rkajitani:
I have developed a de novo metagenome assembler, MetaPlatanus, and I would be grateful if this tool is registered in Bioconda.
The directory of MetaPlatanus' recipe (metaplatanus) was added.

Websites:
[GitHub](https://github.com/rkajitani/MetaPlatanus)
[previous website](http://platanus.bio.titech.ac.jp/metaplatanus)

### DIFF
--- a/recipes/metaplatanus/build.sh
+++ b/recipes/metaplatanus/build.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+mkdir -p $PREFIX/bin/sub_bin
+
+make CXX=${CXX}
+cp metaplatanus $PREFIX/bin
+cp src/scripts/tgsgapcloser_mod $PREFIX/bin
+cp -r sub_bin/* $PREFIX/bin/sub_bin
+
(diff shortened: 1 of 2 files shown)


telatin (commented):
The build has an hard coded URL that will result in a fixed release being downloaded. Is this intended or is it better to have the tool in Bioconda?

rkajitani:
@telatin Thank you for the comment! The hard-coded URL in build.sh was for NextPolish, which is not in conda. Now I have integrated it into the MetaPlatanus repo, and the build script have been simplified. It looks better.

telatin:
> @telatin Thank you for the comment! The hard-coded URL in build.sh was for NextPolish, which is not in conda. Now I have integrated it into the MetaPlatanus repo, and the build script have been simplified. It looks better.

Cool, consider that you might prefer to have it as an external dependencies if you'd like to get updates from the package authors. I'll approve this one

telatin (approved):

PR #29875, opened 2021-08-02
Add spritz

acesnik:
Adding Spritz, a tool for generating proteogenomic databases from RNA-Seq to be used for searching proteomic data.

----

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

<table>
  <tr>
  </tr>
</table>

</details>


acesnik:
@BiocondaBot update

acesnik:
@BiocondaBot please add label

recipes/spritz/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+  url: https://github.com/smith-chem-wisc/Spritz/releases/download/{{ version }}/SpritzCMD.zip
+  sha256: f218e3ebb599b29960ae0245aa4a16ab8734ed7895d98f0772b1f7734dacfc10
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - sysroot_linux-64 2.17  # [linux64]
```
bgruening, line 17: why is this needed?

recipes/spritz/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+  run:
+    - snakemake-minimal =5.27.4
+    - mamba
```
bgruening, line 20: mamba and conda can be removed afaik, or are you installing dependencies during tool runtime?

recipes/spritz/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - python=3.8
```
bgruening, line 21: any reason why you pin python and dotnet?

recipes/spritz/meta.yaml
acesnik: Yes, Spritz installs dependencies during tool runtime using snakemake with conda/mamba.

recipes/spritz/meta.yaml
acesnik: Spritz's commandline tool uses dotnet 5.0, and it doesn't work with other frameworks/runtimes, so that needs to be pinned.

acesnik: It needs python 3. Should I just remove the pin, or pin it to `python=3`?

recipes/spritz/meta.yaml
bgruening: Why does needs to happen during tool runtime? Please be aware that cluster installations exists, where conda is read-only from worker nodes.

recipes/spritz/meta.yaml
acesnik: These are prerequisites accompanying dotnet's need for CentOS 7.

See:
https://conda-forge.org/docs/maintainer/knowledge_base.html#using-centos-7
https://github.com/conda-forge/dotnet-feedstock/blob/master/conda-forge.yml#L5
https://github.com/conda-forge/dotnet-feedstock/blob/master/recipe/meta.yaml#L47

Do you think I should remove this, since CentOS 7 is the default container for bioconda now?

recipes/spritz/meta.yaml
bgruening: We do not build for python2 anymore. So you can really remove that.

recipes/spritz/meta.yaml
acesnik: Spritz has three workflows right now, and not all tools are required if users just need one of the three. In my tests, this is more efficient, not least because spritz uses mamba and users tend to use conda installs by default. Spritz has an option for protected-access cluster setups, where resources (e.g. uniprot, ensembl databases) are downloaded beforehand. I believe snakemake, which is part of the backend of spritz, sets up environments (e.g. using pre-downloaded bioconda/conda-forge repositories) and then activates them before firing up worker nodes. I've tested this via `slurm` on a cluster available to me. Are there other cluster setups that are problematic here?

recipes/spritz/meta.yaml
acesnik: I'll give this recipe a try with these removed.

recipes/spritz/meta.yaml
acesnik: I'm giving it a try with these pins removed, other than for `dotnet-runtime`.

recipes/spritz/meta.yaml
acesnik: Thanks for the heads-up on the possible cluster installation issues, by the way. I'm happy to do some more testing of that issue before merging if that would be best and if I can find such an environment.

bgruening: All I can say is that installation of packages during runtime is not usual, a dependency on conda/mamba as runtime dependency is not usual. Why not splitting up your workflows into 3 packages? We don't have strict rules here about that I just wanted to point it out and this might cause problems down the road for your users.

acesnik: Makes sense!

The workflows aren't separate; they can be combined. For example, users can make databases with variants or transcript isoforms, or they can request both variants and isoforms together which yields a different result than they do separately. Similarly, users can request to do quantification of these features or not.

bgruening: But Bioconda is not a workflow repository but a tool repository :)
I know the lines a blurry, but this is the reason repos and communities like snakepipes, nf-core etc exists. Workflows are a different beast.

acesnik: Oh that's an interesting point. Snakepipes does look like it might be a more appropriate home for Spritz.

acesnik:
I'm going to close this for now and see about integrating spritz into snakepipes.
https://github.com/bioconda/bioconda-recipes/pull/29875#discussion_r685093297
https://github.com/maxplanck-ie/snakepipes

acesnik:
Hi @bgruening, just as an update, it sounds like snakepipes doesn't consider Spritz something they would consider merging. They don't view it as a general repository for snakemake workflows. https://github.com/maxplanck-ie/snakepipes/issues/782

PR #29973, opened 2021-08-09
Curves

svletana:
`build.sh` updated so it copies standard_*.lis files to a hidden `.curvesplus` directory in $HOME.

This is so the user will have these files available to use upon installation, since they are needed for Curves+ execution.


bgruening:
@svletana this is not how conda works. You are copying the files into the HOME dir of your __build__ environment and not in the HOME dir of your user. Also please note that cluster environments do not need to have a HOME dir. So this assumption is not always true (https://academic.oup.com/gigascience/article/8/5/giz054/5497810 - guideline 6).

svletana:
> @svletana this is not how conda works. You are copying the files into the HOME dir of your **build** environment and not in the HOME dir of your user. Also please note that cluster environments do not need to have a HOME dir. So this assumption is not always true (https://academic.oup.com/gigascience/article/8/5/giz054/5497810 - guideline 6).

Thanks for your input! I didn't think about cluster environments now that you mention it. I'll see how to get around this problem.

PR #29975, opened 2021-08-09
Update kaiju meta.yaml

pmenzel:
update sha256sum after re-making the release in the upstream repo

### DIFF
--- a/recipes/kaiju/meta.yaml
+++ b/recipes/kaiju/meta.yaml
@@ -6,10 +6,10 @@ package:
 
 source:
   url: https://github.com/bioinformatics-centre/kaiju/archive/v{{ version }}.tar.gz
-  sha256: 45de5d879a836aaf8f1547af95f592ab4c2ac71d8ac62eff7e3fa4f49772ac97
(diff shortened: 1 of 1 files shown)


recipes/kaiju/meta.yaml
```diff
@@ -6,10 +6,10 @@ package:
 
 source:
   url: https://github.com/bioinformatics-centre/kaiju/archive/v{{ version }}.tar.gz
-  sha256: 45de5d879a836aaf8f1547af95f592ab4c2ac71d8ac62eff7e3fa4f49772ac97
+  sha256: dfb755016fc99838c251b8adf4f5a7f12feef1be12dbdaee6595d837ec1dcf44
```
bgruening, line 9: Can you create a new version instead? We should usually not change the source during a build-number bump. Every change in the source should ideally get a new version. This is, for example, because we backup the tarballs.

pmenzel: hm, in principle I could make a new version, but the old v1.8.0.tar.gz is gone for good already anyways.

druvus (approved):

druvus:
@BiocondaBot please merge

druvus:
@BiocondaBot please update

PR #29986, opened 2021-08-09
Update SpydrPick to v1.3.0

santeripuranen:
Update recipe for SpydrPick v1.2.0 -> v1.3.0

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -261,7 +261,6 @@ recipes/bolt-lmm
 recipes/bmfilter
 recipes/unitig-counter
 recipes/probcons
-recipes/spydrpick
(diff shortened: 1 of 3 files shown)


santeripuranen:
OK, I've tried a few things, but my recipe still seems to be blacklisted. The `lint` stage of `bioconda-testing` says:
```
06:12:39 BIOCONDA INFO Constraining to 1 git modified recipes (spydrpick).
06:12:39 BIOCONDA INFO Ignoring 1 blacklisted recipes (spydrpick).
06:12:39 BIOCONDA INFO Processing 0 recipes.
```
How do I un-blacklist a recipe so that it actually gets built and tested?

PR #30484, opened 2021-09-10
Updated masurca to 4.0.5

dnbenso:
Update to recipe/masurca
* The masurca recipe was stuck on 3.4.2 and the latest version is now 4.0.5.
* It was necessary to pin the compiler to version 10 of gcc as version 11 failed to handle pointer integer comparison in global-1/CA8/src/AS_GKP/fastqAnalyze.C line 73.

@BiocondaBot please update


dnbenso:
Will retry later on to see if memory issue is fixed.

PR #30549, opened 2021-09-15
Add proteomiqon-joinquantpepionswithproteins

ZimmerA:
Results from PSMBasedQuantification contain detailed information about the quantification of every peptide, but only basic informations about the protein they originated from. This information is obtained during the ProteinInference. This tool takes the results from both tools and combines the quantification information of each peptide with the more accurate information about the corresponding protein including its q-value.


bgruening:
@bioconda-bot please merge

PR #30563, opened 2021-09-16
Add recipe for LIQUORICE

peneder:
Added a recipe to build LIQUORICE, a command-line-tool and python package for bias correction and quantification of changes in coverage around regions of interest in cfDNA WGS datasets. It can be used to detect tissue- and cancer specific epigenetic signatures in WGS data from liquid biopsies (see https://liquorice.readthedocs.io/en/latest/intro.html ).


peneder:
@BiocondaBot please add label

peneder:
The most current version of the recipe has now been tested on multiple systems (docker, Linux, macOS) and runs stably. Could someone from @bioconda please review and merge?

recipes/liquorice/conda-recipe/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - LIQUORICE = liquorice.LIQUORICE:main
+    - LIQUORICE_summary = liquorice.LIQUORICE_summary:main
+  noarch: python
+  script: {{PYTHON}} setup.py install
+
+requirements:
+  host:
+    - python >=3.7
+  run:
+    - python ==3.7.9
```
dpryan79, line 21: This really need exactly version 3.7.9 and nothing more recent? Similarly, are the exact version constraints below really needed?

recipes/liquorice/conda-recipe/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - pybedtools ==0.8.2
+    - pybigwig ==0.3.18
+    - scikit-learn ==0.24.2
+    - scipy ==1.7.1
+    - swifter ==1.0.9
+    - parallel
+    - seaborn ==0.11.2
+    - pysam ==0.16.0.1
+  run_constrained:
+    - ray-core ==1.1.0
```
dpryan79, line 39: These can be moved to the `run` section

recipes/liquorice/pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
```
dpryan79, line 1: Please remove this file, setup.cfg and setup.py and then move the meta.yaml into this directory.

dpryan79 (changes requested):
Just a bit of moving files around needed really.

recipes/liquorice/conda-recipe/meta.yaml
peneder: When I had ray-core under "run", the installation failed on macOS, because ray-core is not available for macOS via conda (see https://anaconda.org/conda-forge/ray-core/files). I am therefore keeping it optional (my tool can also run without ray-core, but will be slower). I do need version 1.1.0 of ray-core if installed at all, though, because the program crashes otherwise.  If you have any better ideas I'd be happy to hear!

dpryan79: OK, fair enough

dpryan79 (approved):

dpryan79:
@bioconda-bot please merge

PR #30651, opened 2021-09-22
Specify channel for pbcore to access v1.2.7

bishemma1:
This version of PBHoover (1.0.7) requires pbcore v1.2.7, which is not available on anaconda. I have built a conda package for it and uploaded it to my own channel.

We want people to be able to conda install pbhoover without having to pip install pbcore 1.2.7.

### DIFF
--- a/recipes/pbhoover/meta.yaml
+++ b/recipes/pbhoover/meta.yaml
@@ -25,8 +25,8 @@ requirements:
     - htslib
     - bcftools
     - pyvcf
-    - pbcore >=1.2.10
     - numpy
+    - bishemma1::pbcore
 
 test:
   commands:


simonbray:
Bioconda doesn't allow installation of dependencies from external channels unfortunately - the correct way would be to make the older version available via bioconda, if it is really necessary.

bishemma1:
Thank you, I suppose the pbcore team would need to do that? Or is there a way I can submit it to bioconda? In the meantime we can ask people to pip install this version but ideally they wouldn't have to.

simonbray:
No, you can do it yourself - you should create a subdirectory of `bioconda-recipes/recipes/pbcore` called `1.2.7` and add a modified version of the recipe there. Then you can create a new PR.

bishemma1:
Oh I see, thank you I will do that then.

simonbray:
Feel free to ping me if you need a review

bishemma1:
I created a new PR: #30679

PR #30752, opened 2021-09-28
fqutils Institut Pasteur tools

fmareuil:
Adds fqutils package a series of tools for fastq sequences developed at the Institut Pasteur.

### DIFF
--- a/recipes/fqutils/build.sh
+++ b/recipes/fqutils/build.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+./configure --prefix=$PREFIX PERL=`command -v perl`
+make
+make install
+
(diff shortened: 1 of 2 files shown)


telatin:
Hello! Thanks for adding. Do you know that if your tool is on a git repository  the Biocondabot can create a new version for you every time you make a new release? If this is of interest you might want to change the recipe with the repository url

fmareuil:
Hi @telatin I'm sorry, I have contacted the developer for the git access and he hasn't answered yet. I will keep you informed.

raivivek:
@fmareuil There's also another tool called `fqtools` that does similar work (and maybe more) and seems to be more updated — https://github.com/alastair-droop/fqtools.

If the tool you've added is older and no longer maintained, then perhaps you might be interested creating a recipe for this tool first? Otherwise, we may need a way to disambiguate both.

PR #30793, opened 2021-09-30
Bump up deepbgc to 0.1.29

prihoda:
Bump up deepbgc to 0.1.29

### DIFF
--- a/recipes/deepbgc/meta.yaml
+++ b/recipes/deepbgc/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "deepbgc" %}
-{% set version = "0.1.28" %} # remember to update sha256 hash as well
+{% set version = "0.1.29" %} # remember to update sha256 hash as well
 {% set file_ext = "tar.gz" %}
 
 package:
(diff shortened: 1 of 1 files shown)


prihoda:
@BiocondaBot please add label

rpetit3:
This one was captured by the auto-update bot https://github.com/bioconda/bioconda-recipes/pull/30794, going to close

@prihoda please feel free to reopen if necessary

prihoda:
Thanks @rpetit3, got it. About the auto-update bot: How does it handle changed dependencies? Because in the previous release `0.1.28` https://github.com/bioconda/bioconda-recipes/pull/30602, an autoupdate was merged even though the dependencies in setup.py have changed, which caused a broken release on conda, because it kept the old conda dependencies.

rpetit3:
When there are new dependencies, we have to manually add them. Usually the tests fail due to the missing dependency and we can get them added it. But as you know personally that's not always the case.

prihoda:
Right, that makes sense. I should probably set up better tests :)

PR #30832, opened 2021-10-01
Update to v2.2 and add tests

skoren:
Update canu to latest release. Switch to using git packages which allows support of OS X and Linux. Switch jvm from openjdk which often fails due to memory over-allocation. Lastly, download/run a small assembly to ensure the installation is working as intended.


skoren:
Awesome, thanks

PR #30894, opened 2021-10-06
Add MetaNovo recipe

reid-wagner:
This adds a new recipe for the MetaNovo tool.

A few notes:
- I added `- should_be_noarch_python` to `skip-lints` because this does target specifically `linux-64`.  Other architectures are skipped with `skip: True`. I'm not sure why the linter expects `noarch: python`, but I used the precedent in #16556 to skip the check.
- I added some DSOs to the list of allowable missing DSOs. I expect that the user's Linux system should have these available for the packaged native dependency. In fact, these specific dependencies are not yet used (they may be at a later time) by the tool, but are included in the larger distribution that is used by MetaNovo, e.g. PepNovo in DeNovoGUI.

### DIFF
--- a/recipes/metanovo/build.sh
+++ b/recipes/metanovo/build.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+
+# Copy executable scripts
+declare -a metanovo_scripts=(
+    bin/bash/metanovo.sh
+    bin/bash/compomics.sh
+    bin/python/bp_export_tags.py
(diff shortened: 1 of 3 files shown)


reid-wagner:
@BiocondaBot please add label

reid-wagner:
@BiocondaBot please add label

jj-umn (approved):

jj-umn:
@BiocondaBot please merge

PR #30917, opened 2021-10-07
Add recipe for vafator

priesgo:
This PR adds a recipe for vafator. 

Vafator annotates variants in a VCF file with information about allele frequency and depth of coverage from multiple BAM files from tumor and or normal samples. Even though the functionality of vafator is simple this is useful to homogenise technical annotations like VAF across different variant callers.


dpryan79:
@bioconda-bot please merge

PR #31015, opened 2021-10-13
kaiju v1.8.1

pmenzel:
bump kaiju to v1.8.1 and extend build.sh so that it also copies the kaiju-multi binary file, see https://github.com/bioinformatics-centre/kaiju/issues/186


pmenzel:
Can the test for MacOS be disabled?

rpetit3:
@biocondabot please merge

PR #31428, opened 2021-11-06
[Add]create parallel-meta-suite recipe

qdu-bioinfo:
[Add]create parallel-meta-suite recipe

### DIFF
--- a/recipes/parallel-meta-suite/build.sh
+++ b/recipes/parallel-meta-suite/build.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+mkdir -p ${PREFIX}/bin
+
+make 
+
+cp -r bin/* ${PREFIX}/bin
+
+# # path for database and example
+ParallelMETA_path=${PREFIX}/${PKG_NAME}-${PKG_VERSION}
+mkdir -p ${ParallelMETA_path}
+
+# set VIBRANT DB PATH variable on env activation
+mkdir -p ${PREFIX}/etc/conda/activate.d ${PREFIX}/etc/conda/deactivate.d
+cat <<EOF >> ${PREFIX}/etc/conda/activate.d/pms.sh
+export ParallelMETA=${ParallelMETA_path}
+EOF
+
+cat <<EOF >> ${PREFIX}/etc/conda/deactivate.d/pms.sh
+unset ParallelMETA
+EOF
--- a/recipes/parallel-meta-suite/build.sh
+++ b/recipes/parallel-meta-suite/build.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 mkdir -p ${PREFIX}/bin
-
+cd ${SRC_DIR}
 make 
 
 cp -r bin/* ${PREFIX}/bin
--- a/recipes/parallel-meta-suite/build.sh
+++ b/recipes/parallel-meta-suite/build.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 mkdir -p ${PREFIX}/bin
-cd ${SRC_DIR}
-make 
+
+make CC=$CXX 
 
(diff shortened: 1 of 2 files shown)


qdu-bioinfo:
@epruesse I'm sorry to disturb you, but I have a problem here. While Circle CI test my recipe, I got a `g++: No such file or directory`. But I tested this recipe on my own PC with conda-build, it works well. Could you tell me what to do to solve this problem?

qdu-bioinfo:
@marcelm Sorry to disturb you. I got a problem here. While Circle CI testing my recipe, I got a `g++: No such file or directory`. It seems that the `src` folder can not be found. But I tested this recipe on my own PC with conda-build, it works well. Could you tell me how to solve this problem?

marcelm:
I was also able to build the package locally. The problem is that `make` hardcodes `g++` for compilation, which will not use the compiler provided by Conda, but the one installed on your system. This is one of the lines that is printed during the build:
```
g++ -c src/ExtractRNA.cpp -o bin/ExtractRNA.o -Wno-deprecated
```
It should actually use whichever compiler Conda has configured in the `$CXX` environment variable. This can often be achieved by calling make like `make CXX=${CXX}`.

marcelm:
Looking at the Makefile, I think you need `make CC=$CXX`

qdu-bioinfo:
Thanks a million. I have tried `make CC=$CXX`. It works! But this time I can not pass the MacOS test... The problem is clang compiler does not have `OpenMP` library. I tried to add `OpenMP`  and `llvm-OpenMp` in the `requirements` section, it didn't work. What should I do?

marcelm:
I don’t know how to fix macOS builds, sorry. If supporting only Linux is good enough for you, you can make your build section look like this to skip the macOS build:
```
build:
  number: 0
  skip: True  # [osx]
```
Otherwise, you may want to ask on the [chat](https://gitter.im/bioconda/Lobby#) otherwise.

qdu-bioinfo:
> I don’t know how to fix macOS builds, sorry. If supporting only Linux is good enough for you, you can make your build section look like this to skip the macOS build:
> 
> ```
> build:
>   number: 0
>   skip: True  # [osx]
> ```
> 
> Otherwise, you may want to ask on the [chat](https://gitter.im/bioconda/Lobby#) otherwise.

I will skip macOS and figure this one myself! I truly appreciate your help for the make command!

recipes/parallel-meta-suite/build.sh
```diff
@@ -0,0 +1,20 @@
...
+
+make CC=$CXX 
+
+cp -r bin/* ${PREFIX}/bin
+
+# # path for database and example
+ParallelMETA_path=${PREFIX}/${PKG_NAME}-${PKG_VERSION}
+mkdir -p ${ParallelMETA_path}
+
+# set VIBRANT DB PATH variable on env activation
```
marcelm, line 12: This comment isn’t quite correct, or is it? You can just delete that line, it’s clear what an activate/deactivate script does.

marcelm (approved):

recipes/parallel-meta-suite/build.sh
qdu-bioinfo: The comment is correct. The database which we used is very big, more than 700MB. I don't think add this to the `source` is a good choice. So I define a path in the activate script, the download command will download the database in this folder.

marcelm: I just found it confusing that the comment seems to imply that a variable named `VIBRANT_DB_PATH` or similar would be set, but the actual variable name is `ParallelMETA_path`

marcelm: I’ll just merge it

marcelm:
@biocondabot please merge

PR #31655, opened 2021-11-22
Update odgi to 0.6.3

AndreaGuarracino:


raivivek:
@AndreaGuarracino Looks like the build is timing out due to 1hr limitation on the CircleCI build. Could you check if the build.sh is using all CPU cores to build the package? Perhaps that could speed it enough to work. Setting `-j $CPU_COUNT` or `-j 2` for example to the `conda --build` command should help.

AndreaGuarracino:
@raivivek, thank you for your suggestion. I tried and failed. It looks like it was near the end T_T

raivivek:
Yeah. Looks like it helped but not enough. I'm not sure if we can do anything except to wait for new build system. Perhaps someone else can help.

AndreaGuarracino:
@BiocondaBot please add label

dpryan79 (approved):

PR #31744, opened 2021-11-28
Update meme to 5.4.1

lpryszcz:
@BiocondaBot please update


lpryszcz:
I've tested the build locally and it built ok. 
CI fails due to timeout 
image

lpryszcz:
thx @raivivek although it still fails due to time out even with `make -j2`. how many threads max can we use with CI?

raivivek:
I believe the builds are done on a `Linux Medium` instance which as 2 CPUs.

mencian:
Closing in favor of #54384.

PR #31872, opened 2021-12-07
add new recipe pysbol 0.0.2

kenza12:
pySBOL is a SWIG-Python wrapper around [libSBOL](https://github.com/SynBioDex/libSBOL) a module for reading, writing, and constructing genetic designs according to the standardized specifications of the Synthetic Biology Open Language (SBOL).

----

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

<table>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/recipes/pysbol/meta.yaml
+++ b/recipes/pysbol/meta.yaml
@@ -1,0 +1,42 @@
+{% set name = "pysbol" %}
+{% set version = "0.0.2" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/brsynth/pySBOL/archive/{{ version }}.tar.gz
+  sha256: 65d79b956e1249370307bc2a1e8638f776bd10b0559a6ab507477a65bcedc74c
+
+build:
+  number: 0
+  noarch: python
+  script: {{ PYTHON }} -m pip install -vv .
+
+requirements:
+  host:
+    - python =3.7
+    - pip
+  run:
+    - python =3.7
+    - requests
+
+test:
+  source_files:
+    - tests/
+  imports:
+    - {{ name|lower }}
+  requires:
+    - pytest
+
+about:
+  home: https://github.com/brsynth/pySBOL/
+  summary: A module for reading, writing, and constructing genetic designs according to the standardized specifications of the Synthetic Biology Open Language (SBOL).
+  license: Apache-2.0
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - breakthewall
+    - kenza12
--- a/recipes/pysbol/run_test.sh
+++ b/recipes/pysbol/run_test.sh
@@ -1,0 +1,4 @@
+#!/usr/bin/env bash
+
+echo -e "\n\n*** TEST ***\n\n"
+pytest -v


breakthewall:
Hi, I'm one of the maintainers of this recipe. Thanks!

kenza12:
@BiocondaBot please add label pysbol 0.0.2 and merge it.

recipes/pysbol/run_test.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
```
marcelm, line 1: Instead of adding a separate file, you should add the `pytest -v` command to the recipe itself:
```
test:
  ...
  commands:
    pytest -v
  ...
```
It’s easier to maintain that way.

recipes/pysbol/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  sha256: 65d79b956e1249370307bc2a1e8638f776bd10b0559a6ab507477a65bcedc74c
+
+build:
+  number: 0
+  noarch: python
+  script: {{ PYTHON }} -m pip install -vv .
+
+requirements:
+  host:
+    - python =3.7
```
marcelm, line 19: This is quite restrictive especially for a package that provides a library (bindings in this case). As an application developer, I would not want to be restricted to Python 3.7.

In generaly, applications can have quite strict dependencies, but libraries should be quite open.

marcelm (changes requested):
I think you need to change the way in which you create the bindings in your package. Your repository contains pre-compiled libraries, for example at https://github.com/brsynth/pySBOL/tree/master/pysbol/lib/Linux/.libs . There are various reasons why it is usually not a good idea do this. The libraries are quite system-specific, for example, and it is lilkely that your Linux libraries will not work on old Linux versions. Conda has mechanisms in place to compile things in such a way that they work on many Linux versions, but just re-packaging your bindings would bypass this and it is likely that a broken package would result.

Also, your package includes `libxml2`, `libxslt` and zlib. However, the proper way would be to specify these as dependencies in the Conda recipe.

You should also be aware that if you re-distribute binaries of other libraries, you need to make sure that you follow the licenses. I have not checked the license of these libraries in particular, but you may be required to at least give credit, for example.

marcelm:
Ok, now I found https://pypi.org/project/pySBOL/, which appears to be the actual project that you would like to package. I’m confused. What is the purpose of the repository at https://github.com/brsynth/pySBOL?

marcelm:
Also according to the README at https://github.com/SynBioDex/pySBOL:
> pySBOL has been deprecated. Please use [pySBOL2](https://github.com/synbiodex/pysbol2) instead.

breakthewall:
Hi,

actually we are building a federative platform (galaxy-synbiocad.org) and we are currently publishing tools on anaconda channels. Some tools have been developed a couple of years ago by external partners and some of them leverage on pySBOL, which is a Python wrapper over `libsbol` C library. These collaborators are not involved anymore in the development of these tools and tools work fine as they are. Knowing that pySBOL is not available on anaconda, this is what we tried to make it compliant with conda:

- replace `import sbol` by `import sbol2`: the first program of the chain works good but the output file raises an error in the second program;
- contact `pySBOL` developers to create a tag so that we would be able to create a conda package: ok on macOS and Windows but linking problems on Linux with `_libsbol.so` that is embedded in `pySBOL`;
- install `pySBOL` with pip and copy the content of the package into our first program (`sbml2sbol`), which worked and this package is published in `conda-forge` (!);
- because we have several tools depending on `pySBOL`, we've created our own repo (`brsynth/pySBOL`) with the content of the `pip` package and try to publish on anaconda.

About `libxml2`, `libxslt` and `zlib`, the reason why we included it is because the `_libsbol.so` is linked against specific versions of these dependencies **which are not available** on anaconda.

I hope these explanations answer to your questions and will succeed to convince you about our approach.

recipes/pysbol/run_test.sh
kenza12: Hi @marcelm , I deleted the separate file and add `pytest -v` to the recipe but the tests failed with this error:

```
10:53:08 BIOCONDA INFO (ERR) [Dec  8 10:53:08] SERR bash: pytest: command not found
10:53:08 BIOCONDA INFO (ERR) [Dec  8 10:53:08] ERRO Task processing failed: Unexpected exit code [127] of container [a110877139b2 step-a2306b4b49], container preserved
.10:53:41 BIOCONDA INFO (OUT) ..
10:53:41 BIOCONDA ERROR COMMAND FAILED (exited with 1): mulled-build build-and-test pysbol=0.0.2--pyhdfd78af_0 -n biocontainers --test bash -c '/usr/local/env-execute true && . /usr/local/env-activate.sh && pytest -v' --channels conda-forge,file:///home/circleci/project/miniconda/conda-bld,bioconda,defaults --involucro-path /home/circleci/project/miniconda/lib/python3.7/site-packages/bioconda_utils/involucro

10:53:41 BIOCONDA ERROR TEST FAILED: recipes/pysbol
10:53:41 BIOCONDA INFO (COMMAND) conda build purge
10:53:42 BIOCONDA ERROR BUILD SUMMARY: of 1 recipes, 1 failed and 0 were skipped. Details of recipes and environments follow.
10:53:42 BIOCONDA ERROR BUILD SUMMARY: FAILED recipe recipes/pysbol


Exited with code exit status 1

CircleCI received exit code 1
```

marcelm: Sorry about that, you need to move it back to the script because the 'test: requires:' dependency is not installed during the `mulled-build` test, see https://bioconda.github.io/contributor/guidelines.html#test-dependencies

marcelm:
Let me describe how it would look in an ideal world. Possibly it’s necessary to deviate from this, but at the moment, this is what I would aim for.

Bioconda packages should generally be compiled from sources if available. Packages should not bundle their dependencies, but instead list their dependencies in their metadata such that Conda will install them along with the package. This reduces space on disk because a dependency used by multiple packages only needs to be installed once, but more importantly allows dependencies to be updated independently of the main package in case of bug or security fixes. zlib 1.2.3 is from 2005, and the other libraries also appear to be quite outdated. It’s almost certain there are security problems in them.

So the correct way in my view would be to first create a Bioconda (or possibly conda-forge) package for libSBOL. The libSBOL package would be built from source and properly depend on zlib, libxslt etc. Because libSBOL is built from source, it is not a problem to use newer versions of these libraries.

The second step would then be to create a package for the Python bindings, which depends on libSBOL. Looking closer into this, it appears that libSBOL actually comes with the Python bindings already included, so the second step would be to extend the libsbol recipe such that it also builds the Python bindings.

marcelm:
Hi again,
I now see that have already opened https://github.com/conda-forge/staged-recipes/pull/16673 at conda-forge and gotten some feedback. It would have been wise to mention this when this PR was opened. Your are otherwise wasting the time of volunteers.

I agree with all the advice over there: You cannot re-package binaries. My impression is that Bioconda is less strict about this if everything else would be too complicated, but it would then need to be justified.

My advice stands: Try to first package libsbol instead. You can for example use the `libsbml` package as a starting point because it also uses cmake.

dpryan79:
Yeah, we're much less restrictive about binary repackaging, but it's still a method-of-last-resort, since those things are just so incredibly likely to break. I'll echo what @marcelm said and suggest getting libsbol compiling by itself as a package first.

PR #32093, opened 2021-12-20
Update orsum 1.3.0

ozanozisik:
Update orsum 1.3.0


ozanozisik:
bioconda.bioconda-recipes (lint lint) gives error "build_number_needs_bump: The recipe build number should be incremented" but when I change it CircleCI requires build number reset.

PR #32156, opened 2021-12-23
Update GRIDSS 2.13.0

d-cameron:
Updated recipe to 2.13.0.

Reflects the changed build requirements for gridsstools


d-cameron:
@BiocondaBot please add label

jmarshall:
The _build.sh_ change appears to replace using the system (i.e, Conda's) htslib with building a bundled htslib. What is the reason for using a bundled htslib, and why then does _meta.yaml_ still list (build-time) requirements on the `htslib` package?

d-cameron:
The upstream package uses static linking to a specifc htslib version so this recipe is just reflecting that. Changing to system htslib is an issue for the upstream tool maintainer ;) (https://github.com/PapenfussLab/gridss/issues/498).


> why then does meta.yaml still list (build-time) requirements on the htslib package?

It's there to ensure that the transitive htslib library dependencies (e.g. libdeflate) exist when building. A bit hacking but no more so than having manually specifying them.

dpryan79 (approved):

PR #32200, opened 2021-12-28
Update FOCUS: adding setuptools

linsalrob:
Both setuptools and setuptools_scm need to be included in the meta.yaml to provide pkg_resources which was failing

```
    - setuptools >=38.6.0
    - setuptools_scm
```


linsalrob:
@BiocondaBot please add label

dpryan79 (approved):

PR #32201, opened 2021-12-28
bump/super focus

linsalrob:
Both setuptools and setuptools_scm need to be included in the meta.yaml to provide pkg_resources which was failing

```
    - setuptools >=38.6.0
    - setuptools_scm
```


linsalrob:
Ignore this PR. It is failing because all the intermediate steps

PR #32214, opened 2021-12-29
Update pbhoover dependency versions

emjbishop:
As described in the commit, we discovered that PBHoover was buggy unless using these versions of:
pbh5tools <=0.8.0
bcftools <=1.3.1
pyvcf <=0.6.7
pbcore <=1.2.7

So I updated the meta.yaml to fix this.


dpryan79:
@bioconda-bot please update

dpryan79 (approved):

PR #32300, opened 2022-01-06
Nextclade: add run_test.sh

corneliusroemer:
Add run_test.sh to catch more bugs before pushing out new version

@pvanheus do you have experience with tests and bioconda?

It appears as though our release 1.8.0 is broken for macOS and this wasn't noticed because the test is currently too simple

Hence, I'm trying to add a more sophisticated test than `nextclade --help` through `run_test.sh` but I've never added such a test to a bioconda recipe so I'm not sure I'm doing it right.

### DIFF
--- a/recipes/nextclade/meta.yaml
+++ b/recipes/nextclade/meta.yaml
@@ -14,7 +14,7 @@ source:
     sha256: 5017669852d5d624e9f677036470a3cb9a7225c644b5e357fc27c49a9b602d71                             # [arm64]
 
 build:
-  number: 0
+  number: 1
   binary_relocation: False
 
 requirements:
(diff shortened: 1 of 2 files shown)


corneliusroemer:
It should fail not pass, weird.
https://dev.azure.com/bioconda/bioconda-recipes/_build/results?buildId=9388&view=logs&j=1b052f1d-4456-52f0-9d43-71c4c5bd734d&t=edc48dcd-1fc2-5e3b-7036-7be9cea00123&l=127

corneliusroemer:
`set -e` didn't help.

This is weird. Can someone help me? `run_test.sh` fails but check passes?

https://dev.azure.com/bioconda/bioconda-recipes/_build/results?buildId=9390&view=logs&j=1b052f1d-4456-52f0-9d43-71c4c5bd734d&t=edc48dcd-1fc2-5e3b-7036-7be9cea00123&l=126

dpryan79:
@corneliusroemer Is it still returning 0 as the exit status? That's the most likely explanation.

corneliusroemer:
Thanks for figuring this out @dpryan79, that seems to be the reasons indeed as reported here: https://github.com/nextstrain/nextclade/issues/679

PR #32420, opened 2022-01-14
Remove R dependencies

nh13:
Remove R as a dependency.  It's not strictly required, and inflates the environment significantly.  

For the maintainers, is there any way to make an optional features for this recipe?

### DIFF
--- a/recipes/fgbio/meta.yaml
+++ b/recipes/fgbio/meta.yaml
@@ -16,9 +16,6 @@ requirements:
   run:
     - openjdk >=8
     - python
-    - r-base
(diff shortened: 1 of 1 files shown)


cbrueffer:
Not sure about your question, but Picard suffered from the same issue and we created a `picard-slim` without R dependencies in addition to the full-fledged `picard`.

nh13:
@cbrueffer I think that's a great idea, to have a `fgbio-slim` or `fgbio-minimal` recipe.  I'll close this PR and do that instead.  Thank-you for the direction!

PR #32548, opened 2022-01-22
bump gemBS 3.5.5_IHEC

JakeLehle:
Waiting to have [#88](https://github.com/heathsc/gemBS/issues/88) resolved. I'm updating gemBS to the most current 3.5.5 version to fix the issue with running this step in the pipeline.
```bash
gemBS map
```


JakeLehle:
Okay, I'll update the patch file for this tomorrow and see if I can fix this.

PR #32549, opened 2022-01-23
bump gembs v3.5.5_IHEC

JakeLehle:
I will probably need help with this patch job, but at least I'm learning. Going for v3.5.5, fingers crossed.


JakeLehle:
Okay I'm stumped. The updated patch I have in this pull request currently have should have been able to patch version 3.5.1 up to 3.5.5 which will fix an issue im having with the gemBS map step [#88](https://github.com/heathsc/gemBS/issues/88) but there is an error with applying the patch which I didn't have when I ran it locally after downloading the 3.5.1 tar ball and testing out the patch file that I made.

Any help on this would be greatly apprechiated!

JakeLehle:
Learning more every day!

These changes will push gemBS to v3.5.5 and fix an issue I was having with running v3.5.1.
Hey, @bioconda/core please take a look at my changes and approve them when you get a chance!

Best,
Jake

dpryan79 (approved):

PR #32614, opened 2022-01-26
gemBS troubleshooting

JakeLehle:
I just got gemBS updated to the most recent v3.5.5 but while testing out the installation this morning from anaconda I saw this did not fix a previous error I was having [#88](https://github.com/heathsc/gemBS/issues/88). I looked over everything with the installation and I think the build.sh is jumping in too deep and making the installation file in the tools/utils directory when it really should take one step back in the tools directory and make the installation from there.

This pull request is to see if this changes to the build.sh fixes that.


JakeLehle:
Hey @bioconda/core or @dpryan79 I have a small question that I'm sure that you have seen before and can help me fix quickly.

Okay, I made some improvements to setting up the gemBS package from my own forked repo (replacing submodule linked repos with their actual target files) which I think can get past having to do a bunch of patching to get gemBS to build correctly. HOWEVER, I hit a dumb snag. During the compilation, gcc throws this error.

fatal error: zlib.h: No such file or directory

I know have zlib in the meta.yaml file, and I can see in the configure file for that part of the make process that it found the bzlib.h presence but no zlib.h presence to be found.


I looked it up, and it looks like on linux this can happen if [haven't told the compiler to link the library that contains the functions that you've used](https://stackoverflow.com/questions/22337896/how-to-fix-undefined-references-to-inflate-deflate-functions). Maybe, I don't know. [Looks like this issue happens sometimes](https://stackoverflow.com/questions/36374267/how-to-fix-fatal-error-zlib-h-no-such-file-or-directory), so I'm not alone at least.

That thread suggests linking the files with something like 
gcc -o foo foo.c -lz

I haven't had a chance to write C code yet so this is a good place to learn something new. 

Here is what I'm thinking to try and fix this.
1. Try to link the files in by adding that line of code with the right names to the .configure file for that part of the make process.
2. Add that code to the actual file.c that is throwing the error.
3. Add that code to the build.sh before I invoke the make process.
4. Look for a package like libz-dev or zlib1g-dev to add to the meta.yaml file (I'm pretty sure these should be covered by having zlib in the meta.yaml file already)

Are any of those options going in the right direction? 
Thanks for any advice on this!

JakeLehle:
Update, I fixed my issue with the missing zlib.h file. 
Now I just have to fix all the new things that broke after that part and I'm golden.

JakeLehle:
Okay @bioconda/core  and @bioconda/osx I'm stuck but I'm in the home stretch!

I got the whole gemBS pipeline almost entirely set up (patchless I might add, the way are scientific overlords intended), but right at the part where C compiles the bs_call tool, C was throwing an error that a clock_gettime command couldn't find the real-time library. So  I rolled up my sleeves, read up on C, and made a patch to the tools/bs_call/src/Makefile adding -lrt to the end of the gcc command to link the real-time library to the makefile.

Yay now the code works and the linux build version passes. I'm amazing! 
But boo, the OSX version can't find the glibc library to match that super old -lrt trick. I'm trash. 

I tried adding in glibc (I know it should be standard) or even adding in the extra gsl library after seeing it was a part of the bs_call [meta.yaml](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bs_call/meta.yaml) file. No dice I'm still stuck at OSX. 

But why am I having to add this patch workaround in the first place? Seems like this should have been fixed after glibc was bumped past [2.17 in 2013](https://stackoverflow.com/questions/47703239/ld-library-not-found-for-lrt). I don't have to pull this trick when I build the file locally on Ubuntu 20.04 I just run make in the tools dir sit back and watch it compile on the first try. Does the Linux lint test run on some super old version of linux? 

Please help.
I'm gonna go take a walk.

JakeLehle:
@dpryan79 you beautiful wonderful man. Yeah I would not have known about adding that trick to skip [py < 37]

Do you like what I did with the build.sh to conditionally add in the -lrt option if it's installing on a linux system?
I got that idea from the [dawg build.sh](https://github.com/bioconda/bioconda-recipes/blob/25ee21573c577aa1ae899e0f7fbbc848d8a63865/recipes/dawg/build.sh).

JakeLehle:
Well, that was an absolute slog. I really hope this fixes my previous issue [#88](https://github.com/heathsc/gemBS/issues/88) by adding the files that were linked to the gemBS repo by submodules but weren't being included in the tar.gz and thus weren't being installed in the build process.

@bioconda/core please review and merge this branch and I'll test out the package as soon as it updates on anaconda.org.

wen-chen (approved):

Issue #32718, opened 2022-02-02
Several bowtie versions don't install properly


NickSto:
I'm finding that all bowtie 1.2.* versions will look like they install successfully, but fail when I try to execute, giving this error:
```
$ bowtie --version
bowtie-align: error while loading shared libraries: libtbb.so.2: cannot open shared object file: No such file or directory
```

## Steps to reproduce
```
conda create -y -n test
conda install -y -n test bowtie=1.2.1.1
conda activate test
bowtie --version
```

## Environment
```
$ conda info

     active environment : None
            shell level : 0
       user config file : /home/me/.condarc
 populated config files : /home/me/.condarc
          conda version : 4.11.0
    conda-build version : 3.21.7
         python version : 3.9.10.final.0
       virtual packages : __linux=5.10.0=0
                          __glibc=2.31=0
                          __unix=0=0
                          __archspec=1=x86_64
       base environment : /home/me/src/installations/miniconda3  (writable)
      conda av data dir : /home/me/src/installations/miniconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://conda.anaconda.org/conda-forge/linux-64
                          https://conda.anaconda.org/conda-forge/noarch
                          https://conda.anaconda.org/bioconda/linux-64
                          https://conda.anaconda.org/bioconda/noarch
                          https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /home/me/src/installations/miniconda3/pkgs
                          /home/me/.conda/pkgs
       envs directories : /home/me/src/installations/miniconda3/envs
                          /home/me/.conda/envs
               platform : linux-64
             user-agent : conda/4.11.0 requests/2.27.1 CPython/3.9.10 Linux/5.10.0-1055-oem ubuntu/20.04.3 glibc/2.31
                UID:GID : 1000:1000
             netrc file : None
           offline mode : False

```

NickSto:
So is this just an issue with my system?

I'm running into this while trying to build a new version of my tool (`dunovo`), which depends on `bowtie`. I'd like it to run on my system so I can test it, but it needs to work for everyone else too.

PR #32758, opened 2022-02-04
Update Du Novo: Initial version of 3.0 support.

NickSto:
I'm trying to make a package for Du Novo 3.0. The new tool version has a new dependency: BioPython.

I've also added a few other dependencies which it always depended on but weren't specified before, like samtools, bash, and a number of unix utilities.

I've had a lot of trouble getting this to build and run. I logged one problem in #32718. Based on @bgruening's suggestion, I pinned `tbb`, but even then, it still doesn't run properly unless I first manually install an old version of `tbb` in the conda environment (so far I've only tried 2018.0.3). But it sounds like this might just be my machine?

### DIFF
--- a/recipes/dunovo/build.sh
+++ b/recipes/dunovo/build.sh
@@ -5,7 +5,7 @@ cat > submodules.txt <<EOF
 kalign  0.3.0       makovalab-psu kalign-dunovo c0ef2de4a958aed47311ea86591debb3bada871143a6923bc6338ab2f99f2d5b
 utillib 0.1.1-alpha NickSto       utillib       961f5a3481d1c0dbe00c258b9df2de541e5605f3de4ff25bcc3cec22922e7c06
 ET      0.3         NickSto       ET            6b757b3ab3634b949f78692ac0db72e9264a1dceaf7449e921091d3ad0012eea
-bfx     0.5.0       NickSto       bfx           f5cc5c9c8a627a082f8b505135aec10b47c09647269ad92abe3a1fa2f2059048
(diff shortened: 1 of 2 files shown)


dpryan79:
Do you strictly require bowtie 1.2.* or would 1.3 also work? The metadata for 1.2 requires patching and Anaconda has yet to enable that for our channel.

NickSto:
Du Novo works with 1.0.0 up to 1.3.1, as far as I know. The version restriction for `bowtie` is simply what I could get to build and run. If anyone wants to take a crack at changing it, be my guest. I couldn't get anything else to work.

NickSto:
Oh, FYI there's also Makefiles in the `bfx` and `kalign` subdirectories that get executed.

I can also alter those Makefiles. What's the best practice here, to allow setting the `$(CC)` from outside? Is there a way to let it get set from an environment variable?

bgruening:
Yes, $CC should be configurable, also it would help you I think if you release complete tarballs with submodules included.

NickSto:
> Yes, $CC should be configurable, also it would help you I think if you release complete tarballs with submodules included.

I agree! Is there a Github feature or would I have to manually make and host the tarball?

bgruening:
https://github.community/t/create-release-that-contains-submodule/1329 You have to do this manually or on CI I think.

https://github.blog/2013-07-02-release-your-software/

NickSto:
Okay, `$(CC)` is configurable and the submodules are bundled now. And it looks like it's passing the checks.

dpryan79 (approved):

Issue #33361, opened 2022-02-23
samtools 1.15 exclude htslib 1.15  as dependency


maarten-k:
Samtools 1.15 excluded htslib 1.15 as a dependency.

`
htslib >=1.14,<1.15.0a0`

samtools 1.15 ends up now with an older version of htslib (1.14) as a dependency. 

Would a rebuild fix this problem or am I oversimplifying this issue?

maarten-k:
Artifact of rebuild is still limited to 
   "htslib >=1.14,<1.15.0a0",

maarten-k:
There is a separate file with htslib number in there. I updated it to 1.15

maarten-k:
Looks this is fixed in #33362

jmarshall:
This is now fixed for samtools by PR #33366. (Should also be fixed for bcftools as well, but that is still percolating through anaconda's server caches.)

Issue #30296, opened 2021-08-27
samtools on macosx is incompatible with freebayes


MartinL1106:
Hello, 
I'm working on MacOS Big Sur 11.5.2, I found that when I installed "freebayes" from bionconda, my samtools downgraded from 1.13 to 1.4.1, and then when I can'u use samtools, with the message:
dyld: Library not loaded: @rpath/libcrypto.1.0.0.dylib
  Referenced from: /Users/Use name/opt/miniconda3/bin/samtools
  Reason: image not found
Abort trap: 6

when I re-install samtools, the freebayes was removed..

But I have a python code needs both samtools and freebayes, do you have some idea for fixing this problem? 

Thanks for your help

jmarshall:
Bioconda currently has HTSlib 1.12 pinned, and the current freebayes package has therefore been built against that version.

Samtools 1.13 requires htslib 1.13, and has an exception to the pinning (via its own _conda_build_config.yaml_ file) that means it requires the htslib-1.13 package.

Hence freebayes and samtools-1.13 are incompatible. You could install them into separate conda environments and add both bin directories to $PATH for your python code, or you could use samtools-1.12 with

```
conda create -n myenv freebayes samtools=1.12
```

(Unfortunately the plain installation command with `samtools` rather than `samtools=1.12` has selected the ancient now mostly broken version instead of the 1.12 that it could have selected, so you need to specify it explicitly.)

Note that samtools-1.12 has a serious known bug: samtools/htslib#1279. But otherwise it's a lot better than 1.4.1 and you don't have to juggle openssl/libcrypto dependencies.

This is a pretty good demonstration that Bioconda's current policy of “pin HTSlib, but allow an exception for widely used packages like samtools & bcftools” is not really tenable…

MartinL1106:
> Bioconda currently has HTSlib 1.12 pinned, and the current freebayes package has therefore been built against that version.
> 
> Samtools 1.13 requires htslib 1.13, and has an exception to the pinning (via its own _conda_build_config.yaml_ file) that means it requires the htslib-1.13 package.
> 
> Hence freebayes and samtools-1.13 are incompatible. You could install them into separate conda environments and add both bin directories to $PATH for your python code, or you could use samtools-1.12 with
> 
> ```
> conda create -n myenv freebayes samtools=1.12
> ```
> 
[...]

PR #33362, opened 2022-02-23
update to  rebuild samtools to match with htslib 1.15

maarten-k:
See #33361


maarten-k:
Did not work out as planned:
htslib is still limited until 1.15
   "htslib >=1.14,<1.15.0a0",

maarten-k:
Check the artifacts and now depends on htslib 1.15

jmarshall:
The right way to fix this is to bump bioconda's pinning to htslib-1.15 and remove the _conda_build_config.yaml_ files from the samtools and bcftools recipes. That way everything that depends on htslib will remain compatible with each other, and samtools/bcftools won't be incompatible with everything else (cf #30296).

maarten-k:
So updating https://github.com/bioconda/bioconda-utils/blob/45a9c7216c7b739bbf5a39dcaa8b4402d35d801f/bioconda_utils/bioconda_utils-conda_build_config.yaml#L25 to 1.15 would do the trick? I do not feel too confident to play around with this after a long day. Can someone with more experience pick this up?

jmarshall:
This is #33366 and its companion bioconda-utils PR.

maarten-k:
@jmarshall  Thanks for the  PR (and all your great work on htslib and samtools !)

PR #33464, opened 2022-03-02
Add recipe for phyloacc

gwct:
This PR adds a recipe for [PhyloAcc](https://github.com/phyloacc/PhyloAcc), which is software for estimating substitution rates and rate shifts in conserved non-exonic regions.

The main code base is built and installed and the Python interface is copied to `$PREFIX/bin/` and its associated libraries to `$SP_DIR/phyloacc_lib/`.

### DIFF
--- a/recipes/phyloacc/meta.yaml
+++ b/recipes/phyloacc/meta.yaml
@@ -1,113 +1,66 @@
-# Note: there are many handy hints in comments in this example -- remove them when you've finalized your recipe
+#######################################################################
+# Conda recipe for installing PhyloAcc
+# Supported platforms: Linux
+# Author: Gregg Thomas (gwct@github)
+#######################################################################
 
-# Jinja variables help maintain the recipe as you'll update the version only here.
-# Using the name variable with the URL in line 14 is convenient
-# when copying and pasting from another recipe, but not really needed.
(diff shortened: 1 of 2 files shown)


gwct:
@BiocondaBot please add label

recipes/phyloacc/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - make
+  host:
+    - gsl
+    - openblas
+    - liblapack
+    - armadillo
+    - python >=3.6
```
dpryan79, line 36: Please remove ">=3.6" here and from the "run:" section. Then under "build:" on line 23 put "skip: True  # [py2k]`.

dpryan79 (changes requested):
See comment above

recipes/phyloacc/meta.yaml
gwct: Changes made in [ecf5e4a](https://github.com/bioconda/bioconda-recipes/pull/33464/commits/ecf5e4a4d8ecbf7001dae677576f34f73ae0222d). Thanks!

dpryan79 (approved):

PR #33444, opened 2022-03-01
Remove version pinning for ngscheckmate

SPPearce:
Remove the version pinning for bcftools and samtools in ngscheckmate as discussed after #33443 was merged.

PR #33489, opened 2022-03-03
Pin ngscheckmate samtools to 1.14. samtools mpileup is removed in 1.15

SPPearce:
samtools mpileup is removed from v1.15 so bam mode fails as the python script calls samtools mpileup.
Reverts my previous PR #33444 .


SPPearce:
@BiocondaBot please add label

PR #33498, opened 2022-03-03
Add annotwg recipe

fsandron:
Add annotwg, a tool to annotate big whole genome vcf files.


fsandron:
@BiocondaBot please add label

jpuritz (approved):

PR #33757, opened 2022-03-17
Pin Augur's Python version at 3.9 or lower

huddlej:
Until conda-forge and Bioconda support Python 3.10, we need to require Python 3.9 or lower for Augur.

----

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

<table>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/recipes/augur/meta.yaml
+++ b/recipes/augur/meta.yaml
@@ -9,7 +9,7 @@ source:
   sha256: ec8ae3972a3b786379796c1c7d2dd45e7790b2935402e8fe3a7f311ac1cc8478
 
 build:
-  number: 0
+  number: 1
   noarch: python
   entry_points:
     - augur = augur.__main__:main
@@ -17,11 +17,11 @@ build:
 
 requirements:
   host:
-    - python >=3.7
+    - python >=3.7,<3.10
     - pip
 
   run:
-    - python >=3.7
+    - python >=3.7,<3.10
     - bcbio-gff >=0.6.0,<0.7
     - biopython >=1.73,!=1.77,!=1.78
     - jsonschema >=3.0.0,<4


huddlej:
@BiocondaBot please add label

dpryan79 (approved):

dpryan79:
Why do we need to pin this to python <3.10? The package itself should work with any upcoming python 3 releases, since it's a pure python package. Maybe some dependencies aren't but that doesn't affect this package.

huddlej:
@dpryan79, you may be right that this isn't the correct solution to our problem; I should have provided more context in the PR description. This PR stems from an issue we've just noticed with [our Nextstrain installation documentation for Conda on Mac OS X](https://docs.nextstrain.org/en/latest/install.html#installation-steps) where the full installation no longer works with Python 3.10.

Our installation starts by creating an environment with the Nextstrain CLI, as shown below. This command ends up installing Python 3.10 in the new environment. Users run this command if they end up using our Docker image or if they use the "native" Conda installation.

```bash
$ mamba create -n nextstrain -c bioconda nextstrain-cli --yes
```

We activate the new environment and install Augur and some other tools. If we just install the latest Augur in this environment like so, we get the following error:

```bash
$ conda activate nextstrain
$ mamba install -c conda-forge -c bioconda augur=14.0.0 --yes
[...snip...]
Problem: package augur-14.0.0-pyhdfd78af_0 requires bcbio-gff >=0.6.0,<0.7, but none of the providers can be installed
```

So, the issue appears to be with Augur's dependencies. When we try to create a Python 3.10 environment with bcbio-gff, this fails, too:

```bash
$ mamba create -n bcbio -c conda-forge -c bioconda python=3.10 bcbio-gff
[...snip...]
Problem: package bcbio-gff-0.6.2-py_0 requires bx-python, but none of the providers can be installed
```

When we try to create a Python 3.10 environment with just bx-python, this also fails:

```bash
$ mamba create -n bx-python -c conda-forge -c bioconda python=3.10 bx-python
[...snip...]
Problem: package bx-python-0.7.1-np110py27_1 requires python 2.7*, but none of the providers can be installed
```

Looking at the available bx-python packages, I don't see any that support Python 3.10.

```bash
$ mamba search -c conda-forge -c bioconda bx-python
[...snip...]
bx-python                     0.8.13  py37h26d536a_0  bioconda
bx-python                     0.8.13  py37h2b10027_1  bioconda
bx-python                     0.8.13  py38h27f87a4_1  bioconda
bx-python                     0.8.13  py38h94f7190_0  bioconda
bx-python                     0.8.13  py39hc8e2088_0  bioconda
bx-python                     0.8.13  py39hebbeb7d_1  bioconda
```

This is what prompted @corneliusroemer to ask about maximum Python versions supported by Bioconda yesterday in gitter. I guess the lack of compiled bx-python for Python 3.10 relates to [this link you shared](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml#L691-L695) with the Python versions supposed by conda-forge.

So, given this context where Augur's dependency chain leads to tools that don't support Python 3.10, is there a better solution than pinning the max version for Augur? This solution would at least allow our installation docs to work again, by forcing the environment to downgrade from Python 3.10 to 3.9.

We could also pin the Python version to 3.9 in our docs when we first create the Nextstrain environment like so:

```bash
$ mamba create -n nextstrain -c bioconda nextstrain-cli python=3.9 --yes
```

But, this solution obscures the underlying problem that one of Augur's dependencies doesn't support Python 3.10.

dpryan79:
The proper solution to this is to give mamba the entire list of things you want to install at once:

    mamba create -n nextstrain -c conda-forge -c bioconda --yes nextstrain-cli augur=14.0.0 bcbio-gff bx-python

If the other packages are basically always needed then adding them to the "run:" section of nexstrain-cli makes sense, since then you can omit them when making the env.

huddlej:
@dpryan79 That makes sense. We were trying to minimize forking paths in our installation docs, but your example here makes it clear that we could replace our current `create`-then-`install` approach with two independent `create` commands for each path the user could take. Thank you for your advice here!

PR #34004, opened 2022-03-30
WIP: Update UCSC utils to v428

AngieHinrichs:
Work in progress:

When updating ucsc-fatovcf to v426, a patch was necessary in order to get the build scripts to use $PREFIX when $CONDA_BUILD is set.  The patch is no longer necessary so this removes the patch file and updates the recipe.  When the updated ucsc-fatovcf recipe passes tests, I plan to update other UCSC util recipes since most are out of date and many are in build-fail-blacklist.


aliciaaevans:
Closing as these recipes have been updated to newer versions.

PR #34107, opened 2022-04-04
Update phyloacc

gwct:
This updates the phyloacc recipe to use release version 2.0.0. This is related to #34104 , which was auto-generated, but in addition to changing the version number in the recipe I also need to add a dependency and update the build script. Apologies if this is the incorrect protocol for these types of changes. If this PR is the way to move forward then #34104 can be deleted.

----


gwct:
@BiocondaBot please add label

wen-chen (approved):

wen-chen:
@BiocondaBot please merge

dpryan79:

recipes/phyloacc/meta.yaml
```diff
@@ -32,15 +32,16 @@ requirements:
   host:
     - gsl
     - openblas
+    - blis
     - liblapack
     - armadillo
-    - python
+    - python>=3.7
```
dpryan79, line 38: Please remove `>=3.7` and change the `skip: true` online 25 to `skip: True  # [py < 37]`

dpryan79 (changes requested):
We should build for all supported python versions

recipes/phyloacc/meta.yaml
gwct: Changes made in https://github.com/bioconda/bioconda-recipes/pull/34107/commits/80b20488e3a55493706d1409a4805d4b903c4bb5. Actual change is `skip: True # [py < 38]` because of the numpy version.

dpryan79 (approved):

dpryan79:
@BiocondaBot please merge

PR #34374, opened 2022-04-19
Update verkko to v1.0 release

skoren:
Updated verkko to v1.0 release. Added rust dependency and updated snakemake version. Also added OS X compatibility.

### DIFF
--- a/azure-pipeline-master.yml
+++ b/azure-pipeline-master.yml
@@ -123,6 +123,7 @@ jobs:
 
       export OSTYPE="darwin"
       export CI="true"
+      export HOME=`pwd`
       bioconda-utils build recipes config.yml \
           --anaconda-upload \
           --git-range ${BUILD_SOURCEVERSION}~1 ${BUILD_SOURCEVERSION} \
--- a/azure-pipeline.yml
+++ b/azure-pipeline.yml
@@ -180,6 +180,7 @@ stages:
 
         export OSTYPE="darwin"
         export CI="true"
+        export HOME=`pwd`
         bioconda-utils build recipes config.yml \
             --git-range origin/"$SYSTEM_PULLREQUEST_TARGETBRANCH" HEAD
       displayName: Test
--- a/azure-pipeline.yml
+++ b/azure-pipeline.yml
@@ -181,7 +181,6 @@ stages:
         export OSTYPE="darwin"
         export CI="true"
         export HOME=`pwd`
-        echo "HOME outside of bioconda-utils is $HOME"
         bioconda-utils build recipes config.yml \
             --git-range origin/"$SYSTEM_PULLREQUEST_TARGETBRANCH" HEAD
       displayName: Test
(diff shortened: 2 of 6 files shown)


skoren:
@bioconda/core I'm getting rust errors in the OS X test but this runs fine on my local OS X machine. I think it might be a network connectivity issue or a rust configuration issue in the CI node:
```
2022-04-19T15:56:09.3490640Z INFO:bioconda_utils.utils:(ERR)   Downloaded atty v0.2.14
2022-04-19T15:56:09.3492520Z 15:56:09 BIOCONDA INFO (ERR)   Downloaded atty v0.2.14
2022-04-19T15:56:09.3597410Z INFO:bioconda_utils.utils:(ERR) thread 'main' panicked at '`UNKNOWN/.cargo/registry/src/github.com-1ecc6299db9ec823/atty-0.2.14/UNKNOWN/.cargo/registry/src/github.com-1ecc6299db9ec823/atty-0.2.14/src/lib.rs` is not absolute', src/tools/cargo/src/cargo/core/manifest.rs:252:9
2022-04-19T15:56:09.3600300Z 15:56:09 BIOCONDA INFO (ERR) thread 'main' panicked at '`UNKNOWN/.cargo/registry/src/github.com-1ecc6299db9ec823/atty-0.2.14/UNKNOWN/.cargo/registry/src/github.com-1ecc6299db9ec823/atty-0.2.14/src/lib.rs` is not absolute', src/tools/cargo/src/cargo/core/manifest.rs:252:9
```
The url looks very strange and isn't in our repo anywhere.

dpryan79:
I expect the failure is specific to the CI system since I see a nice big UNKNOWN in the path of the missing file. Can you skip that test and then install the package locally for testing?

skoren:
Wouldn't skipping the test skip it on user systems too? That wouldn't be ideal to confirm the installation worked. I think the issue is the same as I had in the build script before, namely $HOME is undefined on Darwin (or rather UNKNOWN) so any tools that use the home folder like rust or snakemake fail on Darwin. Adding this:
```
# taken from yacrd recipe, see: https://github.com/bioconda/bioconda-recipes/blob/2b02c3db6400499d910bc5f297d23cb20c9db4f8/recipes/yacrd/build.sh
 if [ "$(uname)" == "Darwin" ]; then

     # apparently the HOME variable isn't set correctly, and circle ci output indicates the following as the home directory
     export HOME=`pwd`
     echo "HOME is $HOME"
     mkdir -p $HOME/.cargo/registry/index/
 fi
```
seems to have fixed the build on Darwin so I'll try to see if it fixes the test as well. If not, I'll turn off the test on Darwin and return success so it's CI compliant.

skoren:

dpryan79:
Tests aren't run on user systems, the compiled packages are directly extracted and the resulting binaries immediately used. I should modify the CI to set $HOME on OSX now that you mention it.

dpryan79:
Ah, so HOME isn't one of the variables we re-export into environments it seems. I need to look into this, since then we wouldn't need to rely on zsh setting it on CI providers.

dpryan79 (approved):

dpryan79 (approved):

dpryan79:
@BiocondaBot please merge

I'll work on fixing the $HOME thing over in the bioconda-utils repo.

PR #34637, opened 2022-04-30
Update virchip

mehrankr:
Increased build number and changed the URL and sha256sum from the old bitbucket website to the new GitHub repo.

### DIFF
--- a/recipes/virchip/build.sh
+++ b/recipes/virchip/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+$PYTHON -m pip install -vv --disable-pip-version-check --no-deps --no-cache-dir --ignore-installed . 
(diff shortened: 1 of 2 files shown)


mehrankr:
Thanks for helping me getting the recipe up 2 years ago @dpryan79 
I changed the tarball link from the deprecated bitbucket repo to the GitHub repo.
I bumped up the build number too.

Could you please take a look at this.

Much appreciated,
Mehran

mehrankr:
Hello @jodyphelan 

Thanks for noticing the issue with unchanged build number.
I increased the build number to one and make sure the URL/sha256 are accurate.

Would really appreciate your assistance since this is time sensitive.

Thanks,
Mehran

jodyphelan:
Hi @mehrankr,

I've had a look and using the current source it seems like the sha256 changes for each new download:
image
This causes the bioconda CI pipeline to fail.

I'm not sure why exactly this is happening, but I know the usual way for software to feed into bioconda is to create a github release. If you do this then you can update the source url with one that has a stable sha256

mehrankr:
Thanks a lot, will fix and open a new request

PR #34662, opened 2022-05-02
Added expam package

seansolari:
Add expam package to bioconda.

Expam is a Python-based package for phylogenetic analysis of metagenomes.

### DIFF
--- a/recipes/expam/meta.yaml
+++ b/recipes/expam/meta.yaml
@@ -1,0 +1,50 @@
+{% set name = "expam" %}
+{% set version = "1.0.5" %}
+
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  url: "https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz"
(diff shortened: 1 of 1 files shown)


seansolari:
@BiocondaBot please add label

recipes/expam/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - numpy ==1.22.2
+    - setuptools
+  run:
+    - python
+    - numpy ==1.22.2
+    - matplotlib-base ==3.5.1
+    - pandas ==1.4.1
+    - psutil ==5.9.0
+    - requests ==2.27.1
+    - pytables ==3.7.0
```
dpryan79, line 33: Does it actually require these exact versions? That will make it almost immediately incompatible with other packages.

seansolari: Thanks for that @dpryan79, I modified the requirements and reduced it down to something less restricting. The new version seems to pass all the tests (c6c15d1 below).

dpryan79 (approved):

dpryan79:
@BiocondaBot please merge

PR #34695, opened 2022-05-03
Update virchip

mehrankr:
URL changed from deprecated bitbucket repo to GitHub releases and updated the sha256sum
Increased build number from 0 to 1


mehrankr:
@jodyphelan Thanks for your previous suggestions on bumping up the build number and using GitHub releases. Would appreciate if you could take another look.

PR #34984, opened 2022-05-20
Update SnpEff java path

RomainFeron:
The recipe for SnpEff contains a script used to run it as executable. This script was retrieving the value of JAVA_HOME to build the path to the java executable, but this environment variable was not set, so the system's java executable was used instead of the environment one. Since openjdk is included in the environment with the correct version, I changed the script to point directly to 'java'. If there was a better way to fix that, please tell me!

### DIFF
--- a/recipes/snpeff/meta.yaml
+++ b/recipes/snpeff/meta.yaml
@@ -14,7 +14,7 @@ package:
   version: {{ version }}
 
 build:
-  number: 2
+  number: 3
   noarch: generic
 
 source:
--- a/recipes/snpeff/snpeff.py
+++ b/recipes/snpeff/snpeff.py
@@ -29,13 +29,18 @@ def real_dirname(path):
 
 def java_executable():
     """Return the executable name of the Java interpreter."""
-    java_home = getenv('JAVA_HOME')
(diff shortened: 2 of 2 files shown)


RomainFeron:
@BiocondaBot please  add label

cbrueffer:
Could you elaborate on how the system Java ends up getting used? Looking at the code, JAVA_HOME only ends up getting used when it's set, which it's not in your case as you say. So in your case it ends up in the simple `java` anyway, so why remove the JAVA_HOME case?

I'm also a bit confused that JAVA_HOME isn't set; for me setting JAVA_HOME happens when activating an environment with OpenJDK installed. Are you sure the snpEff environment is activated?

RomainFeron:
After your comment I double checked and you are right, JAVA_HOME is set after activating the environment. It did not work for me the first time because I installed snpEff after activating the environment (which was then empty) and JAVA_HOME was set to the wrong value; after deactivating and activating the environment again, it's set to the correct value. Sorry for this oversight, there's no need for the PR then

cbrueffer:
Glad it's resolved, thanks for confirming!

PR #35008, opened 2022-05-23
Update msmetaenhancer: Remove dependency boundaries

xtrojak:
Simplified dependencies of `MSMetaEnhancer`.

### DIFF
--- a/recipes/msmetaenhancer/meta.yaml
+++ b/recipes/msmetaenhancer/meta.yaml
@@ -16,10 +16,10 @@ build:
 
 requirements:
   host:
-    - python
+    - python >=3.8
     - pip
   run:
-    - python
+    - python >=3.8
     - matchms
     - pandas
     - scipy
--- a/recipes/msmetaenhancer/meta.yaml
+++ b/recipes/msmetaenhancer/meta.yaml
@@ -10,17 +10,17 @@ source:
   sha256: 63a6995cc9e58efa2f902c8a9cc415ae48324810f440a5a6527f86363d659c5a
 
 build:
-  number: 0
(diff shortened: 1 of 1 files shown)


xtrojak:
@BiocondaBot please add label

bgruening:
Can you explain why this is needed, or better do you know why this was added in the first place?

xtrojak:
> Can you explain why this is needed, or better do you know why this was added in the first place?

@bgruening That is an excellent question! In the early stages of the package development, we decided to limit allowed python versions. However, it became unclear why this was needed (I think some issues with `matchms` requirement), so we decided to mitigate these restrictions. Also, it seems to be causing extensive memory consumption during the installation process, which we want to [eliminate](https://github.com/RECETOX/MSMetaEnhancer/issues/106).

martenson:
I suspect making the version req something like `python >=3.9` will limit the search space for dependencies and help the performance of `conda create/install` commands. We probably do not want to support Python 3.6 and such anyways.

bgruening:
Does msmetaenhancer-0.2.3-pyh5e36f6f_1.tar.bz2 work with all python version you would like to support?

bgruening (approved):

xtrojak:
I actually double-checked it locally, and it seems like there are some issues due to some matchms dependencies with python under `3.8`, so I reintroduced the lower bound and changed it to `>=3.8`.

bgruening:
@BiocondaBot please merge

thanks for checking

xtrojak:
@bgruening thank you for your time!

bgruening:
Your welcome!

PR #35185, opened 2022-06-02
Update gimmemotifs to 0.17.1

Maarten-vd-Sande:

----

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

<table>
  <tr>
  </tr>
</table>

</details>


siebrenf:
"No numpy version specified in conda_build_config.yaml.  Falling back to default numpy value of 1.16"

According to [the docs](https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html) that file reduces the amount of numpy builds being created. [It exists](https://github.com/bioconda/bioconda-recipes/blob/3c50a4e3089f8ac1c23af19f52bba00c18bd8a96/recipes/gimmemotifs/conda_build_config.yaml), and has numpy versions specified, but the version are duplicated, outdated, and I dont see addtional versions of numpy being installed in Azure.

Maybe we need to remove the duplicate versions, or we need to remove it completely (since we pin numpy in the recipe), or we need to specify the versions we currently do allow ( >=1.6).

I do however see multiple version of Python being installed in Azure. Perhaps we should probably set `host: python >=3.8` to match the current recipe, as the logs currently show we're building environments with python 3.6+.

siebrenf:
In the old meta.yaml, 
- `build: skip: True # [py2k]` skips the Python 2 build.
- `host: pybedtools` skips the Python 3.10 build.

Both of these were required as gimme does not support Python <3.6 (roughly), and pybedtools does not support >=3.10. This implementation is hacky however. 


Ideally, we'd specify the python versions manually with something like `host: python >=3.6, <3.10` (it has to be `host` section, as this influences the available python versions in the `run` section). This was causing:
```cp: cannot stat '/opt/conda/conda-bld/*/*.tar.bz2': No such file or directory```

An issue suggests this is because we cannot specify an upper bound.


Only pinning the lower bound `python >=3.6` causes:
```RuntimeError: Received dictionary as spec.  Note that pip requirements are not supported in conda-build meta.yaml.  Error message: Invalid version '3.10.0a0<3.10': invalid character(s)```

This probably happens because the build now installs Python 3.10, then tries to tell the `run` requirements the minimum python version is 3.10.0a.


The suggested method of pinning version ranges is though [the conda_build_config.yaml](https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html). However, the docs don't mention that a `zip_keys` rule already exists for Python, Python_imp and Numpy. Therefore, my first attempt was met with:
```
ValueError: Variant configuration errors in argument_variants:
  zip_key entry numpy in group frozenset({'numpy', 'python_impl', 'python'}) is a duplicate, keys can only occur in one group
  zip_key entry python in group frozenset({'numpy', 'python_impl', 'python'}) is a duplicate, keys can only occur in one group
```

Furthermore, the Python options are strictly specified, and the docs suggest options which are simply not allowed 😭 .
```
ValueError: variant config in /home/vsts/work/1/s/recipes/gimmemotifs/conda_build_config.yaml is ambiguous because it
does not fully implement all zipped keys (To be clear: missing ['numpy', 'python_impl'])
or specifies a subspace that is not fully implemented (To be clear:
.. we did not find ['3.6', '3.7', '3.8', '3.9'] from {'python': ['3.6', '3.7', '3.8', '3.9']} in python:['3.9.*  *_cpython', '3.8.*  *_cpython', '3.7.*  *_cpython', '3.6.*  *_cpython', '2.7.*  *_cpython']).
```

Buuuut I found a config example that does work [here](https://github.com/bioconda/bioconda-utils/blob/master/bioconda_utils/bioconda_utils-conda_build_config.yaml), so I'm now attempting the tests with the intersection of python versions both bioconda and gimme support.

siebrenf:
ok, this works. 
Now for the painful part: does it work with all 4 build versions active, and can we add back some tools?

siebrenf (approved):
@Maarten-vd-Sande (@simonvh) gimme is currently working with Python 3.8 and 3.9, without Trawler, Weeder and Yamda. These limitations are solely here so gimme fits in Azure's memory. 
I have no idea how it used to fit before, so I'd personally leave it at this. When the Azure container switches to Mamba, we can revisit this.

siebrenf:
Yes 😎 The container log shows the `gimme --help` output, with the expected config messages (all tools found, except the ones excluded above)

PR #35235, opened 2022-06-06
Update entrez-direct recipe to use extended-base

MillironX:
This should resolve issues with SSL certificates via openssl (which is one
of the use cases mentioned in
https://bioconda.github.io/contributor/troubleshooting.html#using-the-extended-image).

Fixes: https://github.com/BioContainers/containers/issues/441

### DIFF
--- a/recipes/entrez-direct/meta.yaml
+++ b/recipes/entrez-direct/meta.yaml
@@ -17,6 +17,8 @@ requirements:
   build:
     # The installer script downloads statically compiled binaries, so this is effectively just avoiding linter errors
     - {{ compiler('go') }}
+  run:
+    - wget
 
 test:
   commands:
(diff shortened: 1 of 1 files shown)


recipes/entrez-direct/meta.yaml
```diff
@@ -17,6 +17,8 @@ requirements:
   build:
     # The installer script downloads statically compiled binaries, so this is effectively just avoiding linter errors
     - {{ compiler('go') }}
+  run:
+    - wget
```
bgruening, line 21: Why is wget needed?

MillironX: All of the e-utils depend on either `wget` or `curl`. busybox includes `wget`, but `extended-base` doesn't, so this dependency was hidden previously.

MillironX:
@BiocondaBot please add label

bgruening (approved):

bgruening:
@BiocondaBot please merge

PR #35367, opened 2022-06-14
add  recipes for lmz eden-kernel dirtyopts and structout

smautner:
add  recipes for lmz eden-kernel dirtyopts and structout.


smautner:
@BiocondaBot please add label

daler:
@smautner none of these look like they are directly related to biomedical research, so [conda-forge](https://conda-forge.org/#add_recipe) would be the place to submit them.

smautner:
@daler thx i put it on forge :)

PR #35398, opened 2022-06-17
split gimmemotifs recipe

siebrenf:
`Gimmemotifs` contains several motif discovery tools which are (or depend on) some very large packages. 
These tools are required for _de novo_ motif discovery by `gimme motifs` and `gimme location`. Together they add around 150 packages, worth ~1.5 GB of space.

The Bioconda's Azure container is running out of memory. I do now know if this is caused by the solver, or by the package when it's loading into memory. In this PR, I'm attempting to split the load on the solver, by separating `gimmemotifs` into `gimmemotifs-minimal` and `gimmemotifs`, similar to `matplotlib` and `snakemake`. The added benefit is that packages using gimme can save a lot if _de novo_ motif discovery is not required (e.g. `ANANSE`).


siebrenf:
@BiocondaBot please merge

PR #35412, opened 2022-06-21
Add recipe for SCExecute and SCReadCounts 

edwardsnj:
Recipes to add SCExecute and SCReadCounts (python programs) to bioconda. 

These programs are used to analyze NGS single-cell reads.

### DIFF
--- a/recipes/alignoth/build.sh
+++ b/recipes/alignoth/build.sh
@@ -1,0 +1,11 @@
+#!/bin/bash -e
+
+# TODO: Remove the following export when pinning is updated and we use
+#       {{ compiler('rust') }} in the recipe.
+export \
+    CARGO_NET_GIT_FETCH_WITH_CLI=true \
+    CARGO_HOME="${BUILD_PREFIX}/.cargo"
+
+export BINDGEN_EXTRA_CLANG_ARGS="${CFLAGS} ${CPPFLAGS} ${LDFLAGS}"
+
+cargo install --no-track --verbose --root "${PREFIX}" --path .
--- a/recipes/alignoth/meta.yaml
+++ b/recipes/alignoth/meta.yaml
@@ -1,0 +1,43 @@
+{% set version = "0.2.0" %}
+
+package:
+  name: alignoth
+  version: {{ version }}
+
+build:
+  number: 0
+  # TODO fails to build on osx with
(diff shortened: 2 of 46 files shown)


daler:
@edwardsnj looks like you're the author. These packages look rather large. Any way to exclude the data either from the releases or by removing the data directories during conda package creation?

edwardsnj:
Sure, I’ll remove the demo data from the build recipe in the morning. The other thing which is making it large is the use of wxpython, which pulls in a lot of other stuff. Wxpython is not needed for command-line use - which is a more common conda use case. Do you have a position on that?

daler:
re: wxpython, the scexecute and screadcounts packages (as hosted on anaconda) don't contain dependencies. Dependencies are only installed on the users machine at install time. The data, on the other hand, was included in the package and therefore consuming hosting space.

Also, looking more closely at the original code, there's a lot going on here -- have you tested that the conda package correctly runs the snakefile?

edwardsnj:
The snakefile is part of the data-demo directory, which I've now deleted from the package. Yes, I hope to *use* these conda packages with snakemake ultimately (but not the other way around).

daler:
Ah. Have you considered separating these and turning them into proper Python packages? Would make updating and maintaining much easier over the long term.

edwardsnj:
> 



> Ah. Have you considered separating these and turning them into proper Python packages? Would make updating and maintaining much easier over the long term.

Do you mean the snakemake workflow (yes, that is the next step, into its own repository), or the two programs themselves? The short answer is that the projects grew up organically and are distributed as binary packages as well as Python source. Whether the projects could be re-structured into a number of python packages is another question, but its not on my critical path right now. 

Is the size of the built packages more reasonable now?

daler:
Yep, I meant making them [Python packages](https://packaging.python.org/en/latest/), likely with scripts provided as [entry points](https://packaging.python.org/en/latest/specifications/entry-points/).

Understandable if that's not a priority, but it makes it easier for people to use and simplifies the maintenance burden in bioconda since the files would be in standard locations and easier to understand. For example, the unconventional build script and any possible interaction of the wrapper script and `LD_LIBRARY_PATH` within conda environments.

I *think* it should be good, but are you able to locally test the artifacts above with your test data?

Size looks good now.

edwardsnj:
I have just checked that these run correctly in docker containers built from the artifact images. No problems!

daler:
@BiocondaBot please merge

PR #35636, opened 2022-07-03
hapROH recipe

hringbauer:
`hapROH `is a package to analyze human aDNA data.

It is an installable pip package, and this here is now the first recipe for bioconda. 

The recipe is generated with `grayskull`.

I had to disable a LinCheck (`cython_must_be_in_host`), as `cython `cannot be both in `run:` and `host:` otherwise this check fails. Is there a deeper underlying issue? That seems like buggy behavior.


bgruening (approved):

bgruening:
@bioconda-bot please merge

PR #35647, opened 2022-07-04
Update to bedops 2.4.40 and reinstate macOS build

jmarshall:
As described in this PR's _omit-bad-mutable.patch_, bedops contains code that fails to compile with recent versions of Clang. Fortunately none of the problematic code is actually used, so the bioconda build can work around the problem simply by removing it.

Build failure analysis reported upstream as bedops/bedops#267.

----

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

<table>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -51,9 +51,6 @@ recipes/bcftools/1.3.1
 # fitBeta2.cpp:158:93: error: no matching function for call to 'arma::eGlue<arma::Glue<arma::Glue<arma::Op<arma::Mat<double>, arma::op_htrans>, arma::Mat<double>, arma::glue_times>, arma::Mat<double>, arma::glue_times>, arma::Mat<double>, arma::eglue_plus>::i(bool) const'
 recipes/xtail
 
-# unknown c++ error
-recipes/bedops
-
 # needs patching for compilers
 recipes/logol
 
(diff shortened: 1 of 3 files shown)


jmarshall:

jmarshall:
@BiocondaBot please add label

daler:
@jmarshall thanks for figuring this out. Is it worth waiting for upstream to fix this? Or can you confirm that the upstream unit tests are successful after this change?

jmarshall:
Upstream have now applied an appropriate fix, but there is no telling what their release plans are.

Upstream unit test results are the same with this patch as they are with the existing Conda 2.4.39 package and with pristine upstream 2.4.40 (as compiled with the — older — system compiler) and with the upstream development branch (v2p4p41). Some unit tests do not have their expected results hence do not succeed, but this is apparently a known as yet unaddressed issue; see eg bedops/bedops#257 ¯\\\_(ツ)\_/¯

This patch very intentionally only **removes** apparently unused code (rather than changing the functions in question to be non-`const` as the upstream fix does), precisely so that it would be implausible for it to change the behaviour of the program. Either the behaviour is unchanged, or it would fail to compile as the code was not in fact unused.

jmarshall:
Ping.

cbrueffer:
This looks good to me; @daler any other reservations from your end?

daler:
Thanks, and sorry for late reply

daler:
@BiocondaBot please merge

PR #33207, opened 2022-02-16
Add recipe for Neofox python package

priesgo:
This recipe adds the Neofox package for annotation of neoantigens to bioconda.

PR #36235, opened 2022-08-04
Neofox move (second try)

priesgo:
This is a follow PR to #36199

Here the neofox version is updated to 0.6.3 and the requirements versions are using in most cases compatible versions to avoid the issue of some bugfix versions available in pypi but not in conda-forge.

Originally, this was fixing the mess of the neofox recipe folder being in the root folder which was done here #33207.


priesgo:
Found some further issues with pickle-mixin which was not being used, so I removed it altogether + upgraded pysam version + fixed biopython version as we use the alphabet that was removed from later versions (it seems the python `~=` compatibility is broken there)

cbrueffer:
I'm curious about the requirements; https://github.com/tron-bioinformatics/neofox lists e.g. R and BLAST as additional requirements. Are these optional and the package will do something useful without these? I realize these can't be expressed as requirements in pypi, but it's certainly possible here if that's useful.

priesgo:
That is a good point @cbrueffer NeoFox requirements are not simple. 

There are a couple of non optional requirements which have a non commercial use license, these are netMHCpan (https://services.healthtech.dtu.dk/service.php?NetMHCpan-4.1) and netMHC2pan (https://services.healthtech.dtu.dk/service.php?NetMHCIIpan-4.0). To work around that we configured those dependencies via environment variables as explained here https://neofox.readthedocs.io/en/latest/02_installation.html#install-third-party-dependencies.

Once we were in that uncomfortable situation we managed blast and R in the same way, but you are totally right we could easily look for blast and R in the environment variable and if not there default to the path. And add those dependencies in conda. Would you point me to an example?

And regarding your question, no, blast and R are not optional.

priesgo:
I found how to add R and blast dependencies, I also added some R libraries that are also dependencies.

In a subsequent release of NeoFox on which we had been working for a while we will implement the defaulting to the path. I hope it looks better now.

cbrueffer:
Thanks for the clarification and changes. If you want you could also point to the other optional, restrictively licensed, packages in the `notes` section section, see https://github.com/bioconda/bioconda-recipes/blob/master/recipes/astral-tree/meta.yaml for an example. That will be shown upon package installation if I recall correctly.

priesgo:
Thanks for the feedback, this is very helpful. I just added some information pointing to the relevant documentation

priesgo:
@BiocondaBot please add label

cbrueffer (approved):

cbrueffer:
@bioconda-bot please merge

priesgo:
yay! Thanks all!

PR #36412, opened 2022-08-12
Add Biocode recipe

gtonkinhill:
Add Biocode python package recipe. Includes a collection of useful bioinformatics scripts. A patch has been added to remove a script that is not written in python v3.


andersgs:
Could the script be converted to py3 upstream?

recipes/biocode/meta.yaml
```diff
@@ -0,0 +1,141 @@
...
+  sha256: ab48e0ea92f52c7fe48c6c947d4d5aa43437f16cd42a804546c71ef0e5ba2265
+  patches: biocode.patch
+
+build:
+  noarch: python
+  number: 0
+  script: "{{ PYTHON }} -m pip install . --no-deps -vv"
+  
+requirements:
+  host:
```
marcelm, line 19: The host section should only contain those requirements needed to run `pip install`, often these are only `pip` and `python`.

gtonkinhill:
@andersgs thanks very much for reviewing! I think that particular script is problematic as it doesn't handle FASTA files that have entries that extend over multiple lines. It would probably be best fixed with a pull request to the main package which I will try and find time for soon. I was hoping to get the rest of the package up on bioconda in the meantime.

@marcelm thanks! I have removed the unnecessary host requirements.

andersgs (approved):

andersgs:
@gtonkinhill ok. Please do fix it upstream. It is better than patching here.

marcelm:
@andersgs Not sure if this has changed recently, but at least a couple of months ago, the preferred way to merge PRs was to issue the 'BiocondaBot please merge` command because it avoids re-building the recipe.

andersgs:
I completely forgot about that! Thank you @marcelm

PR #36600, opened 2022-08-23
Update rust-bio-tools to 0.41

FelixMoelder:
Try building rust-bio-tools for macOS 10.14

### DIFF
--- a/recipes/rust-bio-tools/build.sh
+++ b/recipes/rust-bio-tools/build.sh
@@ -1,5 +1,11 @@
 #!/bin/bash -e
 
+
+1	#!/bin/bash
+2	if [ `uname` == Darwin ]; then
+3		export MACOSX_DEPLOYMENT_TARGET=10.14
+4	fi
+
 # TODO: Remove the following export when pinning is updated and we use
 #       {{ compiler('rust') }} in the recipe.
 export \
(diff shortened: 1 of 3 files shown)


FelixMoelder:
This reenables building for macOS.
It has been previously disabled as `rocksdb` could not be build for osx <10.14.
With 10.14 as target it works again.

recipes/rust-bio-tools/build.sh
```diff
@@ -1,5 +1,9 @@
 #!/bin/bash -e
 
+if [ `uname` == Darwin ]; then
+    export MACOSX_DEPLOYMENT_TARGET=10.14
```
johanneskoester, line 4: @dpryan79 @daler @epruesse @bgruening do we want to target later OSXs? I.e. do we guarantee a minimum compatible OSX version somewhere which we would break this way?

dpryan79: The CI infrastructure sets its own OSX target version. That was copied from conda-forge, but perhaps they're using a more recent target these days. I know at least previously people used to update OSX pretty religiously, but I don't know how accurate that is these days. If the 10.12 (or whatever we target by default) footprint is miniscule I would definitely support changing to something more recent.

PR #36699, opened 2022-08-30
macOs debug [WIP]

daler:
The bulk branch is still problematic for compiling Mac packages. The same packages seem to work fine locally on my Mac laptop. This PR is intended to be a testing ground for fixing those errors on the CI infrastructure.


jmarshall:
Does the ✔️ on the most recent commit indicate any resolution to the **bulk** branch issues?

I see there has been recent activity in bioconda-utils, and I still hope that bioconda/bioconda-utils#797 will make it into any upcoming bioconda-utils release. (Bioconda cannot update to samtools et al 1.16+ without this or a related change, which is starting to be embarrassing.)

daler:
@jmarshall yep, spending the long holiday weekend on this...osx is now working in bulk. The checkmark meant it was working here, but ended up being stale caches in GitHub Actions that was causing it to fail on bulk. That is now running and will be ramping up the workers to build as fast as possible. Once that settles down then we can update the pinnings and start a new round on bulk to rebuild.

PR #36848, opened 2022-09-07
Add recipe for HAPO-G 1.3.2

bistace:
Hello,

I've added the recipe for the HAPO-G tool [link](https://github.com/institut-de-genomique/HAPO-G/tree/1.3.2) the list of available recipes by creating the relevant folder (`hapog`) and conda build files.

### DIFF
--- a/recipes/hapog/build.sh
+++ b/recipes/hapog/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+
+export C_INCLUDE_PATH=${PREFIX}/include
+export LIBRARY_PATH=${PREFIX}/lib
+echo $PREFIX
+
+python setup.py install --record record.txt
+bash build.sh -l $LIBRARY_PATH
+cp -r hapog_build/hapog ${PREFIX}/bin/hapog_bin
--- a/recipes/hapog/meta.yaml
+++ b/recipes/hapog/meta.yaml
@@ -1,0 +1,38 @@
+package:
+  name: hapog
+  version: 1.3.2
+
+build:
+  number: 0
+
+source:
+  url: https://github.com/institut-de-genomique/HAPO-G/archive/refs/tags/1.3.2.tar.gz
+  sha256: "67d6c055ba2e7c26520ed1e8c41676b7651eea8459da6c43999c3f31cfb9c484"
+
+requirements:
+  build:
+    - gcc
+    - cmake
+    - make
+    - htslib
+  host:
+    - 'python>=3.7'
+    - htslib
+  run:
+    - htslib
+    - 'python>=3.7'
+    - biopython
+    - bwa
+    - samtools
+    - minimap2
+
+test:
+  commands:
+    - hapog -h
+    - hapog_bin -h
+
+about:
+  home: https://github.com/institut-de-genomique/HAPO-G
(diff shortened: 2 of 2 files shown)


recipes/hapog/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - minimap2
+
+test:
+  commands:
+    - hapog -h
+    - hapog_bin -h
+
+about:
+  home: https://github.com/institut-de-genomique/HAPO-G
+  license: OTHER
```
pvanheus, line 39: This should be `license: CECILL-2.1` and `license_family: CECILL`

recipes/hapog/build.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+export C_INCLUDE_PATH=${PREFIX}/include
+export LIBRARY_PATH=${PREFIX}/lib
+
+python setup.py install --single-version-externally-managed --record=record.txt
```
pvanheus, line 6: Looks like Python modules might be installed as well as the binary, if so they should be tested with an [import test](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#python-imports) in the test section.

recipes/hapog/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - 'python>=3.7'
+    - setuptools
+    - biopython
+    - bwa
+    - samtools
+    - minimap2
+
+test:
+  commands:
+    - hapog -h
```
pvanheus, line 34: Which is the command to run? The README of the [upstream tool](https://github.com/institut-de-genomique/HAPO-G#standard-pipeline) talks about running a `hapog.py` - does `hapog` here correspond to that Python file? And is `hapog_bin` a compiled module? Would the user ever run `hapog_bin` directly?

recipes/hapog/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  url: https://github.com/institut-de-genomique/HAPO-G/archive/refs/tags/{{ version }}.tar.gz
+  sha256: "46e0d102467276d0aee5c493e5e678c76120760f0dbe57caf4d880d366dcb285"
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - cmake
+    - make
+  host:
+    - 'python>=3.7'
```
pvanheus, line 20: Quotes are not necessary.

pvanheus (changes requested):

recipes/hapog/meta.yaml
bistace: The command to run is `hapog` as it is a wrapper around a C binary that was called `hapog_bin` here. In most cases, the user won't have to run this manually but in the case they need to, we would prefer if the binary was easily accessible.

pvanheus (approved):

pvanheus:
I see that `CECILL` as a `license_family` is not allowed. I'll raise this issue upstream - for now change `license_family` to `OTHER` and the linting will pass.

bistace:
@pvanheus Thanks a lot for reviewing this.

pvanheus:
@BiocondaBot please merge

PR #37003, opened 2022-09-16
Update Hapo-G to 1.3.3

bistace:
Hello,

this is an update to the Hapo-G recipe which fixes a bug. This bump in version is necessary for the tool to be integrated into Galaxy as it fixes a bug when launched with 1 thread.


bistace:
The update has been merged in another PR, closing this one.

PR #37164, opened 2022-09-27
Update smallgenomeutilities to version 0.3.8

mcarrara-bioinfo:
Update smallgenomeutilities to version 0.3.8

multiple bug fixes on the frameshift reporting script
Increased portability


mcarrara-bioinfo:
@BiocondaBot please add label

recipes/smallgenomeutilities/0.3.8/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+  sha256: {{ sha256 }}
+
+build:
+  noarch: python
+  number: 0
+  script: {{ PYTHON }} -m pip install . --ignore-installed --no-deps -v
+
+requirements:
+  host:
+    - python >=3.6
```
siebrenf, line 21: If you look at the Azure tests you can see your package is only built for python 3.10. You can build it for python 3.7, 3.8, 3.9 and 3.10 with these changes:

```
build:
  skip: True  # [py<37]

requirements:
  host:
    - python
  run:
    - python
```

by not specifying the python version in the requirements, all versions are build. Using the build skip, you then exclude the versions that don't work for you (with the comment).

siebrenf (approved):
Your recipe is fine, but why are you making a new folder for every version? (You might have a reason I don't know, just asking. The feedback is optional)

You could place this `meta.yaml` in `bioconda-recipes/recipes/smallgenomeutilities/` and you'll only have to update the SHA and version. The bioconda-bot should even do this for you automatically from then on! Also, the older versions of your package will still be available.

DrYak:
> why are you making a new folder for every version?

Until very recently there where multiple versions of [V-pipe](https://github.com/cbg-ethz/V-pipe) in production in parallel (e.g.: SARS-CoV-2 processing initially lived as a separate fork), each relying on a separate version of smallgenomeutilities (with potential need for patches and back-ported fixes), thus multiple parallel versions of the package.

By now we have merged the versions into a single common master branch and rely on virus base configs, so indeed in the near future we will progressively weed out the multiple versions. (And if anyone want to recreate the older environment, the historical older package of bioconda should be indeed fine as you suggest).

recipes/smallgenomeutilities/0.3.8/meta.yaml
DrYak: Ooops, we now get the error:
> ERROR: recipes/smallgenomeutilities/0.3.8/meta.yaml:16: should_not_be_noarch_skip: The recipe uses `skip: True` but is marked noarch

DrYak: From the [bioconda documentation](https://bioconda.github.io/contributor/linting.html#noarch-or-not-noarch):

> A noarch package cannot use skip. Packages that are noarch but require specific Python versions should use pinning (e.g. - python >3).

I think we should revert 0bf6bf5.

siebrenf: My bad, sorry! I was reading from the [conda](https://docs.conda.io/projects/conda/en/latest/user-guide/concepts/packages.html#noarch-python) docs, instead of the [bioconda](https://bioconda.github.io/developer/_autosummary/bioconda_utils.lint.check_noarch.html#bioconda_utils.lint.check_noarch.should_not_be_noarch_skip) docs...

DrYak:
@bioconda-bot please merge

PR #37358, opened 2022-10-08
Exclude hesslab-gambit from future builds

jlumpe:
This tool/recipe was renamed to `gambit`, the last version was supposed to be `0.5.1`. I think this is the correct change to achieve that outcome (was made in #36569 but closed for some reason).


jlumpe:
@rpetit3

rpetit3:
@jlumpe lets actually just remove the folder for hesslab-gambit

jlumpe:
Thanks Robert, new PR is #37438.

PR #37568, opened 2022-10-21
Update phyloacc

gwct:
This updates the phyloacc build script to add a post-processing script (phyloacc_post.py) to the environment path.

----

### DIFF
--- a/recipes/phyloacc/build.sh
+++ b/recipes/phyloacc/build.sh
@@ -2,5 +2,6 @@
 
 make install CXX=${CXX}
 cp src/PhyloAcc-interface/phyloacc.py ${PREFIX}/bin/.
+cp src/PhyloAcc-interface/phyloacc_post.py ${PREFIX}/bin/.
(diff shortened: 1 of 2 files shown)


gwct:
@BiocondaBot please add label

mfansler:
@gwct it's preferable to test all entry points. Is there a reasonable test that could be run against this additional script? e.g., would it accept a `--help` argument?

gwct:
> @gwct it's preferable to test all entry points. Is there a reasonable test that could be run against this additional script? e.g., would it accept a `--help` argument?

@mfansler sure, I added the test. I was/am wary because the source might need to be updated for the test to pass and I wanted to update the recipe first. But maybe I should've done it the other way around. Guess we'll see.

dpryan79:

dpryan79 (approved):

dpryan79:
@BiocondaBot please merge

PR #37619, opened 2022-10-25
Update phyloacc to 2.1.0

gwct:
For some reason this didn't autobump when I updated the source, so I'm doing it manually.

----


gwct:
I had a typo in the version tag which is why it didn't autobump... closing.

PR #37931, opened 2022-11-11
chore: use mergify to automatically merge autobump PRs from BiocondaBot after waiting 3 days

johanneskoester:
Still to solve: we need to ensure that the autobump PR does not contain additional commits from others than the BiocondaBot.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -5,6 +5,7 @@ pull_request_rules:
       - "label=autobump"
       - "created-at>=3 days ago"
       - "check-success=bioconda.bioconda-recipes"
+      - "#commits=1"
     actions:
       review:
         type: APPROVE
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -0,0 +1,13 @@
+pull_request_rules:
+  - name: Approve and merge autobump PRs if tests are successful
+    conditions:
+      - "author=BiocondaBot"
+      - "label=autobump"
+      - "created-at>=3 days ago"
+      - "check-success=bioconda.bioconda-recipes"
```
mirand863, line 7: Perhaps just checking if there is a single commit would suffice?

```suggestion
      - "check-success=bioconda.bioconda-recipes"
      - "#commits=1"
```

mirand863 (changes requested):

mbhall88:
The 3-day wait doesn't seem to be happening. I had a PR https://github.com/bioconda/bioconda-recipes/pull/40231 get merged almost immediately after passing. This can be a little annoying if I needed to make changes (i.e. add new dependencies etc.)

PR #38042, opened 2022-11-18
Fix dependency issue between perl and bioperl

simroux:
Fixing dependency issue between perl and bioperl that prevented the recipe to be installed as of Nov 18 2022. See issue https://bitbucket.org/srouxjgi/iphop/issues/17/unsatisfiableerror-the-following


apcamargo (approved):

apcamargo:
@BiocondaBot please merge

PR #38565, opened 2022-12-27
Add vcfbub

AndreaGuarracino:
**Overview**
The VCF output produced by a command like `vg deconstruct -e -a -H '#' ...` includes information about the nesting of variants. With `-a, --all-snarls`, we obtain not just the top-level bubbles, but all nested ones. This exposed snarl tree information can be used to filter the VCF to obtain a set of non-overlapping sites (n.b. "snarl" is a generic model of graph bubbles including tips and loops).

`vcfbub` accomplishes a simple task: we keep sites that are the children of those which we "pop" due to their size. These occur around complex large SVs, such as multi-Mbp inversions and segmental duplications. We often need to remove these, as they provide little information for many downstream applications, such as haplotype panels or other imputation references.

### DIFF
--- a/recipes/vcfbub/build.sh
+++ b/recipes/vcfbub/build.sh
@@ -1,0 +1,6 @@
+#!/bin/bash -euo
+
+RUST_BACKTRACE=1 CARGO_HOME="${BUILD_PREFIX}/.cargo" cargo build --release
+
+mkdir -p $PREFIX/bin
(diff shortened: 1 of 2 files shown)


AndreaGuarracino:
@BiocondaBot please add label

rpetit3 (approved):

rpetit3:
@BiocondaBot please merge

PR #38612, opened 2023-01-02
deeptools np.float is not supported anymore

Maarten-vd-Sande:
Deeptools makes use of np.float and np.int, but those are depcrecated since 1.20 and not supported since 1.24 anymore: https://numpy.org/doc/stable/release/1.24.0-notes.html#expired-deprecations


siebrenf:
@BiocondaBot please merge

PR #38683, opened 2023-01-06
Update spec2vec to 0.8.0

BiocondaBot:
Update [`spec2vec`](https://bioconda.github.io/recipes/spec2vec/README.html): **0.6.0** &rarr; **0.8.0**

[install with bioconda](http://bioconda.github.io/recipes/spec2vec/README.html) [Conda](https://anaconda.org/bioconda/spec2vec/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/spec2vec`](https://github.com//bioconda/bioconda-recipes/tree/bump/spec2vec/recipes/spec2vec) (click to view/edit other files)
Summary | Word2Vec based similarity measure of mass spectrometry data.
Home | [https://github.com/iomega/spec2vec](https://github.com/iomega/spec2vec)
Releases |[https://pypi.org/pypi/spec2vec/json](https://pypi.org/pypi/spec2vec/json)
Recipe Maintainer(s) | @bgruening, @hechth

***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

PR #38716, opened 2023-01-09
Add missing dependency (missing from gensim 4.3.0)

florian-huber:
Seems like `FuzzyTM` is part of gensim's dependencies but still not installed (maybe missing from the conda recipe?).
So I added it to the spec2vec dependencies to pass the tests.


bgruening:
Florian if you need a new build you need to bump the build number.

florian-huber:
@bgruening I thought that because https://github.com/bioconda/bioconda-recipes/pull/38683 wasn't merged the build for 0.7.0 would still be `0`?
(if not I will of course bump the number)

hechth:
@florian-huber I'm working on getting the new gensim dependencies into conda (https://github.com/conda-forge/staged-recipes/pull/21712) and (https://github.com/conda-forge/staged-recipes/pull/21711) + 3 more packages until we have everything it seems for FuzzyTM. If spec2vec still works with the lower versions that is good for now, but it would be good to have the option to update in the future anyway.

Maybe we have to bump the build number since only the PR got renamed?

bgruening:
@florian-huber but then you need to take the changes from https://github.com/bioconda/bioconda-recipes/pull/38683 over here.

bgruening:
See also https://github.com/conda-forge/staged-recipes/pull/21824

bgruening:
We need to wait for conda-forge and then we will start here again. Thanks @florian-huber

PR #38819, opened 2023-01-13
Update SPAdes 3.12.0

kciy:
Using SPAdes 3.12.0 with a Python version >3.9 leads to the error described in [this issue](https://github.com/ablab/spades/issues/873). Python versions 3.10 and above don't have the alias to the old collections class anymore (see [this issue](https://github.com/python/cpython/issues/81505)).

Setting the python version to below 3.10 fixes the problem for SPAdes 3.12.0 (it is anyways fixed in SPAdes >=3.15.4).


wm75 (approved):
Careful work there @kciy. Thanks for the good explanation and helpful links!

bgruening:
@bioconda-bot please merge

PR #38875, opened 2023-01-17
pggb: add the script folder in bin and the igraph dep

AndreaGuarracino:


AndreaGuarracino:
@BiocondaBot please add label

recipes/pggb/meta.yaml
```diff
@@ -29,6 +29,7 @@ requirements:
     - time
     - vg 1.40.0
     - wfmash >=0.10.2
+    - python-igraph ==0.10.2
```
bgruening, line 32: Is this strict pin really necessary?

AndreaGuarracino: Not really. I put it to avoid possible problems with future versions of igraph in case something changes in it. That is the version that we mainly tested. Would you suggest being relaxed and use '>='?

szhorvat: You can check the changelog to see what changed. If you don't use experimental function (which are marked clearly in the docs), all behavioural changes should be the result of bugfixes, i.e. you really shouldn't be relying on the previous version's behaviour.

https://github.com/igraph/igraph/blob/master/CHANGELOG.md#0103---2022-12-30
https://github.com/igraph/python-igraph/blob/master/CHANGELOG.md#0103---2022-12-31

Latest version is 0.10.3, not 0.10.2.

AndreaGuarracino: We don't use experimental functions, so we are good. I've updated the recipe's YAML accordingly, avoiding the constraint to a specific (old) version. Thank you both for the tips!

recipes/pggb/build.sh
```diff
@@ -6,6 +6,8 @@ n=$((n-1))
 sed -i ${n}'a\if [ $show_version == true ]; then echo "pggb '$PKG_VERSION'"; exit; fi' pggb
 
 mkdir -p $PREFIX/bin
+
 mv pggb $PREFIX/bin
 mv partition-before-pggb $PREFIX/bin
 mv scripts/*.py $PREFIX/bin
+mv scripts $PREFIX/bin
```
subwaystation, line 13: EOL @AndreaGuarracino. Then this should be good to go. Thanks!

AndreaGuarracino: Here you are

subwaystation:
@bgruening @szhorvat Could one of you please take a look? Thanks!

subwaystation:
@BiocondaBot please add label

subwaystation:
@BiocondaBot please review & merge

bgruening (approved):

bgruening:
@BiocondaBot please merge

bgruening:
Thanks @AndreaGuarracino

subwaystation:
Thank you very much @bgruening !

PR #39619, opened 2023-02-26
Update plink2 to 2.00 alpha 3.7

cpalmer718:
This is a (no longer simple) version bump from the existing alpha 3.3 build, with modifications:

- skeleton Windows support from existing recipe is removed
- OSX is still doing binary redistribution as in the existing recipe, but with v3.7
- Linux now builds from source and dynamically links against [mkl](https://www.intel.com/content/www/us/en/develop/documentation/get-started-with-mkl-for-dpcpp/top.html)

----

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

<table>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/recipes/plink2/build.sh
+++ b/recipes/plink2/build.sh
@@ -1,3 +1,3 @@
 # Install as plink2
 mkdir -p $PREFIX/bin
-cp plink2 $PREFIX/bin/plink2
+cp plink2* $PREFIX/bin/plink2
--- a/recipes/plink2/build.sh
+++ b/recipes/plink2/build.sh
@@ -1,14 +1,19 @@
-# Build plink2 from source
-# Apparently this source bundle contains both 1.9 and 2.0.
-# Per repo instructions, actual target Makefile is ./build_dynamic/Makefile.
-# Due to how the Makefile includes dependent Makefiles, cwd has to be the build directory.
-cd 2.0/build_dynamic
+# OSX source build is not currently supported due to a lack
+# of an available optimized linear algebra engine (mkl, aocl, atlas)
 
-PLINK2_INCLUDES="-I${SRC_DIR}/2.0/libdeflate -I${SRC_DIR}/2.0/zstd/lib -I${SRC_DIR}/2.0/simde"
-PLINK2_CFLAGS="-O2 -std=gnu99 -I${PREFIX}/include ${PLINK2_INCLUDES}"
-PLINK2_CXXFLAGS="-O2 -std=c++11 -I${PREFIX}/include ${PLINK2_INCLUDES}"
-PLINK2_LINKFLAGS="-L${PREFIX}/lib -lpthread -lm -lz"
-make -f Makefile CC="${CC}" CXX="${CXX}" CFLAGS="${PLINK2_CFLAGS}" CXXFLAGS="${PLINK2_CXXFLAGS}" LINKFLAGS="${PLINK2_LINKFLAGS}" DYNAMIC_MKL="1" MKLROOT="${PREFIX}/lib" MKL_IOMP5_DIR="${PREFIX}/lib" CPU_CHECK=""
+if [[ ! -f "plink2" ]] ; then
+    # Build plink2 from source
+    # Apparently this source bundle contains both 1.9 and 2.0.
+    # Per repo instructions, actual target Makefile is ./build_dynamic/Makefile.
+    # Due to how the Makefile includes dependent Makefiles, cwd has to be the build directory.
+    cd 2.0/build_dynamic
+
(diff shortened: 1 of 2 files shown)


daler:
Hello again :D

Bioconda doesn't support Windows, so the `[win]` lines won't ever be seen by the build infrastructure and can be removed. And I guess the `.exe` addition would also be moot. However, if you do want it available in Windows, then I think it would need to go in conda-forge instead of bioconda (that's where BioPython is hosted, for example, to support Windows users).

And, in general, source distributions are preferred; even [BLAST is compiled](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/blast/build.sh). That can often be a pretty big lift though, and whether or not a contributor gives up and falls back to binary redistribution is largely based on their pain threshold.

cpalmer718:
@daler Thank you so much for the comments. This sounds pretty much like what I was expecting. I've removed the Windows "support" entirely; it was only there because it was in the preexisting recipe, and I have no personal interest in using it on Windows. The package _for linux_ is now compiled from source with appropriate dependencies. For OSX, it's still redistributing the prebuilt binary. I'd be curious to hear your thoughts on this:

- plink2 requires a linear algebra implementation. it looks for mkl, aocl, or atlas, depending on configuration.
- for linux, mkl is available in anaconda and works out of the box quite effectively (honestly to my surprise). aocl doesn't seem to exist anywhere, and I know almost nothing about it; and atlas is unavailable. 
- for osx, mkl is available but fails in build _somewhere_ due to something in the build reporting that mkl is secretly not actually supported for osx. aocl and atlas are also not available.
- [atlas](https://github.com/math-atlas/math-atlas) is a weird one. it's designed to be very heavily optimized for a specific system, so I've struggled to see use cases for it in the past even on an HPC, let alone in a package channel like this. it's seemingly BSD-3 and I don't see why one couldn't package it on e.g. conda-forge, except (1) the CI build time might be unacceptably slow while it does its parameter grid search, and (2) the resulting optimization wouldn't actually be particularly optimal on any particular downstream CPU.

So in short: linux source build works, OSX source build seems out of reach for the time being, and I don't know if it's worth the particular effort to get it working. The recipe is now a hybrid, which is a bit of a weird fit but hey, it's a step in the direction, I guess. I'd be interested in your thoughts on any of this.

druvus (approved):

daler:
There have been some situations like this in the past (linux works, osx doesn't). Agreed this is in the right direction, and probably the best that can be done at this point.

Regarding optimization, we can't really provide optimized builds so in cases where performance is of the utmost importance, conda packages are not a good fit. 

Pretty sure that the `run:` here is redundant with `host:` (see [this part of conda docs](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#export-runtime-requirements)), does it work if you remove the run block entirely?

cpalmer718:
@daler So, amusingly, about 20 minutes before you posted this, I was sniffing around other PRs in this repo and saw a discussion about this, and tried removing the entire run block. The build fails due to it not selecting `mkl` in the run environment. Dropping `zlib` only works, as `libzlib` still ends up in the run dependencies. I'll mess around with it a bit locally with the docker build script to save some CI cycles. At the very least, since the plink2 executable in the OSX build is statically linked, it theoretically shouldn't need BLAS.

cpalmer718:
@daler As this was already tagged, I'll convert the PR to draft so it doesn't accidentally get sucked up by someone just passing through.

daler:
all sounds great, thanks!

cpalmer718:
@daler Ok, so, everything but `mkl` can be removed from `run`. I'll be entirely frank and say I don't actually understand `run_exports` in practice. I think that's partially because the docs themselves are imo a bit misleading. In any case, I tried moving `mkl` to `run_exports` locally and it worked fine with `conda mambabuild`, in that the resulting package correctly pulled in `mkl` when installed. I didn't run `bioconda-utils` locally this time because for some reason it's timing out with remotes during `host` environment solution. With 00b2d77, the CI build failed when the shared library wasn't pulled in at runtime. I find this discrepant behavior relative to local build quite alarming, honestly; I don't see why this should have happened. Maybe the CI build and my local `conda-build` are different enough that they have different behaviors.

In any case, `mkl` is just a standard `run` inclusion now, and the build passes. And BLAS is removed from the OSX package.

daler:
Yeah, others on the bioconda core team have found run exports to be confusing as well. With tests passing here, let's merge it and then maybe you can test locally.

daler:
@bioconda-bot please merge

cpalmer718:
Cool @daler thank you. This has been a very informative and helpful review, I really appreciate it.

PR #39954, opened 2023-03-16
Initial release of pastrami

sgupta1524:
Pastrami is a novel, scalable computational algorithm for rapid human ancestry estimation at population-, subcontinental- and continental-levels. Pastrami works on two key methodologies: exact haplotype matching and non-negative least square (NNLS) optimization.

### DIFF
--- a/recipes/pastrami/build.sh
+++ b/recipes/pastrami/build.sh
@@ -1,0 +1,3 @@
+#!/bin/sh
+
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
--- a/recipes/pastrami/build.sh
+++ b/recipes/pastrami/build.sh
@@ -1,7 +1,0 @@
-#!/bin/bash
-
-#chmod +x $SRC_DIR/pastrami.py
-$PYTHON setup.py install --single-version-externally-managed --record=record.txt
-$PYTHON -m pip install --no-deps .  
-PIP_NO_CACHE_DIR=True
-$PYTHON -m pip install --no-deps .
--- a/recipes/pastrami/build.sh
+++ b/recipes/pastrami/build.sh
@@ -1,4 +1,7 @@
-#!/bin/sh
+#!/bin/bash
 
-PIP_NO_CACHE_DIR=True
+#chmod +x $SRC_DIR/pastrami.py
 $PYTHON setup.py install --single-version-externally-managed --record=record.txt
+$PYTHON -m pip install --no-deps .  
+PIP_NO_CACHE_DIR=True
+$PYTHON -m pip install --no-deps .
--- a/recipes/pastrami/build.sh
+++ b/recipes/pastrami/build.sh
@@ -1,3 +1,4 @@
 #!/bin/sh
 
+PIP_NO_CACHE_DIR=True
 $PYTHON setup.py install --single-version-externally-managed --record=record.txt
--- a/recipes/pastrami/meta.yaml
+++ b/recipes/pastrami/meta.yaml
@@ -6,8 +6,8 @@ package:
   version: "{{ version }}"
 
 source:
-  url: "https://github.com/healthdisparities/pastrami/archive/refs/tags/v{{ version }}.tar.gz"
-  sha256: 949e07383d26b04a49a151b0f35737327f7b5b72218785561d6bbca56ac941a8
+  url: "https://github.com/healthdisparities/pastrami/archive/v{{ version }}.tar.gz"
+  sha256: 4e498a66cf38aef94b307463889d9cb2fdf664a3a6765db8fab254234dac2a31
 
 build:
   noarch: python
@@ -23,6 +23,7 @@ requirements:
     - plink
     - python
     - scipy
+    - bzip2
   run:
     - python >=3.8
     - numpy
@@ -31,6 +32,7 @@ requirements:
     - plink
     - python
     - scipy
+    - bzip2
 
 test:
   imports:
--- a/recipes/pastrami/meta.yaml
+++ b/recipes/pastrami/meta.yaml
@@ -11,124 +11,25 @@ source:
 
 build:
   noarch: python
-  number: '0'
-  string: py_0
+  number: 0
+  script: "{{ PYTHON }} -m pip install . --no-deps -vv"
 
 requirements:
   build:
     - python >=3.8
-    - plink2
     - setuptools
-    - _libgcc_mutex =0.1= conda_forge
-    - _openmp_mutex =4.5= 2_gnu
-    - brotlipy =0.7.0 =py38h0a891b7_1005
-    - bzip2 =1.0.8 =h7f98852_4
-    - ca-certificates =2022.12.7 =ha878542_0
-    - certifi =2022.12.7 =pyhd8ed1ab_0
-    - cffi =1.15.1 =py38h4a40e3a_3
-    - charset-normalizer =2.1.1=pyhd8ed1ab_0
-    - cryptography =39.0.2 =py38h3d167d9_0
-    - dill =0.3.6 =pyhd8ed1ab_1
-    - idna =3.4 =pyhd8ed1ab_0
-    - ld_impl_linux-64 =2.40 =h41732ed_0
-    - libblas =3.9.0 =16_linux64_openblas
-    - libcblas =3.9.0 =16_linux64_openblas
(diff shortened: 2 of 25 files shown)


sgupta1524:
@BiocondaBot please add label

sgupta1524:

sgupta1524:
@BiocondaBot please add label

sgupta1524:

recipes/pastrami/meta.yaml
```diff
@@ -0,0 +1,140 @@
...
+  url: "https://github.com/healthdisparities/pastrami/archive/v{{ version }}.tar.gz"
+  sha256: 4457ad6cda7a2384a1b22109caf3edbff191b3e0a8749f779525276424b4c55c
+
+build:
+  noarch: python
+  number: '0'
+  string: py_0
+
+requirements:
+  build:
```
marcelm, line 18: The `build:` section should only list the requirements needed to run the installation, in this case, this is only `pip` and `python`, maybe `setuptools`.

recipes/pastrami/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+#chmod +x $SRC_DIR/pastrami.py
+$PYTHON setup.py install --single-version-externally-managed --record=record.txt
```
marcelm, line 4: [Calling `setup.py` directly like this is deprecated.](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html)

recipes/pastrami/build.sh
```diff
@@ -0,0 +1,7 @@
...
+$PYTHON -m pip install --no-deps .  
+PIP_NO_CACHE_DIR=True
+$PYTHON -m pip install --no-deps .
```
marcelm, line 7: This runs the installation three times, which should not be necessary. I would suggest that you get rid of `build.sh` entirely and just put `script: "{{ PYTHON }} -m pip install . --no-deps -vv"` into the `build:` section of the recipe.

recipes/pastrami/meta.yaml
```diff
@@ -0,0 +1,140 @@
...
+  version: "{{ version }}"
+
+source:
+  url: "https://github.com/healthdisparities/pastrami/archive/v{{ version }}.tar.gz"
+  sha256: 4457ad6cda7a2384a1b22109caf3edbff191b3e0a8749f779525276424b4c55c
+
+build:
+  noarch: python
+  number: '0'
+  string: py_0
```
marcelm, line 15: Why do you set this? It should not be necessary.

recipes/pastrami/meta.yaml
```diff
@@ -0,0 +1,140 @@
...
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  url: "https://github.com/healthdisparities/pastrami/archive/v{{ version }}.tar.gz"
+  sha256: 4457ad6cda7a2384a1b22109caf3edbff191b3e0a8749f779525276424b4c55c
+
+build:
+  noarch: python
+  number: '0'
```
marcelm, line 14: This may work, but not quoting the build number would be the normal thing to do. (`number: 0`)

recipes/pastrami/meta.yaml
```diff
@@ -0,0 +1,140 @@
...
+    - scipy =1.10.1 =py38h10c12cc_0
+    - setuptools =67.5.1 =pyhd8ed1ab_0
+    - six =1.16.0 =pyh6c4a22f_0
+    - tk =8.6.12 =h27826a3_0
+    - typing-extensions =4.4.0 =hd8ed1ab_0
+    - typing_extensions =4.4.0 =pyha770c72_0
+    - urllib3 =1.26.14 =pyhd8ed1ab_0
+    - wheel =0.38.4 =pyhd8ed1ab_0
+    - xz =5.2.6 =h166bdaf_0
+  run:
```
marcelm, line 77: You should not restrict the dependencies like that if not absolutely necessary, and you should also only list the direct dependencise. Otherwise, it is highly likely that nothing else can be installed along this package into a Conda environment. Also, sometimes individual builds of packages get deleted from anaconda.org because they have a security issue or so, and if this happens, then your package would no longer be installable.

Normally, you should only specify minimum versions (if at all).

marcelm (commented):
Hi, thanks for the recipe! I left a couple of comments.

recipes/pastrami/meta.yaml
sgupta1524: done

recipes/pastrami/build.sh
sgupta1524: Thanks! I was not aware. Done!

recipes/pastrami/build.sh
sgupta1524: Done

recipes/pastrami/meta.yaml
sgupta1524: Done

recipes/pastrami/meta.yaml
sgupta1524: Done

recipes/pastrami/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - python >=3.8
+    - plink2
+    - pathos
+    - numpy
+    - pandas
+    - bzip2
+    - numpy
+    - pandas
+    - pathos
```
marcelm, line 31: These two are in here twice. Maybe put `python >=3.8` as the first one and then sort the rest alphabetically?

recipes/pastrami/meta.yaml
sgupta1524: Done

recipes/pastrami/meta.yaml
sgupta1524: My bad, they shouldn't be repeated

marcelm (approved):
Looks good now!

I took the liberty of removing the `name` variable because it was used only once (and also there’s very low risk that the name changes in a future version of the recipe).

Just minor comments for future recipes/PRs:
- Don’t delete the PR template text with the instructions.
- In those instructions, it said that the PR title should start with the word "Add" if it adds a new recipe.

You can merge this yourself as soon as the tests have passed by issuing the "@ BiocondaBot please merge" command. (Remove the space after the @).

sgupta1524:
@marcelm Thanks for the review and extremely helpful comments! I will keep it in mind for future PRs. 
I had one question -- Do I need to create a new PR for a version bump of package? Or is there any other way to do it?

sgupta1524:
@BiocondaBot please merge

marcelm:
You’re welcome!

When you bump the package version, the BiocondaBot will notice this and automatically generate a version bump PR, see #39963 for an example. The bot runs (I believe) once an hour, so you may need to wait a little wait after you made your release. The bot does not detect when dependencies change, so if that is the case, you should comment on the bot-generated PR and point out which changes should be made to the recipe. Alternatively, you could also open your own PR, but then it would be good to link to the bot-generated PR just so that whoever does the review knows that there are two PRs doing the same thing.

Another alternative is to become a member of the Bioconda organization, then you can just edit the bot-generated PR.

sgupta1524:
@marcelm Thanks for this!
I am able to see my package on bioconda - https://anaconda.org/bioconda/pastrami
but conda install -c bioconda pastrami is still not working. Do you know how long it takes for bioconda to update? Additonally, if I download tar.bz2 file from https://anaconda.org/bioconda/pastrami/files and try to install - it doesn't work as well. If I tried to install using the merged meta.yaml only, the package was working as expected.

marcelm:
I have the same problem with two of my own packages that were merged a couple of hours ago. This is usally available almost immediately. Some other people mentioned the same problem in the [Bioconda chat](https://gitter.im/bioconda/Lobby#).

sgupta1524:
I see! Thanks for that. Hopefully it gets resolved soon then. Another problem is if you download the package from https://anaconda.org/bioconda/pastrami/0.9.5/download/noarch/pastrami-0.9.5-pyhdfd78af_0.tar.bz2 - ithe metal.yaml has the wrong python path in script section. I am wondering if the previous method of setup.py was correct?

marcelm:
Please have patience. The Python path is intended to be that way. Conda replaces it with the correct one at installation time.

PR #40071, opened 2023-03-25
fix builds

daler:
Most things seem to be working after [the key change](https://github.blog/2023-03-23-we-updated-our-rsa-ssh-host-key/) except for the `@bioconda-bot please update` functionality. This attempts to fix that...but that's not quite working, I think it needs to be fixed in the bot container(s).

However, separately and apparently unrelated, the lint step is failing because conda hangs when trying to install bioconda-utils. In the Azure Pipelines, this installation is currently implemented in [createEnv.py](https://github.com/bioconda/bioconda-recipes/blob/master/.azure-pipelines/createEnv.py), which uses conda.

Recently, the bulk branch has been using a unified installation script to install everything on GH Actions. This script uses mamba and forces strict channel priority, which seems to allow bioconda-utils to install.

So this PR refactors the Azure Pipelines CI to use this same unified script (though on the `bioconda-common/fix` branch, since installations into `$HOME` were not working as expected, as they were on GH Actions runners).

**NOTE:** `azure-pipeline-master.yml` and `azure-pipeline-nightly.yml` still need to be similarly ported.


daler:
The [install script](https://github.com/bioconda/bioconda-common/blob/fix/install-and-set-up-conda.sh) is currently installing miniconda from scratch into `/opt/miniconda`. Different providers (GH Actions, Azure Pipelines) seem to have already-installed versions of conda which we explicitly ignore in favor of being more explicit in our setup.

With the changes here, the **entire** `/opt/miniconda` dir is being cached. It's 2 GB and takes a couple minutes to restore.

It would be a smaller cache if we only used just the bioconda env. However, this requires modifying the install script to use the central conda installation and not install miniconda like we've been doing.

While I like how the separate miniconda installation is explicit, it does come with the higher cache size cost. I think I could go either way.

daler:
@bioconda-bot please update

abretaud:
Any progress on this? There are a few PR waiting, can we help?

PR #40299, opened 2023-04-06
pbhoover: add missing runtime dependency on htslib

0xaf1f:
Missing `bgzip` resulted in errors like
```
No such file or directory
```
at the normalization step following variant calling


0xaf1f:
@BiocondaBot please add label

bgruening (approved):

PR #40318, opened 2023-04-07
new recipe for qiskit-xyz2pdb

thepineapplepirate:
Added recipe for qiskit-xyz2pdb. This is a program used to convert XYZ files from Qiskit workflows, into proper PDB files.

### DIFF
--- a/recipes/qiskit-xyz2pdb/meta.yaml
+++ b/recipes/qiskit-xyz2pdb/meta.yaml
@@ -1,0 +1,47 @@
+{% set name = "qiskit-xyz2pdb" %}
+{% set version = "0.1.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: aabd4c37258d9d45db8c0cf89b19368497de6d7c15a514bb9f616b3f5871eab9
+
+build:
+  number: 0
+  noarch: python
+  entry_points:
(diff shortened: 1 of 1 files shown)


recipes/qiskit-xyz2pdb/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  imports:
+    - qiskit-xyz2pdb
+  commands:
+    - pip check
+    - qiskit-xyz2pdb --version
+  requires:
+    - pip
+
+about:
+  home: https://github.com/thepineapplepirate/qiskit-xyz2pdb
```
bgruening, line 37: That one does not exist?

recipes/qiskit-xyz2pdb/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  host:
+    - python >=3.6
+    - pip
+    - setuptools
+  run:
+    - python >=3.6
+
+test:
+  imports:
+    - qiskit-xyz2pdb
```
bgruening, line 29: This can not be the module name, isn't it?

recipes/qiskit-xyz2pdb/meta.yaml
thepineapplepirate: just made public

recipes/qiskit-xyz2pdb/meta.yaml
thepineapplepirate: just updated to correct name :).

bgruening (approved):

thepineapplepirate:
thank you @bgruening !!

PR #40693, opened 2023-04-28
Update MOSCA to 2.1.0 and remove it from blacklist

iquasere:
MOSCA is a snakemake pipeline, that was based on a complex conda environment, which could no longer be built. This put it on the blacklist.

MOSCA has now been updated to use snakemake's environment handling, therefore only requiring snakemake and python as dependencies. It can now be reintegrated into Bioconda.


iquasere:
@bioconda-bot please review & merge

iquasere:
@BiocondaBot please add label

dpryan79:

dpryan79:
As long as everyone is OK with this not being a docker container or working on instances where there's no internet connection then this should be OK.

dpryan79 (approved):

iquasere:
Thank you, @dpryan79 ! Your comment got me thinking, is the "working where there's no internet connection" due to the pipeline downloading loads of databases before being capable of running?

And the "not being a docker container", is it because of the singularity approach? If running a container with the ```-v /var/run/docker.sock:/var/run/docker.sock```, it should work... at least it does on the [testing](https://github.com/iquasere/MOSCA/actions/runs/4831773613). 

Or is there something else I'm missing?

dpryan79:
@iquasere Exactly, the need for a network connection is to allow snakemake to create its environments and databases to be downloaded. You're correctly that if you pass in networking options then the containers will work, though they'll also need to redownload dependencies and such unless you additionally pass in storage for them to make them permanent. None of that's a big deal, just something to be aware of.

PR #40829, opened 2023-05-07
Update angsd recipe enable osx build

isinaltinkaya:
Remove osx build skip statement.
Increase build number since it is a new build of the same version.

### DIFF
--- a/recipes/angsd/meta.yaml
+++ b/recipes/angsd/meta.yaml
@@ -12,8 +12,8 @@ source:
 
 
 build:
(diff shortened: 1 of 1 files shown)


isinaltinkaya:
The osx build error is fixed by [angsd commit 25afb62](https://github.com/ANGSD/angsd/commit/25afb62fd479296394229e317e7a33cf2f2df2cc). This will be included in the next version release and I will send the PR for recipe again with the new version.

Issue #41050, opened 2023-05-18
Missing curl in kaiju recipe


danpal96:
According to https://github.com/bioinformatics-centre/kaiju#creating-the-reference-database-and-index Kaiju needs both wget and curl but only wget is in the recipe

kbseah:
Pull request has been merged, closing this issue. Thanks for the fix!

PR #41051, opened 2023-05-18
Update kaiju recipe with curl requirement

danpal96:
As described in #41050 Kaiju needs curl as a requirement


peterjc (approved):
Looks good, but the CI testing is currently unavailable...

Known issue 41025 (not linking to reduce noise there).

peterjc:
Thanks all :)

PR #41073, opened 2023-05-20
Update recipe

heche-psb:
Update recipe


dpryan79:
Can this use paml 4.10.6? That's available in conda-forge.

heche-psb:
Currently paml 4.9j works proper for sure, not sure whether the higher version will lead to crash or not. I put it as 4.9j for now. Future I will test it with higher versions of paml.

dpryan79:
OK, I ask since 4.9j isn't available.

heche-psb:
I have added a new dependency. Please review and merge.

mencian:
Closing in favor of #47445.

PR #41078, opened 2023-05-21
Add callingcardstools

cmatKhan:
Add [callingcardstools](https://github.com/cmatKhan/callingCardsTools), a package which provides an API and cmd line tools to process Calling Cards data. Calling cards is an assay orthogonal to ChIP-seq. It provides insight into where a given transcription factor binds to the DNA.

@BiocondaBot please add


peterpru:
@BiocondaBot please add label

recipes/callingcardstools/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  url: "https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz"
+  sha256: 74a20d1eb22f7bdf3373e70e57737caefe4ea63f115e975dba6c04be814288fd
+
+build:
+  noarch: generic
```
peterpru, line 13: Should this be `noarch: python`?

recipes/callingcardstools/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+requirements:
+  host:
+    - biopython
+    - edlib
+    - numpy
+    - pandas
+    - pip
+    - pysam
+    - poetry
+    - python
```
peterpru, line 28: For compatibility, is it possible to specify a version of python here? E.g. `python <3` or `python >=3.6`

recipes/callingcardstools/meta.yaml
cmatKhan: yes -- changed in [427ccaf](https://github.com/bioconda/bioconda-recipes/pull/41078/commits/427ccaf1f14fabc7dd87ff0371cba1ca8222c057)

recipes/callingcardstools/meta.yaml
cmatKhan: Yes -- I originally had this specified, and then looked at the pysam meta.yml as an example. There are not specified versions there.

My pyproject.toml has specified software versions. I erred on the side of removing them, under the assumption that the pyproject.toml would act to resolve any issues. However, I wasn't sure.

I added the python specification back, and I'm happy to add specs for the rest of the packages.

peterpru (approved):

PR #41306, opened 2023-06-03
Update bayescode to v1.2.1

ThibaultLatrille:
@BiocondaBot please add label

### DIFF
--- a/recipes/bayescode/build.sh
+++ b/recipes/bayescode/build.sh
@@ -4,7 +4,7 @@ rm -rf data/
 # Create bin directory
 mkdir -p "${PREFIX}"/bin/
 
-top -n 1
+top -n 1 -b | head -n 30
 # Build specifying CPP compiler and C compiler
 make CXX="${CXX}" tiny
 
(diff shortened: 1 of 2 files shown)


ThibaultLatrille:
The tests run smoothly locally but are hanging in the Azure cloud.

Given the logs, the typical pattern is that all dependencies are installed after 13min, the 'make/cmake' starts the compilation of Bayescode but stops at around 15min without exit code, it is just hanging. 
Also, it stops at different checkpoint in the compilation process, sometimes 50%, sometimes 80%, ...

Then after 45min, Azure cloud stop the VM with the error:
 `The hosted runner encountered an error while running your job. (Error Type: Disconnect).`
 
So my interpretation is that the process is running out of memory (RAM and/or swap).
Looking at the memory (with `top`), only 130Mo of RAM are still available in the VM before the compilation starts.
However, I only require :
- compile('cxx')
- make
- cmake

Am I mistaken that it is a memory issue ? How can I free memory so that the tests can pass?

baraaorabi:
I'm also having the same problem with our PR #41444. I hope this gets resolved soon

baraaorabi:
Ok, this is crazy, but I think we got our PR (#41444) deprovisioning issue resolved. Passing `${CPU_COUNT}` to `make` seem to solve our issue 🤷‍♂️

It passed the Linux test and running the OSX test now...

PR #41543, opened 2023-06-17
Update recetox-xMSannotator to 0.10.0

xtrojak:
Update recetox-xMSannotator to 0.10.0. This introduced new dependencies and version pinning.

### DIFF
--- a/recipes/r-recetox-xmsannotator/meta.yaml
+++ b/recipes/r-recetox-xmsannotator/meta.yaml
@@ -8,7 +8,7 @@ package:
 
 source:
   url: "{{ github }}/archive/refs/tags/v{{ version }}.tar.gz"
-  sha256: b59eaae4e76c917be486b3b2568b010350e246ec510fb284c4697e6f71a17299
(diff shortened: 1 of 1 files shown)


xtrojak:
@BiocondaBot please add label

recipes/r-recetox-xmsannotator/meta.yaml
```diff
@@ -8,42 +8,62 @@ package:
...
 
+channels:
+  - conda-forge
+  - bioconda
 
 requirements:
   host:
-    - r-base
-    - r-dplyr >=1
+    - r-base =4.1.1
```
bgruening, line 27: I would not pin it that strict if possible.

hechth: When using the environment script to create the environment, pinning the R version massively reduces the time needed to resolve. Not sure how this maps to the host section of a recipe.

hechth:
Given the global pinnings for R, I think by being so specific we will make the package difficult to use in combination with any other package

hechth (approved):

PR #41637, opened 2023-06-22
Update callingcardstools

cmatKhan:
Adding a version to the software dependency `edlib`. This should address an issue in which `callingcardstools` installs, but does not behave correctly, when an earlier version of `edlib` is used.


cmatKhan:
@BiocondaBot please add label

cmatKhan:
@BiocondaBot please update

rpetit3 (approved):

Issue #41526, opened 2023-06-15
OSX builds fail with ``KeyError: 'https://conda.anaconda.org/conda-forge'``


f0t1h:
```
22:36:43 BIOCONDA ERROR COMMAND FAILED (exited with 1): /opt/mambaforge/envs/bioconda/bin/conda mambabuild --override-channels --no-anaconda-upload -c conda-forge -c bioconda -c defaults -e /opt/mambaforge/envs/bioconda/conda_build_config.yaml -e /opt/mambaforge/envs/bioconda/lib/python3.8/site-packages/bioconda_utils/bioconda_utils-conda_build_config.yaml recipes/tksm/meta.yaml
22:36:43 BIOCONDA ERROR STDOUT+STDERR:

```
```
Traceback (most recent call last):
  File "/opt/mambaforge/envs/bioconda/bin/conda-mambabuild", line 10, in <module>
    sys.exit(main())
  File "/opt/mambaforge/envs/bioconda/lib/python3.8/site-packages/boa/cli/mambabuild.py", line 256, in main
    call_conda_build(action, config)
  File "/opt/mambaforge/envs/bioconda/lib/python3.8/site-packages/boa/cli/mambabuild.py", line 228, in call_conda_build
    result = api.build(
  File "/opt/mambaforge/envs/bioconda/lib/python3.8/site-packages/conda_build/api.py", line 180, in build
    return build_tree(
  File "/opt/mambaforge/envs/bioconda/lib/python3.8/site-packages/conda_build/build.py", line 3078, in build_tree
    packages_from_this = build(metadata, stats,
  File "/opt/mambaforge/envs/bioconda/lib/python3.8/site-packages/conda_build/build.py", line 2038, in build
    output_metas = expand_outputs([(m, need_source_download, need_reparse_in_env)])
  File "/opt/mambaforge/envs/bioconda/lib/python3.8/site-packages/conda_build/render.py", line 787, in expand_outputs
    for (output_dict, m) in deepcopy(_m).get_output_metadata_set(permit_unsatisfiable_variants=False):
  File "/opt/mambaforge/envs/bioconda/lib/python3.8/site-packages/conda_build/metadata.py", line 2524, in get_output_metadata_set
    conda_packages = finalize_outputs_pass(
  File "/opt/mambaforge/envs/bioconda/lib/python3.8/site-packages/conda_build/metadata.py", line 884, in finalize_outputs_pass
    fm = finalize_metadata(
[...]

PR #41751, opened 2023-06-29
Verkko v1.4 re-enable OSX build

skoren:


skoren:
@bioconda/core still having issues with OS X builds and mamba. Different than the error in #41526 but it is trying to install a python 2.7 version of a package (parasail) when I already specified python 3.7+ for the recipe. Parasail seems to exist for newer python: https://anaconda.org/bioconda/parasail-python. This also works ok on Linux and this dependency hasn't changed from the previous verkko release which build on OSX fine. Not sure what to try to make it load the right python linked parasail?

dpryan79:
We're no longer getting the python 2.7 issue, but instead an issue related to graphaligner. This was rebuilt a month ago, but only on Linux. In fact the recipe for this is very odd, since we're hard-coding an old boost into the build. Something tells me it doesn't build with newer boosts (or at least not easily). This is the root of the issue, so if someone can get graphaligner to build with newer boost versions then we should be able to get OSX re-enabled.

dpryan79:
I'll try to debug the graphaligner stuff in a different PR. No promises that I'll get it to work, this can be very non-obvious :(

skoren:
I agree it makes sense to handle graph aligner in another pull request but I saw no way to do that while having this one open.

It looks like it might need to be boost/libbost version 1.73 as that's the latest libboost available. Graphaligner tries to build a static package which only works on Linux, thus the difference between build and run between OSX and Linux in the recipe. I don't think boost-cpp is enough as it doesn't seem to provide the accessory as static libraries.

skoren:
I'm going to close this since it has a bunch of unrelated graphaligner changes. Not sure why graphaligner is broken, I can build on both OSX and Linux fine with new versions of boost but it seems boost is just broken on conda. I'll try a new pull request to experiment with graphaligner builds.

PR #41880, opened 2023-07-07
Remove outdated dependency restrictions for pangolin

wm75:
Since v4.3 pangolin works again with the latest releases of all its dependencies.


wm75:
@BiocondaBot please add label

PR #41953, opened 2023-07-11
Graphaligner fixing OSX build

skoren:
OSX complains about boost v1.67.0 so trying to see if we can update the version and get graphaligner built.


skoren:
@bioconda-bot fetch artifacts

skoren:
@dpryan79 Hey, I started with your base in #41828 and was able to build both by updating boost and turning off static compilation on conda. This was the reason the library was missing, the static boost packages do not get included in the conda build of boost. I think it should be OK to use non-static compilation on conda since we already do that on OSX and it should have all the necessary libraries present at runtime too.

skoren:

skoren:
@maickrau can you review my patch change and confirm it it is OK? This removes the dependency on a fixed very old version of boost.

maickrau:
The ls and echo in build.sh look unnecessary, otherwise look fine to me

skoren:

skoren:
@BiocondaBot please add label

shenwei356 (approved):

PR #42029, opened 2023-07-18
Update cogent3 recipe to version 2023.7.18a1

khiron:
Milestone release of Cogent3 version 2023.7.18a2


jiarong (approved):

recipes/cogent3/meta.yaml
```diff
@@ -16,17 +17,19 @@ build:
 
 requirements:
   host:
-    - python >=3
+    - python >=3.8, <3.12
```
bgruening, line 20: why do you restrict this to 3.12?

khiron: Because of the dependence of cogent3 on `numba`, there is some delay in which versions of Python can be supported.

khiron:
closed to restate updated recipe for cogent3 2023.7.18a1

PR #42145, opened 2023-07-24
Pangolin scikit-learn fix

wm75:


bgruening (approved):

wm75:
Hmm, the issue is that snakemake dropped support for Python < 3.9 (https://github.com/snakemake/snakemake/blob/9cfcd917e6430cec72217a284222ed19574677f3/snakemake/__init__.py#L8-L9) in the very recent "minor" release 7.30.2 without mentioning it in the release notes.

Not sure where this should be fixed, but probably not here?

wm75:
I assume #42147  would be the correct fix?

wm75:
Excluding snakemake 7.30.2 for now since updating to Python 3.9 would also require unpinning Biopython and possibly other changes, which seems like a drastic change for just a new build.
Let's do this only on the next release of pangolin.

PR #42147, opened 2023-07-24
Update snakemake Python dependency

wm75:
7.30.2 requires at least Python 3.9

I hope that's correct @johanneskoester?

### DIFF
--- a/recipes/snakemake/meta.yaml
+++ b/recipes/snakemake/meta.yaml
@@ -9,7 +9,7 @@ source:
   sha256: 0cb86cf9b43b9f2f45d5685cd932595131031c7087690f64c5bc7eaec88df029
 
 build:
-  number: 0
+  number: 1
   noarch: generic
 
 requirements:
@@ -85,11 +85,11 @@ outputs:
 
     requirements:
       host:
-        - python >=3.7
+        - python >=3.9
         - pip
         - setuptools
       run:
-        - python >=3.7
+        - python >=3.9
         - setuptools
         - wrapt
         - requests


tsibley:
I concur with this PR. I noticed the issue when some of our CI with Python 3.8 + Snakemake 7.30.2 from Bioconda started failing with:

```
  File "…/bin/snakemake", line 6, in <module>
    from snakemake import main
  File "…/lib/python3.8/site-packages/snakemake/__init__.py", line 9, in <module>
    raise ValueError("Snakemake requires at least Python 3.9.")
ValueError: Snakemake requires at least Python 3.9.
```

when 7.30.2 should never have been installed on 3.8 to begin with.

tsibley:
@BiocondaBot please add label

corneliusroemer:
Verified correctness of this PR, see
https://github.com/snakemake/snakemake/pull/2341#issuecomment-1648487223
https://github.com/snakemake/snakemake/pull/2370
https://github.com/snakemake/snakemake/issues/2369

corneliusroemer (approved):

corneliusroemer:
I don't know enough about build numbers to know whether Python 3.8 will now try to pick build 0 (before the version requirement bump was part of conda metadata).

tsibley:
Indeed the solver (well, at least libmamba's solver) will just pick the build 0 instead. :-(

I thought it would, but tested anyway and was initially surprised because `snakemake ==7.30.2 hdfd78af_1` (the build from this PR) got installed. Unfortunately it was able to be installed because it pulled in the `pyhdfd78af_0` build of the `snakemake-minimal` package of the same version.

Can one yank releases/builds from Bioconda? I believe Anaconda's repositories let you do this, but who can do that for Bioconda?

Separately, it seems this pattern of the `snakemake` Conda package depending on the `snakemake-minimal` "sub" package should not only pin the exact same version as it does now but also the exact same build string. This doesn't functionally change the situation here, but it would likely help avoid other build mismatch issues in the future.

tsibley:
Until someone can yank these files from Bioconda's channel:

```
snakemake-7.30.2-hdfd78af_0.tar.bz2
snakemake-minimal-7.30.2-pyhdfd78af_0.tar.bz2
```

then anyone with a Conda env with Python 3.7 or 3.8 trying to install Snakemake (without a version restriction) will get a broken version.

A workaround is to exclude 7.30.2, e.g.:

    micromamba install 'snakemake !=7.30.2'

but this version restriction will be required in perpetuity (assuming Snakemake doesn't reintroduce support for 3.7 or 3.8).

PR #42339, opened 2023-08-04
Update meta.yaml

cjw85:
I believe @rrwick can confirm, my perusal of the source suggests that muscle 5 is supported and so the pin can be removed.

### DIFF
--- a/recipes/trycycler/meta.yaml
+++ b/recipes/trycycler/meta.yaml
@@ -7,7 +7,7 @@ package:
 
 build:
   noarch: python
-  number: 0
+  number: 1
   script: {{ PYTHON }} -m pip install --no-deps --ignore-installed .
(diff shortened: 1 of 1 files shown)


rrwick:
While MUSCLE v5 is technically supported, I have found MUSCLE v3 generally works better in Trycycler, and it's much faster ([read more here](https://github.com/rrwick/Trycycler/wiki/FAQ-and-miscellaneous-tips#a-note-on-muscle-versions)). So my preference would be to leave `<4` in the recipe.

cjw85:
Thanks Ryan. I've gone on a crusade to compile packages for aarch64 and did muscle v5 before realising that trycycler wants <4 😞.  I'll go make a muscle v3 pacakge.

PR #42394, opened 2023-08-08
Update scepia

Maarten-vd-Sande:
scepia does not work with gimmemotifs 0.17.2 and higher, plus @rebecza and me are new maintainers of the recipe


bgruening (approved):

bgruening:
Thanks

PR #42442, opened 2023-08-11
Update verkko to v1.4.1

skoren:
Also fixes OSX support, adds dependencies missing in v1.4, and expand tests to be more comprehensive (to catch the missing dependencies in future)


bgruening (approved):
Great! Thanks a lot for your work.

PR #42531, opened 2023-08-17
mob_suite biopython min requirement bumped to 1.80 and higher

kbessonov1984:
Need to specify explicitly minimum biopython requirement to version 1.80 and up. Otherwise conda installs biopython versions that are not compatible with the current tool release. Thus updated the `biopython` recipe requirement 

-


rpetit3 (approved):

rpetit3:
All good here @kbessonov1984 ? If so we can merge, and close the auto bump PR

kbessonov1984:
Yes, it finally works. I am satisfied with the result

PR #42666, opened 2023-08-25
fix: freebayes

kilpert:
Fix freebayes-prarallel because of bug in vcflib=1.0.9 (https://github.com/vcflib/vcflib/issues/389).
Also, htslib=1.17 does not work with freebayes-parallel.

### DIFF
--- a/recipes/freebayes/meta.yaml
+++ b/recipes/freebayes/meta.yaml
@@ -9,7 +9,7 @@ source:
   sha256: 6016c1e58fdf34a1f6f77b720dd8e12e13a127f7cbac9c747e47954561b437f5
 
 build:
-  number: 6
+  number: 7
   skip: true  # [osx]
 
 requirements:
--- a/recipes/freebayes/meta.yaml
+++ b/recipes/freebayes/meta.yaml
@@ -32,7 +32,8 @@ requirements:
     - parallel
     - samtools
     - tabixpp
-    - vcflib
+    - vcflib=1.0.3  # vcflib=1.0.9 does not work with freebayes-parallel, see https://github.com/vcflib/vcflib/issues/389
+    - htslib =1.16  # htslib=1.17 does not work with freebayes-parallel
     
 test:
   commands:
--- a/recipes/freebayes/meta.yaml
+++ b/recipes/freebayes/meta.yaml
@@ -25,13 +25,11 @@ requirements:
   host:
     - bzip2
     - htslib
-    - tabixpp =1.1.0 # tabixpp =1.1.2 no working with freebayes-parallel
-    - vcflib =1.0.3  # vcflib=1.0.9 not working with freebayes-parallel, see https://github.com/vcflib/vcflib/issues/389
+    - vcflib
   run:
     - parallel
     - samtools
-    - tabixpp =1.1.0 # tabixpp =1.1.2 no working with freebayes-parallel
-    - vcflib =1.0.3  # vcflib=1.0.9 does not work with freebayes-parallel, see https://github.com/vcflib/vcflib/issues/389
+    - vcflib
     
 test:
   commands:
--- a/recipes/freebayes/meta.yaml
+++ b/recipes/freebayes/meta.yaml
@@ -11,6 +11,8 @@ source:
 build:
   number: 7
   skip: true  # [osx]
+  run_exports:
+    - {{ pin_subpackage("freebayes", max_pin="x") }}
 
 requirements:
   build:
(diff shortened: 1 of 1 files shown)


bgruening:
I think vcflib should be fixed. There is an underlying issue which we need to understand, isn't it?

johanneskoester:
I agree, but maybe this here can be done as a temporary workaround, such that freebayes at least works for the people for now. It is a widely used package. If you agree @bguening, this PR should however, definitely also point to the vcflib issue URL in the code of the meta.yaml, not only the PR message. And probably it should also add a back reference to that issue in the vcflib meta.yaml.

johanneskoester:
@kilpert what is the issue with htslib though? What error message appears when only vcflib is pinned?

bgruening:
The linked issue above is from 12.6, so everyone is thinking they do a temporary workaround :)
We are releasing here a new package with changes, that we need to revert once the vcflib is fixed. So more work ;) but feel free to merge. Its not a -1, more a -0 ;)

johanneskoester:
I completely agree with you. If I could name somebody to fix the origin of the issue, I would do it. But I don't know anybody who currently has the capacity (including myself) :-(.

johanneskoester:
Maybe @mencian as the author of the changes that seemingly introduce the issue in 1.0.9 would like to have a look?

johanneskoester:
cross post: https://github.com/bioconda/bioconda-recipes/commit/032977aed9bcf6941284034e443b234e36af62e5

johanneskoester:
The htslib issue actually originates from tabixpp not being pinned in vcflib as a dependency. Fixing that now by adding a run_export to tabixpp.
```
freebayes-parallel test.idk 1 -f resources/genome.dna.homo_sapiens.GRCh37.105.fasta results/recal/pF40.bam
vcfuniq: error while loading shared libraries: libtabixpp.so.0: cannot open shared object file: No such file or directory
vcfstreamsort: error while loading shared libraries: libtabixpp.so.0: cannot open shared object file: No such file or directory
```

johanneskoester:
That one is now fixed in PR https://github.com/bioconda/bioconda-recipes/pull/42667

johanneskoester:
Maybe this change is the issue: 

now:
```
cmake -S . -B build \
	-DZIG=ON -DOPENMP=ON -DWFA_GITMODULE=OFF \
	-DCMAKE_BUILD_TYPE=Release \
	-DBUILD_SHARED_LIBS=ON \
	-DCMAKE_INSTALL_PREFIX="${PREFIX}" \
	-DCMAKE_CXX_COMPILER="${CXX}" \
	-DCMAKE_CXX_FLAGS="${CXXFLAGS}" \
	-DCMAKE_C_COMPILER="${CC}" \
	-DCMAKE_INCLUDE_PATH="${PREFIX}/include" \
	-DCMAKE_LIBRARY_PATH="${PREFIX}/lib"
```

before:
```
cmake --build build/ --target install -j 4 -v

#cmake .. -DCMAKE_INSTALL_PREFIX=$PREFIX -DTABIXPP_LOCAL:STRING=$PREFIX/lib
cmake .. -DCMAKE_INSTALL_PREFIX=$PREFIX 
cmake --build . 
cmake --install .
```

The disabling of WFA seems suspicious to me. I think this causes the symbol to be missing.

johanneskoester:
Ok, I think we are close enough to get this done without merging this PR. If anybody could try to make a separate PR that tries to activate WFA building in vcflib again, that would be awesome. I have to prepare a talk now.

johanneskoester:
Related: https://github.com/bioconda/bioconda-recipes/pull/42671

johanneskoester:
And here we can work on the fix. Hopefully my test new case will show whether the problem remains.
https://github.com/bioconda/bioconda-recipes/pull/42673

johanneskoester:
> And here we can work on the fix. Hopefully my test new case will show whether the problem remains. #42673

It does show it. Great, that will make things easier. Contributions welcome!

johanneskoester:
Seems tha vcflib forgets to link vcfstreamsort against libwfa. I have a fix in the PR and also upstream. Hopefully the fix will solve the issue. Luckily, we will immediately see that in the test case I added to the recipe.

johanneskoester:
@bgruening thanks for forcing me to solve this properly ;-)

johanneskoester:
Fixes are merged. Let us rebuild to get the updated run_exports.

johanneskoester (approved):

PR #40675, opened 2023-04-28
Patch repodata to avoid dependencies on libdeflate versions not present in conda-forge's current_repodata.json

jmarshall:
Upstream libdeflate has never bumped its soversion, and all releases to date have been forward compatible. Conda packages for libdeflate changed from .so to .dylib on macOS in v1.3 but after that have also been forward compatible.

Building other conda packages that use libdeflate produces a tight bound on a single x.x libdeflate version, but we can use repodata patching to widen the bounds on previously build packages, due to this forward compatibility. Doing so insulates bioconda from older releases of libdeflate falling out of conda-forge's current_repodata.json, as the widened bounds of older bioconda packages can be satisfied by the newer libdeflates still in current_repodata.json.

This is another approach to preventing libdeflate falling out of conda-forge's _current_repodata.json_ from causing conda/conda#9905 / #24199. Applying this will fix the defect existing again today whereby `conda create -n tmp samtools` selects an ancient samtools version because recent ones all require a libdeflate that is no longer present in conda-forge's _current_repodata.json_.

Add an additional (optional) filtering option to `_pin_looser()` and change it and `_pin_stricter()` to do version comparisons properly (so e.g. 1.9 < 1.11).

----

Please read the [guidelines for Bioconda recipes](https://bioconda.github.io/contributor/guidelines.html) before opening a pull request (PR).

* If this PR adds or updates a recipe, use "Add" or "Update" appropriately as the first word in its title.
* New recipes not directly relevant to the biological sciences need to be submitted to the [conda-forge channel](https://conda-forge.org/docs/) instead of Bioconda.
* PRs require reviews prior to being merged. Once your PR is passing tests and ready to be merged, please issue the `@BiocondaBot please add label` command.
* Please post questions [on Gitter](https://gitter.im/bioconda/Lobby) or ping `@bioconda/core` in a comment.

<details>
  <summary>Please use the following BiocondaBot commands:</summary>

Everyone has access to the following BiocondaBot commands, which can be given in a comment:

<table>
  <tr>
    <td><code>@BiocondaBot please update</code></td>
    <td>Merge the master branch into a PR.</td>
  </tr>
  <tr>
    <td><code>@BiocondaBot please add label</code></td>
    <td>Add the <code>please review & merge</code> label.</td>
  </tr>
  <tr>
    <td><code>@BiocondaBot please fetch artifacts</code></td>
    <td>Post links to CI-built packages/containers. <br />You can use this to test packages locally.</td>
  </tr>
</table>

[...]

Issue #24199, opened 2020-09-02
Newest version of package not installed unless specified


mikecormier:
When installing a package, like `ggd`, an older version of the package is installed unless the newest version is provided. Example `conda install -c bioconda ggd` will install version `0.1.2`, which is a few versions away from the latest version. This problem has been seen with multiple packages from bioconda. 

Additionally, when a package is updated or removed other packages are downgraded to an older version even though the package being updated or removed has no dependencies on the other packages. Again, an example is when a package, say `vep` is updated or removed it downgrades `ggd` to version `0.1.2` even though vep does not depend on `ggd` or any of the dependencies `ggd` has. 

This seems to be a similar problem with the version priority when installing a specific package like `ggd`. How do we set version priority for a package so it doesn't revert back to an older version unless specified to do so?

dpellow:
@mikecormier - have you found a solution to this problem?

jmarshall:
This is a significant general problem impacting Conda's usefulness, also reported as conda/conda#9905 and having received some discussion there.

There are many other similar reports here, e.g., #24621, #24320, #24264, #24182, #24033, and #22824.

jmarshall:
@bioconda/core: This bug is causing confusion for users every day — most recently as reported in samtools/htslib#1192 — and has been ongoing for months. Whether this is properly a conda problem (cf conda/conda#9905) or something that could in principle be fixed with updates to the bioconda repo metadata files, please **pin this issue** and add an explanation of the workarounds (check the version you're offered and explicitly specify the latest version if necessary; mamba apparently) and ideally an overview of the roadmap towards fixing this (if there is one).

dpryan79:
@jmarshall As you're aware, we have no control over this issue. I'm happy to pin this, but be aware that it will not help since regular users almost never read such things. As you indicated, the only solutions are (1) to specify the exact version you want or (2) to use mamba, which lacks this annoying quirk.

jmarshall:
Conda `create`/`install` starts by downloading a trimmed-down _current_repodata.json_ metadata file from each channel, and by default only downloads the full _repodata.json_ files if it needs to. This issue occurs when the newest package version cannot be satisfied via _current_repodata.json_ but an older version can, so it uses that instead of retrying with the full _repodata.json_. Hence it can be worked around as follows:

1. Ask for the particular newest version (to prevent the older version from succeeding):
`conda create`/`install package==version`

2. Skip the _current_repodata.json_ attempt entirely:
[...]

PR #42895, opened 2023-09-07
Patch repodata to relax HTSlib dependency upper bounds

jmarshall:
Somewhat similarly to PR #40675 for libdeflate, this PR recommends patching htslib dependencies to extend version ranges instead of pinning htslib in bioconda. Pinning htslib causes deployment difficulties each time there is an upstream htslib/samtools/bcftools joint release; e.g. #42167 and #42168 have still not been merged six weeks after bcftools and samtools 1.18 were released, because the htslib pinning still has not been updated (cf bioconda/bioconda-utils#915).

The reason for pinning htslib is that it is a common dependency and if several tools using it are all going to be installed in the same environment they all need to be using (and hence built against) the same htslib version.

If we patch repodata to loosen these htslib dependencies, this need will go away. Then we can simplify updates by removing the pinning of htslib (which is now bioconda/bioconda-utils#917).

Upstream HTSlib is careful to maintain binary compatibility between soversion bumps. Note though that those bumps are not correlated with e.g. major version number bumps, as the HTSlib maintainers would use that to reflect a break in source compatibility not binary compatibility.

Hence building other bioconda packages that use htslib produces a tight bound on a single `x.x` htslib version, but we can use repodata patching to widen the bounds on previously build packages, due to this forward compatibility. Doing so will mean bioconda no longer needs to pin HTSlib, as it will be less vital to ensure all dependent packages are built against the exact same version of htslib. In turn, this will simplify packaging htslib/samtools/bcftools updates.

----

### General instructions

```yaml
build:
  run_exports:
    - ...

```

with `...` being one of:

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/recipes/bioconda-repodata-patches/gen_patch_json.py
+++ b/recipes/bioconda-repodata-patches/gen_patch_json.py
@@ -165,6 +165,12 @@ def _gen_new_index(repodata, subdir):
         if record_name.startswith('samtools') and record['subdir']=='linux-64' and not has_dep(record, "openssl") and not has_dep(record, "htslib"):
             deps.append('openssl >=1.1.0,<=1.1.1')
 
+        # future HTSlib versions are binary compatible until they bump their soversion
+        if has_dep(record, 'htslib'):
+            # skip deps prior to 1.10, which was the first with soversion 3
+            # TODO adjust replacement (exclusive) upper bound with each new compatible HTSlib
+            _pin_looser(fn, record, 'htslib', min_lower_bound='1.10', upper_bound='1.19')
+
         # future libdeflate versions are compatible until they bump their soversion; relax dependencies accordingly
         if record_name in ['htslib', 'staden_io_lib', 'fastp'] and has_dep(record, 'libdeflate'):
             # skip deps that allow anything <1.3, which contained an incompatible library filename
--- a/recipes/bioconda-repodata-patches/meta.yaml
+++ b/recipes/bioconda-repodata-patches/meta.yaml
@@ -1,13 +1,13 @@
 package:
   name: bioconda-repodata-patches
-  version: 20230506  # ensure that this is the "current" date, and always higher than the latest version in master
+  version: 20230907  # ensure that this is the "current" date, and always higher than the latest version in master
 
 source:
   path: .
 
 build:
   noarch: generic
-  number: 1
+  number: 0
 
 requirements:
   build:
--- a/recipes/bioconda-repodata-patches/meta.yaml
+++ b/recipes/bioconda-repodata-patches/meta.yaml
@@ -8,6 +8,8 @@ source:
 build:
   noarch: generic
   number: 0
+  run_exports:
+    - {{ pin_subpackage("bioconda-repodata-patches", max_pin=None) }}
 
 requirements:
   build:


dpryan79 (approved):

dpryan79:
@jmarshall This looks great and I completely agree that this should make things easier in the future. Thanks a ton!

jdblischak:
> Hence building other bioconda packages that use htslib produces a tight bound on a single x.x htslib version, but we can use repodata patching to widen the bounds on previously build packages, due to this forward compatibility. Doing so will mean bioconda no longer needs to pin HTSlib, as it will be less vital to ensure all dependent packages are built against the exact same version of htslib. In turn, this will simplify packaging htslib/samtools/bcftools updates.

What about the `run_exports` in the htslib recipe itself? If the repodata patch is automatically loosening the htslib runtime requirement for all recipes, should we loosen (or simply remove) the current `run_exports`?

https://github.com/bioconda/bioconda-recipes/blob/d40dcb531fd430ad56518e0ccbfcfa4b1b29d60a/recipes/htslib/meta.yaml#L9-L10

jmarshall:
Loosening the `run_exports` would require predicting the future. HTSlib has previously bumped its shared library soversion in versions 1.4 and 1.10, and there is no predicting in advance when it will do so again. As [conda does not model soversion dependencies](https://twitter.com/jomarnz/status/1298187919786741762), it needs to build packages using overly conservative version dependencies for libraries instead. See also the discussion (in a somewhat different context) in https://github.com/bioconda/bioconda-recipes/issues/24199#issuecomment-758553486 onwards.

(If you think semver would miraculously solve this problem, reflect on the fourth paragraph in the initial comment above and the fact that semver does not distinguish between source code compatibility and ABI compatibility.)

jdblischak:
> Loosening the `run_exports` would require predicting the future. HTSlib has previously bumped its shared library soversion in versions 1.4 and 1.10, and there is no predicting in advance when it will do so again.

@jmarshall thanks for the explanation! That makes sense

> As [conda does not model soversion dependencies](https://twitter.com/jomarnz/status/1298187919786741762), it needs to build packages using overly conservative version dependencies for libraries instead

Yes, I recall that discussion on Twitter. Reading your other posts helped solidify my understanding

---

I think part of my confusion yesterday was not understanding the arguments to `_pin_looser()`. They aren't documented, but from reading the function definition, I think I understand now:

https://github.com/bioconda/bioconda-recipes/blob/a372835f7348577899bfe26855a3951465d40203/recipes/bioconda-repodata-patches/gen_patch_json.py#L168-L172

translates to:

* if the htslib pin is <1.10, do nothing
* if the htslib pin is >=1.10, then edit the upper bound to be <1.20

Thus a recipe built against htslib 1.10 would be edited to have a pin of `htslib >=1.10,<1.20`, a recipe built against htslib 1.15 would be edited to have a pin of `htslib >=1.15,<1.20`, etc. As far as I can tell, there is no way to reduce the lower pin. But I assume that is because the same SO version indicates forward compatibility and not backwards compatibility.

Now let's say that hypothetically in the future htslib 1.20 introduces a new SO version, and then releases 1.21, 1.22, and 1.23 are compatible. My guess would be that we would add an additional `_pin_looser()` call to edit binaries built against htslib 1.20+, eg

```python
 # future HTSlib versions are binary compatible until they bump their soversion 
 if has_dep(record, 'htslib'): 
     # skip deps prior to 1.10, which was the first with soversion 3 
     # TODO adjust replacement (exclusive) upper bound with each new compatible HTSlib 
     _pin_looser(fn, record, 'htslib', min_lower_bound='1.10', upper_bound='1.20') 
    # htslib 1.20 bumped the soversion
     _pin_looser(fn, record, 'htslib', min_lower_bound='1.20', upper_bound='1.24') 
```

Do I understand the plan correctly?

jmarshall:
Bioconda-repodata-patches's scripts are fairly niche and undocumented. You have translated the existing code correctly.

A program built against htslib-1.15 will be compatible until htslib bumps its soversion (unless they make a mistake!), but as it may well use functions newly introduced in htslib-1.15 it won't in general be compatible with earlier versions such as 1.14. Thus it's mostly not useful to reduce the lower pin — but if it was useful for something such functionality could be added to _gen_patch_json.py_.

Adding another `_pin_looser(…)` call like that is indeed the plan for after HTSlib bumps its soversion again (along with moving the TODO comment down, as only the latest `_pin_looser` ever needs adjusting). I could have added `_pin_looser(…, '1.0', '1.4'); _pin_looser(…, '1.4', '1.10')` initially to do this for those older versions too, but those were already ancient history at the time so best left alone IMHO.

PR #42924, opened 2023-09-08
Add hapnest new recipe

msantorsola:
This PR adds a recipe for [hapnest](https://github.com/intervene-EU-H2020/synthetic_data), a tool for genotype and phenotype data simulation.


recipes/hapnest/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  test:
+   commands:
+    -  generate_pheno | grep "option --config requires an argument"  #julia --threads "$1" run_program.jl --genotype --config "$2" | grep "option --config requires an argument"
+
+about:
+  summary: "Meta-package for hapnest which simulates genotype and phenotype data"
+  license: "The license for this meta-package is GNU General Public License v3.0; individual tools vary"
+
+extra:
+  skip-lints:
```
bgruening, line 40: why do you skip those lints?

recipes/hapnest/meta.yaml
```diff
@@ -0,0 +1,44 @@
+package:
+  name: hapnest
+  version: "1"
+
+build:
+  noarch: generic
+  number: 0
+  run_exports:
+    - julia >=1.6.4
```
bgruening, line 9: This does not look correct. Please look at the PR description for some instructions.

recipes/hapnest/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+source:
+  url: https://github.com/intervene-EU-H2020/synthetic_data/archive/refs/heads/main.zip
```
bgruening, line 12: can you maybe tag a release and use that instead of `main`

recipes/hapnest/meta.yaml
msantorsola: Those lints are now included

recipes/hapnest/meta.yaml
msantorsola: Thank you, I fixed run_exports.

recipes/hapnest/meta.yaml
msantorsola: There are no releases available

recipes/hapnest/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/intervene-EU-H2020/synthetic_data/archive/refs/heads/main.zip
+  sha256: 6457c711e3430420d07b2ca8e22303a08ab60811b34c9f005a4c87bbe1ea1ade
+
+build:
+  number: 0
+  skip: true
```
aliciaaevans, line 14: I think this is why you aren't getting a build.

msantorsola: Thank you @aliciaaevans, skip line deleted.

PR #43194, opened 2023-09-21
Update ninja-nj for OSX

dirkjanvw:
I was looking into the ninja-nj package on bioconda for macOS and I found out that it still is an issue for `mamba install repeatmodeler` on macOS. I noticed that it should be quite simple to fix this for macOS, so that's what this PR changes.

Related to this, #33626 seems to mention that there were problems to add this recipe in the past for macOS. @Juke34 since you originally looked into this issue, could you please confirm whether this PR fixes it for macOS?

### DIFF
--- a/recipes/ninja-nj/meta.yaml
+++ b/recipes/ninja-nj/meta.yaml
@@ -12,7 +12,7 @@ source:
 build:
   number: 3
   run_exports:
-    - {{ pin_subpackage(name, max_pin="x") }}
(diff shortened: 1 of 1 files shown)


dirkjanvw:
@BiocondaBot please add label

Juke34 (approved):

Juke34:
Thanks @dirkjanvw for this update

PR #43251, opened 2023-09-25
Update: Rebuild samtools

kdm9:
I'm having the same issue with samtools 1.17 as described in this old issue from 2018: https://github.com/samtools/samtools/issues/790

Last time a build bump fixed it, so probably there's been an update to ncurses that broke the linkage


mencian (approved):
Looks good to me!

PR #43252, opened 2023-09-25
Update seqhax

kdm9:
Bump seqhax whose version is quite out of date and has library conflicts preventing use


mencian (approved):
Looks good to me!

PR #43288, opened 2023-09-27
Update post-link.sh script for RepeatMasker

dirkjanvw:
On macOS, the current RepeatMasker installation is non-functional because the Dfam database is not correctly downloaded. This is because `zcat` works differently on linux compared to osx. Therefore, I propose in this PR to use `gunzip -c` instead of `zcat`.


dirkjanvw:
@abretaud @bgruening since you were involved in the addition of the `post-link.sh` script to this recipe earlier (#40484), could you please let me know what you think of the change I propose here before I add the label for review and merge?

abretaud:
Ah, no objection if it works the same, but you need to bump the build number too

dirkjanvw:
Ah yes, indeed. I also added the run_exports as it is required now too.

I'll still test this recipe on both macOS and Linux to make sure it works!

dirkjanvw:
Hmm it seems the pipeline can no longer extract the source tarball for repeatmasker (`wget https://www.repeatmasker.org/RepeatMasker/RepeatMasker-4.1.5.tar.gz`), whereas it could previously. Do you know a way around this?

dirkjanvw:
@bioconda/core could you please help out here? how can I let the pipeline succeed (no space left on device error)? it did work out for the previous build of the same version (so same unpacking the same tarball)

dirkjanvw:
@BiocondaBot please add label

mencian:
LGTM! Thanks, @dirkjanvw, for looking into this!

mencian (approved):

abretaud:
@dirkjanvw what's the reason to no longer include the db in the container image in https://github.com/bioconda/bioconda-recipes/pull/43288/commits/013fa23b460a66b98f257061462c96eb4f4f52ca ?
It breaks a galaxy tool that expect the db to be there

mvdbeek:
Can we delete the docker and singularity images in the meantime ?

dirkjanvw:
Ah sorry, I didn't know. The reason that I removed the database as well (next to fixing the post link script) is that the pipeline ran out of space when building the container (see messages above): someone on Gitter suggested to follow the example of lotus to solve this issue but if you know of another way of downloading the full database without breaking the pipeline I see no issue with putting it back.

mvdbeek:
If by pipeline you mean the CI, yes, use of zcat as before should prevent running out of space.

dirkjanvw:
Hmm but that doesn't work for OSX as `zcat` works differently there.

bgruening:
zcat is in the gzip package. Maybe it helps to put this as host dependency for osx.

```
host:
  - gzip # [osx]

abretaud:
Opened https://github.com/bioconda/bioconda-recipes/pull/43994 to fix this

PR #43322, opened 2023-09-28
Add ViralConsensus v0.0.4

niemasd:
This PR adds ViralConsensus v0.0.3 to bioconda.

----

### General instructions

```yaml
build:
  run_exports:
    - ...

```

with `...` being one of:

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

</details>


niemasd:
@BiocondaBot please add label

jmarshall:
Your _meta.yaml_ file lists your package as depending on bioconda's htslib package. This is the right way to build a bioconda package: depend on the common externally packaged library, rather than bundle it within your package.

Corresponding to this, in _build.sh_ you should remove the cloning and building of htslib and instead patch and/or configure ViralConsensus (if necessary) so that it can build against an already installed htslib.

niemasd:
@jmarshall Thanks for the feedback! I've refactored the ViralConsensus code base to use the environment's htslib installation, and I've refactored the bioconda recipe to properly use the conda environment's htslib. It now passes the tests again. Can you review again?

recipes/viral_consensus/build.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# update environment variables
+export CFLAGS="$CFLAGS -I$PREFIX/include"
+export CPPFLAGS="$CPPFLAGS -I$PREFIX/include"
+export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
+export CPATH=${PREFIX}/include
+export INCLUDES="-I${PREFIX}/include"
+export LIBPATH="-L${PREFIX}/lib"
+export LIBRARY_PATH=${PREFIX}/lib
```
jmarshall, line 10: There's no need to define any of these, as you're using conda's compilers. (It's possible that might break and you might need to re-add one or two of these, but I expect not.)

recipes/viral_consensus/build.sh
```diff
@@ -0,0 +1,18 @@
...
+
+# compile ViralConsensus
+make CC=$CC CXX="${CXX}"
+if [ ! -d "$PREFIX/bin" ]; then
+    mkdir $PREFIX/bin;
+    export PATH=$PREFIX/bin:$PATH;
```
jmarshall, line 16: No need to add to PATH for _build.sh_.
Most recipes just use `mkdir -p $PREFIX/bin` instead of a conditional.

recipes/viral_consensus/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  version: {{ version }}
+
+build:
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('viral_consensus', max_pin="x.x.x") }}
+
+source:
+  url: https://github.com/niemasd/ViralConsensus/archive/refs/tags/{{ version }}.tar.gz
+  sha256: d531311c8e5c6b689f4426a9e89cc00426733ae9d4ff71757a861ad6ddbf81f4 # curl -sL URL | openssl sha256
```
jmarshall, line 14: In future the autobump script will show you new sha256 values, so you won't need the mnemonic comment.

recipes/viral_consensus/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - bzip2
+    - htslib
+    - libcurl
+    - make
+    - ncurses
+    - zlib
```
jmarshall, line 25: Just `make` and the two compilers for `build:`

recipes/viral_consensus/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  host:
+    - bzip2
+    - htslib
+    - libcurl
+    - ncurses
+    - zlib
```
jmarshall, line 31: Just `htslib` for `host:` — you don't use `ncurses` at all and the others you don't use directly.

recipes/viral_consensus/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  run:
+    - bzip2
+    - htslib
+    - libcurl
+    - ncurses
+    - zlib
```
jmarshall, line 37: Just `htslib` for `run:` or in fact you should probably leave this empty and the dependency will be automatically derived.

recipes/viral_consensus/build.sh
niemasd: Thanks! I'll remove them, and if something breaks, I'll re-add them as needed (I think the Mac OS X build was failing until I added `export LIBRARY_PATH=${PREFIX}/lib`)

recipes/viral_consensus/build.sh
niemasd: Thanks! I'll replace with just `mkdir -p` and remove the conditional

recipes/viral_consensus/meta.yaml
niemasd: Thanks! I more so left it as a note for myself if I ever add more tools to bioconda in the future, but I'll remove it for cleanliness (it was easy to Google anyways)

recipes/viral_consensus/meta.yaml
niemasd: Thanks! Removed everything else

recipes/viral_consensus/meta.yaml
niemasd: Thanks! Removed everything else

recipes/viral_consensus/meta.yaml
niemasd: Thanks! A similar tool (`ivar`) seems to have `htslib` under `run:`, but I'll try it without it and re-add if it breaks things

jmarshall: In the past it was necessary, but more recently most libraries such as htslib propagate the right thing from `host:`.

niemasd:
Following up after the tests finished: Linux is crash due to not being able to find the `-llzma`, `-lbz2`, `-lz`, `-lcurl`, and `-lhts` libraries. Not sure if it's because I removed these from `build:` in `meta.yaml` or because of the removal of environmental variable assignments in `build.sh`. Will play around and find out

jmarshall:
Make sure `-lhts` appears before the others.

niemasd:
As in edit the `Makefile` in my tool's GitHub repo to move `-lhts` earlier than `-llzma -lbz2 -lz -lcurl -pthread` in the `LIBS` declaration? I could do that, but I'd need to make a new release for it (otherwise the updated `Makefile` isn't in the current tarball)

niemasd:
@jmarshall I just rereleased v0.0.4 with `-lhts` first in the `LIBS` definition in the `Makefile`, and I updated the bioconda `meta.yaml` with the updated `sha256`. The tests are running, so fingers crossed it's okay to merge this time! 😄

niemasd:
@jmarshall Strange, just making this single change (moving `-lhts` first in the `Makefile` definition of `LIBS`) broke the Mac OS X build:

https://github.com/niemasd/ViralConsensus/commit/1ebe1584a50048fb5f68e7a8a96643518c305617

I'm not quite sure how to debug, as I don't have a Mac environment, and I can't seem to tell what failed in the logs:

https://dev.azure.com/bioconda/bioconda-recipes/_build/results?buildId=42408&view=logs&j=1b052f1d-4456-52f0-9d43-71c4c5bd734d&t=edc48dcd-1fc2-5e3b-7036-7be9cea00123&l=353

jmarshall:
This is likely something that I meant to mention previously: you should propagate conda's CXXFLAGS etc settings to your Makefile too. So in _build.sh_:

```sh
make CC="$CC" CXX="$CXX" CFLAGS="$CFLAGS" CPPFLAGS="$CPPFLAGS" CXXFLAGS="$CXXFLAGS" LDFLAGS="$LDFLAGS"
```

I see you had added some others too, but with the full complement of these standard ones you shouldn't need manual ones.

niemasd:
@jmarshall That now causes the Linux build to fail as well, for the same `ld` error message of not being able to find the libraries (I think because the library paths are not being set). I'll revert some of the changes until it gets back to a working state, and I'd rather submit it once it's back to a working state if possible, as I'm not sure how much cleaner it can get (it seems like it's already quite a bit cleaner than many other recipes in bioconda, and this is taking quite some time to tinker with)

niemasd:
@jmarshall After re-adding those library paths to the `make` command variables, it seems to have worked again and is now passing all the tests 😄

jmarshall:
The error messages are what tell us what is wrong with your recipe. I will be examining the previous build logs further.

It would not be appropriate to merge this in its current state, without understanding the causes of the various failures.

niemasd:
Sure, makes sense; thanks for checking the build logs!

jmarshall:
Unsurprisingly, the problem is that your Makefile doesn't actually use those conventional variables. I've pushed a commit to have the recipe put conda's standard settings variables into the `$(RELEASEFLAGS)` variable that the Makefile does use.

This is better than papering over the cracks with explicit `-L` options and `$LIBRARY_PATH` environment variables, as it preserves conda's `-rpath` settings that presumably are important to producing properly usable packaged executables (though you might expect that being missing to cause the test step's trial execution to fail too, which appears not to have happened).

jmarshall:

jmarshall (approved):

niemasd:
Thank you!!

jmarshall:
No problem :smile:

Bit annoying that the system has uploaded one of the already superseded 0.0.3 versions, but it won't really do any harm.

PR #43488, opened 2023-10-05
Update raxml to v8.2.13 that supports partially Linux ARM64

martin-g:
Only the AVX Makefiles do not support ARM64
The SSE ones use sse2neon.h when executed on Linux ARM64

Related-to: https://github.com/stamatak/standard-RAxML/pull/70 & https://github.com/stamatak/standard-RAxML/pull/71


martin-g:
Fixed with https://github.com/bioconda/bioconda-recipes/pull/46469

PR #43571, opened 2023-10-09
Update t-coffee and merge t_coffee

lrauschning:
This PR bumps the version of t-coffee to 13.46 and symlinks t_coffee (an outdated recipe missing quite some dependencies) to t-coffee.
I think that ideally software should only be packaged once, and in any case there should not be different recipes lacking required dependencies.
The symlink was done so as not to break the workflows of users that incorrectly have t_coffee as a dependency (though if I see it correctly, the workflow would crash in most cases anyway).


lrauschning:
The CI run got canceled because it ran for >360 min. Locally, the recipe builds successfully in 25 mins (using --mulled-test). I suspect the Azure node doesn't have enough RAM and starts swapping, making the compile take too long.
I'd be glad if someone else could test it on their machine to confirm that this recipe builds so it can be merged.

lrauschning:

mencian:

mencian:

mencian:
Updating the base branch to trigger CI rebuild seems to have done the trick. Thanks! @lrauschning

PR #43664, opened 2023-10-13
Add ucsc-{bed,big}-tools

mrvollger:
This adds all the UCSC tools starting with bed* or big* as two different packages, ucsc-bed-tools and uscsc-big-tools.


mrvollger:
@BiocondaBot please add label

recipes/ucsc-bed-tools/meta.yaml
```diff
@@ -0,0 +1,74 @@
+{% set package = "ucsc-bed-tools" %}
```
thanhleviet, line 1: Is it possible to separate the recipe per PR each?

mrvollger: What is the reason for doing so, if you don't mind me asking? 

To be honest, I wanted this to be one PR, but the very small file limits are requiring me to split binaries over multiple packages. Do you know if these file size limits can be increased?

thanhleviet: Because each recipe is usually seen as a separate software/tool, it'd be cleaner to have each as a separate PR; it might also affect CI processes if one of the recipes fails, then the entire PR will fail too. Do you want to include the binary in the recipe or where that has file size limits? I see the test instance of this PR completed successfully for both recipes.

mrvollger: I was hoping to make one package that had all of the utilities here:
https://hgdownload.soe.ucsc.edu/admin/exe/linux.x86_64/
A total of about 3.8GB. 

But the recipe for the failed with out-of-space errors, so I started splitting it into multiple packages, one for recipie for binaries starting bed, another for ones starting with big, etc. 

But ideally, all the binaries at the above URL could be part of one package. Do you know if there is a way to appeal the size limit so I could have one package for all these associated UCSC tools?

thanhleviet: Ah, I see, probably a core member will have a better idea @bgruening @rpetit3 ?

bgruening: This is currently not possible. Maybe skipping the container generation will help, but then we do not test the package in the container. Which might be fine in that case. But we would need to add support for skipping container builds / testing.

recipes/ucsc-bed-tools/build.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+mkdir -p "$PREFIX/bin"
+
+if [[ ${target_platform} =~ linux.* ]]; then
+    SRC="rsync://hgdownload.soe.ucsc.edu/genome/admin/exe/linux.x86_64/"
```
bernt-matthias, line 5: This dir should be versioned. Seems to contain version 459 already: https://hgdownload.soe.ucsc.edu/admin/exe/linux.x86_64/FOOTER.txt

Why you want to use these binaries at all instead of compiling as before: 
https://github.com/bioconda/bioconda-recipes/blob/master/recipes/ucsc-bedgraphtobigwig/build.sh

bernt-matthias (commented):
What is the usecase for having these recipes. Seems that for most we already have 1 per tool:

https://github.com/bioconda/bioconda-recipes/blob/master/recipes/ucsc-bedgraphtobigwig/build.sh

mrvollger:
I often use all these tools in combination. But more practically almost all the tools are 8 months out of date and significantly more than that for Mac, and a joint package would make things easier to keep up to date, and it's something the ucsc browser team would like as well. 

But the size limits are preventing a truly joint package (see previous comments) so I am debating the utility of doing anything.

PR #43708, opened 2023-10-16
Skip arm builds if no recipes request the platform

aliciaaevans:
NOTE: This would be a merge to the **arm-builds branch** (not master).

Skip remaining steps if there are no changed recipes that have a setting in meta.yaml for `additional-platforms` for the current platform. This will save time on CircleCI versus installing bioconda-utils and then not actually building anything.

I had to re-download yq because the one baked into the cimg/base image was amd64 and throws errors. (Thus all the debugging commits.)


.circleci/config.yml
```diff
@@ -86,6 +122,42 @@ jobs: # a basic unit of work in a run
...
+          name: Check for Additional Platforms
+          command: |
+            # Download ARM version of yq
+            yq_platform=$(uname)
+            yq_arch=$(uname -m)
+            [[ $yq_arch = "aarch64" ]] && yq_arch="arm64"
+            wget https://github.com/mikefarah/yq/releases/latest/download/yq_${yq_platform}_${yq_arch} -O /usr/local/bin/yq
+
+            # Find recipes changed from this merge
+            files=`git diff --name-only --diff-filter AMR ${CIRCLE_SHA1}~1 ${CIRCLE_SHA1} | grep -E 'meta.yaml$' `
```
martin-g, line 135: I was going to suggest to extract the common logic into `.circleci/check-for-additional-platforms.sh` and just call it here but I noticed this small difference between this step and the one at line 33 (https://github.com/bioconda/bioconda-recipes/pull/43708/files#diff-78a8a19706dbd2a4425dd72bdab0502ed7a2cef16365ab7030a5a0588927bf47R33). Maybe pass the SHAs as arguments to the shell script ?!

aliciaaevans: Good thinking

johanneskoester (approved):

PR #43751, opened 2023-10-18
Add me-PCR v1.0.6

peterjc:
Proposed description (and context to the PR):

> Multithreaded Electronic PCR (me-PCR) was developed by Kevin Murphy at
> Children's Hospital of Philadelphia. It is described in Murphy et al
> (2004) https://doi.org/10.1093/bioinformatics/btg466 and was originally
> available from http://genome.chop.edu/mePCR (defunct). It was based on the
> public domain NCBI tool e-PCR by Gregory D. Schuler, described in Schuler
> (1997) https://doi.org/10.1101/gr.7.5.541 which was origially availble
> from ftp://ncbi.nlm.nih.gov/pub/schuler/e-PCR/ (defunct). The final
> release of me-PCR was v1.0.6 (2008-02-18), and the author wrote that in
> general NCBI e-PCR should be used instead having been improved greatly
> over the years. The NCBI retired and withdrew the e-PCR webservice and
> command line tool on 2017-06-28, suggesting the online-only tool
> Primer-BLAST instead. However, this is not suitable for offline high
> throughput usage. See also Jim Kent's isPCR (aslo available in BioConda).

Quoting Murphy (2004), "Both source and executable code are freely available to the research community at http://genome.chop.edu/mePCR/ "

Quoting Schuler (1997), "The source code for the e-PCR program is freely available (ftp://ncbi.nlm.nih.gov/pub/schuler/e-PCR/)."

i.e. The licensing of me-PCR is not explicit, but is clearly a derivation of the NCBI e-PCR tool, which again is not explicit but I believe to be in the public domain (given the NCBI status).

It would help if anyone knew of a copy of the standalone e-PCR source code? https://ftp.ncbi.nih.gov/pub/schuler/e-PCR/RETIRED_README.txt just says it was removed, links to the announcement, and suggests Primer-BLAST instead.

### Bot commands for PR management

<details>
  <summary>Please use the following BiocondaBot commands:</summary>

Everyone has access to the following BiocondaBot commands, which can be given in a comment:

<table>
  <tr>
    <td><code>@BiocondaBot please update</code></td>
    <td>Merge the master branch into a PR.</td>
  </tr>
  <tr>
    <td><code>@BiocondaBot please add label</code></td>
    <td>Add the <code>please review & merge</code> label.</td>
  </tr>
  <tr>
    <td><code>@BiocondaBot please fetch artifacts</code></td>
    <td>Post links to CI-built packages/containers. <br />You can use this to test packages locally.</td>
  </tr>
[...]

PR #43752, opened 2023-10-18
Update e-PCR source URL and correct license

peterjc:
The NCBI withdrew the downloads, see https://ftp.ncbi.nih.gov/pub/schuler/e-PCR/RETIRED_README.txt and https://ncbiinsights.ncbi.nlm.nih.gov/2017/06/28/e-pcr-is-retiring-use-primer-blast/ - so using the Ubuntu mirror instead.

Also correcting license, NCBI e-PCR is Public Domain (but the Ubuntu/Debian package definition is GPL-2+). See e.g. http://changelogs.ubuntu.com/changelogs/pool/universe/e/epcr/epcr_2.3.12-1-8build1/copyright

Cross reference #43751 for packaging me-PCR, a fork of e-PCR.

### DIFF
--- a/recipes/e-pcr/meta.yaml
+++ b/recipes/e-pcr/meta.yaml
@@ -3,11 +3,12 @@ package:
   version: "2.3.12"
 
 build:
-  number: 1
+  number: 2
   skip: True # [osx]
 
 source:
-  url: ftp://ftp.ncbi.nlm.nih.gov/pub/schuler/e-PCR/e-PCR-2.3.12-1-src.tar.gz
+  # url: ftp://ftp.ncbi.nlm.nih.gov/pub/schuler/e-PCR/e-PCR-2.3.12-1-src.tar.gz
+  url: http://archive.ubuntu.com/ubuntu/pool/universe/e/epcr/epcr_2.3.12-1.orig.tar.gz
   md5: 516b111f78769af24393e61b0bf7fd80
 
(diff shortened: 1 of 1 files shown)


peterjc:
Looks like this recipe takes the provided Linux binaries as is, and thus skips macOS. Ideally we would compile the source - perhaps in a followup PR?

peterjc:
This feels like a false positive due to removing the compiler dependency since the recipe was reusing the NCBI provided Linux binaries,

```
LintMessage(recipe=Recipe "e-pcr", check=<class 'bioconda_utils.lint.check_noarch.should_be_noarch_generic'>, severity=<Severity.ERROR: 30>, title='The recipe should be build as `noarch`', body="\nPlease add:\n\n    build:\n      noarch: generic\n\nPackages that don't require a compiler to build are normally\narchitecture independent and go into the `noarch` subset of\npackages.", start_line=8, end_line=12, fname='recipes/e-pcr/meta.yaml', canfix=True)
The following problems have been found (visit https://bioconda.github.io/contributor/linting.html for details on the particular lints you get below.):

ERROR: recipes/e-pcr/meta.yaml:12: should_be_noarch_generic: The recipe should be build as `noarch`
Errors were found
```

@bioconda/core (in the absence of a C/C++ team), please advise. Is there an easy fix, or should we switch to actually compiling e-PCR here?

PR #43941, opened 2023-10-27
Add Phylogenize

Kekananen:
Adds [Phylogenize](https://github.com/pbradleylab/phylogenize) to conda. Allows for users to link microbial genes to environments, accounting for phylogeny. More specifically, given community composition data, phylogenize links genes in microbial genomes to either microbial prevalence in, or specificity for, a given environment, while also taking into account an important potential confounder: the phylogenetic relationships between microbes.

Paper: [(Bradley and Pollard, 2020)](https://pubmed.ncbi.nlm.nih.gov/31588499/)

### DIFF
--- a/recipes/phylogenize/meta.yaml
+++ b/recipes/phylogenize/meta.yaml
@@ -1,0 +1,47 @@
+{% set version = "0.91" %}
+{% set name = "phylogenize" %}
+
+package:
+  name: phylogenize
+  version: "{{ version }}"
+
+source: 
+  url: https://github.com/pbradleylab/phylogenize/archive/refs/tags/v{{ version }}.tar.gz
+  build:
+    name: 0
+    rpaths:
+      - lib/R/lib/
+      - lib/
+    noarch: generic
+
+requirements:
+  build:
+    - vsearch
+    - bioconductor-ggtree
+    - bioconductor-biomformat
+    - r-devtools
+    - r-ragg
+    - r-phylolm
+    - r-phangorn
+  run:
+    - vsearch
+    - bioconductor-ggtree
+    - bioconductor-biomformat
+    - r-devtools
+    - r-ragg
+    - r-phylolm
+    - r-phangorn
+  test:
+    commands:
+      - '$R -c "library(''{{ name }}'')"'
+
+about:
+  home: https://github.com/bitbucket/phylogenize
+  license: MIT
+  summary: "An R package that allows users to link microbial genes to environments, accounting for phylogeny"
+  summary: "Phylogenize is a tool that allows users to link microbial genes to environments, accounting for phylogeny. More specifically, given community composition data, Phylogenize links genes in microbial genomes to either microbial prevalence in, or specificity for, a given environment, while also taking into account an important potential confounder: the phylogenetic relationships between microbes."
+  extra:
+    recipe-maintainers:
+      - kathrynkananen
+    indentifiers:
+      - doi: https://doi.org/10.1093/bioinformatics/btz722
--- a/recipes/phylogenize/meta.yaml
+++ b/recipes/phylogenize/meta.yaml
@@ -21,25 +21,25 @@ build:
 
 requirements:
   host:
-    - perl 5.32.1
-    - r-base 4.3.1
-    - r-devtools 2.4.5 
-    - vsearch 2.24.0
-    - r-ragg 1.2.5
-    - r-phylolm 2.6.2
-    - r-phangorn 2.11.1
-    - 'bioconductor-ggtree 3.8.0'
-    - 'bioconductor-biomformat 1.28.0'
+    - perl =5.32.1
+    - r-base =4.3.1
+    - r-devtools <=2.4.5 
+    - vsearch =2.24.0
+    - r-ragg <=1.2.6
+    - r-phylolm =2.6.2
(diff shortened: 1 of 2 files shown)


recipes/phylogenize/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+  rpaths:
+    - lib/R/lib/
+    - lib/
+  noarch: generic
+  run_exports:
+    - {{ pin_subpackage(name, max_pin="x.x") }}
```
thanhleviet, line 20: `{{ pin_subpackage("phylogenize", max_pin="x") }}`

recipes/phylogenize/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - $R -e "devtools::install_github('pbradleylab/phylogenize/package/phylogenize')"
+    - '$R -e "?phylogenize"'
+
+about:
+  home: https://github.com/pbradleylab/phylogenize
+  license: MIT
+  summary: "Phylogenize is a tool that allows users to link microbial genes to environments, accounting for phylogeny."
+extra:
+  recipe-maintainers:
+    - kathryn kananen
```
thanhleviet, line 56: `Kekananen`

recipes/phylogenize/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  rpaths:
+    - lib/R/lib/
+    - lib/
+  noarch: generic
+  run_exports:
+    - {{ pin_subpackage(name, max_pin="x.x") }}
+
+requirements:
+  host:
+    - perl
```
thanhleviet, line 24: Do the dependencies need their version pinned?

recipes/phylogenize/meta.yaml
Kekananen: I think the name variable from line 3 will sub into the pinning right?

recipes/phylogenize/meta.yaml
Kekananen: Ah likely yes, if nothing else but to make it more robust! Thanks.

recipes/phylogenize/meta.yaml
Kekananen: Fixed

recipes/phylogenize/meta.yaml
Kekananen: Packages pinned to specific version.

recipes/phylogenize/meta.yaml
thanhleviet: nope, see https://bioconda.github.io/contributor/linting.html#missing-run-exports

recipes/phylogenize/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - lib/R/lib/
+    - lib/
+  noarch: generic
+  run_exports:
+    - {{ pin_subpackage(name, max_pin="x.x") }}
+
+requirements:
+  host:
+    - perl 5.32.1
+    - r-base 4.3.1
```
thanhleviet, line 25: It should be a pinning expression, e.g `r-base >=4.3.1`, see a full example https://github.com/bioconda/bioconda-recipes/blob/master/recipes/bioconductor-a4classif/meta.yaml

recipes/phylogenize/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - 'bioconductor-biomformat 1.28.0'
+    - r-devtools 2.4.5
+    - r-ragg 1.2.5
+    - r-phylolm 2.6.2
+    - r-phangorn 2.11.1
+
+test:
+  commands:
+    - '$R -e "library(devtools)"'
+    - $R -e "devtools::install_github('pbradleylab/phylogenize/package/phylogenize')"
```
thanhleviet, line 47: single quote missing?

recipes/phylogenize/meta.yaml
Kekananen: Yes, I believe it says in there that you can use the name variable.

"Case 1: If the software follows semantic versioning (or it has at least a normal version string (like 1.2.3) and the actual strategy of the devs is unknown), add run_exports to the recipe like this:

build:
  run_exports:
    - {{ pin_subpackage('myrecipe', max_pin="x") }}
with myrecipe being the name of the recipe (you can also use the name variable). This will by default pin the package to >=1.2.0,<2.0.0 where 1.2.0 is the version of the package at build time of the one depending on it and <2.0.0 constrains it to be less than the next major (i.e. potentially not backward compatible) version."

recipes/phylogenize/meta.yaml
Kekananen: Not finding it, so I think we're good.

recipes/phylogenize/meta.yaml
thanhleviet: Ah, my bad, sorry, I was not aware of that.

recipes/phylogenize/meta.yaml
Kekananen: Should be good now

Kekananen:
@BiocondaBot please add label

thanhleviet (approved):

PR #43998, opened 2023-10-31
Add omamer 

alpae:
This PR adds OMAmer (https://doi.org/10.1093/bioinformatics/btab219) to bioconda.

### DIFF
--- a/recipes/omamer/meta.yaml
+++ b/recipes/omamer/meta.yaml
@@ -15,6 +15,8 @@ build:
   noarch: python
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
   number: 0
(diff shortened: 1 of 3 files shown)


recipes/omamer/meta.yaml
```diff
@@ -0,0 +1,63 @@
+{% set name = "omamer" %}
```
thanhleviet, line 1: Please separate each recipe as a PR

alpae: ok, but then I guess I will need to add one after each other so that the dependencies can be resolved?

alpae: (omamer depends on PySAIS and rmath4)

PR #44121, opened 2023-11-06
MultiQC: add missing dependencies, pin to Python >=3.8

vladsavelyev:
After version 1.17, MultiQC stopped supporting Python 3.6 and 3.7, and added the `packaging` dependency to handle the calls of the deprecated `distutils` that was removed in 3.12: https://github.com/ewels/MultiQC/pull/2113

However, we forgot to updated the bioconda recipe accordingly. Thos PR:
- Pins Python to >=3.8
- Adds the missing `packaging` dependency
- Additionally, adds the `humanize` dependency that is added in the dev version and will be used in the future MultiQC 1.18 version.

Addresses https://github.com/nextflow-io/rnaseq-nf/issues/22

### DIFF
--- a/recipes/multiqc/meta.yaml
+++ b/recipes/multiqc/meta.yaml
@@ -22,7 +22,7 @@ requirements:
     - python >=3.8
     - pip
   run:
-    - python >=3.8,<3.12
+    - python >=3.8
     - setuptools
     - click
     - coloredlogs
--- a/recipes/multiqc/meta.yaml
+++ b/recipes/multiqc/meta.yaml
@@ -19,10 +19,10 @@ build:
 
 requirements:
   host:
-    - python >=3.8
+    - python >=3.8,<3.12
     - pip
   run:
-    - python >=3.8
+    - python >=3.8,<3.12
     - setuptools
     - click
     - coloredlogs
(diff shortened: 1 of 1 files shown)


marcodelapierre:
@vladsavelyev  based on this issue of mine  : https://github.com/nextflow-io/rnaseq-nf/issues/22

I would suggest adding:
- dependency on `importlib-metadata` (for `multiqc >= 1.17`)
- constraining `python` to `< 3.12`

vladsavelyev:
Good points, @marcodelapierre! Thanks a lot for the addition :)

recipes/multiqc/meta.yaml
```diff
@@ -19,20 +19,23 @@ build:
 
 requirements:
   host:
-    - python >=3.6
+    - python >=3.8,<3.12
```
ewels, line 22:
```suggestion
    - python >=3.8
```

ewels (commented):
> * constraining `python` to `< 3.12`

I don't think that there is any need for this, as long as the missing dependencies have been added?

marcodelapierre:
> > * constraining `python` to `< 3.12`
> 
> I don't think that there is any need for this, as long as the missing dependencies have been added?

yes you are right Phil

ewels (approved):
LGTM, thanks 👍🏻

PR #44182, opened 2023-11-09
Update Pantools v4.2.3 recipe

dirkjanvw:
Previously, I had assumed that dependencies in the `run` section have to be valid for both Linux and macOS. I have just found out that these can be platform specific too. Therefore, this PR makes two previously excluded dependencies Linux specific since they are not conflicting on Linux. They are still conflicting on macOS and therefore the `# [linux]` has been added for them. This will also make sure that these dependencies are included for the PanTools image on biocontainers.

### DIFF
--- a/recipes/pantools/meta.yaml
+++ b/recipes/pantools/meta.yaml
@@ -17,7 +17,7 @@ build:
     - {{ pin_subpackage("pantools", max_pin="x") }}
 
 requirements:
-  build:
+  host:
     - openjdk =8
     - maven
     - jq
--- a/recipes/pantools/meta.yaml
+++ b/recipes/pantools/meta.yaml
@@ -12,7 +12,7 @@ source:
 
 build:
   noarch: generic
-  number: 0
+  number: 1
   run_exports:
     - {{ pin_subpackage("pantools", max_pin="x") }}
 
@@ -32,13 +32,13 @@ requirements:
     - blast
     - mash =2.3
     - fastani
-      #- busco =5  #causes conflicts on macOS
+    - busco =5  # [linux]
     - r-base >=3.5.0
     - r-ggplot2
     - r-ape
     - graphviz
     - aster =1.3
-      #- bcftools >=1.12  #causes conflicts on macOS
+    - bcftools >=1.12  # [linux]
     - tabix
 
 test:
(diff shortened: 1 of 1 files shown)


dirkjanvw:
@bioconda/core I cannot use the OSX artifacts from the pipeline to check that it works; can you please confirm that it's true what I'm trying here? that the `# [linux]` in the `run` section works as expected and these dependencies are not required on osx?

recipes/pantools/meta.yaml
```diff
@@ -32,13 +32,13 @@ requirements:
     - blast
     - mash =2.3
     - fastani
-      #- busco =5  #causes conflicts on macOS
+    - busco =5  # [linux]
```
bgruening, line 35: What kind of conflicts does it cause? Isn't busco an essential dependency?

bgruening (commented):
What is currently in the build section can be moved into the host section I think.

recipes/pantools/meta.yaml
dirkjanvw: BUSCO is essential only for one subcommand of PanTools luckily which likely no one runs on OSX systems, but it seems like the conflict is related to openMP differences between Linux and OSX (this causes conflicts between BUSCO, fastani, kmc and python). However, I am not familiar enough with openMP myself to fix this.

dirkjanvw:
Moving the current `host` section to a `build` section did not give any errors, but I still can't test the OSX artifacts locally; it is essentially empty.

bgruening:
OSX tests will run here on CI. Maybe that helps you to figure out what is happening with OpenMP? https://conda-forge.org/docs/maintainer/knowledge_base.html#openmp

dirkjanvw:
So these are two different things:
1. OSX artifact: I know where to find the artifacts in the CI pipeline over at Azure. However, the Linux artifact works just fine when I unzip it and do `mamba create -n test_pantools -d -c ./LinuxArtifacts/packages pantools`; but for the OSX artifact it doesn't work because the artifact is empty and doesn't contain the new information for pantools.
2. OpenMP conflict: I debugged this and found out that the issue is likely FastANI which specifies `openmp` instead of `llvm-openmp` for OSX. Do you think this can be true? Because if so, then I would like to change this for the FastANI recipe.

bgruening:
We could rebuild FastANI if this is the bottleneck.

dirkjanvw:
I created a PR for FastANI here: https://github.com/bioconda/bioconda-recipes/pull/44241 But I already see that we will only be able to use it in a future version where we also change our KMC dependency. The KMC dependency version is highly critical for PanTools and can't be changed easily, but we will update this in the future.

dirkjanvw:
FastANI has been rebuild so we can use that for a future release; then the only question remaining is, how do I obtain the correct artifacts zip file for OSX so that I can test it?

dirkjanvw:
We discussed internally and in a following version of PanTools we will fix all conflicting dependencies, so for now we would like this temporary fix to be in bioconda (basically giving busco and bcftools as dependencies for linux only, so that the biocontainer contains both at least)

dirkjanvw:
@BiocondaBot please add label

bgruening (approved):

PR #43995, opened 2023-10-31
Bulk arm builds

aliciaaevans:
Port .github/workflows/bulk.yml to Circle CI for ARM builds.

To Do:

- [x] determine what's needed to commit as bioconda-bot
- [x] test

*Edited by @daler 2023-11-26*. **Note to future readers:**  this PR has become a sort of sandbox for all things ARM. There is no intention that this should ever be merged. Instead, working parts should be cherry-picked or manually ported over to the `arm-builds` branch, https://github.com/bioconda/bioconda-recipes/pull/40550

PR #44245, opened 2023-11-13
CircleCI workflow for Bulk builds on ARM

aliciaaevans:
Port .github/workflows/bulk.yml to Circle CI for ARM builds.

Notes:

- Tested the workflow except for `--anaconda-upload` on https://github.com/bioconda/bioconda-recipes/pull/43995 but wanted to have a less cluttered branch.
- Holding off for now from `--mulled-upload-target biocontainers` per discussion with the core team.
- Used `machine` instead of docker for the `executor` because "nested" docker with mounted volumes is [not supported by CircleCI](https://circleci.com/docs/using-docker/#docker-benefits-and-limitations).
- Do we want to use a branch other than `bulk` to do the first bulk update?
- Not sure if there's anything additional needed to test `--anaconda-upload` to make sure `linux-aarch64` does not conflict with `linux-64`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,10 @@ executors:
     docker:
       - image: cimg/base:2023.06
     resource_class: arm.medium
+  linux-aarch64-bulk:
+    machine:
+      image: ubuntu-2004:current
+    resource_class: arm.medium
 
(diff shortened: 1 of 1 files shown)


martin-g (approved):
LGTM

PR #44369, opened 2023-11-21
update iphop to 1.3.3, and addition of run_exports

simroux:
Updating iphop to 1.3.3, had to modify the meta.yaml to now include a run_exports clause. Using the standard one as the versioning follows standard numbering.


simroux:
Dependencies issues, must fix before new pull request

PR #44975, opened 2023-12-23
mcl: restore bioinformatics tools support

0xaf1f:
The last mcl release mistakenly excluded the `mcxdeblast` (and `mclblastline`) scripts. The developer confirmed this and promised to include it in the next release. Until that happens, the bioconda package can bring them back. This follows the approach by the Debian package maintainers that the developer recommended.

Because of the mistake, the last update of the mcl package broke several packages:
- https://github.com/nf-core/modules/issues/3828
- https://github.com/sanger-pathogens/Roary/issues/217
- as well as `hybran`

See https://github.com/micans/mcl/discussions/25

### DIFF
--- a/recipes/mcl/build.sh
+++ b/recipes/mcl/build.sh
@@ -28,3 +28,8 @@ cd ..
 make -j"${CPU_COUNT}"
 make install
 make clean
+
+# Add back in the mcl/blast scripts.
+# Remove this once the next release re-incorporates them.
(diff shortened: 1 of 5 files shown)


0xaf1f:
@BiocondaBot please add label

mencian:
LGTM, thanks for looking into this @0xaf1f!

mencian (approved):

PR #44991, opened 2023-12-27
Revert "Update keggcharter to 1.0.1"

iquasere:
Reverts bioconda/bioconda-recipes#44961


iquasere:
I'm trying to fix the update to KEGGCharter, since the meta.yaml still required a new command, but it was merged before I could do it. I'll just release another version, it's simpler.

PR #45116, opened 2024-01-08
Update perl-ms: add run_exports, bump version

jvolkening:
Adds run_exports to meta.yml; bump version


jvolkening:
@BiocondaBot please add label

kbseah (changes requested):
looks good, tests pass, but could you please use the SPDX abbreviation for the license, either `GPL-3.0-only` or `GPL-3.0-or-later`, whichever is correct?

kbseah (approved):

PR #45161, opened 2024-01-10
Update igblast to 1.22.0

ressy:
Updates igblast from 1.21.0 to 1.22.0.

I updated the `shared_vdb.patch` file to account for a now-wrapped line in the original source, but other than confirming that it should patch correctly I don't know any details about that patch.  I also defined `run_exports` in the `meta.yaml` to use `max_pin="x"` following the [linter's advice](https://bioconda.github.io/contributor/linting.html#missing-run-exports) (as far as I can tell igblast seems to follow semantic versioning).

### DIFF
--- a/recipes/igblast/build.sh
+++ b/recipes/igblast/build.sh
@@ -13,37 +13,17 @@ mkdir -p $PREFIX/bin
 mkdir -p $SHARE_DIR/bin
 
 if [[ $(uname) == Linux ]]; then
-    export CPPFLAGS="$CPPFLAGS -I$PREFIX/include/ncbi-vdb"
-    export CC_FOR_BUILD=$CC
     export AR="$AR rcs"
 
     cd c++
 
-    # IgBLAST is based on the BLAST source code and building it produces
-    # a similar set of shared libraries if --with-dll is used. To avoid
-    # conflicts when installing IgBLAST and BLAST simultaneously,
-    # we link IgBLAST statically.
-    ./configure.orig \
-        --with-static-exe \
-        --with-mt \
-        --with-openmp \
-        --without-autodep \
-        --without-makefile-auto-update \
-        --with-flat-makefile \
-        --with-caution \
-        --without-lzo \
-        --without-debug \
-        --with-strip \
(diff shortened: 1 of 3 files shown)


ggabernet:
@ressy any idea what is wrong, it would be great to have the updated recipe :D

ressy:
> @ressy any idea what is wrong, it would be great to have the updated recipe 😄

No, unfortunately not!  I keep meaning to take a closer look but haven't had a chance yet.  I can build the existing bioconda igblast package locally, and when I try to build the new version locally I seem to get the same error the azure test gets, but haven't figured out what's actually failing.

marcelm:
I tried to look into this, but probably have to give up. Some notes follow.

Since this recipe is based on the BLAST recipe, I tried to run `./configure` in the same way in which it is done in the BLAST recipe. In particular
- I switched to `make -j1` from `make -j2`
- I tried to run the "flat" Makefile with `make -j1 -f Makefile.flat igblastn.exe igblastp.exe makeblastdb.exe` in the hope that it would only attempt to build the parts really necessary for igblast.

The error I currently get is this one:
```
make[3]: *** No rule to make target '.../conda-bld/debug_1708865365570/work/c++/GCC1320-ReleaseMTDLL64/status/.seqdb.dep', needed by 'libalign_format.so'.  Stop.
```
I tried to just `touch` that file, but then get linker errors.

I thought that maybe providing the programs to build via the `igblastn.exe igblastp.exe makeblastdb.exe` arguments is not the way to go, but when I omit it and restart `make`, I get a similar error about `.../status/.blastinput-static.dep`.

The next thing to try IMO would be to naively run `./configure && make`. If that works, I’d then slowly add the configure options from the build script to see which one is problematic.

Annoyingly, `ccache` doesn’t work, maybe because the configure script injects `-DNCBI_BUILD_SESSION_ID=ce669e3e-cd94-4317-82e0-f395bc8b317e`, which I presume changes for every `configure` invocation.

marcelm:
Wow, I cannot believe this is actually working now ...

marcelm (approved):

PR #45214, opened 2024-01-14
Update deepvariant dependency versions

kdm9:
The deepvariant package pins out-of-date tensorflow versions. The upstream built against 2.11, and we pin 2.0.XXX which is no longer available, making the package uninstallable.

Also added a run-exports section

### DIFF
--- a/recipes/deepvariant/meta.yaml
+++ b/recipes/deepvariant/meta.yaml
@@ -19,7 +19,7 @@ source:
   sha256: '{{ sha256 }}'
 
 build:
-  number: 0
+  number: 1
   skip: true  # [osx or not py36]
 
 requirements:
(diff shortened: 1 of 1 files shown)


kdm9:
By the way, I had contemplated updating this to use deepvariant 1.6.0, but it seems like they no longer provide deepvariant.zip that contains the precompiled innards of deepvaraint. @bgruening, I'm not sure how to continue providing new deepvariant versions via conda, seems the upstream authors are all-in on just giving folks a docker image. Thoughts?

bgruening:
> we pin 2.0.XXX which is no longer available

Do you know why it's not available anymore?

bgruening (approved):

bgruening:
> I'm not sure how to continue providing new deepvariant versions via conda, seems the upstream authors are all-in on just giving folks a docker image. 

All GPU-based tools are a bit complicated to provide packages for. conda-forge went to a large pain to provide some recipes. It's probably a lot of work. So I can understand that its easier to provide Docker containers - I would still argue for properly designed recipes and metadata descriptions. But there is only so much we can do :(

PR #45309, opened 2024-01-19
Update recipe MS2Query

niekdejonge:
The dependencies for MS2Query 1.5 have changed, therefore I have updated the recipe here.


recipes/ms2query/meta.yaml
```diff
@@ -21,19 +21,19 @@ build:
```
martin-g, line 21: I think you will need to bump `build > number` (line14) too

niekdejonge:
@BiocondaBot please add label

recipes/ms2query/meta.yaml
niekdejonge: Thanks. I tried, but the linting did not pass if I set build number to 1.

martin-g (approved):

PR #45376, opened 2024-01-23
Add Acanthophis: a comprehensive plant (holo)genomics pipeline

kdm9:
This adds Acanthophis, a snakmake pipeline for population genomics of plants, and/or their microbes.


gbouras13:
There's an issue with snakemake minimal - I'll have a look into it

kdm9:
the issue is that it seems the builder pins python=3.10, but snakemake requires 3.11. I would have thought that pulling in snakemake would upgrade python to 3.11, but seems not?!?

kdm9:
@bioconda-bot please update

PR #45488, opened 2024-01-30
Update Gubbins

nickjcroucher:
Update of Gubbins to version 3.3.3


nickjcroucher:
@BiocondaBot please review & merge

recipes/gubbins/meta.yaml
```diff
@@ -28,28 +28,27 @@ requirements:
...
+    - multiprocess
     - setuptools
     - raxml =8.2.12
     - iqtree >=2.2
     - rapidnj
     - raxml-ng =1.0.1
     - fasttree =2.1.10
-    - numpy >=1.2
-    - numba >=0.56
+    - numpy <=1.23.0
```
martin-raden, line 50: you dropped the numpy version restriction above for `host:` so is it needed?

nickjcroucher: Well spotted, thanks - I will specify the version for `host` as well for the time being, as that will hopefully consistently avoid the problematic numpy versions, but hopefully we will soon be able to provide a >= constraint.

martin-raden (approved):

Issue #45164, opened 2024-01-10
Interoperability issues with anndata/scanpy-scripts


neoformit:
The current build of `scanpy-scripts` has packaged `anndata==0.7.6`. This is causing issues in downstream Galaxy tools, since `ebi-gxa/anndata_ops` requires `scanpy-scripts` and gets this old `anndata` version. This means that it does not play nicely with other tools in the anndata suite, for example:

```
iuc/anndata_manipulate -> anndata==0.10.3
ebi-galaxy/anndata_ops -> anndata==0.7.8
```

An example of "does not play nicely" is that the output of `anndata_manipulate` will produce an error if read by `anndata_ops`, like:

```
Traceback (most recent call last):
  File "/mnt/tmp/job_working_directory/007/724/7724020/configs/tmp1jvzx33z", line 7, in <module>
    adata = sc.read('input.h5')
  File "/usr/local/lib/python3.9/site-packages/scanpy/readwrite.py", line 112, in read
    return _read(
  File "/usr/local/lib/python3.9/site-packages/scanpy/readwrite.py", line 713, in _read
    return read_h5ad(filename, backed=backed)
  File "/usr/local/lib/python3.9/site-packages/anndata/_io/h5ad.py", line 421, in read_h5ad
    d[k] = read_attribute(f[k])
  File "/usr/local/lib/python3.9/functools.py", line 877, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
  File "/usr/local/lib/python3.9/site-packages/anndata/_io/utils.py", line 183, in func_wrapper
    raise AnnDataReadError(
anndata._io.utils.AnnDataReadError: Above error raised while reading key '/layers' of type <class 'h5py._hl.group.Group'> from /.
```

`ebi-galaxy/anndata_ops` uses the latest `scanpy-scripts==1.1.6`, but the `anndata` version in `scanpy-scripts` is not pinned:

https://github.com/bioconda/bioconda-recipes/blob/8a3900e6e5bdbf7bae2c166a619e0eb2af4361bf/recipes/scanpy-scripts/meta.yaml#L50-L51

- I'm not sure why this would lead to such an old `anndata` version, is it pinned in another dependency?
- Would a rebuild of `scanpy-scripts` pull a newer `anndata` version?
- Is there a way that we can solve the downstream problem, perhaps by declaring `anndata>=0.8.0` as a requirement in the Galaxy tool?

swbioinf:
[...]

PR #45534, opened 2024-02-01
Update scanpy-scripts: require anndata >=0.8.0 (previously unversioned)

swbioinf:
Added a requirement of AnnData >=0.8.0 (previously unversioned) for scanpy-scripts. I have incremented build version from 0=>1.

AnnData objects saved with > 0.8.0 (Mar 2022) are not backwards compatible with objects saved with current AnnData versions (e.g. current 0.10.5) https://anndata.readthedocs.io/en/latest/release-notes/index.html#version-0-8
Scanpy and scanpy-scripts will work in isolation with either version, but when data objects are shared between tools, issues arise. 

I expect for most users/builds this will not affect much about the build (since it should default to the latest anndata), but in the case where an old version of anndata is available in a higher priority channel, this will mean a current version is sought. This particuar edge case has been seen in galaxy tool builds via biocontainers; See https://github.com/bioconda/bioconda-recipes/issues/45164 and https://github.com/ebi-gene-expression-group/container-galaxy-sc-tertiary/issues/311 for details.

Its been a couple of years now, so I anticipate that there will not be many people wanting to stick with the old library.   Is it fair to assume they can seek the previous build version (0) of this recipe if required? This is my first recipie update, so if I should be seeking a different aproach to this pull request please advise and I'll check it out. Thanks.

### DIFF
--- a/recipes/scanpy-scripts/meta.yaml
+++ b/recipes/scanpy-scripts/meta.yaml
@@ -16,7 +16,7 @@ source:
       - scrublet.patch
 
 build:
-  number: 0
+  number: 1
   noarch: python
   entry_points:
     - scanpy-cli=scanpy_scripts.cli:cli
--- a/recipes/scanpy-scripts/meta.yaml
+++ b/recipes/scanpy-scripts/meta.yaml
@@ -48,7 +48,7 @@ requirements:
     - pytoml
     - packaging
   run:
-    - anndata
+    - anndata>=0.8.0
     - bbknn>=1.5.0
     - click<8
     - fa2
--- a/recipes/scanpy-scripts/meta.yaml
+++ b/recipes/scanpy-scripts/meta.yaml
@@ -18,6 +18,8 @@ source:
 build:
   number: 1
   noarch: python
+  run_exports:
+    - {{ pin_subpackage('scanpy-scripts', max_pin="x.x") }}
   entry_points:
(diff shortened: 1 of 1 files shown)


pcm32:
Please do not merge this, as it might create a new container on the same version of scanpy-scripts which will break most of our setup. We are testing the update of scanpy scripts to 1.9.x scanpy which will include a newer AnnData.

pcm32:
To clarify, these tests:

https://github.com/ebi-gene-expression-group/scanpy-scripts/blob/develop/scanpy-scripts-tests.bats

are not tested in bioconda and they need to pass. A fresh install of the bioconda recipe as it is just changing the anndata pin doesn't pass these tests. We have now merged changes on scanpy scripts to use a newer Scanpy (and drag with that a newer AnnData) (see https://github.com/ebi-gene-expression-group/scanpy-scripts/pull/117)

pcm32:
@swbioinf if you want to help us, I'm attaching the planemo report with changes needed on the Galaxy wrappers for Scanpy 1.9.x. in this PR https://github.com/ebi-gene-expression-group/container-galaxy-sc-tertiary/pull/313 - just declare on which tools you will work and contribute to the branch on that PR. Thanks!

swbioinf:
Hi @pcm32 No worries - no I'm discontinuing this particular pull request (will close).  I started this and realised when I hit the pinned version testing errors that it was (of course) not so simple! Thanks.

PR #45648, opened 2024-02-06
metabuli1.0.3

jaebeom-kim:
Metabuli 1.0.3
- New parameter: --tie-ratio in classify module. [default 0.95]
When the best matching species has a score MAX, species with score >= (MAX * --tie-ratio) is considered as a tie to the best score. When tie species occur for a read, the read is classified into their LCA.


mencian:
LGTM, thanks @jaebeom-kim!

PR #45649, opened 2024-02-06
Add Scevan R package recipe

andreiprodan:
Recipe for GitHub Scevan R package


recipes/scevan/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  number: 0
+  noarch: generic  # Specify that the package is noarch
+  run_exports: '{{ pin_compatible("SCEVAN", max_pin="x.x") }}'
+  rpaths:
+    - lib/R/lib
+    - lib
+  script: "$R -e \"install.packages(c('Rcpp', 'remotes', 'devtools'), lib='$PREFIX/lib/R/library', repos='https://cran.rstudio.com/')\""
+
+requirements:
+  build:
```
martin-g, line 19: I think this should be `host`, not `build`.
R, as Python, is an interpreted language, so these dependencies should be listed in `requirements>host`

recipes/scevan/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - bioconductor-ggtree
+    - r-tidytree
+    - r-ggrepel
+    - r-optparse
+
+test:
+  commands:
+    - $R -e "library(devtools)"
+    - $R -e "install_github('miccec/yaGST')"
+    - $R -e "devtools::install_github('AntonioDeFalco/SCEVAN')"
```
martin-g, line 48: Hm!
I think is not correct!

1) This will install the latest code from the default branch. I think you need to specify the tag, e.g. `, ref = "1.0.0"`.
2) Why do you install it with devtools from Github at all ? IMO you should test the locally installed library instead

recipes/scevan/meta.yaml
andreiprodan: will edit, thanks!

recipes/scevan/meta.yaml
andreiprodan: will add the tag!

recipes/scevan/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  run:
+    - r
+    - r-base
+    - r-devtools
+    - r-paralleldist
+    - r-pheatmap
+    - r-forcats
+    - r-dplyr
+    - bioconductor-fgsea
+# - r-yaGST - manually with devtools
```
martin-g, line 33: Why this dependency is left to be (pre-)installed by the user ?
I think it should be a proper runtime dependency and lines https://github.com/bioconda/bioconda-recipes/pull/45649/files#diff-180caf3dc4ce15be03b03267239715f1249fa8f1d7e031a38dc7624e5189ef07R46-R48 / https://github.com/bioconda/bioconda-recipes/pull/45649/files#diff-180caf3dc4ce15be03b03267239715f1249fa8f1d7e031a38dc7624e5189ef07R27 could be removed

martin-g: You may need a build script like https://github.com/bioconda/bioconda-recipes/pull/45645/files#diff-0d601fcf99e217658acf5443fb7fb9fe3d9691742cee85bccdb60bcfbdb200acR2 to install the library locally and test it without `devtools`

andreiprodan: could not a way to install "yaGST" other than with `devtools::install_github()`. "yaGST" is not on CRAN or Bioconductor or conda

martin-g: I see!
Maybe a recipe should be added for `yaGST` first ...
I'll let the Bioconda maintainers to decide what would be the best!

pcm32: I would suggest that this gets shipped as it is and eventually when someone adds that additional recipe, then it could be added as conda dependency here? Otherwise it is harsh on contributors to have to chase every single dependent package. There are other packages that do installations of deps via devtools.

pcm32:
With the above artifact @andreiprodan you could test locally if the code on the SCEVAN Readme works well.

recipes/scevan/meta.yaml
```diff
@@ -0,0 +1,56 @@
+package:
+  name: scevan
```
pcm32, line 2: I think that the package name should have `r-` appended.
```suggestion
  name: r-scevan
```

recipes/scevan/meta.yaml
pcm32: This tag is fairly old though, around a year. I wonder if we can ask the maintainer to make a new release.

andreiprodan: updated to 1.0.1

pcm32:
If the bot doesn't give the artifacts, they can be normally located through CI runs, for instance previous one to your last commit is here: https://dev.azure.com/bioconda/bioconda-recipes/_build/results?buildId=50666&view=artifacts&pathAsName=false&type=publishedArtifacts

pcm32:

recipes/r-scevan/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  noarch: generic  # Specify that the package is noarch
+  run_exports: '{{ pin_compatible("SCEVAN", max_pin="x.x") }}'
+  rpaths:
+    - lib/R/lib
+    - lib
+  script: "$R -e \"install.packages(c('Rcpp', 'remotes', 'devtools'), lib='$PREFIX/lib/R/library', repos='https://cran.rstudio.com/')\""
+
+requirements:
+  host:
+    - r
```
bgruening, line 20:
```suggestion
```

recipes/r-scevan/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - r-base
+    - r-remotes
+    - r-devtools
+  run:
+    - r
```
bgruening, line 25:
```suggestion
```

recipes/r-scevan/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+test:
+  commands:
+    - $R -e "library(devtools)"
+    - $R -e "devtools::install_github('miccec/yaGST')"
+    - $R -e "devtools::install_github('AntonioDeFalco/SCEVAN', ref = 'v1.0.1')"
+    - $R -e "library(SCEVAN)"
+
+about:
+  home: https://github.com/AntonioDeFalco/SCEVAN/
+  license: GNU GPLV3
```
bgruening, line 53: please use SPDX identifiers

andreiprodan: updated license to 'GPL-3.0-only' identifier

andreiprodan:

cbrueffer:
Please remove the `LICENSE` file, you can point to the one installed by `r-base` instead: `license_file: '{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-3'`

cbrueffer (changes requested):

recipes/r-scevan/meta.yaml
```diff
@@ -0,0 +1,54 @@
+package:
+  name: r-scevan
+  version: 1.0.1
+
+source:
+  url: https://github.com/AntonioDeFalco/SCEVAN/archive/refs/tags/v1.0.1.tar.gz  # Replace with the GitHub URL of the release
```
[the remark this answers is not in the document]
andreiprodan: comments have been removed

recipes/r-scevan/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  sha256: 7558d31282c36418db3960b3edb9cb9815db14b0a48d8882287e11f752dfeab1       
+
+build:
+  number: 0
+  noarch: generic  # Specify that the package is noarch
+  run_exports: '{{ pin_compatible("r-scevan", max_pin="x.x") }}'
+  rpaths:
+    - lib/R/lib
+    - lib
+  script: "${R} CMD INSTALL --build . "
```
pcm32, line 16: This is now missing `miccec/yaGST` probably.

recipes/r-scevan/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - bioconductor-ggtree
+    - r-tidytree
+    - r-ggrepel
+    - r-optparse
+
+test:
+  commands:
+    - $R -e "library(devtools)"
+    - $R -e "devtools::install_github('miccec/yaGST')"
+    - $R -e "devtools::install_github('AntonioDeFalco/SCEVAN', ref = 'v1.0.1')"
```
pcm32, line 46: I suspect that these shouldn't be in the tests... yaGST needs to be installed in the script part.

pcm32:
For some reason the build is again re-installing the dependencies from R when doing R BUILD of the SCEVAN, I remember there is a way of telling R to build ignoring dependencies, as in our case they are all installed either by devtools or by conda.

recipes/r-scevan/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - bioconductor-fgsea
+    - r-cluster
+    - r-ggplot2
+    - r-Rtsne
+    - bioconductor-scran
+    - r-ape
+    - bioconductor-ggtree
+    - r-tidytree
+    - r-ggrepel
+    - r-optparse
```
pcm32, line 39: Ok, the issue I think is that all of this are missing in host AFAIK, that is why is trying to install again all these dependencies.

pcm32: putting these in `host:` means that when it reaches the R CMD INSTALL --build for SCEVAN, then it won't try again to install them. Sorry, I hadn't paid attention to `run` and `host` sections. They also need to be present in run of course.

recipes/r-scevan/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  sha256: 7558d31282c36418db3960b3edb9cb9815db14b0a48d8882287e11f752dfeab1       # Generate the checksum using 'sha256sum' command
+
+build:
+  number: 0
+  noarch: generic  # Specify that the package is noarch
+  run_exports: '{{ pin_compatible("SCEVAN", max_pin="x.x") }}'
+  rpaths:
+    - lib/R/lib
+    - lib
+  script: "$R -e \"install.packages(c('Rcpp', 'remotes', 'devtools'), lib='$PREFIX/lib/R/library', repos='https://cran.rstudio.com/')\""
```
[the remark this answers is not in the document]
andreiprodan: done (in a separate build.sh script)

recipes/r-scevan/meta.yaml
andreiprodan: added miccec/yaGST installation via devtools to build.sh script

recipes/r-scevan/meta.yaml
andreiprodan: removed "install_github('AntonioDeFalco/SCEVAN')"; now installed with "R CMD INSTALL --build . "

recipes/r-scevan/meta.yaml
andreiprodan: fixed now

pcm32:
Let me try the artifact.

pcm32:

pcm32:
Great work @andreiprodan , and thanks for the perseverance!

- Artifact installs correctly
- This [example](https://htmlpreview.github.io/?https://github.com/AntonioDeFalco/SCEVAN/blob/main/vignettes/IntratumoralHeterogeneityInGlioblastoma.html) runs to completion, without errors and the figures generated look the same as in the example.

On my side I'm happy with this as it is.

pcm32 (approved):
LGTM and I have tested the artifact to work with real examples.

pcm32:
@BiocondaBot please add label

pcm32:
@cbrueffer I think that your requests have been addressed.

cbrueffer (approved):

cbrueffer:
Nice job everyone!

PR #45780, opened 2024-02-13
MultiQC: pin Plotly version

vladsavelyev:
Pin Plotly version tag (to address issues like https://github.com/MultiQC/MultiQC/issues/2323) and bump build number.


ewels (approved):
LGTM, thanks!

PR #45821, opened 2024-02-15
Add initial RRIkinDP recipe

mwaldl:
Add recipe for RRIkinDP - a tool for evaluating thermodynamic and kinetic features of RNA-RNA interactions.


s-will (approved):
looks good :)

I see no urgent need (if at all) to fix the minor details for this build

PR #45954, opened 2024-02-22
update bio digest recipe

lspindler2509:

----

### General instructions

```yaml
build:
  run_exports:
    - ...

```

with `...` being one of:

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

</details>


lspindler2509:

lspindler2509:

lspindler2509:

lspindler2509:

lspindler2509:

recipes/biodigest/meta.yaml
```diff
@@ -10,16 +10,21 @@ source:
...
-  number: 0
-  script: "{{ PYTHON }} -m pip install . -vvv --no-deps --no-build-isolation"
+  number: 1
+  script: 
+    - "{{ PYTHON }} -m pip install . -vvv --no-deps --no-build-isolation"
   noarch: python
   run_exports:
     - {{ pin_subpackage('biodigest', max_pin="x.x") }}
 
+channels:
```
marcelm, line 20: This should not be needed

lspindler2509:

lspindler2509:
@BiocondaBot please add label

marcelm:
Now that you got the `post-link.sh` script working, I wonder whether this is actually the best way to do it (sorry). post-link scripts are a bit annoying (and the Conda documentation says to avoid them if possible), and especially if it’s running after the actual installation and downloading quite a bit of data, I as a user would be wondering what is going on.

Have you considered just downloading the data as part of the build? Then the data would become part of the Conda package itself. The package itself would become larger, but the user is going to have to download the data anyway. If it comes as part of the package, then they will get at least a nice progress bar. Also, with the post-link script, the data is going to be re-downloaded for each time the package is installed while the conda package is going to be cached.

lspindler2509:
@marcelm I once already tried to call the setup step in the build in the script part after the pip install as the following:

 script: 
    - "{{ PYTHON }} -m pip install . -vvv --no-deps --no-build-isolation"
    - conda install -c conda-forge -c bioconda graph-tool
    - echo "Running setup script..."
    - python -c "from biodigest import setup; setup.main(setup_type='api')"
    - echo "Setup done..."
    
Is this the right starting point for it or should I embed it somewhere else?

marcelm:
It looks like the right place, yes. If the build steps get longer, it may make sense to move them to a separate `build.sh` file, but if you reduce it to the `pip install` and `python -c ...` lines, then I think it’s fine to have it directly in `meta.yaml`. I suggest remove the `echo` lines after you’re done debugging.

What is the graph-tool dependency needed for? You should just add it to the `host` dependencies if actually needed.

lspindler2509:
@marcelm  okay thanks for your help. I updated the meta.yaml but when installing the package this way: conda create -n tmp -c $HOME/.conda/conda-bld biodigest python=3.8 I cannot find the created files from the setup step... and when not specifying the python version, it defaults to python 3.11 even though the meta.yaml specifies a version between python3.7-3.9

lspindler2509:
@marcelm  thanks for your help. I actually already had it that way but as now when retrying, I cannot find the created files from the setup script in envs/env/lib/python3.8/site-packages/biodigest/mapping_files unfortunately. 
I built it as before (there the current logging occurred) and then installed the packages like this: 

conda create -n tmp -c $HOME/.conda/conda-bld biodigest

lspindler2509:
when unpacking the build jar in conda-bld, the files are also not present.

marcelm:
The recipe also needs a test that only passes when the downloaded files are present.

marcelm:
Using the above `build.sh` seems to work, at least I can see that the generated package is much larger.

lspindler2509:
It worked for me locally as well and I will now create a test to check for the files as suggested.
Thank you for your help! @marcelm

recipes/biodigest/meta.yaml
marcelm: Did you accidentally re-add this section?

lspindler2509: yeah, the graph-tool import in host fails otherwise in the pipeline, but I will try again after adding the test wether the build also succeeds when deleting the channel section.

marcelm: It definitely works for me and I’m 99% sure it will work on CI as well because conda-forge and bioconda are preconfigured in the channels list already.

Have you maybe not configured conda-forge and/or bioconda in your `~/.condarc` channels list?

marcelm:
The test you added was a bit too specific. It should not test whether the exact files exist, but it should just run the tool. That way, you don’t have to care about exact file names and can instead let the tool figure it out. Here’s a suggestion:
- Add the [`target_gene_cluster.txt`](https://github.com/bionetslab/digest-tutorial/blob/main/input/target_gene_cluster.txt) file from the tutorial.
- Add a [`source_files:` section to the `tests:` section](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#test-files) that ensures that the `target_gene_cluster.txt` file is copied into the test environment.
- Use this as a test: `python -m biodigest.single_validation -t target_gene_cluster.txt -ti symbol -m clustering`

For me, this fails with `Exception: gene_att_mapping.csvdoes not exist. Please run setup.` before running `biodigest.setup.main`, and it succeeds afterwards.

It takes about 7 minutes to run, which is a bit long for a test; maybe you know a way to reduce the runtime.

lspindler2509:
how can I add the file?
when I just add 
test:
  source_files:
     - target_gene_cluster.txt
     
 the build fails: "Did not find any source_files for test with pattern target_gene_cluster.txt"
 
 The repo itself is not mine. I just want to make it available via conda, so my team can use the tool for our NF pipeline.
 Hence, I sadly don't know, what different test we could use which would be shorter...
 @marcelm

marcelm:
Just `git add` it to the biodigest/ folder in the bioconda-recipes repository (so that it sits next to `meta.yaml`).

recipes/biodigest/run_test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+python -m biodigest.single_validation -t target_gene_cluster.txt -ti symbol -m clustering
```
marcelm, line 3: Since this is so short, it’s easier to maintain if you don’t add this extra file but keep the test in `meta.yaml`:
```yaml
test:
  commands:
    - python -m ...
```

marcelm:
> Hence, I sadly don't know, what different test we could use which would be shorter...

I found something: Adding `-c 1` to the command line makes the test complete in a couple of seconds.

lspindler2509:
the current error about graph-tool with G.Type assertion error, is the one why I added the channels section. I found this tip in some forum when already observing the error earlier. @marcelm

marcelm:
The relevant error message is this one:
```
Failed to load shared library 'libgdk-3.so.0' referenced by the typelib: libXcursor.so.1: cannot open shared object file: No such file or directory
```

This may be an issue with one of the dependencies. This recipe lists some old versions of dependencies, so it is actually possible that the problem has been fixed in a more recent version.

Maybe try adding `xorg-libxcursor` to the dependencies as a workaround.

marcelm:
Here’s the issue you’re hitting: https://github.com/conda-forge/gtk3-feedstock/pull/59

This was fixed in gtk3-3.24.37, build 1, so it would be appropriate to add a dependency on gtk3>=3.24.38. This may not work because there’s a reason conda installs an old gtk3 version into the build environment.

marcelm:
I’m giving up at this point. I don’t know why the CI job is timing out.

I suggest you go back to the working commit 34bdf6c with the `post-link.sh` script.

lspindler2509:
I totally understand. Thank you for your effort! @marcelm

lspindler2509:
@marcelm could you maybe approve the PR? :)

marcelm (approved):

PR #45958, opened 2024-02-22
Add recipe ipk

blinard-BIOINFO:

https://github.com/phylo42/IPK


blinard-BIOINFO:
Dear @dpryan79 ,

I am having a strange CI behavior on this one.
The Lint check fails, and not because of my new recipe **ipk**, but because of recipe **10x_bamtofastq**, which is from the main branch ?

Below log extract from the azure pipeline, Lint step :

```
ERROR: recipes/10x_bamtofastq/meta.yaml:8: build_number_needs_bump: The recipe build number should be incremented
WARNING: recipes/10x_bamtofastq/meta.yaml:34: long_summary: The summary line is rather long
ERROR: recipes/10x_bamtofastq/meta.yaml:0: missing_run_exports: Recipe should have a run_exports statement that ensures correct pinning in downstream packages
Errors were found
```

What should I do ?

PR #45973, opened 2024-02-23
Add recipe ipk

blinard-BIOINFO:

https://github.com/phylo42/IPK


blinard-BIOINFO:
Appears to be an openmp related issue on mac os compilation.
Potential solution found in #8869 .

blinard-BIOINFO:
@BiocondaBot please add label

rpetit3 (approved):

rpetit3:
Nice avatar @blinard-BIOINFO

PR #46020, opened 2024-02-26
Add recipe r-spec2annot

sdechaumet:
This PR add a recipe for the R package [spec2annot](https://github.com/odisce/Spec2Annot) used in Mass Spectrometry and propose a brute-force algorithm to annotate unknown spectra.


sdechaumet:
@BiocondaBot please add label

bgruening:
@sdechaumet can you please submit this recipe to conda-forge?

It does not have a Bioconductor dependency and could be hosted there.

sdechaumet:
Hi @bgruening,

This package constitutes the initial component of a two-package bundle. The second package is dependent on Bioconductor resources ([spec2xtract](https://github.com/odisce/Spec2Xtract)). Both packages are designed for biological applications, specifically in the fields of Metabolomics and Lipidomics. While I could submit **spec2annot** to conda-forge independently, it appears more cohesive to publish both on Bioconda. However, I'm unsure of the optimal approach. Any advice on the best course of action?

bgruening:
We decided at some point that it would be nice to keep all r- only packages in conda-forge. So if you can please submit it to conda-forge.

sdechaumet:
Thanks again for the feedback, I'll push it to conda-forge.

PR #46109, opened 2024-03-01
hybran: limit biopython dependency to <1.82 due to breaking change

0xaf1f:
A deprecation snuck up:

https://github.com/biopython/biopython/blob/master/DEPRECATED.rst#bioseqfeature

> The `location_operator` argument was removed from the SeqFeature initializer in Release 1.82.


0xaf1f:
@BiocondaBot please add label

bgruening (approved):

PR #46283, opened 2024-03-09
Bump version sistr v1.1.2 recipe

kbessonov1984:
SISTR v1.1.2 new recipe to complement other releases on other repos


kbessonov1984:
Will create a new clean PR after figuring out all issues so not to pollute master branch

PR #46328, opened 2024-03-12
Add Snakemake 8 executor plugins for LSF

BEFH:
This adds `snakemake-executor-plugin-lsf-jobstep` and `snakemake-executor-plugin-lsf`, which are necessary for running Snakemake workflows on LSF clusters.


BEFH:
@dlaehnemann, Here is the pull request. Thanks for your advice so far.

PR #46514, opened 2024-03-18
Add recipe for REViewer

micknudsen:
REViewer is a tool for visualizing ExpansionHunter output.

https://github.com/Illumina/REViewer


recipes/reviewer/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  name: reviewer
+  version: '{{ version }}'
+
+source:
+  url: https://github.com/Illumina/REViewer/releases/download/v{{ version }}/REViewer-v{{ version }}-linux_x86_64.gz
+  sha256: '{{ sha256 }}'
+
+build:
+  number: 0
+  noarch: generic
```
martin-g, line 14: The downloaded file says `x86_64`. And it brings pre-built binary for x86_64. It is not really a `noarch` package!

recipes/reviewer/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+mkdir -p "${PREFIX}/bin"
+wget "https://github.com/Illumina/REViewer/releases/download/v${PKG_VERSION}/REViewer-v${PKG_VERSION}-linux_x86_64.gz
```
martin-g, line 4: The file must be already downloaded (and uncompressed) by the recipe.

recipes/reviewer/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  run_exports:
+    - {{ pin_subpackage('myrecipe', max_pin="x.x") }}
```
martin-g, line 16:
```suggestion
    - {{ pin_subpackage('reviewer', max_pin="x.x") }}
```

recipes/reviewer/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+requirements:
+  build:
+    - wget
```
martin-g, line 20: Most probably not needed. See above.

recipes/reviewer/build.sh
micknudsen: I more or less did copy-paste of the `delly` recipe. I'll go ahead and remove this part.

martin-g: You could temporarily use `ls -laR` to see what's available for the build.

recipes/reviewer/meta.yaml
martin-g: IMO it would be better to download the source tarball and build it instead of using the pre-built one.
This way you can try to make the package available for Mac OS (x86_64) and Linux ARM64.

micknudsen: I agree that it would be the best solution, but since I've never complied C code myself before, I'm not really sure how to get started. I thought that downloading the pre-compiled binary would be the easiest way to get a working recipe (which could then be improved later).

martin-g: In that case you may go back to the pre-built binary and let some other user to improve it later.

micknudsen:

recipes/reviewer/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+build:
+  number: 0
+  skip: True # [osx]
+  run_exports:
+    - {{ pin_compatible('reviewer', max_pin="x.x") }}
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - zlib
```
martin-g, line 21: `zlib` should be in `requirements > host` and `requirements > run`
See https://bioconda.github.io/contributor/guidelines.html#c-c

micknudsen: Fixed. Do you think the PR can be merged as it is now? Thanks for assisting!

martin-g (approved):

martin-g:

martin-g:
```

LinuxArtifacts/packages/linux-64 
❯ tar jxvf reviewer-0.2.7-hdcf5f25_0.tar.bz2 
info/files
info/run_exports.json
info/test/run_test.sh
info/recipe/build.sh
info/hash_input.json
info/paths.json
info/index.json
info/recipe/meta.yaml.template
info/recipe/meta.yaml
info/recipe/conda_build_config.yaml
info/about.json
info/licenses/LICENSE.txt
info/recipe/LICENSE.txt
info/git
bin/REViewer

LinuxArtifacts/packages/linux-64 
❯ ./bin/REViewer --help
Program options:
  --help                                Print help message
  --version                             Print version number
  --reads arg                           BAMlet generated by ExpansionHunter
  --vcf arg                             VCF file generated by ExpansionHunter
  --reference arg                       FASTA file with reference genome
  --catalog arg                         Variant catalog
  --locus arg                           Locus to analyze (or a list of 
                                        comma-separated loci)
  --region-extension-length arg (=1000) Length of flanking region (must match 
                                        corresponding ExpansionHunter setting)
  --output-prefix arg                   Prefix for the output files


LinuxArtifacts/packages/linux-64 
❯ file bin/REViewer 
bin/REViewer: ELF 64-bit LSB executable, x86-64, version 1 (GNU/Linux), statically linked, for GNU/Linux 2.6.32, BuildID[sha1]=ce318c4bfa5c714de791b481b1f0cbc3f87ea642, with debug_info, not stripped
```

Looks OK to me!

jmarshall:
The license is listed as GPL and the source appears to be available from the upstream repo, but this recipe just copies a linux-x86_64 binary from an upstream binary artifact. Hence it supports only one of the three platforms supported by bioconda.

Is there a reason this recipe does not build the tool itself, and thus support all three current bioconda platforms?

micknudsen:
> Is there a reason this recipe does not build the tool itself, and thus support all three current bioconda platforms?

@jmarshall I tried to do just that, but I failed to do se. I guess it should be possible, but probably not with my limited experience with C/C++.

jmarshall:
REViewer bundles boost and htslib, while a bioconda package would preferably use conda-forge's boost and bioconda's htslib packages instead of bundling. This is non-trivial for those (such as myself) who are not familiar with cmake :smile: but I am working on getting this building recipe working in #46706.

micknudsen:
@jmarshall Thanks for the effort! I really have no clue how to fix this myself.

micknudsen:

micknudsen:
I fetched artifacts and installed the package on a Linux machine. Everthing looks OK. Is this ready to be merged?

jmarshall:

jmarshall (approved):

PR #46524, opened 2024-03-18
fix busco patch

berkelem:
Patch existed but had not been added to meta.yaml so it wasn't patched.


mencian:
LGTM, thanks @berkelem!

mencian (approved):

PR #46899, opened 2024-03-30
hybran: update to version 1.8

0xaf1f:
New dependencies in this version. Also, we've restored compatibility with latest biopython versions and so I've removed the version pinning.


0xaf1f:
@BiocondaBot please add label

cbrueffer (approved):

PR #47086, opened 2024-04-09
ENH add chardet version specification to argnorm

Vedanth-Ramji:
Specify chardet as requirement due to error during installation with pronto.

### DIFF
--- a/recipes/argnorm/meta.yaml
+++ b/recipes/argnorm/meta.yaml
@@ -11,7 +11,7 @@ source:
 
 build:
   noarch: python
-  number: 1
+  number: 2
   entry_points:
     - argnorm=argnorm.cli:main
   script: "{{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation"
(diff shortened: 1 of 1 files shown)


Vedanth-Ramji:
@BiocondaBot rebuild

luispedro:
Does `argNorm` even depend on `chardet`? It's not listed in its requirement (https://github.com/BigDataBiology/argNorm/blob/main/requirements.txt)

Vedanth-Ramji:
pronto 2.5.6 is dependent on chardet. When I try to install the latest argNorm version I get this error:
image

I am just trying to fix this now.

luispedro:
This is probably best fixed in the pronto recipe (or maybe even upstream)

Vedanth-Ramji:
Just opened a new issue here: https://github.com/althonos/pronto/issues/224

Vedanth-Ramji:
The problem should be with the channels. `conda install -c bioconda argnorm` didn't work, but `conda install -c bioconda -c conda-forge argnorm` did. This makes sense as `chardet` uses the `conda-forge` channel. I think the channels weren't loaded properly from my end.

PR #47294, opened 2024-04-17
Update PhyloAcc recipe

gwct:
I've added a symlink to the `phyloacc.py` executable so it can be called as just `phyloacc`.


gwct:
@BiocondaBot please add label

PR #47358, opened 2024-04-19
Update to version v1.7.1, added better tests

ChrisHIV:
Use new tests that check required external programs can be run and required python modules can be imported


bgruening (approved):

bgruening:
Thanks @ChrisHIV

PR #47340, opened 2024-04-19
Update snakemake to 8.10.8

BiocondaBot:
Update [`snakemake`](https://bioconda.github.io/recipes/snakemake/README.html): **8.10.7** &rarr; **8.10.8**

[install with bioconda](http://bioconda.github.io/recipes/snakemake/README.html) [Conda](https://anaconda.org/bioconda/snakemake/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/snakemake`](https://github.com//bioconda/bioconda-recipes/tree/bump/snakemake/recipes/snakemake) (click to view/edit other files)
Summary | A popular workflow management system aiming at full in-silico reproducibility.
Home | [https://snakemake.github.io](https://snakemake.github.io)
Releases |[https://pypi.org/pypi/snakemake/json](https://pypi.org/pypi/snakemake/json)
Recipe Maintainer(s) | @johanneskoester, @corneliusroemer

***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

PR #47371, opened 2024-04-21
Update snakemake pinned dependency to match setup.cfg

melund:
This PR ensures we use the newest version of smart-open, which was pinned incorrectly here.

It fixes (https://github.com/snakemake/snakemake/issues/2765) which was caused by a bug in smart-open version 3.


melund:
@BiocondaBot please add label

johanneskoester:
Thanks for pointing me to this! I have integrated the fix in https://github.com/bioconda/bioconda-recipes/pull/47340.

PR #47384, opened 2024-04-22
Update shiver: fix non-fatal absence of HXB2 sequence file

ChrisHIV:
Fix non-fatal absence of HXB2 sequence file


mencian (approved):
LGTM, thanks @ChrisHIV!

PR #47390, opened 2024-04-23
update hicexplorer scikit-learn requirement

pavanvidem:
We have the same issue as here with scikit-learn>=1.4: https://github.com/scikit-learn/scikit-learn/discussions/28388


pavanvidem:
This version should fix hicTADClassifier tests

bgruening (approved):

PR #47747, opened 2024-05-07
PATCH: include pysam in libdeflate repodata-patches

lizgehret:
The current build of `pysam` contains a max pin of 1.19 for `libdeflate`, which conflicts with a lot of packages in the QIIME 2 ecosystem. However, it appears as though `pysam` is compatible with the newer version of `libdeflate` - this just needs to be reflected in the bioconda build (see this related [PR](https://github.com/bioconda/bioconda-recipes/pull/40675) and [comment](https://github.com/pysam-developers/pysam/issues/1114#issuecomment-1521641722)).

### DIFF
--- a/recipes/bioconda-repodata-patches/gen_patch_json.py
+++ b/recipes/bioconda-repodata-patches/gen_patch_json.py
@@ -172,7 +172,7 @@ def _gen_new_index(repodata, subdir):
             _pin_looser(fn, record, 'htslib', min_lower_bound='1.10', upper_bound='1.21')
 
         # future libdeflate versions are compatible until they bump their soversion; relax dependencies accordingly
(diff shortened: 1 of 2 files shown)


bgruening:
@jmarshall can you please have a look here? Thanks

jmarshall:
My intention has long been to revert the pysam recipe to using an external htslib package, as it did prior to PR #13093.

IIRC I left pysam out of #40675 in the hope that that reversion was imminent. However I still haven't got around to it, so this is a good improvement that should resolve this libdeflate compatibility problem for extant bioconda pysam packages but not interfere in a future in which pysam no longer directly depends on libdeflate.

jmarshall (approved):

bgruening:
Thanks @jmarshall

lizgehret:
Thanks all, much appreciated!

PR #47803, opened 2024-05-09
Update eBCSgen - add run_exports

xtrojak:
Add `run_exports` to the `meta.yaml` file.


xtrojak:
@BiocondaBot please add label

PR #47886, opened 2024-05-14
update verkko to v2.1

skoren:
Updating verkko to v2.1


skoren:
I think the code is correct and builds a valid package (at least on linux) but I'm having issues because the Azure instances seem unstable and randomly fail with `23:53:09 BIOCONDA INFO (OUT) curl: (18) transfer closed with 24528077 bytes remaining to read` while downloading small 25-ish mb files. I've been able to download the same files multiple times locally on my system so I don't think it's the link or the source. Any ideas why the curl could be failing randomly like this?

skoren:
Got this working but will close and make a cleaner pull request without all the extra commits.

PR #48274, opened 2024-06-04
update recipe

pdimens:


pdimens:
This patch should fix the build recipe for #48273

pdimens:
@BiocondaBot please add label

PR #48328, opened 2024-06-06
Add mmlong2 lite

Serka-M:
Introduction of the mmlong2-lite pipeline into Bioconda

### DIFF
--- a/recipes/mmlong2-lite/build.sh
+++ b/recipes/mmlong2-lite/build.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 
-mkdir -p $PREFIX/bin/ 
-cp -r src/* $PREFIX/bin/.
+mkdir -p $PREFIX/bin/
+cp -r src/* $PREFIX/bin
 chmod +x $PREFIX/bin/mmlong2-lite
--- a/recipes/mmlong2-lite/build.sh
+++ b/recipes/mmlong2-lite/build.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+mkdir -p $PREFIX/bin/ 
+cp -r src/* $PREFIX/bin/.
+chmod +x $PREFIX/bin/mmlong2-lite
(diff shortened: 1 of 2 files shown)


Serka-M:
@BiocondaBot please add label

recipes/mmlong2-lite/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+mkdir -p $PREFIX/bin/ 
+cp -r src/* $PREFIX/bin/.
```
martin-g, line 4: what's the purpose of the trailing dot here ?

recipes/mmlong2-lite/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+build:
+  number: 0
+  noarch: generic
+  run_exports:
+    - {{ pin_subpackage('mmlong2-lite', max_pin="x.x") }}
+
+requirements:
+  run:
+    - snakemake 8.2.3
```
martin-g, line 21: Does it need exactly these versions ?
I guess it will still work with newer versions too ?! Especially for `tar`
Can we relax the pins ?

recipes/mmlong2-lite/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - pigz 2.6
+    - tar 1.34
+
+test:
+  commands:
+    - which mmlong2-lite
+
+about:
+  home: "https://github.com/Serka-M/mmlong2-lite"
+  license: GPL3
```
martin-g, line 34:
```suggestion
  license: GPL-3.0-only
```
or `GPL-3.0-or-newer`
Please see [SPDX](https://spdx.org/licenses/)

martin-g, line 34:
```suggestion
  license: GPL3
  license_file: LICENSE
```
pointing to https://github.com/Serka-M/mmlong2-lite/blob/main/LICENSE

recipes/mmlong2-lite/meta.yaml
Serka-M: With Snakemake, I'd rather stick to a specific version since it gets changed frequently, causing incompatibilities.
The Singularity version is needed and the rest can work with newest versions, although my employer is very adamant about version-controlling everything.

recipes/mmlong2-lite/build.sh
Serka-M: The trailing dot is unnecessary

martin-g (approved):

PR #48356, opened 2024-06-07
Bulk: testing for failures

aliciaaevans:
This is to test out fixes for failures on bulk, specifically osx-64


.circleci/config.yml
```diff
@@ -229,6 +229,7 @@ workflows:
               ignore:
                 - master
                 - bulk
+                - /bulk.*/
```
aliciaaevans, line 232: This is kind of hacky, but I did it because master is hard-coded in this yml file as the root branch and so building this would try to build everything from bulk.

Azure has the option to exclude PRs with the target of bulk so I did that on the bulk branch itself before branching this..

aliciaaevans:
@daler I actually added a separate GHA configuration for this, but I did it on the bulk branch before I created this branch so we could use it for multiple branches. https://github.com/bioconda/bioconda-recipes/blob/bulk-osx-failures/.github/workflows/Bulk-PR.yml 

Sorry I didn't communicate that well when it didn't show up in the PR diff. (Carry on with what you are doing, but it's there if you need it.)

PR #48485, opened 2024-06-14
Bulk

aliciaaevans:
Merging bulk after doing the update for the latest pinnings, including Python 3.11 and 3.12.

Also, removed caching from the bulk CI because it was causing more problems than it solved (already discussed with the core team).

Also added in the build failures updates that has been running on bulk to make the separate bulk page in the wiki that only shows build failures that are not on master.


aliciaaevans:
Successful bulk builds: https://github.com/bioconda/bioconda-recipes/actions/runs/9521946157/job/26250531541 and https://app.circleci.com/pipelines/github/bioconda/bioconda-recipes/120019/workflows/b0f9b88d-4686-40f4-bddb-c54f4bc36c7d

PR #48521, opened 2024-06-17
Add pgscatalog.calc recipe

nebfield:
Add `pgscatalog.calc`, a python package for aggregating and adjusting calculated polygenic scores


recipes/pgscatalog.calc/LICENSE
mencian, line 1: Could you remove the license file? Sorry for being difficult but it is redundant to have two license copies. Otherwise, everything else LGTM!

mencian (changes requested):

recipes/pgscatalog.calc/LICENSE
nebfield: no problem! thanks for the review 😁

PR #48563, opened 2024-06-18
Add CIRI-full recipe

nictru:
Adds the [CIRI-full](https://ciri-cookbook.readthedocs.io/en/latest/CIRI-full.html) package


nictru:
@BiocondaBot please add label

PR #48580, opened 2024-06-19
Add r-doubletfinder recipe

nictru:
This PR adds the [DoubletFiner](https://github.com/chris-mcginnis-ucsf/DoubletFinder) R package

### DIFF
--- a/recipes/r-doubletfinder/meta.yaml
+++ b/recipes/r-doubletfinder/meta.yaml
@@ -37,8 +37,6 @@ about:
   license: unknown
   summary: R package for detecting doublets in single-cell RNA sequencing data
 extra:
-  additional-platforms:
-    - linux-aarch64
   identifiers:
     - doi:10.1016/j.cels.2019.03.003
   recipe-maintainers:
--- a/recipes/r-doubletfinder/meta.yaml
+++ b/recipes/r-doubletfinder/meta.yaml
@@ -30,7 +30,7 @@ requirements:
 
 test:
   commands:
-    - $R -e "library(doubletFinder)"
+    - $R -e "library(DoubletFinder)"
 
 about:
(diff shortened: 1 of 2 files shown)


nictru:
@BiocondaBot please add label

recipes/r-doubletfinder/meta.yaml
```diff
@@ -0,0 +1,45 @@
+package:
+  name: r-doubletfinder
+  version: 1.0.0
+
+source:
+  url: https://github.com/chris-mcginnis-ucsf/DoubletFinder/archive/03e9f37f891ef76a23cc55ea69f940c536ae8f9f.zip
```
martin-g, line 6: Any chance you could ask the author to make a release/tag ?

recipes/r-doubletfinder/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  commands:
+    - $R -e "library(DoubletFinder)"
+
+about:
+  home: https://github.com/chris-mcginnis-ucsf/DoubletFinder
+  license: unknown
+  summary: R package for detecting doublets in single-cell RNA sequencing data
+extra:
+  additional-platforms:
+    - linux-aarch64
```
martin-g, line 41: `additional-platforms` is not needed for `noarch` recipes

martin-g (changes requested):

recipes/r-doubletfinder/meta.yaml
nictru: I could, but would probably take a while. 
Would it be an acceptable intermediary  solution to have it as a 0.0.1 version until we have a proper tag?

recipes/r-doubletfinder/meta.yaml
nictru: Ah, good to know, thanks

PR #48607, opened 2024-06-19
sortmerna recipe relocating to conda-forge

biocodz:
Sortmerna recipe is [being relocated to conda-forge - PR 25438](https://github.com/conda-forge/staged-recipes/pull/25438).
Current sortmerna package v4.3.7 on bioconda channel is Not Valid and should be removed. Sortmerna [issue 423](https://github.com/sortmerna/sortmerna/issues/423)
Starting with  4.3.7 sortmerna will be hosted on conda-forge and will include Windows builds not supported on bioconda.


martin-g (approved):

martin-g:
> v4.3.7 on bioconda channel is Not Valid and should be removed

I cannot help with the removal.

aliciaaevans:
Packages in conda-forge will override those in bioconda because of strict repo priority, so once it's in conda-forge I believe it should not be an issue. If it's causing problems, @bgruening might be able to help with removal.

In any case, let's merge this now to avoid any new updates.

PR #48753, opened 2024-06-26
Fix eukulele

paulzierep:
This tool fails in runtime since it requires busco.


martin-g:
@paulzierep  Please bump the build number

PR #48898, opened 2024-07-02
Update primerforge to 1.5.2

BiocondaBot:
Update [`primerforge`](https://bioconda.github.io/recipes/primerforge/README.html): **1.1.1** &rarr; **1.5.2**

[install with bioconda](http://bioconda.github.io/recipes/primerforge/README.html) [Conda](https://anaconda.org/bioconda/primerforge/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/primerforge`](https://github.com//bioconda/bioconda-recipes/tree/bump/primerforge/recipes/primerforge) (click to view/edit other files)
Summary | primerForge: software to identify primers that can be used to distinguish genomes.
Home | [https://github.com/dr-joe-wirth/primerForge](https://github.com/dr-joe-wirth/primerForge)
Releases |[https://github.com/dr-joe-wirth/primerForge/tags](https://github.com/dr-joe-wirth/primerForge/tags)
Recipe Maintainer(s) | @dr-joe-wirth
Author | @dr-joe-wirth
***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).


recipes/primerforge/meta.yaml
```diff
@@ -28,6 +28,8 @@ requirements:
     - numpy
     - primer3-py >=2.0
     - scipy >=1.10
+    - pip:
```
martin-g, line 31:
```suggestion
```

recipes/primerforge/meta.yaml
```diff
@@ -28,6 +28,8 @@ requirements:
...
+      - khmer >=2.1
```
martin-g, line 32:
```suggestion
    - khmer >=2.1
```

recipes/primerforge/meta.yaml
```diff
@@ -28,6 +28,7 @@ requirements:
     - numpy
     - primer3-py >=2.0
     - scipy >=1.10
+    - khmer >=2.1
```
martin-g, line 31:
```suggestion
    - khmer >=2.1,<3
```

dr-joe-wirth:
not sure how to fix this. in my [environment.yml file](https://github.com/dr-joe-wirth/primerForge/blob/main/environment.yml), it did not work when I specified it as @martin-g did. It only worked when I allowed `pip` to install the package.

i'd gladly change it, but i'm unsure how to track down and resolve the conflict.

martin-g:
Which version has been installed by `pip` ? Maybe we can try with exact version instead of a range ?!

dr-joe-wirth:
> Which version has been installed by `pip` ? Maybe we can try with exact version instead of a range ?!

version 2.1.1. So i had thought that specfiying `khmer >=2.1` would be sufficient.

dr-joe-wirth:
would including a line like `pip install khmer` in a `build.sh` file be compliant with bioconda syntax? or is this still a bad practice?

dr-joe-wirth:
i am going to fork the khmer recipe and update it to work up to 3.11. It is not compatible with 3.12

rpetit3:
I think we have a few options

1. Try building khmer 2.1.1 with latest python pinnings
2. try rebuilding khmer 2.1 with latest python pinnings
3. (do not recommend) pingeon-hole primerforge to only work with python 3.6 (which might not even be supported by primerforge)

rpetit3:
> would including a line like `pip install khmer` in a `build.sh` file be compliant with bioconda syntax? or is this still a bad practice?

You are correct, this wouldn't be allowed

dr-joe-wirth:
@rpetit3 thanks for the input. Here are my thoughts.

> I think we have a few options
> 
> 1. Try building khmer 2.1.1 with latest python pinnings
> 2. try rebuilding khmer 2.1 with latest python pinnings
> 3. (do not recommend) pingeon-hole primerforge to only work with python 3.6 (which might not even be supported by primerforge)

### TL;DR
Option 3 is probably a bad choice; where can I find resources for doing options 1 or 2?

### full response
I think option 1 is the best option, but khmer 2.1.1 does not have a conda recipe. So maybe that means option 2 is better? To be honest, I don't really know how we would go about rebuilding a recipe with the latest python pinnings. Is there a reference available that I could use to figure this out?

Option 3 is not good because I know for sure that the inclusion of type hinting without importing `typing` is not supported below python 3.9. If I had said like `from typing import List, Dict, Tuple` then it may have been compatible with lower versions. Obviously I would prefer not to rewrite the entire code base just to make a functioning conda installation especially when my pip installer works just fine.

As a note, `khmer` does not work with python 3.12 because `SafeConfigParser` was replaced with `ConfigParser`. But the pip installation does work with `python >=3,9,<3.12`.


### additional note
> > would including a line like `pip install khmer` in a `build.sh` file be compliant with bioconda syntax? or is this still a bad practice?
> 
> You are correct, this wouldn't be allowed

Ok thanks for clearing that up. I had started working on this as a solution (this is what I was trying to do in my [`build.sh`](https://github.com/bioconda/bioconda-recipes/blob/f67865935112b48996c91ef7741d891bcd9815aa/recipes/primerforge/build.sh) file, but I will discard this work and move things back to how they were.

dr-joe-wirth:
updated the version, but this recipe is still broken since the recipe for `khmer` is incompatible with the `python3` versions required for `primerForge`

dr-joe-wirth:
@mencian why did you remove the `primerForge --check_install` line from the test? That test specifically checks that all dependencies are installed correctly and if it is failing then the package should not be put into production.

martin-g:
@dr-joe-wirth The commit message says `trailing 'a3' in khmer version causes --check_install to error out`:
```
ValueError: invalid literal for int() with base 10: '0a3'
```
https://github.com/bioconda/bioconda-recipes/actions/runs/12919334584/job/36029731116

PR #48906, opened 2024-07-03
Add orthologer

ftegenfe:
This package contains a software to establish orthology among a set of fasta files. It is described here:

    https://orthologer.ezlab.org/

The online database (OrthoDB) is found here:

    https://www.orthodb.org

The software is also capable of mapping user fasta sequences to existing orthology from OrthoDB.

### DIFF
--- a/recipes/multiqc/meta.yaml
+++ b/recipes/multiqc/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "multiqc" %}
-{% set version = "1.24" %}
+{% set version = "1.24.1" %}
 
 package:
   name: {{ name }}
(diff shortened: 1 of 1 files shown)


ftegenfe:
Fixed test command such that it should allow to be run as root user

recipes/orthologer/meta.yaml
```diff
@@ -0,0 +1,69 @@
+{% set version = "3.2.3" %}
+{% set sha256 = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855" %}
```
martin-g, line 2:
```suggestion
{% set sha256 = "e9c2d7be3939dae479031065d1eac8163f5cb1d6e22c056244b4f80234a2aebb" %}
```

PR #48969, opened 2024-07-05
Update to version 0.1.21

AntonS-bio:
Change python dependency to be >3.10 from =3.9.

### DIFF
--- a/recipes/handyreadgenotyper/meta.yaml
+++ b/recipes/handyreadgenotyper/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "HandyReadGenotyper" %}
-{% set version = "0.1.18" %}
+{% set version = "0.1.20" %}
 
 package:
   name: {{ name|lower }}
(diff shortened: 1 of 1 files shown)


jodyphelan (approved):

corneliusroemer:
Just wondering @AntonS-bio: if it worked with 3.8, and works with 3.10+, why do you need to now exclude 3.8 and 3.9?

You can just say >3.8, if any package has extra dependencies, like pysam, the solver will take care of it!

jodyphelan:
I haven't seen that error before, what exactly does that mean?
Also, I'm pretty sure the checks had all passed when I approved, not sure why the tests on the latest commit failed now.

corneliusroemer:
I don't think there's any failure, all good. No clue what mergify does. I think we can safely ignore. You did everything right.

I reopened an old PR because I didn't know about this PR. So that might have confused mergify (whatever it does)

jodyphelan:
Ah right! 👍🏼

PR #47946, opened 2024-05-20
Update ipk to v0.5.1

blinard-BIOINFO:
Updates IPK to v0.5.1, which introduces several bugfixes.

PR #49000, opened 2024-07-08
Ipk: issue related to boost-cpp version which differs in bioconda-build CI and local install.

blinard-BIOINFO:
Package does not work because of different behavior in conda-build CI and install on some local computer.

@martin-g @Juke34 
I'm getting a weird behavior related to boost following the package development.

In the conda-build CI, package `boost-cpp=1.85` got selected for the conda env.
My rule initially required `boost-cpp>=1.67`.

As one can see in the CI logs of PR #47946 :

```
# see pipeline https://github.com/bioconda/bioconda-recipes/runs/27086383834

ONDA INFO (OUT) The following NEW packages will be INSTALLED:
15:55:23 BIOCONDA INFO (OUT) 
15:55:23 BIOCONDA INFO (OUT)     _libgcc_mutex:    0.1-conda_forge   conda-forge
15:55:23 BIOCONDA INFO (OUT)     _openmp_mutex:    4.5-2_gnu         conda-forge
15:55:23 BIOCONDA INFO (OUT)     boost-cpp:        1.85.0-h44aadfe_2 conda-forge        <<===
15:55:23 BIOCONDA INFO (OUT)     bzip2:            1.0.8-hd590300_5  conda-forge
15:55:23 BIOCONDA INFO (OUT)     icu:              73.2-h59595ed_0   conda-forge
15:55:23 BIOCONDA INFO (OUT)     libboost:         1.85.0-hba137d9_2 conda-forge
15:55:23 BIOCONDA INFO (OUT)     libboost-devel:   1.85.0-h00ab1b0_2 conda-forge
15:55:23 BIOCONDA INFO (OUT)     libboost-headers: 1.85.0-ha770c72_2 conda-forge

[...]

CONDA INFO (OUT) -- Found Boost: $PREFIX/lib/cmake/Boost-1.85.0/BoostConfig.cmake (found version "1.85.0") found components: serialization iostreams system filesystem

```

So the binary is built with boost v1.85 on bioconda side.
However, when installing on a fresh env in my local computer, conda or mamba refuses to use `boost-cpp` 1.85 and stick to boost-cpp 1.84. This results to a conflit of shared boost librairies and the program fails to run.

```
# running the command 

ipk.py build --refalign reference.fasta --reftree tree.rooted.newick --states nucl --workdir . --model GTR

/home/belinard/SOFTWARE_LOCAL/miniconda3/envs/epik/bin/ipk-dna: error while loading shared libraries: libboost_program_options.so.1.85.0: cannot open shared object file: No such file or directory

# the binary from the package look for v1.85 like in conda-build

ldd miniconda3/envs/epik/bin/ipk-dna
	linux-vdso.so.1 (0x00007ffeebbd7000)
	libboost_program_options.so.1.85.0 => not found
	libboost_filesystem.so.1.85.0 => not found
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x0000788eb5097000)
	libboost_serialization.so.1.85.0 => not found
	libboost_iostreams.so.1.85.0 => not found
	libstdc++.so.6 => /home/belinard/SOFTWARE_LOCAL/miniconda3/envs/epik/bin/../lib/libstdc++.so.6 (0x0000788eb4eb2000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x0000788eb4dcb000)
	libgcc_s.so.1 => /home/belinard/SOFTWARE_LOCAL/miniconda3/envs/epik/bin/../lib/libgcc_s.so.1 (0x0000788eb4dac000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x0000788eb4a00000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x0000788eb4da5000)
	/lib64/ld-linux-x86-64.so.2 (0x0000788eb515b000)

# but the command `conda install ipk epik` installs v1.84

# conda list gives:

boost-cpp                 1.84.0               h44aadfe_3    conda-forge
libboost                  1.84.0               hba137d9_3    conda-forge
libboost-devel            1.84.0               h00ab1b0_3    conda-forge
libboost-headers          1.84.0               ha770c72_3    conda-forge

# i can confirm by looking at the files
miniconda3/envs/epik/lib/libboost_serialization.so

```

I looked at the repo libboost-feedstock which maintains `boost-cpp` and it seems this can happen if channel_priority is not set to strict for conda-forge. I check the documentation, but so far did not found a way to force "strict" priority for a bioconda recipe. Is this even possible ?

In this PR, I set the `boost-cpp` version to a strict version in the recipe, hoping this will resolve the issue.

But I wanted to let you know that this weird behavior happens (and may break potential other libboost-based packages, which may be super common for C++ packages).

The worse part is that all looks clean in your CI side, then you merge to make it available online, and then things get messed-up when someone is actually installing the package on his/her computer.


martin-g:
It uses 1.85 for me too!

blinard-BIOINFO:
The issue comes from the fact that my default conda installation uses flexible priority.

```
$  conda config --show channels
channels:
  - conda-forge
  - bioconda
  - defaults
  - r

$  conda config --show channel_priority
channel_priority: flexible
```
As mentionned here: https://github.com/conda-forge/boost-feedstock
The boost-cpp package can be an issue is priority is not set to strict.

I was wondering if there is any way to set in the meat.yaml a directive to forge channels and strict priority for a particular package (e.g boost-cpp in my case) ?

blinard-BIOINFO:
Also, in a freash environement, where I install only package `ipk`, trying to remove it laucnhes an error.
I'm obliged to delete the full environement and do a conda clean to solve the issue.


```
Confirm changes: [Y/n] Y
PackageRecord(_hash=-1232370165032959845, name='__unix', version='0', build='0', build_number=0, channel=Channel("@"), subdir='linux-64', fn='__unix', md5='12345678901234567890123456789012', package_type='virtual_system')

# >>>>>>>>>>>>>>>>>>>>>> ERROR REPORT <<<<<<<<<<<<<<<<<<<<<<

      File "/home/belinard/SOFTWARE_LOCAL/miniconda3/lib/python3.10/site-packages/conda/_vendor/boltons/setutils.py", line 247, in remove
        didx = self.item_index_map.pop(item)
    KeyError: PackageRecord(_hash=-1232370165032959845, name='__unix', version='0', build='0', build_number=0, channel=Channel("@"), subdir='linux-64', fn='__unix', md5='12345678901234567890123456789012', package_type='virtual_system')

```

blinard-BIOINFO:
Last element. 

* both recipes of `ipk` and `epik`requesting the same boost version constraint in the `meta.yaml`
* These two packages are using boost C++ library, which needs to be the same version. `ipk` produces a binary database, and `epik` loads it to perform taxonomic classification

Temporary solution is to manually fix both to `boost-cpp=1.85`, as a more strict versioning constraint.

Juke34:
Usually we prefer to keep compatibility as large as possible, so preferring ` boost-cpp >=1.67`instead of `boost-cpp =1.85` if no perculiar reason push against that.
We also promote the usage of strict channel priority, so I guess most of user would not meet any issue.
Then the synchronous version needed between EPIK and IPK for me is another problem. 
First I'm not sure to understand the problem. **Is the binary database produced impacted by the boost-cpp version?**
That is strange to me, but in that case it should be well stated on EPIK that bugs can occur if boost-cpp used in EPIK and IPK differ. Another solution is to save in the IPK build database the version of boost-cpp used, and check this information when running EPIK throwing a warning explaining the problem when versions differ. Or both tools should be shipped together...

I do not see any problem at short term to set  boost-cpp to 1.85 (both for EPIK and IPK recipes) but if  `boost-cpp` must be the same for EPIK and IPK then problems will occur as soon as the version will be modified in one of the recipe. Synchronizing dependencies versions between recipes seem impossible to me.

What is your thought @martin-g ?

blinard-BIOINFO:
>  First I'm not sure to understand the problem. Is the binary database produced impacted by the boost-cpp version?

Indeed, because it is serialized via boost library. Boost guarantees correct de-serialization only for the same version of the library. Problem is, there is no good mechanism at de-serialization time to verify which library version was used at serialization time... We scratch our heads about this for some time now and the best solution we found was to ensure same boost headers version in the compile environment (I basically do the same in conda).

Initially, the choice of package separation is because IPK creates databases that can be used by different tools. For instance, the package `sherpas` that I created years ago, can also load IPK-built databases, but for different analysis purposes (virus recombination, while epik is for taxonomic classification). 

I agree with you, ultimately the solution could be to merge all tools into one conda recipe. But then, to make the release mechanism work, we will have to create a "merged" github repo, with its own releases, for a single conda package. 

In the meantime, I would be happy to stick to `boost-cpp=1.85`. On our side, we need to discuss all this for future releases.

Juke34 (approved):

PR #49098, opened 2024-07-10
Build raxml for osx-arm64

corneliusroemer:


corneliusroemer:
@martin-g I noticed you upstreamed some changes to raxml to add aarch64 support. Maybe you could do likewise here? I'm using a patch for now as I don't understand the details of the necessary changes, but would be cool if you could have a look at whether we might build with pthreads etc for arm64 as well.

corneliusroemer:
@BiocondaBot please add label

corneliusroemer:

recipes/raxml/build.sh
```diff
@@ -24,8 +33,8 @@ for PTHREADS in "" .PTHREADS; do
     else
       MAKEFILE=${MAKEFILE}.gcc
     fi
-    make -f ${MAKEFILE} CC=$CC
```
martin-g, line 27: Why did you remove `CC=$CC` ?
I think this is the reason why now you need the `patch` (to remove `CC = gcc`.

corneliusroemer:
Ah this is a great point! I don't quite remember, maybe I was trying to
debug locally and didn't have the CC env variable set. I don't know very
much about make yet.

Thanks for your eyes!

recipes/raxml/build.sh
corneliusroemer: 🤦 stupid me, thanks for noticing!

martin-g (approved):

PR #49151, opened 2024-07-11
[Testing only]Test osx-arm64 onGitHub Actions

aliciaaevans:
CircleCI has some limitations, so checking concurrent builds on GHA to see if it works better.

Also there's an error that occurred on multiple PRs about m4, checking to see if that happens on GHA.


aliciaaevans:
The m4 error did occur, so it's not specific to CircleCI.

PR #49199, opened 2024-07-12
Update zDB to version 1.3.1 and update recipe.

njohner:
- Update version pinnings
- Add new maintainer, doc URL and license_family.

### DIFF
--- a/recipes/zdb/meta.yaml
+++ b/recipes/zdb/meta.yaml
@@ -6,7 +6,7 @@ package:
 
 build:
   noarch: generic
-  number: 0
+  number: 1
   run_exports:
     - {{ pin_subpackage('zdb', max_pin="x.x") }}
 
@@ -16,7 +16,7 @@ source:
 
 requirements:
   run:
-    - nextflow >=21.04.0,<=22.10.8
+    - nextflow >=21.04.0
     - mamba
 
 test:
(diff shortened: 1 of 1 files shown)


recipes/zdb/meta.yaml
```diff
@@ -6,17 +6,17 @@ package:
...
 
 source:
   url: https://github.com/metagenlab/zDB/archive/refs/tags/v1.3.0.tar.gz
-  sha256: 3f9496d423875e0d281bc62c3fcc88f9878c167ea14ff9b8afebfea4087466c0
+  sha256: 485cf00d9148e5560f62d4656649c73ba94f3ce434e19759f5306a0fa51ff606
 
 requirements:
   run:
-    - nextflow >=21.04.0,<=22.10.8
+    - nextflow >=21.04.0
```
bgruening, line 19: Why is this upper pin not needed anymore? And why does the sha256 changes in the same version?

njohner: - We ported the nextflow code do DSL2, so we do not need support for DSL1 anymore.
- The sha256 changed because I moved the tag in the source code. I know that's ugly, but I screwed up the release order: zDB uses containers of itself which I forgot to build and bump in the code before doing the 1.3 release. So the release was basically broken. I decided that it's better to move the tag than having a broken release. If that's not acceptable for you guys, I'll move the tag back and make a release 1.3.1.

martin-g: I think making a new release (1.3.1) would be better!

njohner: Ok I moved the 1.3.0 tag back to its original commit in the zDB repository and made a new release 1.3.1. I updated this PR to update the recipe for the 1.3.1 version of zDB.

This PR should replace the [automatically generated PR for the 1.3.1 release](https://github.com/bioconda/bioconda-recipes/pull/49413)

njohner: Ok well [version 1.3.1 has been merged](https://github.com/bioconda/bioconda-recipes/pull/49413), so only update of recipe is left in this PR.

martin-g: I think we need to bump the build number

njohner: Ok hope it's good now.

martin-g (approved):

PR #49226, opened 2024-07-14
Update knotinframe

sjanssen2:
I've updated upstream and reflect this change with this PR. Furthermore, pin_subpackage was added.


recipes/knotinframe/meta.yaml
```diff
@@ -1,37 +1,42 @@
 {% set version = "2.3.1" %}
-{% set sha256 = "d9ab68329e48709cffbf5ed775f905bdddebab470d70836bb522993b971d48af" %}
+{% set packagename = "knotinframe" %}
+{% set sha256 = "0c5b7d91893d0a23189f4af3602e18f110f030eac49a3658f7acede23596fde0" %}
 
 package:
-  name: knotinframe
-  version: {{ version }}
+  name: {{ packagename }}
+  version: 2.2.0
```
epruesse, line 7: Why are you removing the {{version}} template and downgrading from 2.3.1 to 2.2.0 here?

recipes/knotinframe/meta.yaml
```diff
@@ -1,37 +1,42 @@
...
-    - bellmans-gapc >=2020.12.08
+    - bellmans-gapc >=2024.01.12
 
 test:
   commands:
   - echo "AUGC" | knotinframe | grep 'No suitable slippery sites'
 
 about:
-  home: https://bibiserv.cebitec.uni-bielefeld.de/knotinframe
+  home: https://bibiserv.cebitec.uni-bielefeld.de/{{ packagename }}
```
epruesse, line 38: Replacing `knotinframe` with a template `packagename` would make sense if you were building multiple tools from that location in one recipe, which is not what we usually do. Since you aren't using the template throughout, nor building multiple things here, I'm not sure why you are using a template for the URL.

epruesse (changes requested):

recipes/knotinframe/meta.yaml
sjanssen2: Hi @epruesse thanks for your review!
The {{version}} variable shall address the release version of the upstream repository. This repository serves not only knotinframe, but also rnashapes, rnaalishapes, pkiss, palikiss, rapidshapes and acms. 

The version of the software KnotInFrame was previously confused with the upstream repo version and should have been "2.1.0" as defined in https://github.com/jlab/fold-grammars/blob/5b46de8bb107d9b5cf4955136e7ccffdaf0bca75/Misc/Applications/lib/foldGrammars/Settings.pm#L15 and was actually bumped to 2.2.0 by me recently.

My idea is that {{version}} of the upstream package will help recognizing novel releases of the upstream repo and thus help auto bump bioconda packages.

recipes/knotinframe/meta.yaml
sjanssen2: because I was using copy & paste from the above mentioned other packages like https://github.com/bioconda/bioconda-recipes/blob/f20dff4c1331bd49d80d3500eacad472a1212b91/recipes/rnaalishapes/meta.yaml#L41 or https://github.com/bioconda/bioconda-recipes/blob/f20dff4c1331bd49d80d3500eacad472a1212b91/recipes/palikiss/meta.yaml#L41

sjanssen2:
I am now differentiating `fold_grammars_version` and `knotinframe_version` to avoid future confusions.

recipes/knotinframe/meta.yaml
```diff
@@ -18,20 +21,22 @@ requirements:
...
   license: 'GPL-3.0-or-later'
-  license_family: GPL3
```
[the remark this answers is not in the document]
sjanssen2: good catch! Removal was an error on my end :-/

recipes/knotinframe/meta.yaml
```diff
@@ -1,13 +1,17 @@
...
 
 source:
   sha256: {{ sha256 }}
-  url: https://github.com/jlab/fold-grammars/archive/{{ version }}.tar.gz
+  # the repository "fold-grammars" contains sources for several programs, i.e. pKiss, RNAshapes, ...
+  # thus, the version number does not correspond to the {{ packagename }} version number
+  url: https://github.com/jlab/fold-grammars/archive/{{ fold_grammars_version }}.tar.gz
 
 build:
   number: 0
```
martin-g, line 17:
```suggestion
  number: 1
```

sjanssen2:
@epruesse ready for your review :-) Thanks got to @martin-g !

epruesse:
@sjanssen2 apologies for not responding - is this still an active PR?

sjanssen2:
Thanks @epruesse for the follow up. I had to go back to the code and double check, but yes this is still a change I would love seen to be merged. Therefore, a review from your side is highly appreciated!

PR #49227, opened 2024-07-14
Add rnaalishapes

sjanssen2:
`RNAalishapes` extends RNAshapes by also accepting RNA sequence **alignments** as input to produce various predictions for secondary structures.


martin-g:
You could apply the same suggestions from https://github.com/bioconda/bioconda-recipes/pull/49228 !
Same for your other PRs based on the same source repository.

martin-g (approved):

PR #49228, opened 2024-07-14
Add palikiss

sjanssen2:
`pAliKiss` extends `pKiss` by also accepting RNA sequence alignments as input to produce various predictions for secondary structures including H- and K- type pseudoknots.


recipes/palikiss/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+make PREFIX=$PREFIX CC=$CC -C Misc/Applications/pAliKiss all
```
martin-g, line 2:
```suggestion
make -j ${CPU_COUNT} PREFIX=$PREFIX CC=$CC -C Misc/Applications/pAliKiss all
```

recipes/palikiss/build.sh
```diff
@@ -0,0 +1,5 @@
...
+make PREFIX=$PREFIX CC=$CC -C Misc/Applications/pAliKiss install-program
```
martin-g, line 3:
```suggestion
make -j ${CPU_COUNT} PREFIX=$PREFIX CC=$CC -C Misc/Applications/pAliKiss install-program
```

recipes/palikiss/build.sh
```diff
@@ -0,0 +1,5 @@
...
+make PREFIX=$PREFIX CC=$CC -C Misc/Applications/lib install
```
martin-g, line 4:
```suggestion
make -j ${CPU_COUNT} PREFIX=$PREFIX CC=$CC -C Misc/Applications/lib install
```

recipes/palikiss/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
```
martin-g, line 1:
```suggestion
#!/bin/bash

set -xe

```

recipes/palikiss/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+source:
+  sha256: {{ sha256 }}
+  # the repository "fold-grammars" contains sources for several programs, i.e. pKiss, RNAshapes, ...
+  # thus, the version number does not correspond to the {{ packagename }} version number
+  url: https://github.com/jlab/fold-grammars/archive/{{ version }}.tar.gz
+
+build:
+  number: 0
+  skip: False # [osx]
```
martin-g, line 17: Do you really mean `false` here ? Usually the value is `True`, i.e. skip the build on OSX

recipes/palikiss/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - libopenblas
+
+test:
+  commands:
+    - pAliKiss -h
+    - pAliKiss CGCAUGCUAGUCGCG
+
+about:
+  home: https://bibiserv.cebitec.uni-bielefeld.de/{{ packagename }}
+  license: 'GPLv3+'
```
martin-g, line 41: Please use SPDX identifiers, i.e. `GPL-3.0-or-later.
See https://spdx.org/licenses/

recipes/palikiss/build.sh
sjanssen2: Thanks @martin-g. Makes total sense to run make in parallel here.

recipes/palikiss/build.sh
sjanssen2: as this target is basically copies files, I don't think it needs to run in parallel

recipes/palikiss/build.sh
sjanssen2: same here: no real use in running in parallel as it just copies a few set of files

recipes/palikiss/build.sh
sjanssen2: I've changes to bash

recipes/palikiss/meta.yaml
sjanssen2: yes, this is intentionally. I had to invest some time to make compilation work on OSX. Thus, I've started early versions of my software with skipping OSX and nowadays am able to also support OSX (thanks to @corneliusroemer for assistance with mismatching libc). Should I drop OSX in future versions, I thus know directly how to do that, instead of going through conda readmes again.

recipes/palikiss/meta.yaml
sjanssen2: done

recipes/palikiss/build.sh
martin-g: Then why does it need ` CC=$CC` ?

recipes/palikiss/build.sh
martin-g: Why does it need ` CC=$CC` ?

recipes/palikiss/meta.yaml
martin-g: It is confusing. Please either add a comment or remove it!

recipes/palikiss/build.sh
sjanssen2: this is to obtain the ARCHTRIPLE https://github.com/jlab/fold-grammars/blob/0ee5347c5d5559b340589338dea25b2ecb8bdfcb/makefile#L18 and https://github.com/jlab/fold-grammars/blob/0ee5347c5d5559b340589338dea25b2ecb8bdfcb/Misc/Applications/RNAshapes/makefile#L22 which is part of a temporary directory storing compile results

recipes/palikiss/build.sh
sjanssen2: see above

recipes/palikiss/meta.yaml
sjanssen2: I decided to remove

martin-g (approved):

sjanssen2:
Thanks again for your great and prompt assistance @martin-g !

PR #49261, opened 2024-07-15
ci: revert automatic artifact fetching

aliciaaevans:
This bot is a little out of control. I guess GitHub considers each CI platform to be its own check suite?

I'll see if I can figure out a way to get it to do what we want, but for now, I don't want to spam all the PRs.


martin-g (approved):
I see !
Thanks!

PR #49359, opened 2024-07-17
attempting to get khmer compatible with python3.9+

dr-joe-wirth:
previously tried to get `khmer` recipe working with higher versions of python3, but it didn't actually build anything. hopefully i can solve that issue in this PR


dr-joe-wirth:
closing in favor of #49276

PR #49522, opened 2024-07-24
Add carpedeam

LouisPwr:
Adding recipe for a new metagenomics tool "carpedeam".


LouisPwr (commented):
update recipe

LouisPwr:
@BiocondaBot please add label

recipes/carpedeam/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - zlib
+    - bzip2
+
+test:
+  commands:
+    - carpedeam > /dev/null
+
+about:
+  home: https://github.com/LouisPwr/CarpeDeam
+  license: GPLv3
```
martin-g, line 41:
```suggestion
  license: GPL-3.0-only
  license_family: GPL3
```

recipes/carpedeam/meta.yaml
```diff
@@ -0,0 +1,50 @@
+{% set version = "1.0.0" %}
+{% set sha256 = "92deec3d670810f8433c0345758588b47fa460349e61ce8aa549374f5cf4a812" %}
+
+package:
+  name: carpedeam 
+  version: {{ version|replace("-", ".")  }}
```
martin-g, line 6: There is no need of `| replace` since the version does not have any `-`s. But it does not do any harm either.

recipes/carpedeam/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+build:
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('carpedeam', max_pin=None) }}
```
martin-g, line 11: Are you sure about `None` here ?
Please see the PR [template](https://github.com/bioconda/bioconda-recipes/blob/master/.github/PULL_REQUEST_TEMPLATE.md) what are the possible values and their meaning.

recipes/carpedeam/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - zlib
+    - bzip2
+  run:
+    - gawk
+    - zlib
+    - bzip2
+
+test:
+  commands:
+    - carpedeam > /dev/null
```
martin-g, line 37: Such kind of tests usually pipe to `grep`, e.g. `carpedeam --version | grep {{ version }}` or `carpedeam --help | grep "Some expected text"`. If the command returns non-0 status code then it is considered an error.

recipes/carpedeam/meta.yaml
LouisPwr: Thanks for the reference, I think `None` should be fine, but to be safe I changed to `"1.0"`

recipes/carpedeam/meta.yaml
LouisPwr: That is good to know! Thanks. I added a more informative test.

LouisPwr (commented):
I updated the` meta.yaml` file. Thanks for your comments!

recipes/carpedeam/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+{% set sha256 = "92deec3d670810f8433c0345758588b47fa460349e61ce8aa549374f5cf4a812" %}
+
+package:
+  name: carpedeam 
+  version: {{ version|replace("-", ".")  }}
+
+build:
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('carpedeam', max_pin="1.0") }}
```
martin-g, line 11: The value should be `"x"` or `"x.x"`, not `"1.0"`.
If you expect to make breaking changes in 1.x then use `"x.x"`

recipes/carpedeam/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+{% set sha256 = "92deec3d670810f8433c0345758588b47fa460349e61ce8aa549374f5cf4a812" %}
+
+package:
+  name: carpedeam 
+  version: {{ version|replace("-", ".")  }}
+
+build:
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('carpedeam', max_pin="1.x") }}
```
martin-g, line 11:
```suggestion
    - {{ pin_subpackage('carpedeam', max_pin="x.x") }}
```

recipes/carpedeam/meta.yaml
LouisPwr: Ah sorry, I saw my mistake. Thanks for pushing.

martin-g (approved):

PR #49885, opened 2024-08-07
Add CryptKeeper

croots:
CryptKeeper is a negative design tool that supports molecular biologists in identifying and responding to unwanted, cryptic gene expression. CryptKeeper relies on ViennaRNA (one step further down the dependency tree through OSTIR) and TranstermHP, so a Conda recipe would streamline installation for users on MacOS and Linux.


croots:
I'm learning that you apparently can't rely on conda to install pip packages even though that's supported with an environment.yml file...

Will have to come back to this later since I'll need to add several additional conda recipes this depends on.

PR #49950, opened 2024-08-09
Try building ncbi-stxtyper for ARM

evolarjun:
Try adding ARM targets to meta.yaml


evolarjun:
Thanks @peterjc!

It looks like this won't work until the package it's dependent on are available for ARM. I'll close and delete and try again later once I can see that blast is available on ARM. I don't have time right now to work on the blast package.

peterjc:
Funnily enough I'm also hoping to get BLAST on ARM working - there are some perl dependencies etc to sort out first.

Issue #49899, opened 2024-08-08
Linux mummer 3.23-pl5321hdbdd923_19 broke


peterjc:
Unexpected side effect of #49870 which fixed macOS Intel and added macOS ARM.

Build 19 is working nicely for me on Intel Mac:

```
(x86 macOS) $ conda list mummer
# packages in environment at /Users/xxx/mambaforge/envs/pyani-plus_py312:
#
# Name                    Version                   Build  Channel
mummer                    3.23            pl5321h4a92bd6_19    bioconda
(x86 macOS) $ rm -rf tmp* ; nucmer -p tmp NC_002696.fasta NC_014100.fna ; echo "Return $?"
...
Return 0
```

But sadly this has broken on Linux, root cause currently unknown:

```
(x86 Linux) $ conda list mummer
# packages in environment at /home/pjacock/miniforge3/envs/pyani-plus_py312:
#
# Name                    Version                   Build  Channel
mummer                    3.23            pl5321hdbdd923_19    bioconda
(x86 Linux) $ rm -rf tmp* ; nucmer -p tmp NC_002696.fasta NC_014100.fna ; echo "Return $?"
1: PREPARING DATA
2,3: RUNNING mummer AND CREATING CLUSTERS
# reading input file "tmp.ntref" of length 4016949
# construct suffix tree for sequence of length 4016949
# (maximum reference length is 18446744073709551612)
# (maximum query length is 18446744073709551615)
# process 40169 characters per dot
#....................................................................................................
# CONSTRUCTIONTIME /home/pjacock/miniforge3/envs/pyani-plus_py312/opt/mummer-3.23/mummer tmp.ntref 1.22
# reading input file "/home/pjacock/repositories/pyani-plus/tests/fixtures/sequences/NC_014100.fna" of length 4655622
# matching query-file "/home/pjacock/repositories/pyani-plus/tests/fixtures/sequences/NC_014100.fna"
# against subject-file "tmp.ntref"
ERROR: mummer and/or mgaps returned non-zero
Return 1
```

Reverting to build 18 works:

```
(x86 Linux) $ conda list mummer
# packages in environment at /home/pjacock/miniforge3/envs/pyani-plus_py312:
#
# Name                    Version                   Build  Channel
mummer                    3.23            pl5321hdbdd923_18    bioconda
(x86 Linux) $ rm -rf tmp* ; nucmer -p tmp NC_002696.fasta NC_014100.fna ; echo "Return $?"
[...]

PR #50092, opened 2024-08-15
TEST ONLY: test reverting mummer recipe

aliciaaevans:
No intention to merge this at the moment.

Reverting the last set of changes just for this PR to see if the linux build works again.

See https://github.com/bioconda/bioconda-recipes/issues/49899


aliciaaevans:

aliciaaevans:

aliciaaevans:

aliciaaevans:
The sed line actually never gets executed on linux. The error is "file: command not found" because the container doesn't have that command. However it does look like the link and permissions lines are happening and when I build locally on linux I have file and still ended up with a bad build from what's in master.

So far all these Linux builds have worked, though.

aliciaaevans:

PR #50131, opened 2024-08-17
Add a recipe for polap v0.2.4

goshng:
Add a recipe for polap v0.2.4

PR #50134, opened 2024-08-17
Add a recipe for polap-fmlrc v0.2.4 b1

goshng:
Add a recipe for polap-fmlrc v0.2.4 b1

### DIFF
--- a/recipes/polap-fmlrc/meta.yaml
+++ b/recipes/polap-fmlrc/meta.yaml
@@ -11,7 +11,7 @@ source:
   sha256: "{{ sha256 }}"
 
 build:
-  noarch: python
+  noarch: generic
   number: 0
   run_exports:
     - {{ pin_subpackage('polap-fmlrc', max_pin="x.x") }}
--- a/recipes/polap-fmlrc/meta.yaml
+++ b/recipes/polap-fmlrc/meta.yaml
@@ -19,9 +19,9 @@ build:
 requirements:
   run:
     - python =2.7
-    - bioconda::ropebwt2 >=r187
-    - bioconda::fmlrc >=1.0.0
-    - conda-forge::parallel
+    - ropebwt2 >=r187
+    - fmlrc >=1.0.0
+    - parallel
 
 test:
   commands:
@@ -29,7 +29,7 @@ test:
 
 about:
   home: https://github.com/goshng/polap
-  license: GPLv3
+  license: GPL-3.0-or-later
   license_family: GPL
   license_file: LICENSE
   summary: "POLAP: plant organelle long-read assembly pipeline"
--- a/recipes/polap-fmlrc/meta.yaml
+++ b/recipes/polap-fmlrc/meta.yaml
@@ -1,0 +1,41 @@
+{% set name = "polap-fmlrc" %}
+{% set version = "0.2.4" %}
+{% set sha256 = "4463d4a71a8a3273108a0ab2dff0a4c37a565c05af6ee17ee83dc62404294caf" %}
+
+package:
+  name: "{{ name }}"
+  version: "{{ version }}"
+
+source:
+  url: https://github.com/goshng/polap/archive/refs/tags/{{ version }}.tar.gz
+  sha256: "{{ sha256 }}"
+
+build:
+  noarch: python
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('polap-fmlrc', max_pin="x.x") }}
+
+requirements:
+  run:
+    - python =2.7
+    - bioconda::ropebwt2 >=r187
+    - bioconda::fmlrc >=1.0.0
+    - conda-forge::parallel
+
+test:
+  commands:
+    - polap
+
+about:
+  home: https://github.com/goshng/polap
+  license: GPLv3
+  license_family: GPL
+  license_file: LICENSE
+  summary: "POLAP: plant organelle long-read assembly pipeline"
+  dev_url: https://github.com/goshng/polap
+  doc_url: https://goshng.github.io/polap/polap.html
+
+extra:
+  recipe-maintainers:
+    - goshng
--- a/recipes/polap-fmlrc/LICENSE
+++ b/recipes/polap-fmlrc/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
+ of this license document, but changing it is not allowed.
+
+                            Preamble
+
+  The GNU General Public License is a free, copyleft license for
+software and other kinds of works.
+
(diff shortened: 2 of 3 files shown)


recipes/polap-fmlrc/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+build:
+  noarch: python
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('polap-fmlrc', max_pin="x.x") }}
+
+requirements:
+  run:
+    - python =2.7
```
bgruening, line 21: Python2? That does not look good :-(

recipes/polap-fmlrc/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - bioconda::ropebwt2 >=r187
```
bgruening, line 22: Please do not specify channels here

recipes/polap-fmlrc/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - bioconda::fmlrc >=1.0.0
+    - conda-forge::parallel
+
+test:
+  commands:
+    - polap
+
+about:
+  home: https://github.com/goshng/polap
+  license: GPLv3
```
bgruening, line 32: Please use spdx identifiers

recipes/polap-fmlrc/meta.yaml
bgruening: I'm not sure we should add new recipes with such old dependencies. What is the use case here?

recipes/polap-fmlrc/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+package:
+  name: "{{ name }}"
+  version: "{{ version }}"
+
+source:
+  url: https://github.com/goshng/polap/archive/refs/tags/{{ version }}.tar.gz
+  sha256: "{{ sha256 }}"
+
+build:
+  noarch: python
```
bgruening, line 14: You include R packages as well, it's not a noarch python package, rather a generic one I think.

bgruening (commented):
Thanks @goshng ... I'm not sure about the python 2 dependency and if we should add such old tools. Python2 is not maintained since years

recipes/polap-fmlrc/meta.yaml
goshng: Thank you for the reply. I am using a short-read polishing tool called FMLRC version 1 and its a package called kbchoi::msbwt, which is not part of bioconda and needs python 2.7.

recipes/polap-fmlrc/meta.yaml
goshng: Let me see ... so I should not append the channel name to a bioconda and conda-forge package. Right?

recipes/polap-fmlrc/meta.yaml
goshng: Let me study what spdx identifiers are.

recipes/polap-fmlrc/meta.yaml
goshng: I am editing the recipe.

recipes/polap-fmlrc/meta.yaml
goshng: I got it. GPL-3.0-or-later should be one of such spdx identifier for a license, right?

recipes/polap-fmlrc/meta.yaml
goshng: This one does not use R package. I known it is confusing. I have other recipe that needs R packages.

recipes/polap-fmlrc/meta.yaml
goshng: I know that this is somewhat abusing bioconda. I could not add kbchoi::msbwt package in the recipe. Instead, users could install polap-fmlrc and install kbchoi::msbwt separately.

goshng:
> Thanks @goshng ... I'm not sure about the python 2 dependency and if we should add such old tools. Python2 is not maintained since years

Thank you for reviewing the recipe. I have other recipe very similar one available at https://github.com/bioconda/bioconda-recipes/pull/50131 Could you review that one too? Your comments will be equally appreciated because the two recipes are basically from the same github repository but with a different bioconductor package. I have a part of script that needs python 2.7 so I want to have this recipe polap-fmlrc, and most of the script need latest python version, so I made the other recipe for polap. Thank you!

recipes/polap-fmlrc/meta.yaml
goshng: I have corrected them. Thank you.

recipes/polap-fmlrc/meta.yaml
goshng: I hope that now it is better. Please let me know if I need to edit the license part more. Thank you.

bgruening: Yes

recipes/polap-fmlrc/meta.yaml
bgruening: Yes

recipes/polap-fmlrc/meta.yaml
bgruening: Sorry, we can only include packages from conda-forge or bioconda

recipes/polap-fmlrc/meta.yaml
goshng: Actually, I wanted to have recipes only for linux-64 architecture because I have not tested my script with mac-osx. The script is intended to work with only Linux 64bit architecture because of its dependency and its data to deal with. I do not know that it is working with Mac OSX, which I guess not. Thank you.

recipes/polap-fmlrc/meta.yaml
goshng: Thank you

recipes/polap-fmlrc/meta.yaml
goshng: Thank you

recipes/polap-fmlrc/meta.yaml
goshng: Thank you

goshng:
The following is from the installation instruction I wanted to have for the github project.

You could install polap bioconda package in a new conda environment, and then
run [Polap](https://github.com/goshng/polap) with a test dataset.

    (base) $ conda create --name polap bioconda::polap
    (base) $ conda activate polap
    (polap) $ git clone https://github.com/goshng/polap.git
    (polap) $ cd test
    (polap) $ polap assemble --test

Because FMLRC's conda environment was incompatible with that of Polap's,
we need to create one for Polap's FMLRC conda environment.

    (base) $ conda create --name polap-fmlrc bioconda::polap-fmlrc
    (base) $ conda activate polap-fmlrc
    (polap-fmlrc) $ conda install -y kbchoi::msbwt">=0.2.9"
    (polap-fmlrc) $ cd test
    (polap-fmlrc) $ polap prepare-polishing
    (polap-fmlrc) $ polap polish

PR #50270, opened 2024-08-22
art: enable osx-arm64 build

aryarm:
Enable osx-arm64 builds for ART

For some reason, I had to manually check the host and alter the `./configure` command to cross-compile for ARM ([as suggested here](https://stackoverflow.com/a/54070358/16815703)). I hope that's ok.


aryarm:
@BiocondaBot please add label

martin-g (approved):

PR #50314, opened 2024-08-25
Update Inspector to v1.3

czc:
We have updated Inspector to v1.3 with bugs fixed and new functions added. Thank you.

### DIFF
--- a/recipes/inspector/meta.yaml
+++ b/recipes/inspector/meta.yaml
@@ -1,4 +1,9 @@
-{% set version = "1.2" %}
+{% set version = "1.3" %}
+
+channels:
+  - conda-forge
+  - bioconda
+  - defaults
 
 package:
   name: inspector
@@ -6,7 +11,7 @@ package:
 
 source:
   url: https://github.com/ChongLab/Inspector/archive/refs/tags/inspector-v{{ version }}.tar.gz
-  sha256: 6d7ca6b22053d06105a549a064cf9a26efaf5c4a2a1a4502f2f36a4eb248ec4a
+  sha256: 6e1278de333990005a720a85048581783040ed4473b2247d0aa0eeaa8cb9ffda
 
 build:
   number: 0


recipes/inspector/build.sh
```diff
@@ -5,3 +5,5 @@ mkdir -p $PREFIX/bin
 
 cp *.py $PREFIX/bin
 chmod +x *.py
+
+exit 0
```
martin-g, line 9: This is not needed.

recipes/inspector/meta.yaml
```diff
@@ -1,16 +1,23 @@
-{% set version = "1.2" %}
+{% set version = "1.3" %}
+
+channels:
```
martin-g, line 3: This is not needed too. 
The defaults channel is paid since recently.

recipes/inspector/meta.yaml
```diff
@@ -1,16 +1,23 @@
...
 source:
   url: https://github.com/ChongLab/Inspector/archive/refs/tags/inspector-v{{ version }}.tar.gz
-  sha256: 6d7ca6b22053d06105a549a064cf9a26efaf5c4a2a1a4502f2f36a4eb248ec4a
+  sha256: 6e1278de333990005a720a85048581783040ed4473b2247d0aa0eeaa8cb9ffda
 
 build:
   number: 0
   noarch: generic
+  run_exports:
+    - {{ pin_subpackage('inspector', max_pin="x.x") }}
```
martin-g, line 20: Are you sure the max_pin value is "x.x"?

martin-g (changes requested):

recipes/inspector/build.sh
czc: Thanks, Martin. Noticed a "bash exit 1" error so I made this change. I will keep it as is if no potential issues.

recipes/inspector/meta.yaml
czc: Please help clarify. Is only the defaults channel not needed or are all these channels not needed? We found that we cannot install the package via conda if without the channels. We had to use "conda install -c conda-forge -c bioconda -c defaults inspector" to successfully install it.

recipes/inspector/meta.yaml
czc: We will be releasing versions like 1.2, 1.3, 1.4, 2.1, 3.2, etc. Will max_pin="x.x" suffice? Thanks.

recipes/inspector/build.sh
martin-g: OK, but it is useless. 
The `exit 1` must have been due to some error earlier

recipes/inspector/meta.yaml
martin-g: Bioconda builds use only `conda-forge` and `bioconda` channels. It is hardcoded.
AFAIK the `channels` property is not used at all.

martin-g: If you need `-c defaults` then you depend on non-free package.
But the build is green, so I guess you don't really need `defaults`

recipes/inspector/meta.yaml
martin-g: See https://github.com/bioconda/bioconda-recipes/blob/master/.github/PULL_REQUEST_TEMPLATE.md#instructions-for-avoiding-api-abi-and-cli-breakage-issues
If you follow Semantic versioning then you should use `"x"`

recipes/inspector/build.sh
czc: Thanks, Martin. I have removed this line.

recipes/inspector/meta.yaml
czc: Good to know. I have removed all the channels lines.

recipes/inspector/meta.yaml
czc: Thanks again. I have changed it to "x".

martin-g (approved):

PR #50362, opened 2024-08-27
Update telseq/meta.yaml: add samtools to decode CRAM file

lindenb:
telseq cannot read CRAM files. Samtools is added to read CRAM Files ( see: https://github.com/zd1/telseq/issues/26#issuecomment-704019482 )


bgruening:
If I understand the link correctly samtools is not a dependency of telseq. It can be used in conjunction, but this is true for a lot of software. I think you would need to install samtools in the same environment as telseq, but I wouldn't make it a dependency.

lindenb:
@bgruening true, but in the context of a standalone module for nf-core, samtools would be required.

lindenb:
@bgruening  I see. Thanks for the link. Closing

PR #50372, opened 2024-08-28
GW v1.0.3 patch

kcleal:
The patch fixes a bug in opengl context creation api, and adds additional libs


kcleal:
@BiocondaBot please add label

mencian (approved):

PR #50388, opened 2024-08-28
ci: Fix additional platform script

aliciaaevans:
The extra logging added in https://github.com/bioconda/bioconda-recipes/pull/50325 is great, but the CI was using the echoed output. Some broken osx-arm64 builds got merged to master (found them on bulk). 

Updated this to use the exit code instead. I'm not too thrilled with this solution, so open to suggestions.

(draft because I need to test if it will fail with the exit code from the script)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,9 +23,9 @@ jobs: # a basic unit of work in a run
       - run:
           name: Check for Additional Platforms
           command: |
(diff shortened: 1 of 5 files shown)


aliciaaevans:
Yup that didn't work. I think that's why I was using echo.

martin-g:
@aliciaaevans I see you reverted some of the problematic changes with https://github.com/bioconda/bioconda-recipes/pull/50389
Feel free to revert more if needed!
Without my last changes it is a bit hard to figure out what is the problem when there is such but not impossible. So, I am fine to revert everything that breaks the old behavior!

PR #50443, opened 2024-09-02
Update beav to v1.4.0

alexweisberg:
Updated Beav to version 1.4.0 and defined a key environment variable in the meta.yaml recipe.


alexweisberg:
@BiocondaBot please add label

PR #47397, opened 2024-04-23
Update freebayes to 1.3.8

BiocondaBot:
Update [`freebayes`](https://bioconda.github.io/recipes/freebayes/README.html): **1.3.7** &rarr; **1.3.8**

[install with bioconda](http://bioconda.github.io/recipes/freebayes/README.html) [Conda](https://anaconda.org/bioconda/freebayes/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/freebayes`](https://github.com//bioconda/bioconda-recipes/tree/bump/freebayes/recipes/freebayes) (click to view/edit other files)
Summary | Bayesian haplotype-based polymorphism discovery and genotyping
Home | [https://github.com/freebayes/freebayes](https://github.com/freebayes/freebayes)
Releases |[https://github.com/freebayes/freebayes/tags](https://github.com/freebayes/freebayes/tags)
Recipe Maintainer(s) | @jpuritz
Author | `@freebayes`
***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

PR #50531, opened 2024-09-06
Update vcflib

fplazaonate:
Current version of vcflib does not work with CPUs that do not support AVX extensions.
This new recipe disables march=native optimisations by unsetting the EXTRA_FLAGS variable in CMakeLists.txt

https://github.com/vcflib/vcflib/blob/master/CMakeLists.txt#L73

### DIFF
--- a/recipes/vcflib/build.sh
+++ b/recipes/vcflib/build.sh
@@ -41,6 +41,6 @@ cmake -S . -B build \
 	-DCMAKE_BUILD_TYPE=Release \
 	-DCMAKE_INSTALL_PREFIX="${PREFIX}" \
 	-DCMAKE_CXX_FLAGS="${CXXFLAGS}" \
-	-DEXTRA_FLAGS=""
+	-DEXTRA_FLAGS="" # Disable march=native to build a generic executable
 
 cmake --build build/ --target install -j 4 -v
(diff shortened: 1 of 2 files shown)


fplazaonate:
@BiocondaBot please add label

bgruening:
Can you please link to an upstream issue, so we can track when this issue is solved?

fplazaonate:
It is actually discussed in this PR
https://github.com/bioconda/bioconda-recipes/pull/47397

bgruening:
Sorry, I was talking about an upstream issues. So this does not get lost. And maybe then add a comment to the build.sh, so that we can revert this change ones its fixed upstream.

fplazaonate:
Ok. It is reported here:
https://github.com/vcflib/vcflib/issues/401
I will add the comment line in the build script.

recipes/vcflib/build.sh
```diff
@@ -40,6 +40,7 @@ cmake -S . -B build \
 	-DZIG=ON -DOPENMP=ON -DWFA_GITMODULE=OFF \
 	-DCMAKE_BUILD_TYPE=Release \
 	-DCMAKE_INSTALL_PREFIX="${PREFIX}" \
-	-DCMAKE_CXX_FLAGS="${CXXFLAGS}"
+	-DCMAKE_CXX_FLAGS="${CXXFLAGS}" \
+	-DEXTRA_FLAGS="" # Disable march=native to build a generic executable
```
bgruening, line 44:
```suggestion
	-DEXTRA_FLAGS="" # Disable march=native to build a generic executable, see https://github.com/vcflib/vcflib/issues/401
```

bgruening:
Thanks a lot!

PR #50564, opened 2024-09-09
Update vcflib 

fplazaonate:
My previous attempt to disable "march=native" by unsetting EXTRA_FLAGS from the command line failed.
I guess this is because variable set via the command line are defined before those in CMakeLists.txt.
https://github.com/bioconda/bioconda-recipes/pull/50531

This new recipe directy remove EXTRA_FLAGS  from the CMakeLists.txt
I have built this recipe and I have I checked that the "march=native" flag does not appear anymore.

### DIFF
--- a/recipes/vcflib/build.sh
+++ b/recipes/vcflib/build.sh
@@ -40,7 +40,6 @@ cmake -S . -B build \
 	-DZIG=ON -DOPENMP=ON -DWFA_GITMODULE=OFF \
 	-DCMAKE_BUILD_TYPE=Release \
 	-DCMAKE_INSTALL_PREFIX="${PREFIX}" \
-	-DCMAKE_CXX_FLAGS="${CXXFLAGS}" \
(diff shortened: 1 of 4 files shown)


fplazaonate:
@bgruening could you approve this?
Sorry for the waste of time.

fplazaonate:
@BiocondaBot please add label

bgruening:

bgruening:
Can you test those binaries please?

fplazaonate:
I have built the library on a machine with a CPU that supports AVX extensions and installed it on a machine without.
Before the installation (i.e using vcflib-1.0.9-hdcf5f25_7 from bioconda), freebayes crashes with "Illegal instruction" error. After the installation (i.e using vcflib-1.0.9-h2bc3f7f_8 built locally), freebayes finishes normally.

objdump confirms that there are no avx instructions:
```
objdump -d  lib/libvcflib.so.1 | grep -Ec '\bv(add|sub|mul|div|mov|shuf|and|or|xor|perm|blend|cmp)'
0
```

mencian (approved):

PR #50652, opened 2024-09-12
restrict mysql dependency

Maarten-vd-Sande:
Applies: https://github.com/vanheeringen-lab/genomepy/pull/250


martin-g:
You need to bump `build > number` to `2`

PR #50701, opened 2024-09-13
Add osx-arm as build target for dysgu

kcleal:
Updated the meta.yaml to add osx-arm


mencian:
Closing in favor of #55263.

PR #50711, opened 2024-09-13
UPDATE: remove old dendropy pin

lizgehret:
As of sepp 4.5.5, there is no longer a dependency on dendropy 4.5.2 (builds are currently up on [bioconda](https://anaconda.org/bioconda/sepp/files)) - so this max pin can be dropped, which will allow for greater compatibility of busco with other more modern packages.

### DIFF
--- a/recipes/busco/meta.yaml
+++ b/recipes/busco/meta.yaml
@@ -30,7 +30,6 @@ requirements:
     - biopython >=1.79
     - prodigal
     - sepp >=4.3.10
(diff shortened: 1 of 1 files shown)


martin-g:
You need to increase the value of `build > number` with 1

lizgehret:
> You need to increase the value of `build > number` with 1

Got it, thank you @martin-g - build number has been bumped to 1.

PR #51287, opened 2024-10-09
add gfainject

AndreaGuarracino:
This is a pangenome tool to inject mappings/alignments (in BAM and PAF formats) into pangenome graphs (in GFA format).


AndreaGuarracino:
@BiocondaBot please add label

martin-g (approved):

PR #51343, opened 2024-10-11
Add `cosigt`

AndreaGuarracino:
Cosigt computes the cosine similarity between pangenome-based coverage vectors that can be used to genotype read samples.


AndreaGuarracino:
@BiocondaBot please add label

martin-g (approved):

PR #51551, opened 2024-10-22
Treekin

mwaldl:
Add runtime dependency libgfortran3 to meta,yaml file. This resolves the runtime error "treekin: error while loading shared libraries: libgfortran.so.3: cannot open shared object file: No such file or directory". 

Further clean up might be needed, including adding libgfortran as dependency during build.

### DIFF
--- a/recipes/treekin/meta.yaml
+++ b/recipes/treekin/meta.yaml
@@ -31,7 +31,7 @@ requirements:
   run:
     - mlapack
     - lapack <3.9
-    - libgfortran =3
+    - {{ compiler('fortran') }}
 
 test:
   commands:
--- a/recipes/treekin/meta.yaml
+++ b/recipes/treekin/meta.yaml
@@ -10,7 +10,7 @@ build:
   skip: True # [osx]
   number: 5
   run_exports:
-    - {{ pin_subpackage('myrecipe', max_pin="x.x") }}
+    - {{ pin_subpackage(name, max_pin="x.x") }}
 
 source:
   url: https://github.com/ViennaRNA/Treekin/releases/download/v{{ version }}/Treekin-{{ version }}.tar.gz 
--- a/recipes/treekin/meta.yaml
+++ b/recipes/treekin/meta.yaml
@@ -9,6 +9,8 @@ package:
 build:
   skip: True # [osx]
   number: 5
+  run_exports:
+    - {{ pin_subpackage('myrecipe', max_pin="x.x") }}
 
 source:
   url: https://github.com/ViennaRNA/Treekin/releases/download/v{{ version }}/Treekin-{{ version }}.tar.gz 
--- a/recipes/treekin/meta.yaml
+++ b/recipes/treekin/meta.yaml
@@ -8,7 +8,7 @@ package:
 
 build:
   skip: True # [osx]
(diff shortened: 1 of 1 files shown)


martin-g:
@bgruening @aliciaaevans This PR fails with:
```
13:51:51 [36mBIOCONDA DEBUG[0m Found: LintMessage(recipe=Recipe "treekin", check=<class 'bioconda_utils.lint.check_build_help.should_use_compilers'>, severity=<Severity.ERROR: 30>, title='The recipe requires a compiler directly', body="\nSince version 3, `conda-build` uses a special syntax to require\ncompilers for a given language matching the architecture for which\na package is being build. Please use:\n\n    requirements:\n       build:\n         - {{ compiler('language') }}\n\nWhere language is one of `c`, `cxx`, `fortran`, `go` or\n`cgo`. You can specify multiple compilers if needed.\n\nThere is no need to add `libgfortran`, `libgcc`, or\n`toolchain` to the dependencies as this will be handled by\nconda-build itself.", start_line=33, end_line=34, fname='recipes/treekin/meta.yaml', canfix=False)[0m
```
but the recipe properly uses `- {{ compiler("cxx")` and `- {{ compiler("fortran")`.

In addition the linting error probably should mention `rust` too.

mwaldl:
@martin-g @bgruening @aliciaaevans Does the linter fail because  libgfortran is defined as a runtime dependency?
The build dependency is properly defined as - {{ compiler("fortran") }}, however it appears to also be required at runtime. Any suggestions on how to resolve this?

recipes/treekin/meta.yaml
```diff
@@ -8,7 +8,9 @@ package:
 
 build:
   skip: True # [osx]
-  number: 4
+  number: 5
+  run_exports:
+    - {{ pin_subpackage('myrecipe', max_pin="x.x") }}
```
martin-g, line 13:
```suggestion
    - {{ pin_subpackage(name, max_pin="x.x") }}
```

martin-g:
@mwaldl According to https://github.com/bioconda/bioconda-utils/blob/5910a1009b8b783eb4a3e1a9b15ab71b9794899a/bioconda_utils/lint/check_build_help.py#L13-L47 it is not allowed to use `libgfortran` as a dependency.
Let's try with `- compiler("fortran")` as a run dependency!

recipes/treekin/meta.yaml
```diff
@@ -29,6 +31,7 @@ requirements:
   run:
     - mlapack
     - lapack <3.9
+    - libgfortran =3
```
martin-g, line 34:
```suggestion
    - {{ compiler('fortran') }}
```

martin-g:
Now it fails with
```
09:30:20 BIOCONDA DEBUG Found: LintMessage(recipe=Recipe "treekin", check=<class 'bioconda_utils.lint.check_build_help.compilers_must_be_in_build'>, severity=<Severity.ERROR: 30>, title='The recipe requests a compiler in a section other than build', body="\nPlease move the `{{ compiler('language') }}` line into the\n`requirements: build:` section.", start_line=33, end_line=34, fname='recipes/treekin/meta.yaml', canfix=False)
```

PR #51729, opened 2024-10-28
CI: Don't rename image filename

aliciaaevans:
Remove the line that gets rid of `:` in the filename since bioconda-utils uses that to get the image version.


aliciaaevans:
ok that's not going to work because it actually throws an error. Got to change it on the bioconda-utils side.

PR #51830, opened 2024-10-31
Update PhaBOX to 2.1.4

KennthShang:
Update PhaBOX to the 2.1.3 version and add new requirements


KennthShang:
@BiocondaBot please add label

KennthShang:
@bgruening  Sorry for bothering. 

I wonder if it will take a long time to test and if I will need a new approval every time I revise the script. Thanks.

Also, it seems I can easily pass the conda build and upload my package to the Anaconda cloud (my personal account), but I usually fail the Bioconda test. Do you have any suggestions?

Best,
Jiayu

recipes/phabox/meta.yaml
```diff
@@ -1,61 +1,66 @@
...
-    - networkx >=3.4
-    - numpy >=1.26
-    - pandas >=2
-    - prodigal-gv >=2.11.0
-    - pyarrow >=16
-    - pytorch >=2.4
-    - scipy >=1.14
-    - seaborn-base >=0.13.2
-    - transformers >=4
+    - python =3.10.14
```
bernt-matthias, line 26: I thinks such a strong pinning is not needed for most of the used software.

bernt-matthias:
> I wonder if it will take a long time to test and if I will need a new approval every time I revise the script. Thanks.

Yes, an approval is needed for every pull request. 

> Also, it seems I can easily pass the conda build and upload my package to the Anaconda cloud (my personal account), but I usually fail the Bioconda test. Do you have any suggestions?

Bioconda has stricter testing. Many people and projects only use bioconda and conda-forge, because these are quite strictly tested and well maintained sources of software. In my daily practice I would never install software from other conda channels. 

bioconda comes with some additional effort, but it certainly pays off on the longer run (IMO).

bernt-matthias:
I restored the old version pinning (and re-changed the order of the requirements .. which makes it easier to track changes). Lets see if this works. The strict pinning of your commits could not be resolver by conda (wrt the pinning given by the bioconda ecosystem)

KennthShang:
Thanks for the advice. Since it still failed to pass the test, I made some revisions. 

I think the major reason is the version of the deep learning packages.  I left the version as blank to avoid conflict when building my packages, so let's see whether it works here.

KennthShang:
I got a very strange error during the test: `PermissionError: [Errno 13] Permission denied: '.tmpcl7hxzxv'`

So I revised script and try again

KennthShang:
I removed all the unnecessary files in the release and all the os.system() are replaced by subprocess.run(). Provides standard error and return a non-zero exit code in case one of the calls fails in this version

KennthShang:
Seems the `no space` error still occurs. I have to remove some packages from the receipt and let the user download them independently. Hope this time the it will work

bernt-matthias:
Seems not to be the case. The zip is only 6MB, but a few pyc files might be removed: https://github.com/KennthShang/PhaBOX/issues/40

KennthShang:
@bernt-matthias I have removed them now.

@bgruening Do not know why, but after I followed the original rules (receipt from @bernt-matthias ) and added new packages below there, it worked well [#51958](https://github.com/bioconda/bioconda-recipes/pull/51958). So maybe it is caused by the downloaded dependencies during the build.

PR #51910, opened 2024-11-04
Add linux-aarch64 and/or osx-arm64 to various perl-* packages 1 of 2

dslarm:
This breaks PR #51856 up to smaller pieces to get past build-timeout due to size of package list.
This adds osx-arm64 and linux-aarch64 where the packages would already work but for lack of meta.yaml or historic missing dependency in conda-forge.


dslarm:
@martin-g -g

PR #52091, opened 2024-11-13
Add osx-arm64 to r-facets

dslarm:
The r-facets package compiles locally on osx-arm64 - committing this to CI now to test.
[was also r-spp, but this failed CI]


dslarm:
@martin-g can you review?

PR #52152, opened 2024-11-17
perl-io-compress linux aarch64 support

dslarm:
This adds support for linux-aarch64 for perl-io-compress.
Dependencies in conda required by perl-io-compress are now available.
~~This also tests whether osx-arm64 will also build.~~
osx-arm64 fails due to missing perl-compress-raw-bzip2 and perl-scalar-list-utils


dslarm:
@martin-g

PR #52385, opened 2024-11-26
Add linux aarch64 to blast

dslarm:
* This PR enables linux-aarch64 for blast.
* Update the meta.yaml note on missing package requirements for osx-arm64


dslarm:
This is ready to merge - @martin-g can you review? 

The timeout in building Linux-aarch64 was not due to compilation or test speed, but due to failure caused by the 'which' bug.

mencian (approved):
Thanks for looking into this, @dslarm!

PR #52396, opened 2024-11-27
linux-aarch64 control-freec support

dslarm:
Add linux-aarch64 to control-freec.

The use of 'which' is replaced by test -x in the meta.yaml due to https://github.com/conda-forge/which-feedstock/issues/5 which will may take some time to fully resolve given it will impact many packages, so this is a safe alternative.


dslarm:
@martin-g - can you review?

mencian (approved):

PR #52413, opened 2024-11-27
rsem: add linux-aarch64 support and document missing osx-arm64 dependencies.

dslarm:
Added linux-aarch64 and locally tested.
Added zlibz as a dependency for build and run - lack of it was failing the build on linux-aarch64.


rob-p (commented):
This looks good to me now!

rob-p (approved):

PR #52438, opened 2024-11-28
perl-dbd-* and perl-lwp-simple: add linux-aarch64 and partial osx-arm64 support

dslarm:
Adds perl-dbd-* and perl-lwp-simple, documents requirement missing for osx-arm64 where currently failing


dslarm:
@mencian - can you approve?

mencian (approved):

Issue #30370, opened 2021-09-02
Python version range in recipe not respected in package


jlumpe:
The [recipe](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/hesslab-gambit/meta.yaml) for my package includes `python >=3.7` in both the `host` and `run` subsections of `requirements`, but the [built package](https://bioconda.github.io/recipes/hesslab-gambit/README.html#package-hesslab-gambit) says Python 3.9 is required. Is my recipe incorrect somehow?

jlumpe:
Awesome, thanks @marcelm. `conda search` now shows me builds for 3.7-3.9. Should this maybe be added to the contributor documentation?

marcelm:
Yes, absolutely. I’ll make a PR when I find some time for this (unless anyone else wants to have a go). Perhaps this should even be a lint check.

PR #52501, opened 2024-12-01
Update gambit to 1.1.0

jlumpe:
Update gambit to 1.1.0

### DIFF
--- a/recipes/gambit/meta.yaml
+++ b/recipes/gambit/meta.yaml
@@ -19,13 +19,13 @@ build:
 requirements:
   build:
     - {{ compiler('c') }}
+    - libgomp      # [linux]
+    - llvm-openmp  # [osx]
   host:
     # Biopython not currently available for 3.13
     - python >=3.9,<3.13
     - pip
     - cython >=3
-    - libgomp      # [linux]
-    - llvm-openmp  # [osx]
   run:
     - python
     - numpy >=1.13,<2
--- a/recipes/gambit/meta.yaml
+++ b/recipes/gambit/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "1.0.1" %}
+{% set version = "1.1.0" %}
 
 package:
   name: gambit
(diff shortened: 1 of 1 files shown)


jlumpe:
@BiocondaBot please add label

jlumpe:
Small question for someone more experienced than me with Conda recipes: I added dependencies for OpenMP shared libraries (`libomp` / `llvm-openmp`) (not sure why the recipe was working for previous versions without them, probably it was the update to Cython 3). I followed [this section](https://conda-forge.org/docs/maintainer/knowledge_base/#openmp) of conda-forge's maintainer documentation.

I added the dependencies under the `host` section, but then realized that the linked source says to add them under the `build` section. I'm not very clear about the difference between the two. Could someone tell me if it's good as-is or should be changed?

johanneskoester:
From the conda docs (about the host section, see https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#host):

> This section was added in conda-build 3.0. It represents packages that need to be specific to the target platform when the target platform is not necessarily the same as the native build platform.

The conda-forge docs suggest to me that openmp will be just needed at compile time. Hence, it belongs into build since it cannot really be specific to the target platform then.

johanneskoester (approved):

martin-g:
> I added the dependencies under the `host` section, but then realized that the linked source says to add them under the `build` section. I'm not very clear about the difference between the two. Could someone tell me if it's good as-is or should be changed?

My understanding is that those should be in `host`. Only build `tools` should be in `build`. All library dependencies should be in `host`.
I have seen other recipes with both ways.
IMO this is not a stopper to merge the PR!

jlumpe:
@johanneskoester @martin-g Hoping you can help with something else. I'm seeing 1.1.0 as available now but only builds for Python 3.12. The package's [page](https://bioconda.github.io/recipes/gambit/README.html) on the Bioconda website lists the minimum Python version as 3.12, when it should be 3.9. Any ideas about why this might be?

I was able to create an environment with Python 3.9 and the rest of the dependencies copied from the recipe, so I don't think it's a dependency restricting the minimum version.

martin-g:
I also don't have an idea why it selected just 3.12 here.

jlumpe:
I realized for 1.0.1 I didn't have version constraints on the `python` dependency directly and instead used:

```
build:
  skip: True  # [py < 37]
```

(when the minimum supported version was 3.7). This line got removed in #48485, which I don't really understand the reason for.

The current recipe has the Python version range in `host`, but no range in `run`. I'm having a hard time telling from the documentation what's the correct way to do it, and it seems like there is a lot of variation in existing recipes.

jlumpe:
Actually seems like I [asked](#30370) this exact question years ago when I first submitted the question and forgot about it... It seems the correct way to do it is indeed with the `skip: True`, the recipes doing it the other way might all be `noarch` packages.

Submitted a new PR #52571.

PR #52517, opened 2024-12-02
Manta: add linux-aarch64 builds

dslarm:
Adds linux-aarch64 builds by using source instead of binary.


dslarm:
@BiocondaBot please add label

recipes/manta/meta.yaml
```diff
@@ -6,17 +6,28 @@ package:
...
   run_exports:
     - {{ pin_subpackage(name, max_pin="x") }}
+  skip: True  # [osx] # until CI os/x is fixed https://github.com/bioconda/bioconda-recipes/issues/52553
 
 requirements:
+  build:
+    - cmake
+    - make
+    - {{ compiler('cxx') }}
+    - boost-cpp
```
martin-g, line 25: this should be in `host`

recipes/manta/meta.yaml
```diff
@@ -6,17 +6,28 @@ package:
...
+    - python ==2.7.*
```
martin-g, line 26: This is not needed in `build`. `host` should be enough

recipes/manta/meta.yaml
dslarm: That fails to lint:
```
The recipe requests a compiler in a section other than build
Please move the `{{ compiler('language') }}` line into the`requirements: build:` section
```

martin-g: Sorry, I meant just `boost-cpp`.
`build` should list only build **tools**. All software _libraries_ should be in `host`

dslarm: Moving boost-cpp seems to be better - just trying that

recipes/manta/meta.yaml
```diff
@@ -6,17 +6,28 @@ package:
...
   run_exports:
     - {{ pin_subpackage(name, max_pin="x") }}
+  skip: True  # [osx] # until CI os/x is fixed https://github.com/bioconda/bioconda-recipes/issues/52553
 
 requirements:
+  build:
+    - cmake
+    - make
+    - {{ compiler('cxx') }}
+    - python ==2.7.*
```
martin-g, line 25:
```suggestion
```
Interpreters should be in `host` too

martin-g (approved):

PR #52557, opened 2024-12-03
Samtools 0.1.19 and 0.1.18 - linux-aarch64 and osx-arm64 (0.1.18) support

dslarm:
Add support for legacy samtools - these are the only versions that work with perl-bio-samtools.


dslarm:
@BiocondaBot please add label

martin-g (approved):

PR #52576, opened 2024-12-04
perl-bio-samtools: bump samtools minor version and add linux-aarch64 and osx-arm64

dslarm:
Add linux-aarch64 and osx-arm64 (if tests pass)
Bump the required samtools to max 0.1.19 as https://metacpan.org/release/LDS/Bio-SamTools-1.43/source/Changes states:
```
Version 1.41
  * Added missing INSTALL.pl script.
  * Make it clear that this library only works up through
    Samtools 0.1.19.
  * Improvements to INSTALL.pl script and documentation.

```
Enables the 0.1.18 (osx-64) and 0.1.19 (linux-aarch64, build fails on osx-64) to be used.
Resolves: existing version required by perl-bio-samtools meta.yaml (0.1.17) no longer exists in bioconda - so is unbuildable (not a reported bug)


dslarm:
@BiocondaBot please add label

martin-g (approved):

PR #52771, opened 2024-12-12
maxbin2: start enabling linux-aarch64

dslarm:
- Replace perl-lwp-simple with perl-libwww-perl.
- perl-lwp-simple is blacklisted and only installs perl-libwww-perl these days, so let's cut out and go direct to perl-libwww-perl.
- perl-libwww-perl itself is in conda-forge - but required missing perl-mime-base64 for linux-aarch64 (https://github.com/conda-forge/perl-mime-base64-feedstock/pull/2).  Once that's merged and built, this PR should be retested.

### DIFF
--- a/recipes/maxbin2/meta.yaml
+++ b/recipes/maxbin2/meta.yaml
@@ -27,7 +27,7 @@ requirements:
     - bowtie2
     - hmmer
     - idba
-    - perl =5.26
+    - perl
     - perl-libwww-perl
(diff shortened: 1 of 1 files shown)


dslarm:
On linux-aarch64 we're waiting for a couple of packages to merge in conda-forge - perl-mime-base64 and perl-digest-md5.

martin-g (approved):

PR #52783, opened 2024-12-13
recipe for scROSHI package based on CRAN repo

vipints:
Trying to add scROSHI package to bioconda

### DIFF
--- a/recipes/r-scroshi/meta.yaml
+++ b/recipes/r-scroshi/meta.yaml
@@ -1,0 +1,54 @@
+{% set version = '1.0.0.0' %}
+
+package:
+  name: r-scroshi
+  version: {{ version|replace("-", "_") }}
+
+source:
+  url:
+    - {{ cran_mirror }}/src/contrib/scROSHI_{{ version }}.tar.gz
+    - {{ cran_mirror }}/src/contrib/Archive/scROSHI/scROSHI_{{ version }}.tar.gz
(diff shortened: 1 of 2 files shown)


recipes/r-scroshi/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  noarch: generic
+
+requirements:
+  host:
+    - r-base ==4.3.2
+    - bioconductor-s4vectors ==0.40.2
+    - bioconductor-singlecellexperiment ==1.24.0
+    - bioconductor-summarizedexperiment ==1.32.0
+    - bioconductor-limma ==3.58.1
+    - r-uwot ==0.1.16
```
bgruening, line 29: Are those strict pins really needed?

vipints: @bgruening, I have tested the installation with the above package versions. I am not quite sure whether the dependency conflict arises if this is not mentioned explicitly. Let me test that as well.

vipints: I tested and there are no version constraints for the packages but there is a minimum version for r-base: >=3.6.0

bgruening: Ok, then please remove those contraints :) Thanks.

bgruening (approved):

bgruening (approved):

PR #52859, opened 2024-12-17
Change osx-arm to osx-arm64 in additional targets

rob-p:
This PR changes 'osx-arm' to 'osx-arm64' in additional targets so as to build for the proper OSX arm target.


jmarshall:

jmarshall (approved):

PR #52897, opened 2024-12-18
Rename "handyreadgenotyper" to "amplicontyper"

AntonS-bio:
The original name turned out very unwieldy. AmpliconTyper has no obvious conflicts and reasonably conveys the purpose of the package.


recipes/amplicontyper/meta.yaml
```diff
@@ -1,26 +1,30 @@
...
   noarch: python
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
   run_exports:
-      - {{ pin_subpackage('handyreadgenotyper', max_pin="x.x") }}  
+      - {{ pin_subpackage('amplicontyper', max_pin="x.x") }}  
   number: 0
 
 requirements:
+  build:
+    - setuptools
```
martin-g, line 21: Are you sure this is needed?

AntonS-bio: Oddly yes, I didn't have it in previous releases and "conda-build" worked fine, but not this time. I've tried a lot of things and this is the only one removed the "module not found error",

martin-g: I think the one in `host` should be enough.
Let me build without the one in `build` and if it fails I will revert it!

martin-g (approved):

recipes/amplicontyper/meta.yaml
AntonS-bio: You are right, the one in host is enough. I've updated meta.yaml for future releases. Thanks for checking.

PR #52977, opened 2024-12-24
add r-rcppplanc dependency to r-liger

theAeon:
Finally got this one pushed through to conda-forge. The only reason it isn't a hard runtime dependency in DESCRIPTION is because CRAN was/is being glacial about approval. I figure as long as builds exist it makes sense to have a hard dependency on our core backend.


theAeon:
huh. its not finding the version i just merged (and neither is my local). I guess there's a lag-time on that?

theAeon:
@BiocondaBot rebuild

bgruening:
Thanks

PR #53124, opened 2025-01-08
Remove openssh requirement

pavanvidem:
Adding the openssh requirement here still could not overcome the error in the Galaxy tool which is related to ssh. Removing the dependency here and adding the requirement in the tool should hopefully fix it.


pavanvidem:
@bgruening this is the PR with the Galaxy tool https://github.com/galaxyproject/tools-iuc/pull/6212

PR #53141, opened 2025-01-08
fix bioconductor-rbowtie2

aliciaaevans:
Found this change to build.sh that was overwritten by the bioconductor skeleton builder


aliciaaevans:

dpryan79 (approved):

dpryan79:
Good catch. I used to have a list somewhere of weird bioconductor packages that required special modification. That's likely been lost to the ages :(

PR #53196, opened 2025-01-13
ngscheckmate - migrate to python3 and enable linux-aarch64 and osx-arm64 

dslarm:
* change python2 to python3 - largely tabulation and print syntax
* the ngscheckmate_fastq binary is now built by the build script; the source has been updated to be more modern C, and the installation no longer uses the linux-x86_64-only binary that was in this package.
* this package no longer marked as architecture generic.


dslarm:
@martin-g - could you review and merge please?

dslarm:
@BiocondaBot please add label

mencian:
Oops, the commit message should be `python >=2.7`.

PR #53207, opened 2025-01-14
Re-enable broken dialign-tx (dep of t-coffee, dep of pasta), add linux-aarch64

dslarm:
Use archive of dialign-tx (checksum unchanged); original download link is deceased.
 Add linux-aarch64


dslarm:
@BiocondaBot please add label

martin-g (approved):

PR #53412, opened 2025-01-22
ECTYPER v2.0.0 build #2 addressing the bioconda 1.85 incompatibility

kbessonov1984:
This is build #2 where bioconda v1.85 recent update resulted in errors reading FASTA files. This incompatibility was addressed in this build thanks to making source code compatible with both previous and recent biopython versions. 

This addresses issue #99 at https://github.com/phac-nml/ecoli_serotyping/issues/99


kbessonov1984:
@BiocondaBot please add label

PR #53444, opened 2025-01-23
`khmer` v3 does not work the same as v2 and is incompatible

dr-joe-wirth:
primerForge does not work with `khmer` v3. I do not know what has changed between versions, but it does not generate the same results it used to. `khmer` v2.1 does work. I have restricted the build to reflect this.

@mencian the recipes for both primerForge v1.5.2 and primerForge v1.5.3 are broken because of this incompatibility. It passes the bioconda recipe checks, but the unit tests fail.

### DIFF
--- a/recipes/primerforge/meta.yaml
+++ b/recipes/primerforge/meta.yaml
@@ -26,7 +26,7 @@ requirements:
     - python >=3.9,<3.12
     - biopython ==1.81
     - ispcr
-    - khmer >=2.1.1
+    - khmer >=2.1,<3
     - numpy
     - primer3-py >=2.0
     - pyahocorasick


dr-joe-wirth:

dr-joe-wirth:
i am going to run unit tests on this build to ensure it is working. i will review changes once i have confirmed it runs.

dr-joe-wirth:
ok this is what i expected. the version of `khmer` that is compatible with `primerForge` cannot be installed.

@mencian @rpetit3 there are two builds in master that do not work (1.5.2-0 and 1.5.3-0). how should i proceed?

mencian:
@dr-joe-wirth I think we'll have to give option 1 from [rpetit3's comment](https://github.com/bioconda/bioconda-recipes/pull/48898#issuecomment-2209454553) a try and build `khmer` v`2.1.1` on python `3.9-3.11`.

dr-joe-wirth:
@lskatz and I tried this last year and could not get it figured out. If you think you can do it I would be very grateful.

PR #53769, opened 2025-02-09
Add Star 2.7.3a

brainstorm:
Adds support for this specific STAR version since it's needed for an ongoing effort to port `oncoanalyser` pipeline to `linux-aarch64`.

This specific version is needed since latest versions generate runtime issues, as reported by Hartwig.

/cc @scwatts

### DIFF
--- a/recipes/star/2.7.3a/build.sh
+++ b/recipes/star/2.7.3a/build.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+export INCLUDES="-I${PREFIX}/include"
+export LIBPATH="-L${PREFIX}/lib"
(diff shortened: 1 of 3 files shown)


scwatts:
Amazing, thanks for this. I'll check the build artifacts run as expected on Linux AMD64 and hopefully ARM64 but let's see what we get back from @bioconda-bot

scwatts:

scwatts (approved):
These look good! I've checked the Conda packages by aligning simulated WTS reads to GRCh38 then comparing MD5 hashes of output alignments and flagstats for

- the existing build on AMD64,
- this build on AMD64, and
- this build on ARM64

No discrepancies were seen 🙌

martin-g (approved):

PR #53867, opened 2025-02-14
t-coffee: add versions of dependencies and revert the removal of arm64 support

luisas:
Reverts bioconda/bioconda-recipes#53845

This is a revert of a revert. 

We removed the changes introduced because it seeminlgy broke some downstream applications. 

We figured the issue was not the addition of the arm64 support but a change in some dependencies version (tmalign in concrete) which was then not compatible with t-coffee. 

Now all versions are explicitly provided.


luisas:
@martin-g Hi! 

Sorry to bother you again - here it goes the fix we talked about yesterday. 

The real issue was not the arm64 support but a dependency version being wrong. Now all the versions are specified.

Could you review it when you have some time? Thanks!!

luisas:
This is just the undo of the undo we did yesterday of the arm64 support + the versions specifications

luisas:
@BiocondaBot please add label

martin-g (approved):

PR #53951, opened 2025-02-18
GW v1.1.4

kcleal:
This lowers some of the requirements in meta.yaml to prevent conflicts during install


kcleal:
@BiocondaBot please add label

martin-g (approved):

PR #54022, opened 2025-02-21
Update grz-cli to 0.1.4

tedil:
Update grz-cli to 0.1.4 manually.
Also change source url path from pypi.org to pypi.io such that the autobumper can correctly pick up on changes.

### DIFF
--- a/recipes/grz-cli/meta.yaml
+++ b/recipes/grz-cli/meta.yaml
@@ -6,7 +6,7 @@ package:
   version: {{ version }}
 
 source:
-  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/grz_cli-{{ version }}.tar.gz
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/grz_cli-{{ version }}.tar.gz
   sha256: b79ae294576a0a8180bc9678392d5efc367b71b7f439d153d2dae5bd0009d96f
 
 build:
--- a/recipes/grz-cli/meta.yaml
+++ b/recipes/grz-cli/meta.yaml
@@ -1,13 +1,13 @@
 {% set name = "grz-cli" %}
-{% set version = "0.1.1" %}
+{% set version = "0.1.4" %}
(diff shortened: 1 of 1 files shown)


recipes/grz-cli/meta.yaml
```diff
@@ -1,13 +1,13 @@
...
+{% set version = "0.1.4" %}
 
 package:
   name: {{ name|lower }}
   version: {{ version }}
 
 source:
-  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/grz_cli-{{ version }}.tar.gz
-  sha256: 199dd5d96405065db6224d75e991dd385b28685bc7f581e75a8c09d913737fa7
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/grz_cli-{{ version }}.tar.gz
```
dlaehnemann, line 9:
```suggestion
  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/grz_cli-{{ version }}.tar.gz
```
Is the `.io` intentional here? Because I think it should be `.org`, based on these discussions:
* https://github.com/pypi/warehouse/issues/2184

dlaehnemann (approved):
Just one question/suggestion, but otherwise straightforward. So feel free to merge after resolving it either way.

recipes/grz-cli/meta.yaml
tedil: Yes, intentional.
It seems like the autobump machinery did not pick up on the .org URL, not sure why.

dlaehnemann: Maybe an old grayskull version or something?
https://github.com/conda/grayskull/issues/537

tedil: I mean, it was .org before, but for that autobump did not recognize it as a known source url (for pypi), so perhaps there is some issue in the autobump source detection. I'll change it back to .org here and see if I can find the underlying issue.

Issue #54006, opened 2025-02-20
Update Plink v1.9


aarthi-mohan:
Hi there,

I would like to install Plinkv1.9 that was released on 22 Oct 2024 as part of a Snakemake pipeline. Currently bioconda has until v1.90b6.21 (older but couldn't trace release year). Is it possible to update this to a latest version?

Thanks for your help!
Aarthi

martin-g:
https://anaconda.org/bioconda/plink
v1.90b6.12 has been uploaded "2 months and 9 days ago"

Let's see how https://github.com/bioconda/bioconda-recipes/pull/54017 will go!

aarthi-mohan:
Hi Martin, Thanks for looking into this. The version currently available via bioconda is `PLINK v1.90b6.21 64-bit (19 Oct 2020)` but the Plink [website](https://www.cog-genomics.org/plink/) has version `PLINK v1.9.0-b.7.7 64-bit (22 Oct 2024)` with several bug fixes. It would be helpful to have the most recent version, let me know if you need more info. Thanks!

martin-g:
Do you like to send a Pull Request ?
It would be similar to https://github.com/bioconda/bioconda-recipes/pull/54017/files. You need to update the versions and to reset the `build > number` field to `0`

aarthi-mohan:
Sure, I will make the changes suggested and create a PR. Thank you!

PR #54035, opened 2025-02-22
Update plink to v1.9.0-b.7.7 (2024 release)

aarthi-mohan:
Relates to update requested in #54006

### DIFF
--- a/recipes/plink/meta.yaml
+++ b/recipes/plink/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "plink" %}
-{% set version = "1.9.0-b.7.7" %}
+{% set version = "1.9.0b.7.7" %}
 
 package:
   name: {{ name }}
--- a/recipes/plink/meta.yaml
+++ b/recipes/plink/meta.yaml
@@ -1,20 +1,20 @@
 {% set name = "plink" %}
-{% set version = "1.90b6.21" %}
+{% set version = "v1.9.0-b.7.7" %}
 
 package:
   name: {{ name }}
   version: {{ version }}
 
 build:
-  number: 7
+  number: 0
   run_exports:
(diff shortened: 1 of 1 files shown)


recipes/plink/meta.yaml
```diff
@@ -1,20 +1,20 @@
 {% set name = "plink" %}
-{% set version = "1.90b6.21" %}
+{% set version = "1.9.0b.7.7" %}
```
martin-g, line 2: This didn't work well.
https://anaconda.org/bioconda/plink still says that 1.90 is the latest version.
We have the same problem for another recipe too - https://github.com/bioconda/bioconda-recipes/pull/53947

aarthi-mohan: Hi, I just noticed this. Plink1.9 is going to be replaced by Plink2 eventually, so all these are more of bug fixes than active feature development so the version numbers are 1.9-x.x.x, but a lot of us use 1.9 over 2 at this time. 
Any thoughts on what should be done? I am very new to this. Thanks!

martin-g: https://anaconda.org/bioconda/plink/files
All previous versions were `1.90`. The one uploaded by this version is `1.9`. The resolver resolves `1.90` as the latest...
Since no one complained about all previous `1.90`s that they are no really correct I'd suggest to keep using `1.90` (or 1.91, or whatever makes sense).
If you agree please create a new PR that changes the version and resets the `build > number` to `0`.

aarthi-mohan: Thanks, can I instead make it "1.90b7.7"? I hope this will be higher than the "1.90b6.21".

martin-g: Yes, I think so

PR #54085, opened 2025-02-24
Trinity: add linux-aarch64 support

dslarm:
- also removes -m64 flag from x86 configs, this has been default for many many years.


dslarm:
@BiocondaBot please add label

PR #54100, opened 2025-02-25
Add setuptools and core utils to dependencies of kgwasflow

ChristinaPapastolopoulou:
When using the bio container of kgwasflow we noticed that there was an incorrect version of ln. Therefore, I added coreutils to the dependencies of kgwasflow. While updating the recipe we found out that setuptools was missing, as well. 
@akcorut Do you see any big issues with this?

### DIFF
--- a/recipes/kgwasflow/meta.yaml
+++ b/recipes/kgwasflow/meta.yaml
@@ -11,7 +11,7 @@ source:
 
 build:
   noarch: python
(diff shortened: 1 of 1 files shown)


martin-g (approved):

ChristinaPapastolopoulou:
@BiocondaBot please add label

PR #54123, opened 2025-02-26
Update ivar version to 1.4.4 and added osx-arm64 to additional platforms

gkarthik:
* Bumped iVar to v1.4.4.
* Added `osx-arm64` to additional platforms


mencian (approved):

gkarthik:
Thank you @mencian !

PR #54184, opened 2025-02-27
Add rabbitsketch package to Bioconda

tripleldh:
This PR adds the rabbitsketch package to Bioconda. The package has been tested and built successfully.
@BiocondaBot please add label


tripleldh:

tripleldh:

tripleldh:
@BiocondaBot please update

mfansler (changes requested):
Pretty sure the CI does not process git submodules. Please resubmit without using submodules.

tripleldh:
@mfansler I accidentally requested a review before completing my tests. I'll update once everything is ready.

mfansler (changes requested):
The source code should not be added directly - most of the files here need to be removed.

Typical recipes only need a `meta.yaml` and sometimes a build script.

tripleldh:
@mfanslerThank you for your feedback! I have now removed all unnecessary files and kept only `meta.yaml`.

tripleldh:

tripleldh:
@BiocondaBot please review and merge

tripleldh:
@BiocondaBot please add label

recipes/rabbitsketch/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  run_exports:
+    - {{ pin_subpackage('rabbitsketch', max_pin="x.x") }}
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - cmake
+    - make
+    - llvm-openmp
```
martin-g, line 28: This and the following ones should be in 'host'. Not sure about pybind

recipes/rabbitsketch/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - python =3.7
+  commands:
+    - python -c "import sys; print(sys.path)"
+    - python -c "import rabbitsketch; print(rabbitsketch.__file__)"
+    - python -c "import rabbitsketch"
+    - python -c "from rabbitsketch import Kssd"
+
+about:
+  home: https://github.com/RabbitBio/RabbitSketch
+  license: https://github.com/RabbitBio/RabbitSketch/blob/lib/LICENSE.txt
```
martin-g, line 55: The license value should be a SPDX identifier

recipes/rabbitsketch/meta.yaml
tripleldh: The required dependencies have been moved to 'host' as suggested

recipes/rabbitsketch/meta.yaml
tripleldh: The license has been updated to GPL-3.0-or-later with the correct SPDX identifier

recipes/rabbitsketch/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - zlib
+    - pybind11 >=2.6, <2.7
+    - cython
+    - libxcrypt
+    - setuptools
+    - pip
+    - wheel
+  run:
+    - python >=3.7,<3.11
+    - libxcrypt
```
mfansler, line 40: `libxcrypt` has run exports defined, so having it in `host` automatically adds an appropriate requirement `run`:

```suggestion
```

recipes/rabbitsketch/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+build:
+  number: 0
+  skip: True  # [osx]
+  script:
+    - mkdir -p build
+    - cd build
+    - cmake -DCXXAPI=ON .. -DCMAKE_INSTALL_PREFIX=.
+    - make
+    - cd ..
+    - "{{ PYTHON }} -m pip install . --no-deps --ignore-installed -vv"
```
mfansler, line 18: Stylistic suggestions:

- if one has more than one command, then use a `build.sh` instead
- prefer `pushd build` ... `popd` over `cd build` ... `cd ..`

recipes/rabbitsketch/meta.yaml
```diff
@@ -0,0 +1,56 @@
+package:
+  name: rabbitsketch
+  version: "0.1.1"
+
+source:
+  url: https://github.com/RabbitBio/RabbitSketch/archive/refs/heads/bioconda.tar.gz
```
mfansler, line 6: This is substandard. It would be better to use a Release or Tag upstream that includes the `{{ version }}` in the string. That would enable the recipe to autodetect when new versions are released.

mfansler (commented):
Minor notes

recipes/rabbitsketch/meta.yaml
tripleldh: Thank you for your feedback! I have removed libxcrypt from the run section as suggested.

recipes/rabbitsketch/meta.yaml
tripleldh: I have tagged the bioconda branch with v0.1.1 and updated the meta.yaml to use the tag-based source URL.
And I have removed the script section from meta.yaml and kept build.sh as the build script to follow Bioconda’s best practices.

recipes/rabbitsketch/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - make
+  host:
+    - python >=3.7,<3.11
+    - cmake
+    - llvm-openmp
```
mfansler, line 24: I made a comment elsewhere. On Conda Forge, OpenMP should use `libgomp` on Linux platforms, and this should be in `build`:

```suggestion
    - libgomp  # [linux]
  host:
    - python >=3.7,<3.11
    - cmake
```

mfansler (changes requested):
If this doesn't build, then revert - but I would like to see it tried this way.

PR #54283, opened 2025-03-04
Update: Preserve existing PERL5LIB path on run

Adamtaranto:
Wrapper script for `pfam_scan.pl` currently overwrites `PERL5LIB` path. On OSX this causes error locating the IPC run.pm dependency.

This PR appends the existing `PERL5LIB` path at time of execution to fix this issue. No effect on other platforms.


Adamtaranto:
@BiocondaBot please add label

martin-g (approved):

Issue #54304, opened 2025-03-04
ensembl-vep>=112 is missing dependency perl-list-moreutils


feiloo:
running `vep` after installing a ensembl-vep version after 111 produces the following error:

```
Can't locate List/MoreUtils.pm in @INC (you may need to install the List::MoreUtils module) (@INC contains: /home/user/miniconda3/envs/ensembl_test/share/ensembl-vep-113.3-1/modules /home/user/miniconda3/envs/ensembl_test/share/ensembl-vep-113.3-1 /home/user/miniconda3/envs/ensembl_test/lib/perl5/5.32/site_perl /home/user/miniconda3/envs/ensembl_test/lib/perl5/site_perl /home/user/miniconda3/envs/ensembl_test/lib/perl5/5.32/vendor_perl /home/user/miniconda3/envs/ensembl_test/lib/perl5/vendor_perl /home/user/miniconda3/envs/ensembl_test/lib/perl5/5.32/core_perl /home/user/miniconda3/envs/ensembl_test/lib/perl5/core_perl .) at /home/user/miniconda3/envs/ensembl_test/share/ensembl-vep-113.3-1/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm line 102.
BEGIN failed--compilation aborted at /home/user/miniconda3/envs/ensembl_test/share/ensembl-vep-113.3-1/Bio/EnsEMBL/Variation/DBSQL/VariationFeatureAdaptor.pm line 102.
Compilation failed in require at /home/user/miniconda3/envs/ensembl_test/share/ensembl-vep-113.3-1/modules/Bio/EnsEMBL/VEP/BaseVEP.pm line 59.
BEGIN failed--compilation aborted at /home/user/miniconda3/envs/ensembl_test/share/ensembl-vep-113.3-1/modules/Bio/EnsEMBL/VEP/BaseVEP.pm line 59.
Compilation failed in require at /home/user/miniconda3/envs/ensembl_test/lib/perl5/vendor_perl/base.pm line 100.
        ...propagated at /home/user/miniconda3/envs/ensembl_test/lib/perl5/vendor_perl/base.pm line 109.
        BEGIN failed--compilation aborted at /home/user/miniconda3/envs/ensembl_test/share/ensembl-vep-113.3-1/modules/Bio/EnsEMBL/VEP/BaseRunner.pm line 56.
[...]

PR #54330, opened 2025-03-05
fix ensembl-vep missing dependency "perl-list-moreutils"

feiloo:
this adds the missing dependency perl-list-moreutils 

it also removes the insecure curl requests.

note: with my strict conda channel priority perl>=5.27.0 had to be specified for this to build

context:

since bioconda::ensembl-vep>=112.0 the vep command errors out because of missing List/MoreUtils.pm 

see https://github.com/bioconda/bioconda-recipes/issues/54304#issue-2895018904


feiloo:
I think the previous test flaked out because of some network interruption.

martin-g (approved):

PR #54423, opened 2025-03-10
Update r-psminer and rename to r-pathosurveilr

zachary-foster:
This needed to be renamed because there was already a package on CRAN called `psminer` that we did not know about.


zachary-foster:
@BiocondaBot please add label

martin-g (approved):

PR #54446, opened 2025-03-11
Update dysgu recipie

kcleal:
This lists version numbers to try and help lower requirements and prevent conflicts


kcleal:
@BiocondaBot please add label

PR #54479, opened 2025-03-12
linux-aarch64 build for gtdbtk-2.0.0

ZerryNi:
The adaptation of gtdbtk 2.0.0 to linux-aarch64 has passed the test.
gtdbtk-2 0 0


mencian:
`gtdbtk` is already a `noarch` package since `v1.0.2`, adding `linux-aarch64` is redundant.

ZerryNi:
> `gtdbtk` is already a `noarch` package since `v1.0.2`, adding `linux-aarch64` is redundant.

image
Hi，but it can't be installed!

PR #54480, opened 2025-03-12
linux-aarch64 build for perl-db-file

ZerryNi:
Test passed!
perl-db-file

### DIFF
--- a/recipes/perl-db-file/build.sh
+++ b/recipes/perl-db-file/build.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+echo "INCLUDE = $PREFIX/include" > config.in
+echo "LIB = $PREFIX/lib" >> config.in
+echo "PREFIX = size_t" >> config.in
+echo "HASH = u_int32_t" >> config.in
+
+# If it has Build.PL use that, otherwise use Makefile.PL
+if [ -f Build.PL ]; then
+    perl Build.PL
+    ./Build
+    ./Build test
+    # Make sure this goes in site
+    ./Build install --installdirs site
+elif [ -f Makefile.PL ]; then
+    # Make sure this goes in site
+    perl Makefile.PL INSTALLDIRS=site
+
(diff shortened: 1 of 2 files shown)


dslarm:
Thanks @ZerryNi !

Did you know that perl-db-file is replaced by the conda forge package perl-db_file?

I am OK / happy to see this patch - it's good to have perl-db-file exist for linux-aarch to create a build, because some old recipes can't build today and need perl-db-file..   but if you see any recipes that need perl-db-file, please consider changing the requirement to perl-db_file so that it is kept up to date, as this perl-db-file should be removed in the future for every architecture.

bgruening:
But if there is a conda-forge package, we should not put a bioconda one in.

This should be contributed to the conda-forge package IMHO.
Is that ok for you @ZerryNi?

martin-g:
IMO https://anaconda.org/conda-forge/perl-db_file should be used.
If an old recipe uses perl-db-file then it should be updated

ZerryNi:
> But if there is a conda-forge package, we should not put a bioconda one in.
> 
> This should be contributed to the conda-forge package IMHO. Is that ok for you @ZerryNi?

ok

ZerryNi:
> 谢谢@ZerryNi！
> 
> 您是否知道 perl-db-file 已被 conda forge 包 perl-db_file 取代？
> 
> 我很高兴看到这个补丁 - 拥有 perl-db-file 以便 linux-aarch 创建构建是件好事，因为一些旧的配方今天无法构建并且需要 perl-db-file。但是如果您看到任何需要 perl-db-file 的配方，请考虑将要求更改为 perl-db_file，以便它保持最新，因为这个 perl-db-file 将来应该在每个架构中被删除。

ZerryNi:
> IMO https://anaconda.org/conda-forge/perl-db_file should be used. If an old recipe uses perl-db-file then it should be updated

ok,I get it

PR #54482, opened 2025-03-12
linux-aarch64 build for pplacer

ZerryNi:
image

### DIFF
--- a/recipes/aviary/meta.yaml
+++ b/recipes/aviary/meta.yaml
@@ -1,13 +1,13 @@
-{% set version = "0.11.0" %}
+{% set version = "0.11.1" %}
 {% set name = "aviary" %}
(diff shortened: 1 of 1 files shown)


dslarm:
@ZerryNi - this is going to be a harder one to fix.

pplacer is currently a binary tarball (from 2016) which doesn't support aarch.
pplacer needs ocaml to build.

ocaml 4.12 supports aarch64, but ocaml's conda forge recipe is at 4.11, and it doesn't work at 4.12 or above.

I have not been able to understand why it doesn't work: https://github.com/conda-forge/ocaml-feedstock/pull/45
There is a segfault once the binaries have been moved/patched during the package creation of ocaml.  If you could work on that, then this pplacer recipe should be converted to build from source - from the original owner's repo.

PR #54486, opened 2025-03-12
update Tidyp for linux-aarch64,add usage of '-fPIC' flag

ZerryNi:
Test passed!
tidyp

### DIFF
--- a/recipes/tidyp/build.sh
+++ b/recipes/tidyp/build.sh
@@ -4,11 +4,16 @@ set -xe
 
 mkdir -p ${HOME}/bin
 export PATH="${HOME}/bin:$PATH"
-ln -s ${CC} ${HOME}/bin/gcc
+
+ln -fs ${CC} ${HOME}/bin/gcc
 
 pushd build/gmake
+sed -i.bak '85s/^CFLAGS=/CFLAGS=-fPIC/' Makefile
 make
 popd
 
 mkdir -p ${PREFIX}/bin
+
 cp ./bin/tidyp ${PREFIX}/bin
+cp -r ./include ${PREFIX}/
+cp -r ./lib ${PREFIX}/
--- a/recipes/tidyp/meta.yaml
+++ b/recipes/tidyp/meta.yaml
@@ -7,7 +7,7 @@ source:
   sha256: 5f0965041119bed0ed70a7f0be39e8f0162141eea14d77f472b49f8b3303ab0b
 
 build:
-  number: 7
+  number: 8
   run_exports:
     - {{ pin_subpackage('tidyp', max_pin="x") }}
 


ZerryNi:
@BiocondaBot please add label

recipes/tidyp/build.sh
```diff
@@ -4,11 +4,26 @@ set -xe
...
+fi
 
 pushd build/gmake
+if [[ "$OSTYPE" == "darwin"* ]]; then
+  #macOS
+  sed -i '' '85s/^CFLAGS=/CFLAGS=-fPIC/' Makefile
+else
+  #Linux	
+  sed -i '85s/^CFLAGS=/CFLAGS=-fPIC/' Makefile
+fi
```
martin-g, line 19: No need of if/else. Just use 'sed -i.bak ...' for both OSes.

recipes/tidyp/build.sh
```diff
@@ -4,11 +4,26 @@ set -xe
 
 mkdir -p ${HOME}/bin
 export PATH="${HOME}/bin:$PATH"
-ln -s ${CC} ${HOME}/bin/gcc
+
+if [ ! -f "${HOME}/bin/gcc" ]; then
```
martin-g, line 8: I guess you added this check because it falls on the second run when executed locally?
You could use 'ln -f...' to force the link creation. 
Usually recipes use $BUILD_PREFIX/bin instead of $HOME/bin for this workaround.

ZerryNi: Hello! I just want to create a soft link,but not a hard link.So I use 'if else' to judge.

martin-g: I understand!
IMO the best would be to use `$BUILD_PREFIX` instead of `$HOME`. 
Create the symlink before the build command and undo it after it (`unlink ...`)

recipes/tidyp/build.sh
ZerryNi: I've revised it,please refer to.

recipes/tidyp/build.sh
```diff
@@ -4,11 +4,18 @@ set -xe
...
+ln -fs ${CC} ${HOME}/bin/gcc
 
 pushd build/gmake
+sed -i.bak '85s/^CFLAGS=/CFLAGS=-fPIC/' Makefile
 make
 popd
 
 mkdir -p ${PREFIX}/bin
+mkdir -p ${PREFIX}/include
+mkdir -p ${PREFIX}/lib
```
martin-g, line 17: No need to mkdir those folders. They are created with 'cp -r ...' below

ZerryNi: I‘ve revised it,please refer to.

mencian:
Closing in favor of #55675.

PR #54338, opened 2025-03-05
add postmaster

taliaferrojm:
add postmaster recipe

PR #54546, opened 2025-03-13
update pigpen_0.0.6 to include r-lme4 and postmaster

taliaferrojm:
Update recipe for pigpen_0.0.6 to include requirements for r-lme4 and postmaster. Additionally, add tests for bacon.


recipes/pigpen/meta.yaml
```diff
@@ -12,7 +12,8 @@ source:
 
 
 build:
-  number: 0
+  number: 1
+  skip: True # postmaster not available on osx
```
martin-g, line 16:
```suggestion
  skip: True # [osx]
```

recipes/pigpen/meta.yaml
```diff
@@ -60,6 +65,4 @@ extra:
     - taliaferrojm
   identifiers:
     - https://doi.org/10.1101/2024.11.12.623278
-  additional-platforms:
-    - linux-aarch64
```
martin-g, line 64: Why this is removed ?
If postmaster doesn't support it yet then can we add support for it instead of removing it here ?

taliaferrojm: That's a good idea, but it's beyond my capabilities. Postmaster is not my software, and I know next to nothing about Rust. According to [this](https://github.com/bioconda/bioconda-recipes/pull/54338/commits/bd5937e9780f39dea981a6e639c661334bb40a63), postmaster uses an old version of rust-htslib that fails on aarch64.

taliaferrojm: Hello @martin-g 

Just a kind request to please approve this submission when you have time. Thanks.

mencian (approved):

PR #54558, opened 2025-03-14
Update polap recipe to v0.4.3.7

goshng:
Please, update polap recipe to v0.4's first version.


recipes/polap/meta.yaml
```diff
@@ -1,32 +1,41 @@
 {% set name = "polap" %}
 {% set version = "0.4.3.7" %}
-{% set sha256 = "9e225905991025d58ea905fea7ef4645236de3e290225f6db665d5774ebec516" %}
+{% set sha256 = "cec0439a4b4bb79dfed5d0bb5d6c9904921812448a202eed4c317ab2518f0f2a" %}
```
bgruening, line 3: Is there a new release? Have you added all those changes to an already existing release?

goshng: I do not understand what you meant to ask. The sha256 is the sha256sum of the github tar.gz, right?

goshng: One thing that I noted was that build 0 of the version 0.4.3.7 has gone through bioconda repo without a review or approval of a member like you.

Please, see noarch/polap-0.4.3.7-hdfd78af_0.tar.bz2 at the top of the file list available at https://anaconda.org/bioconda/polap/files

I do not know what happened. I am also confused by the 2nd at the top, which is  noarch/polap-0.4.1.1-hdfd78af_0.tar.bz2 because I did not uploaded such recipe.

I only did try to follow the instruction at https://bioconda.github.io/contributor/workflow.html#

but I did not PR. Do you see what happened in the two bioconda packages? It is very weird.

PR #54564, opened 2025-03-14
update quast

ZerryNi:
quast

### DIFF
--- a/recipes/quast/meta.yaml
+++ b/recipes/quast/meta.yaml
@@ -6,7 +6,7 @@ package:
   version: {{ version }}
 
 build:
-  number: 2
+  number: 3
   run_exports:
     - {{ pin_subpackage("quast", max_pin="x") }}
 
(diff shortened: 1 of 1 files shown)


ZerryNi:
@BiocondaBot please add label

recipes/quast/meta.yaml
```diff
@@ -26,32 +26,32 @@ requirements:
...
   host:
-    - perl
-    - python >=3.12
+    - perl 5.32.1.*
+    - python 3.12.*
     - pip
-    - setuptools
-    - cython
-    - matplotlib-base
+    - setuptools 68.2.2.*
```
martin-g, line 32: Why does it need to be specifically 68.2.2.x? 
Would it break with 68.2.3.x ?

The same question is valid for all other changes in this and other PRs of yours.

recipes/quast/meta.yaml
```diff
@@ -26,32 +26,32 @@ requirements:
...
+    - matplotlib-base 3.8.4.*
     - joblib
     - simplejson
-    - blast
+    - blast 2.16.0.*
     - zlib
     - glimmerHMM
     - circos
-    - minimap2
+    - minimap2 >=2.10
```
martin-g, line 41: 👍

recipes/quast/meta.yaml
ZerryNi: Do you mean these depennices can't be versioned?

martin-g: They can!
But what is the need to version them ?

Adding such specific version makes the upgrades harder! Someone has to do it manually. 
Usually most of the packages provide backward compatible new versions, so there is no need to pin them like that.

ZerryNi: Hello。The build of the software metawrap-mg on the linux-aarch platform depends on these software,and if no version is specified,these dependencies conflict with each other when compiling metawrap-mg.

PR #54594, opened 2025-03-14
updated ectyper_v2.0.0 build #3 recipe to fix minor bugs in serotyping

kbessonov1984:
Users identified small issues including
- higher number of mixed O-antigen calls
- [issue 101](https://github.com/phac-nml/ecoli_serotyping/issues/101) with the inconsistent number of output columns depending on `--pathotype` parameter specified or not

### DIFF
--- a/recipes/ectyper/meta.yaml
+++ b/recipes/ectyper/meta.yaml
@@ -7,10 +7,10 @@ package:
 
 source:
     url: https://github.com/phac-nml/ecoli_serotyping/archive/{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


kbessonov1984:
@BiocondaBot please add label

nh13 (changes requested):
@kbessonov1984 I am curious why not release a fix (`2.0.1`) according to [semvar](https://semver.org/) versus overwriting the tag on github?  I would not expect that the same version with a new build # (on bioconda) to have fixes other than for build/compiling issues.

kbessonov1984:
Since this is a very cosmetic issue on the output I felt that it was not worth while to release another version `2.0.1` and confuse users. If there will be more issues, then of course we will bump version to >2.0.0

PR #54602, opened 2025-03-15
Add dflye

goshng:
Please, add dflye.

### DIFF
--- a/recipes/dflye/build.sh
+++ b/recipes/dflye/build.sh
@@ -1,0 +1,28 @@
+#!/bin/bash -euo
+
+# Original: flye's recipes
+
+#zlib headers for minimap
+sed -i.bak 's/CFLAGS=/CFLAGS+=/' lib/minimap2/Makefile
+sed -i.bak 's/INCLUDES=/INCLUDES+=/' lib/minimap2/Makefile
+export CFLAGS="${CFLAGS} -O3 -L$PREFIX/lib"
+export INCLUDES="-I$PREFIX/include"
+
+rm -rf lib/minimap2/*.bak
+
+#zlib headers for cflye binaries
+export CXXFLAGS="$CXXFLAGS -O3 -I$PREFIX/include"
+export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
+
+#install_name_tool error fix
+if [[ "$(uname)" == Darwin ]]; then
+	export LDFLAGS="$LDFLAGS -headerpad_max_install_names"
+fi
+
+#dynamic flag is needed for backtrace printing,
+#but it seems it fails OSX build
+sed -i.bak 's/-rdynamic//' src/Makefile
+
+rm -rf src/*.bak
+
+${PYTHON} -m pip install --no-deps --no-build-isolation --no-cache-dir . -vvv
--- a/recipes/dflye/meta.yaml
+++ b/recipes/dflye/meta.yaml
@@ -1,0 +1,44 @@
+{% set name = "dflye" %}
+{% set version = "2.9.5.0" %}
+{% set sha256 = "b789d9083d3388c186c4fbc8de50df0b9791bb439d9083ca482ec0a334737e0a" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/goshng/Flye/archive/refs/tags/{{ version }}-d.tar.gz
+  sha256: {{ sha256 }}
+  patches:
+    - 0001-Makefile-aarch64.patch
+
+build:
+  number: 0
+  entry_points:
+    - dflye = dflye.main:main
+  run_exports:
+    - {{ pin_subpackage("dflye", max_pin="x") }}
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - make
+  host:
+    - python
+    - pip
+    - zlib
+  run:
+    - python
+    - setuptools  # [osx]
+
+test:
+  commands:
+    - dflye -h | grep "Assembly"
+
+about:
+  home: "https://github.com/goshng/Flye"
+  license: "BSD-3-Clause"
+  license_family: BSD
+  license_file: LICENSE
+  summary: "dflye: Directional reads in the Flye."
+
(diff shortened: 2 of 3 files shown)


goshng:
@BiocondaBot please add label

recipes/dflye/build.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash -euo
+
+# Original: flye's recipes
+
+#zlib headers for minimap
```
martin-g, line 5: Could the recipe use minimap2 as a 'host' dependency instead of the vendored one ?

recipes/dflye/build.sh
```diff
@@ -0,0 +1,28 @@
...
+export LDFLAGS="$LDFLAGS -L$PREFIX/lib"
+
+#install_name_tool error fix
+if [[ "$(uname)" == Darwin ]]; then
+	export LDFLAGS="$LDFLAGS -headerpad_max_install_names"
+fi
+
+#dynamic flag is needed for backtrace printing,
+#but it seems it fails OSX build
+sed -i.bak 's/-rdynamic//' src/Makefile
```
martin-g, line 24: Maybe move it in the 'if Darwin' check above ? This way the Linux builds will still use it

recipes/dflye/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  commands:
+    - dflye -h | grep "Assembly"
+
+about:
+  home: "https://github.com/goshng/Flye"
+  license: "BSD-3-Clause"
+  license_family: BSD
+  license_file: LICENSE
+  summary: "dflye: Directional reads in the Flye."
+
```
martin-g, line 44: Maybe add a build for linux-aarch64 since you patch the Makefile?!
See 'additional-platforms' in other recipes

recipes/dflye/build.sh
goshng: I think that Flye uses a separate version of minimap2 in its implementation. So, this would need extensive testing, which is beyond the scope of dflye's implementation. So, could I skip this, please?

recipes/dflye/build.sh
goshng: I am not sure of this because I copy Flye's current version of bioconda recipe. There must be some reasons of leaving this way although I agree with you about this one. Let me follow your suggestion at the moment.

recipes/dflye/meta.yaml
goshng: I will try it out, although I am sure of it. Thank you!

goshng: I tried to add the build to no avail. As I mentioned earlier, I am using the Flye's recipes in the bioconda. The original author would have some reasons of having that version. Could I skip this additional platform stuff?

recipes/dflye/build.sh
goshng: I also tried this one, but it failed to the conda build in the github. So, I guess that the original Flye's recipe should be kept in the original form. Could I go without the 'if Darwin' check as well?

recipes/dflye/build.sh
goshng: All the of your suggestions are good and they should be considered, which I did. But, there must be some reasons why the original Flye's bioconda recipes is at the current form. Because I use the current copy of Flye's biocoda recipes, I think that reflecting your suggestion by myself is beyond the scope of my changing some of Flye's feature at the moment. So, could I go ahead without reflecting your suggestions here? I really appreciate your comments, but I tried to reflect those to no avail. Thank you!

goshng: The cflye bioconda recipe is in similar situation.

goshng: I appreciate your help.

goshng:
@BiocondaBot please add label

recipes/dflye/build.sh
martin-g: Alright!

recipes/dflye/build.sh
martin-g: What is the relation between flye, cflye and dflye ?
The descriptions of the new recipes (cflye and dflye) are not very clear.

recipes/dflye/meta.yaml
martin-g: Flye supports both linux-aarch64 and osx-arm64 - https://github.com/bioconda/bioconda-recipes/blob/eae742da15961d78b6278969bd992678fc02624e/recipes/flye/meta.yaml#L49-L51

Your recipes seem to copy code from the flye recipe but you have no idea how/why it works :-)
I'll leave the PRs open for review by other Bioconda maintainers.
I am starting to think this might be a `name squatting` attempt - you create recipes with names which are very similar to another recipe, their code is almost the same but there is something different that you cannot explain...

goshng: Thank you for your help. You are almost right about it although I could explain what I added to the original codebase. But, I am not sure whether it could be general enough to be integrated into the Flye codebase, yet. Until then, I guess that a bioconda package of dflye would be great for testing its new feature. What I do not agree about your comment about "there is something different that you cannot explain". As I mentioned, I'd be happy to explain it if I may. Anyway, I hope that this package could be available in the bioconda so that potential users might have easier time to use its feature. I will clearly mention what it does and what it is not. I do not think that users would use it instead of the original assembler, Flye. Could anybody review this, again? I understand your concerns about this package and appreciate your time and efforts. As always, thank you for your help. A bioconda package is crucial in my work.

goshng: > Flye supports both linux-aarch64 and osx-arm64 -
> 
> https://github.com/bioconda/bioconda-recipes/blob/eae742da15961d78b6278969bd992678fc02624e/recipes/flye/meta.yaml#L49-L51
> 
> Your recipes seem to copy code from the flye recipe but you have no idea how/why it works :-) I'll leave the PRs open for review by other Bioconda maintainers. I am starting to think this might be a `name squatting` attempt - you create recipes with names which are very similar to another recipe, their code is almost the same but there is something different that you cannot explain...

Thank you! Your suggestion of additional platforms works.

PR #54620, opened 2025-03-16
Update PlasAnn to 1.0.8

Habib-human:
Update PlasAnn to 1.0.5


recipes/plasann/meta.yaml
```diff
@@ -34,7 +34,7 @@ requirements:
 
 test:
   commands:
-    - PlasAnn --help
+    - python -c "import plasann"
```
bgruening, line 37: if you want to test imports you can do that outside of the commands section.

PR #54651, opened 2025-03-18
linux-aarch64 build for metawrap-mg

ZerryNi:
metawrap-mg


mencian:
This package is already `noarch`, adding `linux-aarch64` is redundant.

ZerryNi:
> This package is already `noarch`, adding `linux-aarch64` is redundant.

image
It's also impossible to install the software

PR #54654, opened 2025-03-18
add pisad

ZhantianXu:


ZhantianXu:
we developed PISAD, a Phsaed Intraspecies Sample Anomalies Detection tool, suitable for multiple species without reference information.We now want to add it to bioconda

recipes/pisad/meta.yaml
```diff
@@ -0,0 +1,61 @@
+package:
+  name: pisad
+  version: 1.1.0
+
+source:
+  url: https://github.com/ZhantianXu/PISAD/releases/download/v1.1.0/PISAD-1.1.0.tar.gz
+  sha256: 3864210c041868e399469647226f641e094e8056ec921ed1e90024832409e4e4
+
+build:
+  skip: True # [win]
```
martin-g, line 10: This is not really needed. Bioconda does not support Windows anyway
```suggestion
```

recipes/pisad/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - scikit-learn
+    - pandas
+    - samtools
+    - hdf5
+    - boost-cpp
+    - gperftools
+    - dsk
+    - matplotlib-base
+    - perl
+    - make
```
martin-g, line 44: Is 'make' really needed at runtime?

ZhantianXu: 'make' is not required at runtime, it has been deleted, thank you for pointing out.

recipes/pisad/meta.yaml
ZhantianXu: it has been deleted, thank you for pointing out.

martin-g (approved):

PR #54657, opened 2025-03-18
Wakhan recipe fix -- added missing requirement

ljwharbers:
Added missing requirement for wakhan


ljwharbers:
@BiocondaBot please add label

martin-g (approved):

PR #54683, opened 2025-03-18
Update cctyper - new build to automatically install database

simroux:
This is an update of the cctyper recipe. The tool is still exactly the same (so no version number change) but the build number was increased as the recipe was changed.
The recipe was changed to now automatically install the CCTyper database upon build, rather than having the user need to install the database separately.


simroux:
@BiocondaBot please add label

mencian (approved):

PR #54731, opened 2025-03-19
Add recipe AdDeam

LouisPwr:
Conda recipe for aDNA tool AdDeam.

### DIFF
--- a/recipes/addeam/build.sh
+++ b/recipes/addeam/build.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+set -e
+
+export CPPFLAGS="-I${PREFIX}/include -I${PREFIX}/include/htslib"
+export LDFLAGS="-L${PREFIX}/lib -lhts -lgsl -lcurl -lgslcblas -llzma -lbz2 -lz  -lpthread"
+
+pushd submodules/lib/libgab
+git submodule update --init --recursive
+make clean
+make libgab.a
+make -C gzstream
+popd
+
+pushd submodules/src
+make clean
+make CXX="${CXX}"   
+install -v -m 0755 bam2prof "${PREFIX}/bin/"
+install -v -m 0755 diffprof "${PREFIX}/bin/"
+popd
+
+install -v -m 0755 addeam-bam2prof.py addeam-cluster.py "${PREFIX}/bin/"
--- a/recipes/addeam/meta.yaml
+++ b/recipes/addeam/meta.yaml
@@ -1,0 +1,69 @@
+{% set name = "addeam" %}
+{% set version = "1.0.0" %}
+{% set sha256 = "db89dff19db7e3fbc64d0be11774ed53e604dd5738a31d0f3ae677fb7385a217" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/LouisPwr/AdDeam/releases/download/{{ version }}/AdDeam-v{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+
+build:
+  number: 0
+  skip: True  # [osx]
+  run_exports:
+    - {{ pin_subpackage("addeam", max_pin="x") }}  
+
+requirements:
+  build:
+    - make
+    - {{ compiler('cxx') }}
+    - gxx >=12.0
+  host:
+    - cython
+    - zlib
+    - xz
+    - curl
+    - bzip2
+    - gsl
+    - htslib
+    - samtools
+  run:
+    - python
+    - numpy >=2
+    - pandas
+    - matplotlib-base
+    - scikit-learn
+    - scipy
+    - opencv
+    - pypdf2
+    - bottleneck
+    - htslib
+    - samtools
+    - zlib
+    - xz
+    - curl
+    - bzip2
+
+
+test:
+  commands:
+    - addeam-bam2prof.py --help | grep "Python wrapper for bam2prof"
+    - addeam-cluster.py --help | grep "Cluster and plot damage profiles."
+
+
+about:
+  home: "https://github.com/LouisPwr/AdDeam"
+  license: GPL-3.0-or-later
+  license_file: LICENSE
+  summary: "A Fast and Scalable Tool for Estimating and Clustering Reference-Level Damage Profiles."
+
+extra:
+  identifiers:
+    - biotools:addeam
+  recipe-maintainers:
+    - gabriel-renaud
+    - louis-kraft
--- a/recipes/addeam/meta.yaml
+++ b/recipes/addeam/meta.yaml
@@ -21,7 +21,6 @@ requirements:
   build:
     - make
     - {{ compiler('cxx') }}
-    - gxx >=12.0
   host:
     - cython
     - zlib


LouisPwr:
please review & merge

recipes/addeam/build.sh
```diff
@@ -0,0 +1,31 @@
...
+
+export LDFLAGS="-L${PREFIX}/lib"
+
+export CFLAGS="-I${SRC_DIR}/submodules/lib/htslib -I${SRC_DIR}/submodules/lib/samtools -I${SRC_DIR}/submodules/lib/libgab -I${PREFIX}/include"
+export CPPFLAGS="-I${SRC_DIR}/submodules/lib/htslib -I${SRC_DIR}/submodules/lib/samtools -I${SRC_DIR}/submodules/lib/libgab -I${PREFIX}/include"
+export CXXFLAGS="-I${SRC_DIR}/submodules/lib/htslib -I${SRC_DIR}/submodules/lib/samtools -I${SRC_DIR}/submodules/lib/libgab -I${PREFIX}/include"
+
+mkdir -p "${PREFIX}/bin"
+
+cd submodules/src/
```
martin-g, line 14:
```suggestion
pushd submodules/src/
```

recipes/addeam/build.sh
```diff
@@ -0,0 +1,31 @@
...
+
+ln -s ${PREFIX}/lib/libncurses.so ${PREFIX}/lib/libcurses.so
```
martin-g, line 16: This looks like a workaround. 
Can't we require the correct host requirement instead ?

recipes/addeam/build.sh
```diff
@@ -0,0 +1,31 @@
...
+
+make clean
+make
```
martin-g, line 19:
```suggestion
make -j"${CPU_COUNT}"
```

recipes/addeam/build.sh
```diff
@@ -0,0 +1,31 @@
...
+mv bam2prof "${PREFIX}/bin/"
```
martin-g, line 20:
```suggestion
install -v -m 0755 bam2prof "${PREFIX}/bin/"
```

recipes/addeam/build.sh
```diff
@@ -0,0 +1,31 @@
...
+chmod +x "${PREFIX}/bin/bam2prof"
```
martin-g, line 21:
```suggestion
```

recipes/addeam/build.sh
```diff
@@ -0,0 +1,31 @@
...
+cd ../..
```
martin-g, line 22:
```suggestion
popd 
```

recipes/addeam/build.sh
```diff
@@ -0,0 +1,31 @@
...
+
+# Ensure Python scripts are installed correctly
+cp addeam-bam2prof.py addeam-cluster.py "${PREFIX}/bin/"
```
martin-g, line 25:
```suggestion
install -v -m 0755 addeam-bam2prof.py addeam-cluster.py "${PREFIX}/bin/"
```

recipes/addeam/build.sh
```diff
@@ -0,0 +1,31 @@
...
+chmod +x "${PREFIX}/bin/addeam-bam2prof.py" "${PREFIX}/bin/addeam-cluster.py"
+
+# Add ${PREFIX}/bin to PATH so tests can find the executables
+export PATH="${PREFIX}/bin:$PATH"
```
martin-g, line 29: This shouldn't be needed. Please remove it!

recipes/addeam/meta.yaml
```diff
@@ -0,0 +1,70 @@
+{% set name = "addeam" %}
+{% set version = "1.0.0" %}
+{% set sha256 = "a93f302e3be9d4b35a80ddf73d7ed7143cc3f00638e93bf782154c2ed454899d" %}
+
+package:
+  name: addeam
```
martin-g, line 6:
```suggestion
  name: {{ name }}
```

recipes/addeam/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+  version: {{ version|replace("-", ".")  }}
```
martin-g, line 7:
```suggestion
  version: {{ version }}
```

recipes/addeam/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+  skip: True  # [osx]
+  run_exports:
+    - {{ pin_subpackage("addeam", max_pin="x") }}  
+
+requirements:
+  build:
+    - make
+    - {{ compiler('cxx') }}
+    - {{ compiler('c') }}
+    - gxx >=12.0
```
martin-g, line 24: Isn't this already provided by compiler(cxx)?!

recipes/addeam/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+test:
+  commands:
+    - addeam-bam2prof.py --help | grep "Python wrapper for bam2prof"
+    - addeam-cluster.py --help | grep "Cluster and plot damage profiles."
+
+
+about:
+  home: "https://github.com/LouisPwr/AdDeam"
+  license: GPLv3
```
martin-g, line 61:
```suggestion
  license: GPL-3.0-or-later
```

martin-g (dismissed):

recipes/addeam/build.sh
LouisPwr: Hi, thanks for your time!

The issue is, that we use code from samtools and htslib. Therefore AdDeam is cloning their repositories and compiling them. samtools uses the "curses" library during compiling. Nowadays there is only "ncurses" available, which is the newer version of "curses" BUT it is backwards compatible. So on many servers "curses" is symlinked to "ncurses". And since conda only provides "ncurses" I needed to use this symlink strategy.

recipes/addeam/build.sh
LouisPwr: There are some submodule dependencies during make and it is safer to use a single thread.

LouisPwr (commented):
applied most

recipes/addeam/build.sh
LouisPwr: Thanks!
Done that now.

recipes/addeam/meta.yaml
LouisPwr: Indeed. Changed.

LouisPwr:
@BiocondaBot please add label

LouisPwr:
@BiocondaBot help needed

recipes/addeam/build.sh
jmarshall: 1. Conda is unusual in not providing _libcurses.so_ etc as symlinks.
2. Samtools only uses `-lcurses` by default. You should be running its _configure_ script, which will detect ncurses and use it.
3. But all that is immaterial in a conda context. You should not be bundling htslib and samtools, and your package will not be accepted like this. Instead your package should depend on bioconda's htslib and samtools packages instead of building this external software itself.

LouisPwr: Hi!
Thanks for your reply. That makes sense. 
I updated it now - using condas htslib and samtools - and it is indeed much cleaner.

LouisPwr:
@BiocondaBot please add label

PR #54971, opened 2025-03-28
ci: update check names in mergify

aliciaaevans:
Mergify hasn't been merging successful autobumps because the "OSX Tests" check name changed to "OSX-64 Tests". Also, we should not merge if the ARM builds failed.


bgruening (approved):

bgruening:
Merci!

PR #54976, opened 2025-03-28
Update el_gato version 1.21.2

jennahamlin:
Update el_gato version 1.21.2

### DIFF
--- a/recipes/el_gato/meta.yaml
+++ b/recipes/el_gato/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "el_gato" %}
-{% set version = "1.21.1" %}
+{% set version = "1.21.2" %}
 {% set hash = "4b16fcf89cd3af033c1eec27b32893e2421f112168e983610ce9db751611692e" %}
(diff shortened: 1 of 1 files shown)


recipes/el_gato/meta.yaml
```diff
@@ -27,12 +27,13 @@ requirements:
...
     - python >=3.8,<3.12
     - pip
+
   run:
     - python >=3.8,<3.12
-    - minimap2 >=2.24
-    - samtools >=1.15.1
-    - blast >=2.13
-    - ispcr >=33.0
+    - minimap2 ==2.24
```
bgruening, line 33: please do not use strict version pins, only use pinning if the version are not compatible

jennahamlin: Got it, thank you

jennahamlin:
@BiocondaBot please add label

recipes/el_gato/meta.yaml
bgruening: do you really need this specific minimap version?

PR #55039, opened 2025-04-01
Update PretextGraph to 0.0.9

guanshaoheng:
Changes:
- removed the normalisation for `repeat_density` extension. The repeat density was normalised by the `bin_size`. But there are some bins with different sizes, which will make the repeat_density signal incorrect.
- removed the dependency of `Wrapper` by adding the `SIMDExtensionWrapper.cpp` inside of the project.
- change the function of `stbsp_snprintf` to `snprintf` to avoid **segfault**.
- update version number to `0.0.9`


guanshaoheng:
Could you please help review and merge? Thank you so much. @martin-g

mencian (approved):

PR #48787, opened 2024-06-27
Update plink2 to 2.00a5.12 final

aryarm:
Update [`plink2`](https://www.cog-genomics.org/plink/2.0): **2.00a5.10** → **2.00a5.12** (released on June 25, 2024)

I simply adapted https://github.com/bioconda/bioconda-recipes/pull/46505 to write this PR

PR #55220, opened 2025-04-09
Update plink2 to 2.0.0-a.6.9

tavareshugo:
Update `plink2` to version 2.0.0-a.6.9 (29 Jan 2025), following what was done on #48787.

### DIFF
--- a/recipes/plink2/meta.yaml
+++ b/recipes/plink2/meta.yaml
@@ -3,7 +3,7 @@ package:
   version: "2.0.0-a.6.9"
 
 build:
-  number: 1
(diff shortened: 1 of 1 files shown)


tavareshugo:
@BiocondaBot please add label

tavareshugo:
To the reviewer of this PR:

Note that I had to modify the version of `plink2` in the recipe, compared to the original. The original version has a hyphen, which I guess is not supported by Conda, as one of the tests failed because of it. I removed the hyphen, which matches previous versions. I hope that is OK.

PR #55229, opened 2025-04-09
FIX incorrect source for r-scbio

SciLiciumTheo:
After https://github.com/amitfrish/scBio/issues/21: scBio release 0.1.4 has not up-to-date archive assets, resulting in arguments for function `CPM()` of the `scBio` `R` library not present in the conda package when they should be present.

This PR fixes the source url in order to get the correct code.


SciLiciumTheo:
@BiocondaBot please add label

PR #55263, opened 2025-04-11
Update recipe for dysgu

kcleal:
Adds `- osx-arm64`


kcleal:
@BiocondaBot please add label

PR #55269, opened 2025-04-11
eskrim does not work with python 3.13

fplazaonate:


fplazaonate:
@BiocondaBot please add label

PR #55277, opened 2025-04-11
Removed wfa2 external dependency in linux

RedPenguin100:
Supposed to fix this issue
https://github.com/vcflib/vcflib/issues/416


RedPenguin100:
@BiocondaBot please add label

PR #55553, opened 2025-04-21
bioconductor-glmgampoi: add osx-arm64

dslarm:


dslarm:
both osx-64 and osx-arm64 are failing due to curl getting an unrecognized option ('curl -L {url}' - which seems ok according to the man page on my local curl)

dslarm:
@BiocondaBot please add label

PR #55600, opened 2025-04-23
Add osx build to alignoth

fxwiegand:
This pull request updates the build configuration for the `alignoth` package in its `meta.yaml` file. The most significant change is the increment of the build number and the removal of the build skip directive for macOS.

### DIFF
--- a/recipes/alignoth/build.sh
+++ b/recipes/alignoth/build.sh
@@ -3,8 +3,9 @@
 export CFLAGS="${CFLAGS} -O3 -fno-rtti"
 export CPPFLAGS="${CPPFLAGS} -I${PREFIX}/include"
 export LDFLAGS="${LDFLAGS} -L${PREFIX}/lib"
-
 export BINDGEN_EXTRA_CLANG_ARGS="${CFLAGS} ${CPPFLAGS} ${LDFLAGS}"
+
+cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
 
 RUST_BACKTRACE=1
 cargo install --no-track --locked --verbose --root "${PREFIX}" --path .
--- a/recipes/alignoth/build.sh
+++ b/recipes/alignoth/build.sh
@@ -1,11 +1,10 @@
 #!/bin/bash -e
 
-# TODO: Remove the following export when pinning is updated and we use
-#       {{ compiler('rust') }} in the recipe.
-export \
-    CARGO_NET_GIT_FETCH_WITH_CLI=true \
-    CARGO_HOME="${BUILD_PREFIX}/.cargo"
+export CFLAGS="${CFLAGS} -O3 -fno-rtti"
+export CPPFLAGS="${CPPFLAGS} -I${PREFIX}/include"
+export LDFLAGS="${LDFLAGS} -L${PREFIX}/lib"
 
 export BINDGEN_EXTRA_CLANG_ARGS="${CFLAGS} ${CPPFLAGS} ${LDFLAGS}"
 
-cargo install --no-track --verbose --root "${PREFIX}" --path .
+RUST_BACKTRACE=1
+cargo install --no-track --locked --verbose --root "${PREFIX}" --path .
--- a/recipes/alignoth/meta.yaml
+++ b/recipes/alignoth/meta.yaml
@@ -1,23 +1,25 @@
+{% set name = "alignoth" %}
 {% set version = "0.13.0" %}
 
 package:
-  name: alignoth
+  name: {{ name }}
   version: {{ version }}
-
-build:
-  number: 2
-  run_exports:
-    - {{ pin_subpackage('alignoth', max_pin="x") }}
 
 source:
   url: https://github.com/alignoth/alignoth/archive/v{{ version }}.tar.gz
   sha256: a955e762d3593ac572424dbe775f1c554bb5ab10f391fb643c7d613cca416cc4
+
+build:
+  number: 2
+  run_exports:
+    - {{ pin_subpackage('alignoth', max_pin="x.x") }}
 
 requirements:
   build:
     - {{ compiler('c') }}
     - {{ compiler("cxx") }}
     - {{ compiler('rust') }}
+    - cargo-bundle-licenses
     - clangdev
     - pkg-config
     - make
@@ -27,19 +29,23 @@ requirements:
     - libcurl
     - openssl
 
-
 test:
   commands:
     - alignoth --help
 
 about:
-  home: https://alignoth.github.io
+  home: "https://alignoth.github.io"
   license: MIT
-  summary: A tool for creating alignment plots from bam files
+  license_family: MIT
+  license_file: LICENSE
+  summary: "A tool for creating alignment plots from bam files."
+  dev_url: "https://github.com/alignoth/alignoth"
+  doc_url: "https://github.com/alignoth/alignoth/blob/v{{ version }}/README.md"
 
 extra:
   recipe-maintainers:
     - fxwiegand
     - johanneskoester
-  additional-platforms:  #add aarch64
+  additional-platforms:
     - linux-aarch64
+    - osx-arm64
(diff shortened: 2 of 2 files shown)


recipes/alignoth/build.sh
```diff
@@ -1,11 +1,10 @@
...
-    CARGO_NET_GIT_FETCH_WITH_CLI=true \
-    CARGO_HOME="${BUILD_PREFIX}/.cargo"
+export CFLAGS="${CFLAGS} -O3 -fno-rtti"
+export CPPFLAGS="${CPPFLAGS} -I${PREFIX}/include"
+export LDFLAGS="${LDFLAGS} -L${PREFIX}/lib"
 
 export BINDGEN_EXTRA_CLANG_ARGS="${CFLAGS} ${CPPFLAGS} ${LDFLAGS}"
 
-cargo install --no-track --verbose --root "${PREFIX}" --path .
+RUST_BACKTRACE=1
```
dlaehnemann, line 9: Are you sure you want to leave this in? I assume this was for debugging the build and install, and is not actually needed for the final package?

This does seem to incur overhead in the binary, if I understand this answer (and the older ones by shepmaster, above, and the linked GitHub discussion) correctly:
https://stackoverflow.com/a/77797462/2352071

So I would suggest:
```suggestion
```

recipes/alignoth/build.sh
```diff
@@ -1,11 +1,10 @@
 #!/bin/bash -e
 
-# TODO: Remove the following export when pinning is updated and we use
-#       {{ compiler('rust') }} in the recipe.
-export \
-    CARGO_NET_GIT_FETCH_WITH_CLI=true \
-    CARGO_HOME="${BUILD_PREFIX}/.cargo"
+export CFLAGS="${CFLAGS} -O3 -fno-rtti"
+export CPPFLAGS="${CPPFLAGS} -I${PREFIX}/include"
+export LDFLAGS="${LDFLAGS} -L${PREFIX}/lib"
```
dlaehnemann, line 5: I guess this is just the new default? Or should this be explained in some comment?

recipes/alignoth/build.sh
```diff
@@ -1,11 +1,10 @@
...
 
 export BINDGEN_EXTRA_CLANG_ARGS="${CFLAGS} ${CPPFLAGS} ${LDFLAGS}"
 
-cargo install --no-track --verbose --root "${PREFIX}" --path .
+RUST_BACKTRACE=1
+cargo install --no-track --locked --verbose --root "${PREFIX}" --path .
```
dlaehnemann, line 10: Any comment, why the `--locked` is added here? Just trying to understand it.

dlaehnemann (commented):
This generally looks good, but I have some questions... :sweat_smile:

recipes/alignoth/build.sh
fxwiegand: The `--locked` is preventing the installation to update any dependency. This was an issue here for example https://github.com/bioconda/bioconda-recipes/pull/53877.

recipes/alignoth/build.sh
fxwiegand: This just sets the backtrace for the package build doesn't it? Therefore it will come in very handy in case something goes wrong (fingers crossed it doesn't).

recipes/alignoth/build.sh
fxwiegand: Hope it is - copied it over from a much newer package.

dlaehnemann: Fingers crossed... :swe

recipes/alignoth/build.sh
dlaehnemann: Ah, OK. So this basically forces it to use the project's `cargo.lock` file, right? That makes a lot of sense, leaving pinning up to the package's developer, a.k.a. as YOU in this case.

recipes/alignoth/build.sh
dlaehnemann: Yeah, but they didn't do any actual testing on this, whereas the rust-playground issue has some numbers that indicate that there is a penalty during execution (with a previous build with `RUST_BACKTRACE=1`). Maybe we should do some quick benchmarking ourselves? :upside_down_face:

dlaehnemann: OK, I did do some benchmarking and there is basically no detectable difference in runtime between the binary compiled with `RUST_BACKTRACE=1` set and with it `unset`.

I did the following setup:

```
# simulate data with mason
micromamba create -n mason mason
micromamba activate mason
mason_genome --contig-length 2000 --contig-length 5000 --seed 17 --out-file simulated_genome.fa
mason_simulator --input-reference simulated_genome.fa --num-fragments 100000 --seed 13 --out-alignment simulated_reads.bam -o simulated_reads_1.fq.gz -or simulated_reads_2.fq.gz
micromamba deactivate

# prepare data with samtools
micromamba create -n samtools samtools
micromamba activate samtools
samtools sort --output-fmt BAM -o simulated_reads.sorted.bam simulated_reads.bam
samtools index simulated_reads.sorted.bam
micromamba deactivate

# setup alignoth
wget https://github.com/alignoth/alignoth/archive/v0.13.0.tar.gz
tar xzfv v0.13.0.tar.gz
cd alignoth-0.13.0/
# the following assumes an installed recent cargo version
cargo --version

# benchmark with hyperfine
micromamba create -n hyperfine hyperfine
micromamba activate hyperfine
# this is needed to drop the cache before running the command; without dropping the cache, the command somehow runs about 3s longer after an initial hyperfine run...
sudo -v 

# case 1: RUST_BACKTRACE=1
RUST_BACKTRACE=1
cargo install --no-track --locked --verbose --root "${PREFIX}" --path .
taskset --cpu-list 2 hyperfine --prepare 'sync; echo 3 | sudo tee /proc/sys/vm/drop_caches' "bin/alignoth --bam-path simulated_reads.sorted.bam --reference simulated_genome.fa --region 2:500-4500 > plot_all.vl.json"

# case 2: unset RUST_BACKTRACE
unset RUST_BACKTRACE
rm -r bin
cargo install --no-track --locked --verbose --root "${PREFIX}" --path .
taskset --cpu-list 2 hyperfine --prepare 'sync; echo 3 | sudo tee /proc/sys/vm/drop_caches' "bin/alignoth --bam-path simulated_reads.sorted.bam --reference simulated_genome.fa --region 2:500-4500 > plot_all.vl.json"
```

And as results I got:

### case 1: RUST_BACKTRACE=1

```
Benchmark 1: ../alignoth-0.13.0/bin/alignoth --bam-path simulated_reads.sorted.bam --reference simulated_genome.fa --region 2:500-4500 > plot_all.vl.json
  Time (mean ± σ):      5.129 s ±  0.017 s    [User: 4.659 s, System: 0.439 s]
  Range (min … max):    5.107 s …  5.165 s    10 runs
```

### case 2: unset RUST_BACKTRACE

```
Benchmark 1: ../alignoth-0.13.0/bin/alignoth --bam-path simulated_reads.sorted.bam --reference simulated_genome.fa --region 2:500-4500 > plot_all.vl.json
  Time (mean ± σ):      5.127 s ±  0.018 s    [User: 4.655 s, System: 0.441 s]
  Range (min … max):    5.107 s …  5.159 s    10 runs
```

dlaehnemann (approved):
Thanks for the clarifications. This look good!

recipes/alignoth/build.sh
fxwiegand: Wow! Thanks a lot for investigating this. I am even a bit sad that there is no difference as this would've been a nice and easy way to speed up things.

dlaehnemann: There's always the chance that I overlooked something in my benchmarking, which is why I documented it above. But I did try to be as thorough as possible. Including this clearing of the cache and pinning the runs onto the same cpu. So I'm hoping this is our definitive answer (for our tool). Other tools might have more overhead if they use backtrace-related crates, from what I gathered from online discussions. And panics will incur a performance cost, but they will blow up your execution, anyway...

And I finally got around to using hyperfine, which was nice!

PR #55749, opened 2025-04-29
fastq-filter: add linux-aarch64

dslarm:


recipes/fastq-filter/meta.yaml
```diff
@@ -10,7 +10,10 @@ source:
   sha256: 8f28ea8124871b7c2cba2b558b7b64abdcbd4b0fdb7197a82056f7e6202b3e4b
 
 build:
-  number: 3
+  number: 4
+  run_exports:
+    - {{ pin_subpackage(name, max_pin = "x.x") }}
+
   skip: true  # [py < 38]
```
aliciaaevans, line 17: This can be removed since we don't build those lower python versions any more.

dslarm: done

aliciaaevans:

PR #55762, opened 2025-04-29
update build recipe

pdimens:


pdimens:
@aliciaaevans forgive me if you aren't an appropriate person to ping for this. The Linux linting is failing and I cannot understand as to why. It's not providing any clear error message related to the recipe.

aliciaaevans:
@pdimens For some reason, it doesn't like that you are trying to merge into the `bump/harpy` branch. I suggest changing the PR base to `master` instead. Then we can just close the autobump PR.

pdimens:
okie dokie, let's try that. Thank you!

PR #56173, opened 2025-05-14
meme 4.11.2: reenable and add linux-aarch64

dslarm:
This is needed for antismash-lite. 
Package was missing liblzma-devel but now builds on linux-aarch64.
build-fail-blacklist entry removed.


dslarm:
@BiocondaBot please fetch artefacts

dslarm:
@BiocondaBot please add label

PR #56177, opened 2025-05-14
Add fastreer 1.13.20

gkanogiannis:
This adds the `fastreer` package — a Python+Java CLI for computing phylogenetic trees and distance matrices from VCF/FASTA.

- Source: https://github.com/gkanogiannis/fastreeR
- Version: 1.13.20
- License: GPL-3
- JAR and CLI install to `$PREFIX/share/fastreer/` and `$PREFIX/bin`
- Verified with: `bioconda-utils build --docker --mulled-test`

Maintainer: @gkanogiannis


gkanogiannis:

recipes/fastreer/meta.yaml
```diff
@@ -0,0 +1,48 @@
+{% set name = "fastreeR" %}
+{% set version = "1.13.20" %}
+{% set sha256 = "81216e03072d4aca37f920fc4c882d04db16b288157d48674e6a3d192c39a901" %}
+
+about:
+  home: https://github.com/gkanogiannis/fastreeR
+  license: GPL-3
```
bgruening, line 7: Can you please use SPDX identifiers?

recipes/fastreer/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+  run_exports:
+    - {{ pin_subpackage('fastreer', max_pin="x") }}
+
+source:
+  url: https://github.com/gkanogiannis/fastreeR/archive/refs/tags/{{ version }}.zip
+  sha256: {{ sha256 }}
+
+requirements:
+  build:
+    - python >=3.7
```
bgruening, line 31: Can you please use the new python_min syntax.

recipes/fastreer/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+  commands:
+    - fastreeR --version
+    - fastreeR --help
+
+extra:
+  recipe-maintainers:
+    - gkanogiannis
+  identifiers:
+    - doi:10.1186/s12859-016-1186-3
+  skip-lints:
```
bgruening, line 47: This should not be needed.

recipes/fastreer/meta.yaml
gkanogiannis: License updated to `GPL-3.0-only`.

recipes/fastreer/meta.yaml
gkanogiannis: Used `python_min` syntax.

recipes/fastreer/meta.yaml
gkanogiannis: Removed unnecessary `skip-lints`.

recipes/fastreer/meta.yaml
```diff
@@ -0,0 +1,48 @@
+{% set name = "fastreeR" %}
+{% set version = "1.13.20" %}
+{% set sha256 = "81216e03072d4aca37f920fc4c882d04db16b288157d48674e6a3d192c39a901" %}
+{% set python_min = "3.7" %}
```
bgruening, line 4:
```suggestion
```

Try removing it. We should pin hopefully this globally. If this does not work, this is great! Thanks!

gkanogiannis: Removed

PR #56248, opened 2025-05-19
Update polap recipe: build 2

goshng:
Update the meta.yaml with minimap2 >= 2.28

@BiocondaBot please add label


goshng:
@BiocondaBot please add label

PR #56269, opened 2025-05-19
Add MavericK v1.0.5

fonors:
This pull request adds MavericK, a program for inferring population structure from genetic data. MavericK uses a mixture modeling framework similar to the Structure program, but with a different implementation. It analyzes population structure using genetic information and offers a unique approach while following familiar methods.

@bobverity - original author


fonors:
Since the Makefile upstream has g++ hardcoded, it doesn't properly used the compiler toolchain available. Due to the simple nature of the Makefile, I changed the recipe to build the package directly.

fonors:
@BiocondaBot please add label

fonors:
Documentation for using MavericK can be found [here](https://doi.org/10.1534/genetics.115.180992), under Suplementary Data -> FileS2 (PDF format).

fonors:

recipes/maverick/meta.yaml
```diff
@@ -0,0 +1,36 @@
...
+    cp MavericK $PREFIX/bin/
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+
+test:
+  commands:
+    - which MavericK
```
bgruening, line 27: can you run Maverick --help or something like that?

fonors: @bgruening In order not to get an exit status 1, MavericK **must** be run with input data and parameters. I have built a minimal test file and included inline on `run_test.sh`. These files are generated on the fly before MavericK is run. Running MavericK with these test files takes about 2 seconds in the GitHub Action default runner, which I don't think is unreasonable. I am aware this is kind of a dirty hack. Please let me know if this solution is acceptable or if you have other suggestions.

mencian (approved):

PR #56728, opened 2025-06-10
Update PhaBOX's meta.yaml to solve a "ValueError" caused by the update of PyTorch

KennthShang:
Solve a "ValueError" caused by the update of PyTorch

```
ValueError: Due to a serious vulnerability issue in torch.load, even with weights_only=True, we now require users to upgrade torch to at least v2.6 in order to use the function. This version restriction does not apply when loading files with safetensors.
```


KennthShang:
@BiocondaBot please add label

PR #56732, opened 2025-06-10
linux-aarch64 build for humann2

ZerryNi:
Test Passed!
image


mencian:
This package is `noarch`, adding `linux-aarch64` is redundant.

PR #56794, opened 2025-06-12
Update coils and enabled aarch64

ZerryNi:
Test passed!
coils


ZerryNi:
@mencian Hello ,can you help me solve this SSL problem? I can compile the software sucessfully by adding ssl_verify: false to ~/.condarc.
image

ZerryNi (commented):
ok

ZerryNi:
> @mencian Hello ,can you help me solve this SSL problem? I can compile the software sucessfully by adding ssl_verify: false to ~/.condarc. image
@mencian Hello，I've seen your modification,but it seems that is still an SSL problem and it won't pass the IC test

mencian:
@ZerryNi I don't think there is an obvious way to work around the `OpenSSL` certificate error, so `coils` may not be rebuilt :(

PR #56838, opened 2025-06-13
update r-rblast

ZerryNi:
Test passed!


mencian:
There is already a `r-rBLAST` package, so this would be redundant. Though, I added your changes in #56846.

jmarshall:
@ZerryNi @mencian This has created `recipes/r-rblast/*` alongside `recipes/r-rBLAST/*`. Please delete one of them, or unify them as appropriate.

PR #56873, opened 2025-06-13
Update mimick to 2.1

BiocondaBot:
Update [`mimick`](https://bioconda.github.io/recipes/mimick/README.html): **2.0.1** &rarr; **2.1**

[install with bioconda](http://bioconda.github.io/recipes/mimick/README.html) [Conda](https://anaconda.org/bioconda/mimick/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/mimick`](https://github.com//bioconda/bioconda-recipes/tree/bump/mimick/recipes/mimick) (click to view/edit other files)
Summary | Simulate linked-read data.
Home | [https://github.com/pdimens/mimick](https://github.com/pdimens/mimick)
Releases |[]()
Recipe Maintainer(s) | @pdimens
Author | `@pdimens`
***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

PR #56874, opened 2025-06-13
update mimick recipe

pdimens:


pdimens:
I'm trying to update the deps for the recipe for #56873 
unclear why the linting is failing, the error log isn't informative (to me, at least)

PR #56900, opened 2025-06-16
Update PhaBOX to a fixed Transformer package version

KennthShang:
The latest version of the Transformer had a conflict with PyTorch.

So we try to fixed the Transformer as a old and stable version.


KennthShang:
@BiocondaBot please add label

KennthShang:
It seems that it passed all the standard tests but failed in the "build and test (ARM)"

And the detail shows a problem with git clone?

Sounds like a network problem.

PR #56994, opened 2025-06-18
update daisysuite

ZerryNi:
Test passed!
image


ZerryNi:
@mencian Hi，can you help me add noarch:generic to this software? It depends on the sak and guastf,which have benn adapted the linux-aarch64 platform and have jist benn integrated.And,I download the source conde package and introduce the local path in meta.yaml.The compilation is sucessful.
image

PR #57004, opened 2025-06-18
update perl-io-prompt

ZerryNi:
Test passed!


recipes/perl-io-prompt/meta.yaml
```diff
@@ -24,18 +24,20 @@ source:
 
 build:
   noarch: generic
-  number: 3
+  number: 5
```
bgruening: Why not 4?

ZerryNi: I've revised it,please review it.

PR #57044, opened 2025-06-19
update deltapd

ZerryNi:
Test passed!
image


ZerryNi:
@mencian Hi,This software is not compiled sucessfully,I see your modification,but stillfailed the test.

PR #57052, opened 2025-06-19
add htslib-1.10 and enbaled linux-aarch64

ZerryNi:
Test passed!
image


ZerryNi:
@mencian Hi,can you help me find this problem? My local build has been successful.On this environment,no matter how L modify it , it fails to pass the test.Demuxlet depends on htslib1.10.If htslib of a later version is  used,an error will be reported.Therefore,htslib1.10 needs to be adapted to the linux-aarch64 platform.

PR #57100, opened 2025-06-20
update remura enabled aarch64

ZerryNi:
Test passed!


cfmin09:
@mencian ，I need build "remurna"，please help remove it from the build fail blacklisted

PR #57122, opened 2025-06-20
Add faststructure-static

fonors:
fastStructure is a fast algorithm for inferring population structure from large SNP genotype data. It is based on a variational Bayesian framework for posterior inference and is written in Python 2.x. Due to this, it's limited to environments running Python 2.x, and as such, can't be used in modern workflows. This pull requests aims to add a slightly modified recipe to build a statically compiled binary that can be used in any environment.

Since this has some ancient dependencies for building, any help would be appreciated.

A working build script (at least on Ubuntu 12.04) can be found [here](https://gitlab.com/StuntsPT/Structure_threader/-/raw/master/helper_scripts/install_faststructure.sh).

Original author - @rajanil


fonors:
Unnecessary due to pull request #57289. Closing.

PR #57159, opened 2025-06-23
update Ripser

ZerryNi:
Test passed!


cfmin09:
@mencian , please help me, (bioconda/ripser) fail to build, reporting"A package of the same name already exists in another channel " (conda-forge/ripser), now what should I do now? They have the same name but different source codes.

mencian:
`conda-forge` takes precedence so the recipe will have be under a different name, i.e., `bio-ripser`. I've updated `bio-ripser` to v`1.2.1` in #57179. Closing in favor of #57179.

cfmin09:
> `conda-forge` takes precedence so the recipe will have be under a different name, i.e., `bio-ripser`. I've updated `bio-ripser` to v`1.2.1` in #57179. Closing in favor of #57179.

ths

PR #57172, opened 2025-06-23
Enable DMRcate

snowpong:
Enable DMRcate to hopefully succeed in 3.20

(re-trying, as previous PR had renamed the file)

PR #57196, opened 2025-06-24
Add IlluminaHumanMethylationEPICv2anno.20a1.hg38 and IlluminaHumanMethylationEPICv2manifest

snowpong:
Adding packages `bioconductor-illuminahumanmethylationepicv2anno.20a1.hg38`
and `bioconductor-illuminahumanmethylationepicv2manifest`.
I've copied the recipes from their V1 equivalents and used those as templates. Essentially I've update the names and versions, added r-utils, and updated the md5s. Hopefully that is all that is needed.

These two package are dependencies that missMethyl (currently blacklisted) rely on, so hopefully this can get missMethyl working again, and then after that DMRcate. See https://github.com/bioconda/bioconda-recipes/pull/57172#issuecomment-2998906860


snowpong:
So it's currently failing because:

`11:21:03 BIOCONDA INFO (OUT) + echo 'ERROR: post-link.sh was unable to download any of the following URLs with the md5sum null:'
`

and that is because I haven't added the md5 and urls to dataURLS.json - but it's not clear to me if this is something I should be doing, either manually, or through a script, or if this should somehow happen automatically.

snowpong:
> and that is because I haven't added the md5 and urls to dataURLS.json - but it's not clear to me if this is something I should be doing, either manually, or through a script, or if this should somehow happen automatically.

It seems the proper way to do this is to use `bioconda-utils bioconductor-skeleton` to both make the meta.yml and update dataURLs.json. But, that revealed a bug in [bioconductor-skeleton](https://github.com/bioconda/bioconda-utils/issues/1052).

snowpong:
All passes. Ready for review. Don't know if that happens automatically or I need to do something?

snowpong:
@BiocondaBot please add label

snowpong:
@mencian 🙏 thanks

PR #57332, opened 2025-06-27
enable dmrcate and bump build version

snowpong:
enable dmrcate and bump build version


snowpong:
@BiocondaBot please add label

PR #57420, opened 2025-07-02
update flams and enabled linux-aarch64

ZerryNi:
Test passed!
image


ZerryNi:
@BiocondaBot please add label

bgruening:
What is in this PR is not what is written in the PR description.

mencian:
@ZerryNi `flams` is a `noarch` package, so adding `additional-platforms` would be redundant.

ZerryNi:
> @ZerryNi `flams` is a `noarch` package, so adding `additional-platforms` would be redundant.

ok,but the versoin for blast  still requrires modification,as version 2.13.0 is not compatible with  linux-aarch64
image

PR #57560, opened 2025-07-08
update fusioncatcher

ZerryNi:
Test passed


ZerryNi:
@BiocondaBot please add label

ZerryNi:
image
@mencian Hi,This software does not seem to be successfully installable on the linux-aarch64 platform.It appears that the PR modified pigz to not restrict the version,but during installation,it still reports the origin version 2.3.

PR #57568, opened 2025-07-08
Add: excavator2

danilotat:
Addition of [excavator2](https://github.com/ctglab/excavator2) v.1.0.1 to Bioconda.

### DIFF
--- a/recipes/excavator2/build.sh
+++ b/recipes/excavator2/build.sh
@@ -10,6 +10,7 @@ mkdir -p ~/.R
 # Let R/conda manage compilers, only add legacy flags
 cat <<EOF > ~/.R/Makevars
 FFLAGS += -std=legacy -ffixed-line-length-none -w
+SHLIB_LDFLAGS += -undefined dynamic_lookup
 EOF
 
 find "${SOURCE_SUBDIR}" -name "*.so" -delete


danilotat:
@BiocondaBot please add label

dpryan79:

recipes/excavator2/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+DEST_DIR="${PREFIX}/share/${PKG_NAME}"
+mkdir -p "${DEST_DIR}"
+SOURCE_SUBDIR="${SRC_DIR}/${PKG_NAME}"
+# copy in share
+cp -r "${SOURCE_SUBDIR}"/* "${DEST_DIR}/"
```
dpryan79, line 19: This is copying some precompiled libraries around, which will likely fail at some point (certainly on OSX!). Can we compile things properly instead like a normal package?

dpryan79 (dismissed):

recipes/excavator2/build.sh
danilotat: Hi @dpryan79 , thanks for the review. We fixed this in the new release. Should I edit this PR or open another one?

dpryan79: Go ahead and update this one, no need for a new PR :)

danilotat:

recipes/excavator2/build.sh
```diff
@@ -0,0 +1,18 @@
...
+for lib_file in ${DEST_DIR}/lib/F77/*.f ; do
```
dpryan79: I have this making the 2 object files now, but I assume you need those combined into a library or something (i.e., a .so or a .dylib on OSX). Do you have a way to test that these are being loaded properly?

danilotat: They're separated libraries that are loaded from R scripts. I may add also a smoke test for ensuring that they could be loaded.

PR #57596, opened 2025-07-09
Update ont-modkit (fix installed version)

tdido:
Current bioconda package v0.5.0 actually installs v0.4.5. This fixes it in a new build.


tdido:
@BiocondaBot please add label

PR #57653, opened 2025-07-11
update maxentpy

ZerryNi:
Test passed


ZerryNi:
@BiocondaBot please add label

recipes/maxentpy/meta.yaml
```diff
@@ -19,8 +19,9 @@ requirements:
     - {{ compiler('c') }}
   host:
     - pip
-    - python
-    - cython
+    - python >=3.8,<3.10
```
dpryan79: Does it not work with python >=3.10?

mencian (approved):

PR #57748, opened 2025-07-16
Bumped STAR version in STAR-Fusion recipe

micknudsen:
This PR bumps the pinned version of STAR in the STAR-Fusion recipe. The current version of STAR is `v2.7.11b,` but the version is pinned to `v2.7.11a` in the recipe.

I don't think there is a way to match both versions in conda because of the `a/b` thing, so my best solution is to make a separate build.


micknudsen:

micknudsen:

PR #58027, opened 2025-07-25
Add ProMod3 recipe

xrobin:
Add new recipe for [ProMod3](https://openstructure.org/promod3) v3.5.0

----

### General instructions

```yaml
build:
  run_exports:
    - ...

```

with `...` being one of:

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

</details>


recipes/promod3/build.sh
```diff
@@ -0,0 +1,44 @@
...
+    CPU_COUNT=8
+  fi
+  export CPU_COUNT
+  echo "CPU_COUNT for build = $CPU_COUNT"
+fi
+if [[ $(uname -m) == "x86_64" ]]; then
+    export CMAKE_ARGS="${CMAKE_ARGS} -DENABLE_SSE=1"
+else
+    export CMAKE_ARGS="${CMAKE_ARGS} -DENABLE_SSE=0"
+fi
```
eunos-1128, line 19: I don't think these lines are needed because there are no heavy GUI-related dependencies as with openstructre.

recipes/promod3/build.sh
```diff
@@ -0,0 +1,44 @@
...
+# Prevent build failures due to insufficient memory in the CI environment
+# Use parallel build because of serial build on osx-arm64 causing occasional errors
+if [[ "${build_platform}" == "linux-aarch64" || "${build_platform}" == "osx-arm64" ]]; then
+  export CPU_COUNT=$(( CPU_COUNT * 70 / 100))
+  if [ "$CPU_COUNT" -gt 8 ]; then
+    CPU_COUNT=8
+  fi
+  export CPU_COUNT
+  echo "CPU_COUNT for build = $CPU_COUNT"
+fi
```
eunos-1128: I don't think these lines are needed because there are no heavy GUI-related dependencies for building as with openstructure.

recipes/promod3/build.sh
```diff
@@ -0,0 +1,44 @@
...
+if [[ $(uname -m) == "x86_64" ]]; then
+    export CMAKE_ARGS="${CMAKE_ARGS} -DENABLE_SSE=1"
+else
+    export CMAKE_ARGS="${CMAKE_ARGS} -DENABLE_SSE=0"
+fi
+
+if [[ "${target_platform}" == "linux-"* ]]; then
+    export LDFLAGS="${LDFLAGS} -Wl,--allow-shlib-undefined,--export-dynamic"
+elif [[ "${target_platform}" == "osx-"* ]]; then
+    export LDFLAGS="${LDFLAGS} -undefined dynamic_lookup -Wl,-export_dynamic -framework OpenGL"
```
eunos-1128: `-framework OpenGL` could not be necessary.

recipes/promod3/conda_build_config.yaml
```diff
@@ -0,0 +1,5 @@
+c_compiler_version: # [osx]
+  - 20 # [osx]
+
+cxx_compiler_version: # [osx]
+  - 20 # [osx]
```
eunos-1128: Since openstructure is not built this time, this file can be deleted.

recipes/promod3/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - python {{ python }}
+    - libboost-devel >=1.86,<1.87
+    - libboost-python-devel >=1.86,<1.87
+    - openmm >=7
+    - eigen
+    - openstructure >=2.11.0
+  run:
+    - python {{ python }}
+    - {{ pin_compatible("openstructure") }}
+
```
eunos-1128, line 35: Add `- {{ pin_compatible("openmm") }}` because `openmm` has no `run_exports` in its recipe. 

In this case, openstructure has openmm as a run dependency, so openmm will come with openstructure. But I think writing explixitly is better.

recipes/promod3/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+  host:
+    - python {{ python }}
+    - libboost-devel >=1.86,<1.87
+    - libboost-python-devel >=1.86,<1.87
+    - openmm >=7
+    - eigen
+    - openstructure >=2.11.0
+  run:
+    - python {{ python }}
+    - {{ pin_compatible("openstructure") }}
```
eunos-1128: I don't think `- {{ pin_compatible("openstructure") }}` is required since `openstructure` has `run_exports` in its recipe.

recipes/promod3/build.sh
```diff
@@ -0,0 +1,44 @@
...
+
+mkdir -p build && cd build
+
+cmake .. \
+    ${CMAKE_ARGS} \
+    -DCMAKE_PREFIX_PATH="${PREFIX}" \
+    -DOST_ROOT="${PREFIX}" \
+    -DCMAKE_CXX_COMPILER="${CXX}" \
+    -DCXX_FLAGS="${CXXFLAGS}" \
+    -DCMAKE_CXX_STANDARD=11 \
```
eunos-1128: The errors in osx builds are probably caused by the following line remaining, which is preventing it from being overwritten by `-DCMAKE_CXX_STANDARD=11`.

https://git.scicore.unibas.ch/schwede/ProMod3/-/blob/conda/cmake_support/PROMOD3.cmake?ref_type=heads#L914

eunos-1128:
Creating `run_test.py` which has code like [these lines](https://github.com/brewsci/homebrew-bio/blob/develop/Formula/promod3.rb#L61-L66) in the recipe dir will enable python bindings tests.

eunos-1128:

eunos-1128:

recipes/promod3/build.sh
eunos-1128: openstructure's shared libraries in the conda package, which is already built, are used to build promod3. openstructure won't be built this time.

xrobin:
Thanks for the review @eunos-1128 , I will try your suggestions next week.

Maybe you can help me with this too: the CircleCI builds time out after an hour, even without running make check. Promod3 is notably slow to build, that's well know, and as it's repeated with 4 different Pythons. With all the conda overhead, it's just not going to work. Disabling optimizations isn't an option as Promod would then be too slow at runtime (and actually I'm not sure if it would even build).

I noticed that some of the openstructure runs took over 2 hours on CircleCI. Do you know how that happened and where the difference comes from? Any advice would be welcome here.

recipes/promod3/meta.yaml
xrobin: But then I probably need to move it to the build section too. That would actually make sense.

recipes/promod3/build.sh
xrobin: Promod3 is making heavy use of c++ templating and is a beast to build too. I will try removing it just to see if it helps.

recipes/promod3/meta.yaml
eunos-1128: There is no need to put `openstructure` in the build section, since **`openstructure >=2.11.0` in the host section is used** to build promod3, and even if it were in the build section, it is probably not used.

The `build` section is mainly used for the compiler and command line tools used only in the build environment.

https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#requirements-section

recipes/promod3/build.sh
eunos-1128: In my experience, heavy compilation of C++ is sometimes problematic, and I have the impression that GUI-related builds often fail in case of insufficient resources.

recipes/promod3/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+  run_exports:
+    - {{ pin_subpackage(name|lower, max_pin="x.x") }}
+
+requirements:
+  build:
+    - {{ compiler("cxx") }}
+    - cmake
+    - make
+    - pkg-config
+    - wget
```
eunos-1128: `wget` can be removed.

recipes/promod3/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - openstructure >=2.11.0
```
eunos-1128: As replied, I think it would not be used.

recipes/promod3/build.sh
```diff
@@ -0,0 +1,35 @@
...
+mkdir -p build && cd build
+
+cmake .. \
+    ${CMAKE_ARGS} \
+    -DCMAKE_PREFIX_PATH="${PREFIX}" \
+    -DOST_ROOT="${PREFIX}" \
+    -DCMAKE_CXX_COMPILER="${CXX}" \
+    -DCXX_FLAGS="${CXXFLAGS}" \
+    -DOPTIMIZE=ON \
+    -DDISABLE_DOCUMENTATION=1
```
eunos-1128: Add `-DPython_EXECUTABLE="${PYTHON}"`. 

I can see errors that the location of the Python headers and libraries is missing. This should solve the problem.

eunos-1128 (changes requested):

eunos-1128:
@xrobin 

I suspect build failures are due to https://github.com/bioconda/bioconda-recipes/pull/58027#discussion_r2247428863 (both for linux and osx) and https://github.com/bioconda/bioconda-recipes/pull/58027#discussion_r2245959676 (for osx)

The formula of [promod3 for homebrew](https://github.com/brewsci/homebrew-bio/blob/develop/Formula/promod3.rb) I wrote may help you.

eunos-1128 (commented):
See new comments on the recipe

eunos-1128:
@xrobin 

> I noticed that some of the openstructure runs took over 2 hours on CircleCI. Do you know how that happened and where the difference comes from? Any advice would be welcome here.

I have no idea about this. 

Has anything changed in CircleCI's resource settings, @bioconda/core?

recipes/promod3/meta.yaml
xrobin: Just to clarify, the goal is that promod3 built with openstructure 2.11 doesn't run with an other version like 2.12. Do you think the definition we have prevents this? If so, we're good 👍

eunos-1128: >  Do you think the definition we have prevents this?

Yes, it should do.

When seeing [openstructure's recipe](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/openstructure/meta.yaml#L14-L15), there is `{{ pin_subpackage(name|lower, max_pin="x.x") }}` as `run_exports`.

This means that when you install a package with `openstructure >=2.11` written in the host section (i.e. this recipe), you will install `openstructure 2.11.x`, which is the same as having `openstructure >=2.11,<2.12` in the run section.

In other words, in this case 2.11.1 or 2.11.7 etc... can be installed, but 2.12.0, 2.12.7 etc... are cannot.

It should work as you expect it to.

recipes/promod3/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - {{ pin_compatible("openmm") }}
+
+test:
+  source_files:
+    - doc/tests/data/1crn.fasta
+    - doc/tests/data/1crn_cut.pdb
+    - doc/tests/data/1CRNA.hhm
+  commands:
+    - pm 2>&1 | grep -q Usage
+
```
eunos-1128, line 48: You can check if the Python module was installed successfully by trying `import promod3`.

```suggestion
  imports:
    - promod3
```

recipes/promod3/build.sh
xrobin: @schdaude please try

xrobin:
I have [this build for OpenStructure in mind for instance, which ran less than two months ago and completed in 2h53](https://app.circleci.com/pipelines/github/bioconda/bioconda-recipes/157713/workflows/63be7ecc-3f62-4cc6-8277-033940af6d29/jobs/285927?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary). Promod3 should build faster than that, and I think we are just running over the 1 hour mark.

eunos-1128:
@xrobin 

I opened [PR](https://github.com/bioconda/bioconda-recipes/pull/58380) to verify building based on this recipe.

[Deleting `${Python_LIBRARIES}`](https://github.com/eunos-1128/bioconda-recipes/blob/1c57d4b8432148ad083c07ca80db8dc8e93acb50/recipes/promod3/build.sh#L17) using `sed` resolved errors which bring about build failures on macOS. 

Fixing it in the upstream source should eliminate the error.

Apparently, the wrong libpython is linked, which causes a segmentation error at building if this is not done.

By the way, [using `ninja` instead of `make`](https://github.com/bioconda/bioconda-recipes/pull/58380/files#diff-84a19e278a5b04ed927774c909582e33ab808af56f1c5e7791934bfedc2719a9R23) seems to speed up the build a bit, and it seems to build python 3.10-3.12 builds reliably when building on linux-aarch64.

For your information.

xrobin:
Thanks a lot @eunos-1128 ! Now that I see it, `${Python_LIBRARIES}` is actually the same fix as what you did for OpenStructure. Good catch! I included the fix upstream directly.

I included your other changes as well, they seem to make a lot of sense.

Let's see if I did it right and the builds complete now.

eunos-1128:
@xrobin 

`Linux` is incorrect. `linux` is correct.

You needn't even add `linux` as `aarch64` includes `linux`.


linux | True if the platform is Linux.
-- | --

aarch64 | True if the platform is Linux and the Python architecture is aarch64.
-- | --

> Defining metadata (meta.yaml) — conda-build 0.0.0.dev0+placeholder documentation https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#preprocessing-selectors

eunos-1128:
`ninja -C build check` can be replaced with `cmake --build build --target check --parallel "${CPU_COUNT}"`.

xrobin:
> `ninja -C build check` can be replaced with `cmake --build build --target check --parallel "${CPU_COUNT}"`.

Unfortunately checks won't run in parallel as several tests touch the same files.

Let's see if your other fix is enough (somehow I thought aarch applied to Mac as well)

eunos-1128:
@xrobin 

Try `cmake --build build --target check --parallel 1` instead of `ninja -C build check`.

`ninja` performs parallel processing by default, so you may need to explicitly set it to serial build.

eunos-1128:

eunos-1128:
@BiocondaBot please add label

PR #58068, opened 2025-07-28
update rtk and enabled to lniux-aarch64

ZerryNi:
Test passed!


mencian:
There are no changes in this PR.

PR #58193, opened 2025-08-01
Update meta.yaml

simroux:
Minor version bump to 0.9.7 - fix bug in create db step for extra function "map to target"


simroux:
thanks @mencian  for the review and improvements :-)

PR #58265, opened 2025-08-04
checkm-genome: bump required pplacer version - to bring in linux-aarch64/osx-arm64

dslarm:
* set required pplacer version to 1.1-alpha20,  to bring in linux-aarch64/osx-arm64


mencian (approved):
Now able to pull in pplacer-alpha.20 with two pins

PR #58362, opened 2025-08-07
Add clasv

Ebenco36:
Lassa virus rapid lineage assignment using Random Forest.

### DIFF
--- a/recipes/clasv/meta.yaml
+++ b/recipes/clasv/meta.yaml
@@ -1,9 +1,13 @@
+{% set name = "CLASV" %}
+{% set version = "1.0.0" %}
+
 package:
-  name: clasv
-  version: "0.1.0"
+  name: {{ name|lower }}
+  version: {{ version }}
 
 source:
-  path: .
+  url: https://github.com/JoiRichi/CLASV/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: d5558cd419c8d46bdc958064cb97f963d1ea793866414c025906ec15033512ed
 
 build:
   noarch: python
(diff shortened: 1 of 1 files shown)


recipes/clasv/meta.yaml
```diff
@@ -0,0 +1,154 @@
...
+
+build:
+  number: 0
+  noarch: python
+  run_exports:
+    - {{ pin_subpackage(name|lower, max_pin="x.x") }}
+  script:
+    - echo "Contents of work directory:"
+    - set +e
+    - chmod +x $HOME/.local/bin/nextclade
```
bgruening: why is this needed? This should not be install the package into the HOME but in the site-package dir for conda. Please have a look at other Python packages.

Ebenco36: Thanks for the feedback—I’ve updated accordingly.

recipes/clasv/meta.yaml
```diff
@@ -0,0 +1,149 @@
...
+  run_exports:
+    - {{ pin_subpackage(name|lower, max_pin="x.x") }}
+  script:
+    - "{{ PYTHON }} -m pip install ."
+    - "{{ PYTHON }} -c 'from CLASV.install_nextclade import install_nextclade; install_nextclade()'"
+
+channels:
+  - conda-forge
+  - bioconda
+  - defaults
```
bgruening:
```suggestion
```

recipes/clasv/meta.yaml
```diff
@@ -0,0 +1,149 @@
...
+    - setuptools
+    - biopython
+    - pandas
+    - requests
+    - nextclade
+    - plotly
+    - joblib
+    - snakemake
+    - scikit-learn
+
```
bgruening:
```suggestion
```

recipes/clasv/meta.yaml
```diff
@@ -0,0 +1,149 @@
...
+  run:
+    - python
+    - nextclade
+    - appdirs ==1.4.4
```
bgruening, line 30: all those very strict pins should be removed.

A conda package is __not__ a conda environment. Its purpose is to be easy to mix and match with other packages.

PR #58385, opened 2025-08-08
update  checkm-genome and enabled linux-aarch64

ZerryNi:
Tsest passed!


ZerryNi:
@BiocondaBot please add label

bgruening:
This only has a bump, is that intended?

ZerryNi:
> This only has a bump, is that intended?

I'm sorry,perhaps someone has already updated it.

ZerryNi:
> This only has a bump, is that intended?

I need to adapt to version 1.2.1 to support the linux-aarch64 platform.

mencian (approved):

PR #58475, opened 2025-08-12
Add distree recipe

Paururo:
Add distree,  a command-line tool written in Rust that extracts a distance matrix from a phylogenetic tree in Newick format. It is designed to handle large trees with thousands of sequences by using a low-memory, parallelized approach.


Paururo:
@BiocondaBot please add label

bgruening (approved):

PR #58476, opened 2025-08-12
Add samtools dependency to PBSIM3

yukiteruono:
This PR adds samtools (>=1.10) to both host and run requirements
to fix missing dependency in the conda package.


mencian:
Closing in favor of #58473.

PR #58504, opened 2025-08-13
Update cuna 0.3.0 pin pyarrow

iris1901:
- Pin pyarrow =20.0.0 to avoid incompatibilities.
- Keep pod5 without version to use default available in bioconda/conda-forge.


iris1901:
@BiocondaBot please add label

PR #58566, opened 2025-08-15
update tgt and enabled linux-aarch64

ZerryNi:
Test passed!


ZerryNi:

aliciaaevans:

recipes/tgt/meta.yaml
```diff
@@ -31,3 +33,7 @@ about:
   license: GPL-3.0+
   license_family: GPL
   summary: 'TextGridTools -- Read, write, and manipulate Praat TextGrid files'
+
+extra:
```
aliciaaevans: Since this is noarch, it doesn't actually need a linux-aarch64 build. You can see the duplicate packages in the artifacts comment.

PR #58660, opened 2025-08-19
update xunit-wrapper and enabled linux-aarch64

ZerryNi:
Test passed! The software galaxy-parsec depends on it ,so it needs to be adapted for the linux-aarch64 platform.


ZerryNi:

PR #58661, opened 2025-08-19
update metawrap-mg recipe

ZerryNi:
Test passed!


ZerryNi:
@BiocondaBot please add label

ZerryNi:
@mencian Hi,please help me review the code.Thank you.

bernt-matthias:
Hi @mencian and @ZerryNi ~metabat2~ metawrap strictly requires python2. I would like to undo the change (and maybe remove the package + container that have been built from this commit?).

What would be the best procedure?

bernt-matthias:
Also the tool has clearly documented dependencies https://github.com/bxlab/metaWRAP/blob/master/installation/dependancies.md?plain=1 ...

ZerryNi:
@bernt-matthias Hi,when I tried to install this software on the linux-aarch64 platform,it reported an error.Could you please help me take a look? Thank you!
image

bernt-matthias:
Dear @ZerryNi . I understand. Unfortunately the solution you suggested here breaks metawrap, i.e. you can install it, but its broken.

Problem is that metawrap is just super old and seems unmaintained. IMO metawrap only contains a one unique part: its bin_refinement module; all other modules just call other software (sometimes they wrap just a single program). 

Since building the recipe with the full set of requirements seems impossible, my idea is to split the recipe such that it produces multiple packages: one per module: https://github.com/bioconda/bioconda-recipes/pull/59915. This will simplify the problem of incompatible packages. 

Alternatives would be: 

- setup a community maintained fork of metawrap (and remove python2)
- add patches to bioconda that remove python2
- ... ?

PR #59246, opened 2025-09-10
Update cargo install to use --all-features flag

donaldcampbelljr:
We are changing our rust-gtars crate to gatekeep some features behind feature flags. Therefore, we are changing the build command to use the --all-features flag.


nleroy917:
FYI: the path in that `cargo install` command might need to change to something like `./gtars-cli` instead

PR #59273, opened 2025-09-11
CryptoGenotyper v1.5.0 build 2 re-name post-link.sh fixes database init

kbessonov1984:
I did not name correctly the `post-link.sh` file so the databases were not initialized during conda install causing downstream writing permission errors in Galaxy wrapper. I misnamed `post-link.sh` as `post_link.sh` in my previous build. This will initialize the databases during install. We want to release Galaxy Project wrapper today https://github.com/phac-nml/galaxy_tools/pull/275. It will be released on Galaxy EU to replace version 1.0 that is hosted there

### DIFF
--- a/recipes/cryptogenotyper/post-link.sh
+++ b/recipes/cryptogenotyper/post-link.sh
@@ -1,0 +1,2 @@
+#!/usr/bin/env bash
+cryptogenotyper_init


kbessonov1984:
@BiocondaBot please add label

kbessonov1984:
@rpetit3 Please help us release this small fix so that we can make conda recipe compatible with read only installs.

PR #59287, opened 2025-09-12
Update sierra local

svarona:
- Fixed nucamino binary
- Made python scripts executable in bin
----

### General instructions

```yaml
build:
  run_exports:
    - ...

```

with `...` being one of:

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/recipes/sierra-local/build.sh
+++ b/recipes/sierra-local/build.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+set -ex
+
+# instalar con pip
+$PYTHON -m pip install . -vvv --no-deps --no-build-isolation --no-cache-dir
+
+# Wrap scripts from sierralocal and add shebang
+for f in sierralocal/*.py; do
+    name=$(basename $f .py)
+    cat > ${PREFIX}/bin/$name << EOF
+#!/usr/bin/env bash
+exec "$PYTHON" -m sierralocal.$name "\$@"
+EOF
+    chmod +x ${PREFIX}/bin/$name
+done
+
+# Wrap scripts from scripts/ and add shebang
+for f in scripts/*.py; do
+    name=$(basename $f .py)
+    cp "$f" "$PREFIX/bin/$name.py"
+    cat > "$PREFIX/bin/$name" << EOF
+#!/usr/bin/env bash
+exec "$PYTHON" "$PREFIX/bin/$name.py" "\$@"
+EOF
+    chmod +x "$PREFIX/bin/$name" "$PREFIX/bin/$name.py"
+done
--- a/recipes/sierra-local/meta.yaml
+++ b/recipes/sierra-local/meta.yaml
@@ -15,8 +15,6 @@ build:
   noarch: generic
   run_exports:
     - {{ pin_subpackage("sierra-local", max_pin="x") }}
-  script: 
-    - $PYTHON -m pip install . -vvv --no-deps --no-build-isolation --no-cache-dir
   binary_relocation:
     - false
   ignore_prefix_files:


svarona:

recipes/sierra-local/build.sh
```diff
@@ -0,0 +1,26 @@
...
+    name=$(basename $f .py)
+    cat > ${PREFIX}/bin/$name << EOF
+#!/usr/bin/env bash
+exec "$PYTHON" -m sierralocal.$name "\$@"
+EOF
+    chmod +x ${PREFIX}/bin/$name
+done
+
+# Wrap scripts from scripts/ and add shebang
+for f in scripts/*.py; do
```
bgruening, line 18: Can you explain why this detour is needed?

svarona: I want some of the python scripts in the repo to be executable through command line, and they are in different folders. So i'm adding them to the bin folder and adding the sheban so I can execute them as `updater.py <whatever>`

recipes/sierra-local/meta.yaml
```diff
@@ -11,12 +11,14 @@ source:
...
+  number: 1
   noarch: generic
   run_exports:
     - {{ pin_subpackage("sierra-local", max_pin="x") }}
-  script: 
-    - $PYTHON -m pip install . -vvv --no-deps --no-build-isolation --no-cache-dir
+  binary_relocation:
+    - false
+  ignore_prefix_files:
+    - sierralocal/bin/nucamino-linux-amd64
```
bgruening, line 21: Is there a way to compile this binary?

We usually don't want to ship precompiled assets.

recipes/sierra-local/build.sh
bgruening: Using the entry_point magic from python packages does not work? Like here: https://github.com/bioconda/bioconda-recipes/blob/424a758f1da0e378c1e7ec9cdaa1396750ff0d60/recipes/brawn/meta.yaml#L19

Thanks @svarona!

svarona: Good point 👌, there’s a key thing here.

The Python scripts in the repo were witten to also be directly executable from the command line as if they were binaries, for example:

`updater.py -updater_outdir data/`

The original [`updater.py`](https://github.com/PoonLab/sierra-local/blob/master/sierralocal/updater.py) script defines a `main(updater_outdir=None)` that expects an explicit argument, and in the `if __name__ == "__main__"`: block it already has an argparse parser that allows -updater_outdir.

The wrapper that conda generates with entry_points just calls main() without passing anything from sys.argv so the option `-updater_outdir` is never recognized.

I don't know how I could manage this rather than copying the script to main and add the shebang but I'm really open to suggestions! Thank you very much for your help!

recipes/sierra-local/meta.yaml
svarona: What you’re proposing would mean recompiling the binaries (nucamino, postalign) from source instead of using the precompiled ones bundled with sierra-local? Like creating their recipes? I don't know if they will remain fully compatible with all the Python scripts in the package.

I decided not to do this when I initially though about this because the original binaries are already tested and validated with sierra-local and recompiling them could break functionality or introduce incompatibilities with the Python code.

If this is a concern, I could try creating separate recipes for the other binaries and test if it works. However, since sierra-ocal itself is not my package, I’m not sure how recompiling these binaries would affect its functionality or compatibility with the Python code.

Let me know your thoughts and thank you for your help!

recipes/sierra-local/build.sh
bgruening: Sorry, hadn't time to look at the upstream code, but it works for all others. Maybe have a look at deeptools, they have a lot of those

PR #59293, opened 2025-09-12
Add plasmaag

Las02:
Add PlasMAAG, a tool to recover plasmids and organisms from metagenomic samples.
repo: https://github.com/RasmussenLab/PlasMAAG
article: https://www.biorxiv.org/content/10.1101/2025.02.26.640269v2.abstract


Las02:
@BiocondaBot please add label

daler (commented):
@Las02 the listed host Python (`==3.11.0`) does not match the listed runtime Python (`==3.11.2`) which is kind of strange.

But more broadly, are all of these strict pinnings necessary? This will restrict the environments in which this package can be installed.

PR #59349, opened 2025-09-16
update r-traminer and enabled linux-aarch64

ZerryNi:


recipes/r-traminer/build.sh
```diff
@@ -0,0 +1,56 @@
...
+# 首先安装缺失的依赖包（确保在构建环境中可用）
+echo "Installing required R dependencies..."
+${R_HOME}/bin/R -e "
+install.packages(
+  c('colorspace', 'RColorBrewer', 'boot', 'vegan', 'cluster', 'MASS', 'nnet'),
+  repos = 'https://cloud.r-project.org/',
+  dependencies = TRUE,
+  INSTALL_opts = '--no-test-load'
+)
+"
```
daler: these should be dependencies in the recipe itself, not installed separately

recipes/r-traminer/build.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+set -ex
+
+# 设置构建环境
```
daler: can you change comments to english to be consistent with the rest of the recipes?

PR #59430, opened 2025-09-18
update PanTools to version v4.3.4

PFlege-cmd:
update PanTools to version v4.3.4


PFlege-cmd:
@BiocondaBot please add label

PR #59438, opened 2025-09-18
Update SnpEff v5.0, v5.1, v5.2 to use AstraZeneca's repository

kysrpex:
Microsoft's SnpEff repository has been deactivated. Data has been migrated to a new repository provided by AstraZeneca.

URLs have been already updated in the configuration file of SnpEff v5.3, but older versions fail to run. Afaik, SnpSift only uses annotations from local files, so the recipes would not need to be updated.

This PR patches snpEff.config so that it points to the new URLs.

### DIFF
--- a/recipes/snpeff/5.0/meta.yaml
+++ b/recipes/snpeff/5.0/meta.yaml
@@ -1,0 +1,41 @@
+{% set snpeff_ver = "v5_0e" %}
+# NOTE: if the version contains a trailing letter, use the <d>.<d>.1<l> format
+# Please update recipes/snpsift as well!
+{% set version = "5.0" %}
+{% set sha256 = "85d907b5dd9e9008a0cf245956e3c9077a31e45f21a1b580d9b98a53fd8dcb9d" %}
+
+about:
+  home: 'http://snpeff.sourceforge.net/'
+  license: "LGPLv3"
+  summary: "Genetic variant annotation and effect prediction toolbox"
+
+package:
+  name: snpeff
+  version: {{ version }}
+
+build:
+  number: 2
+  noarch: generic
+  run_exports:
+    - {{ pin_compatible('snpeff', max_pin="x.x") }}
+
+source:
+  url: https://snpeff.odsp.astrazeneca.com/versions/snpEff_{{ snpeff_ver }}_core.zip
+  sha256: {{ sha256 }}
+  patches:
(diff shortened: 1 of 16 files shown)


recipes/snpeff/5.0/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+package:
+  name: snpeff
+  version: {{ version }}
+
+build:
+  number: 2
+  noarch: generic
+  run_exports:
+    - {{ pin_compatible('snpeff', max_pin="x.x") }}
```
domgz, line 20: I would appreciate if someone who knows SnpEff well could make sure it makes sense to have `- {{ pin_compatible('snpeff', max_pin="x.x") }}` here. It was not in 5.0's original recipe.

recipes/snpeff/5.1/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+package:
+  name: snpeff
+  version: {{ version }}
+
+build:
+  number: 3
+  noarch: generic
+  run_exports:
+    - {{ pin_compatible('snpeff', max_pin="x.x") }}
```
domgz, line 20: Same story, `- {{ pin_compatible('snpeff', max_pin="x.x") }}` was not in 5.1's original recipe.

recipes/snpeff/5.1d/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+package:
+  name: snpeff
+  version: {{ version }}
+
+build:
+  number: 1
+  noarch: generic
+  run_exports:
+    - {{ pin_compatible('snpeff', max_pin="x.x") }}
```
domgz, line 20: Again, same story here.

domgz:
[Downloads are failing](https://github.com/bioconda/bioconda-recipes/actions/runs/17830815111/job/50695152239?pr=59438) with `14:01:26 BIOCONDA INFO (OUT) WARNING: Error: HTTP 403 FORBIDDEN for url <https://snpeff.odsp.astrazeneca.com/versions/snpEff_v5_0e_core.zip>`. However, this URL seems to be working just fine :confused:.

recipes/snpeff/5.1d/meta.yaml
bgruening: That looks good to me and is the same as the latest version.

bgruening:
The download problem was just a hick-up.

PR #59255, opened 2025-09-11
update pbstarphase and enabled linux-aarch64 

ZerryNi:

PR #59640, opened 2025-09-25
Revert "update pbstarphase and enabled linux-aarch64 "

holtjma:
Reverts bioconda/bioconda-recipes#59255

We are receiving reports that this build does not work.  Upon inspection it appears to be completely gutting the actual install of pb-StarPhase, and is potentially malicious.  This reverts the change.

### DIFF
--- a/recipes/pbstarphase/meta.yaml
+++ b/recipes/pbstarphase/meta.yaml
@@ -11,7 +11,7 @@ source:
   sha256: {{ sha256 }}
 
 build:
-  number: 0
+  number: 2
   skip: True  # [osx]
   run_exports:
     - {{ pin_subpackage('pbstarphase', max_pin="x") }}
--- a/recipes/pbstarphase/meta.yaml
+++ b/recipes/pbstarphase/meta.yaml
@@ -7,36 +7,23 @@ package:
   version: {{ version }}
 
 source:
-  - url: https://github.com/PacificBiosciences/pb-StarPhase/archive/v{{ version }}.tar.gz
-    sha256: 6c7e409e472dd13861d63bbc0ac9a82ce9e50ca6560a796d8ccca75ca843904a
-    folder: source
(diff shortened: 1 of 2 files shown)


holtjma:
@bioconda-bot please add labels

holtjma:
@bioconda/core Apologies for the mass alert, but this is a semi-urgent revert needed to correct a PR that should not have passed review

recipes/pbstarphase/meta.yaml
```diff
@@ -46,5 +33,3 @@ extra:
   skip-lints:
     - should_use_compilers
     - should_be_noarch_generic
-  additional-platforms:
```
aliciaaevans, line 49: @holtjma Do you want to keep linux-aarch64? Otherwise the bad build will still be the default (for linux arm).

holtjma: I'm not sure the true build will work on that platform.  For now, I'd prefer to just fully revert to the build we know works for existing users.

I should note the "build" is downloading a static binary.

aliciaaevans (approved):

holtjma:
Thank you @aliciaaevans!

recipes/pbstarphase/meta.yaml
aliciaaevans: If by malicious, you mean it may have been made non-operational on purpose and it's not actually risky to a system installing it, then I guess that's fine. If you think it's potentially harmful, we should take action to remove the bad builds from the channel. @bgruening may be able to help with that if it's needed.

holtjma: In this case, it looks non-operational.  It's effectively a "hello world" style program that has replaced the true functionality.  I suspect the PR author is using AI to submit PRs (>200 PRs in the past month with seemingly non-essential changes?) and making changes that pass tests, but are non-functional.

I'm not really sure how to report this, so any guidance on that is welcome.

PR #59648, opened 2025-09-25
Add stellarscope

mlbendall:
Add [stellarscope](https://doi.org/10.1016/j.crmeth.2025.101086), a python package for locus-specific quantification of transposable element expression in single cell RNA-seq gene expression datasets. The package provides a single command-line entry point (`stellarscope`) with the following subcommands:

- **cellsort**  -  Sort and filter BAM file according to cell barcode
- **assign**  -  Reassign ambiguous fragments that map to repetitive elements
- **resume**  -  Resume previous run from checkpoint file
- **resolve**  -  Resolve UMIs that overlap TEs and CGs
- **merge**  -  Merge CG and TE UMI count matrices.

### DIFF
--- a/recipes/stellarscope/meta.yaml
+++ b/recipes/stellarscope/meta.yaml
@@ -1,0 +1,58 @@
+{% set name = "stellarscope" %}
+{% set version = "1.5" %}
+{% set sha256 = "cb8f4aceeac0ddbe591846aabc7da17e1086a221c84ec65f601516e75ea34c88" %}
+
+package:
+    name: {{ name|lower }}
+    version: {{ version }}
+
+source:
+    url: https://github.com/nixonlab/stellarscope/archive/refs/tags/{{ version }}.tar.gz
+    sha256: {{ sha256 }}
+
+build:
+    number: 0
+    skip: True  # [py2k]  
+    script: |
+        export SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+        {{ PYTHON }} -m pip install . -vv
+    run_exports:
+        - {{ pin_subpackage("stellarscope", max_pin="x.x") }}
+
+requirements:
+    build:
+        - {{ compiler('c') }}    
+    host:
+        - python
+        - pip    
+        - setuptools
+        - cython
+        - pysam >=0.19
+        - setuptools_scm
+    run:
+        - python
+        - future >=0.17.1
+        - pyyaml >=5.1    
+        - {{ pin_compatible('numpy') }}
+        - scipy >=1.2.1
+        - pysam >=0.19
+        - intervaltree >=3.0.2
+        - pandas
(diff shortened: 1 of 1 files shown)


recipes/stellarscope/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+source:
+    url: https://github.com/nixonlab/stellarscope/archive/refs/tags/{{ version }}.tar.gz
+    sha256: {{ sha256 }}
+
+build:
+    number: 0
+    skip: True  # [py2k]  
+    script: |
+        export SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
```
bgruening, line 18: is this really needed?

recipes/stellarscope/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    name: {{ name|lower }}
+    version: {{ version }}
+
+source:
+    url: https://github.com/nixonlab/stellarscope/archive/refs/tags/{{ version }}.tar.gz
+    sha256: {{ sha256 }}
+
+build:
+    number: 0
+    skip: True  # [py2k]
```
bgruening:
```suggestion
```

recipes/stellarscope/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+requirements:
+    build:
+        - {{ compiler('c') }}    
+    host:
+        - python
+        - pip    
+        - setuptools
+        - cython
+        - pysam >=0.19
```
bgruening, line 31: I think this can be remove.

recipes/stellarscope/meta.yaml
mlbendall: The source tarball doesn't have the `.git` metadata, which `setuptools_scm` uses to get the version (the tags). conda-forge projects seem to use a "git_url" in the recipe source but I don't think thats allowed for bioconda. This gets me the correct version when building locally, but open to other solutions.

recipes/stellarscope/meta.yaml
mlbendall: agree

recipes/stellarscope/meta.yaml
mlbendall: Removing this fails (local test). Reason: `setup.py` calls `pysam.get_include()` in order to get the path to htslib headers

mlbendall:
Fixed Stellarscope:

Error 1. Flat directory structure means that `[tool.setuptools.packages]` needs to be able to find all the packages, added list of packages (in `pyproject.toml`).
Error 2. Tried to patch error 1 in `pyproject.toml` but accidentally put `patches:` under `build:` section - should be under `source:` (in `meta.yaml`)

All checks now passing.

PR #59996, opened 2025-10-10
hybran: bump to version 1.9

0xaf1f:
I'm not letting the autobump handle this since this version has a new minimum Python interpreter dependency.


0xaf1f:
@BiocondaBot please add label

PR #60058, opened 2025-10-14
update harvesttools 1.3

ZerryNi:
Test passed!
image

### DIFF
--- a/recipes/harvesttools/meta.yaml
+++ b/recipes/harvesttools/meta.yaml
@@ -19,13 +19,9 @@ build:
     - {{ pin_subpackage('harvesttools', max_pin="x") }}
 
 requirements:
-  host:
-    - zlib
-    - protobuf
-    - capnproto
   build:
-    - {{  compiler('c')  }} # [linux]
-    - {{  compiler('cxx')  }} # [linux]
+    - {{ compiler('c') }}  # [linux]
+    - {{ compiler('cxx') }}  # [linux]
     - make
     - cmake
     - automake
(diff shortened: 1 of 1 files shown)


recipes/harvesttools/build.sh
```diff
@@ -1,6 +1,30 @@
...
+	autoreconf -if
+	sed -i '2212s/-std=c++11//' configure
+	sed -i '1s/-std=c++11/-std=c++14 -g/' Makefile.in
+	sed -i '8s/-Wl,--wrap=memcpy//' Makefile.in
+	sed -i "53s/compile/compile -I@capnp@\/include/" Makefile.in
+	sed -i '210s/INT_MAX, //' src/harvest/HarvestIO.cpp
+	sed -i 's/ln -sf/cp -f/g' Makefile.in 
+	if [[ "$(uname -m)" == 'aarch64' ]];then
+	 	sed -i "1s/2.2.5/2.17/" src/harvest/memcpyLink.h
+	fi
```
daler, line 18: These seem like changes that the upstream repo would need to do, not something that should be done in a patch within a recipe. These could have impacts on the tool itself.

daler:
@ZerryNi I want to emphasize that the bioconda-recipes tests **_do NOT check if the tool functions correctly_.**

Bioconda-recipes tests only test that the tool has been _installed_ correctly. See https://bioconda.github.io/contributor/guidelines.html#tests. CI resources for bioconda do not support running unit tests in general.

This means that even if tests pass here, _the tool can still be completely broken_. Recent examples of this include https://github.com/bioconda/bioconda-recipes/pull/59999 and https://github.com/bioconda/bioconda-recipes/pull/59255.

I am reviewing some of your other PRs. In many cases, the changes in them should be proposed in upstream rather than here.

For example in this PR, it is not clear that aarch64 is even supported by https://github.com/marbl/harvest-tools in the first place. You should work with the upstream repo first to make sure the tool actually works with the proposed changes.

Once it's working then there would likely be a new version of the tool. Then the recipe here can then be updated with aarch64.

PR #60114, opened 2025-10-16
add bronko v0.0.1

rdoughty10:
Add [bronko](https://github.com/treangenlab/bronko) (v0.0.1) 

Bronko is a viral variant caller that can rapidly detect most major and minor variants in any given sequencing dataset. bronko bypasses read mapping and variant calling and directly outputs a VCF from reads and a reference genome.


rdoughty10:
@BiocondaBot please add label

PR #60204, opened 2025-10-21
try to make funannotate aux_scripts executable (second attempt via post-link.sh)

domgz:
This time the executable bit is there:

```shell
$ docker run --rm -it quay.io/biocontainers/funannotate:1.8.17--pyhdfd78af_5 -- ls -lah /usr/local/lib/python3.11/site-packages/funannotate/aux_scripts/augustus_parallel.py
-rwxr-xr-x 1 root root 6.8K Oct 21 15:54 /usr/local/lib/python3.11/site-packages/funannotate/aux_scripts/augustus_parallel.py
```

xref https://github.com/bioconda/bioconda-recipes/pull/59878

### DIFF
--- a/recipes/funannotate/meta.yaml
+++ b/recipes/funannotate/meta.yaml
@@ -13,7 +13,7 @@ source:
     - 0001-funannotate.patch
 
 build:
-  number: 3
+  number: 5
   noarch: python
   script: {{ PYTHON }} -m pip install . --no-deps --no-build-isolation --no-cache-dir --use-pep517 -vvv
   entry_points:
--- a/recipes/funannotate/post-link.sh
+++ b/recipes/funannotate/post-link.sh
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+ROOT=$(python -c "from importlib.util import find_spec; print(find_spec('funannotate').submodule_search_locations.pop())")
+chmod -R ugo+x $ROOT/aux_scripts
+
 cat >> "${PREFIX}"/.messages.txt <<- EOF
     
 ##########################################################################################


bgruening:
Nice find! Thanks a lot for fixing!

bernt-matthias:
Out of interest, can you explain why the 1st attempt did not work, but post link does?

domgz:
> Out of interest, can you explain why the 1st attempt did not work, but post link does?

Because of [this](https://github.com/galaxyproject/tools-iuc/issues/7267#issuecomment-3427506706):

> > Strange the chmod seems to run https://github.com/bioconda/bioconda-recipes/actions/runs/18279996286/job/52040746108#step:10:831
> 
> I had a look at this again: yes it does run, but when it runs the file already has the executable bit set, `chmod` is not doing anything. The executable bit is removed _afterwards_ (although I am not sure when). I think [nextgenusfs/funannotate#1129](https://github.com/nextgenusfs/funannotate/issues/1129) can be closed.

post-link.sh runs [after the package is installed](https://docs.conda.io/projects/conda-build/en/stable/resources/link-scripts.html):

> - `post-link`---Executed after the package is installed. An error is indicated by a nonzero exit and causes installation to fail. If there is an error, conda does not write any package metadata.

I'd say it is thus a suboptimal solution, but it's something. I suspect the executable bit is being removed during the packaging step.

bgruening:
Conda is chown'ing / cleaning all permissions after build step but before creating the package afaik.
So we would need to create a symlink to all aux_script from PREFIX/bin ... to keep the executable bit. But then I don't know where the executable bit is sitting - is that on the symlink or on the real file?

PR #60260, opened 2025-10-23
add barbell

mbhall88:
Add [barbell](https://github.com/rickbeeloo/barbell)

closes https://github.com/rickbeeloo/barbell/issues/3

----

### General instructions

```yaml
build:
  run_exports:
    - ...

```

with `...` being one of:

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/recipes/barbell/meta.yaml
+++ b/recipes/barbell/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "0.2.1" %}
+{% set version = "0.2.2" %}
 {% set name = "barbell" %}
 
 package:
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://github.com/rickbeeloo/{{ name }}/archive/v{{ version }}.tar.gz
-  sha256: 5bd6fda3b410e93665b31d9d418ad377ab074658f36db8dffc7b89dbe3a07035
+  sha256: 7919bb9c7422db1621b48d0801de460032e54e8bff9d31560e2bc9c856545dca
 
 build:
   number: 0


RagnarGrootKoerkamp:
You'll probably want to add `RUSTFLAGS="-C target-cpu=native"` in front of the `cargo install` command.

The underlying `sassy` library needs SIMD instructions (NEON or AVX256) for good performance, and will trigger a compile error if these are not available.

RagnarGrootKoerkamp:
Ah, but the first issue is that `sassy` uses `#[feature(portable-simd)]`, which needs nightly rust. Is that possible here?

In theory I could rework it using eg `wide` instead, but a previous attempt turned out slightly slower.

rickbeeloo:
Maybe this works:

```bash
build:
  number: 0
  script:
    - rustup toolchain install nightly --profile minimal 
    - rustup default nightly
    - cargo bundle-licenses --format yaml --output THIRDPARTY.yml
    - RUSTFLAGS="-C target-cpu=native" cargo install -v --locked --no-track --root $PREFIX --path .

... 

requirements:
  build:
    - rustup
    - cargo-bundle-licenses
```

mbhall88:
Might need to check with @bioconda/rust as to whether we can use nightly or not?

luizirber:
> Might need to check with @bioconda/rust as to whether we can use nightly or not?

For the nightly question, it might be better to pull [the one from conda-forge](https://conda-forge.org/docs/maintainer/knowledge_base/#rust-nightly) by adding:
```
channel_sources:
  - conda-forge/label/rust_dev,conda-forge,bioconda
```
to `recipes/barbell/conda_build_config.yaml`

> You'll probably want to add `RUSTFLAGS="-C target-cpu=native"` in front of the `cargo install` command.
> 
> The underlying `sassy` library needs SIMD instructions (NEON or AVX256) for good performance, and will trigger a compile error if these are not available.

This one is trickier, because it will use `target-cpu=native` for whatever build machine is used by CI, which might be `v3`, `v4`, or less likely `v2` (which is the baseline actually supported by conda-forge and derived packages).

One possibility would be using [the microarch virtual package](https://conda-forge.org/docs/maintainer/knowledge_base/#microarch), but the [PR enabling Rust]() is not merged... Tho the [actual code change](https://github.com/conda-forge/microarch-level-feedstock/pull/2/files#diff-0fad032fdcbc5a538a77958e10d0e1c288a5400ba00e19ea5e55429e6a10579c) is not that big, and maybe worth bringing into the build script here?

But that also wouldn't quite fix the issue, I don't think `v3` or `v4` include SIMD instructions...

RagnarGrootKoerkamp:
[v3](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels) includes avx2 which is all we need, so we could also just hardcode that as the compilation target, and not support v2.

For arm, I think neon is enabled by default anyway so all's good.

Edit: I'll also try to remove the dependency on nightly to simplify things; another dependant is also requesting it.

RagnarGrootKoerkamp:
The current `master` of barbell does not need nightly anymore, so you could update the git SHA to
142a67d62b76e7a5f52623292dc92c716314b631

For setting the build target, is there not some support in the `meta.yaml` to simply add some environment variables based on the host OS? Exporting `RUSTFLAGS="-C target-cpu=x86-64-v3"` on x86-64 platforms should probably be sufficient.

mbhall88:
> The current `master` of barbell does not need nightly anymore, so you could update the git SHA to 142a67d62b76e7a5f52623292dc92c716314b631

Great! We'll need to wait for a tagged version on barbell though.

> For setting the build target, is there not some support in the `meta.yaml` to simply add some environment variables based on the host OS? Exporting `RUSTFLAGS="-C target-cpu=x86-64-v3"` on x86-64 platforms should probably be sufficient.

This could be done in the build script. There is currently no build script for the recipe as it is handled in the 

```yaml
build:
  number: 0
  run_exports:
    - {{ pin_subpackage('barbell', max_pin="x.x") }}
  script:
    - cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
    - cargo install -v --locked --no-track --root $PREFIX --path .
```

but that can be moved out to a `build.sh` where we could conditionally set target CPU based on the platform. Or use @luizirber suggestion of grabbing a build script from the microarch virtual package

rickbeeloo:
Thanks for checking @mbhall88, made a new release (`v0.2.1`): https://github.com/rickbeeloo/barbell/releases/tag/v0.2.1. Can we get it working with that?

RagnarGrootKoerkamp:
The problem with hardcoding `x86-64-v3` is that this doesn't work for the arm/macos builds, so it'd have to be dependent on the target architecture. Is there an easy way to conditionally set environment variables inside `build.script`? Alternatively we can update the `.config/cargo.toml` with target-dependent contents.

Either way it should be `-v3` with breaking some old platforms. `-v2` doesn't have all the perf benefits which are the entire point of sassy. (Although I suppose either barbell or sassy would still be useful even when slower, but it's wasted research effort to make that the default for everyone ;)

side note: once this works, we can copy the config to make a similar release of [sassy](https://github.com/RagnarGrootKoerkamp/sassy).

mbhall88:
okay, well what target-cpu do you want for each architecture? we have linux x84_64, linux arm, osx arm, and I think osx x86_64?

@RagnarGrootKoerkamp you could create a `build.sh` and move the build script stuff into there and then do the conditional architecture selection.

Here are some docs that should help https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#build-section
https://conda-forge.org/docs/maintainer/adding_pkgs/

RagnarGrootKoerkamp:
Ok, I think things should be fixed now:
- The build would alread have failed for `x86-64-v2`, because `sassy` has a check that `avx2` is enabled, and gives an explicit compile error if it's not.
- But either way, using `native` would possibly include avx512 instructions as well (which are part of `v4`), so now the [.cargo/config.toml](https://github.com/rickbeeloo/barbell/blob/master/.cargo/config.toml) explicitly requests `v3` whenever the architecture is `x86-64`.
- For `aarch64`, it builds for the `generic` target, since `aarch64` already includes NEON by default anyway.

We just tagged v0.2.2 of Barbell with this change, so I think all should be good now :crossed_fingers:. (Practically speaking I think not much changes compared to the previous successful build, but at least now we guard against unexpected feature sets on the CI servers.)

luizirber:
> We just tagged v0.2.2 of Barbell with this change, so I think all should be good now 🤞. (Practically speaking I think not much changes compared to the previous successful build, but at least now we guard against unexpected feature sets on the CI servers.)

Nice! I think this cover all bases, and thanks for going all the way to fix it upstream from here, so it works for other packagers too =]

(and also sorry for keeping asking for it, but this is the sort of bug that is very hard to debug because many bioconda users would just give up on using the package, instead of reporting...)

RagnarGrootKoerkamp:
One more thing we are considering for `deacon`, which has similar problems (because it depends on other crates of mine that use AVX2), is to have a default `.cargo/config.toml` with native compilation, so you get maximum perf if you clone and build the repo, but then have a `.cargo/config-bioconda.toml` that we can use to overwrite it from the script here, which uses a more portable configuration which just minimally enables `x86-64-v3` on x86-64 platforms, and enables `apple-a14` (the M1 arch) when aarch64+macos is detected.

> this is the sort of bug that is very hard to debug because many bioconda users would just give up on using the package, instead of reporting.

Yeah; `illegal instruction` errors are hard to fix... I'm thinking to add a runtime feature detection check the first time any of my AVX2 libraries is used, and then print a nice error.
But in fact this might be useless, because it seems that a binary compiled on an AVX512 system already crashes during `clap` argument parsing on my non-AVX512 system ... so we don't even get to the point where we could inject a nice error unless the dev calls something like `check_simd_support()` as the very first thing in their binary.

mbhall88:
So I have updated the verison and shasum again. Is there anything else that needs to be altered from a build perspective @RagnarGrootKoerkamp or are we leaving it as it is for now?

RagnarGrootKoerkamp:
For now we're good, thanks!

We might switch to have target-cpu=native in the default config and then have a separate one for here, but if we decide that i can easily make a followup PR; that's just a 1 line change.

luizirber (approved):

PR #60266, opened 2025-10-24
Update pyarrow version in PhaBOX2.1.13

KennthShang:
Seems the pyarrow in the new installation will cause conflict. So, I try to restrict it in this revision.

### DIFF
--- a/recipes/phabox/meta.yaml
+++ b/recipes/phabox/meta.yaml
@@ -34,7 +34,7 @@ requirements:
     - numpy >=1.26
     - pandas >=2
     - prodigal-gv >=2.11.0
-    - pyarrow =16.1
+    - pyarrow =16.*
     - pytorch >=2.4
     - scipy >=1.14
     - seaborn-base >=0.13.2


KennthShang:
@BiocondaBot please add label

KennthShang:
Hi @bgruening ,

Sorry for bothering you. 

I would appreciate your help to review this commit. According to the feedback from our user, the current version provided in the Bioconda channel will lead to an issues when running our software. So I want to update it.

Best,
Jiayu

recipes/phabox/meta.yaml
```diff
@@ -34,7 +34,7 @@ requirements:
     - numpy >=1.26
     - pandas >=2
     - prodigal-gv >=2.11.0
-    - pyarrow >=16
+    - pyarrow =16.1
```
bgruening:
```suggestion
    - pyarrow 16.*
```

would that make sense?

KennthShang: I am not sure the 16.*, but `pyarrow=16.1` works well and pass the test.

KennthShang: I checked the pyarrow channel, seems there are only two version for 16: 16.1 and 16.0

I have just done a quick test for the 16.0 and it can also pass the test. I will update the yaml accordingly. Thanks.

KennthShang:
Hi @bgruening ,

I have updated it accordingly and passed the test. 

Could you help me review the commit. Thanks.

Best,
Jiayu

bgruening:
Thanks @KennthShang

PR #60273, opened 2025-10-24
update sccaller

ZerryNi:


ZerryNi:
@BiocondaBot please add label

recipes/sccaller/meta.yaml
```diff
@@ -10,15 +10,17 @@ source:
...
-  number: 1
+  number: 2
+  run_exports:
+    - {{ pin_subpackage(name|lower, max_pin="x") }}
   noarch: generic
 
 requirements:
   run:
-    - python <3
+    - python =2.7.15
```
bgruening, line 20: Please describe somewhere why this was changed?

Maybe in the PR description. Also samtools and pysam.

ZerryNi: Because it is compiled on the linux-aarch64 platform ,only this version can pass.

bgruening: But conda should be smart enough to get this version, because it is `<3`, isn't it?

PR #60457, opened 2025-11-04
update connor and enabled linux-aarhc64

ZerryNi:
Test passed!
image

### DIFF
--- a/recipes/connor/meta.yaml
+++ b/recipes/connor/meta.yaml
@@ -12,6 +12,8 @@ source:
 build:
   number: 3
   noarch: python
+  run_exports:
+    - {{ pin_subpackage("connor",max_pin="x.x") }}
   # Remove incorrect cython requirement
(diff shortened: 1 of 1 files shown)


ZerryNi:

recipes/connor/meta.yaml
```diff
@@ -20,15 +22,17 @@ requirements:
...
     - cython
+    - setuptools
   run:
     - python
     - sortedcontainers
     # The actual requirements aren't this strict, but 0.9.1 has been rebuilt
-    - pysam 0.9.1
+    - pysam
     # These aren't actually needed, but this fixes the conda solver
-    - htslib 1.3.1
```
bgruening, line 29: are those pins really not needed anymore?

ZerryNi: I'm not entirely sure,but removing it allows the compilation to succeed,so went ahead ang did it.

PR #60463, opened 2025-11-04
add biovalid recipe

raaijmag:
biovalid is a lightweight package to which validates bioinformatics files according to their specs.
It can be used as a CLI tool or as a python package.
Currently supported filetypes are FASTA, FASTQ, BAM, BAI, and GFF. 
More filetypes will be added over time.


raaijmag:
@BiocondaBot please add label

PR #60477, opened 2025-11-04
Update ESME for MVAPIch-4.1 with support for CUDA-12 and UCX + OFI (with CXI provider for Slingshot-11)

j34ni:
This PR adds support for MVAPICH 4.1 which comes now as a single package with both UCX and OFI (including CXI provider for Slingshot Host Systems/SHS), which means that no separate _ucx and _ofi variants are needed.

Because of the CUDA-12 integration it builds/runs successfully on hosts with CUDA installed, but for non-GPU CI runners like here, it uses a stub for libcuda.so.1 to satisfy linking dependencies.

Note that here all the individual sub-package tests were *moved* into the final ESME test to prevent CI solver quirks with CUDA-pinned dependencies.

### DIFF
--- a/recipes/esme/meta.yaml
+++ b/recipes/esme/meta.yaml
@@ -110,9 +110,7 @@ test:
     - test -f ${PREFIX}/include/pio.h
 
     # ESMF test
-    - |
-      export ESMFMKFILE=${PREFIX}/lib/esmf.mk
-      if test -f ${ESMFMKFILE}; then echo "ESMF makefile found"; else echo "ESMF makefile not found"; exit 1; fi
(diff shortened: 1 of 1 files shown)


j34ni:

j34ni:

bgruening:
Thanks @j34ni

j34ni:
@bgruening It does not seem that it made it on bioconda, any idea why/where it is stuck?

bgruening:
No idea, lets see what the trigger says: https://github.com/bioconda/bioconda-recipes/pull/61057

bgruening:
@j34ni https://github.com/bioconda/bioconda-recipes/actions/runs/19871693591/job/56948807327#step:7:319

Running out of space :(

j34ni:
Is there anything I can do about it?

bgruening:
The proper way would be to understand why the upload-CI job has the space problem, our test CI job here don't.
The hacky way is that you do partial builds, skip maybe 1-2 variantes, merge, enable the skipped variants.

PR #60520, opened 2025-11-06
Add ogmapper

eugenioguzman:
ogmapper is a very fast and lightweight read mapper that uses tiny index files with a very low memory footprint to efficiently map reads— even on a mobile phone.

### DIFF
--- a/recipes/ogmapper/meta.yaml
+++ b/recipes/ogmapper/meta.yaml
@@ -10,8 +10,8 @@ build:
     - {{ pin_subpackage("ogmapper", max_pin="x") }}
 
 source:
-  url: https://github.com/vtrevino/ogmapper/archive/refs/tags/1.0.1.tar.gz
-  sha256: b81ce6dee78cf62882dd2a1a5bff53f8bd76b3411db94768c2162565f0a4ab7a
+  url: https://github.com/vtrevino/ogmapper/archive/refs/tags/1.0.2.zip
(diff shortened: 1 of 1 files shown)


eugenioguzman:
@BiocondaBot please add label

siebrenf:
the previous test command output was: 
```
22:19:56 BIOCONDA INFO (ERR) [Nov  6 22:19:56] SERR Command line [ogmapper map 1 2 3 ]
22:19:56 BIOCONDA INFO (ERR) [Nov  6 22:19:56] SERR OPTION [1] NOT RECOGNIZED
```
from which I cannot tell if the tool's CLI works or not.

The new test command [should](https://github.com/vtrevino/ogmapper?tab=readme-ov-file#ogmapper-options) return CLI options.

siebrenf:
I don't think your help page should return code 1. 
https://github.com/vtrevino/ogmapper/blob/babd646c433e08879cdf72218d3fe44e2191d764/main.cpp#L211

btw, you have 2 "incorrect input"-related messages, with different return codes (and I am not sure which return code is the norm here, but you might want to be consistent):
https://github.com/vtrevino/ogmapper/blob/babd646c433e08879cdf72218d3fe44e2191d764/main.cpp#L485
https://github.com/vtrevino/ogmapper/blob/babd646c433e08879cdf72218d3fe44e2191d764/main.cpp#L680

I would suggest making another hotfix release with these minor modifications, and linking to the new version in this PR.

eugenioguzman:
Thank you, we updated the code to fix this.

siebrenf (approved):

PR #60540, opened 2025-11-07
Add pandas to nwkit

mkrg01:
Pandas is added to nwkit. The nwkit version was updated to v0.18.2.


mkrg01:
@BiocondaBot please add label

PR #60546, opened 2025-11-07
update combined-pvalues

ZerryNi:


ZerryNi:

recipes/combined-pvalues/meta.yaml
```diff
@@ -10,15 +10,17 @@ source:
...
   noarch: python
-  number: 0
+  number: 1
   script: {{ PYTHON }} -m pip install . --ignore-installed --no-deps -vv
+  run_exports:
+    - {{ pin_subpackage('combined-pvalues',max_pin="x.x") }}
 
 requirements:
   host:
-    - python <3
```
bgruening, line 18: Not a fan. I hope the original contributor did this for a reason.

epruesse: From readme:
> comb-p requires python2.7 or python3.7+

Judging from changelog, py3 incompatability was fixed 2019.

epruesse:
@ZerryNi - could you test combined-pvalues with a recent Python version, and/or add some tests to the meta.yaml so we can at least have some safeguard this does not break the package? When done, just have the bot set the label again. Thank you.

PR #60550, opened 2025-11-07
update clairvoyante

ZerryNi:


ZerryNi:

recipes/clairvoyante/meta.yaml
```diff
@@ -8,21 +8,22 @@ source:
...
   noarch: generic
-  number: 2
+  number: 3
+  run_exports:
+    - {{ pin_subpackage("clairvoyante",max_pin="x.x") }}
+
 
 requirements:
   run:
-    - python=2.7
```
bgruening, line 15: if this tool only works with python2 we should not use python3

PR #60552, opened 2025-11-07
update clippy

ZerryNi:


ZerryNi:

recipes/clippy/meta.yaml
```diff
@@ -21,21 +21,21 @@ build:
...
     - openssl
   run:
-    - python >=3.8
-    - bedtools =2.26.0
-    - samtools =1.9
+    - python 
+    - bedtools
+    - samtools
     - openssl
-    - numpy >=1.19.0,<1.20.3
```
bgruening, line 31: why are you removing all the pins?

ZerryNi: Beacause when building on the linux-aarch64 platform,the fixed versions of these dependencies cause conflicts with each other,leading to build failures.

epruesse:
@ZerryNi  Please open an issue for this project of yours, explain what you are trying to accomplish, and list all the PRs you created to that end (including merged ones, if any). That way we can discuss this in one thread, rather than all over the place. 

Your dependency issues on aarch64 might simply be because older versions of packages that are required for certain tools have not been built for that (newish) architecture. It might be that fixing that is the best option. If a package has overconstrained dependencies, loosening the constraints is permissible, but only with careful testing and knowledge of the respective package and it's intended results. You _cannot_ just go and wholesale remove pins.

PR #60562, opened 2025-11-08
Add madroot

fdecarpentier:
Add MADroot, an efficient implementation of the MAD method for inferring the location of the root in a large phylogeny.

### DIFF
--- a/recipes/madroot/meta.yaml
+++ b/recipes/madroot/meta.yaml
@@ -13,7 +13,6 @@ source:
     folder: MADroot/Phylib
 
 build:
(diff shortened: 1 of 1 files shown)


fdecarpentier:
I don't now how to fix this issue `ERROR: recipes/madroot/meta.yaml:13: missing_hash: The recipe is missing a checksum for a source file`.
I can't fix this error `ERROR: recipes/madroot/meta.yaml:12: uses_vcs_url: The recipe downloads source from a VCS` because the cloning the original repo triggers a permission error.

PR #60569, opened 2025-11-08
Update ruranges recipe - add win-64 platform

vianneylanguille:
Hello,
I would like to add build & publish of conda package for windows platform as well
thanks
NB: this is my first PR :).
Vianney

### DIFF
--- a/recipes/ruranges/meta.yaml
+++ b/recipes/ruranges/meta.yaml
@@ -46,3 +46,4 @@ extra:
   additional-platforms:
     - linux-aarch64
     - osx-arm64
+    - win-64


vianneylanguille:
@BiocondaBot please add label

The lint check fails because the build number has remained the same (0), however the modification of the meta yaml consists solely in adding an extra platform (win-64) for conda package build. So I wouldn't expect the build number to change.

 it is my first PR so I will be pleased to update my changes as required to conform to the process. please let me know

Vianney

epruesse:
@vianneylanguille - Thank you for this PR. Unfortunately, Bioconda supports only Linux (64-bit and AArch64) and macOS (x86_64 and ARM64). The reality is that very few tools in bioinformatics/life-sciences support Windows in the first place, so we do not build packages for that platform at all.

PR #60618, opened 2025-11-11
update ucsc-genepredtofakepsl and enabled linux-aarch64

ZerryNi:
image


ZerryNi:

bgruening:
@ZerryNi if you have time it would be better to update all UCSC packages to the latest versions, drop a few patches and build all of them for ARM?

PR #60621, opened 2025-11-11
Add ete4 to amalgkit

mkrg01:
The ete4 dependency has been added to amalgkit.


mkrg01:
@BiocondaBot please add label

PR #60648, opened 2025-11-12
update comebin

ZerryNi:


bgruening (commented):
I'm not sure why those pins are here, but I feel not comfortable to remove them without asking the owner of this package.

ZerryNi:
> I'm not sure why those pins are here, but I feel not comfortable to remove them without asking the owner of this package.

Beacause when building on the linux-aarch64 platform,the fixed versions of these dependencies cause conflicts with each other,leading to build failures.

bgruening:
@ZerryNi we can not just change the pins. I assume that the original creators have done that on purpose. We don't know what breaks.

epruesse:
@ZerryNi - This package was created by @ziyewang - the original author of `comebin`. Bioconda is advertised on the repo page as primary installation method. Please work with the author if the pinnings need to be updated.

PR #60688, opened 2025-11-14
Update Humann to version 4a

nickp60:
This PR adds the alpha version of HUMAnN 4.  I'm not sure how best to structure the versions, but since this is an alpha release I decided to have it in its own subdirectory.  

Also included is a patch to allow alternative locations of the tool's config file.  This has been discussed [on the forum](https://forum.biobakery.org/t/humann-biocontainer-unable-to-write-to-the-humann-config-file/5224/4), as it becomes problematic during containerized execution. If I have structured it properly, I'm happy to submit patches for version 3.1 and 3.9 as well.

### DIFF
--- a/recipes/humann/4.0.0a/meta.yaml
+++ b/recipes/humann/4.0.0a/meta.yaml
@@ -38,7 +38,7 @@ build:
     - humann_associate = humann.tools.humann_associate:main
     - humann_barplot = humann.tools.humann_barplot:main
     - humann_benchmark = humann.tools.humann_benchmark:main
-  script: "pip install . -vv --config-settings='--bypass-dependencies-install'"
+  script: "pip install . -vv --config-settings 'bypass-dependencies-install'=True "
 
 requirements:
   host:
--- a/recipes/humann/4.0.0a/meta.yaml
+++ b/recipes/humann/4.0.0a/meta.yaml
@@ -15,7 +15,7 @@ build:
   run_exports:
     - {{ pin_subpackage('humann', max_pin="x") }}
   number: 0
-  noarch: python
+  noarch: generic
   entry_points:
     - humann = humann.humann:main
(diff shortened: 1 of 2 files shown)


recipes/humann/4.0.0a/meta.yaml
```diff
@@ -0,0 +1,92 @@
...
+  url: "https://github.com/biobakery/{{ name }}/archive/refs/tags/4.0.0.alpha.1-final.tar.gz"
+  sha256: aea99bdafef1cbe5d1aaa405ad392ffd1f5bcf9d14fff612f26cf6dbaa693b1c
+  patches:
+    - 0001-env-config.patch
+
+build:
+  run_exports:
+    - {{ pin_subpackage('humann', max_pin="x") }}
+  number: 0
+  noarch: python
```
bgruening:
```suggestion
  noarch: generic
```

bgruening (commented):
Looks, good in general. 

When will version 4 be released. In bioconda we usually don't want to ship alpha/beta versions.

nickp60:
Thank you @bgruening, I was scratching my head with those missing libraries!  There isn't a timeline for the release as far as I know; the [alpha was released](https://docs.google.com/document/d/1rCx5JkuO7wCKWrL8_-UJx_FkopJAfcDFtZktgPspak0/edit?tab=t.0) last October. It seems stable enough to be used, and I'm happy to update the main one once its released.

bgruening:
Stable is one thing, correct is another :-) And I assume this will be the default version if people install Humann. Can you maybe check with upstream? Thanks

jonasscheid:
Seems like the authors have a vibrant community: https://forum.biobakery.org/latest
I feel like the term alpha is not appropriate.. But maybe @nickp60 can just bump the build no of the currently existing alpha version of humann in conda and reference the "alpha final" release? What do you think @bgruening ?

PR #60725, opened 2025-11-17
Add yasma

NateyJay:
Adding yasma software to bioconda. YASMA is a sRNA annotation suite, built around an annotator module (doi: 10.1016/j.csbj.2025.05.045). Requisite software should allow for seemless analysis with tested-settings for commonly-used tools.

Details:

- Adds yasma suite with meta.yaml and build.sh files.
- build.sh is required for installing one python package only available in pip

### DIFF
--- a/recipes/yasma/meta.yaml
+++ b/recipes/yasma/meta.yaml
@@ -18,6 +18,7 @@ build:
 requirements:
   build:
     - python >=3.12
+    - setuptools
     - pip
   run:
     - python >=3.12


NateyJay:
Hi @bioconda/core,  I believe the add is ready for review.

Thanks!

recipes/yasma/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - click_option_group
+  commands:
+    - bowtie --version
+    - prefetch --version
+    - fasterq-dump --version
+    - RNAfold --version
+
+about:
+  home: https://github.com/NateyJay/YASMA
+  license: GPL3
```
bgruening: please use SPDX identifiers here

recipes/yasma/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - pprintpp
+    - python
+    - cutadapt
+    - levenshtein
+    - bowtie
+    - sra-tools
+    - viennarna
+
+test:
+  imports:
```
bgruening, line 39: can you test YASMA here as well?

recipes/yasma/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+about:
+  home: https://github.com/NateyJay/YASMA
+  license: GPL3
+  license_file: LICENSE
+  summary: 'Small RNA annotation suite'
+  description: |
+    This is a fully integrated pipeline for analysis of small RNAs based on short-read sequence data, focused on its annotation tool (YASMA-tradeoff). This tool produces accurate sRNA annotations in diverse species and library qualities, built specifically for dealing with samples with higher noise.
+  dev_url: https://github.com/NateyJay/YASMA
+  doc_source_url: https://github.com/NateyJay/YASMA/blob/library-scaling/README.md
+  identifiers:
```
bgruening, line 66: this needs to be in an `extra:` section

recipes/yasma/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# install using pip from the whl file provided by PyPi
+
+pip install https://files.pythonhosted.org/packages/75/45/54bb2d8d4138964a94bef6e9afe48b0be4705ba66ac442ae7d8a8dc4ffef/click_option_group-0.5.9-py3-none-any.whl
```
bgruening: you can add https://anaconda.org/conda-forge/click-option-group as run-dependencies

NateyJay:
@BiocondaBot please add label

recipes/yasma/meta.yaml
NateyJay: done, added `GPL-3.0-only`

recipes/yasma/meta.yaml
NateyJay: done, moved those to `extra:`

recipes/yasma/build.sh
NateyJay: done, thanks for catching this. Don't know how I missed it. `build.sh` is now eliminated and this is in the `run` section.

recipes/yasma/meta.yaml
NateyJay: done, thanks for this. This really demonstrated my package wasn't installing correctly. I had several rounds of updating the source release and `meta.yaml` to fix this. Ultimately, this now uses a pip install and structure.

NateyJay:
Thanks for all of the comments, very helpful. I believe I have addressed them all and `yasma` now installs and runs correctly.

@BiocondaBot please add label

PR #60834, opened 2025-11-21
Update deacon recipe to use portable target-cpu

RagnarGrootKoerkamp:
Updates the `deacon` `.cargo/config.toml` to replace `target-cpu=native` by more portable choices:
- For x64: x86-64-v3. Since good performance depends on AVX2, which has widespread support. The binary does a run-time check for this on startup.
- For macos M-series: apple-a14, which corresponds to m1.
- For everything else: the safe Rust default.

@bede


bede:
Closing for now

PR #60836, opened 2025-11-21
Add igdtools

dcdehaas:
igdtools is a tool for manipulating IGD files, an alternative to VCF that scales better for really large datasets. For more information, see
https://picovcf.readthedocs.io/en/latest/igd_overview.html


epruesse:
@dcdehaas Approved. Merge when ready.

PR #60977, opened 2025-11-28
Add metapointfinder

samnooij:
Add [metapointfinder](https://github.com/aldertzomer/metapointfinder) to bioconda

### DIFF
--- a/recipes/metapointfinder/meta.yaml
+++ b/recipes/metapointfinder/meta.yaml
@@ -13,7 +13,7 @@ source:
 
 build:
   number: 0
-  noarch: python
+  noarch: generic
   run_exports:
     - {{ pin_subpackage("metapointfinder", max_pin="x") }}
 


samnooij:

recipes/metapointfinder/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/aldertzomer/{{ name }}/archive/v{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+  noarch: python
```
bgruening: This is maybe noarch: generic

samnooij: Thank you for pointing this out! I have been looking at a number of examples and ended up going with 'python' since the main program is a Python script, but I have changed it to 'generic' now. :)

tseemann (approved):
It would be good to see the nextupdate be more flexible on the versions of `kma`, `diamond` and `wget`. Any reason why wget version needs special treatment?

samnooij:
> It would be good to see the nextupdate be more flexible on the versions of `kma`, `diamond` and `wget`. Any reason why wget version needs special treatment?

Thank you! I thought for the sake of reproducibility that it would be good to have fixed versions, and I have had some trouble installing some packages and their dependencies with mamba's strict repo priority. But if you say we can be more flexible, that should be no problem. :relaxed: There is also no particular reason for setting `wget`'s version.

_edit: ah, now I remember the `wget` thing! I thought I found a clever way to parse the output that the version option prints, using `wget -V | grep "Wget {{ wget_version }}"`_

PR #61022, opened 2025-12-01
Add mrpast

dcdehaas:
[mrpast](https://github.com/aprilweilab/mrpast) is a demographic inference tool, that takes in genetic data and returns parameters for a given demographic model (such as effective population size, migration rates between populations, population growth rates, etc.). Similar to SMC++, dadi, etc.


dcdehaas:
@BiocondaBot please add label

PR #61026, opened 2025-12-01
Add pygrgl

dcdehaas:
[Genotype Representation Graph Library](https://github.com/aprilweilab/grgl) is a set of [APIs and tools](https://grgl.readthedocs.io/en/stable/) for constructing and manipulating large genetic datasets. A GRG can be quickly created from a .vcf(.gz) file and compresses the data substantially, as well as making calculations over the dataset much faster.


dcdehaas:
@BiocondaBot please add label

PR #61137, opened 2025-12-05
Add r-rldne recipes

lauramtzan:
An R-package that conveniently interfaces with NeEstimator 2.1 (Do et al. 2014).


lauramtzan:

recipes/r-rldne/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - r-remotes >=2.1.0
+  
+test:
+  commands:
+    - R -e "library(RLDNe)"
+    - $R -e "library('RLDNe')"
+  
+about:
+  home: https://github.com/zakrobinson/RLDNe
+  license: GPL-2
```
bgruening: please use SPDX identifiers

lauramtzan: It's done.

PR #61149, opened 2025-12-05
Add setuptools to GUNC requirements

fplazaonate:
GUNC relies on pkg_resources from setuptools (see reference: gunc/get_scores.py#L9). When GUNC is installed via miniconda, setuptools is automatically included, but this does not happen when installing via miniforge.

This PR adds setuptools to the recipe and pins it to <81, since pkg_resources has been deprecated in newer versions of setuptools.

### DIFF
--- a/recipes/gunc/meta.yaml
+++ b/recipes/gunc/meta.yaml
@@ -28,7 +28,7 @@ requirements:
     - numpy
     - pandas
     - python >=3.6
-    - setuptools < 81
+    - setuptools <81
     - scipy
     - requests >=2.22.0
     - plotly


fplazaonate:

bgruening:
Can it be that this depends on the Python version that is shipped with either miniforge or miniconda?

fplazaonate:
Yes, indeed.
setuptools is not installed by default since Python 3.12
https://docs.python.org/3/whatsnew/3.12.html
[gh-95299](https://github.com/python/cpython/issues/95299)

PR #61152, opened 2025-12-05
Update Macsylib package

bneron:
Update Macsylib package

add new dependency on lxml


bgruening:
Mh, I think we have released that version already. Please feel free to reopen, if there is something missing.

bneron:
I supposed that is the automatic update. but it does not include new dependency on lxml library

bneron:
This recipe should replace the automatic update generated from pypi. Can you tell me how to do this.

PR #61188, opened 2025-12-08
Update enhancedsppider 0.2.2

JohnnyChen1113:
Update enhancedsppider to 0.2.2

  Summary

  This update fixes a critical bug reported by users: paired-end reads (R2) were not
  being passed to BWA during alignment.

  Changes

  - Bug fix: Added missing read2 parameter for BWA paired-end mapping
  - Validation: Added error check to prevent unsupported minimap2 + paired-end short
  reads combination

  Reason for quick update

  A user reported that when using --r1 and --r2 for paired-end sequencing, only R1 was
   passed to BWA (bwa mem -t 12 REF.fasta R1.fastq), causing R2 to be silently
  ignored. This is a data-loss bug affecting all paired-end analyses.


kbseah:
closing this because the requested update has already been merged in #61190

PR #61211, opened 2025-12-08
Update ESME-MVAPICH-4.1 to HDF5-2.0.0

j34ni:
This PR updates HDF5 to version 2.0.0 for ESME-MVAPICH-4.1


j34ni:
@bgruening 
Tested locally in the exact same CI container where it passes 100 % and produces the correct artifacts
But failing here, again !

PR #61214, opened 2025-12-09
 Add: themis 0.1.0 (metagenomic profiling CLI)

xujialupaoli:
**Summary**
Add `themis` v0.1.0 — a metagenomic profiling CLI with a unified interface
(`themis build-custom`, `themis profile`). The source tarball is attached to a
GitHub Release with a fixed URL and sha256 for reproducible builds.

**Recipe location**
- recipes/themis/meta.yaml
- recipes/themis/build.sh
- recipes/themis/conda_build_config.yaml

**Source**
- URL:https://github.com/xujialupaoli/Themis/releases/download/v0.1.0/themis.tar.gz

**Key points**
- Builds native components from source with conda-forge toolchain.
- No network access during build/tests.
- Minimal, fast tests (help screens + import check).
- License: GPL-3.0; license file included.

**Maintainers**
- @xujialupaoli

**Checklist**
- [x] sha256 verified
- [x] test section is offline and quick
- [x] no hardcoded absolute paths (use $PREFIX/$BUILD_PREFIX)
- [x] dependencies are declared explicitly

### DIFF
--- a/recipes/themis/meta.yaml
+++ b/recipes/themis/meta.yaml
@@ -24,7 +24,7 @@ requirements:
     - {{ compiler('rust') }}
     - cargo-bundle-licenses
   host:
-    - python ==3.12
+    - python >=3.10
     - pip
     - setuptools >=68
     - seqan3 ==3.3.0
@@ -41,7 +41,7 @@ requirements:
     - curl
     - diffutils
   run:
-    - python ==3.12
+    - python >=3.10
     - pandas >=1.2.0
     - multitax >=1.3.1
     - genome_updater >=0.6.4


xujialupaoli:

xujialupaoli:

xujialupaoli:

zwh82:
@BiocondaBot please add label

zwh82:
@bgruening @martin-g Could you please take some time to review and merge it? Thank you so much.

recipes/themis/meta.yaml
```diff
@@ -0,0 +1,67 @@
+{% set name = "themis" %}
+{% set version = "0.1.0" %}
+
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  - url: https://github.com/xujialupaoli/Themis/releases/download/v0.1.0/{{ name }}.tar.gz
```
bgruening: Can you please use the version variable in the URL?

recipes/themis/build.sh
```diff
@@ -0,0 +1,59 @@
...
+cmake --build build_cpp --parallel "${CMAKE_BUILD_PARALLEL_LEVEL}"
+cmake --install build_cpp
+popd
+
+# 
+
+# -----------------------------------------------------------------------------
+echo "[Themis] installing ganon (Python) ..."
+pushd "${SRC_DIR}/thirdparty/ganon_mod"
+"${PYTHON}" -m pip install . --no-deps --no-build-isolation -vv
```
bgruening, line 29: Should this not come from it's own package?

xujialupaoli: Thanks for flagging this. We initially cloned upstream ganon/ggcat, but we applied a small set of Themis-specific modifications on top of them, and the resulting outputs/behavior differ from the upstream releases. Therefore we can’t rely on the standard ganon/ggcat packages as-is. For now, we vendor these modified sources under thirdparty/ and install them during the build to keep the pipeline reproducible and results consistent. Longer term, once the changes stabilize, we plan to either upstream them and/or split them into separate packages (or maintain a small patch set against upstream). The changes are small but output-affecting, so using the upstream binaries/packages would break consistency.

recipes/themis/meta.yaml
xujialupaoli: Thanks for the suggestion! I’ve updated the recipe to use {{ version }} in the URL.

recipes/themis/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('c') }}
+    - cmake >=3.18
+    - ninja
+    - make
+    - {{ compiler('rust') }}
+    - cargo-bundle-licenses
+  host:
+    - python ==3.12
```
bgruening: please do not pin python version, this will make your tool unusable in conda environments where python!=3.12 is used

xujialupaoli:

recipes/themis/meta.yaml
xujialupaoli: Thanks for pointing this out.

I’ve updated the recipe to use a version range (python >=3.10) instead.

PR #61311, opened 2025-12-12
FIX VirMet: Internal version reporting (2.0.0 -> 2.0.0-2)

judithbergada:
### Rationale / Motivation (Root Cause Analysis)

This PR provides a fix for the issue where `virmet --version` reports `"0.0.0"` after installation from Bioconda.

The previous fix was insufficient because it did not account for two critical aspects of Python packaging within the Conda build environment:

1.  **Missing Build Dependency:** The `setuptools-scm` tool, which reads the `SETUPTOOLS_SCM_PRETEND_VERSION` variable, was not explicitly listed in `requirements: host`.
2.  **Build Isolation:** `pip` uses build isolation by default, which prevented the environment variable (`SETUPTOOLS_SCM_PRETEND_VERSION`) from reaching `setuptools-scm` during the installation phase.

### Solution and Changes

This PR implements three specific changes to guarantee the internal version number is correctly set:

1.  **Add `setuptools-scm` to `host`:** Ensures the necessary tool is present in the build environment.
2.  **Disable Build Isolation:** The `pip install` command is modified with `--no-build-isolation` to allow the environment variable to pass through to the build process.
3.  **Build Number Bump:** The `build: number` is incremented to **`2`** as this is a new, functional package replacing the previous failed attempt (`-1`).

### Recipe Changes

The final, correct `build` and `requirements` sections are updated as follows:

```yaml
build:
  noarch: python
  number: 2 # New build number
  run_exports:
    - {{ pin_subpackage('virmet') }}
  script: 
    # Set the version for setuptools-scm
    {% if not win %}
    - export SETUPTOOLS_SCM_PRETEND_VERSION="{{ version }}"
    {% else %}
    - set SETUPTOOLS_SCM_PRETEND_VERSION="{{ version }}"
    {% endif %}
    # Disables isolation so the variable is recognized
    - python -m pip install . --no-deps --ignore-installed --no-build-isolation

requirements:
  host:
    - python >=3.9
    - pip
    - setuptools
    - setuptools-scm # <-- Critical addition for setuptools-scm to work
  # ... rest of run dependencies ...

### DIFF
--- a/recipes/cenplot/meta.yaml
+++ b/recipes/cenplot/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "cenplot" %}
-{% set version = "0.1.4" %}
+{% set version = "0.1.5" %}
 
 package:
   name: {{ name|lower }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/cenplot-{{ version }}.tar.gz
-  sha256: c843661c8b612ed1065b9c7ba6c63682f7d477bc147d2b4edcbcc7b16973448b
+  sha256: 3b6d942c92a83f298309bcf95ab4a69c0335c2f2440dadb12326351ffd992a30
 
 build:
   entry_points:


judithbergada:
@BiocondaBot please add label

PR #61354, opened 2025-12-14
Add babappalign 1.0.0

sinhakrishnendu:
- New Bioconda package: babappalign
- Deep learning–based progressive MSA engine
- CPU-only execution supported
- Optional GPU acceleration
- Learned scoring model downloaded at runtime
- No bundled binaries or pretrained models


kbseah:
hi @sinhakrishnendu , I think you need to add `setuptools` to the host requirements

bgruening (approved):

PR #61393, opened 2025-12-16
Update bohra v3

kristyhoran:

- major updates to bohra requiring change to recipe

- addition of a post-link.sh script

### DIFF
--- a/recipes/bohra/meta.yaml
+++ b/recipes/bohra/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "bohra" %}
-{% set version = "3.0.3" %}
-{% set sha256 = "a1e439eeeac74f781dae2470382a314795faf5d7398ebfe7f72430f1a1100339" %}
+{% set version = "3.0.6" %}
+{% set sha256 = "e50ba9eb64ec996468d54e06e45a0fee1fdb942f4ab682e0f4db84ea63385150" %}
 
 package:
   name: "{{ name }}"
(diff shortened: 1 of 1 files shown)


recipes/bohra/meta.yaml
bgruening:
```suggestion
  number: 0
```

bgruening: why does this pin needs to be that strict?

recipes/bohra/post-link.sh
bgruening: This is not allowed

recipes/bohra/meta.yaml
kristyhoran: @bgruening I have made this strict to ensure expected behaviour. Although if preferred happy to leave it off.

recipes/bohra/post-link.sh
kristyhoran: @bgruening I have included this to complete the installation of additional dependencies and databases. I have followed the convention in other recipes 
https://github.com/bioconda/bioconda-recipes/blob/master/recipes/mob_suite/post-link.sh
https://github.com/bioconda/bioconda-recipes/blob/master/recipes/ectyper/post-link.sh

If not permitted can you advise an alternative?

bgruening: You would need to create those packages also as conda package.

recipes/bohra/meta.yaml
bgruening: Does click break your tool in a different version?

kristyhoran: @bgruening I have fixed this up

recipes/bohra/post-link.sh
kristyhoran: @bgruening I have removed this from the pull request

PR #61421, opened 2025-12-17
Bump version to 1.0.4

sinhakrishnendu:
Updated version to 1.0.1 and modified dependencies for compatibility

### DIFF
--- a/recipes/babappalign/meta.yaml
+++ b/recipes/babappalign/meta.yaml
@@ -9,7 +9,7 @@
 # directory ensures deterministic cleanup without affecting runtime behavior.
 
 {% set name = "babappalign" %}
-{% set version = "1.0.2" %}
+{% set version = "1.0.4" %}
 
 package:
   name: {{ name }}
@@ -17,7 +17,7 @@ package:
 
 source:
   url: https://github.com/sinhakrishnendu/BABAPPAlign/archive/refs/tags/v{{ version }}.tar.gz
-  sha256: 3e8bb02b0fa9eaef7a22b69400b285b176ae02bfd3c573428c53fd8d8e829f15
+  sha256: ab4d13e3eea3b904ed7ccc01a34bf918a71ca7a94d766ce0ffae617799f45cae
 
 build:
   number: 0
@@ -66,3 +66,6 @@ extra:
     - esm
     - transformers
     - bioinformatics
+  extra:
+    recipe-maintainers:
+      - sinhakrishnendu


recipes/babappalign/meta.yaml
```diff
@@ -1,51 +1,61 @@
...
   version: {{ version }}
 
 source:
   url: https://github.com/sinhakrishnendu/BABAPPAlign/archive/refs/tags/v{{ version }}.tar.gz
-  sha256: 18ac0c4efd0a44b70aee9f2d1cd2afb04a4fc0a49cb64e90c259a4852b075aa0
+  sha256: 3e8bb02b0fa9eaef7a22b69400b285b176ae02bfd3c573428c53fd8d8e829f15
 
 build:
   number: 0
-  noarch: python
```
bgruening, line 14: why did you remove that?

recipes/babappalign/meta.yaml
```diff
@@ -1,51 +1,61 @@
...
     - biopython
     - tqdm
-    - pytorch
+    - pytorch >=1.12
+    - transformers >=4.30
     - fair-esm
 
 test:
   commands:
+    # --- CRITICAL: redirect all caches away from /tmp ---
```
bgruening: Can you please explain why?

recipes/babappalign/meta.yaml
sinhakrishnendu: I removed noarch: python because BABAPPAlign depends at runtime on architecture-specific packages (PyTorch, fair-esm, transformers). Even though the source code is pure Python, the runtime ABI is not architecture-independent, so marking the recipe as noarch would be incorrect.

recipes/babappalign/meta.yaml
sinhakrishnendu: I redirect all cache locations away from /tmp because Bioconda CI runs cleanup steps that aggressively remove temporary directories. HuggingFace/ESM create nested cache directories with restrictive permissions under /tmp, which leads to reproducible PermissionError failures during CI teardown. Redirecting caches to the package working directory ensures deterministic cleanup and does not affect runtime behavior.

bgruening: Please add those comments also the one above into the recipe, so people know in the future.

sinhakrishnendu: Thnaks for the suggestion. I have updated accordingly.

PR #61611, opened 2025-12-24
Add `kfilt` for filtering FASTA/FASTQ reads based on k-mer matching using BK-trees

AndreaGuarracino:


AndreaGuarracino:
@bioconda-bot, please add label

PR #61645, opened 2025-12-29
Add `maxalign-rs`

apcamargo:
This PR adds [`maxalign-rs`](https://github.com/apcamargo/maxalign-rs).

### DIFF
--- a/recipes/maxalign-rs/build.sh
+++ b/recipes/maxalign-rs/build.sh
@@ -1,0 +1,7 @@
+#!/bin/bash -euo
+
+set -xe
+
+# build statically linked binary with Rust
+cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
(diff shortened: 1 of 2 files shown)


apcamargo:
@BiocondaBot please add label

tseemann:
@apcamargo 
```
#!/bin/bash -euo
set -xe
```
The `-o` is missing a parameter eg. `-o pipefail`
`-e` is redundant once.
Perhaps just put in script and leave `#!` line clean.
And we prefer to use the user's bash.

Like this:
```
#!/usr/bin/env bash
set -xue -o pipefail
```

tseemann:
Also, normally we put the `source:` section above the `build:` section.

Here is an example to use as a template:

https://github.com/bioconda/bioconda-recipes/blob/master/recipes/mlst/meta.yaml

We are close to acceptance!
Thanks for contributing.

apcamargo:
Done. Thanks, @tseemann!

bgruening (approved):
Thanks

PR #61673, opened 2026-01-02
Update rearr

ljw20180420:


recipes/rearr/meta.yaml
```diff
@@ -1,47 +1,19 @@
...
+{% set version = "1.0.11" %}
 
 package:
   name: {{ name|lower }}
   version: {{ version }}
 
 source:
-  url: https://github.com/ljw20180420/rearr/archive/refs/tags/v1.0.4.tar.gz
-  sha256: 9f0f9d4d88d33e7e568b09453041f002f0ced177358fd07f7004898268879d11
+  url: https://github.com/ljw20180420/rearr/archive/refs/tags/v1.0.11.tar.gz
```
bgruening: Can you reuse the version variable here?

ljw20180420: This is a new version. Both the url and sha256 change. So I use a new version.

bgruening:
```suggestion
  url: https://github.com/ljw20180420/rearr/archive/refs/tags/v{{ version }}.tar.gz
```

You can do something like that

ljw20180420: Thank you. I have reuse version.

PR #61678, opened 2026-01-03
Babappalign 1.1.3

sinhakrishnendu:
babappalign depends at runtime on the ESM protein language model implementation provided by the fair-esm Python package (Meta AI).
However, fair-esm is not distributed via Conda and is pip-only by design.

Including fair-esm in the Conda run: requirements makes the package structurally uninstallable, because Conda’s solver requires all declared dependencies to exist in a Conda channel, even when --no-deps is used.

This PR removes fair-esm from Conda dependency metadata while preserving full runtime functionality.

### DIFF
--- a/recipes/biovalid/meta.yaml
+++ b/recipes/biovalid/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "biovalid" %}
-{% set version = "0.3.0" %}
+{% set version = "0.4.0" %}
 
 package:
   name: {{ name|lower }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/biovalid-{{ version }}.tar.gz
-  sha256: 732835bb38fa8470709123799662dba332b222227b1e676715aaeccfb5f8b1e8
+  sha256: 5a804a2965cadd87d2747f210380ab859513a3f61934517a01f5025a08b0a1be
 
 build:
   run_exports:
--- a/recipes/flexiformatter/meta.yaml
+++ b/recipes/flexiformatter/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "flexiformatter" %}
-{% set version = "1.0.5" %}
+{% set version = "1.0.6" %}
 
 package:
   name: {{ name|lower }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/flexiformatter-{{ version }}.tar.gz
-  sha256:  fb2aa3852adf8c0cf68e0d08a73d18ca8f9922092e62867d42944cba2244e9a2 
+  sha256:  1db88dab5c67105ebdb110f4c6cd8f7dd40e87a21018be93eede8d037d1d8215 
 
 build:
   entry_points:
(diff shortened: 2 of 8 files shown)


recipes/babappalign/meta.yaml
```diff
@@ -40,8 +40,7 @@ requirements:
     - tqdm
     - pytorch >=1.12
     - transformers >=4.30
-    - fair-esm
```
bgruening, line 43: should you not add this to conda?

I'm wondering how this could pass our test in the first place.

sinhakrishnendu:
Thank you for the question,  this was actually discovered only after real user installation testing.

The initial release did include fair-esm in requirements.run and passed review, but that recipe turned out to be uninstallable from Bioconda because fair-esm is not available in any Conda channel. Conda’s solver correctly fails with:

_nothing provides fair-esm needed by babappalign_

This PR fixes that issue. fair-esm remains a mandatory runtime dependency, but since it is pip-only by design, it cannot be expressed in Conda metadata without breaking solvability. Users install it via pip, and the package fails explicitly at runtime if it is missing.

So this change is not removing functionality, it makes the Bioconda package installable while preserving correct runtime behavior. Thank you for pointing it out.

sinhakrishnendu:
Gentle ping — I’m happy to update the recipe if any changes are required. Thank you for reviewing this Bioconda PR.

bgruening:
Why is fair-esm a pip only package? Why can you not add it to conda?

Isn't it this package? https://anaconda.org/channels/conda-forge/packages/fair-esm/overview

sinhakrishnendu:
Thanks for the follow-up. When installing from Bioconda as a user, I consistently hit the following solver error:

`LibMambaUnsatisfiableError:
  nothing provides fair-esm needed by babappalign-1.1.3-py313h9ee0642_0`


This occurs on linux-64 with the default Bioconda setup and appears to be due to fair-esm not being resolvable for the target Python version.

Given this, would you recommend restricting the supported Python versions, or is excluding fair-esm from requirements.run (and enforcing it explicitly at runtime) the preferred approach here? I’m very happy to follow the best practice you suggest.

sinhakrishnendu:
Another thing to bring to you kind attention — this appears to be the root cause. The conda-forge fair-esm package is noarch: python, while babappalign is architecture- and ABI-specific due to PyTorch. When built for py313, the solver fails to resolve fair-esm, making the package uninstallable.

Given this, would you recommend restricting Python versions, or keeping fair-esm out of requirements.run and enforcing it explicitly at runtime?

sinhakrishnendu:
`(base) rajamosai@rajamosai-B450-AORUS-M:~/Desktop$ conda install -c bioconda babappalign
2 channel Terms of Service accepted
Retrieving notices: done
Channels:
 - bioconda
 - defaults
Platform: linux-64
Collecting package metadata (repodata.json): done
Solving environment: failed

LibMambaUnsatisfiableError: Encountered problems while solving:
  - nothing provides fair-esm needed by babappalign-1.1.3-py313h9ee0642_0

Could not solve for environment specs
The following package could not be installed
└─ babappalign =* * is not installable because it requires
   └─ fair-esm =* *, which does not exist (perhaps a missing channel).

(base) rajamosai@rajamosai-B450-AORUS-M:~/Desktop$ 

`

sinhakrishnendu:
Gentle ping — the associated manuscript has been submitted, so I wanted to check in. Happy to update the recipe if any adjustments are required. Thanks for reviewing.

sinhakrishnendu:
A gentle ping

bgruening:
> conda install -c bioconda babappalign

This can not work. You need to use `conda install -c conda-forge -c bioconda babappalign`. Have a look at https://bioconda.github.io/#with-conda

If the conda-forge package does not work with 3.13, please point me to this and maybe we can fix that package.

mencian:
Closing in favor of #63803.

PR #61787, opened 2026-01-08
Update pyfastg to v0.2.0 (and add build_exports and update the summary)

fedarko:
This adds a `build_exports` field, which I think should fix the issue in #61786. Thanks!


fedarko:
@BiocondaBot please add label

bgruening (approved):

PR #61817, opened 2026-01-09
Fix Python compatibility issues with GTDB-tk

fplazaonate:
GTDB-tk is not compatible with Python 3.14
https://github.com/Ecogenomics/GTDBTk/issues/683

PR #61837, opened 2026-01-12
GTDB-tk is not compatible with Python 3.14

fplazaonate:
GTDB-tk is not compatible with Python 3.14
https://github.com/Ecogenomics/GTDBTk/issues/683

I forgot to pin Python to < 3.14 in the "run" requirements in the previous PR (#61817).
Sorry about that.


fplazaonate:
@BiocondaBot please add label

kbseah (approved):
looks good to me

PR #61952, opened 2026-01-16
DIPPERv0.1.3: ROCm and CPU-only support

sumit-walia:
DIPPER now supports CPU-only and AMD GPUs. The files have been modified accordingly.

### DIFF
--- a/recipes/dipper/meta.yaml
+++ b/recipes/dipper/meta.yaml
@@ -33,7 +33,6 @@ requirements:
   run:
     - boost-cpp
     - tbb
-    - zlib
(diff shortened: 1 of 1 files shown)


recipes/dipper/meta.yaml
```diff
@@ -32,12 +32,12 @@ requirements:
     - tbb-devel
   run:
     - boost-cpp
-    - tbb 
-    - tbb-devel
+    - tbb
+    - zlib
```
bgruening:
```suggestion
```

PR #61972, opened 2026-01-19
Update pegas v1.2.1

liviurotiul:
# Update pegas v1.2.1

Added support for non Snakemake workflow; changed recipe to have 2 outputs, one for the Snakemake version and one for the normal conda version.


liviurotiul:
@BiocondaBot please add label

PR #62073, opened 2026-01-22
Update cobra package version

bebatut:
Update cobra package version


bebatut:
Recipe already available on conda-forge

PR #62125, opened 2026-01-24
DIPPERv0.1.3: Separate package for CUDA and CPU-only

sumit-walia:


sumit-walia:
@BiocondaBot, please assign. 

I'm splitting the existing single-output dipper recipe into dipper (CPU) + dipper-cuda (GPU); since dipper 0.1.5-0 is already published, build: number: 0 fails the Linux Test check, while number: 1 fails Lint with build_number_needs_reset. Kindly suggest which build number this should carry, or whether the restructure needs a version bump instead.

PR #62224, opened 2026-01-28
Update Harpy build deps

pdimens:
This update only sets an upper bound on the `snakemake-minimal` dependency and increments the build number.


pdimens:
@BiocondaBot please add label

pdimens:
There are breaking changes that break the software

PR #62348, opened 2026-02-03
Add zna

mkiyer:
## Description
Adds zna package - a high-performance binary format for compressed nucleic acid sequences.

This package creates a maximally compressed high throughput sequence file format for storing pure sequence information without associated quality scores, alignment results, or other flags. This is meant to support the need for ultrafast file I/O capability.

## Features
- Supports single-end, paired-end, and interleaved reads
- Block-based streaming architecture
- C++ acceleration with pure Python fallback

## Performance
- support for FASTA/FASTQ formats
- 135 MB/s roundtrip throughput
- 2.8+ GB/s for long read processing
- 3.7-4.0x compression ratio (zstandard level 3)
     
## Checklist
- [x] Recipe follows Bioconda guidelines
- [x] Builds successfully on Linux
- [x] Includes tests
- [x] License file included
- [x] C++ compiler dependencies specified

### DIFF
--- a/recipes/seqnado/meta.yaml
+++ b/recipes/seqnado/meta.yaml
@@ -9,9 +9,8 @@ source:
   url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/seqnado-{{ version }}.tar.gz
   sha256: b53bb50e8db341e9d1dabcc543d7077701f5de2576137f6aa70268e9818889f0
 
-
 build:
-  number: 0
+  number: 1
   noarch: python
   script_env:
     - SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
@@ -29,7 +28,8 @@ requirements:
     - setuptools-scm >=6.2
     - pip
   run:
-    - pandas
+    - numpy >=1.24, <=2.1.0
+    - pandas >=2.0, <=2.3.3
     - pandera
     - pulp <=2.9.0
     - pydantic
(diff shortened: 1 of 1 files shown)


mkiyer:
python 3.10 or higher required due to 'slots' in dataclass

mkiyer:
version 0.1.5 with major compression ratio gains, now ~8-10X compression ratio with minimal performance impact.

mkiyer:
Please review.. I am ready to publish package

mkiyer:
@BiocondaBot please add label

mkiyer:
Hi @bioconda/core, this new recipe for "zna" is ready for review. It passed local linting and represents a new compressed nucleic acid format. I think it will have an important niche.. my goal was to develop a tool for storing training data for LLM (AI).

recipes/zna/build.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# Build script for conda on Unix-like systems (Linux, macOS)
+
+set -ex
+
+# Configure and build with CMake via scikit-build-core
+${PYTHON} -m pip install . -vv --no-deps --no-build-isolation
+
+# Verify the installation
+${PYTHON} -c "import zna; print(f'ZNA version: {zna.__version__ if hasattr(zna, \"__version__\") else \"0.1.0\"}')"
```
kbseah: looks like this is duplicated in the test section of meta.yaml

maybe this could be removed from the build script, and since the pip install step is only a single line, that could be moved to the build section of the recipe.

this would make it easier to update the recipe if the API changes in the future

kbseah (changes requested):
thanks for the contribution!

package is relevant to bioconda and tests pass

i just have a suggestion to avoid duplication in the tests

recipes/zna/build.sh
mkiyer: Thank you for the review and suggestion! I removed build.sh and inlined it into meta.yaml.

PR #62353, opened 2026-02-03
Add tncomp_finder recipe

fmaguire:
- Add build.sh to handle binary path and database install
- Include minimal test
- Passes tests/build locally

### DIFF
--- a/recipes/bio2zarr/meta.yaml
+++ b/recipes/bio2zarr/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "bio2zarr" %}
-{% set version = "0.1.6" %}
+{% set version = "0.1.7" %}
 
 package:
   name: {{ name|lower }}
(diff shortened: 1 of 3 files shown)


fmaguire:
@BiocondaBot please add label

BioWilko (commented):
Is it viable to get a more standard pip install method for this repo? It would be preferable to the slightly concerning build shell script if at all possible

If not just say, I'll approve

rpetit3:
I agree with @BioWilko on preferring pip installs, but given this tool v1 was released in 2019 (https://github.com/danillo-alvarenga/tncomp_finder) haha its not going to be worth the effort

fmaguire:
Yeah not ideal... thanks both, I think in an ideal world probably just simplify/combine the common functions between this and their tn3a tool but alas.

PR #62362, opened 2026-02-04
Pin antiSMASH diamond dependency to >= 2.1.21

kblin:
Based our debugging of crashing runs reported by our users, diamond 2.1.19 and 2.1.20 have a memory corruption bug that occasionally crashes the tool during antiSMASH runs. Downgrading to diamond 2.1.18 fixes the issue, so make that the default for now.

### DIFF
--- a/recipes/antismash/meta.yaml
+++ b/recipes/antismash/meta.yaml
@@ -11,7 +11,7 @@ source:
   sha256: {{ sha256 }}
 
 build:
-  number: 0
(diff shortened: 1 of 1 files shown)


tseemann:
@kblin i believe the diamond bug is fixed now?

kblin:
I've updated the PR to depend on diamond >= 2.1.21 now, I'll update the PR title as well

PR #62397, opened 2026-02-05
Add a liner

mokuno3430:
* [x] I have read the Bioconda contribution guidelines.

This PR adds the Bioconda recipe for **a-liner**.

### DIFF
--- a/recipes/a-liner/meta.yaml
+++ b/recipes/a-liner/meta.yaml
@@ -13,6 +13,8 @@ build:
   number: 0
   noarch: python
   script: "{{ PYTHON }} -m pip install . --no-deps --no-build-isolation --no-cache-dir -vvv"
+  run_exports:
+    - {{ pin_subpackage(name, max_pin="x") }}
 
 requirements:
   host:


mokuno3430:
@BiocondaBot please add label

recipes/a-liner/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/mokuno3430/a-liner/archive/refs/tags/v1.0.0.tar.gz
+  sha256: e00e17346e0414d8ffe063448dc89a71f7367de357f6a9e84b377c2963b47198
+
+build:
+  number: 0
+  noarch: python
+  script: "{{ PYTHON }} -m pip install . --no-deps -vv"
```
thanhleviet: **Should fix:** Missing `--no-build-isolation` flag. This is required for pip installs in conda build environments to ensure pip uses conda-provided packages rather than downloading its own:

```suggestion
  script: "{{ PYTHON }} -m pip install . --no-deps --no-build-isolation -vvv"
```

recipes/a-liner/meta.yaml
```diff
@@ -0,0 +1,46 @@
+{% set name = "a-liner" %}
+{% set version = "1.0.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/mokuno3430/a-liner/archive/refs/tags/v1.0.0.tar.gz
+  sha256: e00e17346e0414d8ffe063448dc89a71f7367de357f6a9e84b377c2963b47198
```
thanhleviet, line 10: **Should fix:** The version is hardcoded in the source URL. Use the Jinja variable so that future version bumps only need to change one line:

```suggestion
  url: https://github.com/mokuno3430/a-liner/archive/refs/tags/v{{ version }}.tar.gz
```

thanhleviet (commented):
sha256 verified, all CI passes, dependencies match upstream exactly (good use of `matplotlib-base`). Two items to fix below.

recipes/a-liner/meta.yaml
mokuno3430: Thanks for the suggestions! Applied.

recipes/a-liner/meta.yaml
```diff
@@ -0,0 +1,46 @@
+{% set name = "a-liner" %}
+{% set version = "1.0.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/mokuno3430/a-liner/archive/refs/tags/v1.0.0.tar.gz
```
thanhleviet: The source URL should use the Jinja `version` variable, and the `refs/tags/` prefix is unnecessary (GitHub resolves both the same way):

```suggestion
  url: https://github.com/mokuno3430/a-liner/archive/v{{ version }}.tar.gz
```

recipes/a-liner/meta.yaml
thanhleviet: `--no-build-isolation` should be added so pip uses the host-provided setuptools instead of downloading its own. Also `--no-cache-dir` is conventional:

```suggestion
  script: "{{ PYTHON }} -m pip install . --no-deps --no-build-isolation --no-cache-dir -vvv"
```

mokuno3430: Thanks for the suggestions! Applied.

recipes/a-liner/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+source:
+  url: https://github.com/mokuno3430/a-liner/archive/v{{ version }}.tar.gz
+  sha256: e00e17346e0414d8ffe063448dc89a71f7367de357f6a9e84b377c2963b47198
+
+build:
+  number: 0
+  noarch: python
+  script: "{{ PYTHON }} -m pip install . --no-deps --no-build-isolation --no-cache-dir -vvv"
+  run_exports:
+    - {{ pin_subpackage(name, max_pin="x") }}
```
thanhleviet, line 17: Optional nit: `run_exports` is unnecessary for a `noarch: python` CLI tool — no other recipe would ever list `a-liner` as a `host` dependency, so this section has no effect. Removing it would reduce noise, but it's harmless if you prefer to keep it.

```suggestion
```

mokuno3430:
After removing run_exports, the build passes but lint now fails with:
missing_run_exports: Recipe should have a run_exports statement
How would you prefer to handle this?

recipes/a-liner/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/mokuno3430/a-liner/archive/v{{ version }}.tar.gz
+  sha256: e00e17346e0414d8ffe063448dc89a71f7367de357f6a9e84b377c2963b47198
+
+build:
+  number: 0
+  noarch: python
+  script: "{{ PYTHON }} -m pip install . --no-deps --no-build-isolation --no-cache-dir -vvv"
```
thanhleviet, line 15: The `run_exports` section needs to be added back here. My earlier suggestion to remove it was incorrect — the linter requires it regardless of package type. Since this is v1.0.0 (semver), `max_pin="x"` is correct:

```suggestion
  script: "{{ PYTHON }} -m pip install . --no-deps --no-build-isolation --no-cache-dir -vvv"
  run_exports:
    - {{ pin_subpackage(name, max_pin="x") }}
```

PR #62481, opened 2026-02-08
Update rhinotype

Ephantus-Wambui:
> Updates rhinotype to version 2.5.0.
  >
  > Changes:
  > - Bumps version to 2.5.0.
  > - Adds plotly as a new dependency for interactive SNP visualization.
  > - Updates minimum versions for numpy, pandas, scipy, and biopython to match pyproject.toml.
  > - Updated SHA256 hash for the new PyPI source.

### DIFF
--- a/recipes/rhinotype/meta.yaml
+++ b/recipes/rhinotype/meta.yaml
@@ -12,6 +12,8 @@ source:
 build:
   noarch: python
   number: 0
+  run_exports:
+    - {{ pin_subpackage(name, max_pin="x.x") }}
   script: {{ PYTHON }} -m pip install . --no-deps -vv
   entry_points:
     - rhinotype = rhinotype.main:main


kbseah:
@Ephantus-Wambui could i check if you want this PR to be merged? if so, it should be tagged with the "please review and merge" label so that maintainers are aware.

Ephantus-Wambui:
> @Ephantus-Wambui could i check if you want this PR to be merged? if so, it should be tagged with the "please review and merge" label so that maintainers are aware.

Hi @kbseah, I'd like to merge this, but I don't have permission to add labels to this repository. Could you please apply the 'please review and merge' label for me? Thank you!

kbseah:
sure! for future reference, you can ask the BiocondaBot to tag the PR for you: https://bioconda.github.io/tutorials/2024-adding-bioinformatic-software-to-bioconda.html

recipes/rhinotype/meta.yaml
```diff
@@ -51,4 +52,3 @@ extra:
     - laitanawe
   skip-lints:
     - missing_run_exports
```
kbseah: Could you please clarify why this is skipped?

Ephantus-Wambui: Rhinotype is a standalone Python CLI tool intended for end-users and is not designed to be a build-time dependency for other recipes. Since it doesn't provide a library or API that other packages would link against, I skipped the missing_run_exports lint to keep the metadata focused on its primary use case.

kbseah: I see what you mean. However given how diverse bioinformatics packages are, we can't exclude the possibility of someone using it as a dependency via its CLI in e.g. a snakemake pipeline.

Ephantus-Wambui: I have updated the dependencies now the rhinotype tool can be used by other tools with no issues. I think you can go ahead and merge

PR #62557, opened 2026-02-10
Update `r-heattree`: Reduce version pin requirement and trigger rebuild for new R version

zachary-foster:
I am getting a failed run on a recipe that uses this recipe:

https://github.com/bioconda/bioconda-recipes/pull/62532

Im not sure why a specific version of R is required for `r-heattree` when that is not specified anywhere in the recipe or in the R package itself. I am assuming it is defined when the recipe is built based on the R version used? 

Either way, I wanted to reduce the version requirement for recipes that use this recipe anyway so this might fix both issues by triggering a rebuild.

### DIFF
--- a/recipes/r-heattree/meta.yaml
+++ b/recipes/r-heattree/meta.yaml
@@ -16,8 +16,6 @@ build:
   rpaths:
     - lib/R/lib/
     - lib/
-  run_exports:
-    - {{ pin_subpackage('r-heattree', max_pin="x.x") }}
 
 requirements:
   host:
--- a/recipes/r-heattree/meta.yaml
+++ b/recipes/r-heattree/meta.yaml
@@ -16,6 +16,8 @@ build:
   rpaths:
     - lib/R/lib/
     - lib/
+  run_exports:
+    - {{ pin_subpackage('r-heattree', max_pin="x") }}
 
 requirements:
   host:
--- a/recipes/r-heattree/meta.yaml
+++ b/recipes/r-heattree/meta.yaml
@@ -10,7 +10,7 @@ source:
   sha256: {{ sha256 }}
 
 build:
-  number: 0
+  number: 1
   script: ${R} CMD INSTALL --build . ${R_ARGS}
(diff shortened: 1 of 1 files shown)


zachary-foster:
@BiocondaBot please add label

kbseah:
R base was pinned 4.4 until recently (see https://bioconda.github.io/faqs.html#global-pinning and https://github.com/bioconda/bioconda-utils/pull/1070), maybe that is the reason as you surmised

recipes/r-heattree/meta.yaml
```diff
@@ -10,14 +10,14 @@ source:
...
-  number: 0
+  number: 1
   script: ${R} CMD INSTALL --build . ${R_ARGS}
   noarch: generic
   rpaths:
     - lib/R/lib/
     - lib/
   run_exports:
-    - {{ pin_subpackage('r-heattree', max_pin="x.x") }}
+    - {{ pin_subpackage('r-heattree', max_pin="x") }}
```
kbseah, line 20: could you please clarify why this was changed? didn't quite understand

kbseah (changes requested):
rebuild to update r version looks fine to me. requested clarification about change to version pin

recipes/r-heattree/meta.yaml
zachary-foster: I thought it would lead to less version conflicts in the future. The kind of application that this package would be used in would not typically require the most up to date versions for compatibility. Im fine wither way though. I mostly wanted to rebuild.

kbseah: Thanks for the clarification. The point of pin_subpackage is more to ensure that future versions of this package don't break other recipes depending on it, if for example version 0.3.0 introduces an API change, another package that requires this one would attempt to use 0.3.0 to build if max_pin is set to "x" but break. If I haven't misunderstood, please consider changing to back to "x.x"

mencian:
Closing in favor of #64527.

PR #62570, opened 2026-02-11
Update RNAstructure to version 6.5

Asperatus22:
Summary
This PR updates the rnastructure recipe from version 6.4 to 6.5.

Changes
Version Bump: Updated to the latest stable release (6.5).

Source URL: Switched to the version-specific archive to ensure hash stability.

OpenMP Support:

Improved OpenMP handling for both Linux and macOS.

Added a patch for macOS build to use llvm-openmp by modifying compiler.h.

Environment Activation: Added an activation script (etc/conda/activate.d/) to automatically set the DATAPATH environment variable. This ensures that thermodynamic data tables are found by the binaries regardless of the installation prefix.

Testing:

Added Fold -h to the test suite.

Note: Some binaries in RNAstructure 6.5 return exit code 1 when displaying help. I have adjusted the tests to account for this behavior while still verifying binary execution and linkage.

Pinning and ABI (run_exports)
I have ensured that run_exports is correctly set to handle potential breaking changes in minor versions:

YAML
build:
  run_exports:
    - {{ pin_subpackage(name | lower, max_pin="x.x") }}
 
Rationale: RNAstructure does not strictly follow semantic versioning for its CLI/API, and minor versions often introduce significant changes in data table formats or binary arguments.

Checklist
[x] Title starts with "Update".

[x] run_exports section is present in meta.yaml.

[x] Recipe is relevant to biological sciences.

[x] Tested locally using conda-build.

### DIFF
--- a/recipes/rnastructure/meta.yaml
+++ b/recipes/rnastructure/meta.yaml
@@ -36,19 +36,17 @@ test:
     - fold -h || true
 
 about:
-  home: https://rna.urmc.rochester.edu/RNAstructure.html
+  home: http://rna.urmc.rochester.edu/RNAstructure.html
   license: GPL-2.0-only
-  license_file: LICENSE
-  summary: 'RNAstructure is a complete package for RNA and DNA secondary structure prediction and analysis.'
-  description: |
-    RNAstructure is a complete package for RNA and DNA secondary structure 
-    prediction and analysis. It includes algorithms for secondary structure 
-    prediction, including predicting maximum expected accuracy structures 
-    and pseudoknotted structures. It also includes programs for predicting 
-    base pair probabilities and calculating partition functions. 
-    RNAstructure can also predict secondary structures for multiple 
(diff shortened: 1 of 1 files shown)


Asperatus22:
@BiocondaBot please add label

recipes/rnastructure/meta.yaml
```diff
@@ -47,11 +49,4 @@ about:
     which is much more accurate  than single sequence secondary structure prediction.
     Finally, RNAstructure  can take a number of different types of experiment mapping
     data to constrain  or restrain structure prediction. These include chemical mapping,
-    enzymatic  mapping, NMR, and SHAPE data.'
-
-extra:
```
bgruening, line 52: Any reason why you removed this?

Asperatus22: Sorry, I accidentally truncated the description. It is now restored. Thanks for catching that!

PR #62585, opened 2026-02-11
Add ratatoskr v1.0.2

Fabian-Bastiaanssen:
Initial Bioconda recipe for ratatoskr (v1.0.2), a bioinformatics tool for
programmatically collecting and downloading taxonomic type strain data
from biological data providers.

ratatoskr is a noarch Python package that provides both a command-line
interface and Python API for accessing, aggregating, and organizing
microbial taxonomy and type strain metadata, supporting reproducible
biological data retrieval workflows.

The recipe includes a run_exports section to ensure downstream
compatibility across non-breaking releases, in line with Bioconda
guidelines.

Previously had HTTP errors for linting unrelated to Ratatoskr, hoping a repull will fix that

### DIFF
--- a/recipes/alphagenome/meta.yaml
+++ b/recipes/alphagenome/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "alphagenome" %}
-{% set version = "0.5.1" %}
+{% set version = "0.6.0" %}
 
 package:
   name: {{ name }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
-  sha256: e7ca21dcb431ffa522d6f911ae7979d5f1e1ef63c621244d1e00db875ef8703b
+  sha256: 2b23b93dc599e80d933a42512606d56c9c2e88be92e2484fba3fb39accb3f6dc
 
 build:
   number: 0
--- a/recipes/atlas-fastq-provider/meta.yaml
+++ b/recipes/atlas-fastq-provider/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "0.4.7" %}
+{% set version = "0.4.8" %}
 
 package:
   name: atlas-fastq-provider
@@ -6,7 +6,7 @@ package:
 
 source:
   url: https://github.com/ebi-gene-expression-group/atlas-fastq-provider/archive/v{{ version }}.tar.gz
-  sha256: 578d1b69a7582fa47073ba4ad7b3bf98df4aceea6fe96a25bb38674a1d362f33
+  sha256: db5ad4a435d41b5516f330d909d09e5085e6e839eb119206e5a2741ea7bd2011
 
 build:
   number: 0
--- a/recipes/atol-bpa-datamapper/meta.yaml
+++ b/recipes/atol-bpa-datamapper/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "atol-bpa-datamapper" %}
-{% set version = "0.1.17" %}
+{% set version = "0.2.0" %}
 
 package:
   name: {{ name }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://github.com/TomHarrop/atol-bpa-datamapper/archive/refs/tags/v{{ version }}.tar.gz
-  sha256: a0fee6be37bd24552523b5a3c9febb4b669b45bca29306b639b68a7e58b82c2c
+  sha256: 3eee9e65e8dd57f37e3ed4378da245ac952eccdf3890c446aa3d683911f66ca2
 
 build:
   number: 0
--- a/recipes/atol-genome-launcher/meta.yaml
+++ b/recipes/atol-genome-launcher/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "atol-genome-launcher" %}
-{% set version = "0.2.1" %}
+{% set version = "0.4.1" %}
 
 package:
   name: {{ name }}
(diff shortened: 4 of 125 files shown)


recipes/ratatoskr/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  commands:
+    - ratatoskr --help
+    - pip check
+  requires:
+    - python {{ python_min }}
+    - pip
+
+about:
+  home: https://github.com/Fabian-Bastiaanssen/Ratatoskr
+  summary: Tool for collecting and downloading taxonomic type strain data.
```
tseemann: Does this also need `licence_family:` ?

recipes/ratatoskr/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+requirements:
+  host:
+    - python {{ python_min }}
+    - setuptools <=80
+    - wheel
+    - pip
+    - hatchling
+  run:
+    - python {{ python_min }}
+    - pip
```
tseemann: Is `pip` needed for users to run the tool?

recipes/ratatoskr/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  - {{ pin_subpackage(name|lower, max_pin="x") }}
+requirements:
+  host:
+    - python {{ python_min }}
+    - setuptools <=80
+    - wheel
+    - pip
+    - hatchling
+  run:
+    - python {{ python_min }}
```
thanhleviet: `python {{ python_min }}` renders to `python 3.12`, which in conda means `>=3.12,<3.13` — only 3.12.x is allowed. But the upstream `setup.py` declares `python_requires=">=3.12"`, which includes 3.13+. This should be:
```yaml
- python >={{ python_min }}
```

recipes/ratatoskr/meta.yaml
thanhleviet: `pip` should not be a runtime dependency. It's not in the upstream `install_requires` and is only needed for `pip check` in the test section (where it's already listed under `test: requires`).

recipes/ratatoskr/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps
+  number: 0
+  run_exports:
+  - {{ pin_subpackage(name|lower, max_pin="x") }}
+requirements:
+  host:
+    - python {{ python_min }}
+    - setuptools <=80
+    - wheel
```
thanhleviet, line 22: `hatchling` is listed here but the upstream `pyproject.toml` uses `setuptools.build_meta` as the build backend, not hatchling. This is unused and should be removed.

recipes/ratatoskr/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+source:
+  url: https://github.com/Fabian-Bastiaanssen/Ratatoskr/archive/v{{ version }}.tar.gz
+  sha256: 96b350c31d9f8356af81c6118596b747e0361c84dd6719d3e649af5aeb9b0266
+
+build:
+  entry_points:
+    - ratatoskr=ratatoskr.__main__:main
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps
```
thanhleviet: Missing `--no-build-isolation` flag. Without it, pip may attempt to create an isolated build environment and fetch build deps from PyPI, which won't work in the conda build environment. Standard pattern:
```yaml
script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```

recipes/ratatoskr/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/Fabian-Bastiaanssen/Ratatoskr/archive/v{{ version }}.tar.gz
+  sha256: 96b350c31d9f8356af81c6118596b747e0361c84dd6719d3e649af5aeb9b0266
+
+build:
+  entry_points:
+    - ratatoskr=ratatoskr.__main__:main
```
thanhleviet: The `entry_points` section is redundant — `setup.py` already declares the `ratatoskr` console script entry point. When using `pip install`, entry points from `setup.py` are installed automatically. Declaring them again in `meta.yaml` can cause conflicts or duplicate wrapper scripts.

recipes/ratatoskr/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - ncbi-datasets-cli >=18.10.2
+    - biopython >=1.86
+    - loguru >=0.7.2
+    - polars >=1.32.2
+    - pyarrow >=13.0.0
+    - rich-click >=1.8.8
+    - setuptools <=80
+    - tqdm >=4.66.1
+    - python-keycloak
+    - aiohttp
```
thanhleviet: `python-keycloak` and `aiohttp` (line 40) are not in the upstream `setup.py`'s `install_requires`. Are these transitive dependencies of `async-dsmz` or `bacdive`? If so, they should be tracked by those packages' own metadata rather than listed here explicitly.

recipes/ratatoskr/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  run:
+    - python {{ python_min }}
+    - pip
+    - ncbi-datasets-cli >=18.10.2
+    - biopython >=1.86
+    - loguru >=0.7.2
+    - polars >=1.32.2
+    - pyarrow >=13.0.0
+    - rich-click >=1.8.8
+    - setuptools <=80
```
thanhleviet: `setuptools<=80` as a runtime dependency is unusual — it typically means the package uses `pkg_resources` at runtime (a deprecated pattern). Also, the `<=80` upper bound will eventually become stale. Consider flagging this upstream.

thanhleviet (approved):
All previous review comments have been addressed. CI passes on all platforms, sha256 verified, dependencies match upstream (with `ncbi-datasets-cli` as a valid external CLI dep). Recipe looks good to merge.

thanhleviet:
@BiocondaBot merge

PR #62271, opened 2026-01-30
Update thermorawfileparser to 2.0.0.dev

jonasscheid:
## Summary
- Update thermorawfileparser from 1.4.5 to 2.0.0.dev
- Remove mono dependency (now self-contained .NET 8 build)
- Remove wrapper script (native executable runs directly)
- Change from noarch to linux64-only (platform-specific binary)

## Changes
- Version: 1.4.5 → 2.0.0.dev
- Source: Now using self-contained Linux binary from release
- Dependencies: Removed `mono >=5` requirement
- Build: Changed from `noarch: generic` to `skip: True  # [not linux64]`

## Upstream Release
https://github.com/CompOmics/ThermoRawFileParser/releases/tag/v.2.0.0-dev

Notable upstream changes:
- Code fully migrated to .NET 8 (mono dependency obsolete)
- Parquet format upgraded to v0.2
- Charge state arrays now supported in mzML and MGF
- Precursor ion intensity added to MGF
- Latest instrument CV terms (PSI-MS 4.1.204) integrated

PR #62665, opened 2026-02-14
Update ThermoRawFileParser to v2.0.0-dev

caetera:
- Remove mono dependency
- Change to full binary release 
- Remove wrapper script

### DIFF
--- a/recipes/thermorawfileparser/meta.yaml
+++ b/recipes/thermorawfileparser/meta.yaml
@@ -1,5 +1,7 @@
 {% set name = "ThermoRawFileParser" %}
 {% set version = "2.0.0" %}
+{% set tag = "v.2.0.0-dev" %}
+
 
 package:
   name: {{ name|lower }}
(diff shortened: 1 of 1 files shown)


caetera:
It should be possible to build it within conda, if this way is preferred.  We already build binaries on every release for all major platforms (the ones supported by `dotnet publish`), initially I tried to use a self-contained release (all dependencies included), but still got unresolved dependencies. In the current implementation I use framework-based release , i.e. `dotnet 8` is required during runtime.

CLI of 2.x is generally the same as 1.x; the main breaking change is switch to Net8, there is a change in the format of Parquet output, and some minor improvements (the complete release notes https://github.com/CompOmics/ThermoRawFileParser/releases/tag/v.2.0.0-dev).

caetera:
I have updated the recipe to build the tool with conda from the "tagged" source. To keep it platform-agnostic, it is built as framework-dependent, but requires dotnet/aspnet-runtime. Finally, a more thorough test is included.

caetera:
Duplicate of #62271

PR #62669, opened 2026-02-15
ipcr: build and test all CLIs

ericksamera:
Build and install ipcr-probe, ipcr-nested, ipcr-multiplex, and ipcr-thermo from source. Add smoke tests for each CLI. Require go>=1.20 to support Go modules. Bump build number to 2 for v4.1.1.


ericksamera:
@BiocondaBot please add label

PR #62719, opened 2026-02-17
Add recipe: binge 1.1.0

zkstewart:
Pull request to add a new recipe for the BINge software available from https://github.com/zkstewart/BINge.

This tool is intended for clustering of transcripts from one or more species. Reference genomes are used to produce clusters based on locus orthology. The results are intended for use during multispecies differential gene expression (DGE) analysis.

Pull request involves build from program version 1.1.0 as tagged/released in the repository.

### DIFF
--- a/recipes/binge/build.sh
+++ b/recipes/binge/build.sh
@@ -3,9 +3,9 @@
 mkdir -p $PREFIX/bin
 
 cp BINge.py $PREFIX/bin/BINge.py
-cp BINge.py $PREFIX/bin/BINge
+ln -s $PREFIX/bin/BINge.py $PREFIX/bin/BINge
 cp BINge_post.py $PREFIX/bin/BINge_post.py
-cp BINge_post.py $PREFIX/bin/BINge_post
+ln -s $PREFIX/bin/BINge_post.py $PREFIX/bin/BINge_post
 cp _version.py $PREFIX/bin/_version.py
 
 mkdir -p $PREFIX/bin/modules


zkstewart:
I think the failed test is resulting from a connection issue during the test. If re-run when connection to GitHub is possible, this error should not occur.

zkstewart:
@BiocondaBot please add label

recipes/binge/build.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+mkdir -p $PREFIX/bin
+
+cp BINge.py $PREFIX/bin/BINge.py
```
kbseah, line 5: Could you clarify why BINge.py and BINge_post.py are each copied over twice under different names?

recipes/binge/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - BINge_post -h
+    - gmap --help
+    - mmseqs -h
+    - salmon -h
+
+about:
+  home: https://github.com/zkstewart/BINge
+  dev_url: https://github.com/zkstewart/BINge
+  doc_url: https://github.com/zkstewart/BINge/wiki
+  license: GPL-3
```
kbseah: Could you clarify if this is `GPL-3.0-only` or `GPL-3.0-or-later`?

kbseah (changes requested):
Thanks for the contribution! I've highlighted two points to be clarified

recipes/binge/build.sh
zkstewart: My intent was to provide the ability to invoke the software as if it is an executable (i.e., `BINge <args>` or `BINge_post <args>`) or through Python as a script file (i.e., `python /location/of/BINge.py <args>`) if the default / system PATH Python that gets called by the shebang line is not the one someone wants to invoke BINge with.

In practice this is probably not important since the default Python should be the one that's part of their conda environment...

To avoid unnecessary duplication of the files, the `.py` files could be omitted.

recipes/binge/meta.yaml
zkstewart: As of now the software is `GPL-3.0-only`, although I don't have any problems with adopting the `or-later` clause. To keep things simple, it is `only`.

kbseah: Thanks, it is ultimately your call to make, just wanted to make sure it's consistent between the recipe and the repository. If you could update it here that would be great

recipes/binge/build.sh
kbseah: I see... Maybe if you symlink BINge and BINge_post to the .py files that would be cleaner?

zkstewart: Agreed, and thanks for this suggestion!

PR #62731, opened 2026-02-17
Superintervals update recipe

kcleal:
Trigger build for newer python versions
Removed cython from run


kcleal:
@BiocondaBot please add label

PR #62760, opened 2026-02-18
Rename AMRFior to GeneFior and added additional tooling

NickJD:
AMRFior has now been absorbed into GeneFior.

### DIFF
--- a/recipes/atol-genome-launcher/meta.yaml
+++ b/recipes/atol-genome-launcher/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "atol-genome-launcher" %}
-{% set version = "0.3.1" %}
+{% set version = "0.4.1" %}
 
 package:
   name: {{ name }}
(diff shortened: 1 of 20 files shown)


tseemann:
@bgruening is this the correct protocol to rename a package?

bgruening:
Yes, just note that the old packages will NOT disappear. They will stay.

If you want to move people over ... you could release one final version, that prints a warning during install or depends on the new one and prints a warning ... but all of that is optional and depends on your community.

NickJD:
> Yes, just note that the old packages will NOT disappear. They will stay.
> 
> If you want to move people over ... you could release one final version, that prints a warning during install or depends on the new one and prints a warning ... but all of that is optional and depends on your community.

Ah ok great and thanks both.
I couldn't find any guide on what I should do.

edit - I will add a final update to AMRfior and add GeneFior separately

PR #62775, opened 2026-02-18
Add fgumi 0.1.0

nh13:
Add recipe for [fgumi](https://github.com/fulcrumgenomics/fgumi), high-performance tools for UMI-tagged sequencing data: extraction, grouping, and consensus calling.

### DIFF
--- a/recipes/fgumi/meta.yaml
+++ b/recipes/fgumi/meta.yaml
@@ -37,7 +37,7 @@ about:
     - THIRDPARTY.yml
   summary: "High-performance tools for UMI-tagged sequencing data."
   dev_url: "https://github.com/fulcrumgenomics/{{ name }}"
-  doc_url: "https://github.com/fulcrumgenomics/{{ name }}/blob/v{{ version }}/README.md"
+  doc_url: "https://github.com/fulcrumgenomics/{{ name }}/blob/{{ name }}-v{{ version }}/README.md"
 
 extra:
   additional-platforms:


recipes/fgumi/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - {{ compiler('rust') }}
+    - cargo-bundle-licenses
+  run:
+    - r-base
+    - r-ggplot2
+    - r-scales
+
+test:
+  commands:
+    - fgumi --help
```
tseemann: Do you have a test that will ensure the `R` deps will work?

recipes/fgumi/meta.yaml
```diff
@@ -0,0 +1,46 @@
+{% set name = "fgumi" %}
+{% set version = "0.1.0" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/fulcrumgenomics/{{ name }}/archive/v{{ version }}.tar.gz
```
thanhleviet: This is the root cause of both CI failures:
```
RuntimeError: Could not download https://github.com/fulcrumgenomics/fgumi/archive/v0.1.0.tar.gz
WARNING: Error: HTTP 404 NOT FOUND
```

The repo has no `v0.1.0` tag — the actual tag is `fgumi-v0.1.0` (Cargo workspace convention). The URL should be:
```yaml
url: https://github.com/fulcrumgenomics/{{ name }}/archive/{{ name }}-v{{ version }}.tar.gz
```
Note: the URL currently resolves (302 redirect) but it's unreliable since there's no matching git ref — only `fgumi-v0.1.0` exists in `refs/tags/`.

recipes/fgumi/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - {{ pin_subpackage('fgumi', max_pin="x.x") }}
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('rust') }}
+    - cargo-bundle-licenses
+  run:
+    - r-base
+    - r-ggplot2
```
thanhleviet, line 24: These R runtime dependencies (`r-base`, `r-ggplot2`, `r-scales`) don't appear to belong here. The project is 99.8% Rust, the `Cargo.toml` has zero R-related dependencies, and the README makes no mention of R. Is this a copy-paste from another recipe?

If fgumi does have an optional R-based plotting feature, please document where it's used. Otherwise these should be removed — they add significant install weight (~500MB+ for r-base alone).

recipes/fgumi/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -e
+
+cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
+
+cargo install --no-track --verbose --root "${PREFIX}" --path .
```
thanhleviet: Missing `--locked` flag. This ensures the build uses the exact dependency versions from `Cargo.lock` for reproducibility:
```bash
cargo install --no-track --locked --verbose --root "${PREFIX}" --path .
```

recipes/fgumi/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+test:
+  commands:
+    - fgumi --help
+
+about:
+  home: "https://github.com/fulcrumgenomics/{{ name }}"
+  license: MIT
+  license_family: MIT
+  license_file: LICENSE
```
thanhleviet: The `build.sh` generates `THIRDPARTY.yml` via `cargo-bundle-licenses`, but it's not included in `license_file`. For a Rust package with many statically-linked crates, this should be:
```yaml
license_file:
  - LICENSE
  - THIRDPARTY.yml
```

recipes/fgumi/meta.yaml
nh13: The R deps are intentional. fgumi embeds an R script (`resources/CollectDuplexSeqMetrics.R`) via `include_str!` at compile time, then writes it to a temp file and executes it via `Rscript` at runtime. It uses ggplot2 and scales for duplex sequencing QC plots.

recipes/fgumi/meta.yaml
nh13: Fixed. Updated to use `fgumi-v0.1.0` tag and recomputed sha256.

recipes/fgumi/build.sh
nh13: Added `--locked`.

recipes/fgumi/meta.yaml
nh13: Fixed. license_file now includes both LICENSE and THIRDPARTY.yml.

recipes/fgumi/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+about:
+  home: "https://github.com/fulcrumgenomics/{{ name }}"
+  license: MIT
+  license_family: MIT
+  license_file:
+    - LICENSE
+    - THIRDPARTY.yml
+  summary: "High-performance tools for UMI-tagged sequencing data."
+  dev_url: "https://github.com/fulcrumgenomics/{{ name }}"
+  doc_url: "https://github.com/fulcrumgenomics/{{ name }}/blob/v{{ version }}/README.md"
```
thanhleviet: Very minor nit: `doc_url` references `v{{ version }}` but the repo's tags use `fgumi-v{{ version }}`. This link may 404. Consider:
```yaml
doc_url: "https://github.com/fulcrumgenomics/{{ name }}/blob/{{ name }}-v{{ version }}/README.md"
```
Non-blocking.

thanhleviet (approved):
Lint, Linux, and OSX-64 tests all pass. ARM still pending but should be fine for a standard Rust build. LGTM.

PR #62788, opened 2026-02-19
Add scna-spice 0.1.0

tlkaufmann:
Added the package SPICE a package for the analysis of somatic copy-number alterations (SCNA) in human cancers.

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/recipes/scna-spice/meta.yaml
+++ b/recipes/scna-spice/meta.yaml
@@ -21,10 +21,10 @@ build:
 
 requirements:
   host:
-    - python >=3.8
+    - python >=3.8,<3.13   # ortools only supports up to 3.12
     - pip
   run:
-    - python >=3.8
+    - python >=3.8,<3.13
     - medicc2 >=1.0
     - numpy
     - pandas


recipes/scna-spice/meta.yaml
```diff
@@ -0,0 +1,56 @@
+{% set name = "scna-spice" %}
+{% set version = "0.1.0" %}
+{% set hash = "5dacad73df5397b95dc129d80ee18d598f42d7f6134fdc21ed5e6fd90c483bf3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: "https://files.pythonhosted.org/packages/6c/ef/bf33715252533ad6647f00610c47a0dc30054e5cd514d006743c400d5dca/scna_spice-0.1.0.tar.gz"
```
thanhleviet, line 10: The source URL is a hardcoded full PyPI path with the content-hash directory. This will break if PyPI reorganizes storage. Use the standard templated format instead:
```yaml
url: https://pypi.org/packages/source/s/scna-spice/scna_spice-{{ version }}.tar.gz
```

recipes/scna-spice/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  sha256: {{ hash }}
+
+build:
+  noarch: python
+  number: 0
+  script:
+    - {{ PYTHON }} -m pip install ortools==9.8.3296
```
thanhleviet, line 17: This is the root cause of the Linux Test failure:
```
ERROR: Could not find a version that satisfies the requirement ortools==9.8.3296 (from versions: none)
```

Installing dependencies via `pip` in the build script is a bioconda anti-pattern — dependencies must be declared in the `requirements` section so conda can track them. `ortools-python` is available on [conda-forge](https://anaconda.org/conda-forge/ortools-python), so replace this line and add to `run` requirements:
```yaml
run:
  - ortools-python ==9.8.3296
```
(or relax the pin if compatible with newer versions)

recipes/scna-spice/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - {{ PYTHON }} -m pip install . --no-deps --no-build-isolation -vv
+  run_exports:
+    - {{ pin_subpackage(name, max_pin="x.x") }}
+
+requirements:
+  host:
+    - python >=3.8,<3.13   # ortools only supports up to 3.12
```
thanhleviet, line 24: If the Python constraint is due to `ortools` compatibility, this should be handled by the `ortools-python` conda package's own metadata rather than hardcoded here. Once `ortools-python` is a proper conda dep, the solver will enforce the Python constraint automatically.

recipes/scna-spice/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - pip
+  run:
+    - python >=3.8,<3.13
+    - medicc2 >=1.0
```
thanhleviet, line 28: `medicc2` is not in the upstream `setup.py`'s `install_requires` — it appears to be an external tool dependency invoked via subprocess rather than a Python import. This is fine for a bioconda recipe, but worth a comment noting why it's here (e.g., `# external tool required by SPICE workflow`).

PR #63183, opened 2026-03-04
Update zol build

raufs:
Updating zol build to use a fork of the gravis repo that has been edited to be more compatible recent versions of Python.

----

### General instructions

```yaml
build:
  run_exports:
    - ...

```

with `...` being one of:

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/recipes/zol/build.sh
+++ b/recipes/zol/build.sh
@@ -16,7 +16,7 @@ chmod +x ${PREFIX}/bin/splitDiamondResultsForFai
 
 # setup gravis from a pinned Git commit to ensure reproducible builds
 # We point pip directly to the git repository 
-curl -sL https://github.com/raufs/gravis/archive/refs/tags/v0.1.1.tar.gz | sha256sum | grep -q "d8a8b09a23522ab746eea740adc05f22cb6a1d5190c1b64ee4f304e62b37268b" && echo "Matching SHA256 for gravis" || echo "Mismatch SHA256 for gravis" && exit 1
+curl -sL https://github.com/raufs/gravis/archive/refs/tags/v0.1.1.tar.gz | sha256sum | grep -q "d8a8b09a23522ab746eea740adc05f22cb6a1d5190c1b64ee4f304e62b37268b  -" && echo "Matching SHA256 for gravis" || echo "Mismatch SHA256 for gravis" && exit 1
 $PYTHON -m pip install https://github.com/raufs/gravis/archive/refs/tags/v1.0.0.tar.gz --upgrade --no-deps --no-build-isolation -vvv
 
 # create folder for database download
--- a/recipes/zol/build.sh
+++ b/recipes/zol/build.sh
@@ -39,4 +39,4 @@ unset ZOL_DATA_PATH
 unset ZOL_EXEC_PATH
 EOF
 
-echo "Users of zol, please note, we use an unofficial fork of gravis (v1.0.0; https://github.com/raufs/gravis), not the original/official gravis (https://github.com/robert-haas/gravis)"
+echo "Users of zol, please note, we use an unofficial fork of gravis (v0.1.1; https://github.com/raufs/gravis), not the original/official gravis (https://github.com/robert-haas/gravis)"
--- a/recipes/zol/build.sh
+++ b/recipes/zol/build.sh
@@ -17,7 +17,7 @@ chmod +x ${PREFIX}/bin/splitDiamondResultsForFai
 # setup gravis from a pinned Git commit to ensure reproducible builds
 # We point pip directly to the git repository 
 curl -sL https://github.com/raufs/gravis/archive/refs/tags/v0.1.1.tar.gz -o gravis_v0.1.1.tar.gz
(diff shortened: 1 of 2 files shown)


raufs:
@BiocondaBot please add label

recipes/zol/build.sh
```diff
@@ -14,6 +14,10 @@ chmod +x ${PREFIX}/bin/runRBH
 chmod +x ${PREFIX}/bin/splitDiamondResults
 chmod +x ${PREFIX}/bin/splitDiamondResultsForFai
 
+# setup gravis from a pinned Git commit to ensure reproducible builds
+# We point pip directly to the git repository 
+$PYTHON -m pip install git+https://github.com/raufs/gravis.git@master --upgrade --no-deps --no-build-isolation -vvv
```
kbseah: Dependencies should be installed by conda and not by pip from a git repo, see https://bioconda.github.io/contributor/guidelines.html#stable-urls

I also noted that this points to a personal fork of gravis. Did the upstream author not respond to requests for an update?

The issue is that installing the dependency like this is potentially unstable, and may lead to inconsistencies between builds down the line.

One alternative is to patch the gravis recipe and rebuild it for higher python versions.

recipes/zol/meta.yaml
```diff
@@ -10,7 +10,7 @@ source:
   sha256: f6751f61e9d71cd2430da45d5bf7bff70a7116fb55665b331f8e91b9022f345a
 
 build:
-  number: 1
+  number: 3
```
kbseah, line 17: increment was skipped

kbseah (changes requested):
Thanks for the update! I don't think the gravis dependency should be installed by pip from a personal fork git repo. Please see my comment below

recipes/zol/meta.yaml
raufs: Thank you - it seems there was some more general update and looking at the files for zol at: https://anaconda.org/channels/bioconda/packages/zol/files - there is already a second build for this version that is not reflected in the recipe. I also get an error when attempting to set build to 2, so I think 3 is appropriate here.

recipes/zol/build.sh
raufs: I agree that this is not ideal. The author of gravis unfortunately is not responding to PRs and the changes made in my fork of it are likely too involved to patch, involving dependency removal/replacement. 

I have made a release of my fork which is in compliance I think with the section you sent of the Bioconda docs as it provides a stable link. Am also attempting to add sha256 checking as a safety step prior to pip installation. Getting some issues on arm64 - but hopefully will figure it out soon. Will ping you after and be grateful for another review.

raufs:
Hi @kbseah , thank you again for reviewing here! I believe both concerns should be now addressed. Namely, the pip installation now references a release (tar.gz) of my fork of gravis, which should meet the criteria of a stable link according to the Bioconda doc you referenced. I also added sha256 checking to add further safety for consistent behavior across platforms. Then, while not reflected in zol's recipe, there are already 2.0 builds for this version, thus the build bump needs to be to 3 or otherwise I get errors.

recipes/zol/meta.yaml
kbseah: thanks for clarifying!

kbseah:
hi @raufs , thanks for reworking the recipe to address my earlier review! I'm still not sure that specifying, then overwriting, the gravis dependency is a good practice here. I looked into whether there is another way this could be done, and the conda-build documentation on [building from multiple sources](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#source-from-multiple-sources): both the zol and your gravis fork source code could be specified as sources, downloaded then extracted to different subfolders.

Checksums for both would be specified in the recipe and checked by conda-build so you won't need to do that separately for the gravis fork in the build script. You'd have to modify the build script so that pip installs them both from the downloaded packages.

Perhaps @bioconda/core could chime in on the best practice. Another alternative is to add a new recipe for the gravis fork under another name.

raufs:
@kbseah , thanks again, that is an interesting alternative to use multiple sources! I am about to release a new version of zol today which will likely get detected by the autobump here so I will ultimately just close this ticket. I will likely just transfer our progress from here there and try this alternate approach.

kbseah:
hi @raufs , I've converted the autobump PR to draft so that it won't be automatically merged. thanks for your patience in trying to find a solution for this issue! perhaps you could post in the gitter chatroom to get some other opinions on the best way forward.

raufs:
Hi @kbseah - I think zol builds are currently broke so I would like to expedite getting this recipe updated. However, it seems that the installation you suggested of using multiple sources works from my testing here and is indeed a more elegant solution than what I had previously, so thank you for your suggestions! 

I will probably just accept it if it passes the tests on the new autobump PR, but feel free to post in the gitter chatroom / in issues and reference this case. If you and the Bioconda core team thinks it is problematic, I am happy to create another manual PR and take another solution. One would be to just pin setuptools to 77.0.3 which should allow the official gravis on conda-forge to be used as is, but it is not ideal for moving forward to do this I think.

PR #63201, opened 2026-03-04
Update hits 0.5.1 to ensure a 3.12 build

jeffhussmann:
Adds a `conda_build_config.yaml` modeled after the one in pysam to ensure that a python 3.12 build happens.

PR #63214, opened 2026-03-04
Update hits 0.5.1 to ensure a 3.12 without using conda_build_config.yaml

jeffhussmann:
Has identical goal as https://github.com/bioconda/bioconda-recipes/pull/63201, but avoids using conda_build_config.yaml to avoid issues that pull request has run into.

This approach has been separated into a new pull request to not interfere with gitter discussion of the other one.


bgruening (commented):
You haven't changed the run dependencies.

Can you remind me which problem you are trying to solve here?

jeffhussmann:
The high-level goal is to update the recipe for another package, knock-knock, which now requires hits >=0.5.1.

Both hits and knock-knock now require python >=3.12. When trying to build knock-knock, it appears that some of its other dependencies have only been built for 3.12, so if hits is only built for 3.13, knock-knock cannot build.

So the immediate goal is to ensure that hits is built for python 3.12.

My understanding is that the preferred way to do this would be using `conda_build_config.yaml`, but I have been unable to get that approach to work - see https://github.com/bioconda/bioconda-recipes/pull/63201/changes, and a longer description on the biconda gitter channel.

Any guidance would be appreciated!

bgruening:
> Both hits and knock-knock now require python >=3.12. When trying to build knock-knock, it appears that some of its other dependencies have only been built for 3.12, so if hits is only built for 3.13, knock-knock cannot build.

Which dependencies are only build for 3.12. And why does knock-knock needs python 3.12 and higher? Some new Python snytax you are using?

jeffhussmann:
Yes, the python3.12 requirements come from backwards-incompatible changes to f-strings - https://docs.python.org/3/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings.

`viennarna>=2.6.4` is the knock-knock dependency that causes trouble. Running `conda search viennarna --info` appears to confirm that while there are earlier viennarna versions with 3.13 builds, there are none >=2.6.4.

mencian (approved):

PR #63598, opened 2026-03-17
Add Tir learner4

KGerhardt:
I've completely rewritten the TIR-Learner tool as part of the Ou lab's efforts to keep the tool appropriately modernized. Because TIR-Learner v3 is still in review for publication, we would like to publish this version as a separate TIR-Learner4.

### DIFF
--- a/recipes/tir-learner4/build.sh
+++ b/recipes/tir-learner4/build.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+set -euo pipefail
+
+# Install package files
+mkdir -p "$PREFIX/lib/tir-learner4"
+cp -r TIR-Learner3/* "$PREFIX/lib/tir-learner4/"
+
+# Fix shebang in main script (original has #!/usr/app/env python3, a typo)
+sed -i '1s|.*|#!/usr/bin/env python3|' "$PREFIX/lib/tir-learner4/TIR-Learner.py"
+
+# Create CLI wrapper
+mkdir -p "$PREFIX/bin"
+cat > "$PREFIX/bin/tirlearner4" << 'EOF'
+#!/bin/bash
+exec python3 "$(dirname "$(dirname "$(readlink -f "$0")")")/lib/tir-learner4/TIR-Learner.py" "$@"
+EOF
+chmod +x "$PREFIX/bin/tirlearner4"
--- a/recipes/tir-learner4/meta.yaml
+++ b/recipes/tir-learner4/meta.yaml
@@ -7,7 +7,8 @@ source:
   sha256: f36828693f1c3bbf4f7c2441b7ba8d64d8a42653d0cb3738e52b0c365444e2e8
 
 build:
-  number: 2
+  number: 3
+  skip: True  # [osx]
 
 requirements:
   build:
--- a/recipes/tir-learner4/meta.yaml
+++ b/recipes/tir-learner4/meta.yaml
@@ -1,0 +1,46 @@
+package:
+  name: tir-learner4
+  version: "1.0"
+
+source:
+  url: https://github.com/KGerhardt/TIR-Learner/archive/refs/tags/v4.0.1.tar.gz
+  sha256: f36828693f1c3bbf4f7c2441b7ba8d64d8a42653d0cb3738e52b0c365444e2e8
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+  host:
+    - python 3.9.*
+    - numpy
+    - cython
+    - pip
+    - setuptools
+    - wheel
+  run:
+    - python 3.9.*
+    - numpy
+    - pyfastx
+    - pywfa
+    - keras
+    - pytorch
+    - genometools-genometools
+    - genericrepeatfinder
+    - blast
+    - pigz
+
+test:
+  commands:
+    - tirlearner4 --help
+    
+about:
+  home: https://github.com/KGerhardt/TIR-Learner
+  license: GPL-3.0
+  summary: "Accelerated tool for identifying TIR transposable elements"
+
+extra:
+  recipe-maintainers:
+    - KGerhardt
--- a/recipes/tir-learner4/meta.yaml
+++ b/recipes/tir-learner4/meta.yaml
@@ -35,7 +35,9 @@ requirements:
 
 test:
   commands:
-    - tirlearner4 --help
+  license: GPL-3.0-or-later
+  license_family: GPL3
+  license_file: LICENSE
     
 about:
   home: https://github.com/KGerhardt/TIR-Learner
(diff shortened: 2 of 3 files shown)


KGerhardt:
@Juke34

This includes a change to the pywfa script, too - it seems that OSX is a problem for building pywfa and thus TIR-Learner4. I've added a clause to skip OSX on the build for both.

oushujun:
Hi @thanhleviet, this is an update to the existing TIR-Learner recipe (https://github.com/bioconda/bioconda-recipes/pull/55111) but from a different author/maintainer. Can they use the same recipe handle to avoid confusion when resolving?

Shujun

recipes/tir-learner4/meta.yaml
```diff
@@ -0,0 +1,47 @@
+package:
+  name: tir-learner4
+  version: "1.0"
```
thanhleviet: **Version mismatch (important):** The recipe says `version: "1.0"` but the source tag is `v4.0.1`. This should be `"4.0.1"` to match the upstream release.

```suggestion
  version: "4.0.1"
```

recipes/tir-learner4/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  name: tir-learner4
+  version: "1.0"
+
+source:
+  url: https://github.com/KGerhardt/TIR-Learner/archive/refs/tags/v4.0.1.tar.gz
+  sha256: f36828693f1c3bbf4f7c2441b7ba8d64d8a42653d0cb3738e52b0c365444e2e8
+
+build:
+  number: 3
+  skip: True  # [osx]
```
thanhleviet, line 18: **Build number should be 0 (important):** For a new recipe at its initial version, the build number should start at 0. Increment only when rebuilding the same version.

```suggestion
  number: 0
```

thanhleviet (commented):
Thanks for the recipe! All CI checks pass and the sha256 is verified. A few issues to address before this can be merged.

**Re @oushujun's question:** This is a naming decision for the maintainers. Having both `tir-learner` (v3) and `tir-learner4` in bioconda could confuse users. If v4 supersedes v3, updating the existing `tir-learner` recipe to v4.0.1 (with the new source URL) would be cleaner. If both versions must coexist, keeping them separate is fine — but the version string should reflect reality (see below).

recipes/tir-learner4/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - keras
+    - pytorch
+    - genometools-genometools
+    - genericrepeatfinder
+    - blast
+    - pigz
+
+test:
+  commands:
+    - tirlearner4 --help
```
thanhleviet: **Missing `license_file` (critical):** GPL-3.0 recipes **must** include `license_file`. The upstream repo has a `LICENSE` file. Also, the SPDX identifier should be `GPL-3.0-or-later` (matching the existing `tir-learner` recipe).

```suggestion
  license: GPL-3.0-or-later
  license_family: GPL3
  license_file: LICENSE
```

recipes/tir-learner4/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+source:
+  url: https://github.com/KGerhardt/TIR-Learner/archive/refs/tags/v4.0.1.tar.gz
+  sha256: f36828693f1c3bbf4f7c2441b7ba8d64d8a42653d0cb3738e52b0c365444e2e8
+
+build:
+  number: 3
+  skip: True  # [osx]
+
+requirements:
+  build:
```
thanhleviet, line 21: **Unnecessary build/host dependencies (important):** The `build.sh` just copies files and creates a wrapper script — there's no C/C++ compilation, no `pip install`, and no Cython build step. The compilers, `cython`, `pip`, `setuptools`, and `wheel` in host are all unnecessary.

Since this is a pure file-copy install (no compiled code), this could be `noarch: generic` — just like the existing `tir-learner` recipe. That would simplify the recipe significantly:

```yaml
build:
  number: 0
  noarch: generic
  run_exports:
    - {{ pin_subpackage('tir-learner4', max_pin='x') }}

requirements:
  run:
    - python >=3.9,<3.13
    - numpy
    - pyfastx
    - pywfa
    - keras
    - pytorch
    - genometools-genometools
    - genericrepeatfinder
    - blast
    - pigz
```

If there *is* a compilation step I'm missing, please clarify.

recipes/tir-learner4/meta.yaml
```diff
@@ -0,0 +1,47 @@
+package:
```
thanhleviet, line 5: **Missing Jinja variables (nice to have):** Following bioconda conventions, use `{% set %}` variables at the top:

```yaml
{% set name = "tir-learner4" %}
{% set version = "4.0.1" %}
{% set sha256 = "f36828693f1c3bbf4f7c2441b7ba8d64d8a42653d0cb3738e52b0c365444e2e8" %}

package:
  name: {{ name }}
  version: {{ version }}

source:
  url: https://github.com/KGerhardt/TIR-Learner/archive/refs/tags/v{{ version }}.tar.gz
  sha256: {{ sha256 }}
```

recipes/tir-learner4/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - pytorch
+    - genometools-genometools
+    - genericrepeatfinder
+    - blast
+    - pigz
+
+test:
+  commands:
+    - tirlearner4 --help
+
```
thanhleviet, line 53: **Missing `dev_url` / `doc_url` (nice to have):**

```suggestion
  summary: "Accelerated tool for identifying TIR transposable elements"
  dev_url: https://github.com/KGerhardt/TIR-Learner
  doc_url: https://github.com/KGerhardt/TIR-Learner/blob/main/README.md
```

recipes/tir-learner4/build.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+set -euo pipefail
+
+# Install package files
+mkdir -p "$PREFIX/lib/tir-learner4"
+cp -r TIR-Learner3/* "$PREFIX/lib/tir-learner4/"
+
```
thanhleviet, line 7: **Hardcoded source directory name (important):** The source archive extracts to `TIR-Learner-4.0.1/TIR-Learner3/`. The inner directory is called `TIR-Learner3` even though this is v4. This works, but is worth a comment explaining it — or better yet, the upstream repo could rename the directory.

Also, the `readlink -f` on line 14 is a GNU extension that doesn't work on macOS. Since you skip osx builds this is fine, but if you ever re-enable osx you'll need to handle this (e.g., use `$PREFIX/lib/tir-learner4` directly instead of resolving symlinks).

Juke34:
I would remove change from pywfa here, that make difficult to follow. One PR one recipe changed. 
Please remove change from pywfa from here.

oushujun:
@thanhleviet Thanks a ton for the detailed walk-through and fixing of the recipe. 

@KGerhardt I suggest changing the repo name from tir-learner4 back to tir-learner, and changing the version name to v4.0.1. Folder names in the repo should also be consistent so future versions don't get confused.

Shujun

PR #63727, opened 2026-03-20
Add gwaslab

lyh970817:
Add Bioconda recipe for `gwaslab`, a Python toolkit for handling GWAS summary statistics.

This package is directly relevant to genomics workflows.

The recipe includes `run_exports` because `gwaslab` is intended for reuse as a downstream dependency.

The recipe also carries a small packaging patch to defer the optional `sumstats-liftover` import so `import gwaslab` and `gwaslab --help` work in standalone recipe CI without requiring an unpublished dependency.

### DIFF
--- a/recipes/gwaslab/0001-defer-sumstats-liftover-import.patch
+++ b/recipes/gwaslab/0001-defer-sumstats-liftover-import.patch
@@ -1,0 +1,31 @@
+--- a/src/gwaslab/hm/hm_liftover_v2.py
++++ b/src/gwaslab/hm/hm_liftover_v2.py
+@@ -16,9 +16,15 @@
+ try:
+     from sumstats_liftover import liftover_df
+ except ImportError:
+-    raise ImportError(
+-        "sumstats_liftover package is required. Install it with: pip install sumstats-liftover"
+-    )
++    liftover_df = None
++
++
++def _get_liftover_df():
++    if liftover_df is None:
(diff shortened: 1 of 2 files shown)


recipes/gwaslab/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+  sha256: 618424c34745b0bec4131056eb73166acbc262b3ffb1cada239212465c154408
+
+build:
+  number: 0
+  noarch: python
+  run_exports:
+    - {{ pin_subpackage(name, max_pin="x") }}
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+
+requirements:
```
tseemann: Do you need `build:` section?

recipes/gwaslab/0001-defer-sumstats-liftover-import.patch
```diff
@@ -0,0 +1,31 @@
+--- a/src/gwaslab/hm/hm_liftover_v2.py
++++ b/src/gwaslab/hm/hm_liftover_v2.py
+@@ -16,9 +16,15 @@
+ try:
+     from sumstats_liftover import liftover_df
+ except ImportError:
+-    raise ImportError(
```
tseemann: Can these pathces be made in the upstream package?

mencian (approved):

PR #63738, opened 2026-03-20
Update Gblocks version and source URLs

m-nikolaidis:
Updates:

- Added newest version of gblocks
- Updated URLs
- Added osx-arm64 and linux-64 as additional platforms


m-nikolaidis:
@BiocondaBot please add label

thanhleviet (commented):
sha256 verified for both tarballs, URLs resolve over HTTPS, build number reset correctly. One critical issue with the additional-platforms declaration.

recipes/gblocks/meta.yaml
```diff
@@ -22,13 +22,16 @@ test:
     - which Gblocks
 
 about:
-  home: http://molevol.cmima.csic.es/castresana/Gblocks.html
+  home: https://molevol-ibe.csic.es/Gblocks.html
```
thanhleviet, line 27: The Linux binary is **x86-64 only** (`ELF 64-bit LSB pie executable, x86-64`). There is no aarch64 Linux binary available upstream -- the [download page](https://molevol-ibe.csic.es/Gblocks.html) only offers one Linux build. This will produce a broken package on ARM Linux.

Please remove `linux-aarch64`:

```suggestion
  additional-platforms:
    - osx-arm64
```

`osx-arm64` is fine -- the macOS binary is a universal binary containing both x86_64 and arm64 slices.

recipes/gblocks/meta.yaml
```diff
@@ -1,17 +1,17 @@
 package:
   name: gblocks
-  version: "0.91b"
+  version: "1.0"
```
thanhleviet, line 4: FYI: This tarball is double-gzipped (gzip wrapping another gzip wrapping a tar). conda-build typically expects a single gzip layer. CI is still pending so we'll see if it handles it, but if the build fails to extract, this could be the cause.

thanhleviet:
Lint is now failing with `missing_run_exports`. The recipe needs a `run_exports` section. Since this is semver 1.x, add:

```yaml
build:
  binary_relocation: False
  number: 0
  run_exports:
    - {{ pin_subpackage("gblocks", max_pin="x") }}
```

mencian (approved):

PR #63754, opened 2026-03-22
Add r-tidygwas

lyh970817:

----

This PR adds `r-tidygwas` 1.0.0, an R package for GWAS summary-statistics quality control and harmonization.

Why it belongs in Bioconda:
- It provides quality-control and harmonization utilities for GWAS summary statistics used in statistical genetics workflows.

Dependency note:
- `r-tidygwas` depends on `r-ldsr`, which is already available in Bioconda.

Packaging notes:
- `noarch: generic`
- `run_exports: {{ pin_subpackage("r-tidygwas", max_pin="x.x") }}`
- `ignore_run_exports_from: r-ldsr` with an explicit runtime pin `>=0.4.0,<0.5.0a0`
- Pinned to the supported `r_base 4.4.*` build variant to match `r-ldsr`.
- Source: https://github.com/Ararder/tidyGWAS
- License: MIT


lyh970817:
@BiocondaBot please add label

lyh970817:
@thanhleviet Hi this PR is ready for review with checks passing. Could you please have a look? Thank you!

recipes/r-tidygwas/meta.yaml
```diff
@@ -0,0 +1,92 @@
...
+    - r-ldsr >={{ ldsr_version }},<0.5.0a0
+
+test:
+  commands:
+    - $R -e "library('tidyGWAS')"
+
+about:
+  home: https://ararder.github.io/tidyGWAS/
+  license: MIT
+  license_file: LICENSE
```
thanhleviet: The `LICENSE` file in this package is only R's stub (`YEAR` + `COPYRIGHT HOLDER`) and contains no license text — the actual MIT terms live in `LICENSE.md`. To bundle the real license text into the package, point `license_file` at both files:

```suggestion
  license_file:
    - LICENSE
    - LICENSE.md
```

recipes/r-tidygwas/meta.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+source:
+  url: https://github.com/{{ github_user }}/{{ github_repo }}/archive/{{ github_commit }}.tar.gz
+  sha256: b76ecea03525c6b28ea02b0a37881cbfd312e2596c08daea63afe652fd61b423
+
+build:
+  number: 0
+  noarch: generic
+  run_exports:
+    - {{ pin_subpackage('r-tidygwas', max_pin="x.x") }}
```
thanhleviet: Version is `1.0.0` (semver 1.x), so the bioconda convention is `max_pin="x"` — downstream gets `>=1.0.0,<2`. `max_pin="x.x"` pins downstream to `<1.1`, which is unnecessarily tight and forces rebuilds on every minor release. (If you expect tidyGWAS to break API across minor versions, the tight pin is defensible — your call.)

```suggestion
    - {{ pin_subpackage('r-tidygwas', max_pin="x") }}
```

thanhleviet (commented):
Thanks for the recipe — this is well put together. CI is green, all 21 upstream `Imports` map correctly to their `r-*` packages, and the `ignore_run_exports_from: r-ldsr` + explicit pin is a clean way to override the dependency's run_exports. The bare-commit source pin is the right call here since upstream has no tags or releases.

Two non-blocking should-fixes inline: the `license_file` currently ships only R's stub (no MIT terms), and the `run_exports` pin is tighter than the convention for a 1.0.0 release.

recipes/r-tidygwas/meta.yaml
lyh970817: Fixed in 509afaf8c1 by listing both `LICENSE` and `LICENSE.md` under `license_file`, so the packaged recipe includes the R license stub and the full MIT terms.

recipes/r-tidygwas/meta.yaml
lyh970817: Fixed in 2c478efb18. I agree that `max_pin="x"` is the better default for the 1.0.0 release: it keeps downstream packages compatible across non-breaking 1.x updates, while still requiring review/rebuilds for a future 2.x API break.

thanhleviet (approved):
Both points addressed and CI is green on the fresh run — `license_file` now bundles the full MIT text alongside the R stub, and the run_export pin is `max_pin="x"`. Thanks for the quick turnaround! LGTM.

PR #64183, opened 2026-04-06
New recipe for TIR-Learner4, a tool for identifying TIR transposable

KGerhardt:
Bioconda package try number 2 for TIR-Learner v4. Incorporated fixes suggested from pull 63598 as well as some code improvements. I've closed the old pull.

The code publishes a new TIR-Learner package separately from TIR-Learner v3. The v3 program is still in review for publication and we don't want to overwrite or interrupt it, but do want to switch to this version of TIR-Learner for our own use.

@thanhleviet

@oushujun 

@Juke34

### DIFF
--- a/recipes/tir-learner4/build.sh
+++ b/recipes/tir-learner4/build.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+set -euo pipefail
+
+# Install package files
+mkdir -p "$PREFIX/lib/tir-learner4"
+cp -r TIR-Learner4/* "$PREFIX/lib/tir-learner4/"
+
+# Fix shebang in main script (original has #!/usr/app/env python3, a typo)
+sed -i '1s|.*|#!/usr/bin/env python3|' "$PREFIX/lib/tir-learner4/TIR-Learner.py"
+
+# Create CLI wrapper
+mkdir -p "$PREFIX/bin"
+cat > "$PREFIX/bin/tirlearner4" << 'EOF'
+#!/bin/bash
+exec python3 "$(dirname "$(dirname "$(readlink -f "$0")")")/lib/tir-learner4/TIR-Learner.py" "$@"
+EOF
+chmod +x "$PREFIX/bin/tirlearner4"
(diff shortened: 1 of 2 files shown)


recipes/tir-learner4/build.sh
```diff
@@ -0,0 +1,17 @@
...
+mkdir -p "$PREFIX/lib/tir-learner4"
+cp -r TIR-Learner4/* "$PREFIX/lib/tir-learner4/"
+
+# Fix shebang in main script (original has #!/usr/app/env python3, a typo)
+sed -i '1s|.*|#!/usr/bin/env python3|' "$PREFIX/lib/tir-learner4/TIR-Learner.py"
+
+# Create CLI wrapper
+mkdir -p "$PREFIX/bin"
+cat > "$PREFIX/bin/tirlearner4" << 'EOF'
+#!/bin/bash
```
tseemann: Use `#!/usr/bin/env bash`

recipes/tir-learner4/build.sh
```diff
@@ -0,0 +1,17 @@
...
```
tseemann: Use `#!/usr/bin/env bash`

recipes/tir-learner4/meta.yaml
```diff
@@ -0,0 +1,43 @@
+{% set name = "tir-learner4" %}
```
tseemann: This should maybe be `TIR-Learner`
Then you can use `{{ name }}` in all your URLs;
But use `{{ name | lower ||` for the package name.

tseemann (changes requested):

recipes/tir-learner4/build.sh
KGerhardt: Fixed

recipes/tir-learner4/build.sh
KGerhardt: Fixed.

recipes/tir-learner4/meta.yaml
KGerhardt: That'll cause a conflict with the old TIR-Learner, I think, but we can try.

PR #64256, opened 2026-04-08
Add metastrand

mmpust:
### Add metastrand

This PR adds a new Bioconda recipe for **metastrand**, a pipeline for analyzing sense (sMTX) and antisense (asMTX) transcriptional signals in metagenomic/metatranscriptomic data.

- Source: https://gitlab.com/mpust/metastrand
- License: MIT
- Language: Python + shell scripts

### Functionality
Metastrand provides a workflow for:
- mapping MTX/MGX reads
- generating count matrices
- analyzing antisense transcription patterns

### Notes
- This is a wrapper-style package (no compilation required)
- Tested via `metastrand --help` and bundled config templates
- Dependencies are available on conda-forge/bioconda

### Checklist
- [x] Recipe builds locally
- [x] Tests pass locally
- [x] Source tarball is versioned and reproducible

### DIFF
--- a/recipes/metastrand/meta.yaml
+++ b/recipes/metastrand/meta.yaml
@@ -1,0 +1,42 @@
+package:
+  name: metastrand
+  version: "0.1.1"
+
+source:
+  url: https://gitlab.com/mpust/metastrand/-/archive/v0.1.1/metastrand-v0.1.1.tar.gz
+  sha256: edb49260149298c4f6a64d6a003bb5d506b7844c797af459cbcfebbe6d7aa29c
+
+build:
+  number: 0
+
+requirements:
+  run:
+    - python
+    - pandas
+    - pysam
+    - numpy
+    - scipy
+    - bwa
+    - samtools=1.23
+    - subread=2.1.1
+    - htslib=1.23
+    - bedtools=2.31.1
+    - sambamba=1.0.1
+    - bbmap=39.52
+
+about:
+  home: "https://gitlab.com/mpust/metastrand"
+  license: MIT
+  license_file: LICENSE
+  summary: "Metastrand: Making sense of antisense"
+
+extra:
+  recipe-maintainers:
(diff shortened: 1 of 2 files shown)


mmpust:
@BiocondaBot please add label

thanhleviet (commented):
CI passes, sha256 verified. Two items to address.

recipes/metastrand/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+build:
+  number: 0
+  noarch: generic
+  run_exports:
+    - {{ pin_subpackage('metastrand', max_pin="x.x") }}
+
+requirements:
+  run:
+    - python
```
thanhleviet: `python` has no version constraint. The upstream environment uses Python 3.13, and the scripts use modern Python features. Consider adding a lower bound:

```suggestion
    - python >=3.9
```

recipes/metastrand/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - pandas
+    - pysam
+    - numpy
+    - scipy
+    - bwa
+    - samtools =1.23
```
thanhleviet: All tool dependencies use exact `=` pins (e.g., `samtools =1.23`). While this matches the upstream `environment.yml`, exact pins in conda make the package hard to co-install with other bioconda tools that may need different versions. Since metastrand calls these tools via CLI (not linking against their libraries), relaxing to `>=` lower bounds would improve installability without affecting functionality.

mmpust: Thanks, addressed both points.
I updated the recipe to:
	•	set python >=3.9
	•	relax CLI tool dependencies from exact = pins to >= lower bounds for better co-installability

kbseah (approved):
issues raised in previous review have been addressed. looks good to me!

PR #64306, opened 2026-04-09
Add recipe for TIR-Learner

KGerhardt:
## Summary
- New bioconda recipe for TIR-Learner, an accelerated tool for identifying TIR transposable elements
- Incorporates reviewer feedback from #64183 (shebang fixes, name/URL cleanup)

@thanhleviet @oushujun @Juke34

### DIFF
--- a/recipes/tir-learner/build.sh
+++ b/recipes/tir-learner/build.sh
@@ -1,15 +1,17 @@
-#!/bin/bash
-set -x -e
+#!/usr/bin/env bash
+set -euo pipefail
 
-TIR_LEARNER_DIR="${PREFIX}/share/TIR-Learner3"
+# Install package files
+mkdir -p "$PREFIX/lib/tir-learner4"
+cp -r TIR-Learner4/* "$PREFIX/lib/tir-learner4/"
 
-mkdir -p ${PREFIX}/bin
-mkdir -p ${TIR_LEARNER_DIR}
-cp -rf TIR-Learner3/* ${TIR_LEARNER_DIR}
+# Fix shebang in main script (original has #!/usr/app/env python3, a typo)
+sed -i '1s|.*|#!/usr/bin/env python3|' "$PREFIX/lib/tir-learner4/TIR-Learner.py"
 
-cat <<END >>${PREFIX}/bin/TIR-Learner
-#!/bin/bash
-python3 ${TIR_LEARNER_DIR}/TIR-Learner.py \$@
-END
-
-chmod 0755 ${PREFIX}/bin/TIR-Learner
+# Create CLI wrapper
+mkdir -p "$PREFIX/bin"
+cat > "$PREFIX/bin/tirlearner4" << 'EOF'
+#!/usr/bin/env bash
+exec python3 "$(dirname "$(dirname "$(readlink -f "$0")")")/lib/tir-learner4/TIR-Learner.py" "$@"
+EOF
+chmod +x "$PREFIX/bin/tirlearner4"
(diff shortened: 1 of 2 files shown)


thanhleviet (commented):
CI passes, sha256 verified, license correct, dependencies match upstream imports. Two items to discuss.

recipes/tir-learner/meta.yaml
```diff
@@ -1,49 +1,43 @@
 {% set name = "TIR-Learner" %}
-{% set version = "3.0.7" %}
-{% set sha256 = "5c574aee3f21aa17a4cec0026282ed6671c13e6d9f9fd89963eb2be8dfde2cb6" %}
+{% set version = "1.0" %}
```
thanhleviet: The version goes from `3.0.7` (current in bioconda from lutianyu2001's repo) down to `1.0`. Conda considers `1.0 < 3.0.7`, so users with the existing package won't receive this as an upgrade via `conda update tir-learner`.

Since this is effectively "TIR-Learner 4" (the code lives in `TIR-Learner4/` and `VERSION = "v1.0"`), would it make sense to use version `4.1.0` to reflect the lineage and ensure upgradeability? Or if `1.0` is the intended upstream version, you could add `epoch: 1` in the build section to force conda to treat it as newer.

recipes/tir-learner/build.sh
```diff
@@ -1,15 +1,17 @@
...
+
+# Fix shebang in main script (original has #!/usr/app/env python3, a typo)
+sed -i '1s|.*|#!/usr/bin/env python3|' "$PREFIX/lib/tir-learner4/TIR-Learner.py"
+
+# Create CLI wrapper
+mkdir -p "$PREFIX/bin"
+cat > "$PREFIX/bin/tirlearner4" << 'EOF'
+#!/usr/bin/env bash
+exec python3 "$(dirname "$(dirname "$(readlink -f "$0")")")/lib/tir-learner4/TIR-Learner.py" "$@"
+EOF
```
thanhleviet: `readlink -f` is GNU-specific and may not work on older macOS (pre-Ventura). Since `$PREFIX` is known at build time, the standard conda pattern is to expand it during installation using a non-quoted heredoc:

```suggestion
cat > "$PREFIX/bin/tirlearner4" << EOF
#!/usr/bin/env bash
exec python3 "$PREFIX/lib/tir-learner4/TIR-Learner.py" "\$@"
EOF
```

This bakes the absolute install path into the wrapper, avoiding any runtime symlink resolution.

recipes/tir-learner/meta.yaml
KGerhardt: We can up-version it to 4.1.0 since it's replacing TIR-Learner 3 now. That should be fine.

recipes/tir-learner/build.sh
KGerhardt: Got it.

recipes/tir-learner/meta.yaml
oushujun: It should be 4.0.2 up from 4.0.1. The latest release used a wrong label (v1.0): https://github.com/KGerhardt/TIR-Learner/releases

oushujun: @KGerhardt

KGerhardt:
@thanhleviet @Juke34

I think it's all ready.

thanhleviet (approved):
Verified sha256 matches upstream tarball. CI passes on Linux, OSX-64, and ARM. Build.sh improvements from earlier review look clean. Upgrade path 3.0.7 → 4.02 works correctly in conda version ordering.

One heads-up: the tag is `v4.02` but conda parses this as `4.2.0` (stripping the leading zero in the minor component). A future `v4.1.0` release would rank *below* this version. If you cut a patch-level release soon, consider re-tagging as `v4.0.2` upstream (or mapping version to `4.0.2` in the recipe with a manual source URL) to avoid a future regression. Not blocking.

Thanks!

PR #64404, opened 2026-04-13
Csem recipe bug fix

descostesn:
**[build.sh]** — Fix the Perl module location

Added csem_perl_utils.pm to the cp command that installs binaries into $PREFIX/bin/, so the module lives next to the scripts (where use lib $FindBin::Bin looks for it)
Removed the separate "Install Perl module to site lib" block that was copying it to the wrong location ($Config{installsitelib})

**[meta.yaml]** — Improve tests and add dependency
Added perl-perldoc as a runtime dependency
Replaced the unreliable run-csem --help 2>&1 | grep -qi "csem" test (which passed even on crashes) with individual test -f checks for all 5 binaries: run-csem, csem, csem-bam2wig, csem-bam-processor, csem-generate-input

**[run_test.sh]** — New file, detailed functional tests


descostesn:
Dear @thanhleviet,

I think that the package should work now and can be merged. I tried the suggestion of copilot below but I fall back on the "not found error":

```
- string: h47932c3_{{ PKG_BUILDNUM }}
+ string: h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}
```

The following modifications have been made:


**[build.sh]** — Fix the Perl module location

Added csem_perl_utils.pm to the cp command that installs binaries into $PREFIX/bin/, so the module lives next to the scripts (where use lib $FindBin::Bin looks for it)
Removed the separate "Install Perl module to site lib" block that was copying it to the wrong location ($Config{installsitelib})

**[meta.yaml]** — Improve tests and add dependency

Added perl-perldoc as a runtime dependency
Replaced the unreliable run-csem --help 2>&1 | grep -qi "csem" test (which passed even on crashes) with individual test -f checks for all 5 binaries: run-csem, csem, csem-bam2wig, csem-bam-processor, csem-generate-input

**[run_test.sh]** — New file, detailed functional tests.


Thank you.

Nicolas

recipes/csem/build.sh
```diff
@@ -38,15 +38,14 @@ make
 
 # --- Install binaries and Perl module ---
 mkdir -p "${PREFIX}/bin"
-cp csem csem-bam2wig csem-bam-processor csem-generate-input run-csem "${PREFIX}/bin/"
+cp csem csem-bam2wig csem-bam-processor csem-generate-input run-csem csem_perl_utils.pm "${PREFIX}/bin/"
+
+# --- Fix Perl shebangs to use conda environment perl ---
+sed -i "1s|#!/usr/bin/perl|#!${PREFIX}/bin/perl|" "${PREFIX}/bin/run-csem"
+sed -i "1s|#!/usr/bin/perl|#!${PREFIX}/bin/perl|" "${PREFIX}/bin/csem-generate-input"
```
thanhleviet: Optional: consider `#!/usr/bin/env perl` instead of hardcoding the conda prefix. Prefix replacement makes the current form work, but `env perl` is shorter, avoids shebang length limits on deep env paths, and works if someone copies the script outside the env.

```suggestion
# --- Fix Perl shebangs to use env-based lookup ---
sed -i "1s|#!/usr/bin/perl|#!/usr/bin/env perl|" "${PREFIX}/bin/run-csem"
sed -i "1s|#!/usr/bin/perl|#!/usr/bin/env perl|" "${PREFIX}/bin/csem-generate-input"
```

recipes/csem/run_test.sh
```diff
@@ -0,0 +1,44 @@
...
+    exit 1
+fi
+
+if ! { run-csem --help 2>&1 || true; } | grep -q "run-csem \\[options\\] input_file fragment_length output_name"; then
+    echo "FAIL: run-csem --help did not print expected usage string"
+    exit 1
+fi
+
+# Test 2: csem prints expected usage
+ if ! { csem 2>&1 || true; } | grep -qE "Usage[[:space:]:]+csem([[:space:]]|$)"; then
```
thanhleviet: Stray leading space on this line — the rest of the file has no indent at the `if` level.

```suggestion
if ! { csem 2>&1 || true; } | grep -qE "Usage[[:space:]:]+csem([[:space:]]|$)"; then
```

thanhleviet (approved):
Nice cleanup. The Perl module colocation fix plus the functional tests in `run_test.sh` are a clear improvement over the previous `grep -qi "csem"` check, which would have passed on a crashed binary. CI is green across Linux/OSX/ARM.

Left two optional nits — neither is a blocker.

mencian (approved):

PR #64406, opened 2026-04-13
Add mutopia

AllenWLynch:
Add the "MuTopia" package to bioconda: a python package for model-based analysis of mutational topography and mutational processes.


AllenWLynch:
@BiocondaBot please add label

tseemann (approved):
1. Can `bcftool index` replace `tabix` ?

2. Does `bcftools` already depend on `htslib` ?
3.

PR #64434, opened 2026-04-14
Add ohe-rs: ultra-fast one-hot encoding for bioinformatics

andreaderuvo:
## Summary

Add [ohe-rs](https://github.com/genpat-it/ohe-rs), an ultra-fast one-hot encoding library for bioinformatics and ML.

- **Language:** Rust core with Python bindings (PyO3/maturin)
- **Performance:** 15-35x faster than scikit-learn OneHotEncoder
- **Use case:** cgMLST allele profiles, large-scale categorical data
- **Features:** parallel encoding (rayon), sparse CSR output, multi-column encoding, optional CUDA GPU support

## Tests

```python
from ohe_rs import encode_sparse
import numpy as np
v, i, p, k = encode_sparse(np.array([0, 1, 2, 0], dtype=np.int64))
assert k == 3
```

### DIFF
--- a/recipes/ohe-rs/meta.yaml
+++ b/recipes/ohe-rs/meta.yaml
@@ -1,0 +1,58 @@
+{% set name = "ohe-rs" %}
+{% set version = "0.0.1" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/genpat-it/ohe-rs/archive/v{{ version }}.tar.gz
+  sha256: 9e06ab483d636a69dbc99f4def3cb22d0b0bc36f430499339e85275aff0f2a48
+
+build:
+  number: 0
+  script: |
+    maturin build --release --interpreter ${PYTHON}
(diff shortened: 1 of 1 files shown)


andreaderuvo:
Hi bioconda team,

I'm struggling with a persistent hash mismatch between the Docker build container and the host verification step. The build succeeds inside Docker (all 4 Python versions produce `.conda` files), but the host looks for a different hash.

Example:
- Docker produces: `ohe-rs-0.1.0-py311h78a2e26_0.conda`
- Host expects: `ohe-rs-0.1.0-py311ha295b23_0.conda`

I've tried:
- Removing/adding `compiler('c')` and `stdlib('c')`
- Removing `numpy` from host deps
- Different `run_exports` configurations
- Following the `kcounter` and `gb-io` patterns exactly

ARM and Lint pass consistently — only Linux x86_64 fails with "the built package cannot be found".

This is a Rust+PyO3 package built with maturin. Any guidance on what causes the hash divergence between Docker and host for this type of package?

Thanks!

mencian (approved):

PR #64488, opened 2026-04-15
Add baysor

quentinblampey:
Adding recipe for [Baysor](https://github.com/kharchenkolab/Baysor)
Based on a discussion with the nf-core developers (see slack channel #[request-core](https://nfcore.slack.com/archives/C09H6NYHR9T/p1776088219167789))

FYI @VPetukhov - let us know if you are interested in this recipe, and if you have ideas to help us fix the CI!


quentinblampey:
@BiocondaBot please add label

recipes/baysor/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  url: https://github.com/kharchenkolab/Baysor/releases/download/v{{ version }}/baysor-x86_x64-linux-v{{ version }}_build.zip
+  sha256: f37f02695a068fd360ade7abd0f91b07c5199a3cce6e91ded550689459fbe5c0
+
+build:
+  number: 0
+  skip: True  # [not linux]
+  platforms:
+    - linux-64
+  run_exports:
+    - {{ pin_subpackage('baysor', max_pin="x") }}
```
ewels: Pre v1 so should be `x.x`
```suggestion
    - {{ pin_subpackage('baysor', max_pin="x.x") }}
```

recipes/baysor/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/kharchenkolab/Baysor/releases/download/v{{ version }}/baysor-x86_x64-linux-v{{ version }}_build.zip
+  sha256: f37f02695a068fd360ade7abd0f91b07c5199a3cce6e91ded550689459fbe5c0
+
+build:
+  number: 0
+  skip: True  # [not linux]
```
ewels: I don't really understand the `# [not linux]` here - we _do_ want to skip the build on linux, right? As we have the pre-built binaries? And we're only working with `linux-64` here anyway..

quentinblampey: I wanted to say that the recipe is only for linux, so I wanted to skip it for the other OS
But maybe it's not the right command to do that, I can remove it if you think it has to be removed

recipes/baysor/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/kharchenkolab/Baysor/releases/download/v{{ version }}/baysor-x86_x64-linux-v{{ version }}_build.zip
+  sha256: f37f02695a068fd360ade7abd0f91b07c5199a3cce6e91ded550689459fbe5c0
+
+build:
+  number: 0
+  skip: True
```
bgruening: this is not building the recipe at all?

quentinblampey: Yes, I tried multiple options based on the comments of @ewels, but it seems the first version was the only one to work (i.e., `skip: True # [not linux]`). What do you think @ewels?

quentinblampey: I added `# [not linux]` again

bgruening:
@quentinblampey is there any reason you are not compiling here on bioconda. We encourage all our contributors to compile from source and not ship magic binaries.

ewels:
@quentinblampey I pushed a commit which specifies which gcc to use (the one in the host image is very old). I think that's probably the issue you were hitting.

Unfortunately the build is now timing out / hitting an OOM (it has >300 dependencies 🤯 ). I've tried restarting it in the hope it might work on a second go. But not sure where to go from here if it doesn't work.

quentinblampey:
> Unfortunately the build is now timing out / hitting an OOM (it has >300 dependencies 🤯 ). I've tried restarting it in the hope it might work on a second go. But not sure where to go from here if it doesn't work.

Thanks @ewels for trying it! It looks like the second run also failed. Is it possible to increase the memory/time for the action?
Else, I guess the only solution would be to go back to the binary installation (which was apparently working, see [this action](https://github.com/bioconda/bioconda-recipes/actions/runs/24551735443/job/71778962016))?

> is there any reason you are not compiling here on bioconda. We encourage all our contributors to compile from source and not ship magic binaries.

@bgruening it's because the author of Baysor suggests the binary installation as a preferred/default installation, but I get your point!

ewels:
@quentinblampey - I don't know what else we can do to get it to compile here sorry.

Unless @bioconda/build-system folks have any suggestions, I suggest that we revert back to 42aeca2 just to get it released.

quentinblampey:
Closing as #65052 will be merged instead

PR #64549, opened 2026-04-16
Fixed Staphscope path issue

bbeckley-hub:
@thanhleviet Thank you for merging the recipe. However, after testing the installed package, we discovered that the MLST and SCCmec databases are being installed into $PREFIX/share/staphscope/modules/, while the staphscope orchestrator looks for them in $PREFIX/lib/python*/site-packages/staphscope/modules/. This causes the MLST and SCCmec modules to fail because they cannot find their data.

To fix this, we have updated the build.sh scripts for the two data packages (staphscope-mlst-data and staphscope-sccmec-data) to install directly into $SP_DIR/staphscope/modules/ (the site-packages directory). The version remains 1.2.0, but the build number is bumped to 1.

We have also removed the redundant entry_points block from the main recipe as you suggested.

This new PR contains the corrected recipes. All tests pass locally. Please review and merge.


bbeckley-hub:
@BiocondaBot please add label

PR #64551, opened 2026-04-16
Fix database installation path: use SP_DIR, bump build number to 1

bbeckley-hub:
@thanhleviet Thank you for merging the recipe. However, after testing the installed package, we discovered that the MLST and SCCmec databases are being installed into $PREFIX/share/staphscope/modules/, while the staphscope orchestrator looks for them in $PREFIX/lib/python*/site-packages/staphscope/modules/. This causes the MLST and SCCmec modules to fail because they cannot find their data.  To fix this, we have updated the build.sh scripts for the two data packages (staphscope-mlst-data and staphscope-sccmec-data) to install directly into $SP_DIR/staphscope/modules/ (the site-packages directory). The version remains 1.2.0, but the build number is bumped to 1.  We have also removed the redundant entry_points block from the main recipe as you suggested.  This new PR contains the corrected recipes. All tests pass locally. Please review and merge.


bbeckley-hub:

bbeckley-hub:

bbeckley-hub:
@thanhleviet, after a few "adventures" with the database path (think of it as our package taking the scenic route through $PREFIX/share instead of the express lane to site-packages), we finally got everything back on track. The build now places the MLST and SCCmec databases exactly where the orchestrator expects them – no more treasure hunts for missing files.

Thanks for your patience and for the earlier suggestion to remove the redundant entry_points – that’s also included.

This PR fixes the issue and bumps the build numbers. All tests are green. Please review when you have a moment. And yes, we kind of made it back again. 😅

recipes/staphscope-mlst-data/build.sh
```diff
@@ -1,9 +1,20 @@
...
+    cp -rv "$SRC_DIR/mlst_module/." "$DEST/"
+else
+    echo "=== No mlst_module directory, copying all contents"
+    cp -rv "$SRC_DIR/." "$DEST/"
+fi
 
-# Copy all contents from SRC_DIR (these are the mlst_module files) into DEST
-cp -r $SRC_DIR/* "$DEST/"
+echo "=== Debug: Final contents of $DEST"
+ls -la "$DEST"
```
thanhleviet, line 20: Optional cleanup (for a future bump, not a blocker for this PR): the `echo "=== Debug: ..."` lines, the two `ls -la` statements, and the `-v` flag on `cp -rv` look like leftover iteration artifacts. They'll show up in every conda-build log for users who build from source. Same applies to `recipes/staphscope-sccmec-data/build.sh`. A minimal form would be:

```bash
DEST="$SP_DIR/staphscope/modules/mlst_module"
mkdir -p "$DEST"

if [ -d "$SRC_DIR/mlst_module" ]; then
    cp -r "$SRC_DIR/mlst_module/." "$DEST/"
else
    cp -r "$SRC_DIR/." "$DEST/"
fi
```

recipes/staphscope-mlst-data/build.sh
```diff
@@ -1,9 +1,20 @@
 #!/bin/bash
 set -e
 
-# Destination directory for mlst_module
-DEST="$PREFIX/share/staphscope/modules/mlst_module"
+echo "=== Debug: SRC_DIR = $SRC_DIR"
+ls -la $SRC_DIR
+
+DEST="$SP_DIR/staphscope/modules/mlst_module"
```
thanhleviet, line 7: Heads-up for a follow-up (not blocking): `$SP_DIR` resolves to `$PREFIX/lib/pythonX.Y/site-packages` at **build time**, based on the build env's Python. Because this package is `noarch: generic`, that Python version is baked into the tarball's file paths. `staphscope` pins `python >=3.8,<3.13`, so a user on Python 3.8 would see the data at `lib/python3.12/site-packages/…` instead of `lib/python3.8/site-packages/…`.

The idiomatic fix is to make the data packages `noarch: python` with `python` as a run dep — then conda resolves `$SP_DIR` at install time. Worth considering for a future rebuild. Same concern applies to `staphscope-sccmec-data`.

thanhleviet (approved):
Thanks for iterating on this — the database paths are now where the orchestrator expects, and removing the redundant `entry_points` matches the earlier suggestion. CI green across Linux/OSX/ARM.

Two non-blocking notes below: debug output cleanup, and a heads-up on a latent `noarch: generic` + `$SP_DIR` gotcha for a future follow-up.

recipes/staphscope-mlst-data/build.sh
bbeckley-hub: @thanhleviet, thank you for the thorough review and for catching those debug leftovers – I guess my build scripts wanted to leave a breadcrumb trail in case the databases got lost again. 😄

I'll clean up the echo and ls lines in a future follow‑up (along with the noarch: generic nuance you pointed out). And yes, that idea about noarch: python for data packages is gold – I'll definitely carry it over to one of my ESKAPE tools. Consider it officially adopted.

Thanks again for the patience and the green light. The databases are finally where they belong, and the entry_points are enjoying their retirement.

PR #64565, opened 2026-04-17
Add cuteSV-Trio v1.0.0

XLeeHIT:
This PR adds a new recipe for cuteSV-Trio(v1.0.0), a Haplotype-Resolved and De Novo Structural Variant Detection from Low-Coverage Long-Read Family Trios data

### DIFF
--- a/recipes/cutesv-trio/meta.yaml
+++ b/recipes/cutesv-trio/meta.yaml
@@ -1,0 +1,47 @@
+{% set version = "1.0.0" %}
+
+package:
+  name: cutesv-trio
+  version: {{ version }}
+
+source:
+  url: https://github.com/XLeeHIT/cuteSV-Trio/archive/refs/tags/v1.0.0.tar.gz
+  sha256: 56acf0b09c6ff3637099ec9a2e3abf2d3a7855ab73054a36c7b47fd325e526da
+
+build:
+  number: 0
+  noarch: python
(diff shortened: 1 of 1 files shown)


XLeeHIT:

recipes/cutesv-trio/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - pip
+    - python >=3
+    - setuptools
+  run:
+    - biopython
+    - cigar
+    - numpy
+    - pysam >=0.15.0
+    - python >=3
+    - pyvcf
```
thanhleviet, line 29: Upstream's `setup.py` declares `pyvcf3`, not `pyvcf`. These are separate conda packages:
- `pyvcf` (bioconda 0.6.8) is the legacy package, unmaintained since 2017 and broken on Python 3.10+
- `pyvcf3` (bioconda 1.0.4) is the Python 3-compatible fork that replaces it

Both provide `import vcf`, which is likely why `cuteSVTrio -h` passes CI — the import succeeds either way at argument-parsing time. But for actual VCF handling on newer Python, users would hit runtime errors. Best to match what upstream requires:

```suggestion
    - pyvcf3
```

thanhleviet (commented):
Thanks for the new recipe — sha256 verifies, CI is green, and the structure is clean. One dependency mismatch worth fixing before merge: upstream declares `pyvcf3` in `install_requires`, but the recipe lists the legacy `pyvcf`. Details inline.

PR #64575, opened 2026-04-17
Update rnaframework PERL5LIB in conda activate/deactivate hooks

Vicbeg:
Adds activate.d/deactivate.d scripts to set PERL5LIB on conda activation, fixing deployment in Galaxy and Snakemake environments where the environment may not be fully activated interactively. Bumps build number to 2.

### DIFF
--- a/recipes/rnaframework/build.sh
+++ b/recipes/rnaframework/build.sh
@@ -3,6 +3,8 @@ set -euo pipefail
 
 mkdir -p "${PREFIX}/share/rnaframework/lib"
 mkdir -p "${PREFIX}/bin"
+mkdir -p "${PREFIX}/etc/conda/activate.d"
+mkdir -p "${PREFIX}/etc/conda/deactivate.d"
 
 export DATAPATH="${PREFIX}/share/rnastructure/data_tables/"
 
@@ -22,5 +24,26 @@ for script in rf-*; do
         echo "WARNING: expected script '${script}' not found in source tree" >&2
     fi
 done
+
+cat > "${PREFIX}/etc/conda/activate.d/${PKG_NAME}_perl5lib.sh" <<'EOF'
+if [ "${PERL5LIB+x}" = x ]; then
+    export _RF_OLD_PERL5LIB="${PERL5LIB}"
+    export _RF_HAD_PERL5LIB=1
+else
+    unset _RF_OLD_PERL5LIB
+    export _RF_HAD_PERL5LIB=0
+fi
+export PERL5LIB="${CONDA_PREFIX}/share/rnaframework/lib${PERL5LIB:+:${PERL5LIB}}"
+EOF
+
+cat > "${PREFIX}/etc/conda/deactivate.d/${PKG_NAME}_perl5lib.sh" <<'EOF'
+if [ "${_RF_HAD_PERL5LIB:-0}" = 1 ]; then
+    export PERL5LIB="${_RF_OLD_PERL5LIB}"
+else
+    unset PERL5LIB
+fi
+unset _RF_OLD_PERL5LIB
+unset _RF_HAD_PERL5LIB
+EOF
 
 echo "RNAFramework ${PKG_VERSION} installed successfully."
--- a/recipes/rnaframework/build.sh
+++ b/recipes/rnaframework/build.sh
@@ -26,24 +26,21 @@ for script in rf-*; do
 done
 
 cat > "${PREFIX}/etc/conda/activate.d/${PKG_NAME}_perl5lib.sh" <<'EOF'
-if [ "${PERL5LIB+x}" = x ]; then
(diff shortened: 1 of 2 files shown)


Vicbeg:
@BiocondaBot please add label

bgruening:
I'm not a perl expert but this looks dangerous :)

Can you describe a bit the problem that you are facing?

Vicbeg:
Hi bgruening,

the root cause is that RNAFramework's scripts (e.g. rf-map, rf-count) hardcode a use lib "/path/to/env/share/rnaframework/lib"; line pointing to their own Perl modules. This works fine in an interactive conda environment, but in non-interactive contexts like Galaxy tool execution or a Snakemake pipeline running on a server, the shell may not fully activate the conda environment — so Perl falls back to whatever PERL5LIB is set to in the calling environment, which can point to a different Perl installation or simply be missing the rnaframework lib path entirely.

A user hit exactly this: running rnaframework tools via Galaxy/Snakemake failed with Perl module errors, and the fix was to manually add export PERL5LIB="$CONDA_PREFIX/share/rnaframework/lib:$PERL5LIB" to their environment. This commit bakes that fix into the conda recipe via activate.d/deactivate.d hooks, so the correct lib path is set automatically whenever the environment is activated — even in non-interactive contexts — and cleanly restored on deactivation.

I hope this explains the fix :)

thanhleviet (commented):
Thanks for the detailed explanation of the use case. Adding on to @bgruening's question with a couple of technical concerns — CI is green so not blocking, but worth considering before merge.

recipes/rnaframework/build.sh
```diff
@@ -23,4 +25,25 @@ for script in rf-*; do
...
+cat > "${PREFIX}/etc/conda/activate.d/${PKG_NAME}_perl5lib.sh" <<'EOF'
+if [ "${PERL5LIB+x}" = x ]; then
+    export _RF_OLD_PERL5LIB="${PERL5LIB}"
+    export _RF_HAD_PERL5LIB=1
+else
+    unset _RF_OLD_PERL5LIB
+    export _RF_HAD_PERL5LIB=0
+fi
+export PERL5LIB="${CONDA_PREFIX}/share/rnaframework/lib${PERL5LIB:+:${PERL5LIB}}"
+EOF
```
thanhleviet, line 32: Since the existing `build.sh` already injects `use lib "${PREFIX}/share/rnaframework/lib";` into every `rf-*` script, `@INC` is populated without PERL5LIB. Could you clarify which failure mode this actually fixes?

My suspicion is child Perl processes spawned by `rf-*` via `system`/backticks — they inherit `PERL5LIB` but not the parent's `@INC`. If that's the case, adding a short comment in the activate script explaining this would help future maintainers (otherwise they'll see the hardcoded `use lib` and wonder why this hook exists).

recipes/rnaframework/build.sh
```diff
@@ -23,4 +25,25 @@ for script in rf-*; do
...
+
+cat > "${PREFIX}/etc/conda/deactivate.d/${PKG_NAME}_perl5lib.sh" <<'EOF'
+if [ "${_RF_HAD_PERL5LIB:-0}" = 1 ]; then
+    export PERL5LIB="${_RF_OLD_PERL5LIB}"
+else
+    unset PERL5LIB
+fi
+unset _RF_OLD_PERL5LIB
+unset _RF_HAD_PERL5LIB
+EOF
```
thanhleviet, line 44: The snapshot-and-restore pattern here can clobber `PERL5LIB` changes made by activate.d hooks from *other* packages in the same env. If another package's activate.d appends to `PERL5LIB` after this one runs, deactivating will restore the snapshot taken at activation time and drop that other package's contribution.

Safer pattern: on deactivation, strip only the specific path this hook added. Something like:

```bash
# deactivate.d
_rf_lib="${CONDA_PREFIX}/share/rnaframework/lib"
if [ -n "${PERL5LIB:-}" ]; then
    PERL5LIB=":${PERL5LIB}:"
    PERL5LIB="${PERL5LIB//:${_rf_lib}:/:}"
    PERL5LIB="${PERL5LIB#:}"
    PERL5LIB="${PERL5LIB%:}"
    [ -z "${PERL5LIB}" ] && unset PERL5LIB || export PERL5LIB
fi
unset _rf_lib
```

This is also simpler — no need for the `_RF_OLD_PERL5LIB` / `_RF_HAD_PERL5LIB` helper vars.

recipes/rnaframework/build.sh
Vicbeg: You are right, I will add a comment to explain this.

recipes/rnaframework/build.sh
Vicbeg: Good catch

Vicbeg:
Thank you for the revision. I have amended with your suggestions.

thanhleviet (approved):

PR #64613, opened 2026-04-19
staphscope: update data packages to install into share/ (build 2)

bbeckley-hub:
This PR updates the staphscope data packages (mlst and sccmec) to install into `$PREFIX/share/staphscope/modules/` (version‑independent location) and uses the updated orchestrator (build 2) that looks there first. This resolves the Python version conflict that caused MLST and SCCmec to fail in environments with multiple Python versions.

- Main package: source updated to `v1.2.0-fix`, build number 2
- `staphscope-mlst-data`: now `noarch: generic`, build number 2, installs into share/
- `staphscope-sccmec-data`: now `noarch: generic`, build number 2, installs into share/

All tests pass locally. Please review and merge.

### DIFF
--- a/recipes/staphscope-mlst-data/meta.yaml
+++ b/recipes/staphscope-mlst-data/meta.yaml
@@ -16,7 +16,7 @@ build:
 
 test:
   commands:
-    - test -f $PREFIX/share/staphscope/modules/mlst_module/mlst_module.py
+    - test -n "$(find $PREFIX -name mlst_module.py -print -quit)"
 
 about:
   home: https://github.com/bbeckley-hub/staphscope-typing-tool
--- a/recipes/staphscope-mlst-data/meta.yaml
+++ b/recipes/staphscope-mlst-data/meta.yaml
@@ -5,20 +5,20 @@ package:
   version: {{ version }}
 
 source:
-  url: https://github.com/bbeckley-hub/staphscope-typing-tool/releases/download/v{{ version }}/staphscope-mlst-data-v{{ version }}.tar.gz
+  url: https://github.com/bbeckley-hub/staphscope-typing-tool/releases/download/v{{ version }}/staphscope-mlst-data-v{{ version }}-full.tar.gz
   sha256: 8738c00aa778e22c497fc50735a51152da9cac169aa3edfb2635cf49b07f434a
 
 build:
   noarch: generic
-  number: 1
+  number: 2
   run_exports:
     - {{ pin_subpackage('staphscope-mlst-data', max_pin="x") }}
 
 test:
   commands:
-    - test -n "$(find $PREFIX -name mlst_module.py -print -quit)"
+    - test -f $PREFIX/share/staphscope/modules/mlst_module/mlst_module.py
 
 about:
   home: https://github.com/bbeckley-hub/staphscope-typing-tool
-  summary: 'MLST typing database for StaphScope'
   license: MIT
+  summary: 'MLST typing database for StaphScope'
--- a/recipes/staphscope-mlst-data/meta.yaml
+++ b/recipes/staphscope-mlst-data/meta.yaml
@@ -5,7 +5,7 @@ package:
   version: {{ version }}
 
 source:
-  url: https://github.com/bbeckley-hub/staphscope-typing-tool/releases/download/v{{ version }}/staphscope-mlst-data-v{{ version }}-full.tar.gz
+  url: https://github.com/bbeckley-hub/staphscope-typing-tool/releases/download/v{{ version }}-fix/staphscope-mlst-data-v{{ version }}-full.tar.gz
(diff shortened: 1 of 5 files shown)


bbeckley-hub:
@BiocondaBot please add label

recipes/staphscope/meta.yaml
```diff
@@ -47,7 +47,7 @@ requirements:
     - perl-data-dumper
     - perl-getopt-long
     - perl-file-which
-    # Data packages
+    # Data packages (build 2)
     - staphscope-mlst-data ==1.2.0
```
bgruening, line 51:
```suggestion
    - staphscope-mlst-data {{ version }}
```

Maybe?

bbeckley-hub: Let's stick! I'll consider that next updates. Thanks for the review @bgruening

bbeckley-hub:
@thanhleviet This green has been waiting for you!

thanhleviet (commented):
Nice cleanup — moving data into `$PREFIX/share/` is a clear improvement over the previous site-packages approach. CI is green. A few follow-ups (non-blocking) to tighten the recipe before shipping:

recipes/staphscope/meta.yaml
```diff
@@ -47,7 +47,7 @@ requirements:
...
     - staphscope-sccmec-data ==1.2.0
```
thanhleviet, line 52: Seconding @bgruening here — worth doing this now rather than "next update," since future bumps will otherwise require edits in 4 places (version jinja + 3 literal pins) and the literal pins tend to get forgotten:

```suggestion
    - staphscope-mlst-data =={{ version }}
    - staphscope-sccmec-data =={{ version }}
```

recipes/staphscope/meta.yaml
```diff
@@ -6,13 +6,13 @@ package:
   version: {{ version }}
 
 source:
-  url: https://github.com/bbeckley-hub/staphscope-typing-tool/archive/v{{ version }}.tar.gz
-  sha256: c737c024f9ba5b84cd6b1fd0eb0617771c81c7277dd5e5f9aad63516016c7f5e
+  url: https://github.com/bbeckley-hub/staphscope-typing-tool/archive/v{{ version }}-fix.tar.gz
```
thanhleviet, line 9: The `v{{ version }}-fix` tag with `version: 1.2.0` is confusing: the recipe says 1.2.0 but the source content isn't the original 1.2.0 tarball. Conda differentiates by `build_number` (which you bumped to 2, good), so this works — but future maintainers will wonder why a clean 1.2.0 rebuild points to a different tag.

For next time, a proper `v1.2.1` upstream tag is cleaner than a suffix. Not blocking this PR.

recipes/staphscope-mlst-data/meta.yaml
```diff
@@ -20,5 +20,5 @@ test:
 
 about:
   home: https://github.com/bbeckley-hub/staphscope-typing-tool
-  summary: 'MLST typing database for StaphScope'
   license: MIT
```
thanhleviet, line 23: Bioconda convention is to bundle the license text for MIT packages. If the repo has a LICENSE file, adding `license_file: LICENSE` is a small improvement:

```suggestion
  license: MIT
  license_file: LICENSE
```

(Assuming the tarball contains LICENSE — if not, this can be skipped.)

recipes/staphscope-sccmec-data/meta.yaml
```diff
@@ -5,20 +5,20 @@ package:
...
 
 test:
   commands:
-    - test -n "$(find $PREFIX -name mec_database_20171117.fasta -print -quit)"
+    - test -n "$(find $PREFIX -name run_sccmec_batch.sh -print -quit)"
 
 about:
   home: https://github.com/bbeckley-hub/staphscope-typing-tool
-  summary: 'SCCmec typing database for StaphScope'
   license: MIT
```
thanhleviet, line 23: Same as the mlst-data recipe — `license_file: LICENSE` is recommended for MIT if the tarball contains it:

```suggestion
  license: MIT
  license_file: LICENSE
```

thanhleviet:
One more thought (for a future version, not this PR): since `staphscope`, `staphscope-mlst-data`, and `staphscope-sccmec-data` always bump in lockstep and pin each other with `==1.2.0`, they're really a candidate for a **multi-output recipe** rather than three separate recipes.

A single `recipes/staphscope/meta.yaml` with an `outputs:` block produces all three packages from one source tarball and one `{% set version %}`. `pin_subpackage(..., exact=True)` handles the cross-pin automatically, so every bump is a one-line change. The current layout structurally invites the drift problem — every release needs coordinated edits across 4 files.

No action needed for this PR. Worth considering for the next major bump.

PR #64906, opened 2026-04-28
Update iobrpy to 0.1.8

hhn0527:
Update the iobrpy 0.1.8 recipe with a revised build.

Changes:
- bump build number from 0 to 1
- add samtools 1.21 as a runtime dependency

The source tarball and sha256 are unchanged because this is a recipe-only rebuild for the existing 0.1.8 release.


hhn0527:
@BiocondaBot please add label

bgruening (approved):

PR #64915, opened 2026-04-28
Update muat to version 0.1.18

primasanjaya:
Add features:

- input file from tsv
- fixing ensemble prediction

- fixing genome reference calling


kbseah:
PR has been superseded, v0.1.18 is available

PR #64919, opened 2026-04-28
Rebuild r-bpcells with current R pinning

koefoeden:
This rebuilds `r-bpcells` against Bioconda's current R pinning by bumping the build number.

The current uploaded `r-bpcells` build was produced for R 4.4, which prevents installation into environments using the current R 4.5 pinning. The recipe remains generic (`r-base`) so existing R 4.4 builds remain available while Bioconda can publish a new build for the current R stack.

This also adds the now-required `{{ stdlib('c') }}` build requirement for recipes using C/C++ compilers.

### DIFF
--- a/recipes/r-bpcells/conda_build_config.yaml
+++ b/recipes/r-bpcells/conda_build_config.yaml
@@ -1,0 +1,2 @@
+c_stdlib_version:  # [osx and x86_64]
+  - 10.15          # [osx and x86_64]
--- a/recipes/r-bpcells/conda_build_config.yaml
+++ b/recipes/r-bpcells/conda_build_config.yaml
@@ -1,2 +1,0 @@
-c_stdlib_version:  # [osx and x86_64]
-  - 10.15          # [osx and x86_64]
--- a/recipes/r-bpcells/meta.yaml
+++ b/recipes/r-bpcells/meta.yaml
@@ -25,6 +25,7 @@ requirements:
     - cross-r-base {{ r_base }}  # [build_platform != target_platform]
     - {{ compiler('c') }}
     - {{ compiler('cxx') }}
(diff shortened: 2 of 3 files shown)


recipes/r-bpcells/conda_build_config.yaml
```diff
@@ -0,0 +1,2 @@
+c_stdlib_version:  # [osx and x86_64]
```
bgruening: Is that really needed?

koefoeden: **EDIT: See follow-up below**

You're right, this recipe-local pin was not needed. The lint failure only required adding `{{ stdlib('c') }}` to the build requirements because the recipe uses compilers. I have dropped `recipes/r-bpcells/conda_build_config.yaml` and kept the stdlib requirement plus the configure patch that passes Conda's CPPFLAGS into the BPCells probes.

koefoeden: Follow-up: CI confirmed this pin is needed in Bioconda's current OSX-64 environment. After dropping the recipe-local file, the failed OSX-64 job rendered `macosx_deployment_target_osx-64 10.13` / `__osx >=10.13` and BPCells failed its C++17 `std::filesystem` configure probe. The earlier run with this file rendered `macosx_deployment_target_osx-64 10.15` / `__osx >=10.15` and passed. I have restored the recipe-local `c_stdlib_version: 10.15` pin while keeping `{{ stdlib('c') }}` in the build requirements.

koefoeden: I hope this is the optimal approach, otherwise please let me know - I'm not very experienced in creating these recipes!

PR #64925, opened 2026-04-28
Add kegg_pull 3.2.2

LucoDevro:
This PR brings the PyPi package kegg_pull to Bioconda, enabling its use as a dependency in other Bioconda packages.


LucoDevro:
@BiocondaBot please add label

bgruening (approved):

PR #64937, opened 2026-04-28
Add acinetoscope v1.1.0 with bundled databases

bbeckley-hub:
This PR adds acinetoscope v1.1.0, a complete genomic typing pipeline for Acinetobacter baumannii. The source tarball already contains all the necessary MLST, Kaptive, AMR, and virulence databases – no separate data packages, no post‑install downloads, no “where did my database go?” surprises. 😅

- Bundled everything – MLST schemes, Kaptive reference, ABRicate databases, etc.

- Parallel execution – QC, MLST, Kaptive, AMR, and ABRicate run concurrently (where possible).

- Clean recipe – uses matplotlib-base, run_exports, and a simple --help test.

- Tested locally – all green.

Ready for review and merge. 🙏

### DIFF
--- a/recipes/acinetoscope/meta.yaml
+++ b/recipes/acinetoscope/meta.yaml
@@ -1,0 +1,90 @@
+{% set name = "acinetoscope" %}
+{% set version = "1.1.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/bbeckley-hub/acinetoscope/archive/v{{ version }}.tar.gz
+  sha256: 5228805a1fca2d2203215bad80b1961fc67dec2e39976c349919f55d2b758bfe
+
+build:
+  noarch: python
+  number: 0
+  script: "{{ PYTHON }} -m pip install . --no-deps --no-build-isolation -vvv"
+  run_exports:
+    - {{ pin_subpackage(name, max_pin="x") }}
+
+requirements:
+  host:
+    - python >=3.9
+    - pip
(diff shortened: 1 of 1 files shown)


bbeckley-hub:
@BiocondaBot please add label

bbeckley-hub:
@bgruening @thanhleviet – this PR has been ready for few weeks and all checks are green. Could one of you please take a look and merge? Thank you! 🙏

recipes/acinetoscope/meta.yaml
```diff
@@ -0,0 +1,90 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/bbeckley-hub/acinetoscope/archive/v{{ version }}.tar.gz
+  sha256: f9ace682414a6b0ba4888992d59a1596ec207991d5ee0a9c4f8fc32850cb90b6
+
+build:
+  noarch: python
```
bgruening, line 13:
```suggestion
  noarch: generic
```
Would be an option? Not sure how deep the perl integration is.

bbeckley-hub: @bgruening thank you for the review. The Perl scripts are called as external subprocesses (via subprocess.run), not as Python modules, so noarch: python works fine – and it worked for our other tools (like staphscope) that were merged with the same setting. The CI has passed, so I’d prefer to keep it as is unless you strongly recommend changing to noarch: generic. Please let me know. Thanks again!

PR #64983, opened 2026-04-30
Add TEsmall Bioconda recipe

niekwit:
Add recipe for TEsmall 2.0.9

TEsmall is a pipeline for the simultaneous processing and analysis of a variety of small RNAs in a single integrated workflow.

**Notes:**
- `bokeh` is pinned to `<3.0.0` because `bokeh.models.widgets` was removed in bokeh 3.0, which breaks TEsmall at import (`summary.py`). This is a known upstream incompatibility.
- `numpy` is pinned to `<2.0` because bokeh 2.x uses `np.bool8`, which was removed in NumPy 2.0.
- `run_exports` is included with `max_pin="x.x"` given the 0.x.x-style versioning.

### DIFF
--- a/recipes/tesmall/meta.yaml
+++ b/recipes/tesmall/meta.yaml
@@ -1,0 +1,63 @@
+{% set name = "tesmall" %}
+{% set version = "2.0.9" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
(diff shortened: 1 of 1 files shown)


niekwit:
@BiocondaBot please add label

bgruening (approved):

PR #65006, opened 2026-05-01
Add dipgenie v1.0

gsc74:
## Add dipgenie v1.0

DipGenie is a tool for pangenome graph-based phased diploid genome inference. It estimates full haplotype sequences from low-coverage short-read sequencing data for both haploid and diploid genomes.

- **Home:** https://github.com/gsc74/DipGenie
- **Publication:** Ghanshyam Chandra, William T. Doan and Daniel Gibney. *Haplotype-resolved diploid genome inference on pangenome graphs.* bioRxiv (doi: [10.1101/2025.11.26.690754](https://doi.org/10.1101/2025.11.26.690754))
- **License:** MIT AND Apache-2.0 (composite; main code is public domain, with MIT/Apache components from minigraph, vg, MurmurHash3)
- **Platform:** linux-x86_64

`run_exports` is set with `max_pin="x.x"` since the tool is at version 1.0 and uses semantic versioning.


gsc74:
@Mergifyio queue

gsc74:
@thanhleviet  please add to merge queue

recipes/dipgenie/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+source:
+  url: https://github.com/gsc74/DipGenie/archive/v{{ version }}.tar.gz
+  sha256: 624de036528498de2f8ae88b1bbc5f7d571ca080a36904b4925552b01ff00a38
+
+build:
+  number: 0
+  skip: true  # [not linux or not x86_64]
+  run_exports:
+    - {{ pin_subpackage(name, max_pin="x.x") }}
```
thanhleviet: For semver 1.x+ packages, bioconda convention is `max_pin="x"` (pin only the major). `max_pin="x.x"` is reserved for `0.x.x` releases where minor bumps may be breaking. The PR description states the package follows semver, so a major-only pin is the right choice here.

```suggestion
    - {{ pin_subpackage(name, max_pin="x") }}
```

recipes/dipgenie/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ stdlib('c') }}
+    - make
+  host:
+    - zlib
+    - libgomp  # [linux]
+  run:
```
thanhleviet, line 26: Optional: `zlib` (host) already declares a `run_exports` on `libzlib`, so an explicit runtime entry isn't strictly needed. If you prefer to keep it explicit, use `libzlib` (the actual runtime library) rather than `zlib` (the headers package). Not blocking.

recipes/dipgenie/build.sh
```diff
@@ -0,0 +1,11 @@
...
+set -ex
+
+# Build DipGenie (override -march=native for portability)
+make CXX="${CXX} -O3 -std=c++17" \
+     CXXFLAGS="-fopenmp -pthread -O3 -lm -lz -lpthread -ldl -I${PREFIX}/include -L${PREFIX}/lib" \
+     -j${CPU_COUNT}
+
+# Install binary
+mkdir -p "${PREFIX}/bin"
+install -m 755 DipGenie "${PREFIX}/bin/DipGenie"
```
thanhleviet, line 11: Optional: upstream ships a `vcf2gfa.py` helper alongside the `DipGenie` binary. If users are expected to use it for input prep, consider installing it too:

```bash
install -m 755 vcf2gfa.py "${PREFIX}/bin/vcf2gfa.py"
```

and adding `python` to `run`. Skip if it's an internal-only script.

thanhleviet (commented):
Thanks for the recipe. CI is green, sha256 verified against the upstream tarball, and the build script correctly overrides `-march=native` for portability. One important pin convention nit and a couple of optional suggestions inline.

recipes/dipgenie/build.sh
gsc74: @thanhleviet,  Thanks, I'll skip this since `vcf2gfa.py` is an internal helper script and not part of the intended user-facing workflow.

I think this is ready to proceed with merge.

gsc74: @thanhleviet anything left for merging?

thanhleviet (approved):

PR #65050, opened 2026-05-03
Cryfa 26.05

smortezah:
This pull request updates the `cryfa` recipe to a new version and adjusts its dependencies. The most significant changes are the version bump, checksum update, and the addition of the `cryptopp` library as a dependency.

Version and source update:
* Updated the `cryfa` version from `26.04` to `26.05`, reset the build number to `0`, and updated the source archive's SHA256 checksum to match the new release.

Dependency changes:
* Added `cryptopp` as a dependency in both the `host` and `run` sections to ensure it is available during build and runtime.


kbseah (approved):
thanks for the update, looks good to me!

PR #65051, opened 2026-05-03
Fix TEsmall 2.0.9 recipe: add missing wget dependency

niekwit:
`wget` is required at runtime for downloading reference genome and annotation files but was missing from the original recipe.


niekwit:
@BiocondaBot please add label

kbseah (approved):
Update adds missing run requirement, build and tests pass. Looks good to me!

PR #65074, opened 2026-05-04
Update Smashpp to 26.05

smortezah:
Update smashpp to v26.05 and pass the package version through CMake so the CLI version test reports 26.05.


kbseah (approved):
thanks for the fix! looks good to me

PR #65110, opened 2026-05-05
Fix GECCO dependency

gl-eb:
GECCO depends on pyhmmer 0.12.0 since version 0.10.2. See [GECCO's changelog](https://github.com/zellerlab/GECCO/blob/e837e629e1264b17ce800acc4595f468c8fae1b9/CHANGELOG.md?plain=1#L25).


gl-eb:
@BiocondaBot please add label

mbhall88 (approved):

PR #65168, opened 2026-05-07
Add r nmf

Yuri-Kawahara:
Add r-nmf (v0.28), a framework for Non-negative Matrix Factorization.

This recipe was migrated from conda-forge because `NMF` depends on `bioconductor-biobase`.


Yuri-Kawahara:
Added skip-lints for in_other_channels because v0.28 introduces a dependency on bioconductor-biobase which cannot be added to conda-forge.

PR #65260, opened 2026-05-10
Update hicstuff to 3.2.5

js2264:
- Version bump 3.2.4 → 3.2.5
- Update entry point: `hicstuff.main:main` → `hicstuff.cli:cli`
- Require Python ≥3.8
- Replace `docopt` with `rich-click >=1.7`
- Add `wheel` to host requirements

### DIFF
--- a/recipes/hicstuff/meta.yaml
+++ b/recipes/hicstuff/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "3.2.4" %}
+{% set version = "3.2.5" %}
 
 package:
   name: hicstuff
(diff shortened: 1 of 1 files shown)


js2264:
This fixes auto-bump in #65244.

js2264:
@BiocondaBot please add label

recipes/hicstuff/meta.yaml
```diff
@@ -1,31 +1,32 @@
...
 source:
   url: https://pypi.io/packages/source/h/hicstuff/hicstuff-{{ version }}.tar.gz
-  sha256: 7d74bcbddaff40029a07ed18283701110b338617fbbda13b9b6ec6e3cc107d02
+  sha256: 8db7b39fee781bdbbe81add343e2ff49d43ee5b41d21cd3e023962e62b1149b6
 
 build:
   number: 0
   noarch: python
   entry_points:
-    - hicstuff=hicstuff.main:main
```
kbseah: The `entry_points` field can be removed entirely because it is already specified in the upstream `pyproject.toml`

kbseah (changes requested):
Thanks for the update! Suggesting a change below

recipes/hicstuff/meta.yaml
js2264: fixed by e0a9c9d

js2264:
Hi @kbseah thanks for your feedback. I've removed `entry_points` key from the recipe yaml. Let me know if anything else would be needed, thanks for your help!

kbseah (approved):
Thanks, build and tests pass after the update. Looks good to me!

PR #65327, opened 2026-05-12
Update latch test imports and document `websockets` pin

msto:
Upstream `latch` declares `gql==3.5.0` with `websockets>=15.0.1`, which are not jointly satisfiable on conda — conda-forge `gql 3.5.x` carries `websockets <12` as a `run_constrained` on its base output. The recipe already pins `websockets ==11.0.3` to land inside that range; this PR documents the divergence in-tree and extends the test surface so future dep bumps surface module-level breakage at build time.

`python_min` is introduced as the standard noarch-python jinja variable; the floor is unchanged.

Tracking upstream:

- latchbio/latch#591
- conda-forge/gql-feedstock#37

Build number 2 → 3.

## Test plan

- Build succeeds with bumped number.
- Expanded `imports:` succeed on the resolved environment.
- `pip check` reports the upstream metadata mismatch without failing the test.
- `latch --help` and `latch --version` exit 0.

cc @nh13 (recipe maintainer)

### DIFF
--- a/recipes/latch/meta.yaml
+++ b/recipes/latch/meta.yaml
@@ -1,5 +1,6 @@
 {% set name = "latch" %}
 {% set version = "2.71.2" %}
+{% set python_min = "3.10" %}
 
 package:
   name: {{ name }}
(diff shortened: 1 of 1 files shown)


msto:
@BiocondaBot please add label

nh13 (approved):

PR #65332, opened 2026-05-12
Update grz-common to 2.0.1

Hoeze:
- **Update grz-common to 2.0.1**
- **Update maintainers**
- **Remove transitive dependencies**


tedil (changes requested):
~nvm, need to fix dependencies first~
I got confused because there were two 2.0.1 update PRs.

tedil (approved):

PR #65446, opened 2026-05-15
Bump build number to 1 and add zlib dependency

bgruening:
@FelixKrueger it seems zlib is missing as dependency? Can you please have a look at this PR if it makes sense to you?

Would it be possible if we can build this package natively here, or is that not possible?

----

### General instructions

```yaml
build:
  run_exports:
    - ...

```

with `...` being one of:

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

</details>

### DIFF
--- a/recipes/trim-galore/meta.yaml
+++ b/recipes/trim-galore/meta.yaml
@@ -14,13 +14,14 @@ source:
     sha256: cac32ddcd2ecd6837e770d097892002af2564781c9b3843ca70d75dda862d15b  # [osx and arm64]
 
 build:
-  number: 0
+  number: 1
   skip: True  # [osx and x86_64]
   run_exports:
     - {{ pin_subpackage(name, max_pin="x") }}
 
 requirements:
-  run: []
+  host:
+    - zlib
 
 test:
   commands:


FelixKrueger:
Thanks for catching this, @bgruening! You're right that zlib is needed, and the reason is a bit subtle.

My `Cargo.toml` configures `flate2` with `default-features = false, features = ["zlib-rs"]` — i.e. requesting the pure-Rust backend with no C dependency. But two transitive crates pull `flate2` with conflicting feature sets:

- `fastqc-rust 1.0.1` enables its `native-zlib` feature, which propagates through `zip` → `deflate-flate2` and into `flate2`'s C zlib backend.
- `noodles-bgzf 0.35.0` uses `uses_default_features: True`, which includes `flate2`'s default `zlib` feature.

Cargo's feature unification is additive across the whole dependency graph, so the unified flate2 feature set becomes `zlib-rs + libz-sys + zlib + any_c_zlib + ...` — both backends get compiled, and `libz-sys`'s default behavior on Linux is "prefer system zlib via pkg-config", so the Linux release binary dynamically links `libz.so.1`. Most environments don't notice because zlib is on every distro, but a sandboxed conda env without an explicit dep can hit `error while loading shared libraries: libz.so.1`.

**One small note on this PR's placement**: I think `zlib` belongs in `run:` rather than `host:`. This recipe installs a prebuilt binary with no compilation step, so `host:` (build-time) has no effect — the dependency we need is at *runtime* when the binary execs and `libz.so.1` is dlopen'd. Suggested change:

```yaml
requirements:
  run:
    - zlib
```

Happy for this PR to land with that small adjustment as the immediate fix.

**On building natively**: yes, completely feasible and probably the better durable answer. Switching the recipe to `cargo build --release --locked` from the source tarball would:

1. Declare libz correctly through conda's standard Rust toolchain (no manual `run:` entry needed).
2. Resolve the `glibc >= 2.39` baseline concern noted in the existing recipe comment (the prebuilt Linux binaries are built on `ubuntu-latest`, which uses a newer glibc than bioconda CI's CentOS 7 base).
3. Let us drop the `skip: True  # [osx and x86_64]` line — we don't ship an Intel Mac prebuilt because GitHub Actions retired that runner, but bioconda CI can build natively.

The Rust toolchain (`rust >= 1.88`) is available in conda-forge. I'm happy to drive a follow-up PR with the native-build recipe; would you prefer it as a separate PR after this one lands, or to fold both changes into this one?

PR #65453, opened 2026-05-16
Update metagenomescope to 1.3.0

BiocondaBot:
Update [`metagenomescope`](https://bioconda.github.io/recipes/metagenomescope/README.html): **1.2.0** &rarr; **1.3.0**

[install with bioconda](http://bioconda.github.io/recipes/metagenomescope/README.html) [Conda](https://anaconda.org/bioconda/metagenomescope/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/metagenomescope`](https://github.com//bioconda/bioconda-recipes/tree/bump/metagenomescope/recipes/metagenomescope) (click to view/edit other files)
Summary | Visualization tool for (meta)genome assembly graphs
Home | [https://github.com/marbl/MetagenomeScope](https://github.com/marbl/MetagenomeScope)
Releases |[https://pypi.org/pypi/metagenomescope/json](https://pypi.org/pypi/metagenomescope/json)
Recipe Maintainer(s) | @fedarko

***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

PR #65454, opened 2026-05-16
Update MetagenomeScope to 1.3.0, and update dependencies

fedarko:
It looks like I don't have permission to edit #65453 - this is intended to replace that.

Version 1.3.0 of [MetagenomeScope](https://github.com/marbl/MetagenomeScope/) removes the dependency on Gfapy, so this PR just applies the changes from #65453 and removes the `gfapy` line from the list of `run` requirements. Thanks!


fedarko:
@BiocondaBot please add label

kbseah (approved):
Thanks for the update, looks good to me!

PR #65462, opened 2026-05-16
Update gedi: pin openjdk to >=11,<18 (Java 25 SIGSEGV)

pinin4fjords:
With the previous `openjdk >=11` pin, conda resolves to `openjdk 25.0.2` from conda-forge HEAD. On Java 25, `gedi -e Price` SIGSEGVs at class-discovery time (`StringLatin1.replace` in `java.base@25.0.2`).

Tested across LTS Java versions via Wave-built containers:

| Java | `gedi -h` | `gedi -e Version` | `bamlist2cit -h` | `gedi -e Price -h` |
|------|-----------|-------------------|------------------|--------------------|
| 11   | pass      | pass              | pass             | pass               |
| 17   | pass      | pass              | pass             | pass               |
| 21   | pass      | pass              | pass             | SIGSEGV (`jdk.internal.jimage.ImageReader`) |
| 25   | pass      | pass              | pass             | SIGSEGV (`java.lang.StringLatin1.replace`) |

Pinning `openjdk >=11,<18` keeps the supported LTS range (11, 17) and excludes both broken versions. Java 21 fails too, so a wider `<22` would still be broken.

Also added `gedi -e Price -h` to `test.commands` so this regression would have been caught on the original recipe.

Bumps `build.number` from 0 to 1.

Follow-up to #65445.

### DIFF
--- a/recipes/gedi/build.sh
+++ b/recipes/gedi/build.sh
@@ -7,7 +7,9 @@ set -euxo pipefail
 # invoke maven against the Gedi module from inside it.
 mkdir -p build_out
 cd build_out
-mvn -f "${SRC_DIR}/Gedi/pom.xml" package -DskipTests=true -B
+mvn -f "${SRC_DIR}/Gedi/pom.xml" package -DskipTests=true -B \
+    -Dproject.build.sourceEncoding=UTF-8 \
+    -Dfile.encoding=UTF-8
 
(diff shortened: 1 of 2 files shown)


matthdsm (approved):
Lgtm, but linting failed

PR #65549, opened 2026-05-19
Update psQTL to v1.3.9

zkstewart:
- Clarify GPL licence.
- Make executables via symlink rather than copying the entire file.


zkstewart:
@BiocondaBot please add label

kbseah (approved):

PR #65570, opened 2026-05-19
Add gzip to spechla runtime deps (provides zless)

jonasscheid:
SpecHLA's `script/whole/SpecHLA.sh` invokes `zless file.vcf.gz | grep "#"`, which fails inside the current biocontainer with `zless: command not found`. Adding `gzip` (conda-forge) to runtime requirements provides `zless` and resolves the missing-dep at the packaging level.


jonasscheid:
Closing since fix will ship via https://github.com/bioconda/bioconda-recipes/pull/65578

PR #65614, opened 2026-05-20
Update Locityper to v1.5.4

tprodanov:
For some reason, autobumps did not work for Locityper since v1.3.4. In this commit I try to fix clang to v21 to check if this helps.

### DIFF
--- a/recipes/locityper/build.sh
+++ b/recipes/locityper/build.sh
@@ -7,10 +7,16 @@ fi
 
 export C_INCLUDE_PATH="${PREFIX}/include"
 export LIBRARY_PATH="${PREFIX}/lib"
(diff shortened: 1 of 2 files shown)


tprodanov:
@BiocondaBot please add label

bgruening:
Autobump is not building recipes. I would assume it's not related to this. Have you tried running autobump locally with bioconda-utils?

tprodanov:
@bgruening I think there were several issues with the autobump, first, rust-htslib dependency stopped compiling under clang v22, and after I did some other changes the patch in the recipe stopped applying. But here I think I fixed the issues, so it would be possible to disregard the autobump and merge this PR

bgruening (approved):

bgruening:
Thanks!

PR #65617, opened 2026-05-20
Update `sopa`

quentinblampey:
- Used `python-igraph` instead of `igraph`, as recommended [here](https://python.igraph.org/en/main/install.html).
- Updated the documentation links (the repo was recently moved to a new org).

NB: I did the same PR for novae, see #65618

### DIFF
--- a/recipes/sopa/meta.yaml
+++ b/recipes/sopa/meta.yaml
@@ -9,7 +9,6 @@ source:
   sha256: 0e4a490109ddc9ae4af024d0219587ba101bfacf20267b9f53b4ce42d84119f2
 
 build:
-  entry_points:
   noarch: python
   script: {{ PYTHON }} -m pip install . --no-deps --no-build-isolation --no-cache-dir -vvv
   number: 1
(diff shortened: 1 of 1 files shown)


quentinblampey:
@BiocondaBot please add label

recipes/sopa/meta.yaml
```diff
@@ -13,7 +13,7 @@ build:
     - sopa = sopa.main:app
```
kbseah:
```suggestion
```
The entry_points field can be removed because it is specified in upstream pyproject.toml

kbseah: I wanted to suggest removing both lines 12 and 13, but somehow only one of them shows up in the suggested change. Please check that both are removed, if you choose to accept the change.

kbseah (approved):
Thanks for the update! A small suggestion below

quentinblampey:
Thanks for your review and comment @kbseah, it's updated! (I indeed had to also remove line 12 in another commit)

kbseah (approved):

PR #65618, opened 2026-05-20
Update `novae`

quentinblampey:
- Used `python-igraph` instead of `igraph`, as recommended [here](https://python.igraph.org/en/main/install.html).
- Updated the documentation links (the repo was recently moved to a new org).

NB: I did the same PR for sopa, see #65617.


quentinblampey:
@BiocondaBot please add label

kbseah (approved):

PR #65635, opened 2026-05-20
Update ipcr to 5.0.0

ericksamera:
Updates ipcr to 5.0.0 and fixes the Bioconda lint error by adding stdlib('c') for the C compiler requirement.


kbseah (approved):
thanks for the update! looks good to me

PR #65640, opened 2026-05-21
Update openstructure to use the bioconda numpy version

s22chan:
There is no C linkage with numpy anymore due to https://git.scicore.unibas.ch/schwede/openstructure/-/commit/728ad1c9334ece304c0796c904ab650b3c59855c

So the python dependencies can be less restrictive for an env.


s22chan:
@BiocondaBot please add label

bgruening (approved):

PR #65653, opened 2026-05-21
Add mumu v1.1.4

ffillouxdev:
mumu is a fast, stand-alone C++20 reimplementation of the R package lulu.
It performs post-clustering curation of amplicon-based metabarcoding data
by detecting spurious OTUs and merging them with their parent OTUs based
on co-occurrence patterns and sequence similarity.


bgruening (approved):
Merci!

PR #65706, opened 2026-05-22
Add tapir pipeline

LymF:
TAPIR is an end-to-end pipeline for RNA virus discovery from metatranscriptomic data.
PyPI: https://pypi.org/project/tapir-pipeline/

### DIFF
--- a/recipes/tapir-pipeline/post-link.sh
+++ b/recipes/tapir-pipeline/post-link.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 # Install pip-only dependencies not available in bioconda/conda-forge
-"${PREFIX}/bin/pip" install --no-deps viralquest 2>&1 || true
+"${PREFIX}/bin/pip" install viralquest 2>&1 || true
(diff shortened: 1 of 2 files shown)


recipes/tapir-pipeline/post-link.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+# Install pip-only dependencies not available in bioconda/conda-forge
```
bgruening, line 2: Please add this to conda as well.

bgruening (commented):
Thanks @LymF

recipes/tapir-pipeline/post-link.sh
LymF: viralquest is a third-party PyPI package (maintained by a separate author) that bundles pre-compiled binaries (BLAST, DIAMOND, CAP3) within the package itself and depends on several pip-only packages (langchain, langchain-openai, langchain-anthropic, langchain-google-genai, ollama) that are currently not available in bioconda or conda-forge. Packaging it for bioconda would require significant upstream changes to the viralquest project (removing bundled binaries, splitting them into proper conda dependencies), which is outside the scope of this PR. The post-link.sh approach was used as the only viable workaround at this time. Do you have any suggestions on how best to handle this case within bioconda's guidelines?

bgruening: This sounds even worse. Less reproducible, abusing pypi as a file storage and everyone that uses your package will tained their environment with pypi packages, which is not recommended :-(

PR #65728, opened 2026-05-24
openstructure - relax py3.10 pin, scipy 2.13 is incompatible with py 3.13

s22chan:
py pin workaround was introduced in https://github.com/bioconda/bioconda-recipes/commit/a7cde3373308ebdb21b2633ca8d4a491e1e71e82 and it marked the problem as numpy related, but the real issue is that scipy 1.13.x doesn't have a py 3.13 build and the resolver gets confused.

### DIFF
--- a/recipes/openstructure/conda_build_config.yaml
+++ b/recipes/openstructure/conda_build_config.yaml
@@ -3,6 +3,3 @@ c_compiler_version: # [osx]
 
 cxx_compiler_version: # [osx]
   - 20 # [osx]
-
-python:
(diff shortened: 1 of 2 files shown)


s22chan:
@BiocondaBot please add label

s22chan:
ping @bgruening and @johanneskoester for a review. this reenables osx builds and py 3.11/3.12 support.

bgruening:

s22chan:
thank you @bgruening!

PR #65730, opened 2026-05-24
Update panoptes-ui to 1.0.0

fgypas:
panoptes 1.0.0 adds Snakemake 9 support and drops Python < 3.11. `pytest` and `requests` are no longer runtime dependencies (they moved to a `[dev]` extra upstream), so they are removed from `host`/`run`; `python` is pinned to `>=3.11` and `setuptools` is added to `host`. A `run_exports` (`pin_subpackage`, `max_pin='x'`) is added per the recipe guidelines.

This is the human-curated version of the bump (dependency/metadata changes beyond a plain version+sha autobump).

Upstream release: https://github.com/panoptes-organization/panoptes/releases/tag/v1.0.0

### DIFF
--- a/recipes/panoptes-ui/meta.yaml
+++ b/recipes/panoptes-ui/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "panoptes-ui" %}
-{% set version = "0.2.3" %}
+{% set version = "1.0.0" %}
 
 package:
   name: "{{ name|lower }}"
@@ -7,7 +7,7 @@ package:
 
 source:
   url: "https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz"
-  sha256: 5fa903711786d68fb014e0015d647dd56b80b844fa2d6520554487663e45635d
+  sha256: 5f539b9f30ea88a3459d7cb647d2b9e52ee83c21e3a965ff2d6592581886f038
 
 build:
   noarch: python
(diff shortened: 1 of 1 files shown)


fgypas:
@BiocondaBot please add label

fgypas:

recipes/panoptes-ui/meta.yaml
```diff
@@ -1,17 +1,19 @@
...
+  sha256: 5f539b9f30ea88a3459d7cb647d2b9e52ee83c21e3a965ff2d6592581886f038
 
 build:
   noarch: python
   number: 0
+  run_exports:
+    - {{ pin_subpackage(name, max_pin='x') }}
   entry_points:
     - panoptes=panoptes:main
   script: "{{ PYTHON }} -m pip install . -vv"
```
thanhleviet: This PR adds `setuptools` to `host`, but the build script still uses default build isolation — so pip spins up an isolated build env and re-downloads setuptools from PyPI, leaving the host `setuptools` unused. Adding `--no-deps --no-build-isolation` makes the host backend actually used and keeps the build deterministic (the bioconda guideline for pip installs). The two changes go together: adding the host dep implies opting out of build isolation.

```suggestion
  script: "{{ PYTHON }} -m pip install . --no-deps --no-build-isolation -vv"
```

thanhleviet (commented):
Verified: sha256 matches the PEP 625 underscore sdist (`panoptes_ui-1.0.0.tar.gz`), and the `name|replace('-', '_')` URL fix is correct — the hyphenated URL 404s. `run` deps now match upstream `install_requires` exactly (`pytest`/`requests` correctly dropped to the `dev` extra), `python >=3.11` matches `requires_python`, and `run_exports` `max_pin='x'` is right for a 1.x release. No duplicate/autobump PR open for this version. CI is green across Lint/Linux/OSX/ARM.

One should-fix on the build script inline. Optional, non-blocking: the sdist ships `LICENSE.md`, so `license_file: LICENSE.md` could be added (pre-existing omission); the `entry_points` block duplicates upstream's `console_scripts` (also pre-existing, harmless).

fgypas:
Hi @thanhleviet I made the necessary adjustments. Please let me know if something else is missing.

thanhleviet (approved):
The build script now opts out of build isolation so the host `setuptools` is actually used, and thanks for adding `license_file: LICENSE.md` on top — CI confirms it ships in the sdist. All green. LGTM!

PR #65732, opened 2026-05-24
Meme 5.5.9 rebuild

bcoltman:
This rebuilds MEME so Bioconda provides a build against the current conda-forge stack, rather than the existing build
tied to an older ICU 73.

Changes:
- bump build number from 0 to 1
- remove `nodejs >=16.4.1` from `requirements: host`

Rationale:
- The existing Bioconda build was built against ICU 73, which causes install conflicts with newer packages.
- A local rebuild produced a package against the newer stack, using ICU 75.1 in the test environment.
- MEME 5.5.9 builds and tests successfully without `nodejs`.
- No Node.js build setup was found in the source tarball, and upstream install docs do not list Node.js as a requirement.

Validation:
- `bioconda-utils build --docker --mulled-test --packages meme` passed in a clean isolated build root.
- Conda-build tests passed.
- Mulled/container test passed.


bcoltman:
@BiocondaBot please add label

thanhleviet (approved):
Clean rebuild, well documented. Verified: source sha256 `0406fb...` matches the recipe; the `build: number` bump 0 -> 1 is correct for a metadata-only rebuild (re-pins MEME against the current ICU 75 stack); `stdlib('c')` is the right addition alongside `compiler('c')` for current lint. Confirmed the `nodejs` removal is justified — there is no node/npm usage anywhere in the MEME build system (no `package.json`, and the only `node` matches in `configure` are autotools internals like `am__nodep`); it sat in `host` and was unused. `additional-platforms` is appropriate here since meme is compiled, not noarch, and ARM CI is green. CI passes across Lint/Linux/OSX/ARM.

LGTM.

bcoltman:
Thanks @thanhleviet !
I accidentally clicked the button to merge master into this branch. No changes made but I don't know if the "reviewed" status changes.

PR #65752, opened 2026-05-25
Constrain graphbin2 to Python <3.13

ehsanestaji:
## Summary
- Constrain `graphbin2` to Python `>=3.9,<3.13`, matching upstream package metadata.
- Bump the build number so the corrected dependency metadata is rebuilt.
- Make the pip install invocation use `--no-deps --no-build-isolation --no-cache-dir` for a more deterministic Bioconda build.

## Why
Upstream `graphbin2` 1.3.3 declares `requires_python = >=3.9,<3.13`, but the Bioconda recipe only constrained runtime Python to `>=3.9`. This can allow incompatible Python 3.13+ environments for an assembly-graph binning tool.

## Checks
- [x] `bioconda-utils lint --packages graphbin2`
- [x] `bioconda-utils build --docker --mulled-test --force --pkg-dir "$PWD/.bioconda-build" --packages graphbin2`

### DIFF
--- a/recipes/graphbin2/meta.yaml
+++ b/recipes/graphbin2/meta.yaml
@@ -10,21 +10,21 @@ source:
   sha256: 03ab89dbb6d26c9b414ff808598c295cf22fd3af22816a5c9fada94104a79686
 
 build:
-  number: 0
+  number: 1
   noarch: python
   run_exports:
     - {{ pin_subpackage('graphbin2', max_pin="x.x") }}
   script:
-    - "{{ PYTHON }} -m pip install . -vv"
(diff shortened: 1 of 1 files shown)


thanhleviet (approved):
Verified against upstream: `graphbin2` `pyproject.toml` (v1.3.3) declares `requires-python = ">=3.9,<3.13"`, and all 1.3.x PyPI releases agree, so constraining both `host` and `run` python to `>=3.9,<3.13` is correct. The `build.number` bump `0 -> 1` is right for a metadata-only rebuild of the same version, and `--no-deps --no-build-isolation` is safe here since the `flit-core` backend is already present in `host` (CI confirms). Runtime deps still match upstream. CI is green across Lint/Linux/OSX/ARM.

LGTM.

PR #65763, opened 2026-05-26
Update mags-visualization 0.0.9 - Corrected python-kaleido pin

alexandrah1704:
Rebuild of mags-visualization 0.0.9 with corrected python-kaleido pin. The previous build did not include the version pin, causing kaleido v1.3.0 to be installed which requires Chrome. Pinning to 0.2.1.


alexandrah1704:
Hi @bgruening, this PR contains a small update. Could you take a look when you have a chance?
Thank you!

PR #65860, opened 2026-05-29
Update libxcrypt1 to runtime requirements to avoid perl errors in newer systems

KasperThystrup:
On some of the newer installs, perl would run into an error regarding a missing libcrypt.so.1, which was deprecated long time ago. This can be fixed by adding the above conda-forge repo to the install requirements -> https://github.com/conda-forge/perl-feedstock/issues/37#issuecomment-2556755284

This is my FIRST attempt to updating ANY conda package, so please let me know how to test it proper and let me know how to improve the PR!

[x] I have read and followed the description to the best of my knowledge :-)


KasperThystrup:
Made a mistake with versioning, incremented the number to make sure that the check passed, but later check failed because official repo didn't have that particular version. Increment should not be applied, as the source haven't changed, its just an extra dependecy added to avoid potential perl err.

KasperThystrup:
@BiocondaBot please add label

I posted this simultanously as clicking merge changes into branch. CLI passed before the merge

KasperThystrup:
@BiocondaBot please add label

kbseah (approved):
thanks for the fix, looks good to me!

PR #65885, opened 2026-05-29
Update rad to 1.0.2

indianewok:
## Update `rad` to 1.0.1

The published **0.6.0** package shipped only `bin/rad` and **no `resources/`**, so every command needing a read layout or barcode whitelist failed at runtime with `Cannot locate 'resources' folder` (the `cmake --install` was a no-op — the project had no `install()` rules — and a manual copy grabbed just the binary; the `rad --help`-only test masked it).

**What changed**
- **1.0.0** added real CMake `install()` rules: the binary goes to `libexec/rad/rad`, a self-locating `bin/rad` wrapper sets `RAD_RESOURCES`, and the bundled read layouts + barcode whitelists install to `$PREFIX/share/rad/resources`. `build.sh` drops the old manual binary copy.
- **1.0.1** discovers `pigz` on `PATH` at runtime (so a conda `pigz` is actually used, falling back to zlib) and adds `pigz` as a run dependency.
- Added `bzip2` to host: 1.0.x links `Boost::iostreams` (new vs 0.6.0, which linked filesystem+system), whose compression backends need zlib **and** bzip2.
- The test now asserts the binary runs **and** the bundled resources are installed/discoverable.

**Verification:** built locally with `conda build` on osx-arm64 — build + test phase pass, resources install correctly, `rad --help`/`--version`/`list` run in a clean env, and the binary links the declared `llvm-openmp`.


indianewok:
@BiocondaBot please merge

indianewok:
@BiocondaBot please add label

recipes/rad/meta.yaml
kbseah: The `max_pin` parameter is missing here. Was this intentional?

kbseah (changes requested):
Thanks for the update! Requesting clarification on the run_exports statement, please see the comment below

recipes/rad/meta.yaml
indianewok: hey @kbseah ! thanks for calling it out--no, it's a holdover from boilerplate-type stuff that I schlebbed over like a year ago. Just got rid of it! Thanks for your help :)

indianewok: Update--I'm a liar! Removed it. Lint test immediately failed. Clearly I'm new to this. Schlebbing it back as   run_exports:
    - {{ pin_subpackage('rad', max_pin="x") }}

If it craps out again, well, I'm sure I'll have done something equally stupid again lol. Thank you for your help!

kbseah: thanks for looking into this! resolved now

kbseah (approved):
issue raised in previous review has been resolved. looks good to me! thanks again

PR #65937, opened 2026-06-02
Add phanotate-rs recipe

Yasas1994:
PHANOTATE is a gene prediction tool specifically tailored for predicting phage genes. PHANOTATE-rs is a fast (10x faster) Rust re-implementation with  additional features such as automatic genetic-code detection.


Yasas1994:
@BiocondaBot please add label

recipes/phanotate-rs/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - openssl  # [linux]
+
+  run:
+    - openssl  # [linux]
+
+test:
+  commands:
+    - phanotate-rs --version
+    - phanotate-rs --help
+
```
thanhleviet, line 28: `build.sh` generates `THIRDPARTY.yml` via `cargo-bundle-licenses`, but it isn't listed here, so conda-build never packages it and the bundled third-party license info is discarded. For a GPL Rust package, include it:

```suggestion
  license_file:
    - LICENSE
    - THIRDPARTY.yml
```

thanhleviet (commented):
Thanks for the recipe! sha256 verified and all CI is green (Lint, Linux, OSX-64, ARM). Two things before merge:

1. `THIRDPARTY.yml` is generated by `build.sh` but not listed in `license_file`, so it isn't packaged. Please add it (see inline suggestion). This matters for GPL compliance on the bundled Rust dependencies.
2. `openssl` in `host`/`run` looks unnecessary - there's no openssl/TLS crate in `Cargo.lock`. Can it be dropped?

recipes/phanotate-rs/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+requirements:
+  build:
+    - {{ compiler('rust') }}
+    - {{ compiler('c') }}
+    - {{ stdlib('c') }}
+    - cargo-bundle-licenses    
+
+  host:
+    - openssl  # [linux]
```
thanhleviet: `openssl` appears unnecessary here (and in `run` below). The upstream `Cargo.lock` has no `openssl`/`native-tls`/`reqwest`/`git2` crates - dependencies are anyhow, atty, clap, indicatif, num-bigint, rayon. Unless a build-time `-sys` crate needs it, you can likely drop `openssl` from both `host` and `run`.

recipes/phanotate-rs/meta.yaml
Yasas1994: @thanhleviet thanks for fixing this

recipes/phanotate-rs/meta.yaml
Yasas1994: I removed it from the recipe

Yasas1994:
@thanhleviet please check if everything is correct

recipes/phanotate-rs/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - phanotate-rs --version
+    - phanotate-rs --help
+  license_file:
+    - LICENSE
+    - THIRDPARTY.yml
+about:
+  home: https://github.com/Yasas1994/PHANOTATE-rs
+  license: GPL-3.0-only
+  license_family: GPL3
+  license_file: LICENSE
```
thanhleviet: The `license_file` block ended up nested under `test:`, which conda-build ignores - so `THIRDPARTY.yml` still isn't bundled. It belongs under `about:`. This suggestion removes the stray block and adds both files to `about.license_file`:

```suggestion

about:
  home: https://github.com/Yasas1994/PHANOTATE-rs
  license: GPL-3.0-only
  license_family: GPL3
  license_file:
    - LICENSE
    - THIRDPARTY.yml
```

thanhleviet (commented):
Thanks for the updates - openssl removal looks good. One issue: the `license_file` suggestion got applied in the wrong place. After the rebase it landed nested under `test:` (where `license_file` does nothing), while `about.license_file` still lists only `LICENSE`. So `THIRDPARTY.yml` is still not packaged. The inline suggestion below moves it to the correct spot.

thanhleviet (approved):
LGTM. Both earlier points are resolved: `about.license_file` now bundles `LICENSE` + `THIRDPARTY.yml`, and the unused openssl deps are gone. sha256 verified against the upstream tarball, `Cargo.lock` present so `--locked` works, GPL-3.0-only license file shipped, `run_exports` pinned `max_pin="x.x"` for the 0.x series, build number 0. All CI green (Lint, Linux, OSX-64, ARM). Thanks for the recipe!

PR #65949, opened 2026-06-02
Telescope stranded mode

hanalysis:
This is a slight update to the telescope fix I added a couple of months ago - again, all code comes from the original contributors, but it updates to support stranded alignments.


hanalysis:
@BiocondaBot please add label

kbseah (commented):
Thanks for maintaining this for the community. I noted the source url is your fork of the original software but the URLs in the about section only point to the original. Could a brief explanation be added e.g. in a description section so that users are aware?

kbseah (approved):
Thanks for the update. Issues raised in previous review have been addressed. Looks good to me!

PR #65959, opened 2026-06-02
tir-learner: v4.04 — grfmite-rs dependency, CLI TIR-Learner + tirlearner4 alias

KGerhardt:
Updates the `tir-learner` recipe:

- Bumps source to the [v4.04 release](https://github.com/KGerhardt/TIR-Learner/releases/tag/v4.04), which adds `-k`/`--cnn_ratio` (CPU cores per CNN process; #processes = `-p // k`, fewer in-memory model copies / lower peak RAM; default 4).
- Installs the CLI wrapper as `TIR-Learner` instead of `tirlearner4` so that EDTA, which locates the tool via `command -v TIR-Learner`, resolves it.

Build number stays 0 (new version).

### DIFF
--- a/recipes/tir-learner/build.sh
+++ b/recipes/tir-learner/build.sh
@@ -8,10 +8,11 @@ cp -r TIR-Learner4/* "$PREFIX/lib/tir-learner4/"
 # Fix shebang in main script (original has #!/usr/app/env python3, a typo)
 sed -i '1s|.*|#!/usr/bin/env python3|' "$PREFIX/lib/tir-learner4/TIR-Learner.py"
 
-# Create CLI wrapper
+# Create CLI wrapper. TIR-Learner is the canonical command name, and what EDTA
+# locates via `command -v TIR-Learner`, so install directly under it.
 mkdir -p "$PREFIX/bin"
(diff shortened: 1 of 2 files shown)


KGerhardt:
@BiocondaBot please merge

kbseah:
Thanks for the update, @KGerhardt !

I see that you've changed the CLI name. Would it be better to add the new name as an alias, in case other users have already incorporated this tool into their workflows under the previous name?

FYI the bot command should be "please add label". I've added it manually now

KGerhardt:
I considered adding the alias in, but I had figured that compatibility with existing TIR-Learner installs was the primary target, I can likely add the tirlearner4 invocation in again, too.

KGerhardt:
Updated: v4.04 re-cut so GRF MITE detection runs via [`grfmite-rs`](https://anaconda.org/bioconda/grfmite-rs) (now on bioconda) instead of `grf-main`. Recipe changes: `genericrepeatfinder` → `grfmite-rs`, new v4.04 sha256, and a `tirlearner4` alias alongside the default `TIR-Learner` command (both tested).

KGerhardt:
@Juke34

Juke34 (approved):

PR #65977, opened 2026-06-03
Add recipe for count_logan_tig_coverage v0.1.0

pierrepeterlongo:
## Package

`count_logan_tig_coverage` — compute length-weighted mean and median coverage from Logan contig/unitig FASTA files.

Reads coverage values from the `ka:f:` tag present in [Logan](https://github.com/IndexThePlanet/Logan/) contig and unitig headers.

## Details

- Rust binary, builds with `cargo build --release`
- Source: https://github.com/pierrepeterlongo/count_logan_tig_coverage
- License: GPL-3.0-only
- Used by [logan_blaster](https://github.com/pierrepeterlongo/logan_blaster) to report coverage statistics per SRA accession

## Test

```bash
count_logan_tig_coverage --help
```

PR #65978, opened 2026-06-03
logan_blaster: add count_logan_tig_coverage as run dependency

pierrepeterlongo:
## Summary

Adds `count_logan_tig_coverage` as a run dependency of `logan_blaster`.

`count_logan_tig_coverage` computes length-weighted mean and median coverage from Logan contig/unitig FASTA files (reads the `ka:f:` tag in Logan headers). It is called by `logan_blaster` after each download and after each k-mer recruitment step to report per-accession coverage statistics.

The tool is optional at runtime (logan_blaster falls back gracefully with a warning if not found), but is listed as a run dependency so that conda installs include it automatically.

**Depends on**: bioconda/bioconda-recipes#65977 (`count_logan_tig_coverage` must be available first)


pierrepeterlongo:
@BiocondaBot please add label

pierrepeterlongo:
Hello there, 

Now that dependency https://github.com/bioconda/bioconda-recipes/pull/65977 (count_logan_tig_coverage) is available, would it be possible to merge this PR? 

Thanks!
Pierre

kbseah:
hi @pierrepeterlongo , I think you may have accidentally changed the recipe back to the original version with the latest commit, so no changes are actually proposed in this PR. Could you please look into this?

kbseah:
FYI syncing of the CDN has been laggy in the past few days, may have to wait a while before restarting the build https://github.com/bioconda/bioconda-recipes/issues/66614

pierrepeterlongo:
Thanks @kbseah 
I reached the same conclusions, but I wasn't sure.

kbseah:
hi @pierrepeterlongo there is a checksum mismatch, was there an update under the same version number?

pierrepeterlongo:
Sorry for the mistake @kbseah 
It should be fixed now. 
Thanks again, 
Pierre

kbseah (approved):
Thanks for updating the recipe! Looks good to me. Tests and build pass

pierrepeterlongo:
Super, thanks !

PR #65986, opened 2026-06-03
Update Fix libgsl.so.25 missing for bcftools

KasperThystrup:
On newer installations libgsl.so.25 is not always present which leads to an error. I downgraded dependency to ensure creation of the relevant file


KasperThystrup:
@bioconda/core
Uncertain how to fix the OSx64 AMR build issue. It states gsl 2.5 is not available however it is available in e.g. conda-forge.

Unsure whether ready for label

epruesse:
So is a pin of libgsl now necessary or not? Also, you changed the maxpin, is that necessary? IIRC bcftools is really a command line tool. Usually, those don't need x.x.x?

KasperThystrup:
@epruesse Thank you for looking through this PR. I'm sorry, I only changed the pin because I thought I had to be descriptive as to which version I was updating. Asking co-pilot for clarification, I seem to have misinterpreted the use of run exports.

If understood correctly, I should use run_exports to pin down only the major version (as cmd-line tools are not as important to keep track on a patch version basis.

Since there is a new version, I have updated to that and reverted back to the original maxpin.

KasperThystrup:
@BiocondaBot please add label

mencian (approved):

PR #66007, opened 2026-06-04
Add susiex

lyh970817:
Add a new Bioconda recipe for SuSiEx v1.1.2.

SuSiEx is a C++ command-line tool for cross-ancestry fine-mapping using GWAS summary statistics and LD reference panels, so it is directly relevant to statistical genetics workflows.

Packaging notes:
- Builds the upstream C++ executable from the tagged source archive rather than using the bundled static binary.
- Patches the upstream Makefile to use Conda's C++ compiler and build flags instead of hard-coded g++.
- Declares OpenMP runtime dependencies for Linux and macOS.
- Declares plink as a runtime dependency because SuSiEx shells out to PLINK for LD calculations.
- Includes run_exports with major-version pinning.

Validation performed locally:
- Ran the repository indentation fixer on the recipe.
- Ran git diff --check for the new recipe.
- Ran bash -n on build.sh.
- Ran a source-level build smoke test using the new build.sh; the installed SuSiEx executable printed the expected help banner.
- Ran bioconda-utils lint for susiex with only the channel-repodata checks excluded because this environment cannot reliably reach conda.anaconda.org; all remaining checks passed.

### DIFF
--- a/recipes/susiex/build.sh
+++ b/recipes/susiex/build.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+set -eux -o pipefail
+
+# Upstream hard-codes g++, so patch the Makefile to use Conda's compiler and flags.
+sed -i.bak \
+  -e 's|g++ main.cpp data.o model.o -O3|$(CXX) $(CXXFLAGS) $(CPPFLAGS) main.cpp data.o model.o $(LDFLAGS) -O3|' \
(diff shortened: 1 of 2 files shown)


lyh970817:
@BiocondaBot please add label

kbseah (approved):
thanks for the contribution! package is relevant to bioconda, license and tests are appropriate. looks good to me

PR #66038, opened 2026-06-05
Add cytospace-sc

LuisHeinzlmeier:
- add cytospace

### DIFF
--- a/recipes/cytospace-sc/meta.yaml
+++ b/recipes/cytospace-sc/meta.yaml
@@ -23,7 +23,7 @@ requirements:
     - pip
     - setuptools
   run:
-    - setuptools
+    - setuptools <82
     - python >=3.9
     - numpy >=1.26
     - scipy >=1.11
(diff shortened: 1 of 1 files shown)


LuisHeinzlmeier:
@BiocondaBot please add label

LuisHeinzlmeier:
This is my very first Bioconda recipe. I'd appreciate any feedback. It should only work on Linux because [ortools-python](https://anaconda.org/channels/conda-forge/packages/ortools-python/overview) is only available on Linux. Or are there ways to make it available for macOS as well?

LuisHeinzlmeier:

kbseah:
Thanks for the contribution! I'm not sure if the license is compatible though. It should allow redistribution, whereas the the license in the repo states that the license is non-transferable. Not a lawyer, so I can't say more

LuisHeinzlmeier:
Thank you @kbseah! [I reached out to the developers.](https://github.com/digitalcytometry/cytospace/issues/113)

LuisHeinzlmeier:
@kbseah, unfortunately, I haven't received a reply. How do we proceed?

kbseah:
If the license really does not permit redistribution, we can't package it for Bioconda. Perhaps you could check on the Bioconda gitter chat if anyone else has had a similar experience.

PR #66143, opened 2026-06-09
Update get_mnv to 1.1.4

Paururo:
Update get_mnv to version 1.1.4.

Changes:
- bump version from 1.1.3 to 1.1.4
- update the source sha256 for the 1.1.4 tag

Validation:
- verified source tarball sha256
- ran git diff --check

The existing recipe structure from the previous get_mnv update is unchanged.

### DIFF
--- a/recipes/get_mnv/meta.yaml
+++ b/recipes/get_mnv/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "get_mnv" %}
-{% set version = "1.1.3" %}
-{% set sha256 = "85116fddf639eefdf4244d870391748ccd93ac95de69ad224a3217cedfdb3fc8" %}
+{% set version = "1.1.4" %}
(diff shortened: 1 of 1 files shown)


Paururo:
@BiocondaBot please add label

kbseah (changes requested):
Thanks for the update! Could the recipe be updated to bundle dependency licenses, per the [current Bioconda guidelines](https://bioconda.github.io/contributor/guidelines.html#rust) for Rust recipes?

Paururo:
Thanks @kbseah! I've updated the recipe to bundle the dependency licenses with `cargo-bundle-licenses`, following the current Rust guidelines: it now generates `THIRDPARTY.yml` at build time and includes it in `license_file`. Could you take another look when you have a moment? Thanks!

kbseah (approved):
Thanks! Issue raised in previous review has been resolved. Looks good to me

PR #66154, opened 2026-06-09
Update grfmite-rs to 0.3.0

BiocondaBot:
Update [`grfmite-rs`](https://bioconda.github.io/recipes/grfmite-rs/README.html): **0.2.1** &rarr; **0.3.0**

[install with bioconda](http://bioconda.github.io/recipes/grfmite-rs/README.html) [Conda](https://anaconda.org/bioconda/grfmite-rs/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/grfmite-rs`](https://github.com//bioconda/bioconda-recipes/tree/bump/grfmite_rs/recipes/grfmite-rs) (click to view/edit other files)
Summary | Fast parallel Rust reimplementation of GRF MITE detection (grf-main -c 1), used by TIR-Learner.
Home | [https://github.com/KGerhardt/grfmite-rs](https://github.com/KGerhardt/grfmite-rs)
Releases |[https://github.com/KGerhardt/grfmite-rs/tags](https://github.com/KGerhardt/grfmite-rs/tags)
Recipe Maintainer(s) | @KGerhardt
Author | `@KGerhardt`
***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

PR #66096, opened 2026-06-08
Update tir-learner to 4.07

BiocondaBot:
Update [`tir-learner`](https://bioconda.github.io/recipes/tir-learner/README.html): **4.05** &rarr; **4.07**

[install with bioconda](http://bioconda.github.io/recipes/tir-learner/README.html) [Conda](https://anaconda.org/bioconda/tir-learner/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/tir-learner`](https://github.com//bioconda/bioconda-recipes/tree/bump/tir_learner/recipes/tir-learner) (click to view/edit other files)
Summary | Accelerated tool for identifying TIR transposable elements
Home | [https://github.com/KGerhardt/TIR-Learner](https://github.com/KGerhardt/TIR-Learner)
Releases |[https://github.com/KGerhardt/TIR-Learner/tags](https://github.com/KGerhardt/TIR-Learner/tags)
Author | `@KGerhardt`
***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

PR #66155, opened 2026-06-09
Update tir-learner to 4.07 (pin grfmite-rs >=0.3.0)

KGerhardt:
Supersedes the autobump #66096, which bumps to 4.07 but leaves `grfmite-rs` unpinned.

TIR-Learner v4.07 reads grfmite-rs's new streamed line format (`<base>.grf.txt`) candidate output. grfmite-rs **< 0.3.0** emits the previous columnar `candidate.json`, so tir-learner 4.07 + grfmite-rs 0.2.x is an installable-but-broken combination. This is identical to the autobump (version 4.07, same sha256) plus the required run pin `grfmite-rs >=0.3.0`. grfmite-rs 0.3.0 is in #66154 and should merge first.


KGerhardt:
@Juke34

I hope this is the last one.

Juke34 (approved):

PR #66193, opened 2026-06-10
Update gottcha2 entry point

poeli-lanl:
This PR fixes the `gottcha2` Bioconda package for version 2.4.0.

The current Bioconda recipe creates the console entry point:

```yaml
gottcha2 = gottcha.scripts.cmd:gottcha2_command
```

However, the upstream package now defines the entry point as:

```yaml
gottcha2 = gottcha.gottcha2:cli
```

Because the old `gottcha.scripts.cmd` module no longer exists in the current upstream source, installing `gottcha2` from Bioconda currently produces this runtime error:

```text
ModuleNotFoundError: No module named 'gottcha.scripts'
```

This PR updates the recipe to match the current upstream package layout.

Changes included:

* Updates the `gottcha2` entry point to `gottcha.gottcha2:cli`
* Bumps the build number from `0` to `1`
* Updates the Python requirement to match the upstream package
* Adds `pysam` as a runtime dependency
* Updates the test commands to verify the installed Python module and CLI
* Keeps the existing `run_exports` section unchanged

This is a recipe-only rebuild for the existing 2.4.0 release.

### DIFF
--- a/recipes/gottcha2/meta.yaml
+++ b/recipes/gottcha2/meta.yaml
@@ -7,25 +7,25 @@ package:
 
 build:
   noarch: python
-  number: 0
+  number: 1
   script: {{ PYTHON }} -m pip install -vvv --no-deps --no-build-isolation --no-cache-dir .
   entry_points:
-    - gottcha2 = gottcha.scripts.cmd:gottcha2_command
+    - gottcha2 = gottcha.gottcha2:cli
   run_exports:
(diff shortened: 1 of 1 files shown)


poeli-lanl:
@BiocondaBot please add label

kbseah (approved):
Thanks for fixing this. Actually the entry_points statement is unnecessary as it is already defined in the upstream pyproject.toml. Feel free to remove it for future versions

PR #66205, opened 2026-06-11
bump mehari and mehari-python to 0.43.3

tedil:
For some reason, autobump didn't pick these up correctly.

### DIFF
--- a/recipes/mehari-python/meta.yaml
+++ b/recipes/mehari-python/meta.yaml
@@ -14,7 +14,7 @@ build:
   number: 0
   skip: True  # [osx or py<312]
   run_exports:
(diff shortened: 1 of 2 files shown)


stolpeo (approved):
LGTM

dlaehnemann (approved):
This looks good to me. Once we have linting working for multi-output recipes, I would recommend to integrating these two recipes into one. That way you only have to update one recipe each time, with the same version and build number used between them. For reference, this is happening here:
https://github.com/bioconda/bioconda-utils/pull/1086

But we'll first test this new setup out with existing multi output recipes and can then revisit the recipe here.

PR #66237, opened 2026-06-12
Update iobrpy to 0.1.9

hhn0527:
Update `iobrpy` from 0.1.8 to 0.1.9.

Changes:
- Bump version to `0.1.9`
- Update PyPI sdist SHA256
- Reset build number to `0`

Local checks:
- `bioconda-utils lint recipes config.yml --packages iobrpy` passed
- `bioconda-utils build recipes config.yml --packages iobrpy` passed


hhn0527:
@BiocondaBot please add label

recipes/iobrpy/meta.yaml
kbseah, line 72: This appears to be a python package which doesn't require compilation. Shouldn't it be `noarch: python` under the `build` section?

kbseah (commented):
Thanks for the update! Could you please clarify why this isn't `noarch: python`?

recipes/iobrpy/meta.yaml
hhn0527: Thanks for the review. This package is not pure Python at the moment: it vendors Linux-specific native executables/shared libraries under `src/iobrpy/SpecHLA/bin/` such as `bcftools`, `blastn`, and `libwfa2.so.0`. Because those files are platform-specific, `noarch: python` would not be appropriate here. The recipe is therefore limited to `linux64` and built per Python version.

kbseah: i see, could a comment be added to the recipe to explain this? thank you

hhn0527: Thanks for the suggestion. I have added a comment in the `build` section explaining that this recipe is not `noarch: python` because the PyPI source distribution vendors Linux-specific executables/shared libraries under `src/iobrpy/SpecHLA/bin/`.

kbseah (approved):
New build adds comment to explain why recipe is not noarch:python

PR #66284, opened 2026-06-13
Add salmon-cpp: the final C++ salmon (1.12.0)

rob-p:
This adds a new recipe, **`salmon-cpp`**, packaging the final **C++** release of
salmon (`1.12.0`).

salmon 2.0 is a from-scratch **Rust** rewrite; the primary `salmon` recipe is
moving to that build (a separate autobump PR). `salmon-cpp` preserves the C++
line for reproducibility and emergency fixes.

Notes for reviewers:
- It builds the `v1.12.0` source tarball with the same CMake build the `salmon`
  1.12.0 recipe used, with the pufferfish dependency pinned to `5dce7f41` (the
  streaming `getRefPos` k-mer-orientation fix shipped in 1.12.0).
- It installs a `salmon` binary, so it **conflicts with the `salmon` package** by
  design — users install one or the other. (salmon-cpp is for those who need the
  C++ line specifically.)
- Maintainers: rob-p (salmon author), k3yavi, mjsteinbaugh.

@BiocondaBot please add label

### DIFF
--- a/recipes/salmon-cpp/build.sh
+++ b/recipes/salmon-cpp/build.sh
@@ -1,0 +1,102 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+BUILD_DIR="${SRC_DIR}/build"
+DEPS_PREFIX="${SRC_DIR}/_deps_prefix"
+THIRDPARTY_DIR="${SRC_DIR}/_thirdparty"
+CMAKE_SHIM_DIR="${SRC_DIR}/_cmake_shims"
+
+ZLIBNG_VERSION="2.3.3"
+ZLIBNG_URL="https://github.com/zlib-ng/zlib-ng/archive/refs/tags/${ZLIBNG_VERSION}.tar.gz"
+ZLIBNG_TARBALL="${THIRDPARTY_DIR}/zlib-ng-${ZLIBNG_VERSION}.tar.gz"
+ZLIBNG_SRC_DIR="${THIRDPARTY_DIR}/zlib-ng-${ZLIBNG_VERSION}"
+
+mkdir -p "${BUILD_DIR}" "${DEPS_PREFIX}" "${THIRDPARTY_DIR}" "${CMAKE_SHIM_DIR}"
+
+# Build zlib-ng (compat mode)
+curl -L --fail "${ZLIBNG_URL}" -o "${ZLIBNG_TARBALL}"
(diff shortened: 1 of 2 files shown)


rob-p:
Hi @bgruening,

Is there a good precedent for how we should handle this? The salmon rewrite is in rust and is mostly back compat but not fully to clear the way for future dev. That package is pushed as salmon now, but I wanted to maintain a separate salmon-cpp package for folks who need all of the long tail of legacy functionality. Since the executable names are currently the same, is there a recommended way to handle this?

Thanks!

bgruening:
@rob-p are you planning to maintain the 1.2.* line? I don't see any hard reasons to create a new package(-name).

You can maintain the 1.2.* line in the same folder and call it still `salmon`. We would need to communicate to the community that 1.2 is C++ and 2.0 is rust and ofc it has slightly different behavior .... but this is indicated in the 2.0 version. That would also solve the "don't install 1.2 and 2.0 in one env" problem naturally.

If you like that, just drop a subfolder in the salmon folder, something like recipes/salmon/1.2.x/

rob-p:
My thought process is that I'd like to retain the 1.12 line in case of bug fixes and other regular maintenance stuff, and make it easy for people who need the "legacy" version to install it.

I don't know much about (haven't messed with) multi-package builds in the past. Right now the C++ releases are based on a separate branch and will be cut from there while the new rust releases will be cut from master.  So this recipe tracks the `cpp` branch of the repo while the main salmon recipe tracks master.

bgruening:
I think you can do that without having the extra package `salmon-cpp`, just maintain the 1.2.* line and people need to explicitly `conda create --name salmon-legacy salmon=1.2`

rob-p:
That's cool, but do you know how I have releases trigger builds of the different versions/tags if they are the same recipe?

bgruening:
Its not the same recipe, its a sub-recipe. But no, I don't know if the auto-bump bot will work for the 1.2 release. it might be if you point it to the specific branch, never tested this. But would that be a problem? We don't assume many 1.2.* releases ideally, do we?

rob-p:
Thanks @bgruening — the subfolder approach makes sense

Regarding auto-bump: I dug into autobump.py and see subrecipes are excluded by default (ExcludeSubrecipe). This means that for the rare future bugfix a manual version bump to the subfolder will be required. The only downside to this is that I'll need, in that case, someone else besides myself to approve those bumps.

I'll open the new PR shortly. Thanks for the guidance!

PR #66285, opened 2026-06-13
Bump jaeger-bio to 1.26.4

Yasas1994:
Update jaeger-bio to version 1.26.4.


Yasas1994:
@BiocondaBot please add label

recipes/jaeger-bio/meta.yaml
```diff
@@ -1,48 +1,70 @@
...
+  url: "https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name | replace('-', '_') }}-{{ version }}.tar.gz"
+  # NOTE: update sha256 after releasing to PyPI
+  sha256: 262a896fd7254f30461155ab0157f75a350c3f2b9f7ad38c082273981af32dd7
 
 build:
   number: 0
   script: "{{ PYTHON }} -m pip install . --no-deps --no-build-isolation --no-cache-dir -vvv"
   run_exports:
-    - {{ pin_subpackage('jaeger-bio', max_pin="x") }}
+    - {{ pin_subpackage('jaeger-bio', max_pin="x.x.x") }}
```
kbseah: Why was the max_pin changed? Is it because the versioning does not follow semver?

recipes/jaeger-bio/meta.yaml
```diff
@@ -1,48 +1,70 @@
...
   noarch: python
 
 requirements:
   host:
-    - python >=3.9,<3.12
+    - python >=3.11,<3.14
```
kbseah: Could the upper python pin be relaxed?

kbseah (changes requested):
Thanks for the update! Some questions below, could you please have a look?

recipes/jaeger-bio/meta.yaml
Yasas1994: No strong justification for the max_pin change — happy to revert to "x".

recipes/jaeger-bio/meta.yaml
Yasas1994: I haven’t tested against Python 3.14 yet, but happy to remove the upper pin if that’s preferred.

kbseah: ok I see the upstream specifies <3.14, so pip won't install if that is removed

siebrenf (approved):

kbseah (approved):

PR #66335, opened 2026-06-15
Update percolator to 3.9

timosachsenberg:
Updates `percolator` 3.7.1 → 3.9.

### What changed
- **Drops the `xerces-c` / `xsd` dependency.** Upstream removed the XML/XSD code path from the percolator core ([percolator/percolator#399](https://github.com/percolator/percolator/pull/399), *"Remove XML/XSD support (incompatible with modern C++)"*). The XSD/Xerces-based converters (`sqt2pin`, `msgf2pin`, `tandem2pin`) no longer compile against modern `xerces-c`, so they are no longer built. The core `percolator` + `qvality` binaries (tab-delimited PIN/POUT I/O) are unaffected.
- Removes the obsolete `boost-dynamic.patch` — the `Boost_USE_STATIC_LIBS` line it patched no longer exists in 3.9's `CommonCMake.txt`.
- Drops the `WRITABLE_DIR` `sed` fixup (removed upstream) and the `-std=c++14` override (3.9 sets `CMAKE_CXX_STANDARD 17`).

### Why
Besides being the current release, this removes the `xerces-c <3.3.0` constraint that prevents `percolator` from being co-installed with packages built against `xerces-c 3.3.0` (e.g. OpenMS / `openms-thirdparty`).

### Notes
- The dropped converters were the only consumers of `xerces-c`/`xsd`; they could be re-added as a separate output later if needed.

---
Pull request checklist (bioconda):
- [x] Bumped version + reset build number
- [x] `sha256` verified against the upstream tarball
- [x] `run_exports` retained


recipes/percolator/meta.yaml
```diff
@@ -54,9 +47,6 @@ about:
 extra:
   additional-platforms:
     - linux-aarch64
-
-  container:
-    extended-base: True  # Add UTF-8 for percolator converters
```
bgruening, line 59: Utf-8 not a problem anymore?

timosachsenberg: I am pretty sure those are only needed for the xerces stuff that got removed

PR #66337, opened 2026-06-15
Add salmon 1.12 subrecipe: the final C++ salmon

rob-p:
Adds a version-pinned subrecipe `recipes/salmon/1.12/` that builds the **same `salmon` package** at the final C++ release (**1.12.0**), alongside the existing `recipes/salmon/` (the Rust 2.0 line).

This follows @bgruening's suggestion in #66284 to use a subfolder rather than a separate `salmon-cpp` package. Closing #66284 in favor of this.

- Same package name `salmon`; users install the legacy C++ line with `conda install "salmon=1.12"` (ideally in a dedicated env). `conda install salmon` continues to get the latest (Rust 2.0).
- Sourced from the immutable `v1.12.0` tag (the C++ tree also lives on the `cpp` branch for future maintenance).
- Marked `extra: autobump: enable: false` — subrecipes are excluded from autobump by default anyway; 1.12.0 is the final C++ release, so the rare future bugfix will be a manual version bump here. The top-level recipe keeps auto-bumping the 2.0 line.
- C++ build via CMake; pufferfish pinned to `5dce7f41` (the streaming `getRefPos` orientation fix shipped in 1.12.0).

Supersedes #66284.

### DIFF
--- a/recipes/salmon/1.12/build.sh
+++ b/recipes/salmon/1.12/build.sh
@@ -1,0 +1,102 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+BUILD_DIR="${SRC_DIR}/build"
+DEPS_PREFIX="${SRC_DIR}/_deps_prefix"
+THIRDPARTY_DIR="${SRC_DIR}/_thirdparty"
+CMAKE_SHIM_DIR="${SRC_DIR}/_cmake_shims"
+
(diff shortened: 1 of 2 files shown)


rob-p:
@BiocondaBot please add label (cc @bgruening following up on the separate recipe PR).

bgruening (approved):

PR #66362, opened 2026-06-16
Update oxbreaker to 1.1

rc-reding:
Version bump, plus helper to create a desktop shortcut after installation - so user can start the software with a double click instead.


rc-reding:
@BiocondaBot please add label

recipes/oxbreaker/meta.yaml
```diff
@@ -34,21 +33,22 @@ requirements:
...
+
 test:
   commands:
     - oxbreaker --help
     - nextflow -version
 
 about:
   home: "https://gitlab.com/rc-reding/OxBreaker"
-  license: "LGPL-2.1-only"
-  license_family: LGPL
```
bgruening, line 45: why are you removing the metadata and the dev_url

recipes/oxbreaker/oxbreaker.desktop
bgruening, line 1: Why is this Desktop stuff needed? Its unusal, just wondering :)

rc-reding: Sure. Since the command `oxbreaker` opens up a GUI, so the user can interact easier with the underlying NextFlow pipeline, I thought having the chance to create a desktop shortcut - so it shows up in the system menu - would help those less comfortable with the terminal :)

recipes/oxbreaker/meta.yaml
rc-reding: I missed this - sorry! Having two versions of the same file and edit them with different editors (neovim / helix) seems like a recipe for disaster.

rc-reding: `dev_url` and `home` are the same url

bgruening (approved):

recipes/oxbreaker/oxbreaker.desktop
bgruening: Oha

PR #66336, opened 2026-06-15
Add tirvish-rs 0.1.0

KGerhardt:
Adds a recipe for **tirvish-rs** (v0.1.0): a from-scratch Rust reimplementation of
GenomeTools' `gt tirvish` de-novo terminal-inverted-repeat (TIR) transposon
detection, used by TIR-Learner to replace the slow `gt tirvish` step on
repeat-rich genomes. Output is **bit-exact** with `gt tirvish` (all five stages
validated against an instrumented 1.6.5 reference).

- Source: https://github.com/KGerhardt/tirvish_rs (GPL-3.0-or-later), tag `v0.1.0`
- Builds with `cargo install --bin tirvish` (Rust + C compiler for the `libsais` dep); installs `tirvish` and a `tirvish_rs` alias.
- Same packaging pattern as the existing `grfmite-rs` recipe (same maintainer / project family).
- Recipe build verified locally.

PR #66375, opened 2026-06-16
Add tirvish-rs 0.1.0

KGerhardt:
Adds **tirvish-rs**, a faithful Rust reimplementation of GenomeTools' `gt tirvish` de-novo TIR transposon detection, used by TIR-Learner to replace the slow `gt tirvish` step on repeat-rich genomes.

- Output is validated bit-exact against `gt tirvish` on a 1153-element oracle (every per-element coordinate field + similarity).
- Runs directly on a FASTA (no pre-built mirrored suffix-array index) and accepts the `gt tirvish` options.
- Installs the `tirvish` binary (and a `tirvish_rs` alias).
- GPL-3.0-or-later; source: https://github.com/KGerhardt/tirvish_rs

(Re-opens the work from #66336, which was closed while the output format was finalized.)

### DIFF
--- a/recipes/tirvish-rs/build.sh
+++ b/recipes/tirvish-rs/build.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+set -euo pipefail
+
+# Build + install only the tirvish binary (the crate also carries dev-only
+# validator bins) into the conda prefix.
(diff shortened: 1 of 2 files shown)


KGerhardt:
@Juke34 @kbseah

Juke34 (approved):

PR #66378, opened 2026-06-16
Update methylsieve build number to 1 (fix missing linux-64)

tfenne:
## Why

`methylsieve` 0.1.0 was added in #66289. All PR checks passed (including Linux Tests), but the post-merge **Upload** workflow failed on the **Linux Upload** job:

```
Conflict: ('Owner bioconda already have a package named methylsieve', 409)
anaconda ... upload .../linux-64/methylsieve-0.1.0-h54198d6_0.conda
```

This was a first-publish race: because the package namespace did not yet exist on anaconda.org, the concurrent platform uploads (ARM via CircleCI, osx-64, linux-64) all tried to create it. One won; the linux-64 upload then hit a 409 and its `.conda` never landed. The channel currently has `osx-64`, `osx-arm64`, and `linux-aarch64` — but **not `linux-64`**.

## What

Bump the build number `0 → 1` to re-trigger the build/upload for all platforms. The namespace now exists, so the race cannot recur and linux-64 will publish.

No recipe content changes.

### DIFF
--- a/recipes/methylsieve/meta.yaml
+++ b/recipes/methylsieve/meta.yaml
@@ -10,7 +10,7 @@ source:
   sha256: 5b643592f4da414f4a5fc8fe0b92a69e67c40867aed86a51ee5b4ba3494b4861
 
 build:
-  number: 0
+  number: 1
   run_exports:
     - {{ pin_subpackage("methylsieve", max_pin="x.x") }}
 


tfenne:
@biocondabot please add label

nh13 (approved):

PR #66416, opened 2026-06-17
fastga: install ANOtoBED, BEDtoANO, ANOstat, ANOshow, FastKS

nekrut:
This adds the five binaries that FASTGA's `Makefile` already builds at `v1.5`
(they are in the `ALL` target, so `make` compiles them today) but that
`build.sh` does not install:

- **ANOtoBED**, **BEDtoANO**, **ANOstat**, **ANOshow** — the documented `.1ano`
  annotation toolset (see the FASTGA README command list). `ANOtoBED` in
  particular is needed by downstream tools that convert FASTGA/FasTAN
  annotation output to BED.
- **FastKS** — the standalone adaptamer k-mer table builder (has its own
  `Usage`).

No source/version bump and no patch are required — the tools are already
compiled by the existing `make` invocation; this only adds them to the
`install` list and adds `2>&1 | grep Usage` test commands. Build number bumped
`0 -> 1`.

`ONEalnTEST` is intentionally **not** installed: it is the ONElib/alncode API
demonstration program referenced by the README ("see the make command for
ONEalnTEST"), not a user-facing command.

---
##### Checklist
- [x] Recipe builds the added binaries (already in `ALL`).
- [x] `build.number` bumped.
- [x] Tests added for each new command.
- [ ] (maintainer) `@BiocondaBot please add label` if appropriate.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/recipes/fastga/build.sh
+++ b/recipes/fastga/build.sh
@@ -10,4 +10,5 @@ install -v -m 0755 FastGA \
     FAtoGDB GIXmake ALNtoPAF ALNtoPSL \
     GDBshow GDBstat GIXshow ALNshow ALNplot \
     GDBtoFA GIXrm GIXcp GIXmv ALNchain ALNreset PAFtoALN PAFtoPSL \
-    ONEview $PREFIX/bin
+    ONEview FastKS \
+    ANOtoBED BEDtoANO ANOstat ANOshow $PREFIX/bin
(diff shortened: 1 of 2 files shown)


kbseah (approved):
Thanks for the update

PR #66454, opened 2026-06-19
minibwa: statically link vendored zlib-ng for faster gzip input

nh13:
minibwa reads gzipped FASTQ through a single-threaded `kseq`/`gzread` path. At high thread counts one inflate stream becomes the bottleneck that starves the mappers. This builds zlib-ng in `ZLIB_COMPAT` mode (a drop-in `libz`/`zlib.h` with a much faster inflate) and statically links it in place of the stock `-lz`, with no source changes to minibwa.

Why vendored + static rather than a dependency: the conda-forge `zlib-ng-compat` package (the `track_features` anchor that swaps the system zlib) is not on the main channel, so it can't be used as a bioconda dependency. The plain `zlib-ng` package only ships the `zng_`-prefixed API, not a drop-in `libz`. Vendoring zlib-ng in compat mode is the same approach the `salmon` recipe uses. zlib-ng is declared as a second `source:` entry (pinned by sha256) so the build stays hermetic — no network fetch in `build.sh`.

Validation (on `linux-aarch64`, Graviton4):
- Built with `bioconda-utils build`; the binary statically embeds zlib-ng (no dynamic `libz`).
- Produces byte-identical SAM output vs the stock-zlib build (md5-identical over a 200k-read set).
- ~7% faster at 96 threads on a 20M read-pair hg38 alignment (40.7s vs 43.5s, 3 runs each, ±<0.4s); the gap scales with thread count (tied at 16/32t, 1.3% at 64t, 7.0% at 96t), confirming the inflate bottleneck. Measured on the v0.1 binary; the gzip I/O path is unchanged in v0.2.

Build number bumped to 1 (recipe change, same version).

### DIFF
--- a/recipes/minibwa/build.sh
+++ b/recipes/minibwa/build.sh
@@ -4,15 +4,49 @@ set -xe
 
 mkdir -p "${PREFIX}/bin" "${PREFIX}/lib" "${PREFIX}/include"
 
+# Build the vendored zlib-ng (fetched as the `_zlib_ng` source) in ZLIB_COMPAT
+# mode: a static, drop-in libz.a + zlib.h with a faster inflate. Static so the
+# speedup is baked in without adding a runtime dependency or a track_features
+# anchor.
+DEPS="${SRC_DIR}/_deps"
+mkdir -p "${DEPS}"
+cmake -S "${SRC_DIR}/_zlib_ng" -B "${SRC_DIR}/_zlib_ng/build" -G Ninja \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_PREFIX="${DEPS}" \
+    -DZLIB_COMPAT=ON \
+    -DBUILD_SHARED_LIBS=OFF \
+    -DCMAKE_POSITION_INDEPENDENT_CODE=ON \
+    -DZLIB_ENABLE_TESTS=OFF \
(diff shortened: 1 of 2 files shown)


bgruening:
> s not on the main channel, so it can't be used as a bioconda dependency. 


What is the main channel here? Anaconda main channel?

nh13:
Sorry, "main channel" was sloppy — I meant Anaconda `defaults`, but that's beside the point since bioconda builds against conda-forge, where `zlib-ng-compat` does exist.

The real reason to vendor + static-link rather than depend on `zlib-ng-compat`: it ships its own `libz` and is meant to *replace* `zlib` globally, so pulling it into one recipe's `host` puts two `libz` providers in the solve and collides with anything else that pulls in `zlib` (which bioconda's pinning targets). Static-linking bakes the faster inflate into the binary — no runtime dep, no chance the solver swaps back to the shared `zlib`. Same approach as `recipes/salmon/1.12/build.sh`, just more hermetic (sha256-pinned second `source:` instead of a `curl`).

Rebased onto 0.3 (build number → 2), which also clears the conflict.

nh13:
@BiocondaBot please add label

bgruening (approved):

bgruening:
@nh13 thanks!

PR #66467, opened 2026-06-19
Add cs-tag

jiangyun-fun:
## New recipe: **cs-tag** v0.2.2

`cs-tag` is a Rust CLI that generates minimap2-like **CS tags** for BAM files — it computes the compact query/reference alignment string from an existing BAM alignment plus a reference FASTA and writes the result as an auxiliary tag into the output BAM.

- **Upstream:** https://github.com/jiangyun-fun/cs-tag
- **License:** MIT
- **Tests:** `cs-tag --help`, `cs-tag --version`
- **Maintainer:** @jiangyun-fun

### Build notes
cs-tag depends on `rust-htslib` 1.0.0, which forces the `bindgen` feature on `hts-sys`. Running bindgen at build time produces opaque struct bindings under conda's cross-compilation sysroot, so `build.sh`:

1. `cargo vendor`s the dependency tree,
2. patches `vendor/rust-htslib/Cargo.toml` to drop the `bindgen` feature so `hts-sys` uses its **pre-built bindings**,
3. fixes two type mismatches in those pre-built bindings (`size_t` → `usize`; the `bam1_core_t.isize` field rename) and refreshes the vendored `.cargo-checksum.json`,
4. builds with `cargo install` and bundles third-party licenses via `cargo-bundle-licenses`.

This is the same recipe already building successfully on conda-forge staged-recipes.

Scoped to **Linux** (`skip: not linux`) because the pre-built bindings path is Linux-specific; macOS support can follow in a follow-up PR.

Checklist:
- [x] Package not already in bioconda
- [x] Linter passes locally (`bioconda-utils` / recipe structure follows existing rust-htslib recipes: longshot, wgatools, var-agg)
- [x] License and third-party licenses bundled

### DIFF
--- a/recipes/cs-tag/build.sh
+++ b/recipes/cs-tag/build.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+export CARGO_PROFILE_RELEASE_STRIP=symbols
+export CARGO_PROFILE_RELEASE_LTO=fat
+
+# Tell bindgen where to find libclang
+export LIBCLANG_PATH="${BUILD_PREFIX}/lib"
+
+# The Cargo project lives in the nested cs-tag/ directory of the source tree.
+cd cs-tag
+
+# Vendor all dependencies locally so we can patch them.
+mkdir -p .cargo
+cargo vendor vendor/ > .cargo/config.toml
+
+# rust-htslib 1.0.0 hardcodes features = ["bindgen"] on hts-sys in its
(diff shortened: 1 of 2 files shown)


jiangyun-fun:
Closing in favor of the renamed package **cs-tagger** (see replacement PR above). The upstream project was renamed cs-tag → cs-tagger (repo + Cargo package + binary) and re-released as v0.3.0, so this cs-tag recipe is superseded. Thanks!

PR #66471, opened 2026-06-19
Add cs-tagger

jiangyun-fun:
## New recipe: **cs-tagger** v0.3.0

`cs-tagger` (renamed from `cs-tag`) is a Rust CLI that generates minimap2-like **CS tags** for BAM files — it computes the compact query/reference alignment string from an existing BAM alignment plus a reference FASTA and writes it as an auxiliary tag into the output BAM.

- **Upstream:** https://github.com/jiangyun-fun/cs-tagger
- **License:** MIT
- **Tests:** `cs-tagger --help`, `cs-tagger --version`
- **Maintainer:** @jiangyun-fun

> Note: this supersedes #66467, which added the same tool under its previous name `cs-tag`. The upstream project was renamed to `cs-tagger` (repo + Cargo package + binary) and re-released as v0.3.0; this PR tracks the new name.

### Build notes
cs-tagger depends on `rust-htslib` 1.0.0, which forces the `bindgen` feature on `hts-sys`. Running bindgen at build time produces opaque struct bindings under conda's cross-compilation sysroot, so `build.sh`:

1. `cargo vendor`s the dependency tree,
2. patches `vendor/rust-htslib/Cargo.toml` to drop the `bindgen` feature so `hts-sys` uses its **pre-built bindings**,
3. fixes two type mismatches in those pre-built bindings (`size_t` → `usize`; the `bam1_core_t.isize` field rename) and refreshes the vendored `.cargo-checksum.json`,
4. builds with `cargo install` and bundles third-party licenses via `cargo-bundle-licenses`.

This is the same recipe that built successfully on conda-forge staged-recipes (and as `cs-tag` in #66467, where all checks were green).

Scoped to **Linux** (`skip: not linux`) because the pre-built bindings path is Linux-specific; macOS support can follow.

Checklist:
- [x] Package not already in bioconda
- [x] License and third-party licenses bundled

### DIFF
--- a/recipes/cs-tagger/build.sh
+++ b/recipes/cs-tagger/build.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+export CARGO_PROFILE_RELEASE_STRIP=symbols
+export CARGO_PROFILE_RELEASE_LTO=fat
+
+# Tell bindgen where to find libclang
+export LIBCLANG_PATH="${BUILD_PREFIX}/lib"
+
+# The Cargo project lives in the nested cs-tagger/ directory of the source tree.
+cd cs-tagger
+
+# Vendor all dependencies locally so we can patch them.
+mkdir -p .cargo
+cargo vendor vendor/ > .cargo/config.toml
+
+# rust-htslib 1.0.0 hardcodes features = ["bindgen"] on hts-sys in its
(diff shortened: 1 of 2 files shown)


jiangyun-fun:
@BiocondaBot please add label

bgruening (approved):

PR #66594, opened 2026-06-22
Add lifton 1.0.13

Kuanhao-Chao:
This PR adds a new recipe for **LiftOn** v1.0.9.

[LiftOn](https://github.com/Kuanhao-Chao/LiftOn) is a homology-based genome-annotation lift-over tool that combines DNA-level alignment (minimap2 / Liftoff) with protein-level alignment (miniprot) via a protein-maximization CDS-chaining algorithm and an ORF-rescue pass, producing more accurate lifted annotations than either signal alone. It also installs a standalone GFF3 validator (`gff3-validate`).

- New recipe: `recipes/lifton/meta.yaml`
- `noarch: python`; console entry points `lifton` and `gff3-validate`
- Source: PyPI sdist (https://pypi.org/project/lifton/1.0.9/), `sha256` pinned
- All run dependencies are available on conda-forge/bioconda; `minimap2` and `miniprot` are included as run deps because LiftOn invokes them as subprocesses
- Publication: https://doi.org/10.1101/gr.279620.124


rzelle-lallemand:
@BiocondaBot please add label

(per https://bioconda.github.io/contributor/workflow.html#create-a-pull-request )

Kuanhao-Chao:
Bumped this recipe from 1.0.9 to **1.0.13**, which is the current release and the one that should land first.

What changed since the original submission:

- **`cigar` removed.** It was dropped from LiftOn in v1.0.10 because its sdist could not be built in a clean environment at all (its `setup.py` downloads a setuptools egg from a URL that no longer serves one).
- **`setuptools <81` host cap removed**, now `setuptools >=61`. That cap existed only for `cigar`.
- **`mappy` became an optional extra upstream in v1.0.13** so that a plain `pip install lifton` needs no C compiler ([#78](https://github.com/Kuanhao-Chao/LiftOn/issues/78)). Conda has a prebuilt `mappy`, so it stays a run dependency here and the test asserts the native binding imports.
- **`minimap2` and `miniprot` remain run dependencies.** LiftOn invokes them as subprocesses and pip cannot install them, which is the other half of #78 — a conda install is the one-step route for users. The tests now check both answer `--version`.
- `sha256` is from the sdist PyPI serves for 1.0.13; `doc_url` updated.

`bioconda-utils lint recipes config.yml --packages lifton` → **All checks OK**.

Thanks @rzelle-lallemand for flagging the missing review label.

rzelle-lallemand:
@Kuanhao-Chao, thanks!

(your second GitHub issue link is currently to # 78 in this bioconda-recipes repo, but should be to https://github.com/Kuanhao-Chao/LiftOn/issues/78 ).

PR #65470, opened 2026-05-16
Update bwa-mem3 to 0.2.1; use libsais and sse2neon from conda-forge

nh13:
Bumps bwa-mem3 from 0.2.0 → 0.2.1 and switches two bundled submodules to conda-forge packages.

### Version bump (0.2.0 → 0.2.1)

Upstream 0.2.1 drops the bundled `intel/safestringlib` submodule ([fg-labs/bwa-mem3#105](https://github.com/fg-labs/bwa-mem3/pull/105): −278 lines, replace `*_s` wrappers with plain C, delete the `memcpy_bwamem` chunking wrapper). Reflected in the recipe by dropping `ext/safestringlib/LICENSE` from `license_file` and removing the `SAFE_EXTRA_CFLAGS` workaround from `build.sh`. 0.2.1 also picks up two bug fixes (mapq propagation in `--supp-rep-hard-cap`, smem `enc_qdb` byte-capacity tracking) and an htslib build improvement.

### libsais (linked dependency)

- conda-forge `libsais` 2.10.4 across `linux-64`, `linux-aarch64`, `osx-64`, `osx-arm64`, `win-64`. The conda-forge build pins the same upstream v2.10.4 commit that the bwa-mem3 submodule pins, so the ABI is identical.
- conda-forge's `libsais` is built with `LIBSAIS_USE_OPENMP=ON` and `LIBSAIS_BUILD_SHARED_LIB=ON`; its `libsais.{so,dylib}` carries a SONAME dependency on `libgomp` / `libomp`. bwa-mem3's Makefile already defines `-DLIBSAIS_OPENMP` and injects OpenMP compile/link flags, so OpenMP-parallel suffix-array construction continues to work unchanged.

### sse2neon (header-only, ARM only)

- conda-forge `sse2neon` 1.9.1 is bit-identical to bwa-mem3's bundled `ext/sse2neon` submodule (commit `92f6de1` = `v1.9.1` tag). The resulting ARM binary is unchanged; this is a packaging-only change for consistency with the libsais swap.
- Header-only with no runtime ABI; gated to ARM via the `# [aarch64 or arm64]` selector.
- conda-forge installs the header at `${PREFIX}/include/sse2neon/sse2neon.h` (subdir), so `SSE2NEON_INCLUDES` points one level deeper to let bwa-mem3's `#include "sse2neon.h"` resolve.
- `ext/sse2neon/LICENSE` stays in `license_file` — the header is still inlined into the binary at compile time, so the MIT attribution remains required.

[...]

PR #66643, opened 2026-06-24
Add python-minibwa 0.1.0

nh13:
Adds a recipe for **python-minibwa** — Python bindings (pyo3 + maturin) for [minibwa](https://github.com/lh3/minibwa), Heng Li's lightweight bwa-mem successor. Source: https://github.com/fg-labs/minibwa-bindings (PyPI: `minibwa`).

The conda package is named `python-minibwa` because the `minibwa` name is already taken by the lh3 C tool recipe; the Python import remains `import minibwa`.

### Notes for reviewers

- **Source is the full-repo tarball, not a maturin sdist.** `minibwa-py` depends on the sibling `minibwa`/`minibwa-sys` crates by path (and `minibwa-sys` vendors the minibwa C), so an sdist of the Python package alone would be unbuildable. The recipe builds from the workspace release archive and `cd minibwa-py`.
- **License is `MIT AND Apache-2.0`.** The default build statically links the vendored minibwa C (MIT, Dana-Farber) and libsais (Apache-2.0) into the extension; `license_file` ships `LICENSE`, `THIRD-PARTY.md`, and the vendored upstream notices. The GPL `bwtgen` path is an opt-in Cargo feature (`minibwa-sys` `default = []`) and is **not** enabled.
- **libsais is vendored rather than taken from conda-forge.** Unlike bwa-mem3 (#65470), `minibwa-sys`'s `build.rs` compiles `libsais.c`/`libsais64.c` inline via the `cc` crate with no env hook to substitute an external library, so unbundling would require an upstream `build.rs` change. The Apache-2.0 text is packaged.
- Build uses `--locked` (via `MATURIN_PEP517_ARGS`) against the committed `Cargo.lock`.
- `additional-platforms`: `linux-aarch64`, `osx-arm64` (matches the project's published wheel targets).

---

- [x] I am the original author / maintainer of this software (recipe-maintainer: `nh13`).

@BiocondaBot please add label

### DIFF
--- a/recipes/python-minibwa/build.sh
+++ b/recipes/python-minibwa/build.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+# minibwa-py is the pyo3/maturin package. It depends on the sibling `minibwa`
+# and `minibwa-sys` crates by path (and minibwa-sys vendors the minibwa C), so
+# build from within its subdirectory of the workspace source tree.
+cd minibwa-py
+
+# minibwa-sys generates its FFI via bindgen, whose clang-sys backend needs to
+# locate libclang.so at build time. The conda `clangdev` build dep installs it
+# under ${BUILD_PREFIX}/lib; point bindgen at it explicitly.
+export LIBCLANG_PATH="${BUILD_PREFIX}/lib"
+
+# minibwa-sys emits `cargo:rustc-link-lib=z`. On Linux rustc's link step does
+# not search the host prefix by default (it ignores conda's LDFLAGS), so `-lz`
+# fails to resolve; add ${PREFIX}/lib explicitly. On macOS we deliberately do
+# NOT add it: `-lz` then resolves against the system libSystem zlib, which
+# maturin leaves alone when bundling. Linking conda's libz.1.dylib instead
+# breaks maturin's wheel repair ("@rpath/libz.1.dylib could not be located").
+# (`pthread`/`m` resolve from the compiler sysroot on both platforms.)
+if [[ "$(uname)" != "Darwin" ]]; then
+  export RUSTFLAGS="${RUSTFLAGS:-} -L native=${PREFIX}/lib"
+fi
+
+# Build the cargo dependency graph from the committed Cargo.lock (`--locked`)
+# so the wheel pins the exact crate versions the project tests against, rather
+# than re-resolving at build time. maturin forwards this cargo flag, and pip
+# (PEP 517) forwards maturin args via MATURIN_PEP517_ARGS.
+export MATURIN_PEP517_ARGS="--locked"
+
+${PYTHON} -m pip install . --no-deps --no-build-isolation -vv
--- a/recipes/python-minibwa/meta.yaml
+++ b/recipes/python-minibwa/meta.yaml
@@ -1,0 +1,71 @@
+{% set name = "python-minibwa" %}
+{% set version = "0.1.0" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  # Full-repo tarball: the Python package (minibwa-py) depends on the sibling
+  # `minibwa`/`minibwa-sys` crates via path, and minibwa-sys vendors the
+  # minibwa C source. A maturin sdist of minibwa-py alone would miss those, so
+  # the recipe builds from the release source archive of the whole workspace.
+  url: https://github.com/fg-labs/minibwa-bindings/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 6d506c604b7f81ee05e4aae8c666b609f99868f437119fc0fad61bb216f5edf3
+
+build:
+  number: 0
+  run_exports:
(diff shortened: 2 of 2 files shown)


nh13:
@BiocondaBot please add label

recipes/python-minibwa/build.sh
```diff
@@ -0,0 +1,32 @@
...
+if [[ "$(uname)" != "Darwin" ]]; then
+  export RUSTFLAGS="${RUSTFLAGS:-} -L native=${PREFIX}/lib"
+fi
+
+# Build the cargo dependency graph from the committed Cargo.lock (`--locked`)
+# so the wheel pins the exact crate versions the project tests against, rather
+# than re-resolving at build time. maturin forwards this cargo flag, and pip
+# (PEP 517) forwards maturin args via MATURIN_PEP517_ARGS.
+export MATURIN_PEP517_ARGS="--locked"
+
```
clintval, line 31:
```suggestion

cargo-bundle-licenses --format yaml --output THIRDPARTY.yml

```

recipes/python-minibwa/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+about:
+  home: https://github.com/fg-labs/minibwa-bindings
+  # The default build statically links libsais (Apache-2.0) and the vendored
+  # minibwa C (MIT) into the extension. The GPL bwtgen path is an opt-in Cargo
+  # feature (minibwa-sys `default = []`) and is NOT enabled here.
+  license: MIT AND Apache-2.0
+  license_family: MIT
+  license_file:
+    - LICENSE
+    - THIRD-PARTY.md
```
clintval, line 60:
```suggestion
    - THIRD-PARTY.md
    - THIRD-PARTY.yml
```

recipes/python-minibwa/build.sh
```diff
@@ -0,0 +1,32 @@
...
+# (`pthread`/`m` resolve from the compiler sysroot on both platforms.)
+if [[ "$(uname)" != "Darwin" ]]; then
+  export RUSTFLAGS="${RUSTFLAGS:-} -L native=${PREFIX}/lib"
+fi
+
+# Build the cargo dependency graph from the committed Cargo.lock (`--locked`)
+# so the wheel pins the exact crate versions the project tests against, rather
+# than re-resolving at build time. maturin forwards this cargo flag, and pip
+# (PEP 517) forwards maturin args via MATURIN_PEP517_ARGS.
+export MATURIN_PEP517_ARGS="--locked"
```
clintval, line 30: Claude says to add `--aduitwheel skip` so you can use the conda-provided zlib directly. Worth a shot?

recipes/python-minibwa/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+    - minibwa
+  commands:
+    # Exercise the compiled pyo3 extension, not just the import machinery.
+    - python -c "import minibwa; minibwa.Opts()"
+
+about:
+  home: https://github.com/fg-labs/minibwa-bindings
+  # The default build statically links libsais (Apache-2.0) and the vendored
+  # minibwa C (MIT) into the extension. The GPL bwtgen path is an opt-in Cargo
+  # feature (minibwa-sys `default = []`) and is NOT enabled here.
```
clintval, line 55: In `license_file` you aren't actualy providing a copy of the `libsais` Apache license. Upstream should be encouraged to vendor the license *next to* `libsais.c` or you could add the license file here directly as as a source.

clintval (approved):
Looks good with a few suggestions to enhance the recipe.

recipes/python-minibwa/build.sh
nh13: Followed up in bioconda/bioconda-recipes#66657, but I left this one out on purpose — here's the reasoning.

`--auditwheel skip` controls maturin's *post-link bundling* step, not how `rustc` *resolves* `-lz` at link time. So it wouldn't actually remove the platform-conditional `RUSTFLAGS` block: on Linux, rustc still ignores conda's `LDFLAGS`, so `-L native=${PREFIX}/lib` is still needed for the link to succeed at all — that's a linker concern, a layer below bundling.

It also wouldn't cleanly "use conda's zlib on macOS." The current macOS path deliberately links system libSystem zlib so nothing dangling ships. Switching to "link conda's `libz.1.dylib` but skip the repair that bundles/relinks it" would leave a bare `@rpath/libz.1.dylib` dependency — trading a loud, reproducible build-time failure for a quiet, host-dependent runtime one. And `--auditwheel`'s semantics are manylinux-shaped; macOS wheel repair is a separately re-implemented delocate path (maturin#3114), so it's not safe to assume `skip` governs both backends symmetrically — that asymmetry is exactly what produced the `@rpath/libz.1.dylib could not be located` break the comment in build.sh documents.

Net: it'd cost a few verification steps on a real osx-arm64 build to remove ~4 lines of working, already-debugged shell. Happy to revisit if you've seen `--auditwheel skip` work cleanly for a maturin recipe that links a conda C lib on macOS.

PR #66647, opened 2026-06-24
Add plastanno 2.0.0

maidinhvn:
New recipe for Plastanno v2 — a dual-evidence, confidence-aware chloroplast genome (plastome) annotation pipeline. noarch python; external tools (blast, exonerate, hmmer, aragorn) as run dependencies; the ~266 MB reference database is fetched post-install via `plastanno fetch-db` (not bundled). Tests cover import + `plastanno --help`.

### DIFF
--- a/recipes/plastanno/meta.yaml
+++ b/recipes/plastanno/meta.yaml
@@ -33,8 +33,7 @@ requirements:
     - exonerate
     - hmmer
     - aragorn
-  run_constrained:
-    - trnascan-se        # optional, only for the --trnascan source
+    - trnascan-se        # only used by the optional --trnascan tRNA source
 
 test:
   imports:
--- a/recipes/plastanno/meta.yaml
+++ b/recipes/plastanno/meta.yaml
@@ -1,0 +1,59 @@
+{% set name = "plastanno" %}
+{% set version = "2.0.0" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/maidinhvn/plastanno/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 092762cdc9b93ef769566615ca1901ce41437200625fac413a07db1af886c77f
+
+build:
+  noarch: python
+  number: 0
+  script: {{ PYTHON }} -m pip install . --no-deps --no-build-isolation -vv
+  entry_points:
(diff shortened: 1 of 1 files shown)


maidinhvn:
@BiocondaBot please add label

recipes/plastanno/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - biopython
+    - pandas
+    - matplotlib-base
+    - platformdirs
+    - blast
+    - exonerate
+    - hmmer
+    - aragorn
+  run_constrained:
+    - trnascan-se        # optional, only for the --trnascan source
```
bgruening: I think you need to add this under `run`. run_contrained is something slightly different: https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-constrained

maidinhvn: Thanks @bgruening — you're right. Moved trnascan-se to run in 7c0a22cc. It's only used by the optional --trnascan tRNA source (default uses aragorn), but as a real runtime tool it belongs under run, not run_constrained.

maidinhvn:
Hi @bgruening, thanks again for the review. I've moved trnascan-se to run as you suggested (commit 7c0a22c) and resolved the conversation. All checks are green. Could you take another look when you have a moment? 🙏

recipes/plastanno/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  url: https://github.com/maidinhvn/plastanno/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 092762cdc9b93ef769566615ca1901ce41437200625fac413a07db1af886c77f
+
+build:
+  noarch: python
+  number: 0
+  run_exports:
+    - {{ pin_subpackage("plastanno", max_pin="x") }}
+  script: {{ PYTHON }} -m pip install . --no-deps --no-build-isolation -vv
+  entry_points:
```
kbseah, line 18: just a comment: strictly speaking the entry_points statement can be removed because it is defined in the upstream pyproject.toml

if you refactor the code later, this would be one less thing to update in the recipe.

kbseah (approved):
thanks for the contribution! Package and license are appropriate for bioconda, and tests pass. looks good to me

PR #66669, opened 2026-06-25
Fix MoGAAAP recipe

dirkjanvw:
Because `pip` for `mogaaap` recipe has `--no-deps` specified to have the dependencies handled by conda, the dependencies should be specified in the recipe. This was not done previously, so adding with this PR.


dirkjanvw:
@BiocondaBot please add label

kbseah (approved):
Thanks for the update

PR #66687, opened 2026-06-25
chore: Remove dead scripts

paperbenni:
this removes dead python scripts. the functionality of these has been moved elsewhere and they import things from bioconda-utils which no longer exists.


aliciaaevans:
I'm all for removing things that aren't useful any more, but missingCranPackages.py might still be useful for checking on R migrations. Example: https://github.com/bioconda/bioconda-recipes/issues/49778

paperbenni:
> I'm all for removing things that aren't useful any more, but missingCranPackages.py might still be useful for checking on R migrations. Example: #49778

I didn't notice that one wasn't moved to bioconda-utils, I restored it

aliciaaevans (approved):

PR #66699, opened 2026-06-26
Add bammmotif2

ubercomrade:
This PR adds a Bioconda recipe for BaMMmotif2 v2.0.

BaMMmotif2 is a bioinformatics command-line tool for de novo motif discovery using higher-order Bayesian Markov models. This recipe packages the core C++ command-line tools:

* `BaMMmotif`
* `BaMMScan`
* `FDR`

The recipe intentionally does not add the optional R plotting/logo dependencies. The goal of this package is to provide the core motif discovery and scanning algorithms for command-line use and for downstream Python wrappers.

## Build notes

The upstream v2.0 `CMakeLists.txt` requires a few small conda-specific compatibility adjustments during the build:

* update `cmake_minimum_required(VERSION 2.8.11)` to `cmake_minimum_required(VERSION 3.5)`, because current CMake versions no longer support compatibility with CMake versions older than 3.5;
* use shared Boost from conda-forge instead of static Boost;
* use `Boost_INCLUDE_DIRS`;
* preserve conda-build `CXXFLAGS` instead of overwriting `CMAKE_CXX_FLAGS`.

These substitutions are applied in `build.sh` with `sed` because the upstream v2.0 `CMakeLists.txt` is formatted as a single long line, which makes a normal multi-line patch fragile.

The changes are limited to build-system compatibility and should not affect the BaMMmotif2 algorithms.

I can open an upstream PR for the CMake compatibility changes if preferred; for this recipe they are kept minimal and limited to conda build-system compatibility.

## run_exports

The recipe includes:

```yaml
build:
  run_exports:
    - {{ pin_subpackage('bammmotif2', max_pin="x.x") }}
```

I used `max_pin="x.x"` conservatively because BaMMmotif2 provides command-line tools rather than a library API, and downstream recipes are most likely to depend on the CLI behavior and output formats. A minor-version change could plausibly introduce CLI or output-format changes that affect downstream wrappers or workflows.

## Tests

The recipe includes smoke tests that verify the installed executables are present:

* `BaMMmotif`
* `BaMMScan`
* `FDR`

The test also captures `BaMMmotif --help` output and checks for expected usage/help text.

`BaMMmotif --help` prints the help text but returns a non-zero exit code because it first reports missing required positional arguments, so the test validates the output rather than relying on the exit status.

## Local validation

I tested the recipe locally with:

```bash
conda build recipes/bammmotif2 -c conda-forge -c bioconda
bioconda-utils lint --packages bammmotif2 recipes config.yml
bioconda-utils build --docker --mulled-test --packages bammmotif2 recipes config.yml
```

The Docker/mulled test passed locally.

### DIFF
--- a/recipes/bammmotif2/build.sh
+++ b/recipes/bammmotif2/build.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+set -euxo pipefail
+
+# Upstream v2.0 has CMakeLists.txt formatted as a single long line, so a
+# normal multi-line patch is fragile. Keep these conda-specific edits here.
+sed -i.bak \
+  -e 's/cmake_minimum_required (VERSION 2.8.11 FATAL_ERROR)/cmake_minimum_required (VERSION 3.5 FATAL_ERROR)/' \
+  -e 's/set (Boost_USE_STATIC_LIBS ON CACHE BOOL "use static libraries from Boost")/set (Boost_USE_STATIC_LIBS OFF CACHE BOOL "use shared libraries from Boost")/' \
+  -e 's/include_directories (${Boost_INCLUDEDIR})/include_directories (${Boost_INCLUDE_DIRS})/' \
+  -e 's/set (CMAKE_CXX_FLAGS "-std=c++11 -Wall")/set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall")/' \
+  CMakeLists.txt
+
+# Fail early if any expected replacement did not happen.
+grep -q 'cmake_minimum_required (VERSION 3.5 FATAL_ERROR)' CMakeLists.txt
+grep -q 'Boost_USE_STATIC_LIBS OFF' CMakeLists.txt
+grep -q 'Boost_INCLUDE_DIRS' CMakeLists.txt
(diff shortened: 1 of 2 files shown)


ubercomrade:
@BiocondaBot please add label

thanhleviet (approved):
Verified locally and looks good:

- sha256 matches the v2.0 release tarball exactly
- All four sed substitutions match the upstream CMakeLists.txt, and each is followed by a grep -q guard that fails the build if a substitution did not apply — good defensive practice
- The key Boost_USE_STATIC_LIBS ON->OFF patch correctly switches to conda's shared Boost, which makes boost-cpp a genuine runtime dep (correctly listed in host + run)
- build.sh passes ${CMAKE_ARGS} so the conda cross-compile toolchain is applied (ARM CI passes)
- License GPL-3.0-or-later with license_file present
- {{ stdlib('c') }} present; tag-vs-version handled cleanly via {% set tag %}
- All CI checks pass (Lint, Linux, OSX-64, ARM)

On run_exports: max_pin="x.x" is more conservative than the usual max_pin="x" for a 1.x+ release, but since this is a leaf CLI application that nothing links against, run_exports is largely immaterial here — fine to keep as you have it, and your rationale in the PR description is reasonable.

LGTM.

PR #66733, opened 2026-06-29
Add pairassembler Bioconda recipe

nrminor:
This PR adds a Bioconda recipe for `pairassembler`, a specialized paired-end sequencing read merger. The package installs the command-line executable as `pairasm`.

The recipe builds from the GitHub `v0.1.3` source tag using the checked-in `Cargo.lock` via `cargo install --locked`. It bundles third-party Rust dependency licenses with `cargo-bundle-licenses` and tests the installed executable with `pairasm --version` and `pairasm --help`.

Additionaly, the upstream `pairassembler` repository has a local-build Cargo config that uses `target-cpu=native`, but I made sure to disable it in the recipe so Bioconda builds remain portable across user machines and HPC environments.

The following validation passed locally before I opened this PR:

```sh
pixi exec -c conda-forge -c bioconda -s bioconda-utils \
  bioconda-utils lint --packages pairassembler recipes config.yml
```

Let me know if anything looks awry or there are any more checks I should run--this is my first Bioconda submission!


nrminor:
Hello! Is there anything I can do to help move this along? It looks like one of the build tests was held back but because it was cancelled, not because it failed. Thanks in advance!

mencian (approved):

PR #66744, opened 2026-06-29
chore: Remove unused bioconda-utils args and scripts

paperbenni:
`--full-report` in bioconda-utils does nothing now and will be removed in the next release. This removes its usage from bioconda-recipes. 

The travis scripts were also unused for quite some time and can be safely removed


aliciaaevans (approved):
Thanks!

PR #66787, opened 2026-06-30
Update GeneFior meta.yaml

NickJD:
Added fastp >=1.3.5 to list of dependencies

### DIFF
--- a/recipes/genefior/meta.yaml
+++ b/recipes/genefior/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "GeneFior" %}
-{% set version = "0.10.2" %}
+{% set version = "0.10.1" %}
 
 package:
   name: {{ name|lower }}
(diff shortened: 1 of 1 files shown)


recipes/genefior/meta.yaml
```diff
@@ -1,5 +1,5 @@
 {% set name = "GeneFior" %}
-{% set version = "0.10.1" %}
+{% set version = "0.10.2" %}
```
bgruening: This number does not seem to exist.

NickJD: I'm trying to add a dependency to the meta build and was told to update the conda build version too... I haven't updates the PyPi.
So I guess I shouldn't update the conda build number..?

bgruening: you should update the build number but not the version number :) see below under buid/number

NickJD: Done! I think I understand now - Thank you

kbseah:
please remember to tag the PR for review when you are satisfied with it, thanks!

PR #66827, opened 2026-07-01
Add remeta

james-edward-hart:
Add a new Bioconda recipe for `remeta` v0.11.2.

`remeta` is a C++ tool for meta-analysis of single variants and gene-based tests using REGENIE summary statistics. The recipe builds from the upstream GitHub release tarball and installs the `remeta` executable.

Citation: Joseph, T., Mbatchou, J., et al. Computationally efficient meta-analysis of gene-based tests using summary statistics in large-scale genetic studies. Nature Genetics, 2025. https://doi.org/10.1038/s41588-025-02390-0

This recipe is Linux-only for the time being. It includes a small patch to adapt upstream CMake configuration for Bioconda by preserving Conda compiler flags, resolving dependencies from the Conda prefix, and linking against Conda shared libraries instead of static/system libraries. The patch also fixes small compatibility issues with the Conda-provided Eigen/compiler stack.

This Conda compatibility issue has been raised with the remeta team: https://github.com/rgcgithub/remeta/issues/9

The recipe includes `run_exports` using `max_pin="x.x"` because `remeta` is currently in the `0.x.x` version series.

Testing performed:
  - `bioconda-utils lint --packages remeta recipes config.yml`
  - Native Linux build/test:
  `bioconda-utils build --packages remeta --force --pkg-dir /tmp/bioconda-pkgs           recipes config.yml`
   - I exercised a Docker-based Linux build locally on macOS. The recipe compiled, packaged, and ran its tests in the Bioconda Docker build environment, producing a `linux-64` package. That local Docker wrapper run ended non-zero due to a host-side expected package filename mismatch. The clean passing run is the linux native build.

The smoke test checks for the executable, runs`remeta --version`, checks command usage output, and runs `remeta index-anno` on a tiny bgzipped annotation file.

A clean Docker/mulled run was not available because Docker was unavailable on the Linux cluster used for final testing.


james-edward-hart:
@BiocondaBot please add label

recipes/remeta/meta.yaml
```diff
@@ -0,0 +1,79 @@
...
+  summary: Meta-analysis of single variants and gene-based tests using REGENIE summary statistics.
+  description: |
+    remeta is a collection of C++ programs for meta-analysis of single
+    variants and gene-based tests using summary statistics from regenie.
+    It supports effect-size meta-analysis, p-value meta-analysis, burden,
+    SKATO, ACATV, LD matrix construction, and annotation indexing.
+
+extra:
+  recipe-maintainers:
+    - james-edward-hart
```
eseiler, line 79:
```suggestion
    - james-edward-hart
  identifiers:
    - doi:10.1038/s41588-025-02390-0
```

you can also link dois if you want to

recipes/remeta/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+    - libdeflate
+    - zlib
+    - bzip2
+    - xz
+    - libcurl
+    - openssl
+    - libopenblas
+    - liblapack
+    - liblapacke
+    - libgomp  # [linux]
```
eseiler, line 57: I think this could be condensed.

For some libs, it's enough to have them in the host section, because they are "exported" to run by default, like `bzip2` and `zlib`.

`libgomp` isn't explicitly needed since it's part of gcc. It's not wrong, just redundant. 
If this recipe was built for macOS too, `- llvm-openmp # [osx]` would be required, because `openmp` is not part of the `llvm` cxx compiler.

This is all optional.

I could take a look if you want me to (also as follow up PR), but in my opinion, it's also fine as is.

eseiler (approved):

recipes/remeta/meta.yaml
james-edward-hart: Thanks for the thoughtful review, this is my first Conda recipe, so this info is helpful! For now I'll leave it as is, but I've got it on my to-do list to condense these dependencies and update for macOS.

PR #66613, opened 2026-06-23
Add look4ltrs, mdl-repeat, te-looker (de novo TE-detection tools for Pan_TE)

unavailable-2374:
Adds three de novo transposable-element / repeat-detection tools that together form the detection backend of the **Pan_TE** pan-genome TE-annotation pipeline. They are submitted as one coupled set because a follow-up `pan_te` recipe depends on all three; `pan_te` will be opened as a separate PR once these are merged.

| Recipe | Language / build | License | Role |
|--------|------------------|---------|------|
| `look4ltrs` | C++17 / CMake | AGPL-1.0-only | LTR retrotransposon detection |
| `mdl-repeat` | C / Make | GPL-3.0-or-later | de novo repeat discovery (MDL principle) |
| `te-looker` | Rust / cargo (`dtr` + helpers) | GPL-3.0-or-later | divergent / "twilight-zone" TE discovery |

All three are public, tagged releases with pinned sha256, and each was build-verified locally with `conda build` against `conda-forge` + `bioconda` (binary produced + tests pass) on linux-64.

- `look4ltrs` v1.0.0 — https://github.com/unavailable-2374/Look4LTRs
- `mdl-repeat` v1.0.0 — https://github.com/unavailable-2374/mdl-repeat
- `te-looker` v0.3.0 — https://github.com/unavailable-2374/te-looker

External tool paths are resolved from `$PATH` / env overrides (no hardcoded paths). `te-looker` runtime-depends on `abpoa` + `spoa` (already in bioconda).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #66832, opened 2026-07-01
Add look4ltrs

unavailable-2374:
Adds `look4ltrs`, a C++17/OpenMP LTR-retrotransposon detector, used as the LTR-detection track of the Pan_TE pan-genome TE annotation pipeline.

- Source: https://github.com/unavailable-2374/Look4LTRs (v1.0.0, AGPL-1.0)
- Linux-only (`skip: True # [osx]`): CMakeLists requires GCC + OpenMP and a GCC-specific -O0 workaround; rejects Clang.

Split out from #66613 per maintainer request to submit new recipes separately.

### DIFF
--- a/recipes/look4ltrs/build.sh
+++ b/recipes/look4ltrs/build.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+set -ex
+
+# CMake build of the look4ltrs target (C++17 + OpenMP).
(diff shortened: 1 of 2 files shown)


unavailable-2374:
@BiocondaBot please add label

eseiler (approved):

PR #66857, opened 2026-07-02
rebuild dnaio for other python versions

eseiler:
`dnaio 1.2.3` is currently only available for `python 3.13`. This may cause [problems](https://github.com/bioconda/bioconda-recipes/pull/66848#issuecomment-4866994037).

* remove `>=3.9` to build for all python versions
* change source url: resolves to same server; the pypi URL seems more common
* script is just the default script
* `{{ stdlib('c') }}` is a new pin

local `bioconda-utils build --docker --mulled-test --packages dnaio` was successful and produced packages for python 3.{10,11,12,13,14}

After this, we can also try bumping the version to 1.2.4 - there is a stale autobump PR somewhere 

---

ping @marcelm (recipe maintainer)

### DIFF
--- a/recipes/dnaio/meta.yaml
+++ b/recipes/dnaio/meta.yaml
@@ -1,12 +1,11 @@
-{% set name = "dnaio" %}
 {% set version = "1.2.3" %}
 
 package:
-  name: {{ name }}
+  name: dnaio
   version: {{ version }}
 
 source:
-  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


recipes/dnaio/meta.yaml
```diff
@@ -1,30 +1,31 @@
+{% set name = "dnaio" %}
```
marcelm: Thanks, I appreciate the ping and the update!

I know I’m fighting windmills here, but I don’t think introducing the package name as a variable is an improvement as it is never going to change anyway. If you just hardcode the name and the URL, everything gets shorter and is IMO more readable. And the Bioconda bot does know how to handle this.

eseiler: I don't mind either way :)

marcelm (approved):

PR #66919, opened 2026-07-05
Update plannotate to 2.0.0

mmcguffi:
## Description

Updates the `plannotate` recipe to 2.0.0.

This is a structural change, not just a version bump: 2.0.0 moved the
annotation databases out of the conda package entirely. They are no
longer bundled at build time via a second `source:` entry — they're
downloaded on demand at runtime via `plannotate setupdb`. The old
`BLAST_dbs.tar.gz` release asset no longer exists (2.0.0 ships
`plannotate-databases-v2.tar.gz` instead, resolved dynamically by the
installed package's own version), so the previous recipe would 404
trying to fetch it.

Changes:
- Drop the second `source:` entry and the `build.sh` step that moved
  `BLAST_dbs` into `site-packages`.
- Fix the entry point: `plannotate.main:main` (was
  `plannotate.pLannotate:main`, stale from before an internal
  refactor).
- Update `run:` requirements to the leaner 2.0.0 core dependencies
  (`biopython`, `numpy`, `pandas`, `pyyaml`, `rich`, `typer`) plus the
  external search tools (`blast`, `diamond`, `infernal`). `infernal`
  was previously missing from the recipe entirely even though Rfam
  search requires it. Kept `bokeh` so `--html` output works out of the
  box, matching prior behavior.
- Dropped `streamlit`/`altair`/`protobuf`/`click`/`curl`/`ripgrep`/
  `trnascan-se`, which are no longer runtime requirements of the core
  CLI in 2.0.0 (the web app is now an optional `server` extra of the
  PyPI-style install, not part of the base CLI).

## Checklist

- [x] Bumped version and sha256 for the new release
- [x] Updated entry point and run requirements to match 2.0.0
- [x] `bioconda-utils` lint/build (will run in CI on this PR)

### DIFF
--- a/recipes/plannotate/meta.yaml
+++ b/recipes/plannotate/meta.yaml
@@ -1,23 +1,18 @@
 {% set name = "plannotate" %}
-{% set version = "1.2.5" %}
-{% set sha256 = "9219c4bfaf423f124737ec532709a1a9650df9f0b4b8da387a7c2cb970565740" %}
+{% set version = "2.0.0" %}
+{% set sha256 = "c0b83e6895cca4895143d5e05704f9e99c9da71389a9575b2e6f2bc913b3d4d6" %}
 
 package:
   name: {{ name }}
   version: {{ version }}
 
 source:
-  - url: https://github.com/mmcguffi/{{ name }}/archive/refs/tags/v{{ version }}.tar.gz
-    sha256: {{ sha256 }}
-  - url: https://github.com/mmcguffi/{{ name }}/releases/download/v{{ version }}/BLAST_dbs.tar.gz
-    sha256: 34c7bacb1c73bd75129e16990653f73b3eba7e3cdb3816a55d3989a7601f2137
-    folder: BLAST_dbs
+  url: https://github.com/mmcguffi/{{ name }}/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: {{ sha256 }}
 
 build:
   number: 0
   noarch: python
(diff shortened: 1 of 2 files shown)


mmcguffi:
@BiocondaBot please add label

recipes/plannotate/meta.yaml
```diff
@@ -1,23 +1,20 @@
...
-  - url: https://github.com/mmcguffi/{{ name }}/releases/download/v{{ version }}/BLAST_dbs.tar.gz
-    sha256: 34c7bacb1c73bd75129e16990653f73b3eba7e3cdb3816a55d3989a7601f2137
-    folder: BLAST_dbs
+  url: https://github.com/mmcguffi/{{ name }}/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: {{ sha256 }}
 
 build:
   number: 0
   noarch: python
   entry_points:
```
kbseah: The `entry_points` statement is not strictly necessary because it is defined in the upstream pyproject.toml.

kbseah (changes requested):
Thanks for the update! Small change suggested below

kbseah (approved):
previously raised issues have been addressed. thanks!

recipes/plannotate/meta.yaml
mmcguffi: thanks @kbseah ! `conda` recipes are still a bit mysterious to me -- I appreciate the direction

PR #66938, opened 2026-07-06
rebuild ngs45 with updated dependencies

eseiler:
After #66893, `cutadapt` and its dependencies are available for all Python versions.

### Major change

* `cutadapt <5.0` -> `cutadapt`

### Small changes

* `name` variable is only used once, so just paste the name directly at this one location
* `{{ python_min }}` isn't really needed (unless `python_min` is modified in the recipe)
* `python` doesn't need to be a test requirement, because the package already depends on `python`
* lexicographically sorted the external tools section in the run dependencies

Ping @maidinhvn (recipe maintainer)

### DIFF
--- a/recipes/ngs45/meta.yaml
+++ b/recipes/ngs45/meta.yaml
@@ -1,8 +1,7 @@
-{% set name = "ngs45" %}
 {% set version = "0.1.0" %}
 
 package:
-  name: {{ name }}
+  name: ngs45
   version: {{ version }}
 
 source:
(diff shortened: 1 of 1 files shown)


maidinhvn:
@eseiler this 0.1.0 rebuild is superseded by #66923 (ngs45 0.2.0). After an
upstream history rewrite the v0.1.0 tarball sha256 also changed (old
53c6817e… is now 6a9fdd22…), so this would fail to build. Could you close it
in favour of #66923? Thanks!

PR #66958, opened 2026-07-06
Add pmotools package

kathrynmurie:
This PR adds the Python package pmotools.

The package is available on PyPI:
https://pypi.org/project/pmotools/

Source:
https://github.com/PlasmoGenEpi/pmotools

### DIFF
--- a/recipes/pmotools/meta.yaml
+++ b/recipes/pmotools/meta.yaml
@@ -1,0 +1,50 @@
+{% set version = "1.1.0" %}
+
+package:
+  name: pmotools
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/p/pmotools/pmotools-{{ version }}.tar.gz
+  sha256: 84f0064eabaa9f470677641264216ede0ee9528a91ce67f11948654cc0edce93
+
+build:
+  entry_points:
+    - pmotools-python = pmotools.cli:main
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python >=3.11
+    - uv-build >=0.8.11,<0.9.0
+    - pip
+  run:
+    - python >=3.11
+    - pandas >=2.2.2
+    - biopython >=1.83
+    - jsonschema >=4.23.0
+    - pre-commit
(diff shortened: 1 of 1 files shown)


kathrynmurie:
@BiocondaBot please add label

recipes/pmotools/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  commands:
+    - pip check
+    - pmotools-python --help
+  requires:
+    - pip
+    - python
+
+about:
+  home: https://pypi.org/project/pmotools/
+  summary: Tools for building and analyzing PMO files
```
kbseah: Suggestion: A description field would be helpful, at least to explain what "PMO files" are.

recipes/pmotools/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+package:
+  name: pmotools
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/p/pmotools/pmotools-{{ version }}.tar.gz
+  sha256: 84f0064eabaa9f470677641264216ede0ee9528a91ce67f11948654cc0edce93
+
+build:
+  entry_points:
```
kbseah: `entry_points` statement is not strictly necessary as it is already defined in the upstream pyproject.toml. If the module is refactored this will be one fewer thing to update in the recipe.

recipes/pmotools/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  imports:
+    - pmotools
+  commands:
+    - pip check
+    - pmotools-python --help
+  requires:
+    - pip
+    - python
+
+about:
```
kbseah, line 40: dev_url would be useful to have here too

kbseah (changes requested):
Thanks for the contribution! Package and license are appropriate for Bioconda. Tests pass. Some small suggestions below

kathrynmurie:
@kbseah Thanks for the review! I have added your suggestions

recipes/pmotools/meta.yaml
kathrynmurie: I removed entry_points

kbseah (approved):
Thank you for the contribution! Previously raised issues have been resolved. Looks good to me

PR #66967, opened 2026-07-06
Update Merqury to v1.4.1

arangrhie:
Please update Merqury to v1.4.1.

### DIFF
--- a/recipes/merqury/meta.yaml
+++ b/recipes/merqury/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "merqury" %}
-{% set version = "1.3" %}
-{% set sha256 = "b6e3f54a6fe30acfe6f2b15ebc26a103a51980adfc03522e712a2baa5e79131a" %}
+{% set version = "1.4.1" %}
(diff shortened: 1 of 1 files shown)


recipes/merqury/meta.yaml
```diff
@@ -25,7 +25,7 @@ requirements:
     - r-ggplot2  >=3.3.2,<=3.3.6
     - r-scales  >=1.1.1
     - bedtools  >=2.29.2
-    - samtools  >=1.10
+    - samtools  >=1.22
```
peterjc, line 28: Why the dependency version bump? I saw nothing about samtools 1.22 over at https://github.com/marbl/merqury

arangrhie: samtools index is used in a few places, I tested the latest release version with 1.22.

arangrhie:
@peterjc would you mind shipping this release? I don't think samtools version is a real blocker

peterjc (approved):
I've not checked the formal policies, but my gut instinct is not to overly narrow the dependencies without good reason. Samtools 1.22 was released a year ago, so fine.

arangrhie:
Thank you!

Issue #58217, opened 2025-08-01
Building times out on CiecleCI checks over an hour


eunos-1128:
Hi,

When building on linux-aarch64, it seems that a timeout error occurs after more than an hour.

https://github.com/bioconda/bioconda-recipes/pull/58027#issuecomment-3143622860

Previously, there were no errors even after more than an hour, but has there been a recent change in the CircleCI settings?

Or is there a problem with the recipe?

Thank you,

xrobin:
I have for instance [this build for OpenStructure, which ran less than two months ago and completed in 2h53](https://app.circleci.com/pipelines/github/bioconda/bioconda-recipes/157713/workflows/63be7ecc-3f62-4cc6-8277-033940af6d29/jobs/285927?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary).

Promod3 should build faster than that, and I think we are just running over the 1 hour mark. (The checks on GitHub take 1h26, but seem to run a tad slower as CircleCI). But I also don't really see how to cut that down further. Any suggestions on how to proceed are welcome.

aliciaaevans:
In my previous experience, it has always been a 1 hour timeout,  so I am actually surprised by the one that went longer. 

For osx-arm, unfortunately we only get 1 runner, so it can get queued behind other recipes. That can explain why the whole thing takes a while, but not why the build step is > 1 hr. 

If you don't care about supporting Python 3.9, you could add `skip: True # [py<310]` to the `build` section. That might get it over the finish line.

xrobin:
Thanks Alicia for the suggestion. Skipping Python 3.9 was just enough to bring the total build time to 59 minutes. Hopefully future builds won't fail, but it would still be good to understand where the difference comes from.

(PS: Mac builds fail at the moment unrelated to this timeout issue. We're looking into it).

xrobin:
Now we have a problem with openstructure in #58392. 
We're not going down from 3 hours to 1 by disabling checks and Python 3.9. 

We never had timeouts before (the recipe has only been available for a few months). So definitely some things changed.

eunos-1128:
Might the settings on the Circle CI side have changed?

[The CI settings on the bioconda](https://github.com/bioconda/bioconda-recipes/blob/master/.circleci/config.yml) side haven't changed in nine months, so maybe there's been some change on the CircleCI side.

eunos-1128:
[...]

PR #66974, opened 2026-07-07
Update pyopenms to 3.5.0

jonasscheid:
Updates `pyopenms` from 3.4.1 to 3.5.0, bringing the Python bindings in line with the C++ packages (`libopenms`/`openms`/`openms-thirdparty`), which are already at 3.5.0 (#61322).

### Why the earlier bump (#61347) failed and what's different here

#61347 has been red since December: the autowrap → Cython codegen aborts with `C class 'ArrayWrapperFloat' already implemented` (across all split modules), so the wrapper sources are never generated and the compile fails.

Root cause: the open-ended `autowrap >=0.24` resolves to autowrap **0.27.0**, which was released *after* pyOpenMS 3.5.0 and whose generated bindings are rejected by Cython 3.2.x. OpenMS built and shipped pyOpenMS 3.5.0 against **autowrap 0.24.0** (`src/pyOpenMS/requirements_bld.txt`: `autowrap >= 0.24.0`, `Cython >= 3.1.0`).

Fix: cap `autowrap >=0.24.0,<0.25`. Cython 3.2.x itself is fine.

### Verified locally

Builds cleanly in the bioconda build container (`linux-64`, py3.12) with `autowrap 0.24.0` + `cython 3.2.8` + `libopenms 3.5.0`; the `import pyopenms` test passes and prints `3.5.0`.

Supersedes #61347.

### DIFF
--- a/recipes/pyopenms/meta.yaml
+++ b/recipes/pyopenms/meta.yaml
@@ -71,7 +71,9 @@ about:
 
 extra:
   additional-platforms:
-    - linux-aarch64
+    # linux-aarch64 temporarily skipped: pyopenms 3.5.0 is a heavy build that
+    # repeatedly tripped CircleCI arm.large infrastructure_fail (not a recipe issue;
+    # it builds fine locally within the runner's memory and time budget).
     - osx-arm64
 {% if "dev" in version %}
   skip-lints:
--- a/recipes/pyopenms/meta.yaml
+++ b/recipes/pyopenms/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "pyopenms" %}
-{% set version = "3.4.1" %}  # if ends with 'dev' it is considered a development release and pulled from GitHub
+{% set version = "3.5.0" %}  # if ends with 'dev' it is considered a development release and pulled from GitHub
 
 package:
   name: {{ name }}
@@ -14,11 +14,11 @@ source:
 
 source:
   url: https://github.com/OpenMS/OpenMS/releases/download/release%2F{{ version }}/OpenMS-{{ version }}.tar.gz
-  md5: 8d0b4d1d333f069959f7d5ff1c77a90f
+  sha256: 29ddff876b6f7e604edca962cbd2f57fe2467262db4b4f039aa43148589f2dbd
 {% endif %}
 
 build:
-  number: 3
+  number: 0
   run_exports:
     - {{ pin_subpackage(name, max_pin="x") }}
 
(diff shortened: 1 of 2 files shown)


recipes/pyopenms/meta.yaml
```diff
@@ -78,3 +80,5 @@ extra:
   identifiers:
     - biotools:openms
     - doi:10.1038/s41592-024-02197-7
+  skip-lints:
+    - compiler_needs_stdlib_c
```
eseiler, line 86: Are we sure that nothing is is linking against the C standard library?
 https://conda-forge.org/news/2024/03/24/stdlib-migration/

Or is this somewhat transitive because it's linked against libopenms (in which case a comment would be helpful)

recipes/pyopenms/meta.yaml
```diff
@@ -14,11 +14,11 @@ source:
 
 source:
   url: https://github.com/OpenMS/OpenMS/releases/download/release%2F{{ version }}/OpenMS-{{ version }}.tar.gz
```
eseiler, line 16: I was wondering if it would be helpful to build a pypi source package in addition to the wheels that OpenMS builds for pypi.
From other recipes, it seems rather straightforward to just use a pypi source package. But I don't know whether 1) it's easy to build a source package and 2) there is a reason the Conda package is built from source.

Doesn't matter for this PR, but I'll inquire what OpenMS thinks about this.

jpfeuffer: Building pyopenms from source from a source package will just be frustration and lead to a lot of unnecessary noise on our issue tracker I think. We have enough trouble to make it work on our CI with all the dependencies.

jpfeuffer: It is of course possible and we might accept a PR but in an ideal world, I wouldn't even want to support pypi at all and go fully pixi/conda.

eseiler:
build_and_test-linux-aarch64 times out on circleCI. The non-configurable runtime limit is 1 hour.
I reran it, but I don't expect a different result.
Maybe we need to use 3 threads...

CircleCI also does the build_and_upload for arm once merged.

image

jonasscheid:
Nice to see the interest, thanks for the hints, 3 threads did the job locally, lets see what CI says

jonasscheid:
Didn't help, I think we need large arm from core team for this.. Afaik arm.xlarge

eseiler:
Not sure if a xlarge runner for aarch64 is available for open source projects in CircleCI.

Related: https://github.com/bioconda/bioconda-recipes/issues/58217

Maybe conditionally restricting python versions for aarch64 + using 2 threads (aarch64-only) would succeed.

Or we just skip linux-aarch64 for now.

jonasscheid:
Dragging in @bgruening, could you help us define the solution here? Thanks in advance!

jonasscheid:
Maybe @kbseah could help pushing this forward?

kbseah:
sorry I don't have experience with this build system, unfortunately!

jpfeuffer:
@jonasscheid do you need arm?? Otherwise just skip until 3.6 which should have lighter python builds.

Or disable compiler optimizations or something specifically on arm.

jonasscheid:
No I don't, then I'll skip if 3.6 will solve it, thanks for the hint 🙏🏼

jpfeuffer:
No worries. Thanks for trying!

jpfeuffer:
I can't guarantee that 3.6 will solve it but we switched to nanobind and made quite some other compile time improvements. I have hope.

bgruening (approved):
Thanks? Sorry to be late

PR #66981, opened 2026-07-07
rebuild anglerfish

eseiler:
Pin minimap2 version (https://github.com/NationalGenomicsInfrastructure/anglerfish/issues/94)

### DIFF
--- a/recipes/anglerfish/meta.yaml
+++ b/recipes/anglerfish/meta.yaml
@@ -1,18 +1,19 @@
-{% set name = "anglerfish" %}
 {% set version = "0.7.0" %}
 
 package:
-  name: "{{ name|lower }}"
-  version: "{{ version }}"
+  name: anglerfish
+  version: {{ version }}
 
 source:
-  url: "https://pypi.io/packages/source/b/bio-{{ name }}/bio_{{ name }}-{{ version }}.tar.gz"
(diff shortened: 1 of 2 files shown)


recipes/anglerfish/meta.yaml
```diff
@@ -1,18 +1,19 @@
-{% set name = "anglerfish" %}
```
peterjc, line 1: Why remove the `name` variable usage?

eseiler: It's only used in the URL, so we can also just "hardcode" it. The name usually also doesn't change, hence why a variable?

I general, it's just a style question.

peterjc: I wondered if there was a policy shift away from the variable style?

peterjc (approved):

recipes/anglerfish/meta.yaml
eseiler: Not that I am aware of.

conda has an example without variable: https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html

bioconda has an example with variable: https://bioconda.github.io/tutorials/2024-adding-bioinformatic-software-to-bioconda.html#writing-the-recipe (a bit further down, the variable is superfluous because it's only used for the package name)

PR #66993, opened 2026-07-08
*Add riboss recipe

SiYangming:
RIBOSS is a Python package that integrates long- and short-read RNA sequencing data with ribosome profiling to identify novel translational events. This recipe installs the Python library via pip and also ships the `riboprof` companion binary.

### Why this package is Linux-only

During review we discovered that `bin/riboprof` (shipped by upstream in the source tarball) is **not a Python script** but a **pre-compiled Linux ELF binary** (≈ 392 KB, x86-64, dynamically linked against libz, libstdc++, libgcc_s, etc.). Because the binary cannot run on macOS or Windows, the package must be platform-specific.

Therefore:

- `noarch: python` was **not used**.
- `skip: True # [not linux]` ensures the package is built only on Linux.
- `binary_relocation: False` is set because the ELF binary is pre-compiled and does not need relocation.
- `skip-lints: should_be_noarch_generic` is added to suppress the linter warning, following the same pattern used by other repackaged-binary recipes in Bioconda (e.g., `lima`, `bax2bam`).

The build script copies the ELF binary as-is to `$PREFIX/bin/` and sets executable permissions (no shebang injection, which would be inappropriate for a binary).

### Package information

| Field       | Value                                                  |
| :---------- | :----------------------------------------------------- |
| **Package** | riboss                                                 |
| **Version** | 0.46                                                   |
| **Source**  | tarball from fork with v0.46 tag (`SiYangming/riboss`) |
| **License** | MIT                                                    |

### Recipe details

- **Source**: `url` + `sha256` (avoids `git describe` / setuptools_scm issues in the build environment).
- **Build**: copies `bin/riboprof` (ELF binary) to `$PREFIX/bin/` and runs `pip install .` for the Python library.
- **Run exports**: `{{ pin_subpackage("riboss", max_pin="x") }}`.
- **Tests**: imports `riboss` and `riboss.orfs`; verifies `riboprof` exists and is executable.


SiYangming:
@BiocondaBot please add label

recipes/riboss/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
kbseah, line 1: this file is not necessary as the license file is present in the upstream repository

recipes/riboss/meta.yaml
```diff
@@ -0,0 +1,77 @@
+{% set name = "riboss" %}
+{% set version = "0.46" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/SiYangming/riboss/archive/refs/tags/v{{ version }}.tar.gz
```
kbseah, line 9: The source repo is not the same as the home and dev repo. Appears to be a fork by the maintainer. It's better if the source is the original repo.

I see that the original repo does not have any tags or releases. Was this the main reason for forking? If so, could you file an issue in the original repo asking for a tagged release?

recipes/riboss/meta.yaml
```diff
@@ -0,0 +1,77 @@
...
+  summary: "Integrating RNA-seq and ribosome profiling to identify novel translational events"
+  description: |
+    RIBOSS integrates long- and short-read RNA sequencing data with ribosome 
+    profiling to identify novel translational events beyond annotated regions, 
+    supporting both prokaryotic and eukaryotic species.
+  dev_url: https://github.com/lcscs12345/riboss
+  doc_url: https://github.com/lcscs12345/riboss/blob/master/README.md
+
+extra:
+  channels:
```
kbseah, line 71: `channels` statement is not necessary. Also, Bioconda recipes should not use the the defaults channel.

kbseah (changes requested):
Thanks for the contribution!

Could you please clarify the relationship between the fork and the original authors' repo? It's better not to use a fork if possible.

PR #64896, opened 2026-04-27
Update pyani-plus to 1.0.1

BiocondaBot:
Update [`pyani-plus`](https://bioconda.github.io/recipes/pyani-plus/README.html): **1.0.0** &rarr; **1.0.1**

[install with bioconda](http://bioconda.github.io/recipes/pyani-plus/README.html) [Conda](https://anaconda.org/bioconda/pyani-plus/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/pyani-plus`](https://github.com//bioconda/bioconda-recipes/tree/bump/pyani_plus/recipes/pyani-plus) (click to view/edit other files)
Summary | Whole-genome classification of microbes using Average Nucleotide Identity and similar methods.
Home | [https://github.com/pyani-plus/pyani-plus](https://github.com/pyani-plus/pyani-plus)
Releases |[https://pypi.org/pypi/pyani_plus/json](https://pypi.org/pypi/pyani_plus/json)

***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

PR #67006, opened 2026-07-08
Bump pyani-plus build to trigger re-build and upload

peterjc:
See #64896, the v1.0.1 upload never happened.

### DIFF
--- a/recipes/pyani-plus/meta.yaml
+++ b/recipes/pyani-plus/meta.yaml
@@ -10,7 +10,7 @@ source:
   sha256: 94f863e4b0d96e6deb7522bed7d60acc01bc49c665a5c6a61de4ccc364f16785
 
 build:
-  number: 1
+  number: 0
   noarch: python
   entry_points:
     - .pyani-plus-private-cli = pyani_plus.private_cli:app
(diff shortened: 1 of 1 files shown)


peterjc:
Catch 22,

```
11:05:57 BIOCONDA DEBUG Found: LintMessage(recipe=Recipe "pyani-plus", check=<class 'bioconda_utils.lint.check_repo.build_number_needs_reset'>, severity=<Severity.ERROR: 30>, title='The recipe build number should be reset to 0', body='\nNo previous build of a package of this name and this version exists,\nthe build number should therefore be 0.', start_line=12, end_line=13, fname='recipes/pyani-plus/meta.yaml', canfix=True)
The following problems have been found (visit https://bioconda.github.io/contributor/linting.html for details on the particular lints you get below.):

ERROR: recipes/pyani-plus/meta.yaml:13: build_number_needs_reset: The recipe build number should be reset to 0
```

recipes/pyani-plus/meta.yaml
eseiler, line 41:
```suggestion
    - typer >=0.12
    - click
```

recipes/pyani-plus/meta.yaml
```diff
@@ -31,14 +31,15 @@ requirements:
     - python >=3.11
```
eseiler, line 31: If I'm not mistaken, this will cause the recipe to be built only for one python version.
I'll double check locally

peterjc:
Good catch on the `click` dependency - we'll drop that in the next release, https://github.com/pyani-plus/pyani-plus/pull/437

recipes/pyani-plus/meta.yaml
peterjc: It's a pure python recipe so there will be only one.

eseiler (approved):

PR #67011, opened 2026-07-08
blacklist kalign3

eseiler:
Migrated to conda-forge. Currently breaks nightlies because it has no `meta.yaml`.


eseiler:
Ping @bgruening (or should I ping core for such things?)

I can also do a follow-up PR removing recipe-directories that have been moved to conda-forge (and remove them from the blacklist).

bgruening (approved):

bgruening:
removing recipes that are in conda forge would be very appreciated! Thanks!

PR #67015, opened 2026-07-08
fungar: fix diamond dependency

henriqueantoniolli:
## Description
This PR pins the `diamond` dependency to `<2.2` for the `fungar` package. 

## Reason for Change
The recent release of DIAMOND v2.2 introduced structural and breaking changes to output formats and command behavior. This caused the FUNGAR pipeline to crash silently or fail to parse files immediately after completing the R1 alignment step. Pinning the dependency ensures backward compatibility with DIAMOND v2.1.x, which functions correctly.

- [x] Fixed the `diamond` dependency pin in `meta.yaml`
- [x] Bumped the build number from `0` to `1`


kbseah (approved):
Thanks for the update!

PR #67016, opened 2026-07-08
Update teloscope to 0.1.5

BiocondaBot:
Update [`teloscope`](https://bioconda.github.io/recipes/teloscope/README.html): **0.1.3** &rarr; **0.1.5**

[install with bioconda](http://bioconda.github.io/recipes/teloscope/README.html) [Conda](https://anaconda.org/bioconda/teloscope/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/teloscope`](https://github.com//bioconda/bioconda-recipes/tree/bump/teloscope/recipes/teloscope) (click to view/edit other files)
Summary | A telomere annotation tools for genome assemblies.
Home | [https://github.com/vgl-hub/teloscope](https://github.com/vgl-hub/teloscope)
Releases |[]()
Author | `@vgl-hub`
***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

PR #67017, opened 2026-07-08
Update teloscope to 0.1.5

AldhairMedico:
Updates `teloscope` to 0.1.5.

This supersedes the failing auto-bump PR #67016. That PR's version and sha256 are correct, but its Lint check fails on `compiler_needs_stdlib_c`: the recipe requests a compiler but has no `stdlib`. The auto-bump bot only rewrites the version/sha lines, so it could not add the missing entry.

Changes here, on top of the current recipe:

- Add `{{ stdlib('c') }}` to `requirements.build` (fixes the lint failure that also gated the ARM build).
- Fix the `license_familiy` -> `license_family` typo.
- Bump to 0.1.5 with `sha256: a644271c88a6648d6f27c7db4e92efce6cab97e534de30aef5a4cc21cbf5261c`, verified against the released `teloscope.v0.1.5-with_submodules.zip`.

The teloscope release workflow has also been updated upstream so future releases attach all assets before publishing, keeping auto-bumps reliable.

### DIFF
--- a/recipes/teloscope/conda_build_config.yaml
+++ b/recipes/teloscope/conda_build_config.yaml
@@ -1,0 +1,9 @@
+# teloscope 0.1.5+ uses std::filesystem (canonical/parent_path in
+# src/main.cpp), which on macOS requires a 10.15+ deployment target. The
+# default osx-64 target is older, so bump it for the Intel macOS build
+# (osx-arm64 is already >= 11.0). See:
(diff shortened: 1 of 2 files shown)


AldhairMedico:
@BiocondaBot please add label

AldhairMedico:
Dear @kbseah 
Please let me know if there's something else needed for this PR.
Our Galaxy wrapper relies on the conda recipe, so we would like to update that there as well

Thanks,
Jack

kbseah (approved):

PR #67036, opened 2026-07-09
Add isovar 1.4.24

jonasscheid:
Adds `isovar` 1.4.24, which assembles RNA reads supporting a variant to determine the mutant protein sequence a tumour actually expresses.

Needed as a dependency of `vaxrank`, which is in turn required by `pVACtools`.

All of its runtime dependencies (`pyensembl`, `varcode`, `pandas`, `pysam`, `psutil`) are already in bioconda.

### Test plan
- [x] Builds locally with `conda-build`
- [x] `import isovar`, `import isovar.cli`, and `isovar --help` all succeed
- [ ] Bioconda CI

---

### Dependency chain (pVACtools)

This PR is one of a set adding [pVACtools](https://github.com/griffithlab/pVACtools) to bioconda. pVACtools was not previously packaged; it needs four new dependencies and two repaired recipes. Suggested merge order:

| # | PR | Recipe | |
|---|----|--------|-|
| 1 | #67034 | `shellinford` 0.4.1 | new (C++ ext) |
| 2 | #67035 | `mhctools` 1.9.0 | new |
| 3 | #67036 | `isovar` 1.4.24 | new |
| 4 | #67037 | `pdfkit` 1.0.0 | fix — published builds are py2.7–3.6 only |
| 5 | #67038 | `mhcnuggets` 2.4.1 build 1 | fix — pins `python =3.6`; breaks under Keras 3 |
| 6 | #67039 | `vaxrank` 1.4.0 | new — needs 1–4 |
| 7 | #67040 | `pvactools` 7.0.1 | new — needs 5–6 |

PRs 1–5 are independent and can be reviewed and merged in any order. 6 and 7 are drafts until their dependencies land in the channel.

PR #67035, opened 2026-07-09
Add mhctools 1.9.0

jonasscheid:
Adds `mhctools` 1.9.0, a uniform Python interface to MHC binding predictors (NetMHC family, MHCflurry, IEDB web servers).

Needed as a dependency of `vaxrank`, which is in turn required by `pVACtools`.

All of its runtime dependencies (`varcode`, `pyensembl`, `sercol`, `mhcflurry`, `mhcnames`) are already in bioconda.

### Test plan
- [x] Builds locally with `conda-build`
- [x] `import mhctools`, `import mhctools.cli`, and `mhctools --help` all succeed
- [ ] Bioconda CI

---

### Dependency chain (pVACtools)

This PR is one of a set adding [pVACtools](https://github.com/griffithlab/pVACtools) to bioconda. pVACtools was not previously packaged; it needs four new dependencies and two repaired recipes. Suggested merge order:

| # | PR | Recipe | |
|---|----|--------|-|
| 1 | #67034 | `shellinford` 0.4.1 | new (C++ ext) |
| 2 | #67035 | `mhctools` 1.9.0 | new |
| 3 | #67036 | `isovar` 1.4.24 | new |
| 4 | #67037 | `pdfkit` 1.0.0 | fix — published builds are py2.7–3.6 only |
| 5 | #67038 | `mhcnuggets` 2.4.1 build 1 | fix — pins `python =3.6`; breaks under Keras 3 |
| 6 | #67039 | `vaxrank` 1.4.0 | new — needs 1–4 |
| 7 | #67040 | `pvactools` 7.0.1 | new — needs 5–6 |

PRs 1–5 are independent and can be reviewed and merged in any order. 6 and 7 are drafts until their dependencies land in the channel.

PR #67034, opened 2026-07-09
Add shellinford 0.4.1

jonasscheid:
Adds `shellinford` 0.4.1, a Python binding for the shellinford C++ library (FM-index over a wavelet matrix/tree).

Needed as a dependency of `vaxrank`, which is in turn required by `pVACtools`.

### Notes
- Compiled recipe (C++ extension); the sdist ships a pre-generated SWIG wrapper, so no SWIG is needed at build time — only a C++ compiler.
- The 2014-era wrapper was verified to compile and run on both Python 3.11 and 3.13, so no python cap is applied.

### Test plan
- [x] Builds locally with `conda-build` (linux-64)
- [x] `import shellinford` and a round-trip `FMIndex` build/search return the expected hit
- [ ] Bioconda CI

---

### Dependency chain (pVACtools)

This PR is one of a set adding [pVACtools](https://github.com/griffithlab/pVACtools) to bioconda. pVACtools was not previously packaged; it needs four new dependencies and two repaired recipes. Suggested merge order:

| # | PR | Recipe | |
|---|----|--------|-|
| 1 | #67034 | `shellinford` 0.4.1 | new (C++ ext) |
| 2 | #67035 | `mhctools` 1.9.0 | new |
| 3 | #67036 | `isovar` 1.4.24 | new |
| 4 | #67037 | `pdfkit` 1.0.0 | fix — published builds are py2.7–3.6 only |
| 5 | #67038 | `mhcnuggets` 2.4.1 build 1 | fix — pins `python =3.6`; breaks under Keras 3 |
| 6 | #67039 | `vaxrank` 1.4.0 | new — needs 1–4 |
| 7 | #67040 | `pvactools` 7.0.1 | new — needs 5–6 |

PRs 1–5 are independent and can be reviewed and merged in any order. 6 and 7 are drafts until their dependencies land in the channel.

PR #67037, opened 2026-07-09
Update pdfkit to 1.0.0 and rebuild as noarch

jonasscheid:
Updates `pdfkit` 0.6.1 → 1.0.0 and rebuilds it as a genuinely noarch package.

### Why

The recipe already declared `noarch: python`, but the only artifacts ever published were `py2.7` / `py3.4` / `py3.5` / `py3.6` builds. As a result `pdfkit` cannot be resolved on any currently supported Python:

```
pdfkit [0.5.0|0.6.1] would require python [=2.7|>=2.7,<2.8.0a0] ...
pdfkit 0.5.0 would require python =3.4 ...
```

This surfaced while packaging `vaxrank` (a `pVACtools` dependency), which depends on `pdfkit`.

### Dropping the `wkhtmltopdf` run dependency

`wkhtmltopdf` is packaged only for `linux-64` (bioconda) and `linux-64`/`win-64` (conda-forge). A `noarch` package that hard-depends on it is therefore uninstallable on osx.

It is not a Python dependency — upstream `pdfkit` declares none — but an external binary invoked through `subprocess`, and only when actually rendering a PDF. Importing and using the library's API does not require it. It is now documented in the recipe and in `about.description` instead. Users who need to render PDFs install `wkhtmltopdf` alongside.

### Test plan
- [x] Builds locally with `conda-build`
- [x] `import pdfkit` succeeds on Python 3.11
- [ ] Bioconda CI

---

### Dependency chain (pVACtools)

This PR is one of a set adding [pVACtools](https://github.com/griffithlab/pVACtools) to bioconda. pVACtools was not previously packaged; it needs four new dependencies and two repaired recipes. Suggested merge order:

| # | PR | Recipe | |
|---|----|--------|-|
| 1 | #67034 | `shellinford` 0.4.1 | new (C++ ext) |
| 2 | #67035 | `mhctools` 1.9.0 | new |
| 3 | #67036 | `isovar` 1.4.24 | new |
| 4 | #67037 | `pdfkit` 1.0.0 | fix — published builds are py2.7–3.6 only |
| 5 | #67038 | `mhcnuggets` 2.4.1 build 1 | fix — pins `python =3.6`; breaks under Keras 3 |
| 6 | #67039 | `vaxrank` 1.4.0 | new — needs 1–4 |
| 7 | #67040 | `pvactools` 7.0.1 | new — needs 5–6 |

PRs 1–5 are independent and can be reviewed and merged in any order. 6 and 7 are drafts until their dependencies land in the channel.

### DIFF
--- a/build-fail-blacklist
+++ b/build-fail-blacklist
@@ -56,7 +56,6 @@ recipes/justbackoff
 recipes/kid
 recipes/myriad
 recipes/nose-capturestderr
-recipes/pdfkit
 recipes/perl-pcap
 recipes/perl-getopt-argvfile
 recipes/perl-probe-perl
--- a/recipes/pdfkit/meta.yaml
+++ b/recipes/pdfkit/meta.yaml
@@ -1,29 +1,49 @@
-package:
-  name: pdfkit
-  version: "0.6.1"
+{% set name = "pdfkit" %}
+{% set version = "1.0.0" %}
 
-build:
-  noarch: python
-  number: 0
-  script: {{ PYTHON }} -m pip install . --ignore-installed --no-deps -vv
+package:
+  name: {{ name }}
+  version: {{ version }}
 
 source:
-  url: https://files.pythonhosted.org/packages/a1/98/6988328f72fe3be4cbfcb6cbfc3066a00bf111ca7821a83dd0ce56e2cf57/pdfkit-0.6.1.tar.gz
-  sha256: ef1da35b78d534197e7ce4a604a4a190e9aa769e56634957535f3479a50d8cd1
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 992f821e1e18fc8a0e701ecae24b51a2d598296a180caee0a24c0af181da02a9
+
+build:
+  number: 0
+  noarch: python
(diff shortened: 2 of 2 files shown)


eseiler:
pdfkit was moved to conda-forge (hence blacklisted)
https://github.com/conda-forge/python-pdfkit-feedstock

jonasscheid:
@BiocondaBot please add label

eseiler:
Moved to conda-forge (`python-pdfkit`). At some point the directories of moved packages will be deleted, until then, they are on the blacklist.

jonasscheid:
Thanks @eseiler, that explains the blacklist — I misread it as a stale entry rather than an intentional move. Closing was right; sorry for the noise.

I've dropped the bioconda `pdfkit` dependency from `vaxrank` (#67039) rather than repointing it at `conda-forge::python-pdfkit`, because `python-pdfkit` requires `wkhtmltopdf`, which has no osx build:

```
$ CONDA_SUBDIR=osx-64 mamba create --dry-run -c conda-forge -c bioconda python-pdfkit
  - nothing provides wkhtmltopdf needed by python-pdfkit-0.6.1-py_0
$ CONDA_SUBDIR=linux-64 mamba create --dry-run -c conda-forge -c bioconda python-pdfkit
  → solves
```

Details in #67039.

PR #67038, opened 2026-07-09
Rebuild mhcnuggets for python >=3.9 and cap keras <3

jonasscheid:
Rebuilds `mhcnuggets` 2.4.1 (build 0 → 1) so that it is installable and actually works.

### Why

Two independent problems made the published package unusable:

**1. It pins `python =3.6`.** Python 3.6 has been EOL since 2021, so the package cannot be installed into any currently supported environment. Now `python >=3.9`.

**2. It calls the Keras 2 API but depended on unbounded `keras`/`tensorflow`.** `mhcnuggets.src.models` tries `import keras` first and falls back to `tensorflow.keras`. Under Keras 3 the import *succeeds*, so the breakage is silent until prediction time:

```
ValueError: Argument(s) not recognized: {'lr': 0.001}
```

(Keras 3 removed the `lr=` alias for `learning_rate`.) Since TensorFlow ≥ 2.16 vendors Keras 3, both are now capped: `keras >=2.6,<3` and `tensorflow >=2.6,<2.16`.

Also switched the build script to the standard `pip install . --no-deps --no-build-isolation --no-cache-dir` and added `run_exports`.

### Verification

With `keras 2.15` / `tensorflow 2.15` on Python 3.11, prediction runs and gives sensible affinities — `GILGFVFTL` (a well-characterised strong HLA-A\*02:01 binder) scores 589 nM:

```
peptide,ic50
SIINFEKL,5600.06
GILGFVFTL,589.25
```

The same call under Keras 3 fails with the `ValueError` above.

### Test plan
- [x] Builds locally with `conda-build`
- [x] Import tests pass; `mhcnuggets.src.predict.predict` verified numerically
- [ ] Bioconda CI

---

### Dependency chain (pVACtools)

This PR is one of a set adding [pVACtools](https://github.com/griffithlab/pVACtools) to bioconda. pVACtools was not previously packaged; it needs four new dependencies and two repaired recipes. Suggested merge order:

| # | PR | Recipe | |
|---|----|--------|-|
| 1 | #67034 | `shellinford` 0.4.1 | new (C++ ext) |
| 2 | #67035 | `mhctools` 1.9.0 | new |
| 3 | #67036 | `isovar` 1.4.24 | new |
| 4 | #67037 | `pdfkit` 1.0.0 | fix — published builds are py2.7–3.6 only |
| 5 | #67038 | `mhcnuggets` 2.4.1 build 1 | fix — pins `python =3.6`; breaks under Keras 3 |
[...]

PR #67039, opened 2026-07-09
Add vaxrank 1.4.0

jonasscheid:
Adds `vaxrank` 1.4.0, which selects and ranks mutated peptides for personalized cancer vaccines.

Required by `pVACtools`, which uses `vaxrank.manufacturability` to compute manufacturability metrics.

All four dependencies are now merged and on the channel, so this is ready for review.

### Why 1.4.0 and not the latest (3.1.8)

`pVACtools` 7.0.1 declares `vaxrank>=1.1.0`, but every `vaxrank >= 1.7.0` requires `numpy>=2.0`, and 1.6.1 requires `pandas>=2.1.4`. Both conflict with pVACtools' own `numpy==1.26.4` / `pandas<2.1.0` pins. `vaxrank 1.4.0` is the newest version that is actually co-installable, and is also what pip's resolver selects for `pip install pvactools==7.0.1`.

### Patches

**`drop-removed-jinja2-autoescape-extension.patch`** removes `extensions=['jinja2.ext.autoescape']` from the report `Environment`. Jinja2 3.0 removed that extension — autoescaping became a core `Environment` keyword back in 2.9 — so `vaxrank --help` died on import with `AttributeError: module 'jinja2.ext' has no attribute 'autoescape'`. The call already passes `autoescape=False` explicitly, so dropping the extension is a no-op. Preferable to pinning `jinja2 <3`, which would drag `markupsafe <2.1` along with it.

**`lazy-pdfkit-import.patch`** moves `import pdfkit` from module scope into `make_pdf_report()`.

`pdfkit` now lives in conda-forge as `python-pdfkit`, which requires the `wkhtmltopdf` binary. `wkhtmltopdf` has no osx build:

```
$ CONDA_SUBDIR=osx-64 mamba create --dry-run -c conda-forge -c bioconda python-pdfkit
  - nothing provides wkhtmltopdf needed by python-pdfkit-0.6.1-py_0
```

Since `vaxrank` is `noarch`, a hard dependency on `python-pdfkit` would make it uninstallable on osx while enabling nothing there (PDF rendering was never possible on osx). `pdfkit` is used in exactly one function, backing the optional `--output-pdf-report` flag — one of four output formats. The lazy import mirrors how upstream already handles the adjacent `xvfbwrapper` import. Users who want PDF reports install `conda-forge::python-pdfkit`, as documented in the recipe.

`setuptools <82` is an explicit run dependency because `vaxrank.cli` imports `pkg_resources` at module level.

### Test plan
- [x] Builds locally with `conda-build`
- [x] `import vaxrank`, `import vaxrank.manufacturability`, `vaxrank --help` all succeed
- [x] Verified in an environment with `pdfkit` absent: `vaxrank.report` imports, `make_pdf_report()` raises `ImportError` only when called, manufacturability scoring returns correct values
- [ ] Bioconda CI

---

### Dependency chain (pVACtools)

This PR is one of a set adding [pVACtools](https://github.com/griffithlab/pVACtools) to bioconda.

| # | PR | Recipe | Status |
|---|----|--------|--------|
| 1 | #67034 | `shellinford` 0.4.1 | merged |
| 2 | #67035 | `mhctools` 1.9.0 | merged |
| 3 | #67036 | `isovar` 1.4.24 | merged |
| 4 | #67038 | `mhcnuggets` 2.4.1 build 1 | merged |
| 5 | #67039 | `vaxrank` 1.4.0 | needs 1–4 |
| 6 | #67040 | `pvactools` 7.0.1 | needs 4–5 |

#67037 (`pdfkit`) was closed: pdfkit lives in conda-forge as `python-pdfkit`. `vaxrank` now imports it lazily instead of depending on it — see #67039.

### DIFF
--- a/recipes/vaxrank/meta.yaml
+++ b/recipes/vaxrank/meta.yaml
@@ -1,0 +1,71 @@
+{% set name = "vaxrank" %}
+{% set version = "1.4.0" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 0a287076eae0542a26fd0131848b230c803714de3ff5ba14d9add0e8d7c99be7
+  patches:
+    - drop-removed-jinja2-autoescape-extension.patch
+
+build:
+  number: 0
+  noarch: python
+  entry_points:
+    - vaxrank = vaxrank.cli:main
+  script: {{ PYTHON }} -m pip install . --no-deps --no-build-isolation --no-cache-dir -vvv
+  run_exports:
+    - {{ pin_subpackage(name, max_pin="x") }}
+
+requirements:
+  host:
+    - python >=3.9
+    - pip
+    - setuptools
+  run:
+    - python >=3.9
+    - six
+    - numpy >=1.14.0
+    - pandas
+    - pyensembl >=1.5.0
+    - varcode >=0.5.9
+    - isovar >=1.1.1
+    - mhctools >=1.5.0
+    - roman
+    - jinja2
+    - pdfkit
+    - pypandoc
+    - shellinford >=0.3.4
+    - xlrd
+    - xlsxwriter
+    - xvfbwrapper
+    - future >=0.16.0
+    - astropy
+    # vaxrank.cli imports pkg_resources at runtime
+    - setuptools <82
+
+test:
+  imports:
+    - vaxrank
+    - vaxrank.manufacturability
+  commands:
+    - vaxrank --help
+
+about:
+  home: https://github.com/openvax/vaxrank
+  summary: Ranked vaccine peptides for personalized cancer immunotherapy
+  description: |
+    Vaxrank selects mutated peptides for therapeutic personalized cancer
+    vaccines, ranking candidates by a combination of predicted MHC binding
+    affinity, RNA expression support and manufacturability.
+  license: Apache-2.0
+  license_family: APACHE
+  license_file: LICENSE
+  dev_url: https://github.com/openvax/vaxrank
+
+extra:
+  recipe-maintainers:
+    - jonasscheid
--- a/recipes/vaxrank/meta.yaml
+++ b/recipes/vaxrank/meta.yaml
@@ -10,6 +10,7 @@ source:
   sha256: 0a287076eae0542a26fd0131848b230c803714de3ff5ba14d9add0e8d7c99be7
   patches:
     - drop-removed-jinja2-autoescape-extension.patch
+    - lazy-pdfkit-import.patch
 
 build:
   number: 0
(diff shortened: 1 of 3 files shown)


recipes/vaxrank/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+    - six
+    - numpy >=1.14.0
+    - pandas
+    - pyensembl >=1.5.0
+    - varcode >=0.5.9
+    - isovar >=1.1.1
+    - mhctools >=1.5.0
+    - roman
+    - jinja2
+    - pdfkit
```
eseiler:
```suggestion
    - python-pdfkit
```

eseiler: this should do the trick for pdfkit

jonasscheid:
Thanks for the fix @eseiler — I hadn't realised `pdfkit` had moved to conda-forge as `python-pdfkit`.

I went one step further than your edit, and want to flag why in case you'd rather I revert it.

**`python-pdfkit` cannot be installed on osx**, because it requires `wkhtmltopdf`, which is only built for `linux-64` (bioconda) and `linux-64`/`win-64` (conda-forge):

```
$ CONDA_SUBDIR=osx-64 mamba create --dry-run -c conda-forge -c bioconda python-pdfkit
Could not solve for environment specs
  - nothing provides wkhtmltopdf needed by python-pdfkit-0.6.1-py_0

$ CONDA_SUBDIR=linux-64 mamba create --dry-run -c conda-forge -c bioconda python-pdfkit
  → solves fine
```

Since `vaxrank` is `noarch`, a hard dependency on `python-pdfkit` would make it uninstallable on osx — and it would buy nothing there, because PDF rendering was never possible on osx to begin with (no `wkhtmltopdf`). It would also propagate: `pvactools` (#67040) depends on `vaxrank` and is otherwise fully osx-compatible.

`pdfkit` is used in exactly one place, `make_pdf_report()`, which backs the optional `--output-pdf-report` flag — one of four output formats (`--output-ascii-report`, `--output-html-report`, `--output-pdf-report`, `--output-csv-report`). So instead of a hard dependency I've added `lazy-pdfkit-import.patch`, which moves `import pdfkit` from module scope into `make_pdf_report()`. This mirrors how the adjacent `xvfbwrapper` import is already handled by upstream, a few lines below.

Result: `vaxrank` installs and runs everywhere; `--output-pdf-report` raises a plain `ImportError` unless the user installs `conda-forge::python-pdfkit`, which is documented in a comment where the dependency would have been.

Verified locally with `python-pdfkit` absent from the environment entirely:

```
pdfkit installed? False
import vaxrank.report OK (lazy import works)
calling make_pdf_report raises ImportError as designed: No module named 'pdfkit'
manufacturability still works, cysteine_count = 2
$ vaxrank --help   → rc=0
```

Happy to switch back to a hard `python-pdfkit` dependency (accepting that vaxrank and pvactools then become linux-only) if you'd prefer that trade-off — just say the word.

eseiler (approved):

jonasscheid:
@BiocondaBot please add label

jonasscheid:
@mergifyio queue

PR #67040, opened 2026-07-09
Add pvactools 7.0.1

jonasscheid:
Adds [pVACtools](https://github.com/griffithlab/pVACtools) 7.0.1, a cancer immunotherapy tools suite (pVACseq/bind/fuse/splice/vector/view). Last of the pvactools recipe set; all deps now merged.

Non-obvious points:
- Deps pinned to what pip resolves for `pvactools==7.0.1` (`vaxrank 1.4.0`, `isovar 1.4.24`, `mhctools 1.9.0`, `shellinford 0.4.1`); newer vaxrank needs numpy>=2, which pVACtools forbids.
- `relax-unavailable-pins.patch` relaxes only two unsatisfiable pins: `biopython==1.77`→`>=1.77` (1.77 has no py>=3.9 build) and `imblearn`→`imbalanced-learn` (the real package).
- `xorg-libx11` run dep: `pvactools.lib` imports `turtle`→`tkinter`→libtk, which needs libX11 that conda-forge `tk` doesn't declare (fails in the minimal container otherwise).
- Build script strips top-level `tests/docs/api_status_tests/predictor_tests` that `find_namespace_packages()` would otherwise install.
- `run_constrained: pvacseq <0a0` — the legacy `pvacseq` recipe ships a colliding `bin/pvacseq`.

Dependency chain: #67034 shellinford, #67035 mhctools, #67036 isovar, #67038 mhcnuggets, #67039 vaxrank (all merged); #67037 pdfkit closed (moved to conda-forge as `python-pdfkit`).

### DIFF
--- a/recipes/pvactools/meta.yaml
+++ b/recipes/pvactools/meta.yaml
@@ -69,6 +69,11 @@ requirements:
     - deepdiff
     - scikit-learn >=1.6.0,<1.8.0
     - imbalanced-learn
+    # pvactools.lib.vector_visualization imports turtle -> tkinter, and
+    # libtk links libX11.so.6. conda-forge's tk does not declare that
+    # dependency, so without this even `pvactools --version` fails in a
+    # minimal container.
+    - xorg-libx11
   run_constrained:
     # the legacy pvacseq package (pVAC-Seq, superseded by pVACtools) also
     # installs bin/pvacseq, so the two cannot share a prefix
--- a/recipes/pvactools/meta.yaml
+++ b/recipes/pvactools/meta.yaml
@@ -1,0 +1,111 @@
+{% set name = "pvactools" %}
+{% set version = "7.0.1" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/griffithlab/pVACtools/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 23e84a47fb6143330b6783f41320eea7985e32e279f987b6d33b5a04b8ecd5a6
+  patches:
+    # biopython 1.77 has no build for python >=3.9 and `imblearn` is only a
+    # deprecated shim around `imbalanced-learn`, which is what conda packages.
+    - relax-unavailable-pins.patch
+
+build:
+  number: 0
+  noarch: python
+  entry_points:
+    - pvactools = pvactools.tools.main:main
+    - pvacseq = pvactools.tools.pvacseq.main:main
+    - pvacbind = pvactools.tools.pvacbind.main:main
+    - pvacfuse = pvactools.tools.pvacfuse.main:main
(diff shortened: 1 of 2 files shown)


jonasscheid:
@BiocondaBot please add label

jonasscheid:
@kbseah @bgruening can we merge this? :)

recipes/pvactools/meta.yaml
```diff
@@ -0,0 +1,116 @@
...
+  sha256: 23e84a47fb6143330b6783f41320eea7985e32e279f987b6d33b5a04b8ecd5a6
+  patches:
+    # biopython 1.77 has no build for python >=3.9 and `imblearn` is only a
+    # deprecated shim around `imbalanced-learn`, which is what conda packages.
+    - relax-unavailable-pins.patch
+
+build:
+  number: 0
+  noarch: python
+  entry_points:
```
kbseah: entry_points statement is not strictly necessary as it is defined in the upstream setup.py. One less thing to update if the code is refactored in the future

recipes/pvactools/meta.yaml
```diff
@@ -0,0 +1,116 @@
...
+    - fsspec <=2025.3.0
+    - packaging
+    - pyarrow
+    - polars ==0.16.18
+    - xlsxwriter
+    - openpyxl
+    - deepdiff
+    - scikit-learn >=1.6.0,<1.8.0
+    - imbalanced-learn
+    # pvactools.lib.vector_visualization imports turtle -> tkinter, and
```
kbseah, line 64: Suggestion: Could this be filed as an issue on conda-forge?

kbseah (approved):
Thanks for the contribution! Package and license are appropriate to Bioconda. Python package with import and CLI tests. The patch is explained with a comment in recipe. Two suggestions below. Looks good to me

recipes/pvactools/meta.yaml
jonasscheid: Good call — removed in 4997d439ba. Confirmed the CLI wrappers are still generated: `pip install` writes the console scripts from upstream `setup.py` into the dist-info, and conda/mamba regenerates `bin/pvac*` from that at install time, so all seven commands remain on PATH (verified `pvactools --version` → 7.0.1 in a fresh env).

recipes/pvactools/meta.yaml
jonasscheid: Filed upstream as conda-forge/tk-feedstock#79 (tk hard-links `libX11.so.6` but does not declare `xorg-libx11`). Referenced it in the comment next to the workaround in 4997d439ba; keeping `xorg-libx11` here until that lands.

bgruening (approved):

bgruening:
Thanks @jonasscheid

PR #67062, opened 2026-07-09
Update python-minibwa to 0.1.3

nh13:
Bumps **python-minibwa** 0.1.0 → 0.1.3 (version + sha256; build number reset to 0).

Upstream [fg-labs/minibwa-bindings](https://github.com/fg-labs/minibwa-bindings) v0.1.3 brings `minibwa-py` into the workspace so its version tracks the crate in lockstep — this is the first release to actually publish the wheel to PyPI. `build.sh` and the vendored-license files are unchanged and remain valid against the new tarball.


nh13:
@BiocondaBot please add label

siebrenf (approved):

PR #67064, opened 2026-07-09
Add cellpheno-nis

Chrisa142857:
Adds `cellpheno-nis` — a C++/LibTorch/CUDA executable for whole-brain 3D nuclei **instance** segmentation of light-sheet microscopy (LSFM).

- **Upstream:** https://github.com/Chrisa142857/Lightsheet_microscopy_image_3D_nuclei_instance_segmentation (I'm the author)
- **Source:** pinned to the `v1.0.1` release tarball + sha256
- **License:** MIT

### Notes for reviewers
This is a **GPU / CUDA** package (like a few existing pytorch-dependent recipes):

- It builds a C++ binary that links CUDA-enabled `libtorch` + `libopencv`. It has **no custom `.cu` kernels**, but CUDA-enabled libtorch's CMake config (`Caffe2Config`) calls `enable_language(CUDA)`, so the recipe needs `{{ compiler('cuda') }}`.
- `libtorch_cuda.so` references NCCL symbols, so `nccl` is a host dependency.
- The upstream binary is generically named `main`; `build.sh` installs it as `cellpheno-nis` so it doesn't collide on PATH.
- `conda_build_config.yaml` pins a CUDA version — happy to align it with whatever your CUDA CI/pinning expects; guidance welcome.
- I locally confirmed the sources compile cleanly against conda-forge `libtorch` 2.12.

This will back an nf-core module (nf-core/modules#12179); a reviewer there suggested publishing on bioconda rather than shipping a build-from-source Dockerfile.

##### Please edit this to add/remove reviewers/labels as needed. Thanks for reviewing!

### DIFF
--- a/recipes/cellpheno-nis/build.sh
+++ b/recipes/cellpheno-nis/build.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+set -euo pipefail
+
+# Build the NIS C++/LibTorch/OpenCV/CUDA executable against the conda environment.
+# find_package(Torch)/find_package(OpenCV) resolve from $PREFIX via CMAKE_PREFIX_PATH.
+cd cpp
+mkdir -p build && cd build
+cmake .. \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_PREFIX_PATH="${PREFIX}" \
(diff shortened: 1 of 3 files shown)


Chrisa142857:
@BiocondaBot please add label

PR #67073, opened 2026-07-10
Rebuild ms2rescore-rs for Python 3.13

jonasscheid:
Rebuilds `ms2rescore-rs` 0.4.3 so that Python 3.13 artifacts are published.

### Why

`ms2rescore-rs` was last built on 2025-09-30 and exists only for py38–py312 — there is no py313 build on any subdir:

```
conda create --dry-run -c conda-forge -c bioconda python=3.13 ms2rescore-rs
  -> LibMambaUnsatisfiableError
conda create --dry-run -c conda-forge -c bioconda python=3.12 ms2rescore-rs
  -> ok (0.4.3)
```

This blocks #61768 (`ms2pip` 4.1.2), which run-depends on `ms2rescore-rs >=0.4,<2`. Because that is a *run* dependency, the compile succeeds and the failure surfaces in conda-build's `_test_env` solve. @RalfG already identified this in https://github.com/bioconda/bioconda-recipes/pull/61768#issuecomment-3816802489.

Upstream already publishes `cp313` wheels for 0.4.3, and the recipe's only skip is `py < 37 or (osx and x86_64)`, so a build-number bump is all that's required to produce the missing artifacts.

### Changes

* `build.number: 0 -> 1` — publish py313 artifacts.
* Add `{{ stdlib('c') }}` — required by the `compiler_needs_stdlib_c` lint, which this recipe did not previously satisfy (it only surfaces once the recipe is touched).

### Verified locally

Built in `quay.io/bioconda/bioconda-utils-build-env-cos7:latest` (linux-64), using bioconda's own `conda_build_config.yaml`:

```
conda build --python 3.13 -m bioconda_utils-conda_build_config.yaml recipes/ms2rescore-rs
  -> ms2rescore-rs-0.4.3-py313h9b6b6b8_1.conda   (import ms2rescore_rs + pip check pass)
conda build --python 3.12 recipes/ms2rescore-rs
  -> ms2rescore-rs-0.4.3-py312h3fd9d12_1.conda   (import ms2rescore_rs + pip check pass)
```

### Follow-up

A subsequent PR will update `ms2pip` to 4.1.2 (superseding #61768). Beyond the missing py313 artifact, #61768 needs two further fixes, both verified locally:

[...]

PR #61768, opened 2026-01-07
Update ms2pip to 4.2.0

BiocondaBot:
Update [`ms2pip`](https://bioconda.github.io/recipes/ms2pip/README.html): **4.1.0** &rarr; **4.2.0**

[install with bioconda](http://bioconda.github.io/recipes/ms2pip/README.html) [Conda](https://anaconda.org/bioconda/ms2pip/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/ms2pip`](https://github.com//bioconda/bioconda-recipes/tree/bump/ms2pip/recipes/ms2pip) (click to view/edit other files)
Summary | MS²PIP: MS² Peak Intensity Prediction
Home | [https://github.com/compomics/ms2pip/](https://github.com/compomics/ms2pip/)
Releases |[https://pypi.org/pypi/ms2pip/json](https://pypi.org/pypi/ms2pip/json)
Recipe Maintainer(s) | @CompOmics, @RalfG

***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

PR #67077, opened 2026-07-10
Update ms2pip to 4.1.2

jonasscheid:
Updates `ms2pip` 4.1.0 → 4.1.2. Supersedes #61768.

`ms2pip 4.1.0` pins `sqlalchemy <2`, which conflicts with `ms2rescore >=3.2` (`sqlalchemy >=2`) and forces the numpy-1 build of ms2pip, dragging the whole stack down to `pyopenms 3.3.0`. Upstream relaxed this in 4.1.1 (CompOmics/ms2pip#249).

Beyond #61768, this needed:
- py313 `ms2rescore-rs`, added in #67073 (merged);
- `setuptools` in `host` (PEP 517 backend, else py313 fails with `Cannot import 'setuptools.build_meta'`);
- keep `- python` unpinned (an explicit spec suppresses the python variant → everything builds `py314`); floor via `skip: # [py < 310]`;
- `{{ stdlib('c') }}` for the `compiler_needs_stdlib_c` lint;
- `sqlalchemy >=1.4,<3` to match upstream.

`linux-aarch64` dropped: importing xgboost there fails with `libgomp.so.1: cannot allocate memory in static TLS block` (aarch64 dependency issue, not this recipe). `osx-arm64` retained.

Verified locally on py312/py313. sha256 checked against PyPI.

### DIFF
--- a/recipes/ms2pip/meta.yaml
+++ b/recipes/ms2pip/meta.yaml
@@ -69,5 +69,9 @@ extra:
     - CompOmics
     - RalfG
   additional-platforms:
-    - linux-aarch64
+    # linux-aarch64 temporarily dropped: on aarch64, importing xgboost in a clean
+    # environment fails with "libgomp.so.1: cannot allocate memory in static TLS
+    # block". The same libxgboost 2.1.4 / pyarrow 24 / libgomp 15.2.0 versions are
+    # resolved on linux-64, where the import succeeds, so this is an aarch64-specific
+    # dependency issue rather than an ms2pip problem. Restore once it is fixed upstream.
     - osx-arm64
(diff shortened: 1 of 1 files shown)


jonasscheid:
@BiocondaBot please add label

recipes/ms2pip/meta.yaml
```diff
@@ -67,5 +69,9 @@ extra:
     - CompOmics
     - RalfG
   additional-platforms:
-    - linux-aarch64
+    # linux-aarch64 temporarily dropped: on aarch64, importing xgboost in a clean
```
pinin4fjords: Can we reduce narrative?

pinin4fjords (approved):

jonasscheid:
@mergifyio queue

PR #67097, opened 2026-07-10
Add cytozip 0.3.4

DingWB:
add cytozip
@BiocondaBot please add label

### DIFF
--- a/recipes/cytozip/meta.yaml
+++ b/recipes/cytozip/meta.yaml
@@ -1,0 +1,97 @@
+{% set name = "cytozip" %}
+{% set version = "0.3.4" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: a952626b317b407a2255f65f9ebd16e98affd308de11ea9e2718c98719ddc533 # filled in by bioconda autobump on each release
+
+build:
+  number: 0
+  skip: true  # [py<38 or win]
+  script: |
+    # htslib is guaranteed present in ``host:`` requirements below, so
+    # force setup.py to build the ``cytozip._bam_pileup`` extension even
+    # if the auto-detection based on ``$PREFIX`` / ``$CONDA_PREFIX``
+    # misses the header path for any reason. Without this, the ``test:``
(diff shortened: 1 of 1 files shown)


DingWB:
please merge @bioconda/core

DingWB:
@BiocondaBot please add label

recipes/cytozip/meta.yaml
```diff
@@ -0,0 +1,114 @@
...
+    # ABI-compatible with the numpy 2.0 headers used at build time.
+    - numpy >=1.19
+    - pandas
+    - loguru
+    - pysam
+    - fsspec
+    - requests
+    - anndata
+    - scipy
+    # NOTE: ``fast-fisher`` is PyPI-only (not on conda-forge / bioconda).
```
bgruening: maybe you can add fast-fisher to conda-forge

recipes/cytozip/meta.yaml
```diff
@@ -0,0 +1,114 @@
...
+    - setuptools_scm
+    - cython >=3.0
+    # Build against numpy 2.0 headers. Cython 3.x generates C that
+    # references NumPy 2.0-only C API (``PyDataType_KIND``,
+    # ``_PyUFuncObject_GET_ITEM_DATA`` …), so building against
+    # numpy 1.x fails with ``implicit declaration of function`` errors.
+    # NumPy 2.0 ABI is forward-compatible with 1.19+ at runtime, so
+    # ``run: numpy >=1.19`` keeps installability wide.
+    - numpy >=2.0
+    # C-library dependencies — both headers AND libs are needed at
```
bgruening: those comments are a bit useless imho

recipes/cytozip/meta.yaml
DingWB: @bgruening Thanks for the review! 

Regarding `fast-fisher`: it's only used by `cytozip.pivot.pivot_fisher` and is imported lazily inside that one function, so it's an optional dependency rather than a hard requirement for the package to work. To keep this PR simple I've left it out of `run:` for now, but I'm happy to submit a `fast-fisher` recipe to conda-forge and wire it in as a proper dependency in a follow-up. Let me know if you'd prefer I do that before merging.

recipes/cytozip/meta.yaml
DingWB: @bgruening Thanks for the review! I've removed the unnecessary comments from `meta.yaml` as suggested.

bgruening:
Thanks

PR #67103, opened 2026-07-11
Update gambit to 1.2.0

jlumpe:
Update `gambit` recipe to 1.2.0.

One thing I'm unsure about - should `libgomp` and `llvm-openmp` go under `host:` / `run:` instead of `build:`? I see it both ways in other recipes.


jlumpe:
@BiocondaBot  please add label

mencian (approved):

PR #67123, opened 2026-07-13
Add admixld recipe

yzfranci:
Adds a new recipe for `admixld`, a command-line tool for fast genome-scale
inference of ancestry disequilibrium in hybrid zones from VCF or MSP input
(https://github.com/yzfranci/AdmixLD).

-- Builds via CMake against htslib and Eigen, with optional OpenMP support.
-- `run_exports` is set to `pin_subpackage("admixld", max_pin="x.x")` since
  the package is pre-1.0 (0.10.0).
-- Verified locally with `bioconda-utils lint` and `bioconda-utils build`.


yzfranci:
@BiocondaBot please add label

PR #67129, opened 2026-07-13
Remove moved packages

eseiler:
This removes packages that were moved to conda-forge.

Follow-up of #67011 

| bioconda | conda-forge   |
|----------|---------------|
| cityhash | cityhash      |
| kalign3  | kalign3       |
| pdfkit   | python-pdfkit |
| perl-uri | perl-uri      |
| pysvg    | pysvg-py3     |
| pyx      | pyx           |

The `cityhash` and `perl-uri` directories were already removed, but the entries were still on the blacklist.


bgruening (approved):
Thanks @eseiler

PR #62973, opened 2026-02-25
Update amplirust to 0.3.0

BiocondaBot:
Update [`amplirust`](https://bioconda.github.io/recipes/amplirust/README.html): **0.2.0** &rarr; **0.3.0**

[install with bioconda](http://bioconda.github.io/recipes/amplirust/README.html) [Conda](https://anaconda.org/bioconda/amplirust/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/amplirust`](https://github.com//bioconda/bioconda-recipes/tree/bump/amplirust/recipes/amplirust) (click to view/edit other files)
Summary | In-silico PCR primer matching and product extraction tool.
Home | [https://github.com/erdikilic/amplirust](https://github.com/erdikilic/amplirust)
Releases |[https://github.com/erdikilic/amplirust/tags](https://github.com/erdikilic/amplirust/tags)
Recipe Maintainer(s) | @erdikilic
Author | `@erdikilic`
***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

PR #67169, opened 2026-07-14
Update amplirust to 0.3.0 (fix aarch64 build)

erdikilic:
Updates **amplirust** to 0.3.0 and fixes the aarch64/arm64 build.

Supersedes the autobump PR #62973, which fails the ARM build. Root cause: `build.sh`
set `RUSTFLAGS="-C target-cpu=aarch64-unknown-linux-gnu"` (and `...=aarch64-apple-darwin`)
— those are target **triples**, not CPU names, so `rustc` rejects them and drops the
default NEON feature. amplirust depends on `sassy`, whose `ensure_simd` then aborts the
build with a `compile_error!` because neither AVX2 (x86-64) nor NEON (aarch64) is enabled.

Fix: keep `-C target-cpu=x86-64-v3` for x86_64 (needed for sassy's AVX2), and apply no
override on aarch64/arm64 so the conda toolchain's default NEON is used. This mirrors the
`sassy` recipe and the pending `whittle` recipe, both of which build cleanly on ARM this way.

- [x] This PR updates an existing recipe (version bump + build fix).

### DIFF
--- a/recipes/amplirust/meta.yaml
+++ b/recipes/amplirust/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "amplirust" %}
-{% set version = "0.2.0" %}
+{% set version = "0.3.0" %}
 
 package:
   name: {{ name|lower }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://github.com/erdikilic/{{ name }}/archive/refs/tags/{{ version }}.tar.gz
-  sha256: 04b92cb2744fe6bc9414d93d5e5e1100f5661f613d2c5fa64690a658ae039e85
+  sha256: a65ec18e54f63ef28a4291973b952516f0341750e7d3b5837b672edc10e87303
 
 build:
   number: 0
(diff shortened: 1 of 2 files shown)


erdikilic:
@BiocondaBot please add label

recipes/amplirust/meta.yaml
```diff
@@ -44,5 +44,10 @@ about:
   dev_url: https://github.com/erdikilic/amplirust
 
 extra:
+  additional-platforms:
+    - linux-aarch64
+    - osx-arm64
   recipe-maintainers:
     - erdikilic
+  skip-lints:
```
bgruening: Can you add stdlib to the build section, instead?

erdikilic: Done. Added `{{ stdlib('c') }}` to `requirements/build` and removed the `compiler_needs_stdlib_c` skip-lint. Thanks!

mencian (approved):

PR #64747, opened 2026-04-22
Update wasp2 to 1.4.1

BiocondaBot:
Update [`wasp2`](https://bioconda.github.io/recipes/wasp2/README.html): **1.4.0** &rarr; **1.4.1**

[install with bioconda](http://bioconda.github.io/recipes/wasp2/README.html) [Conda](https://anaconda.org/bioconda/wasp2/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/wasp2`](https://github.com//bioconda/bioconda-recipes/tree/bump/wasp2/recipes/wasp2) (click to view/edit other files)
Summary | Allele-specific analysis of next-generation sequencing data with Rust acceleration
Home | [https://github.com/mcvickerlab/WASP2](https://github.com/mcvickerlab/WASP2)
Releases |[https://pypi.org/pypi/wasp2/json](https://pypi.org/pypi/wasp2/json)
Recipe Maintainer(s) | @Jaureguy760

***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

PR #67178, opened 2026-07-14
Update wasp2 to 1.4.1

Jaureguy760:
Supersedes #64747 with a minimal version bump that preserves the working 1.4.0 build strategy.\n\nChanges:\n- bump WASP2 from 1.4.0 to 1.4.1\n- reset build number to 0\n- update the PyPI sdist SHA-256 (independently verified)\n\nThe bot PR rewrote the Rust/HTSlib build and fails against rust-htslib 0.49. WASP2 1.4.1 uses rust-htslib 1.0 and retains the 1.4.0 Python dependency contract, so no build-script or dependency changes are required.

### DIFF
--- a/recipes/wasp2/meta.yaml
+++ b/recipes/wasp2/meta.yaml
@@ -1,7 +1,7 @@
 {# Version: keep in sync with rust/Cargo.toml (single source of truth) #}
 {# Run scripts/check-version-consistency.sh to verify #}
 {% set name = "wasp2" %}
-{% set version = "1.4.0" %}
+{% set version = "1.4.1" %}
 
 package:
   name: {{ name|lower }}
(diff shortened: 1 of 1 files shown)


Jaureguy760:
All recipe checks are now green: Linux, ARM, macOS, lint, and summary. This replaces the stale failing bot PR #64747, which I cannot close because it is bot-authored.

recipes/wasp2/meta.yaml
```diff
@@ -60,6 +63,7 @@ requirements:
     # External tools (bioconda)
     - samtools >=1.10
     - bcftools
```
bgruening, line 65: we just fixed bcftools, so mabye it works without your htslib pin

mencian (approved):

PR #67181, opened 2026-07-14
Superintervals v1.0.0

kcleal:
Added {{ stdlib('c') }}

### DIFF
--- a/recipes/superintervals/meta.yaml
+++ b/recipes/superintervals/meta.yaml
@@ -10,7 +10,7 @@ source:
   sha256: 23ba7acda203e9a11b7cb1a4c690e13c70785e916768bcd05f443036e298db61
 
 build:
-  number: 1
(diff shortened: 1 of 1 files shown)


kbseah:
hi @kcleal , the version number and source should be updated for version 1.0.0 too

kcleal:

recipes/superintervals/meta.yaml
kbseah: I think this skip can be removed since cython is now in the host requirements section.

kbseah (changes requested):
Thanks for the update! Looks good to me. One small suggestion below

kbseah:
the OSX test failure looked like a transient issue, works after re-run

kbseah (approved):
issues raised previously have been addressed. looks good to me

PR #67216, opened 2026-07-15
Add Sentieon 201911

pinin4fjords:
Adds a subrecipe to build **Sentieon 201911**, backfilling the gap between the existing `201808.08` and `202010.02` subrecipes. Follows the per-version subdirectory pattern introduced in #55437.

### Why 201911 specifically

201911 is the last release to ship `sentieon tnhapfilter`, a somatic-filtering step removed in 202010+ in favour of `driver --algo TNfilter`. The two produce different somatic call sets, so 201911 is needed to reproduce results generated with that tool.

### Notes on the recipe

- **x86_64 only.** 201911 predates the ARM builds (there is no `arm-sentieon-genomics-201911.tar.gz`), so this follows the `201808.08` shape rather than `202010.02`: no `additional-platforms`, and no `STAR` in the test block since that subcommand is not present in this release.
- **`python >=2.7,<3.0a0`.** `tnhapfilter` is a Python 2 script (`libexec/tnhapfilter` runs `${SENTIEON_PYTHON:-python} .../tnhapfilter.py`, and the script uses `dict.iteritems()` / `cmp()`), so it needs a Python 2 interpreter on `PATH` at runtime. Python 2.7.15 resolves cleanly from conda-forge for `linux-64`; nothing here is built against Python, it is only a runtime dependency of the prebuilt binaries.
- **No `zlib` dependency.** The release bundles its own `libz.so.1.2.8` (alongside `libbz2`, `liblzma`, `libhdf5`, etc.) under `lib/`, loaded via the wrapper's `LD_LIBRARY_PATH`. A conda `zlib` pulls in `libzlib >=1.3.2` via run-export, which is incompatible with Python 2.7's `zlib <1.3` requirement; it is also unnecessary, matching how the other bundled libraries are already handled.
- Patches, `build.sh`, `activate.sh` and `deactivate.sh` are taken from the `201808.08` subrecipe; the symlink patches apply cleanly to the 201911 `bin/sentieon` and `bin/bwa`.

### Testing

Built and tested locally with `bioconda-utils build` on `linux-64` (native x86_64). The test block runs `sentieon driver`, `sentieon driver --help --algo TNscope`, `sentieon util --help` and `sentieon tnhapfilter --help`; the last exercises the Python 2 path.


FriederikeHanssen (approved):

pinin4fjords:
Thanks @FriederikeHanssen !

PR #67217, opened 2026-07-15
Add r-scarhrd 0.1.1

pinin4fjords:
Adds a recipe for [scarHRD](https://github.com/sztup/scarHRD) 0.1.1, an R package that estimates genomic scar signatures (LST, TAI, LOH) and homologous recombination deficiency from WXS/WGS data (Sztupinszki et al., npj Breast Cancer 2018, doi:10.1038/s41523-018-0066-6).

The package is GitHub-only (not on CRAN/Bioconductor) and has no upstream tags/releases, so the source points at the commit corresponding to version 0.1.1 (`c98f8bc`). It is pure R (`noarch: generic`). Dependencies mirror its `DESCRIPTION`/`NAMESPACE`: `r-data.table` and `r-sequenza`, both available on the conda-forge/bioconda channels.

---

Please read the [guidelines for bioconda recipes](https://bioconda.github.io/contributor/guidelines.html) before opening a pull request (PR).

- [x] AUTOMATED TESTS: This PR contains automated tests (the `test:` block runs `library('scarHRD')`).
- [x] The recipe includes a `run_exports` section, so downstream packages are guaranteed compatibility.

### DIFF
--- a/recipes/r-scarhrd/build.sh
+++ b/recipes/r-scarhrd/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+${R} CMD INSTALL --build . ${R_ARGS}
--- a/recipes/r-scarhrd/meta.yaml
+++ b/recipes/r-scarhrd/meta.yaml
@@ -1,0 +1,47 @@
+{% set version = "0.1.1" %}
+{% set github_commit = "c98f8bc42ed0810393a98677d415114360616725" %}
+
+package:
+  name: r-scarhrd
+  version: {{ version }}
+
+source:
+  url: https://github.com/sztup/scarHRD/archive/{{ github_commit }}.tar.gz
+  sha256: 608b984eef261c07db61ce75c40c40587bf49e91009e7388426f67b65c00c7ee
+
+build:
+  number: 0
+  noarch: generic
+  rpaths:
+    - lib/R/lib/
+    - lib/
+  run_exports:
+    - {{ pin_subpackage("r-scarhrd", max_pin="x.x") }}
+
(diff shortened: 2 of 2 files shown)


pinin4fjords:
The Linux Tests failure here isn't caused by this recipe - it's a transitive dependency issue: `r-sequenza` pulls in `bioconductor-genomeinfodbdata`, whose post-link script shells out to `yq` -> `jq`, and `jq` isn't guaranteed to be present in the build/install environment. That's the same root cause as #67204.

Rather than wait on #67204 to merge and rebuild `bioconductor-data-packages`, I've added `jq` directly as a host/run dependency here so this recipe is self-contained and CI can validate it now.

FriederikeHanssen (approved):

pinin4fjords:
Follow-up on the jq dependency: I first added it directly to this recipe, then tried removing it since #67204 had merged and the fixed `bioconductor-data-packages` build (`20260207-hdfd78af_1`) was already live on the channel. That retest failed - CI's solver picked the older, broken build (`_0`) instead, since `bioconductor-genomeinfodbdata`'s pin (`>=20221102`, no upper bound) doesn't force a preference for the newer build number.

Re-added jq, confirmed green, then tried removing it a second time about 20 minutes later - that time it passed, and a local solve confirmed the channel index had caught up and now resolves to the fixed build. So the failure was channel propagation lag, not a permanent solver issue, and this recipe doesn't need to carry `jq` explicitly.

All checks (Lint, Linux, OSX-64, ARM) are green on the current commit.

PR #67225, opened 2026-07-15
Add recipe for fqxv: lossless FASTQ archiver

jayhesselberth:
Adds a recipe for [`fqxv`](https://github.com/rnabioco/fqxv) 0.2.0, a lossless (opt-in lossy) archiver for short-read FASTQ. Built from the Rust workspace member crate `crates/fqxv-cli`.

Tracking issue: rnabioco/fqxv#16. Modeled on the existing `sracha` recipe (same author, same Rust-workspace-CLI pattern).

Notes:
- Dual-licensed MIT OR Apache-2.0; both `LICENSE-MIT` and `LICENSE-APACHE` are bundled.
- `Cargo.lock` is committed upstream, so the build uses `cargo install --locked`.
- `build.sh` scopes `RUSTFLAGS="-C target-cpu=x86-64"` to the `linux-64` build to override the `x86-64-v3` pin in the upstream `.cargo/config.toml`. That pin lives in a `[target.<triple>]` config entry, which outranks `CARGO_BUILD_RUSTFLAGS`/`build.rustflags`, so `RUSTFLAGS` (higher precedence) is required to get a portable baseline binary. AVX2 fast paths runtime-dispatch via `is_x86_feature_detected!`, so the generic baseline loses nothing.
- Additional platforms requested: `linux-aarch64`, `osx-arm64`.

---

* [x] AICF-compliant: license files are packaged (`license_file`).


jayhesselberth:
@BiocondaBot please add label

jayhesselberth:

recipes/fqxv/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+test:
+  commands:
+    - fqxv --help
+
+about:
+  home: https://github.com/rnabioco/fqxv
+  license: MIT OR Apache-2.0
+  license_file:
+    - LICENSE-MIT
+    - LICENSE-APACHE
```
thanhleviet, line 33: build.sh generates `THIRDPARTY.yml` with cargo-bundle-licenses, but it never ships because it's not in `license_file` - Bioconda's Rust convention requires the bundled dependency licenses in the package:

```suggestion
    - LICENSE-APACHE
    - THIRDPARTY.yml
```

recipes/fqxv/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+build:
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('fqxv', max_pin="x.x") }}   # 0.x: pin minor while unstable
+
+requirements:
+  build:
+    - {{ compiler('rust') }}
+    - {{ compiler('c') }}
```
thanhleviet, line 20: The `compiler_needs_stdlib_c` lint is pointing at a real gap rather than a false positive - the standard fix is to add the stdlib jinja alongside the C compiler:

```suggestion
    - {{ compiler('c') }}
    - {{ stdlib('c') }}
```

recipes/fqxv/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  summary: Lossless (opt-in lossy) archiver for short- and long-read FASTQ.
+  dev_url: https://github.com/rnabioco/fqxv
+  doc_url: https://github.com/rnabioco/fqxv#readme
+
+extra:
+  additional-platforms:
+    - linux-aarch64
+    - osx-arm64
+  skip-lints:
+    - compiler_needs_stdlib_c
```
thanhleviet: With `{{ stdlib('c') }}` added above, this lint passes on its own and the suppression can go:

```suggestion
```

thanhleviet (commented):
Verified the sha256 against a fresh download and confirmed the RUSTFLAGS override is doing exactly what its comment says (upstream's .cargo/config.toml does pin target-cpu=x86-64-v3 for linux-64, and RUSTFLAGS is the only knob that outranks it) - nice portable-binary catch. Two recipe fixes below. Also: the OSX-64 failure is not your recipe - the log shows the package built and tested fine, then the job died uploading artifacts (`Failed to CreateArtifact: ENOTFOUND`, a GitHub Actions network hiccup). Pushing the fixes will re-trigger CI, which should come back green.

jayhesselberth:
@thanhleviet should be ready to merge?

thanhleviet (approved):
All review items addressed, and I re-verified the 0.7.0 bump: sha256 matches a fresh download, license unchanged (MIT OR Apache-2.0, both files present), and upstream still pins target-cpu=x86-64-v3 so the RUSTFLAGS override stays correct. CI is green across all platforms - the earlier OSX-64 red was just the artifact-upload network hiccup as suspected. LGTM, ready to merge.

jayhesselberth:
@thanhleviet the linux-64 build didn't ship: https://anaconda.org/channels/bioconda/packages/fqxv/files

should I file a separate issue for this?

thanhleviet:
No need for a separate issue. The linux-64 build succeeded, but the post-merge upload hit a transient race: the concurrent aarch64/osx uploads created the new `fqxv` package name on anaconda.org at the same moment, so the linux-64 upload failed with a 409 conflict. It has been sorted out and linux-64/fqxv-0.7.0-hfa8f182_0.conda should now be on the bioconda channel: https://anaconda.org/bioconda/fqxv/files

PR #67253, opened 2026-07-16
Update bwa-mem3 to build 1: statically link mimalloc on Linux

nh13:
### Fix inert mimalloc on Linux (bwa-mem3, build 1)

On Linux the recipe dynamically links conda-forge's `libmimalloc.so`, which exports only the `mi_*` API — **not** an overriding `malloc`/`free` (it is built without `MI_OVERRIDE` symbol interposition). The binary loads mimalloc (so `bwa-mem3 version` prints a mimalloc version via `mi_version()`) but every allocation falls through to glibc `malloc`, costing **~30% wall on multi-threaded WGS alignment**.

The old `build.sh` comment claimed `-lmimalloc` before `-lc` is sufficient for ELF interposition; that only holds when the `.so` exports `malloc`, which conda-forge's does not.

**Fix:** build and statically `--whole-archive` the bundled `MI_OVERRIDE=ON` `libmimalloc.a` (the upstream Makefile default) on Linux. macOS is unaffected — there `libmimalloc.dylib` interposes via dyld `__interpose` regardless of symbol export — so the Darwin branch keeps the conda-forge mimalloc, and `mimalloc` becomes an `# [osx]`-only requirement. An `ldd` test guard fails the build if a dynamic (inert) libmimalloc link regresses.

**Verified** on a c6a.4xlarge (AVX2), same v0.5.0 source built two ways, 1M read pairs vs hg38, `-t 16`, 3 reps:

| link | median wall | allocator |
|------|-------------|-----------|
| dynamic `-lmimalloc` (current) | 43.4 s | glibc (no mimalloc arenas under `MIMALLOC_SHOW_STATS=1`) |
| static whole-archive (this PR) | **34.0 s** | mimalloc (22 GiB arenas reserved) |

Alignments byte-identical between the two.

Source unchanged (same tarball / sha256); build number bumped `0 → 1`.

---
* [x] I have read the [guidelines](https://bioconda.github.io/contributor/index.html) for bioconda recipes.
* [x] This PR does not touch fewer than 10 recipes (it touches 1).
* [x] Tests pass locally.

### DIFF
--- a/recipes/bwa-mem3/meta.yaml
+++ b/recipes/bwa-mem3/meta.yaml
@@ -10,7 +10,7 @@ source:
   sha256: 5e8bf61d13976a24508201374694f07efec789b6f3154ba8945ec8258bff07b1
 
 build:
-  number: 0
+  number: 1
   run_exports:
     - {{ pin_subpackage('bwa-mem3', max_pin="x") }}
 
@@ -23,14 +23,16 @@ requirements:
     - {{ stdlib('c') }}
   host:
     - zlib
-    - mimalloc >=3.3,<4
+    # Linux statically links the bundled mimalloc (see build.sh); only macOS
+    # links the conda-forge mimalloc, whose dylib interposes malloc via dyld.
+    - mimalloc >=3.3,<4  # [osx]
     - htslib >=1.21,<2
     - libsais >=2.10.4,<3
     - sse2neon ==1.9.1  # [aarch64 or arm64]
     - llvm-openmp  # [osx]
     - libgomp      # [linux]
   run:
-    - mimalloc >=3.3,<4
+    - mimalloc >=3.3,<4  # [osx]
     - htslib >=1.21,<2
     - libsais >=2.10.4,<3
     - llvm-openmp  # [osx]
@@ -39,14 +41,25 @@ requirements:
 test:
   commands:
     - bwa-mem3 version
+    # Guard the Linux mimalloc regression: the bundled allocator is statically
+    # whole-archived, so the binary must NOT dynamically link libmimalloc. A
+    # dynamic libmimalloc.so link (conda-forge's) loads but does not override
+    # malloc, silently falling back to glibc. Fail the build if it reappears.
+    - 'if ldd "${PREFIX}/bin/bwa-mem3" | grep -q libmimalloc; then echo "FAIL - bwa-mem3 dynamically links libmimalloc (malloc override inert)"; exit 1; fi'  # [linux]
 
 about:
   home: "https://github.com/fg-labs/bwa-mem3"
-  license: "MIT AND Apache-2.0"
+  license: "MIT AND Apache-2.0 AND Zlib"
   license_family: MIT
   license_file:
     - LICENSE
     - ext/sse2neon/LICENSE
+    # mimalloc (MIT) is statically linked into the Linux binary (see build.sh);
+    # bundle its license. Harmless on macOS where it is dynamically linked.
+    - ext/mimalloc/LICENSE
+    # zlib-ng (Zlib) is statically linked on all platforms (vendored fast gzip
+    # reader); bundle its license and reflect Zlib in the SPDX expression above.
+    - ext/zlib-ng/LICENSE.md
   summary: "Short-read aligner derived from bwa-mem2 with correctness fixes, performance improvements, and methylation alignment."
   dev_url: "https://github.com/fg-labs/bwa-mem3"
   doc_url: "https://bwa-mem3.readthedocs.io"
--- a/recipes/bwa-mem3/build.sh
+++ b/recipes/bwa-mem3/build.sh
@@ -65,16 +65,24 @@ else
   # bwa-mem3 calls shm_open/shm_unlink, which on conda's CentOS-7 sysroot
   # (glibc 2.17) live in librt. Newer Linux glibc (>= 2.34) folds them
   # into libc, so upstream CI on ubuntu-latest doesn't need this.
-  # Use conda-forge's mimalloc and htslib. conda-forge ships only the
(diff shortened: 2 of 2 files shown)


nh13:
@BiocondaBot please add label

nh13:

recipes/bwa-mem3/meta.yaml
```diff
@@ -39,6 +41,11 @@ requirements:
 test:
   commands:
     - bwa-mem3 version
+    # Guard the Linux mimalloc regression: the bundled allocator is statically
+    # whole-archived, so the binary must NOT dynamically link libmimalloc. A
+    # dynamic libmimalloc.so link (conda-forge's) loads but does not override
+    # malloc, silently falling back to glibc. Fail the build if it reappears.
+    - 'if ldd "${PREFIX}/bin/bwa-mem3" | grep -q libmimalloc; then echo "FAIL - bwa-mem3 dynamically links libmimalloc (malloc override inert)"; exit 1; fi'  # [linux]
```
geoffjentry, line 48: FWIW this came from Claude so do with it what you will:

Consider — the guard proves "not dynamically linked", not "override active"

  The ldd guard is a sound negative regression check, but note its limit: it confirms libmimalloc.so isn't dynamically linked, not that the static override is
  actually intercepting malloc. A truly positive check (e.g. MIMALLOC_SHOW_STATS=1 over a real alignment showing reserved arenas) needs reference data and is too
  heavy for a conda test — and symbol/nm checks are fragile against conda's binary stripping. Given whole-archive of an MI_OVERRIDE=ON .a is deterministic at build
  time, the negative guard is a reasonable pragmatic proxy. Worth a one-line acknowledgment, not a blocker. There is no equivalent guard on macOS, which is
  consistent with the PR's argument that dyld interposition is export-independent.

recipes/bwa-mem3/meta.yaml
```diff
@@ -10,7 +10,7 @@ source:
   sha256: 5e8bf61d13976a24508201374694f07efec789b6f3154ba8945ec8258bff07b1
```
geoffjentry, line 10: From Claude, so FWIW. And I don't think it can be attached cleanly to this diff:

 This is the one thing I'd want addressed before merge. Previously mimalloc was a separate dynamically-linked conda package that carried its own license. As of this
  PR, mimalloc (MIT) is statically compiled into the Linux binary, so bioconda/conda-forge convention is to bundle its license text alongside the others:

    license_file:
      - LICENSE
      - ext/sse2neon/LICENSE
      - ext/mimalloc/LICENSE   # add: mimalloc is now statically linked on Linux

  (Please confirm the exact path in the submodule tree — upstream mimalloc ships LICENSE at its root, so ext/mimalloc/LICENSE is expected.) The SPDX expression MIT
  AND Apache-2.0 already covers MIT, so no change to license/license_family is needed — only the physical file bundling. Including it unconditionally is fine; the
  file exists in the tarball on all platforms.

geoffjentry (approved):

nh13:
### Verified the build-1 artifact (functional + performance)

Fetched the linux-64 artifact from this PR and checked both the linkage and the runtime behavior the fix restores.

**Linkage** — `DT_NEEDED` no longer lists `libmimalloc.so`; the allocator is statically whole-archived (`mi_malloc`/`mi_free`/`mi_version` defined in `.text`), and `bwa-mem3 version` reports `mimalloc 3.3.0` (the vendored submodule) rather than conda-forge's `3.3.2`.

**Positive allocator check** — re: the `ldd`-guard note, agreed it only proves "not dynamically linked," not "override active." As the positive confirmation: I ran a real alignment under `MIMALLOC_SHOW_STATS=1` on a c6a.4xlarge — this build reserves mimalloc arenas (**active**), while the published build 0 reserves none (falls through to glibc malloc). So the whole-archive override is confirmed live, not just present.

**Performance** — c6a.4xlarge (AVX2, `-t 16`), hg38, 1M read pairs, 3 reps, published build 0 vs this PR's build 1:

| package | mimalloc | median wall |
|---------|----------|-------------|
| `bwa-mem3-0.5.0-h5814d7d_0` (published) | inert | 44.17 s |
| `bwa-mem3-0.5.0-h5814d7d_1` (this PR) | active | 33.78 s |

Alignments byte-identical between the two (`md5` over QNAME/FLAG/RNAME/POS/MAPQ/CIGAR); source and `sha256` unchanged — only the mimalloc link differs.

**Licensing** — pushed a commit bundling the licenses of the statically-linked vendored deps: `ext/mimalloc/LICENSE` (MIT, now static on Linux) and `ext/zlib-ng/LICENSE.md` (Zlib, statically linked on all platforms for the vendored gzip reader), and added `Zlib` to the SPDX expression. Thanks @geoffjentry for the license catch.

recipes/bwa-mem3/meta.yaml
nh13: Acknowledged — agreed the guard only proves "not dynamically linked," not "override active." I ran the positive `MIMALLOC_SHOW_STATS=1` check over a real alignment on a c6a.4xlarge (reserved mimalloc arenas => active; published build 0 reserves none), results in the PR comment above. Keeping the lightweight `ldd` guard in-tree as the deterministic build-time proxy. Thanks!

recipes/bwa-mem3/meta.yaml
nh13: Done — added `ext/mimalloc/LICENSE` (MIT, now statically linked on Linux). Also folded in `ext/zlib-ng/LICENSE.md` (Zlib, statically linked on all platforms for the vendored gzip reader) and added `Zlib` to the SPDX expression. Thanks for the catch!

tfenne (approved):

PR #67308, opened 2026-07-18
Update finaletoolkit to 1.0.0

ravibandaru-lab:
Updates `finaletoolkit` to 1.0.0.

This supersedes #67223, which is stalled on a failing **Linux Tests** check. That failure is caused by a missing runtime dependency: 1.0.0's CLI was refactored to use `rich-click` (`finaletoolkit/cli/main_cli.py` does `import rich_click as click`), so the recipe's `finaletoolkit --help` test errors with `ModuleNotFoundError: No module named 'rich_click'`. This PR carries the same 1.0.0 update plus `rich-click >=1.8` in the run requirements to fix it.

Opened by a FinaleToolkit maintainer (epifluidlab). #67223 can be closed in favor of this.

---

* [x] This PR adds a new recipe / updates an existing recipe.
* [x] `bioconda-utils` linting passes.

### DIFF
--- a/recipes/finaletoolkit/meta.yaml
+++ b/recipes/finaletoolkit/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "finaletoolkit" %}
-{% set version = "0.11.1" %}
+{% set version = "1.0.0" %}
 {% set loess_version = "2.1.2" %}
 {% set plotbin_version = "3.1.7" %}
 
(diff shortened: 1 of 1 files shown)


ravibandaru-lab:
@BiocondaBot please add label

kbseah (approved):
Thanks for the update, looks good to me

PR #67336, opened 2026-07-20
Rebuild ms2rescore to pick up ms2pip 4.1.2

jonasscheid:
Build-number bump so the biocontainer resolves against the newly published ms2pip 4.1.2, giving numpy 2 / pyopenms 3.5.0 instead of the old numpy-1 stack.


jonasscheid:
@BiocondaBot please add label

kbseah (approved):
Looks good to me!

PR #66122, opened 2026-06-08
Add rneat

joshfactorial:
Adds a recipe for **rneat** (rusty-neat), a Rust port and extension of NEAT for simulating NGS read data (FASTQ/VCF) with a golden BAM/VCF truth set, structural-variant simulation, and an end-to-end tumor/normal cancer workflow.

- Upstream: https://github.com/ncsa/rusty-neat
- License: BSD-3-Clause
- Source: `v1.16.0` release tarball
- Concept DOI: 10.5281/zenodo.20100558

Opened as a draft so the Bioconda CI can lint and build the recipe on linux-64 / osx-64. Will mark ready once CI is green.

---

This pull request:

- [x] is a new recipe or an update to an existing recipe, following the [contributing guidelines](https://bioconda.github.io/contributor/index.html).
- [x] builds a compiled package; `run_exports` is set with an appropriate `pin_subpackage`.
- [x] includes a `license_file` (`LICENSE.md`) and a basic `test` (`rneat --help`).
- [ ] CI is passing (will confirm after the first build).

PR #67420, opened 2026-07-22
Add eidolon (renamed from rneat)

joshfactorial:
Adds **eidolon**, the renamed successor to the existing **rneat** package.

Upstream `ncsa/rusty-neat` was renamed to `ncsa/eidolon` at **v2.0.0** — same tool, a breaking *identity* change only (no functional change). Same maintainer (@joshfactorial) and the same build approach as the current `rneat` recipe: `cargo install --locked` plus `cargo-bundle-licenses` bundling all transitive Rust licenses into `THIRDPARTY.yml` (`--previous … --check-previous`).

- New package: `recipes/eidolon` (v2.0.0), sha256 of the `ncsa/eidolon` v2.0.0 tarball.
- **Supersedes `recipes/rneat`** — please retire/deprecate the old `rneat` recipe in favor of this one. `eidolon` still ships a deprecated `rneat` alias binary for one transition release, so existing users aren't broken.

The original `rneat` recipe was added in #66122; this recipe is that one with the name/URL/sha updated for the rename.

- [x] I have read the [Bioconda guidelines](https://bioconda.github.io/contributor/guidelines.html).
- [x] Recipe includes a `recipe-maintainers` entry.


joshfactorial:
@BiocondaBot please add label

PR #67451, opened 2026-07-22
Add recipe for reveal-tools 1.0.0

SarahSaadain:
Add recipe for reveal-tools 1.0.0

This PR adds a recipe for REVEAL, a Python toolkit for converting BAM/SAM
alignment files to sequence overview (SO) format, with support for variant
detection (SNPs, indels), coverage normalization, and visualization-ready
outputs.

What it does:
- Converts BAM/SAM alignments to sequence overview (.so) format via pysam
- Detects SNPs and indels with configurable thresholds
- Normalizes coverage using single-copy genes (SCGs) as reference
- Computes average copy number from coverage data
- Generates R/ggplot2-ready plotable output for visualization (bam2so,
  normalize, estimate, so2plotable, plot, and stats/compare subcommands)

Use case:
REVEAL is particularly useful for analyzing genomic regions of interest
(e.g. TEs, genes, symbionts) and their coverage and variation from
short-read alignments.

Notes:
- noarch: generic — pure Python/R pipeline, no compiled code
- The package is named `reveal-tools` rather than `reveal` because
  bioconda already has an unrelated package named `reveal` (a
  graph-based genome aligner). The CLI command installed by this
  package is still `REVEAL`.
- Tests verify --help and run the project's pytest suite
- The source code is at https://github.com/SarahSaadain/REVEAL and is
  MIT licensed.

### DIFF
--- a/recipes/reveal-tools/build.sh
+++ b/recipes/reveal-tools/build.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -euo pipefail
+
+mkdir -p "$PREFIX/bin"
+mkdir -p "$PREFIX/lib/reveal"
+
+cp src/*.py "$PREFIX/lib/reveal/"
+cp src/*.R  "$PREFIX/lib/reveal/"
+
+# Create the REVEAL entry point
+cat > "$PREFIX/bin/REVEAL" << 'EOF'
+#!/bin/bash
+exec python "$CONDA_PREFIX/lib/reveal/reveal.py" "$@"
+EOF
+chmod +x "$PREFIX/bin/REVEAL"
--- a/recipes/reveal-tools/build.sh
+++ b/recipes/reveal-tools/build.sh
@@ -11,6 +11,9 @@ cp src/*.R  "$PREFIX/lib/reveal/"
 # Create the REVEAL entry point
 cat > "$PREFIX/bin/REVEAL" << 'EOF'
 #!/bin/bash
-exec python "$CONDA_PREFIX/lib/reveal/reveal.py" "$@"
+# Resolve the install prefix from the wrapper's own location, so REVEAL works
+# when called by absolute path or from a different active environment.
+_dir="$(cd "$(dirname "$0")" && pwd)"
+exec "$_dir/python" "$_dir/../lib/reveal/reveal.py" "$@"
 EOF
 chmod +x "$PREFIX/bin/REVEAL"
--- a/recipes/reveal-tools/meta.yaml
+++ b/recipes/reveal-tools/meta.yaml
@@ -26,14 +26,8 @@ requirements:
     - r-tidyverse >=2.0.0
 
 test:
-  requires:
-    - pytest
-    - numpy
-  source_files:
-    - tests/
   commands:
     - REVEAL --help
-    - PYTHONPATH="$PREFIX/lib/reveal" pytest tests/ -v
 
 about:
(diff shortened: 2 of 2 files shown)


SarahSaadain:
@BiocondaBot please add label

SarahSaadain:
In case it looks strange that there are only a few commits: The tool is the 3rd iteration of this package. Started with teplotter as stated on the github page. it was included in https://github.com/SarahSaadain/pastForward (https://github.com/SarahSaadain/pastForward/tree/develop/workflow/scripts/dynamics/seqVista) as seqVista. As there is another tool with same name doing also sequence vizualization, it was renamed to REVEAL and extracted from pastForward as a stand alone tool.

SarahSaadain:
Hi @bioconda/core. 

Just a friendly bump on this one. All checks are green (lint, Linux/OSX-64 tests, ARM build). It's been open about three weeks now. Would really appreciate a review whenever someone has a few minutes. 

Thanks so much!

recipes/reveal-tools/build.sh
```diff
@@ -0,0 +1,16 @@
...
+mkdir -p "$PREFIX/bin"
+mkdir -p "$PREFIX/lib/reveal"
+
+cp src/*.py "$PREFIX/lib/reveal/"
+cp src/*.R  "$PREFIX/lib/reveal/"
+
+# Create the REVEAL entry point
+cat > "$PREFIX/bin/REVEAL" << 'EOF'
+#!/bin/bash
+exec python "$CONDA_PREFIX/lib/reveal/reveal.py" "$@"
```
thanhleviet: `$CONDA_PREFIX` is only set when an environment is activated, and it points to the currently active env rather than the env this package is installed in. If the wrapper is called by absolute path without activation (or from a different active env), it will look for `reveal.py` in the wrong prefix. Resolving the path relative to the wrapper itself avoids both problems:

```suggestion
exec python "$(cd "$(dirname "$0")" && pwd)/../lib/reveal/reveal.py" "$@"
```

(`reveal.py` already dispatches subcommands via `Path(__file__).parent` and `sys.executable`, so this one line is the only place that needs to locate the install.)

thanhleviet (commented):
Thanks for the recipe. Verified: sha256 matches the upstream v1.0.0 tarball, declared run deps match the actual imports (pysam, numpy, pandas, tidyverse), and CI is green on all platforms. One portability issue in the wrapper script, suggestion inline.

recipes/reveal-tools/build.sh
SarahSaadain: Thanks, good catch. Fixed in f01ee5b.

I took it a bit further than the suggestion: bare `python` has the same problem, since it resolves from `PATH` and picks up the wrong interpreter in the same cases. And because `reveal.py` dispatches subcommands with `sys.executable`, a wrong interpreter there carries through to every subcommand. So the wrapper now resolves both from its own location:

```bash
#!/bin/bash
_dir="$(cd "$(dirname "$0")" && pwd)"
exec "$_dir/python" "$_dir/../lib/reveal/reveal.py" "$@"
```

The heredoc is quoted (`<< 'EOF'`), so this still gets written out unexpanded. I also checked that it runs by absolute path with an empty environment, no `CONDA_PREFIX` and no `PATH`.

thanhleviet (approved):
Thanks for the fix. All checks are green. LGTM.

PR #67659, opened 2026-07-29
Update seqwish to 0.7.12

AndreaGuarracino:
Updates seqwish to v0.7.12.

seqwish has been rewritten from C++ to Rust, so this switches the recipe from the CMake/C++ build to cargo:

- `source` now points at the `v0.7.12` GitHub tag archive.
- `build.sh` uses `cargo install`; removed the CMake build.
- `requirements.build`: `{{ compiler('rust') }}` + `{{ compiler('c') }}` (the latter for the bundled `zstd-sys`, pulled in via AGC support). Dropped cmake/make/cxx/openmp/jemalloc/zlib/python.
- Removed `skip: True  # [osx]` and added `osx-arm64` — the Rust version builds on macOS.
- `build.sh` removes the repo `.cargo/config.toml` (`target-cpu=native`) so the package is portable across CPUs.

Note for reviewers: one dependency (`ragc-core`, for AGC input) is a git dependency resolved at build time from Cargo.lock, so cargo fetches it during the build.

### DIFF
--- a/recipes/seqwish/build.sh
+++ b/recipes/seqwish/build.sh
@@ -7,4 +7,7 @@ set -xe
 # different CPU. Remove it so a portable baseline target is used.
 rm -f .cargo/config.toml .cargo/config
 
+# Bundle third-party crate licenses (bioconda Rust guideline).
+cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
+
 cargo install --locked --no-track --root "${PREFIX}" --path .
--- a/recipes/seqwish/build.sh
+++ b/recipes/seqwish/build.sh
@@ -2,24 +2,9 @@
 
 set -xe
 
-export LIBRARY_PATH=${PREFIX}/lib
-export LD_LIBRARY_PATH=${PREFIX}/lib
-export CPATH=${PREFIX}/include
-export C_INCLUDE_PATH=${PREFIX}/include
-export CPLUS_INCLUDE_PATH=${PREFIX}/include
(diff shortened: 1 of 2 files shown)


recipes/seqwish/build.sh
```diff
@@ -2,24 +2,9 @@
...
-        EXTRA_FLAGS="${EXTRA_FLAGS} -march=sandybridge"
-        ;;
-    *) ;;
-esac
-
-cmake -H. -Bbuild -DCMAKE_BUILD_TYPE=Generic -DEXTRA_FLAGS="${EXTRA_FLAGS}"
-cmake --build build -j ${CPU_COUNT}
-mkdir -p $PREFIX/bin
-mv bin/* $PREFIX/bin
+cargo install --locked --no-track --root "${PREFIX}" --path .
```
kbseah, line 13: Third party licenses should be bundled in the recipe too, per https://bioconda.github.io/contributor/guidelines.html#rust

kbseah (changes requested):
Thanks for the update! Verified that the upstream repo has been moved. One requested change below

AndreaGuarracino:
Thanks @kbseah!. Added `cargo-bundle-licenses` to the build, generated `THIRDPARTY.yml` in `build.sh`, and listed it under `about: license_file` alongside `LICENSE`.

kbseah (approved):
issues raised previously have been addressed. thanks!

PR #67672, opened 2026-07-29
Update cgmlst-dists-py to 0.1.6

andreaderuvo:
Update `cgmlst-dists-py` to 0.1.6.

Release notes: https://github.com/genpat-it/cgmlst-dists-py/releases/tag/0.1.6

Version and sha256 only; no changes to the dependency list. The new `-y/--missing-handler`
option defaults to the previous behaviour, so the package's output is unchanged.

I am the recipe maintainer and the upstream author.

### DIFF
--- a/recipes/cgmlst-dists-py/meta.yaml
+++ b/recipes/cgmlst-dists-py/meta.yaml
@@ -31,17 +31,16 @@ requirements:
     - pyarrow
 
 test:
-  source_files:
-    - test/boring.tab
-    - validation/boring_c.tab
   commands:
     - cgmlst-dists.py --version | grep -F '{{ version }}'
     - cgmlst-dists.py -h
-    # Compute a real matrix and compare it to the reference produced by the
-    # original C implementation, so the package is verified to work and not
-    # merely to start.
-    - cgmlst-dists.py -i test/boring.tab -c -s > out.tsv
-    - diff out.tsv validation/boring_c.tab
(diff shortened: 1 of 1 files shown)


andreaderuvo:
Pushed two additions:

- `pyarrow` added to the run requirements. Upstream it is an optional accelerator (the loader auto-detects it and falls back to pandas), but conda is the primary install route for this package, and without it the input loader is 3-4x slower on large profile tables. The Docker image has always bundled it; this brings conda users in line.
- The test section now computes a real distance matrix from the bundled `test/boring.tab` and diffs it against `validation/boring_c.tab`, the reference output of the original C `cgmlst-dists`, plus an explicit `import pyarrow`. Previously the tests only checked that the tool started.

Both were verified locally against the released tarball before pushing.

andreaderuvo:
Sorry for the noise — my previous commit broke the Linux tests. The functional test used `source_files`, which works for the conda test but not for the mulled container test, where the source tree is absent:

```
diff: can't stat 'validation/boring_c.tab': No such file or directory
```

The test now generates its own two-sample input inline and asserts the resulting distance, so it needs nothing from the source tree and works in both test environments. Verified by running the exact command list in an empty directory.

Worth noting the build itself succeeded, so adding `pyarrow` to the run requirements resolves fine; only the test step was at fault.

andreaderuvo:

kbseah (approved):
Thanks for the update!

PR #67695, opened 2026-07-30
Update cgmlst-dists-py to 0.1.7

andreaderuvo:
Update `cgmlst-dists-py` to 0.1.7.

Release notes: https://github.com/genpat-it/cgmlst-dists-py/releases/tag/0.1.7

**`numba` is removed from the run requirements.** As of 0.1.7 it is only needed for `--gpu`: the CPU path computes with numpy and, since numba is now imported lazily, never loads it at all. With llvmlite it added ~199 MB to every environment, and it also constrained which Python versions this package could be installed on, since numba trails new releases. Without it, `--gpu` warns on stderr and computes on the CPU with identical results instead of failing, so nothing breaks for existing users; GPU users can `conda install numba` alongside. Existing environments upgrading in place keep the numba they already have.

`pyarrow` stays, as agreed in #67672: conda is the main install route and the Arrow loader is significantly faster on large profile tables.

The test section still computes a real distance matrix and checks the value, which now doubles as proof that the tool works without numba installed.

I am the recipe maintainer and the upstream author.

### DIFF
--- a/recipes/cgmlst-dists-py/meta.yaml
+++ b/recipes/cgmlst-dists-py/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "cgmlst-dists-py" %}
-{% set version = "0.1.6" %}
-{% set sha256 = "267ac181e30f38ccbf76aa21f55060aa92ba8654eadbda3283bed202a9fe14da" %}
+{% set version = "0.1.7" %}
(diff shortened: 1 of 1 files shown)


andreaderuvo:

mencian (approved):

PR #67786, opened 2026-08-02
feat: ARM biocontainers

paperbenni:
# What

## ARM biocontainers

Make use of the new ARM biocontainer support from new bioconda-utils by
- enabling biocontainer builds for ARM there
- inserting a new step to build and publish multi-arch manifests
- accounting for the breaking changes in the bioconda-utils CLI since the typer migration

the build and publish step have shared state (the image metadata to publish), which is why ARM builds have been moved to GitHub Actions (cross-CI coordination and artifact sharing would be messy and probably unreliable)

## ARM as a first class citizen

ARM and x86 previously had duplicated drifting workflows with strictly allocated concurrency quota. 
This consolidates both architectures into a matrix, which prevents drift and duplication (still being able to parameterize things and allowing for singletons like the repodata package) and allows quota to be allocated dynamically between the two

## Moving things to GitHub Actions

Previously Nightly builds were split between CircleCI, GHA and Azure. This again means that manifest creation was prone to races, as it needs artifacts from multiple build runners. This has been solved by moving the entire nightly build to GitHub Actions

## Increased GitHub Actions quota

This will slightly increase usage of GitHub Actions, but since the amount of minutes for standard runners is quoted as "unlimited" for Open Source and this now dynamically allocates concurrency, performance impact should be minimal. 

Some actions workflows could also be moved to [blacksmith](https://www.blacksmith.sh/) which also has Open Source tiers and would only require swapping out the `runs-on` line in workflows

# Pixi migration

Bioconda-utils 5.0.0 has a breaking change which cannot be done with a backwards compat shim because typer genuinely does not (and should not) support some of the ways in which argh parsed arguments. 
This means pinning the bioconda-version in bioconda-common poses a problem

- update bioconda-common first -> old workflows run with new syntax -> crash
- update workflows first -> old bioconda-utils with new workflows -> crash

bioconda-common also forces all branches to use the same bioconda-utils version even if (like in this case) that makes no sense

The bioconda-utils version pin was therefore moved to this repo, and the bioconda-common setup scripts were replaced with a shared action. It also gets installed using pixi now, enabling some simpler workflows and faster setup time (building still happens using conda though)


.github/workflows/Bulk.yml
```diff
@@ -189,6 +209,14 @@ jobs:
           echo '============'
           bioconda-utils build recipes config.yml \
             --worker-offset ${{ matrix.runner }} --n-workers 1 \
-            --lint --anaconda-upload --record-build-failures --skiplist-leafs \
+            --lint --anaconda-upload --record-build-failures --skiplist-leaves \
             --exclude bioconda-repodata-patches
           conda clean -y --all
+
+  publish-mulled-manifests:
+    if: always() && !cancelled()
```
aliciaaevans: These bulk jobs tend to time out after 6 hours (not great but a separate issue). I've also sometimes seen GHA mark a job as cancelled when it times out. Although I seem to remember the timeouts on these bulk builds causing a fail, I'm not 100% sure about the circumstances where it's cancelled vs failed. So there may be an issue here where it doesn't publish the manifests. Or if this fails for some other reason, what's the best way forward to recover from that?

.github/workflows/PR.yml
```diff
@@ -194,11 +192,8 @@ jobs:
   build-osx-64:
     name: OSX-64 Tests
     runs-on: macos-15-intel
-    strategy:
-      fail-fast: true
-      max-parallel: 4
     # Limited concurrency for osx, so first make sure linux can pass
-    needs: build-linux
+    needs: [build-linux, build-linux-arm64]
```
aliciaaevans: I don't think we need linux-arm64 to pass before building osx. Just using linux-64 to make sure there's not something inherently wrong with the recipe. There have already been some complaints about the extra time spent not building them concurrently.

.github/workflows/nightly.yml
```diff
@@ -48,4 +45,82 @@ jobs:
...
+  nightly-linux:
+    name: Nightly ${{ matrix.package_platform }} Builds
+    if: github.repository == 'bioconda/bioconda-recipes'
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - package_platform: linux-64
+            runner: ubuntu-22.04
```
aliciaaevans: Deprecation of ubuntu-22.04 starts September 17 so probably should go to 24.04. That could be a separate PR if you don't want to muddy the waters with more changes.

https://github.com/actions/runner-images/issues/14254

recipes/bioconda-utils/meta.yaml
```diff
@@ -37,7 +37,7 @@ requirements:
     - conda-index =0.10
     - conda-package-streaming =0.12
     - mamba =2.5
-    - argh =0.31
+    - typer =0.26
```
aliciaaevans: Not sure why this is being updated here. I think it's just going to cause it get picked up on the nightly. This should be in a separate PR.

azure-pipeline.yml
```diff
@@ -38,7 +38,7 @@ stages:
               eval "$(conda shell.bash hook)"
               conda activate bioconda
               source=`echo $(Build.SourceBranch) | sed -e "s#^refs/pull/#pull/#"`
-              bioconda-utils lint --loglevel debug --git-range origin/"$SYSTEM_PULLREQUEST_TARGETBRANCH" "$source"
+              bioconda-utils lint --loglevel debug --git-range origin/"$SYSTEM_PULLREQUEST_TARGETBRANCH"..."$source"
```
aliciaaevans: I think maybe it's time to just remove this file.

azure-pipeline-master.yml
```diff
@@ -37,11 +37,18 @@ jobs:
 
           export OSTYPE="linux-gnu"
           export CI="true"
+          mkdir -p /tmp/mulled-images
```
aliciaaevans: Let's just remove these old pipeline configuration files, unless you can think of a reason not to. It's been almost 2 years since everything except nightly ran on there.

aliciaaevans (commented):
Just a few comments. Going to hold off on an approval until the bioconda-utils changes are released.

I imagine bulk is going to be a pain, as always. We need some way to recover if for some reason we end up with a bunch of images missing manifests. That could be a separate thing, though, as long as there's a plan before the next bulk effort.

aliciaaevans:
@paperbenni can you make a PR with the bulk branch as the target so I can test out a few recipes before we merge this? (Also open to any other ideas you have about testing this.)

If all checks out on bulk, we can merge this PR before merging bulk back to master. (That way we have a master PR in the history and not just a squashed bulk PR which I think is confusing when looking back at the history.)

azure-pipeline-master.yml
paperbenni: Removed :)

.github/workflows/nightly.yml
paperbenni: I doubt this will have large implications, builds are using other images already. I bumped it, since there were already some inconsistencies between different Ubuntu versions being used in different places, which is probably worse

recipes/bioconda-utils/meta.yaml
paperbenni: I reverted it :+1:

.github/workflows/PR.yml
paperbenni: I reverted it :+1:

PR #67850, opened 2026-08-04
Bump/nemo 2.4.2

fredgui:
Updates `nemo` to 2.4.2.

2.4.2 is a bugfix release. `RAND::RandULong()` returned only 32 random bits, leaving the upper 32 bits of every 64-bit word zero, so within each word the loci at positions 32–63 were always inherited from the same parental chromosome instead of recombining. This affected free recombination in the bitstring trait architectures (`bdmi`, `delet`, `ntrl`, `quant`).

Beyond the version bump, a few things in the recipe were wrong and are fixed here:

**Dropped unused dependencies** — `mpich-mpicxx`, `libgomp`, `llvm-openmp` and `openblas`. None of them are used: `build.sh` never sets `MPI=1`, so `-DUSE_MPI` is never defined; there is no `#pragma omp` anywhere in `src/`; and BLAS comes from GSL's own `-lgslcblas`. They only constrained the solve. `sed` moved from `host` to `build`, where a tool used by `build.sh` belongs.

**A real test.** The previous test was `nemo{{ version }} | grep "{{ version }}"`, which a binary can pass while being unable to run a simulation. It is replaced by `smoke.ini` + `run_test.sh`, a ~500-generation quantitative-trait run that exercises the GSL linkage, the RNG and the stat machinery, then asserts that a non-zero additive genetic variance was recorded.

**`build.sh` now actually applies conda's flags.** `C_OPTS="..." make` is a shell *environment* assignment, and in GNU make the environment loses to the makefile's own `C_OPTS=` line — so conda's `CXXFLAGS`/`CPPFLAGS` never reached the compiler. The argument form `make C_OPTS=...` would win, but it also clobbers the `C_OPTS += -DHAS_GSL` that the `ifdef GSL` block appends, and the build then fails in `Uniform.h`. Patching the line in place is the only form that does both. `LDFLAGS` is now passed to the linker as well, and the script verifies `src/version.h` agrees with `PKG_VERSION` so a mis-tagged release fails loudly rather than shipping a mislabelled binary.

**License** corrected to `GPL-3.0-or-later`, matching `COPYING` as of 2.4.1.

The `skip: true  # [osx and x86_64]` is retained. Every virtual member of `TTQuanti_diallelic_bitstring_no_pleio` now has an out-of-line definition, so the missing-vtable link error may well be gone, but there is no osx x86_64 machine available to verify it, so this is not the PR to drop it in.

Built and tested on linux-64, linux-aarch64 and osx-arm64 before submitting.

### DIFF
--- a/recipes/nemo/build.sh
+++ b/recipes/nemo/build.sh
@@ -5,8 +5,14 @@ export LDFLAGS="${LDFLAGS} -L${PREFIX}/lib"
 export CXXFLAGS="${CXXFLAGS} -O3"
 
 mkdir -p "${PREFIX}/bin"
+# The release tarball is a git archive and does not ship the (gitignored) bin/
+# output directory, so create it before building.
+mkdir -p bin/
 
-case $(uname -m) in
+
+ARCH=$(uname -m)
+
+case "$ARCH" in
     aarch64)
 	export CXXFLAGS="${CXXFLAGS} -march=armv8-a"
 	;;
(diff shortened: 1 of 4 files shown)


fredgui:
@BiocondaBot please add label

mencian:
Closing in favor of #67899.

PR #67833, opened 2026-08-04
Update crisprme to 2.1.13

BiocondaBot:
Update [`crisprme`](https://bioconda.github.io/recipes/crisprme/README.html): **2.1.12** &rarr; **2.1.13**

[install with bioconda](http://bioconda.github.io/recipes/crisprme/README.html) [Conda](https://anaconda.org/bioconda/crisprme/files)

Info | Link or Description
-----|--------------------
Recipe | [`recipes/crisprme`](https://github.com//bioconda/bioconda-recipes/tree/bump/crisprme/recipes/crisprme) (click to view/edit other files)
Summary | CRISPRme, tool package for CRISPR experiments assessment and analysis.
Home | [https://github.com/pinellolab/CRISPRme](https://github.com/pinellolab/CRISPRme)
Releases |[https://github.com/pinellolab/CRISPRme/tags](https://github.com/pinellolab/CRISPRme/tags)
Author | `@pinellolab`
***

This pull request was automatically generated (see [docs](https://bioconda.github.io/contributor/updating.html)).

PR #67861, opened 2026-08-04
crisprme: add lucapinello as recipe-maintainer

lucapinello:
Add the upstream author (@lucapinello, pinellolab/CRISPRme) as a `recipe-maintainer` so future `crisprme` version bumps can be self-serviced via `@BiocondaBot please merge`. No functional/recipe changes.


lucapinello:
Closing — rather than add the maintainer to the already-shipped 2.1.12 recipe (which the pending 2.1.13 autobump #67833 would then drop), we'll carry `recipe-maintainers: [lucapinello]` into the CRISPRme **2.2.0** recipe, where it persists. Thanks!

PR #67949, opened 2026-08-07
optitype: add coincbc and yara dependencies

jonasscheid:
Adds the CBC solver and the YARA mapper as runtime dependencies of `optitype`, so the alternative solver/mapper backends that OptiType 1.5.0 already supports work out of the box.

- `coincbc` (conda-forge) and `yara` (bioconda) added to `run`; both are recommended by OptiType's own docs and `check-deps` output.
- Build number bumped to 1 — 1.5.0 is the current upstream release, so this is a dependency-only rebuild, not a version bump.
- Added `optitype check-deps` to the recipe tests; it verifies exactly these binaries are on `PATH`.

Verified locally: `optitype check-deps` in an env built from these deps reports RazerS3, YARA mapper/indexer, GLPK and CBC all `[OK]` and exits 0.


jonasscheid:

PR #67973, opened 2026-08-08
Add biocframe 0.7.3

mjsteinbaugh:
New recipe for biocframe, a BiocPy container providing flexible dataframe representation to support nested structures.

- noarch: python, pure setuptools/setuptools_scm build, no compiled code
- Runtime dep: numpy, biocutils >=0.3.3 (biocutils is already live on bioconda)
- License: MIT
- Homepage: https://github.com/BiocPy/biocframe
- This unblocks acidgenomes, an existing acidgenomics bioconda package family (r-acidgenomes) whose Python counterpart depends transitively on biocframe/genomicranges/iranges.

PR #67974, opened 2026-08-08
Add compressed-lists 0.4.5

mjsteinbaugh:
New recipe for compressed-lists (PyPI name compressed_lists, import name compressed_lists), a BiocPy memory-efficient container for list-like objects.

- noarch: python, pure setuptools/setuptools_scm build, no compiled code
- Runtime deps: numpy, biocutils >=0.3.3, biocframe >=0.7.2 (both already on bioconda or in this PR series)
- License: MIT
- Homepage: https://github.com/biocpy/compressed-lists
- Depends on biocframe (#67973, tier 2 in this series) — CI will show an unsatisfiable dep until that merges and rebuilds; part of a 4-recipe dependency chain unblocking acidgenomes (biocframe -> compressed-lists -> iranges -> genomicranges).


kbseah (approved):
thanks for the contribution! looks good to me

PR #67975, opened 2026-08-08
Add iranges 0.7.2

mjsteinbaugh:
New recipe for iranges, BiocPy's Python implementation of the Bioconductor IRanges package.

Compiled pybind11/C++ extension, not noarch. Its lib/CMakeLists.txt uses CMake FetchContent to clone LTLA/nclist-cpp during the build, which won't work in a network-restricted CI sandbox -- this recipe instead vendors nclist-cpp v0.3.0 as a second source entry (folder: nclist-cpp-src) and points FetchContent at it via FETCHCONTENT_SOURCE_DIR_NCLIST + FETCHCONTENT_FULLY_DISCONNECTED=ON, passed through iranges's own setup.py MORE_CMAKE_OPTIONS env var (no source patch needed). Same vendoring pattern already used in recipes/bubblefinder/build.sh.

- Build deps: cmake, make, cxx compiler; host: pybind11, numpy, setuptools
- Runtime deps: numpy, biocutils >=0.3.3, biocframe >=0.7.2, compressed-lists >=0.4.3
- additional-platforms: linux-aarch64, osx-arm64 (matches recipes/fastppm, another pybind11 C++ extension)
- License: MIT (both iranges and nclist-cpp)
- Homepage: https://github.com/biocpy/iranges
- Depends on biocframe (#67973) and compressed-lists (#67974), tiers 2-3 in this series -- CI will show unsatisfiable deps until those merge and rebuild. Part of a 4-recipe chain unblocking acidgenomes.


kbseah (approved):
Package and license are appropriate for Bioconda, build and tests pass. Thanks for the contribution!

PR #68007, opened 2026-08-10
 Update pangbank cli to 0.3.0

JeanMainguy:
This PR updates `pangbank-cli` to version `0.3.0`.

* Update pangbank-api to 0.6.0
* Add `httpx` as a dependency to use pangbank-api SDK.
* Fix the repository URL where the repository name was incorrectly written in lowercase. This prevents Bioconda autobump from correctly detecting the repository and updating the recipe automatically.


kbseah (approved):
thanks for the updates!

PR #68015, opened 2026-08-10
Update oxbreaker to 1.2

rc-reding:
This PR updates oxbreaker from 1.1 -> 1.2

- All shebangs removed from python files, letting conda manage which interpreter to use [fixes crash in the specific case of installing anaconda through homebrew in macOS]
- Refactoring of the helper script `oxbreaker-desktop-helper` to let the users install a desktop shortcut to start OxBreaker with the mouse. Now also supporting macOS.


rc-reding:
@BiocondaBot please add label

mencian (approved):

PR #68065, opened 2026-08-11
fix: make `bioconductor-flames` compile for `linux` again

dlaehnemann:
The latest versions of `bioconductor-flames` are not on `bioconda` for `linux`, but only for `osx`. This PR attempts to provide all the implicit unmentioned dependencies that might be necessary to make this build properly. Here are some cross-references:

* [a `Dockerfile` that gives a bunch of build and runtime requirements installed via apt, or manually copied as a binary](https://github.com/mritchielab/FLAMES/blob/f8c31c2cec3e1b4e5558e0502cf8647461ace6f4/Dockerfile#L65)
* [an issue describing the need for `k8` to be present, which should be ensured with the `bioconda` package of `minimap2`](https://github.com/mritchielab/FLAMES/issues/26)
* [an issue talking about problems with the `blaze2` installation](https://github.com/mritchielab/FLAMES/issues/34) --- while `blaze2` exists as a `bioconda` package, `FLAMES` insists on installing it via `basilisk`, which in turn creates python environments using pip, so we won't bother stuffing this into the conda package

Generally, I think the cleanest solution would be for `FLAMES` to not be a Bioconductor package, but a full-blown workflow with separate software package environments for distinct steps, that pull in the dependencies required for individual `FLAMES` scripts. Will try to propose this to the authors and cross-link here.

This is joint work with @vsuppiyar, who initially stumbled upon the missing `bioconda` package versions and dug out all the needed `conda-forge` packages.

----

### General instructions

```yaml
build:
  run_exports:
    - ...

```

with `...` being one of:

<details>

<table>
  <tr>
  </tr>
  <tr>
  </tr>
</table>

</details>


dlaehnemann:
I put a bunch of [the research about the dependencies of this package in an issue over at the package's GitHub repository](https://github.com/mritchielab/FLAMES/issues/74).

One important fix I added through testing this build in the CI was removing the internal building of a copy of the `minimap2` code in the package repository and a specific `oarfish` version via `cargo`. Both these are now simply specified as runtime dependencies in the `meta.yaml`.

dlaehnemann:
@ChangqingW, would you like to provide feedback on these recipe updates, before we merge them and get out a new version of `FLAMES` on `bioconda`? Maybe you can spot dependencies that are superfluous or version constraints that are not optimal?

dlaehnemann:
Maybe someone else from the @bioconda/r team could also have a look at this fix for a bioconductor package. If you want to dive into the dependency research that went into this, check out [the respective issue I opened in the package's GitHub repo](https://github.com/mritchielab/FLAMES/issues/74).

mfansler:
Review GUI lags due to large patch file, so just commenting here. Also just a quick glance.

Rather than patching out the vendored code, it's probably best to dynamically remove it in the `build.sh` (or even just ignore it). Otherwise, the artifact still ends up redistributing all the vendored code, just in the form of the patch. The patch can then focus on just the changes need to not build the vendored code.

On Conda Forge, `pkg-config` is usually a `build:` dependency rather than `host:`.

It seems like there should be a more functional test than just a library call.

dlaehnemann:
Thanks for those good points.

For the dynamic removal, does it make sense to nevertheless remove the code from the `Makevars` file with a patch and then remove all the copied code with a command in the `build.sh`? Also, do you know if this will survive the bulk update machinery for `bioconductor` recipes?

Regarding the `pkg-config`, I wasn't sure if maybe this is needed somehow for the basilisk installation. But as `bioconductor-basilisk` uses `r-reticulate` in the background, I would assume that the dependency chain there should take care of such things. So as this just came from the packages installed in an example `Dockerfile`,  I'll just remove that dependency altogether and see if this works for people...

And I will try to include a little test script, to actually attempt a run of something.

mfansler:
> For the dynamic removal, does it make sense to nevertheless remove the code from the `Makevars` file with a patch and then remove all the copied code with a command in the `build.sh`? 

Yes, patch the `Makevars`, as you are. 👍  Removing the copied code in `build.sh` shouldn't make a difference in the end, but it helps ensure nothing else needed by the `R CMD BUILD` execution depended on/referenced it.

> Also, do you know if this will survive the bulk update machinery for `bioconductor` recipes?

Sorry, I don't know the implementation.

dlaehnemann:
The `Makevars` file auto-detects and links every `.c` and `.cpp` file in any subdirectory, so we need to remove everything we don't want added into the package:

https://github.com/mritchielab/FLAMES/blob/164abaced4e0838bf7cc167a45d08d819aa4907f/src/Makevars#L5-L14

It tries to filter out the `submodules` folder, but it was nevertheless picking this up, even with a patched `Makevars` file... :shrug:

aliciaaevans:
> Also, do you know if this will survive the bulk update machinery for `bioconductor` recipes?

Unfortunately, it will not as that has been a problem in the past with other recipes. Generally, whoever is running the bulk update catches it when they see the build fails for that recipe and they check the previous meta.yaml and build.sh to manually reapply the fixes.

If/when the bulk process gets overhauled, there will (hopefully) be multiple PRs and it will be easier to see something important was clobbered.

ChangqingW:
I could modify Makevars to optionally disable vendering oarfish and minimap2 within FLAMES if that helps. E.g. with `FLAMES_BUILD_TOOLS` that you can set with the recipe. Or perhaps detect those tools from PATH in Makevers.

dlaehnemann:
> > Also, do you know if this will survive the bulk update machinery for `bioconductor` recipes?
> 
> Unfortunately, it will not as that has been a problem in the past with other recipes. Generally, whoever is running the bulk update catches it when they see the build fails for that recipe and they check the previous meta.yaml and build.sh to manually reapply the fixes.
> 
> If/when the bulk process gets overhauled, there will (hopefully) be multiple PRs and it will be easier to see something important was clobbered.

Thanks for chiming in and providing these infos!


> I could modify Makevars to optionally disable vendering oarfish and minimap2 within FLAMES if that helps. E.g. with `FLAMES_BUILD_TOOLS` that you can set with the recipe. Or perhaps detect those tools from PATH in Makevers.

This won't help us for this version, as this relies on something already packaged for `Bioconductor`. But for future versions that would be great. Optimally, this would not require setting any environment variables, as these will have to be re-added upon every `Bioconductor` release update. Instead, automatically detecting binaries that are present would be great. We then just have to make sure that these are dependencies in `host:` (I think), so they are present at build time.

dlaehnemann:
This is ready for a final review. I tried to comment everything as detailed as possible, giving reasons for the odd requirement specifications and the patches.

@mfansler @aliciaaevans @ChangqingW ?

ChangqingW:
Looks good to me. FLAMES::example_pipeline could be used in the tests to make those test pipelines. I'll update the Makevars and suboptimal version constraints in the next release.

dlaehnemann:
Would the `FLAMES::example()` pipeline be more comprehensive than [the `run_test.r` pipeline I included here](https://github.com/bioconda/bioconda-recipes/pull/68065/changes#diff-4d374cfb916ecfd7b6fd153ddbb807a824dbe1ee45e2a73ffb9220096cdb0a60)? This was basically just using [the first example from the `FLAMES` vignette](https://bioconductor.org/packages/3.22/bioc/vignettes/FLAMES/inst/doc/FLAMES_vignette.html#creating-a-pipeline).

dlaehnemann:
I think the most important part for the packaging here, is that the test case triggers the `basilisk`/`reticulate` environment to get created and used. That way we ensure (to a certain degree), that this will work for end-users of the `bioconda` package, as well.

ChangqingW:
It basically does the same thing and can provide bulk and multi-sample pipeline as well. If you want to test the other pipelines / conditions it could be handy, otherwise it is fine.

dlaehnemann:
I'm only worried about testing all the dependencies, here. And it is definitely testing the use of the `basilisk` environment and I can see it is running `FLEXIPLEX`, `minimap2` and `oarfish`. And I assume the isoform identification is using `bambu`? Are there any other dependencies that this minimal workflow is not using, that we should test separately?

ChangqingW:
The alternative demlutiplexer BLAZE is not tested here.
```
  pipeline <- example_pipeline("MultiSampleSCPipeline") # or SingleCellPipeline
  pipeline@config$pipeline_parameters$bambu_isoform_identification <- FALSE #
  pipeline@config$pipeline_parameters$oarfish_quantification <- FALSE       #  might as well test the built-in Python scripts
  pipeline@config$pipeline_parameters$demultiplexer <- "BLAZE"
  pipeline@expect_cell_number <- rep(100L, length(pipeline@fastq))
  pipeline@controllers <- list(default = crew::crew_controller_local()) # also test if the crew controller works
```

dlaehnemann:
Alright. Let's see if this also runs or if we hit any further issues.

dlaehnemann:
@ChangqingW are you sure your above code should work with BLAZE2? It gives me [this error in the CI](https://github.com/bioconda/bioconda-recipes/actions/runs/32150434296/job/95755174696?pr=68065#step:10:3405):

```
✖ isoform_identification (failed: Error:
! no lines available in input
)
ℹ read_realignment (pending)
ℹ transcript_quantification (pending)
$step
[1] "isoform_identification"

$error
<error/crew_error>
Error:
! no lines available in input
---
Backtrace:
     ▆
  1. ├─FLAMES::run_FLAMES(pipeline2)
  2. └─FLAMES::run_FLAMES(pipeline2)
  3.   ├─base::tryCatch(...)
  4.   │ └─base (local) tryCatchList(expr, classes, parentenv, handlers)
  5.   │   └─base (local) tryCatchOne(expr, names, parentenv, handlers[[1L]])
  6.   │     └─base (local) doTryCatch(return(expr), name, parentenv, handler)
  7.   ├─FLAMES::run_step(pipeline, step, disable_controller = FALSE)
  8.   └─FLAMES::run_step(pipeline, step, disable_controller = FALSE)
  9.     └─controller$pop(error = "stop")
 10.       ├─crew:::if_any(...)
 11.       └─crew:::crew_error(message = .subset2(out, "error"))
 12.         └─crew:::crew_stop(message = message, class = c("crew_error", "crew"))
 13.           └─rlang::abort(message = message, class = class, call = emptyenv())

$traceback
[1] "No traceback available "
```

And the `no lines available in input` there sounds like a data problem to me, not some dependencies or something?

ChangqingW:
Ops this is a bug with FLAMES itself. The test data does not have enough read to support any isoform and led to an empty file. Maybe keep `pipeline@config$pipeline_parameters$bambu_isoform_identification <- TRUE` for now and I'll fix it in the next release.

vsuppiyar:
To follow-up on this -- I believe it's caused by this code block: https://github.com/mritchielab/FLAMES/blob/devel/inst/python/sc_longread.py#L1205, the conditional check in `group_bam2isoform`. When the BAM file has insufficient aligned reads, the condition `if len(tmp_isoform) > 0` evaluates to `False`, so the isoform processing is skipped entirely, causing downstream failure. You could either provide as test data a BAM file with sufficient aligned reads or handle empty isoform cases within this function.

dlaehnemann:
@ChangqingW OK, I think we are still running into bugs with this second example pipeline:

```
[410](https://github.com/bioconda/bioconda-recipes/actions/runs/32351606672/job/96372230276?pr=68065#step:10:3411)
✖ isoform_identification (failed: Error:
! BiocParallel errors
  1 remote errors, element index: 2
  2 unevaluated and other errors
  first remote error:
Error in full_join(eqClassCount, minEquiRC, by = c("eqClassById", "GENEID", : Can't join `x$eqClassById` with `y$eqClassById` due to incompatible types.
)
ℹ read_realignment (pending)
ℹ transcript_quantification (pending)
$step
[1] "isoform_identification"

$error
<error/crew_error>
Error:
! BiocParallel errors
  1 remote errors, element index: 2
  2 unevaluated and other errors
  first remote error:
Error in full_join(eqClassCount, minEquiRC, by = c("eqClassById", "GENEID", : Can't join `x$eqClassById` with `y$eqClassById` due to incompatible types.
```

ChangqingW:
I see. The FASTQ files for MultiSampleSCPipeline example_pipeline function was not sampling enough reads and causing bambu to [error](https://github.com/GoekeLab/bambu/pull/522 ). I fixed this in [devel](https://github.com/mritchielab/FLAMES/commit/077f59fa3b4409b904cb8ea05add117d5d339e7f#diff-433326e120cee4586c8563f4fd23ece1b908675531561f950f83d8d6c3b56fa6R230) but did not push to release. You could either use the updated example_pipeline function in the test scripts or only test BLAZE with SingleCellPipeline.

dlaehnemann:
That looks like a bigger patch, so this is probably left for a later release once this makes its way through a Bioconductor release and then the bulk update of the bioconda recipe. So I'll try testing BLAZE and non-oarfish quantification with the SingleCellPipeline for now.

dlaehnemann:
OK, this is finally looking good. I added a full test case after @mfansler 's suggestion and added a further test case as suggested by the package author @ChangqingW. And these are all passing, so this recipe should cover most dependencies, including the optional ones. And this is ready for a final review and hopefully a quick merge.

@mfansler is listed as busy, so maybe @dpryan79 or @aliciaaevans could have a look?

johanneskoester (approved):
Impressive work!

PR #68096, opened 2026-08-12
Update crispritz to 2.8.1

lucapinello:
Update `crispritz` 2.8.0 → 2.8.1.

### Changes
- **Version bump** 2.8.0 → 2.8.1 (new `sha256`; build number reset to 0).
- **Install the compiled `enricher`** that 2.8.1's `Makefile_conda` builds.
  `crispritz.py`'s `_enricher_command()` looks for an `enricher` binary on `PATH`
  and beside `Python_Scripts/Enrichment/`; if it is absent, crispritz silently
  falls back to a ~15× slower pure-Python `enricher.py`. The previous recipe
  compiled but never installed it.
- **Test section**: assert the four compiled binaries are present and the
  `enricher` is installed in both resolution paths, plus the reported version.

No dependency changes vs the current 2.8.0 recipe.

### Local verification
On a native **linux-aarch64** build I confirmed all four C++/OpenMP binaries
(`buildTST`, `searchTST`, `searchBruteForce`, `enricher`) compile and install, the
`enricher` binary links `libz` and runs, and the test-section commands pass.


lucapinello:
@bioconda/review  hello! Anyone that has some time to help in merging this? Thanks!

recipes/crispritz/build.sh
```diff
@@ -13,3 +13,10 @@ cp -R \
     searchBruteForce \
     sourceCode/Python_Scripts \
     "${PREFIX}/opt/crispritz/"
+# Install the compiled (fast) enricher in BOTH places crispritz.py
+# _enricher_command() checks: a binary named 'enricher' on PATH, and one beside
+# the copied Python_Scripts/Enrichment/. Makefile_conda builds it at the repo
+# root (not inside Python_Scripts), so it must be copied in explicitly; without
+# it crispritz silently falls back to the ~15x slower pure-Python enricher.py.
+cp enricher "${PREFIX}/bin/enricher"
```
bgruening, line 23: why can this not be a symlink?

lucapinello: Thanks @bgruening! Good call on avoiding a duplicate binary.

Line 21 (`$PREFIX/bin/enricher`) does have to be a real copy: `enricher` is compiled by `Makefile_conda` in the (ephemeral) build directory at the repo root, so there's no persistent target for `$PREFIX/bin/enricher` to symlink to — it has to be installed as a copy.

The **second** install, line 22 (`.../Python_Scripts/Enrichment/enricher`), is a duplicate of that same binary and can absolutely be a symlink. `crispritz.py`'s `_enricher_command()` looks for the binary both on PATH and beside the bundled `Python_Scripts/Enrichment/`, so it needs to be present there — but a relative symlink to `$PREFIX/bin/enricher` satisfies that without shipping the binary twice. I'll switch line 22 to:

```sh
ln -s ../../../../bin/enricher "${PREFIX}/opt/crispritz/Python_Scripts/Enrichment/enricher"
```

so the package ships a single enricher binary. Pushing that shortly.

lucapinello: Done — the second install is now a relative symlink (`ln -sf ../../../../bin/enricher $PREFIX/opt/crispritz/Python_Scripts/Enrichment/enricher`), so the package ships a single `enricher` binary. All checks are green (Lint, Linux, OSX-64, ARM). I also bumped the build number to 1 since a 2.8.1/build 0 was already in the channel. Thanks for the review!

lucapinello:
Closing: crispritz 2.8.1 already landed via #67999, and 2.8.2 (a one-line fix so the -index search ignores hidden metadata sidecars like a .pam_build/.display_label file) is bumping via the autobump PR #68147. Superseded.

PR #68223, opened 2026-08-17
Update tesseract-assembler to 1.2.4

iowa69:
Follow-up to #68032, which merged at 1.2.2.

That version installs commands named `tessera` and `tessera-model`. Upstream has since renamed everything to match the project's own name — [TesserACT](https://github.com/iowa69/TesserACT) — so those commands no longer exist. They are now `tesseract-asm` and `tesseract-model`.

Deliberately **not** plain `tesseract`: that binary belongs to the OCR engine in conda-forge, and a second package installing `bin/tesseract` would be a file conflict.

This also adds **`tesseract-klebsiella`**, new upstream. The previous package was the two binaries alone, which left a user to find the release page, download a 339 MB model by hand and work out the flags. The runner locates the model, verifies its checksum against a pinned SHA-256 and assembles.

### Testing

Built from the 1.2.4 release tarball with the `c_stdlib` pin CI supplies:

- package test passes
- the functional test assembles a 6 kb genome and asserts a single contig of the right length — a binary that installs but cannot assemble would pass a `--version` check
- help output is checked for the flags a pipeline may depend on
- installing the built package into a clean environment yields all three commands, reporting `TesserACT 1.2.4`

Linux-64 only here; I will follow whatever your CI reports for osx.

### DIFF
--- a/recipes/tesseract-assembler/build.sh
+++ b/recipes/tesseract-assembler/build.sh
@@ -6,5 +6,15 @@ set -euo pipefail
 make -j"${CPU_COUNT:-4}" CXX="${CXX:-g++}"
 
 mkdir -p "${PREFIX}/bin"
-install -m 0755 tessera "${PREFIX}/bin/tessera"
-install -m 0755 tessera-model "${PREFIX}/bin/tessera-model"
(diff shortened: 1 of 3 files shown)


mencian:
Closing in favor of #68226.

PR #68253, opened 2026-08-18
Add cfm-id 4.4.10

RJMW:
Adds CFM-ID 4 as a new recipe rather than an update to `cfm`.

`recipes/cfm` is at version 33, which is the CFM-ID 3 generation. Both install binaries called `cfm-id`, `cfm-predict` and `fraggraph-gen`, so they cannot share an environment. This recipe sets `run_constrained: cfm <0a0` so the solver rejects the combination up front instead of it failing later as clobbered files. Existing `cfm` users are unaffected. Bumping `cfm` in place would also mean a version decrease, since CFM-ID renumbered from 33 to 4.4.10.

The source is pinned to the commit behind the `CFM-ID_4.4.10` tag, because Bitbucket's generated tag archives are not byte-stable.

Three patches are needed to build against a current RDKit. One makes `CMAKE_CXX_STANDARD` overridable from the command line so `build.sh` can ask for C++17, which RDKit's headers now require. One drops a `.get()` call, since `ROMol::operator[]` returns a raw `Bond *` in current RDKit. One adds a missing `<cfloat>` include. Each patch header says what it does and when it can be dropped. They are not reported upstream yet: Bitbucket is removing Issues and the repository has no contact address, so I am contacting the authors directly.

### DIFF
--- a/recipes/cfm-id/add-cfloat-include.patch
+++ b/recipes/cfm-id/add-cfloat-include.patch
@@ -1,0 +1,40 @@
+Include <cfloat> where DBL_MAX is used.
+
+EmModel.cpp and cfm-train/main.cpp use DBL_MAX but never include <cfloat>.
+They compiled historically because older libstdc++ headers pulled it in
+transitively; current libstdc++ does not, so the build fails:
+
+    EmModel.cpp: error: 'DBL_MAX' was not declared in this scope
+
+This is unrelated to RDKit -- it is ordinary transitive-include tightening, and
+the fix is the include that should always have been there.
+
+Upstream status: not reported, because there is currently no channel to report to.
+Atlassian is sunsetting Bitbucket Issues and Wikis across Cloud; they could no
+longer be enabled from April 2026 and existing issues are removed on 20 August
+2026, so wishartlab/cfm-id-code has no usable tracker. The repository README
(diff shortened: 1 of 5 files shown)


RJMW:
@BiocondaBot please add label

bgruening (commented):
@RJMW could you maybe clean please the over-verbose LLM stuff? Thanks!

RJMW:
@bgruening , I cleaned up the comments throughout the YAML as suggested.

bgruening:
But not committed? Is anyone still here, or are those just bots running around :)?

RJMW:
@bgruening Sorry for the mix-up. I had another recipe open and commented on the wrong PR. It’s actually done now. As requested, I’ve removed or trimmed the comments throughout.

bgruening (approved):

bgruening:
Thanks!

RJMW:
@bgruening I noticed cfm-id is only published for linux-64. The osx-64 job was still queued I believe when this merged.

The previous commit did build successfully and created an artefact that was built and tested (containing cfm-id-4.4.10-h9fc8b65_0.conda). Any idea?

Can that artefact be uploaded, or would you rather I open a PR bumping the build number? Or what would you recommend to fix this?

PR #68258, opened 2026-08-18
Update oxbreaker to 1.2.1

rc-reding:
Internal updates to oxbreaker to support macOS.

* Updated .desktop (shortcut file for linux when using `oxbreaker-desktop-helper`)
* Updated `oxbreaker-desktop-helper` to better handle the .app structure in macOS and install shortcut in /Applications so it shows up in finder.


rc-reding:
@BiocondaBot please add label

PR #68301, opened 2026-08-19
Update gtdb-genomes to 0.3.0

asuq:
Supersedes #68294.

Update `gtdb-genomes` to 0.3.0 using the verified published sdist checksum. Unlike the automated version bump, this also removes the obsolete `unzip` runtime dependency because upstream now uses Python's standard library for ZIP extraction.

Validation:

- `bioconda-utils lint` passed.
- The noarch package built locally and all recipe tests passed on Python 3.14.
- Upstream wheel and sdist release validation passed.

Thank you for reviewing this update.


asuq:
@BiocondaBot please add label

kbseah (approved):
Thanks for the update!

PR #68319, opened 2026-08-20
Update sirius-csifingerid to 6.3.12

RJMW:
Updates `sirius-csifingerid` from 5.8.6 to 6.3.12 — a major-series jump, so this is more than a version bump.

### Built and tested on all three platforms locally

The recipe deliberately carries **no platform `skip`**, even though every published build (5.8.2–5.8.6) is linux-64. That looks like macOS is unsupported; it is not. All three were built from this recipe before opening the PR, and every test assertion passed on each:

| platform | result | gradle target |
|---|---|---|
| linux-64 | pass | `LINUX_x86_64` |
| osx-arm64 | pass | `MAC_ARM64` |
| osx-64 | pass | `MAC_x86_64` |

`build.sh` passes conda's `$target_platform` into `-P build.sirius.platform`, and `SiriusPlatform.fromDescriptor()` substring-matches it, so `osx-64` resolves to `MAC_x86_64` rather than silently falling back to the host. So `OSX-64 Tests` going green here is the tested expectation, not a hope.

### Changes

- **Source org** `boecker-lab` → `sirius-ms`. The old org is stale.
- **openjdk 17 → 25.** No compile toolchain is declared anywhere in the Gradle build, so the build JDK is the packager's choice. `sirius_dist/build.gradle` sets `jre_version = '25.0.3'` for the JRE upstream bundles for end users, which is the best available signal.
- **Tests moved from `run_test.sh` into `test.commands`**, and `Kaempferol.ms` removed with the script that used it. The 5.8.6 test used SIRIUS 5 syntax and asserted on a fragmentation-tree directory that no longer exists under the opaque 6.x project space.

  It cannot be replaced with an equivalent SIRIUS 6 analysis: **every analysis subcommand now requires a logged-in account**, including ones the CLI labels `<STANDALONE>` — `sirius decomp -m 180.0634`, a pure mass decomposition needing no network, still answers `Please Login`. The tests therefore assert on the CLI surface only.

  Every assertion checks **output**, not exit status, because **SIRIUS exits 0 when it fails** — verified for both a login refusal and a fatal argument error, which prints a stack trace and still returns 0. A bare `sirius ... > /dev/null` would pass against a completely broken package.

- **Licence corrected** `AGPL-3.0-only` → `AGPL-3.0-or-later`. `COPYING.txt` grants the software under the AGPL *"either version 3 of the License, or (at your option) any later version"*, so `-only` asserted the opposite. This is an error inherited from the 5.8.6 recipe, not a new one.

- **Adds `recipe-maintainers`**, which this recipe did not previously have.

### One thing worth flagging for review

`AND OTHER` is retained, and it is **not a valid SPDX identifier** — SPDX expresses an unlisted licence as `LicenseRef-`. I have kept it because it is what the recipe already said and because it is doing real work: it covers both the third-party dependency inventory in `licenses.md` (2,326 lines — MIT, LGPL-3.0, BSD-2/3-Clause, EPL-2.0, CDDL, OpenJDK's GPLv2+CE, MPL) **and SIRIUS's own submodules**, 51 of which ship their own `LICENSE.txt`; of the 40 sampled, 38 are LGPL-3.0, one LGPL-2.1 and one GPL-3.0. `COPYING.txt` also carries four AGPL §7 additional-permission exceptions.

Happy to change it to something SPDX-valid if that is preferred — it seemed a separate question from correcting a demonstrably wrong version suffix.


RJMW:
@BiocondaBot please add label

mencian (approved):

PR #68320, opened 2026-08-20
Add molnetpack 1.4.0

RJMW:
New recipe for **3DMolMS** — a deep neural network predicting MS/MS spectra from three-dimensional molecular conformations, with qTOF and Orbitrap models, plus retention-time and collision-cross-section prediction.

- Upstream: https://github.com/JosieHong/3DMolMS
- Paper: [doi:10.1093/bioinformatics/btad354](https://doi.org/10.1093/bioinformatics/btad354)

### On the package name

The software is called **3DMolMS**; **molnetpack** is its PyPI distribution and import name. Upstream's own install instruction reads *"3DMolMS is available on PyPI: `pip install molnetpack`"*, so the recipe is named for the distribution — the package ships no console script, so it is a library and `conda install` / `import` / error messages all agree. "3DMolMS" is carried in `summary`, `home`, `doc_url` and `description` for discoverability.

### Built and tested

- **linux-64**, in `quay.io/bioconda/bioconda-utils-build-env-cos7` under `--platform linux/amd64`
- **osx-arm64**

Both test commands ran on each; `MolNet(torch.device('cpu'), seed=42)` reports `MolNetPack version: 1.4.0`, which also confirms the declared dependency set is sufficient to import and instantiate. `bioconda-utils lint` (4.4.1) clean.

The tests deliberately stop at instantiation: `pred_*` methods **download checkpoints on demand**, so reaching one would put a network fetch inside the build.

### One deliberate departure from upstream metadata

`scikit-learn` is declared as a runtime dependency although 1.4.0 moved it to an optional `train` extra.

It is reachable through exactly one path — `fit_scaler()`, called from `MolNet.train(task="rt", use_scaler=True)` — and is imported lazily inside that function, so prediction genuinely does not need it. It is declared anyway so that retention-time training works from the installed package. Cost is 5 packages (`scikit-learn`, `scipy`, `joblib`, `threadpoolctl`, `narwhals`), negligible beside pytorch.

Flagging it because it looks like an error against `pyproject.toml` and would otherwise be a reasonable thing for a reviewer to ask me to remove.

### Licence

**CC-BY-NC-SA-4.0** — not an OSI-approved licence, and non-commercial. Carried accurately in the recipe rather than glossed.

### DIFF
--- a/recipes/molnetpack/meta.yaml
+++ b/recipes/molnetpack/meta.yaml
@@ -36,8 +36,6 @@ requirements:
     - platformdirs
     - requests
     - tqdm
-    # An optional `train` extra upstream, imported lazily in fit_scaler.
-    # Declared so MolNet.train(task="rt", use_scaler=True) works as installed.
     - scikit-learn
 
 test:
@@ -45,7 +43,6 @@ test:
     - molnetpack
   commands:
     - python -c "from molnetpack import MolNet; print('molnetpack import OK')"
-    # Checkpoints download on demand, so tests must not reach a pred_* method.
     - python -c "import torch; from molnetpack import MolNet; MolNet(torch.device('cpu'), seed=42)"
 
 about:
--- a/recipes/molnetpack/meta.yaml
+++ b/recipes/molnetpack/meta.yaml
@@ -1,0 +1,80 @@
+{% set name = "molnetpack" %}
+{% set version = "1.4.0" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 43810265d2a16637b12db32b4ced9d051890032edaa372ddb7a559b8325e7d40
+
+build:
+  number: 0
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  run_exports:
+    - {{ pin_subpackage(name, max_pin="x.x") }}
+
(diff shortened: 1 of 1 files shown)


RJMW:
@BiocondaBot please add label

recipes/molnetpack/meta.yaml
```diff
@@ -0,0 +1,77 @@
...
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 43810265d2a16637b12db32b4ced9d051890032edaa372ddb7a559b8325e7d40
+
+build:
+  number: 0
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  run_exports:
+    - {{ pin_subpackage(name, max_pin="x.x") }}
```
thanhleviet, line 17: Optional: for a 1.x version the usual bioconda convention is `max_pin="x"`. If the stricter `x.x` pin is deliberate because checkpoints are minor-version-specific (as the description suggests), this is fine as is - just confirming it's intentional rather than a leftover from a 0.x draft.

thanhleviet (approved):
Verified: sha256 matches the PyPI sdist, dependencies match pyproject.toml (with the correct pytorch/matplotlib-base substitutions), the scikit-learn addition checks out as described (lazy import in model.py, training path only), the LICENSE file matches the CC-BY-NC-SA-4.0 identifier, and CI is green on all platforms. Thanks for the thorough PR description - it answered every question in advance. One optional question inline; fine to merge as is.

PR #68360, opened 2026-08-21
Update mrsfast: paired-end segfault fix

descostesn:
Adds a one-line fix for a paired-end segmentation fault as a patch on top of upstream 3.4.2, and bumps the build number to 6. Version and source are unchanged.

## The bug

`initRead()` in `Reads.c` allocates the buffer-2 read counters in the paired-end branch but never initialises them:

```c
_r_buf1      = getMem(10000000);
_r_buf1_pos  = getMem(sizeof(int));
_r_buf1_size = getMem(sizeof(int));
*_r_buf1_size = *_r_buf1_pos = 0;        /* buffer 1: initialised   */
if ( pairedEndMode && fileName2 != NULL )
{
    _r_buf2      = getMem(10000000);
    _r_buf2_pos  = getMem(sizeof(int));
    _r_buf2_size = getMem(sizeof(int));  /* buffer 2: never zeroed  */
}
else
{
    _r_buf2      = _r_buf1;              /* single-end aliases buf 1 */
    _r_buf2_pos  = _r_buf1_pos;
    _r_buf2_size = _r_buf1_size;
}
```

Both counters are read before they are written, so paired-end runs operate on uninitialised heap memory. Single-end runs are unaffected, because that branch aliases the buffer-1 pointers, which are initialised on the line immediately above.

`paired-end-buffer-init.patch` adds the missing line, restoring the symmetry the file already has for buffer 1:

```c
*_r_buf2_size = *_r_buf2_pos = 0;
```

## Upstream

Offered upstream at https://github.com/sfu-compbio/mrsfast/pull/21. The repository's most recent commit is 2020-06-10 — the same commit the `v3.4.2` tag points at — and the PR has had no maintainer response, so I am submitting here in the meantime per the [patching guidelines](https://bioconda.github.io/contributor/guidelines.html#patching). If upstream merges it, this patch can be dropped in favour of a version bump.

## Testing

- The patch applies cleanly with `patch -p1` to the `v3.4.2` archive already referenced by the recipe (sha256 unchanged and verified).
- Built with `bioconda-utils build recipes config.yml --mulled-test --packages mrsfast`
- Functional check on several paired-end ChIP-seq data using storage on [anaconda](https://anaconda.org/channels/descostesn/packages/mrsfast/overview)
- Single-end ChIP-seq does not trigger the segfault 

## Notes

- `run_exports` added as requested. I have used `max_pin="x.x"` rather than `"x"`. To be explicit, this is **not** a "known breakage in minor versions" claim — I have no evidence of that, and upstream has published no release since 2020. It is simply the conservative choice for a project that documents no versioning policy. mrsFAST ships only executables (`mrsfast`, `snp_indexer`) and nothing links against it, so the practical difference is small; happy to switch to `max_pin="x"` if you prefer the table's default for a 3.x version.

### DIFF
--- a/recipes/mrsfast/meta.yaml
+++ b/recipes/mrsfast/meta.yaml
@@ -8,10 +8,14 @@ package:
 source:
   url: https://github.com/sfu-compbio/mrsfast/archive/v{{ version }}.zip
   sha256: {{ sha256 }}
-  patches: Makefile.patch
+  patches:
+    - Makefile.patch
+    - paired-end-buffer-init.patch
 
 build:
-  number: 5
+  number: 6
+  run_exports:
+    - {{ pin_subpackage('mrsfast', max_pin='x.x') }}
 
 requirements:
   build:
--- a/recipes/mrsfast/meta.yaml
+++ b/recipes/mrsfast/meta.yaml
@@ -20,6 +20,7 @@ build:
 requirements:
   build:
     - {{ compiler('c') }}
+    - {{ stdlib('c') }}
     - make
     - zlib
     - groff
--- a/recipes/mrsfast/meta.yaml
+++ b/recipes/mrsfast/meta.yaml
@@ -14,6 +14,8 @@ source:
 
 build:
   number: 6
+  # work/conda_build.sh: line 13:  5999 Segmentation fault: 11  ./mrsfast --help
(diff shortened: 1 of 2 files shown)


mencian (approved):

descostesn:
Thank you!

PR #68386, opened 2026-08-22
Update bioleads to 0.2.0

dangerisom:
Version bump for [isomlab/bioleads v0.2.0](https://github.com/isomlab/bioleads/releases/tag/v0.2.0). I am the package author and the recipe maintainer.

Notable change in this release: term clustering is now density-based (HDBSCAN infers the cluster count) instead of KMeans with a fixed k. No new dependencies — `scikit-learn >=1.3`, already pinned, is where `sklearn.cluster.HDBSCAN` lives.

This also drops `pymupdf`, which backed a `pdf` extra that no longer exists upstream: the package has no `--pdf` flag and no PyMuPDF import, so the recipe has been carrying an unused runtime dependency.

Build number reset to 0 for the new version.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/recipes/bioleads/meta.yaml
+++ b/recipes/bioleads/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "bioleads" %}
-{% set version = "0.1.0" %}
+{% set version = "0.2.0" %}
 
 package:
   name: {{ name|lower }}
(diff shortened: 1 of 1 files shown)


dangerisom:
@BiocondaBot please add label

PR #68439, opened 2026-08-24
Update ploidypatch to 1.0.1

707728642li:
PloidyPatch 1.0.1 is a metadata-only patch release.

- records Taishan Li and Gaigai Du consistently as software authors
- updates the public installation description now that Bioconda 1.0.0 is available
- does not change scientific methods, command behavior, dependencies, or the recipe build number

The source SHA-256 is bound to the immutable v1.0.1 GitHub release. The recipe was rendered, built, installed, and tested locally; `ploidypatch --version`, command help, Python import, and `pip check` all passed.

### DIFF
--- a/recipes/ploidypatch/meta.yaml
+++ b/recipes/ploidypatch/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "ploidypatch" %}
-{% set version = "1.0.0" %}
-{% set release_tag = "v1.0" %}
+{% set version = "1.0.1" %}
+{% set release_tag = "v1.0.1" %}
 
 package:
   name: {{ name }}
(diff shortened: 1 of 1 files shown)


707728642li:
@BiocondaBot please add label

recipes/ploidypatch/meta.yaml
```diff
@@ -34,7 +34,7 @@ test:
     - ploidypatch --version
     - ploidypatch --help
     - ploidypatch patch --help
-    - python -c "import ploidypatch; assert ploidypatch.__version__ == '1.0.0'"
+    - python -c "import ploidypatch; assert ploidypatch.__version__ == '1.0.1'"
```
kbseah, line 37: Could consider using the jinja variable here in the future

kbseah (approved):
Thanks for the update!
A small suggestion below

PR #68483, opened 2026-08-24
Add actinn-jax

iandriver:
Adds a recipe for [actinn-jax](https://github.com/iandriver/actinn-jax), a JAX
reimplementation of ACTINN for cell-type annotation of single-cell RNA-seq data. It trains
a classifier from a labelled AnnData reference and annotates query AnnData objects on CPU,
with a pre-trained human reference included in the package.

- Pure Python, so `noarch: python`.
- All run dependencies are available on conda-forge.
- Source is the PyPI sdist; sha256 verified against a local download.
- The sdist contains `LICENSE` and the bundled reference data, so `actinn_jax.bundled_reference()`
  works from a conda install without a network fetch.
- `run_exports` uses `max_pin="x.x"` per the 0.x.x row of the table in the PR template.

I am the author and maintainer of the package.

### DIFF
--- a/recipes/actinn-jax/meta.yaml
+++ b/recipes/actinn-jax/meta.yaml
@@ -1,0 +1,55 @@
+{% set name = "actinn-jax" %}
+{% set version = "0.3.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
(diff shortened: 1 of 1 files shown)


iandriver:
@BiocondaBot please add label

iandriver:
@BiocondaBot please add label

PR #68487, opened 2026-08-24
Add pyeuk 0.7.0

nekrut:
Adds a new recipe for **PyEuk** ([github.com/spond/pyeuk](https://github.com/spond/pyeuk)), a Python framework for molecular typing, genetic distance estimation, and foodborne/waterborne outbreak cluster detection in eukaryotic and microbial pathogens — *Cyclospora cayetanensis*, *Cryptosporidium parvum/hominis*, and general MLST/cgMLST schemes.

It provides reference-free de novo locus and haplotype discovery from assembled contigs, a KING-weighted identity-by-state (wIBS) distance engine with pairwise-complete dropout tolerance and PSD Gram-matrix projection, and unsupervised Ward hierarchical clustering with automatic knee-based selection of the cluster count.

### Recipe notes

- `noarch: python`, pure Python; runtime deps are `numpy`, `scipy`, `pandas`, `scikit-learn`, and `numba`, all already in conda-forge.
- Two console scripts are installed, `pyeuk` and `cyclospora-typing`; both are exercised in the `test:` section along with the module imports.
- `run_exports` uses `max_pin="x.x"` per the 0.x semantic-versioning row in the contributor guidelines.
- Source is the upstream `v0.3.0` tag tarball. Upstream added `LICENSE` (Apache-2.0) and a PEP 621 `pyproject.toml` in [spond/pyeuk#9](https://github.com/spond/pyeuk/pull/9) specifically so this recipe has a checksummable tarball with a `license_file` inside it.
- SPAdes is *optional* at runtime — `micro_assembly.py` looks it up with `shutil.which` and degrades gracefully when it is absent — so it is deliberately not a run dependency. Happy to add `spades` if reviewers would rather have that path work out of the box.

### Verification

Against the exact `v0.3.0` tarball this recipe pins (sha256 `8eccd6e4…b3e6`), in a clean environment:

```
$ pip install ./pyeuk-0.3.0
$ pyeuk --help              # OK
$ cyclospora-typing --help  # OK
$ python -c "import cyclospora_pyeuk, cyclospora_pyeuk.distance_engine, cyclospora_pyeuk.clustering"
imports OK 0.3.0
$ python -m pytest tests -q
27 passed
```

### DIFF
--- a/recipes/pyeuk/meta.yaml
+++ b/recipes/pyeuk/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "pyeuk" %}
-{% set version = "0.5.0" %}
+{% set version = "0.6.0" %}
 
 package:
   name: {{ name }}
@@ -7,14 +7,14 @@ package:
 
 source:
   url: https://github.com/spond/pyeuk/archive/refs/tags/v{{ version }}.tar.gz
-  sha256: 3bb4f26e8fb4a1ec42c3a34822022a6ccbae5da582911aae90c706dd7a336e58
+  sha256: 74dee55ed9dbcd6ec5569dfc25c9e5ce67e6befebbe7e5a6a87668dec7c7781a
 
 build:
   number: 0
   noarch: python
   entry_points:
-    - pyeuk=cyclospora_pyeuk.cli:main
-    - cyclospora-typing=cyclospora_pyeuk.cli:main
+    - pyeuk=pyeuk.cli:main
+    - cyclospora-typing=pyeuk.cli:main
   script: {{ PYTHON }} -m pip install . --no-deps --no-build-isolation --no-cache-dir -vvv
   run_exports:
     - {{ pin_subpackage("pyeuk", max_pin="x.x") }}
@@ -35,11 +35,12 @@ requirements:
 
 test:
   imports:
+    - pyeuk
+    - pyeuk.distance_engine
+    - pyeuk.clustering
+    - pyeuk.naming
+    - pyeuk.amplicon
     - cyclospora_pyeuk
-    - cyclospora_pyeuk.distance_engine
-    - cyclospora_pyeuk.clustering
-    - cyclospora_pyeuk.naming
-    - cyclospora_pyeuk.amplicon
   commands:
     - pyeuk --help
     - cyclospora-typing --help
--- a/recipes/pyeuk/meta.yaml
+++ b/recipes/pyeuk/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "pyeuk" %}
-{% set version = "0.6.0" %}
+{% set version = "0.7.0" %}
 
 package:
   name: {{ name }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://github.com/spond/pyeuk/archive/refs/tags/v{{ version }}.tar.gz
-  sha256: 74dee55ed9dbcd6ec5569dfc25c9e5ce67e6befebbe7e5a6a87668dec7c7781a
+  sha256: 4184570bdff8622f9ad8906940547932e64b21753745c72764202c49fcdf7275
 
 build:
   number: 0
@@ -32,6 +32,7 @@ requirements:
     - scikit-learn >=1.0.0
     - numba >=0.53.0
     - pysam >=0.22
+    - pillow >=9
 
 test:
   imports:
@@ -40,6 +41,7 @@ test:
     - pyeuk.clustering
     - pyeuk.naming
     - pyeuk.amplicon
+    - pyeuk.report
     - cyclospora_pyeuk
   commands:
     - pyeuk --help
@@ -47,6 +49,7 @@ test:
     - pyeuk define-windows --help
     - pyeuk call-haplotypes --help
     - pyeuk build-sheet --help
+    - pyeuk report --help
 
 about:
   home: https://github.com/spond/pyeuk
--- a/recipes/pyeuk/meta.yaml
+++ b/recipes/pyeuk/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "pyeuk" %}
-{% set version = "0.4.0" %}
+{% set version = "0.5.0" %}
 
 package:
   name: {{ name }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://github.com/spond/pyeuk/archive/refs/tags/v{{ version }}.tar.gz
-  sha256: e730e6d3a906971e20dbccf8670b236eabbef7fb57e49392a3a770aeea8aa1b2
+  sha256: 3bb4f26e8fb4a1ec42c3a34822022a6ccbae5da582911aae90c706dd7a336e58
 
 build:
   number: 0
@@ -31,15 +31,21 @@ requirements:
     - pandas >=1.3.0
     - scikit-learn >=1.0.0
     - numba >=0.53.0
+    - pysam >=0.22
 
 test:
   imports:
     - cyclospora_pyeuk
     - cyclospora_pyeuk.distance_engine
     - cyclospora_pyeuk.clustering
+    - cyclospora_pyeuk.naming
+    - cyclospora_pyeuk.amplicon
   commands:
     - pyeuk --help
     - cyclospora-typing --help
+    - pyeuk define-windows --help
+    - pyeuk call-haplotypes --help
+    - pyeuk build-sheet --help
 
 about:
   home: https://github.com/spond/pyeuk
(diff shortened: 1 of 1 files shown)


nekrut:
@BiocondaBot please add label

nekrut:
Flagging this before it merges: the recipe is correct, but the **tag it points at predates a substantive upstream fix** and I think this should wait for a re-tag rather than ship 0.3.0 and supersede it immediately.

`v0.3.0` is commit `43d2ff8`. The upstream fix in [`1f61e15`](https://github.com/spond/pyeuk/commit/1f61e15) ("Fix wIBS weighting inversion for presence/absence sheets", resolving spond/pyeuk#11) landed after it. Checked against the exact tarball this recipe hashes — the sha256 still matches, so the tag has not moved:

```
weight_mode occurrences in tagged distance_engine.py: 0
line 193:  w_j[k] = 1.0 / np.sqrt(p_clip * (1.0 - p_clip))
line 286:  w_j    = 1.0 / np.sqrt(np.maximum(p_j * (1.0 - p_j), 1e-6))
```

That weight is minimised at p=0.5 and diverges as p→0, so on presence/absence input it weights rare alleles hardest while the discriminating signal sits in balanced columns. On a 153-specimen *Cyclospora* sheet it scored ARI −0.0057 against epidemiological labels; the fixed default (`weight_mode="heterozygosity"`) scores 0.95–1.00 on the same sheet. The tagged code has no `weight_mode`, no `min_maf` and no `project_psd` toggle, so there is no way for a user to opt out of the old behaviour.

Packaging 0.3.0 would make that the version most people install.

Two smaller things in the same recipe, both downstream of the same point:

- The description says "a KING-weighted identity-by-state (wIBS) distance engine". That stopped being the default at `1f61e15` and would need rewording once the version bumps.
- `run_exports: pin_subpackage(max_pin="x.x")` pins downstream to 0.3.x, which is right if the re-tag is 0.3.1 and wrong if upstream cuts 0.4.0 instead.

Suggested path: ask upstream to tag `1f61e15` or later, then bump `version`, `url` and `sha256` here and reword the description. Everything else in the recipe — entry points, `noarch: python`, the host/run split, the import and `--help` tests — looks right to me as written and shouldn't need to change.

Happy to push the bump to this branch as soon as a tag exists.

nekrut:
Status update on the version question above: I've opened [spond/pyeuk#13](https://github.com/spond/pyeuk/pull/13) upstream to bump the version and add a changelog, so a `v0.4.0` tag can be cut.

**0.4.0 rather than 0.3.1**, which affects this recipe in two places. The distance engine's default weighting changed (`weight_mode` now defaults to heterozygosity rather than the KING form), so identical input yields different distances than `0.3.0`; and four new public arguments/attributes were added. That is a minor bump under semver, not a patch. It also means `run_exports: pin_subpackage(max_pin="x.x")` stays correct as written — it will pin downstream to 0.4.x, which is what you want given the numerical change.

Worth noting that `master` upstream still declares `0.3.0` in `pyproject.toml` and `__init__.py`, so tagging today without that PR would have produced a `v0.4.0` tarball whose installed metadata reports `0.3.0`. Good thing to have caught before the recipe pinned a sha256 to it.

When the tag lands I'll push three changes here — `version` to `0.4.0`, the `url` accordingly, and the recomputed `sha256` — plus a reworded `summary`/`description`, since the current text advertises "a KING-weighted identity-by-state (wIBS) distance engine" and that is no longer the default.

Nothing needed from reviewers until then; leaving this open rather than closing so the history stays in one place.

nekrut:
Updated to **0.4.0** — this is now ready for review.

Upstream cut [`v0.4.0`](https://github.com/spond/pyeuk/releases/tag/v0.4.0) (commit `7d33516`, on `master`) with the version bump landed as `50b53e2`, so the four commits that had accumulated past `v0.3.0` are now in a tagged release.

Changes in this push:

- `version` → `0.4.0`, `sha256` → `e730e6d3a906971e20dbccf8670b236eabbef7fb57e49392a3a770aeea8aa1b2`
- Description no longer says "KING-weighted"; heterozygosity is the default weighting now and the KING form is one selectable option among several

`run_exports: max_pin="x.x"` is unchanged and still correct — 0.4.0 is a minor rather than patch bump precisely because the default numerical behaviour changed, so pinning downstream to 0.4.x is the intended effect.

Checked against the tagged tarball before pushing, since the previous tag had the metadata mismatch that started this:

```
pyproject.toml:7           version = "0.4.0"
__init__.py:17             __version__ = "0.4.0"
imports OK, __version__ =  0.4.0
entry point cyclospora_pyeuk.cli:main resolves OK
```

Both `[project.scripts]` entries (`pyeuk`, `cyclospora-typing`) point at `cyclospora_pyeuk.cli:main`, and all three imports in the test section resolve, so the test block should pass as written.

nekrut:
Bumped to **0.5.0** and added `pysam` as a run dependency. Same branch, no new PR.

**Why the bump.** Upstream tagged v0.5.0 after this PR was opened. 0.4.0 does not contain the
amplicon front end at all — a `conda install` of it would give a `pyeuk --help` with no
`define-windows`, `call-haplotypes` or `build-sheet`. Since the recipe fixes whichever version
most people end up installing, it seemed better to pin the one that is complete.

**Why `pysam` is a run dependency rather than left out.** Upstream declares it as the
`[amplicon]` extra, which is correct for pip. Conda has no extras, so without listing it the
three amplicon subcommands fail on every conda install and there is no conda-level way to fix
that. Upstream agreed (spond/pyeuk#17). It builds on all four platforms tested here: linux-64,
osx-64, linux-aarch64, osx-arm64.

This is a different case from SPAdes, which I deliberately left out and noted in the PR body:
that one is resolved through `shutil.which` and degrades gracefully when absent. `pysam` is a
hard import — the step simply cannot run.

**Verified before pinning**, against the v0.5.0 tarball:

- sha256 `3bb4f26e…a336e58` matches the tag tarball
- `pyproject.toml` and `cyclospora_pyeuk/__init__.py` both declare `0.5.0`; `LICENSE` present
- `packages` includes `cyclospora_pyeuk.amplicon`, so the new subpackage actually installs
- clean venv, `pip install .[amplicon]` → `pytest tests -q` = **51 passed**
- all three new entry points respond to `--help`

The test block now imports `cyclospora_pyeuk.amplicon` and exercises the three subcommands, so
a build that silently dropped the subpackage from `packages` would fail here instead of
shipping broken.

Nothing else changed. Still one file, still `noarch: python`, still sourcing the canonical
`spond/pyeuk` tarball.

nekrut:
**Please hold this one — do not merge yet.** Apologies for the moving target; I would rather flag this before publication than break users afterwards.

The conda package name `pyeuk` is settled and correct. The **import** package is
`cyclospora_pyeuk`, and that is what the test block imports and what users will write in their
scripts.

That name is now wrong. The package began as a *Cyclospora cayetanensis* typing tool, but it has
since been validated on *Plasmodium vivax* (three separate panels), *Teladorsagia circumcincta*
and *Cryptosporidium*, and 0.5.0 adds a general amplicon front end. Upstream is renaming the
import package to `pyeuk`, keeping `cyclospora_pyeuk` as a deprecation shim so nothing breaks.

Once this recipe is published, the import name is what users will have written into their code.
Changing it after that is a breaking change for everyone who installed early; changing it before
costs nothing.

So I would like to bump this to the renamed release rather than publish 0.5.0 and immediately
supersede it. Tracking upstream at spond/pyeuk. I will post here with the new version, URL and
sha256 as soon as the tag exists, on this same branch — no new PR.

Everything else stands: green on Linux, OSX-64 and ARM at head `93d9d5bf`, single file, `noarch:
python`, sourcing the canonical `spond/pyeuk` tarball.

Sorry for the extra round trip, and thank you for your patience.

nekrut:
**Hold lifted — this is ready for review.** Bumped to **0.6.0**, same branch.

The rename I flagged has landed upstream (spond/pyeuk#18). The import package is now `pyeuk`,
matching the conda package name, with `cyclospora_pyeuk` kept as a deprecation shim so nothing
that already imports the old name breaks.

This was the release worth waiting for. A user installing from this recipe now writes `import
pyeuk`, which is both the package they installed and a name that does not claim one organism
while the tool works on several.

Recipe changes are the version, the sha256, the entry points (they now resolve `pyeuk.cli`), and
the test block. The test block imports the new module names **and** `cyclospora_pyeuk`, so if the
shim is ever removed without a major version bump, this recipe catches it rather than a user.

Verified against the v0.6.0 tarball in a clean virtualenv:

- sha256 `74dee55e…c7781a` matches the tag tarball
- `pyproject.toml` and `pyeuk/__init__.py` both declare `0.6.0`; `LICENSE` present
- `packages` lists `pyeuk`, `pyeuk.amplicon`, and both shim packages
- all five console entry points respond to `--help`
- the shim imports, reports the right version, and raises `DeprecationWarning`
- `pytest tests -q` → **54 passed**

Unchanged from before: one file, `noarch: python`, canonical `spond/pyeuk` tarball, and `pysam`
as a run dependency because conda has no extras.

Thanks for your patience through the version churn — I would rather have published the right
name once than the wrong one twice.

nekrut:
Gentle bump — this one is settled and ready whenever a reviewer has time. No changes since the 0.6.0 push on 27 Aug: single file, `noarch: python`, green on Linux, OSX-64 and ARM. The earlier version churn is finished; I am not planning further pushes to this branch. Happy to answer anything or adjust the recipe if it would help. Thanks!

nekrut:
Bumped the recipe to **pyeuk 0.7.0** (spond/pyeuk was just tagged `v0.7.0`): version + sha256 updated (source is the GitHub tag tarball — pyeuk is not on PyPI). Added `pillow >=9` to run deps and a `pyeuk.report` import / `pyeuk report --help` test for the new graphical-report command. 0.7.0 adds the sweep-default `cluster` output, ~10× amplicon windowing, a lossless sub-quadratic haplotype denoise, and `pyeuk report`.

bgruening (approved):

bgruening:
Talking to a human would be nice ;-), otherwise looks good.

